PR #305, opened 2019-08-22
Have unversioned views hit versioned views for parquet unnesting

jklukas:
flod reported in `#fx-metrics` that submission_date_s3 was missing
for the clients_daily view. That view and other unversioned views
were directly referencing telemetry_derived tables and missing out
on transformations such as parquet unnesting.


jklukas:
I went ahead and deployed these changes, but will wait for review to merge.

jklukas:
Actually, considering it looks like we could have view auto-deploys sooner than later (https://github.com/mozilla-services/mozilla-pipeline-schemas/issues/390), I'm going to merge this now and ask for retroactive review so we don't accidentally revert the views.

PR #308, opened 2019-08-23
Make copy_deduplicate query more efficient for large tables

### LINKED ISSUE #307: Main ping copy_deduplicate query raises memory exception
tl;dr -

> Resources exceeded during query execution: The query could not be executed in the allotted memory

Looks like a single day of main ping is too much for the copy_deduplicate query as currently expressed. The Airflow job failed last night. From [the logs](https://workflow.telemetry.mozilla.org/log?execution_date=2019-08-22T01%3A00%3A00%2B00%3A00&task_id=copy_deduplicate_main_ping&dag_id=main_summary):

```
[2019-08-23 04:57:44,743] {logging_mixin.py:95} INFO - [2019-08-23 04:57:44,743] {pod_launcher.py:104} INFO -     raise exceptions.from_http_response(response)
[2019-08-23 04:57:44,747] {logging_mixin.py:95} INFO - [2019-08-23 04:57:44,747] {pod_launcher.py:104} INFO - google.api_core.exceptions.BadRequest: 400 GET https://www.googleapis.com/bigquery/v2/projects/moz-fx-data-derived-datasets/queries/c31b5f34-69c7-45d7-8b89-ec0d89153f70?maxResults=0&location=US: Resources exceeded during query execution: The query could not be executed in the allotted memory. Peak usage: 101% of limit.
[2019-08-23 04:57:44,748] {logging_mixin.py:95} INFO - [2019-08-23 04:57:44,748] {pod_launcher.py:104} INFO - Top memory consumer(s):
[2019-08-23 04:57:44,749] {logging_mixin.py:95} INFO - [2019-08-23 04:57:44,749] {pod_launcher.py:104} INFO -   query parsing and optimization: 5%
[2019-08-23 04:57:44,749] {logging_mixin.py:95} INFO - [2019-08-23 04:57:44,749] {pod_launcher.py:104} INFO -   other/unattributed: 95%
```

I will look this morning into whether it's possible to recast this query to be more efficient. It may be necessary to break this into two steps with a temp table in between.

cc @relud @whd

jklukas:
Closes #307

### DIFF
--- a/script/copy_deduplicate
+++ b/script/copy_deduplicate
@@ -19,21 +19,67 @@ from google.cloud import bigquery
 
 QUERY_TEMPLATE = """\
 WITH
-  numbered_duplicates AS (
+  -- Our 'base' is a single day of a live ping table.
+  base AS (
   SELECT
-    *,
-    ROW_NUMBER() OVER (PARTITION BY document_id ORDER BY submission_timestamp) AS _n
+    *
   FROM
     `{source_table_spec}`
   WHERE
-    DATE(submission_timestamp) = @submission_date )
+    DATE(submission_timestamp) = @submission_date ),
   --
-SELECT
-  * EXCEPT (_n)
-FROM
-  numbered_duplicates
-WHERE
-  _n = 1
+  -- We assume that the pipeline has already filtered out most duplicates,
+  -- so a list of all document_ids that appear more than once should be
+  -- reasonably small and fit in memory as a lookup table on each node.
+  duplicate_ids AS (
+  SELECT
+    document_id
+  FROM
+    base
+  GROUP BY
+    document_id
+  HAVING
+    COUNT(document_id) > 1),
+  --
+  -- We can directly include all singly-occurring documents in the final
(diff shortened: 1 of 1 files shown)


jklukas:
Your explanation of the algorithm is spot on.

> What tests have been done here? By eye, it seems like it should be fine, but it would be nice to see the queries return the same results on a dataset with known duplicates.

I have run the above query to a tmp table on top of main_v4. It yields 277615380 rows compared to 277667049 in the live table, so showing a dupe rate of 0.02%. I will check if that matches our expectations from previous looks at deduplication as to how many records should be removed. I will also run a few more tests against what's in the stable tables for ones that successfully ran last night.

jklukas:
For core_v10, this query gives 62998475, which exactly matches the stable day for 2019-08-22. The live table contains 63041461 rows, for a 0.07% duplicate rate.

jklukas:
For activity_stream.impression_stats_v1, we see the following counts.

query in this PR: 80486255
stable table: 80486255
live table: 80493347

Dupe ratio of 0.1%

PR #329, opened 2019-09-04
Add semi-compatible date_trunc version to udf_legacy

sunahsuh:
One more legacy function which should knock out a couple more compatibility cases


sunahsuh:
Hmm, I think returning a DATETIME when a DATE is passed in might have unintended effects (the way this is used is often for cohort analysis using a DATETIME as a viz label vs an existing DATE will sometimes make the layout wonky). The way this works now we'll know there was an error at query runtime if a DATETIME is passed in, and we'll just manually edit the query to use the built-in `DATETIME_TRUNC` function

sunahsuh:
I made the catchall warning a little stronger :P

PR #330, opened 2019-09-04
Fix UDF publisher script

sunahsuh:
I tried publishing the udf_legacy directory today and this script had a few issues that prevented it from running.

Might be worth adding a basic test for the script if we expect it to be in use for perpetuity

### DIFF
--- a/script/publish_persistent_udfs
+++ b/script/publish_persistent_udfs
@@ -24,7 +24,7 @@ from bigquery_etl.parse_udf import (
 )
 
 
-UDF_RE = re.compile(r"udf_(?:js_)?([a-zA-z0-9_]+)")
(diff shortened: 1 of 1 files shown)


script/publish_persistent_udfs
```diff
@@ -24,7 +24,7 @@ from bigquery_etl.parse_udf import (
 )
 
 
-UDF_RE = re.compile(r"udf_(?:js_)?([a-zA-z0-9_]+)")
+UDF_RE = re.compile(r"udf_(?:js_|legacy_)?([a-zA-z0-9_]+)")
```
relud: we should probably consider doing something more flexible here in the future

relud:
> Might be worth adding a basic test for the script if we expect it to be in use for perpetuity

filed #331

PR #337, opened 2019-09-06
Refactor glean_clients_daily to clients_daily in fenix datasets

jklukas:
This is in prep for a refactor of smoot tables.

See associated change in https://github.com/mozilla/telemetry-airflow/pull/595

My plan would be to directly copy these tables into the new locations (since the actual schemas have not changed), rerun CI (since dry run is currently failing without the tables existing), merge both PRs, update views, and delete the old tables.


templates/org_mozilla_fenix_derived/clients_daily_v1/query.sql
```diff
@@ -35,20 +35,6 @@ WITH
...
-  FROM
-    base ),
-  -- Deduplicating on document_id is necessary to get valid SUM values.
-  deduplicated AS (
-  SELECT
-    * EXCEPT (_n)
-  FROM
-    numbered_duplicates
-  WHERE
-    _n = 1 ),
```
jklukas: deduplication no longer needed since we're using the stable tables.

templates/org_mozilla_fenix_derived/clients_last_seen_v1/init.sql
```diff
@@ -1,15 +1,12 @@
...
-  glean_clients_last_seen_raw_v1
+  `moz-fx-data-shared-prod.org_mozilla_fenix_derived.clients_last_seen_v1`
 PARTITION BY
   submission_date
 CLUSTER BY
   app_name,
   os
-  -- We are receiving errors if we include sample_id as a clustering field;
-  -- see https://enterprise.google.com/supportcenter/managecases#Case/001000000040sBR/U-19888772
-  --,sample_id
```
jklukas: This should be fixed now.

relud: if this is fixed should we include sample_id ?

templates/org_mozilla_fenix/clients_last_seen/view.sql
```diff
@@ -0,0 +1,11 @@
...
+  `moz-fx-data-shared-prod.org_mozilla_fenix.clients_last_seen` AS
+SELECT
+  -- Include date_last_seen for compatibility with existing queries.
+  DATE_SUB(submission_date, INTERVAL days_since_seen DAY) AS date_last_seen,
+  -- We cannot use UDFs in a view, so we paste the body of udf_bitpos(bits) literally here.
+  CAST(SAFE.LOG(days_seen_bits & -days_seen_bits, 2) AS INT64) AS days_since_seen,
+  CAST(SAFE.LOG(days_created_profile_bits & -days_created_profile_bits, 2) AS INT64) AS days_since_created_profile,
+  *
+FROM
+  `moz-fx-data-derived-datasets.org_mozilla_fenix.clients_last_seen_v1`
```
relud: I don't see `org_mozilla_fenix.clients_last_seen_v1` defined in this PR, but it seems like these columns should be defined in a view named `org_mozilla_fenix.clients_last_seen_v1` that reads from `org_mozilla_fenix_derived.clients_last_seen_v1`, while `org_mozilla_fenix.clients_last_seen` would be created by `generate_and_publish_views`

templates/org_mozilla_fenix_derived/clients_last_seen_v1/init.sql
jklukas: Yes, that was my intent. Will fix.

templates/org_mozilla_fenix/clients_last_seen/view.sql
jklukas: My standpoint is that going forward, we generally do not want include a version suffix on user-facing views unless there is a specific need to. Similar to how we only present `telemetry.core` as a union of the underlying tables, I expect users to interact with this data as `org_mozilla_fenix.clients_last_seen`.

I had imagined that the "generate" portion of `generate_and_publish_views` would only apply to `_stable` tables. For derived tables, there is not necessarily a 1:1 relationship between underlying tables and user-facing views, so my assumption is that we always write an explicit view and check it into this repo.

sql/org_mozilla_fenix_derived/clients_last_seen_v1/init.sql
```diff
@@ -1,15 +1,12 @@
 CREATE TABLE
-  glean_clients_last_seen_raw_v1
+  `moz-fx-data-shared-prod.org_mozilla_fenix_derived.clients_last_seen_v1`
```
relud:
```suggestion
  clients_last_seen_v1
```

sql/org_mozilla_fenix_derived/clients_last_seen_v1/query.sql
```diff
@@ -32,15 +32,15 @@ WITH
...
+    `moz-fx-data-shared-prod.org_mozilla_fenix_derived.clients_daily_v1`
   WHERE
     submission_date = @submission_date ),
   --
   _previous AS (
   SELECT
     * EXCEPT (submission_date)
   FROM
-    glean_clients_last_seen_raw_v1 AS cls
+    `moz-fx-data-shared-prod.org_mozilla_fenix_derived.clients_last_seen_v1` AS cls
```
relud:
```suggestion
    clients_last_seen_v1 AS cls
```

sql/org_mozilla_fenix_derived/clients_last_seen_v1/query.sql
```diff
@@ -32,15 +32,15 @@ WITH
       DATE_DIFF(submission_date, first_run_date, DAY)) AS days_created_profile_bits,
     * EXCEPT (submission_date)
   FROM
-    glean_clients_daily_v1
+    `moz-fx-data-shared-prod.org_mozilla_fenix_derived.clients_daily_v1`
```
relud:
```suggestion
    clients_daily_v1
```

sql/org_mozilla_fenix/clients_last_seen/view.sql
```diff
@@ -0,0 +1,11 @@
...
+  `moz-fx-data-shared-prod.org_mozilla_fenix.clients_last_seen` AS
+SELECT
+  -- Include date_last_seen for compatibility with existing queries.
+  DATE_SUB(submission_date, INTERVAL days_since_seen DAY) AS date_last_seen,
+  -- We cannot use UDFs in a view, so we paste the body of udf_bitpos(bits) literally here.
+  CAST(SAFE.LOG(days_seen_bits & -days_seen_bits, 2) AS INT64) AS days_since_seen,
+  CAST(SAFE.LOG(days_created_profile_bits & -days_created_profile_bits, 2) AS INT64) AS days_since_created_profile,
+  *
+FROM
+  `moz-fx-data-derived-datasets.org_mozilla_fenix.clients_last_seen_v1`
```
relud:
```suggestion
  `moz-fx-data-shared-prod.org_mozilla_fenix.clients_last_seen_v1`
```

templates/org_mozilla_fenix/clients_last_seen/view.sql
relud: lgtm

relud: docs issue filed for later at #343

PR #272, opened 2019-07-31
UDF for unzipping gzipped bytes

scholtzan:
This addresses https://github.com/mozilla/bigquery-etl/issues/270

I added a UDF for unzipping gzipped bytes and updated the `publish_persistent_udfs` script to upload JavaScript files that are external dependencies for UDFs to GCS. The dependencies are in the `udf_dependencies` directory, but perhaps it makes more sense to move them to somewhere else (eg. inside `udf_js`)?

PR #338, opened 2019-09-06
Update addon_aggregates views to fix table rename bug1579478

jasonthomas:
Update views to use the updated table name in https://github.com/mozilla/telemetry-airflow/pull/596.

### DIFF
--- a/sql/telemetry/addon_aggregates/view.sql
+++ b/sql/telemetry/addon_aggregates/view.sql
@@ -1,0 +1,4 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.addon_aggregates`
+AS SELECT * FROM
+  `moz-fx-data-derived-datasets.telemetry_derived.addon_aggregates_v2`
--- a/sql/telemetry/addon_aggregates/view.sql
+++ b/sql/telemetry/addon_aggregates/view.sql
@@ -1,4 +1,4 @@
 CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.telemetry.addon_aggregates`
 AS SELECT * FROM
-  `moz-fx-data-derived-datasets.telemetry_derived.addon_aggregates_v2`
+  `moz-fx-data-shared-prod.telemetry.addon_aggregates_v2`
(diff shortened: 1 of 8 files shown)


jasonthomas:
@relud are you aware of any recent changes to the google credential configuration for CircleCI `build`?

jklukas:
> @relud are you aware of any recent changes to the google credential configuration for CircleCI `build`?

It changed ~2 weeks ago when we merged https://github.com/mozilla/bigquery-etl/pull/272 . We should think a bit more carefully about how we want that to work, since it's indeed a problem right now that PRs submitted from anyone not in the `telemetry` GitHub org fail CI.

relud:
> since it's indeed a problem right now that PRs submitted from anyone not in the `mozilla/telemetry` GitHub team fail CI.

I'm generally in favor of avoiding this by submitting PRs from non-fork branches. I don't see any problem with granting everyone in `kparlante-all` and `moconnor-all` write permissions to this repo (esp with master protected), I just don't know if there are github teams for that.

jklukas:
It's my understanding that moving to CircleCI contexts basically erases the problem of fork vs. non-fork, but shifts the problem to whether the user is in the associated GitHub team or not. It does seem like we should have a GitHub team that should roughly cover the people we'd want to be able to submit PRs here.

jasonthomas:
> @jasonthomas could you grant access to the data-eng-circleci-deployment context to the svcops team and if possible grant data-eng-circleci-tests access to the entire mozilla org?

FWIW I was already in the telemetry mozilla github group. 

I've created a new group named `dataops` and added it to `data-eng-circleci-deployment` context and added 'All members' of mozilla github org to`data-eng-circleci-tests`. It looks like adding groups to the security context is not sufficient, I believe we will also need `Pass secrets to builds from forked pull requests` enabled on the CircleCI repo configuration. 

I think the next steps would be - 

* Move all current environment variables into restricted contexts, specifically DOCKERHUB related credentials. 
* Specify security context for each job in CircleCI workflow configuration
* Enable `Pass secrets to builds from forked pull requests`

Based on these changes I believe the forked pull requests will receive the correct environment variables based restricted context membership.

jklukas:
I am wary of turning on `Pass secrets to builds from forked pull requests` since it relies on discipline to make sure we put secrets only in CircleCI contexts and don't add any secrets directly to the job.

Other than that, @jasonthomas's proposed steps sound like they should work. I particularly like the idea of a future where only the job that does Docker releases has DockerHub creds, etc.

PR #345, opened 2019-09-09
Add test support for view-indirection and fix ssl_ratios_v1

relud:


sql/telemetry_derived/ssl_ratios_v1/query.sql
```diff
@@ -14,7 +14,11 @@ SELECT
   SUM(udf_get_key(histogram_parent_http_pageload_is_ssl, 0)) AS non_ssl_loads,
   SUM(udf_get_key(histogram_parent_http_pageload_is_ssl, 1)) AS ssl_loads,
   -- ratio of pings that have the probe
-  COUNTIF(ARRAY_LENGTH(histogram_parent_http_pageload_is_ssl) > 0) / COUNT(*) AS reporting_ratio
+  -- It is only possible for histogram_parent_http_pageload_is_ssl to be NULL
+  -- because telemetry.main_summary_v4 is a view that selects it from a STRUCT.
+  -- If the underlying table changes to not nest this column in a STRUCT this must become:
+  -- COUNTIF(ARRAY_LENGTH(histogram_parent_http_pageload_is_ssl) > 0) / COUNT(*) AS reporting_ratio
+  COUNT(histogram_parent_http_pageload_is_ssl) / COUNT(*) AS reporting_ratio
```
jklukas: What does `ARRAY_LENGTH(NULL)` return? Can we future-proof this by making it:

       COUNTIF(histogram_parent_http_pageload_is_ssl IS NOT NULL AND ARRAY_LENGTH(histogram_parent_http_pageload_is_ssl) > 0) / COUNT(*) AS reporting_ratio

relud: `ARRAY_LENGTH(NULL)` returns `NULL`, which is `FALSE` for the purposes of `COUNTIF`, so your suggestion is the same as using `COUNTIF(ARRAY_LENGTH(histogram_parent_http_pageload_is_ssl) > 0)`.

We can't future proof this query right now because it must exactly match legacy behavior so that we can migrate to it.

PR #356, opened 2019-09-16
Implement udf_sample_id in SQL

relud:
I did this as an exercise, I have not compared the performance to `udf_js_sample_id`


udf/crc32_uuid.sql
```diff
@@ -0,0 +1,108 @@
...
+    SUBSTR(value, 9, 8)
+  )
+);
+CREATE TEMP FUNCTION udf_crc32_partial_32(crc INT64, value BYTES) AS (
+  udf_crc32_partial_16(
+    udf_crc32_partial_16(crc, SUBSTR(value, 1, 16)),
+    SUBSTR(value, 17, 16)
+  )
+);
+CREATE TEMP FUNCTION udf_crc32_partial_36(crc INT64, value BYTES) AS (
```
jklukas: Do all of these temp functions become real functions that have to be published? That's my main concern here is polluting the udf namespace. Maybe we can invent a naming convention for "private" UDFs?

udf/crc32_uuid.sql
```diff
@@ -0,0 +1,108 @@
...
+See https://en.wikipedia.org/wiki/Cyclic_redundancy_check
+*/
+CREATE TEMP FUNCTION udf_crc32_table() AS (
+  -- precalculated from this algorithm:
+  -- for c in range(256):
+  --   for _ in range(8):
+  --     c = (0xEDB88320 if c & 1 else 0) ^ (c >> 1)
+  --   yield c
+  [
+    0, 1996959894, 3993919788, 2567524794, 124634137, 1886057615, 3915621685,
```
relud: this isn't formatted compatibly with the sql style guide because that makes it unreasonably long

udf/sample_id.sql
```diff
@@ -0,0 +1,24 @@
+/*
+Stably hash a client_id to an integer between 0 and 99.
+This function is technically defined in SQL, but it calls a JS UDF
+implementation of a CRC-32 hash, so we defined it here to make it
+clear that its performance may be limited by BigQuery's JavaScript
+UDF environment.
```
jklukas: Outdated docs here.

udf/crc32_uuid.sql
```diff
@@ -0,0 +1,108 @@
...
+  )
+);
+CREATE TEMP FUNCTION udf_crc32_uuid(value BYTES) AS (
+  udf_crc32_partial_36(0xFFFFFFFF, value) ^ 0xFFFFFFFF
+);
+SELECT
+  assert_equals(
+    308953907,
+    udf_crc32_uuid(b"51baf8b4-75d1-3648-b96d-809569b89a12")
+  )
```
jklukas: Can we add a test for the non-32-char case? Perhaps in the body of the `udf_crc32_uuid` function itself we should check the length and call `ERROR` with a useful message.

udf/crc32_uuid.sql
```diff
@@ -0,0 +1,108 @@
...
+);
+CREATE TEMP FUNCTION udf_crc32_partial_36(crc INT64, value BYTES) AS (
+  udf_crc32_partial_4(
+    udf_crc32_partial_32(crc, SUBSTR(value, 1, 32)),
+    SUBSTR(value, 33, 4)
+  )
+);
+CREATE TEMP FUNCTION udf_crc32_uuid(value BYTES) AS (
+  udf_crc32_partial_36(0xFFFFFFFF, value) ^ 0xFFFFFFFF
```
relud: I chose to only fully implement 36-byte crc32 because every instance of client id in mozilla-pipeline-schemas is validated as that length, except in the metadata schemas.

udf/crc32_uuid.sql
relud: i think that we have to publish all of the functions.

i'm a fan of `udf__helper_func`, because it results in `udf._helper_func`, which follows the python private-if-we-had-that syntax.

udf/crc32_uuid.sql
relud: we can't have tests that hit a case with an `error` clause right now. Instead I've added a `safe_` prefix to indicate that invalid values will return NULL, and handled that case.

package.json
```diff
@@ -0,0 +1,23 @@
...
+  "license": "MPL-2.0",
+  "dependencies": {
+    "sql-formatter": "2.3.2"
+  },
+  "devDependencies": {
+    "eslint": "^5.16.0",
+    "eslint-config-airbnb-base": "13.1.0",
+    "eslint-plugin-import": "2.17.3"
+  }
+}
```
jklukas: Looks like this spilled over from the formatting PR.

relud: 😣 indeed. will remove.

Issue #348, opened 2019-09-11
Start importing amplitude events from FxA oauth server into BQ datasets


irrationalagent:
We need to start counting FxA amplitude events from the oauth server logs, e.g. `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_oauth_YYYYMMDD`

We currently only import events from the auth and content servers into bigquery. 

I believe this to be the cause of at least part, perhaps all of the discrepancy between the GUD dashboard's account of FxA MAU and amplitude https://analytics.amplitude.com/mozilla-corp/chart/sxp5s3v.

We will also need this for skyline reporting purposes. 

IIUC it should be possible to write a similar pipeline to the current jobs that slurp from the auth/content servers. Then we would also need to update the derived datasets so that they start counting these events.

jklukas:
Sounds reasonable. Should be fairly straightforward.

jklukas:
I'm not seeing `jsonPayload.type = 'amplitudeEvent'` like in other tables, but I do see `*.amplitudeEvent`. Should I accept any event where the type ends with amplitudeEvent?

irrationalagent:
@jklukas I think this would be ok but checking with @philbooth to confirm?

philbooth:
> I'm not seeing `jsonPayload.type = 'amplitudeEvent'` like in other tables, but I do see `*.amplitudeEvent`. Should I accept any event where the type ends with amplitudeEvent?

Just looked at the code in the oauth server, so are you seeing e.g. `routes.verify.amplitudeEvent`? It seems unfortunate we did it like that, but yeah, I don't see anything else logging with that string so you should be fine to match it like that.

jklukas:
> routes.verify.amplitudeEvent

Yes, that's one of the variants. There are a few. I'll proceed with `jsonPayload.type LIKE '%.amplitudeEvent'` in the query

jklukas:
It looks like `jsonPayload.fields.op` equals "amplitudeEvent" for all cases where the jsonPayload.type ends in amplitudeEvent. Would that be a more reasonable thing to filter on?

philbooth:
Yep, definitely, do that!

irrationalagent:
> It's possible to recalculate dependent tables so that we get a full recalculation of MAU with `oauth` included. Is that desirable?

I think it is. We are already counting these events in amplitude toward our tracking of MAU so I don't see an obvious reason why wouldn't try to bring these tables into alignment with that.

[...]

PR #362, opened 2019-09-19
Import events from FxA oauth server

jklukas:
I've already created the new table and backfilled existing dates.

Addresses #348

### DIFF
--- a/templates/telemetry/fxa_all_events_v1/view.sql
+++ b/templates/telemetry/fxa_all_events_v1/view.sql
@@ -45,6 +45,20 @@ WITH
       jsonPayload.fields.event_type
     FROM
       `moz-fx-data-derived-datasets.telemetry.fxa_content_events_v1`
+  ),
+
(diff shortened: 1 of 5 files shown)


templates/telemetry/fxa_all_events_v1/view.sql
```diff
@@ -45,10 +45,26 @@ WITH
       jsonPayload.fields.event_type
     FROM
       `moz-fx-data-derived-datasets.telemetry.fxa_content_events_v1`
+  ),
+
+  fxa_oauth_events AS (
+    SELECT
+      timestamp AS submission_timestamp,
+      jsonPayload.fields.user_id,
+      CAST(NULL AS STRING) AS country,
```
irrationalagent: were these missing from the oauth events?

jklukas: Correct. I do not see any of these fields present in the schema for oauth events.

PR #366, opened 2019-09-19
Update kpi_dashboard query to exclude FirefoxForFireTV

jklukas:
Matches updates made to https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/67817/command/95274

### DIFF
--- a/sql/telemetry/firefox_kpi_dashboard_v1/query.sql
+++ b/sql/telemetry/firefox_kpi_dashboard_v1/query.sql
@@ -87,12 +87,9 @@ WITH
     datasource NOT IN ('nondesktop_global', 'nondesktop_tier1')
   UNION ALL
   SELECT
-    * EXCEPT (asofdate) REPLACE (REPLACE(datasource, '_nofire', '') AS datasource)
+    * REPLACE (REPLACE(datasource, '_nofire', '') AS datasource)
   FROM
-    `moz-fx-data-derived-datasets.analysis.jmccrosky_test`
-  WHERE
-    datasource LIKE 'nondesktop_nofire_%'
-    AND asofdate = (SELECT MAX(asofdate) AS asofdate FROM `moz-fx-data-derived-datasets.analysis.jmccrosky_test`)
+    `moz-fx-data-derived-datasets.analysis.growth_dashboard_forecasts_nofire`
   ),
   --
   desktop_base AS (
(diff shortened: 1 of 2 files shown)


sql/telemetry/firefox_kpi_dashboard_v1/query.sql
```diff
@@ -81,7 +81,19 @@ WITH
...
-    `moz-fx-data-derived-datasets.analysis.growth_dashboard_forecasts_current` ),
+    `moz-fx-data-derived-datasets.analysis.growth_dashboard_forecasts_current`
+  -- We patch in nondesktop forecasts that exclude Firefox for Fire TV.
+  WHERE
+    datasource NOT IN ('nondesktop_global', 'nondesktop_tier1')
+  UNION ALL
+  SELECT
+    * EXCEPT (asofdate) REPLACE (REPLACE(datasource, '_nofire', '') AS datasource)
+  FROM
+    `moz-fx-data-derived-datasets.analysis.jmccrosky_test`
```
relud: we should copy this table out of analysis and name it something more descriptive than `jmccrosky_test`

jklukas: I agree. And we should probably block merging this query update until that's the case. I will pursue with @jmccrosky.

jklukas:
@jmccrosky is on PTO Monday, but will pursue changing the table on Tuesday. The existing table is called `analysis.growth_dashboard_forecasts`. I'd advocate for calling the table something like `analysis.growth_dashboard_forecasts_nofire` if we can populate that with just the nofire forecasts. Or actually, now that I think about it this could be a view.

jklukas:
I created the following view:

```
CREATE OR REPLACE VIEW
  `moz-fx-data-derived-datasets.analysis.growth_dashboard_forecasts_nofire` AS
WITH
  base AS (
  SELECT
    *
  FROM
    `moz-fx-data-derived-datasets.analysis.jmccrosky_test` )
  --
SELECT
  * EXCEPT (asofdate)
FROM
  base
WHERE
  datasource LIKE 'nondesktop_nofire_%'
  AND asofdate = (
  SELECT
    MAX(asofdate) AS asofdate
  FROM
    base)
```

It points at jmccrosky_test for now, but we can change where the view points as Jesse adapts to a new, more productionized table. Or perhaps we replace the view with a table Jesse writes into. I am going to update the query here to hit the view, and then merge this PR.

PR #370, opened 2019-09-21
Add --hourly option to copy_deduplicate

### LINKED ISSUE #367: More reliable copy_deduplicate for main_v4
For the first time today, the copy_deduplicate job for main_v4 failed all 3 of its attempts. We also had significant trouble being able to copy full days of main_v4 in https://github.com/mozilla/bigquery-etl/issues/355.

We have a [case open with Google](https://enterprise.google.com/supportcenter/managecases#Case/001000000040sBR/U-20528460) to discuss the issue, which may give us some more advice on how to proceed. So far, their primary suggestion has been to run this query in a project with flat-rate billing. That _may_ give us more reliable access to large numbers of concurrent slots.

We could also break out the logic into multiple queries, as @relud has suggested. We'd run one query to materialize a table of duplicated doc IDs, then we run multiple queries for various time chunks of the live table (perhaps 24 queries that each copy 1 hour of data).

relud:
fixes #367

blocks https://github.com/mozilla/telemetry-airflow/pull/617

for sept 19th splitting up deduplication hourly totaled ~9.04TiB processed, vs ~8.75TiB for the single-query deduplication, or ~3.3% increase in cost. I ran it with `--parallelism=24` so that the queries would be simultaneous, all of the queries succeeded on the first try, and the slowest query ran for 32 minutes, followed by a 14 second copy operation. For comparison, the same day took 4 tries in airflow, the final attempt ran for 48 minutes, and including time for retries usually runs for 3-5 hours on week days.

### DIFF
--- a/script/copy_deduplicate
+++ b/script/copy_deduplicate
@@ -11,15 +11,16 @@ or to process only a specific list of tables.
 """
 
 from argparse import ArgumentParser
-from datetime import datetime
+from datetime import datetime, timedelta
 from fnmatch import fnmatch
+from itertools import groupby
 from multiprocessing.pool import ThreadPool
+from uuid import uuid4
 
 from google.cloud import bigquery
 
-QUERY_TEMPLATE = """\
-WITH
-  -- Our 'base' is a single day of a live ping table.
+QUERY_DAY_BASE = """
+  -- A single day of a live ping table.
   base AS (
   SELECT
     *
@@ -27,6 +28,32 @@ WITH
     `{source_table_spec}`
   WHERE
     DATE(submission_timestamp) = @submission_date ),
+"""
+
+QUERY_HOUR_BASE = """
+  -- Document_ids already seen in previous hours on this day
+  previous_hours_document_ids AS (
+  SELECT
+    DISTINCT document_id
+  FROM
+    `{source_table_spec}`
+  WHERE
+    DATE(submission_timestamp) = DATE(@submission_hour)
+    AND submission_timestamp < @submission_hour),
+  -- A single hour of a live ping table.
+  base AS (
+  SELECT
+    *
+  FROM
+    `{source_table_spec}`
+  WHERE
+    TIMESTAMP_TRUNC(submission_timestamp, HOUR) = @submission_hour
+    AND document_id NOT IN (SELECT * FROM previous_hours_document_ids)),
+"""
+
+QUERY_TEMPLATE = """\
+WITH
+  {base}
   --
   -- We assume that the pipeline has already filtered out most duplicates,
   -- so a list of all document_ids that appear more than once should be
@@ -110,6 +137,26 @@ parser.add_argument(
         "and bytes that would be processed"
     ),
 )
+parser.add_argument(
+    "--priority",
+    default="INTERACTIVE",
+    type=str.upper,
+    choices=["BATCH", "INTERACTIVE"],
+    help=(
+        "Deduplicate one hour at a time, rather than for whole days at once; "
+        "avoids memory overflow at the cost of less effective clustering; "
+        "recommended only for tables failing due to memory overflow"
+    ),
+)
+parser.add_argument(
+    "--hourly",
+    action="store_true",
+    help=(
+        "Deduplicate one hour at a time, rather than for whole days at once; "
+        "avoids memory overflow at the cost of less effective clustering; "
+        "recommended only for tables failing due to memory overflow"
+    ),
+)
 group = parser.add_mutually_exclusive_group()
 group.add_argument(
     "--only",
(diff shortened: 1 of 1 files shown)


script/copy_deduplicate
```diff
@@ -130,42 +167,87 @@ group.add_argument(
...
+        query_job.result()
+    return stable_table, query_job
+
+
+def copy_join_parts(client, stable_table, query_jobs):
+    total_bytes = sum(query.total_bytes_processed for query in query_jobs)
+    if query_jobs[0].dry_run:
+        api_repr = query_jobs[0].to_api_repr()
+        if len(query_jobs) > 1:
+            print(f"Would process {total_bytes} bytes: [{api_repr},...]")
```
relud: hourly deduplciation depends on clustering, so total_bytes here is an upper bound. That isn't so bad until you realize that this estimates deduplicating main_summary hourly will scan the full day 24 times and cost $1087.50 😨

relud: actual usage for hourly on main_v4 was only 3.3% more than daily, so we should be good.

script/copy_deduplicate
```diff
@@ -130,42 +167,88 @@ group.add_argument(
...
+        # look up the dataset used for query results without a destination
+        dry_run = bigquery.QueryJobConfig(dry_run=True)
+        destination = client.query("SELECT 1", dry_run).destination
+        temporary_dataset = client.dataset(destination.dataset_id, destination.project)
+    return temporary_dataset
+
+
+def get_temporary_table(client, date):
+    """Generate a random writable anonynmous table reference."""
+    return get_temporary_dataset(client).table(f"anon{uuid4().hex}${date:%Y%m%d}")
```
jklukas: It would be nice to avoid having this logic. Is it possible to let BQ assign temp tables, and then fetch the name from the query results? Or do we have to set an explicit table name because of the size of the results?

Also, do we know that BQ will automatically clean up tables in this dataset, even if we've created them explicitly rather than having BQ generate them?

Also, does this function follow the exact format of BQ's generated table names? If so, we should document that intention.

script/copy_deduplicate
```diff
@@ -130,42 +167,88 @@ group.add_argument(
...
-    query_job = client.query(sql, job_config)
-    if dry_run:
-        print(
-            "Would process {} bytes: {}".format(
-                query_job.total_bytes_processed, query_job.to_api_repr()
+
+def get_query_job_configs(client, stable_table, date, dry_run, hourly):
+    kwargs = dict(use_legacy_sql=False, dry_run=dry_run)
+    if hourly:
+        kwargs["allow_large_results"] = True
```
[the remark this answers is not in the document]
relud: I can attest that `allow_large_results` is still required if `destination` is not set. I forgot to remove this line when I added `def get_temporary_table` and started actually setting `destination`.

script/copy_deduplicate
```diff
@@ -130,42 +167,88 @@ group.add_argument(
...
+        stable_table = client.get_table(stable_table)
+        for hour in range(24):
+            value = (date + timedelta(hours=hour)).isoformat()
+            yield bigquery.QueryJobConfig(
+                destination=get_temporary_table(client, date),
```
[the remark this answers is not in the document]
relud: `WRITE_APPEND` works fine, but using temporary tables with a multi-source copy makes the day atomic and idempotent, which I feel very strongly about preserving.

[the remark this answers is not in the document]
relud: I'll add a flag to allow setting `batch` priority.

script/copy_deduplicate
relud: > It would be nice to avoid having this logic. Is it possible to let BQ assign temp tables, and then fetch the name from the query results? Or do we have to set an explicit table name because of the size of the results?

we have to set an explicit table name because of result size, `time_partitioning`, and `clustering_fields`. result size is fine if we set `allow_large_results`, but `time_partitioning` and `clustering_fields` require `destination` to be set.

I also tried leaving out `time_partitioning` and `clustering_fields` while setting `dry_run`, then re-using that destination, but that returns the error message: `Cannot explicitly modify anonymous table`.

> Also, do we know that BQ will automatically clean up tables in this dataset, even if we've created them explicitly rather than having BQ generate them?

The temporary results dataset has a default expiration of 1 day, but I can confirm from my tests that it only impacts partition expiration, which is why i'll be adding explicit table expiration.

> Also, does this function follow the exact format of BQ's generated table names? If so, we should document that intention.

it follows it closely enough that the web console shows a link to `Temporary table` as the destination, but not closely enough that it would collide with a name generated by BigQuery. I will add an explanation in the docstring.

PR #373, opened 2019-09-23
create new fxa_auth_content_oauth view

irrationalagent:
Following on #362, I'd like to add this view which is an expansion of `fxa_content_auth_events` to also include the new oauth events. I opted to make a new view instead of replacing the existing one (just to avoid breaking existing queries), but if we'd prefer to keep views to a minimum, I'd be OK with just renaming/replacing the old view. 

sorry for the PR out of the blue, just trying to get the FxA data in BQ up to snuff before skyline.

### DIFF
--- a/sql/telemetry/fxa_content_auth_oauth_events_v1/view.sql
+++ b/sql/telemetry/fxa_content_auth_oauth_events_v1/view.sql
@@ -1,0 +1,89 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.fxa_content_auth_oauth_events_v1` AS
+  --
+  WITH
+    content as (
+    SELECT
+      jsonPayload.logger,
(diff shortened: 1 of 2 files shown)


sql/telemetry/fxa_content_auth_oauth_events_v1/view.sql
```diff
@@ -0,0 +1,89 @@
+CREATE OR REPLACE VIEW
```
jklukas: This file needs to be defined under `templates/` rather than `sql`; see https://github.com/mozilla/bigquery-etl#contributing ; this instructions may be new since you last contributed.

irrationalagent: oops will fix

jklukas:
This is deployed.

PR #374, opened 2019-09-23
Set temp table expiration in copy_deduplicate

relud:
and delete temp tables when finished, because there's no need to leave them around for a full day.

### DIFF
--- a/script/copy_deduplicate
+++ b/script/copy_deduplicate
@@ -191,12 +191,13 @@ def get_temporary_dataset(client):
     return temporary_dataset
 
 
-def get_temporary_table(client, date):
-    """Generate a random writable temporary table reference.
+def get_temporary_table(client, schema, clustering_fields, time_partitioning, date):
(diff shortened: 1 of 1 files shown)


script/copy_deduplicate
```diff
@@ -213,11 +214,17 @@ def get_temporary_table(client, date):
...
+    table.expires = datetime.now() + timedelta(days=1)
+    table.clustering_fields = clustering_fields
+    table.time_partitioning = time_partitioning
+    table = client.create_table(table, exists_ok=False)
+    return f"{sql_full_table_id(table)}${date:%Y%m%d}"
 
 
 def sql_full_table_id(table):
-    return table.full_table_id.replace(":", ".")
+    return f"{table.project}.{table.dataset_id}.{table.table_id}"
```
relud: `table.full_table_id` is only available when set by the server on `Table` and `TableListItem` objects, so I updated this function correctly handle `TableReference` found in `query_job.destination` rather than make extra api calls.

PR #381, opened 2019-09-24
Add additional fields to search datasets

Ben-Wu:
Search aggregates assumes search clients daily is written to a table called `search_clients_daily_v8_bq_test` which will need to be changed later.  

@wlach Could we just use `search_clients_daily_v8` for validation?  If it doesn't pass validation we would delete the table anyway.

v8 schema changes for both search_clients and aggregates:
- removed `submission_date_s3` field; `submission_date` is the date field and no longer a string
- `sample_id` is now an integer


templates/search/search_clients_daily_v8/query.sql
```diff
@@ -1,3 +1,14 @@
+-- Return the version of the search addon if it exists, null otherwise
+CREATE TEMP FUNCTION
+  get_search_addon_version(addon_list ANY TYPE) AS ((
+    SELECT
+      element.version
+    FROM
+      UNNEST(addon_list)
+    WHERE
+      element.addon_id = 'followonsearch@mozilla.com'
```
relud: this needs a limit clause to prevent query errors if someone sends us garbage
```suggestion
      element.addon_id = 'followonsearch@mozilla.com'
    LIMIT
      1
```

templates/search/search_aggregates_v8/query.sql
```diff
@@ -16,16 +20,19 @@ SELECT
   SUM(search_with_ads) as search_with_ads,
   SUM(unknown) as unknown
 FROM
-  search_clients_daily_v8
+  search_clients_daily_v8_bq_test
```
relud: this seems like it shouldn't be changed

templates/search/search_clients_daily_v8/query.sql
relud: alternately we could attempt to get the max or mode-last version

templates/search/search_aggregates_v8/query.sql
```diff
@@ -16,16 +20,19 @@ SELECT
...
 GROUP BY
   submission_date,
+  addon_version,
   app_version,
   country,
   distribution_id,
+  engine,
   locale,
+  os,
+  os_version,
```
relud: these should move to the end of the `GROUP BY` list, to keep consistent with the column order

templates/search/search_aggregates_v8/query.sql
```diff
@@ -8,6 +9,9 @@ SELECT
   search_cohort,
   source,
   default_search_engine,
+  os,
+  os_version,
+  COUNT(DISTINCT client_id) as client_count,
```
relud: reading from `search_clients_daily_v8` means we already every row has a `DISTINCT client_id`, so we should use `COUNT(*)` here instead:
```suggestion
  COUNT(*) as client_count,
```

tests/search/search_aggregates_v8/test_aggregation/search_clients_daily_v8_bq_test.schema.json
```diff
@@ -2,102 +2,96 @@
...
   },
   {
     "type": "STRING",
-    "name": "channel",
-    "mode": "NULLABLE"
-  },
-  {
-    "type": "FLOAT",
-    "name": "active_addons_count_mean",
+    "name": "sample_id",
```
relud: `sample_id` should be an integer here

wlach:
> @wlach Could we just use `search_clients_daily_v8` for validation? If it doesn't pass validation we would delete the table anyway.

I am worried that people would think they were supposed to use `search_clients_daily_v8` if it was there, so my preference would be to not do that. Having to submit another PR to swap table names is not so bad IMO. @relud do you have any thoughts on this?

tests/search/search_aggregates_v8/test_aggregation/search_clients_daily_v8_bq_test.schema.json
```diff
@@ -2,102 +2,96 @@
...
-    "mode": "REQUIRED"
+    "type": "STRING",
+    "name": "client_id",
+    "mode": "NULLABLE"
   },
-
   {
     "type": "STRING",
-    "name": "app_version",
+    "name": "engine",
```
relud: modifying the test to change the column order makes the actual changes to the schema hard to indentify, and chance you could restore the original column order, or put changing the order in a separate PR?

tests/search/search_aggregates_v8/test_aggregation/expect.ndjson
```diff
@@ -0,0 +1,5 @@
...
+{"submission_date": "2017-01-01", "addon_version": "0.9.5", "app_version": "54.0.1", "country": "DE", "engine": "yahoo", "locale": "de", "source": "urlbar", "default_search_engine": "google", "os": "windows", "os_version": "10.0", "client_count": 1, "organic": 0, "tagged_sap": 0, "tagged_follow_on": 0, "sap": 4, "ad_click": 0, "search_with_ads": 0, "unknown": 0}
+{"submission_date": "2017-01-01", "addon_version": "0.9.6", "app_version": "54.0.1", "country": "US", "engine": "google", "locale": "de", "source": "urlbar", "default_search_engine": "google", "os": "windows", "os_version": "10.0", "client_count": 1, "organic": 0, "tagged_sap": 0, "tagged_follow_on": 0, "sap": 4, "ad_click": 0, "search_with_ads": 0, "unknown": 0}
+{"submission_date": "2017-01-01", "app_version": "54.0.1", "country": "DE", "locale": "de", "default_search_engine": "google", "os": "windows", "os_version": "10.0", "client_count": 1, "organic": 0, "tagged_sap": 0, "tagged_follow_on": 0, "sap": 0, "ad_click": 0, "search_with_ads": 0, "unknown": 0}
```
relud: this file is missing a trailing newline

tests/search/search_clients_daily_v8/test_aggregation/telemetry_derived.main_summary_v4.schema.json
```diff
@@ -1,7 +1,7 @@
 [
   {
     "type": "DATE",
-    "name": "submission_date_s3",
+    "name": "submission_date",
     "mode": "REQUIRED"
   },
```
relud: sample id in this schema needs to be modified to INTEGER

relud:
> > @wlach Could we just use `search_clients_daily_v8` for validation? If it doesn't pass validation we would delete the table anyway.
> 
> I am worried that people would think they were supposed to use `search_clients_daily_v8` if it was there, so my preference would be to not do that. Having to submit another PR to swap table names is not so bad IMO. @relud do you have any thoughts on this?

in that case let's put this in a different bigquery dataset when we deploy it, so that promoting to prod only modifies airflow and there's no confusion around whether or not to use this.

tests/search/search_aggregates_v8/test_aggregation/search_clients_daily_v8_bq_test.schema.json
Ben-Wu: `sample_id` is a string in the `search_clients_daily_v7` even though the values are integers.  I'm not sure it's worth changing.

Ben-Wu: Unless we decided that all `sample_id` will be integers from now on

relud: we did decide that, also they are integers in main_summary (in bigquery), so they will be integers in search_clients_daily_v8, so they will be integers here

Ben-Wu:
Reverted schema order and using `search_clients_daily_v8` instead of `search_clients_daily_v8_bq_test`

templates/search/search_clients_daily_v8/query.sql
```diff
@@ -1,3 +1,16 @@
...
+    LIMIT
```
relud: upon further consideration, this should definitely be selecting mode_last:
```suggestion
    GROUP BY
      element.version
    ORDER BY
      COUNT(element.version) DESC,
      MAX(OFFSET) DESC
    LIMIT
```

templates/search/search_clients_daily_v8/query.sql
```diff
@@ -1,3 +1,16 @@
+-- Return the version of the search addon if it exists, null otherwise
+CREATE TEMP FUNCTION
+  get_search_addon_version(addon_list ANY TYPE) AS ((
+    SELECT
+      element.version
+    FROM
+      UNNEST(addon_list)
+    WHERE
```
relud:
```suggestion
    WITH OFFSET
    WHERE
```

templates/search/search_clients_daily_v8/query.sql
```diff
@@ -59,11 +72,12 @@ WITH
   windowed AS (
   SELECT
     ROW_NUMBER() OVER w1_unframed AS _n,
-    submission_date_s3 AS submission_date,
+    submission_date,
     client_id,
     engine,
     source,
     udf_mode_last(ARRAY_AGG(country) OVER w1) AS country,
+    get_search_addon_version(active_addons.list) AS addon_version,
```
relud: this should aggregate active_addons from all of the rows for the day:
```suggestion
    get_search_addon_version(ARRAY_CONCAT_AGG(active_addons.list) OVER w1) AS addon_version,
```

Ben-Wu: This gives an error `Analytic function ARRAY_CONCAT_AGG is not supported`.  I think it's because `OVER` isn't supported by `ARRAY_CONCAT_AGG`.  Is that correct?  Any ideas for workarounds?  `ARRAY_AGG(active_addons)` might work but some other changes need to be made.

jklukas: Yes, you're correct. Looks like ARRAY_CONCAT_AGG can only be used as an aggregate function, not an analytic window function. I don't have an immediate idea for an elegant workaround, though.

relud: you can work around this by nesting the aggregation, like so:

```suggestion
    get_search_addon_version(ARRAY_AGG(active_addons) OVER w1) AS addon_version,
```

and then

```sql
CREATE TEMP FUNCTION get_search_addon_version(active_addons ANY type) AS (
  (
    SELECT AS STRUCT
      element.version,
      MAX(_offset_1),
      MAX(_offset_2)
    FROM
      (
        SELECT
          list,
          _offset_1
        FROM
          UNNEST(active_addons)
        WITH
          OFFSET AS _offset_1
      ),
      UNNEST(list) AS value
    WITH
      OFFSET AS _offset_2
    WHERE
      element.addon_id = 'followonsearch@mozilla.com'
    GROUP BY
      element.version
    ORDER BY
      COUNT(element.version) DESC,
      MAX(_offset_1) DESC,
      MAX(_offset_2) DESC
    LIMIT
      1
  )
);
```

Ben-Wu:
Added aggregation for addon version (thanks @relud).  Also added a test for the case where a user has multiple search addons.

PR #382, opened 2019-09-25
Improve list_tables speed for script/copy_deduplicate

relud:
this PR cuts the time to run `time script/copy_deduplicate --only telemetry_live.main_v4 --dry_run --date=2019-09-19 --parallelism=24 --slices=24` for me in from 26 seconds to 9 seconds.

optimizations:
- distribute `list_tables` calls to the thread pool
- skip `list_datasets` call when `--only` doesn't have globs in the datasets
- skip `list_tables` calls when `--only` doesn't have globs at all
- only compile glob patterns once, into a single regex
- distribute `get_query_job_configs` calls to 

this also makes `stable_table` a string, which is more readable for logging output
~~and fixes a bug where the `end_time` parameter was named  `start_time`~~ (moved to #383)

### DIFF
--- a/script/copy_deduplicate
+++ b/script/copy_deduplicate
@@ -12,10 +12,12 @@ or to process only a specific list of tables.
 
 from argparse import ArgumentParser
 from datetime import datetime, timedelta
-from fnmatch import fnmatch
 from itertools import groupby
 from multiprocessing.pool import ThreadPool
 from uuid import uuid4
+
+import fnmatch
+import re
 
 from google.cloud import bigquery
 
@@ -192,7 +194,9 @@ def sql_full_table_id(table):
     return f"{table.project}.{table.dataset_id}.{table.table_id}"
 
 
-def get_query_job_configs(client, stable_table, date, dry_run, slices, priority):
+def get_query_job_configs(client, live_table, date, dry_run, slices, priority):
+    sql = QUERY_TEMPLATE.format(live_table=live_table)
+    stable_table = f"{live_table.replace('_live.', '_stable.', 1)}${date:%Y%m%d}"
     kwargs = dict(use_legacy_sql=False, dry_run=dry_run, priority=priority)
     start_time = datetime(*date.timetuple()[:6])
     end_time = start_time + timedelta(days=1)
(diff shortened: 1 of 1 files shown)


script/copy_deduplicate
```diff
@@ -12,12 +12,13 @@ or to process only a specific list of tables.
...
 from datetime import datetime, timedelta
-from fnmatch import fnmatch
 from itertools import groupby
+from google.cloud import bigquery
 from multiprocessing.pool import ThreadPool
 from uuid import uuid4
 
-from google.cloud import bigquery
+import fnmatch
+import re
```
jklukas: I'm not sure why these were moved around. This was previously following the convention of having stdlib imports together alphabetically, blank line, then external imports.

script/copy_deduplicate
```diff
@@ -266,49 +270,85 @@ def copy_join_parts(client, stable_table, query_jobs):
...
+    return pattern.split(".", 1)[0]
+
+
+def compile_glob_patterns(patterns):
+    return re.compile("|".join(fnmatch.translate(pattern) for pattern in patterns))
+
+
+def glob_predicate(match, table, arg):
+    matched = match(table) is not None
+    if (arg == "only") ^ matched:
```
jklukas: Can this be expressed without `^`, or can we add a comment to explain what that operator is doing here?

script/copy_deduplicate
```diff
@@ -266,49 +270,85 @@ def copy_join_parts(client, stable_table, query_jobs):
...
-                for job_config in get_query_job_configs(
-                    client=client,
-                    stable_table=stable_table,
-                    date=args.date,
-                    dry_run=args.dry_run,
-                    slices=args.slices,
-                    priority=args.priority,
-                )
+    with ThreadPool(args.parallelism) as pool:
+        live_tables = list_live_tables(
```
jklukas: Do we risk running into throttling by parallelizing these bq api calls?

script/copy_deduplicate
relud: i wasn't sure which way to order them, so i picked one. i'll switch it.

script/copy_deduplicate
relud: the client automatically handles retries/backoff, and the relevant quotas are 100req/s per user, and 300 concurrent req/s per user, so unless parallelism is set very high this shouldn't be an issue.

script/copy_deduplicate
relud: sure, i'll change it to:
```suggestion
    if (arg == "only" and not matched) or (arg == "except" and matched):
```

Issue #360, opened 2019-09-17
Generate per-namespace views for errors tables


whd:
As `payload_bytes_error` sees increased usage, we will want to set up user-facing views for accessing the data. Engineering currently has access to these tables via the BQ console, but the plan is to deprecate that access by creating an authorized view that elides some fields. This would also have the property of making the errors data more widely accessible beyond engineering e.g. to re:dash.

Since we provision datasets per namespace (not pipeline family), it would probably be useful to partitions error views by namespace as well so that e.g. `org-mozilla-fenix` users see an errors view in the `org-mozilla-fenix` dataset that has only fenix errors. We could do this in the view by matching on `uri`, probably.

The work of generating and managing these views might be done here or in the infra repo, I'm not sure presently. The ops infra currently calls generate/publish_views as part of schemas deploys, so hooking the view generation for errors into that process seems reasonable.

jklukas:
> The ops infra currently calls generate/publish_views as part of schemas deploys, so hooking the view generation for errors into that process seems reasonable.

Agreed, although these will need to be authorized views, correct? I have not set one up before, so not exactly sure what the process is. It looks like it can all be done via the python api, so this is probably fine.

Do we know what group we want to grant view access to?

whd:
See https://cloud.google.com/bigquery/docs/authorized-views and https://cloud.google.com/bigquery/docs/share-access-views. I believe you need the source dataset to grant access to the dataset+table the view will be in, which the ops infra has permission to do. We recently tested this with https://github.com/mozilla-services/cloudops-infra/pull/1379/files#diff-28af500d7cdc48a0f9b9728d8e886d22R146 and it appears to work.

relud:
also noteworthy is that the view has to be re-granted access any time it changes

mreid-moz:
In general, we no longer need views to make errors data accessible, but it may still be worthwhile to split the errors data up per-namespace to make it easier to look at errors for one namespace at a time.

whd:
A note to update this that errors data is once again (as of late 2021) not generally accessible and so this issue is still relevant.

ANich:
Seems like we could do this via `sql_generator` for the `document_namespace`'s that we care about, querying the already authorized views? cc @akkomar @kik-kik

whd:
https://github.com/mozilla-services/cloudops-infra/pull/6192

PR #388, opened 2019-09-27
Pipeline sql

fbertsch:
I've been writing queries to monitor the pipeline, but I think it may be good to include them all here.


fbertsch:
Needs conversations around:
- Which dataset do these kinds of tables belong in?
- Should we have a query in here using `json_extract_missing_cols`?

templates/pipeline/structured_error_counts_v1/view.sql
```diff
@@ -0,0 +1,45 @@
...
+  SELECT
+    TIMESTAMP_TRUNC(submission_timestamp, HOUR) AS hour,
+    metadata.document_namespace,
+    metadata.document_type,
+    metadata.document_version,
+    COUNT(*) AS ping_count
+  FROM
+    `payload_bytes_decoded.structured_*`
+  WHERE
+    submission_timestamp >= TIMESTAMP_SUB(current_timestamp, INTERVAL 28 * 24 HOUR)
```
fbertsch: The predicate-pushdown for `WHERE hour > ..` works, but the query processor complains if we don't explicitly limit using this kind of filter.

udf_js/json_extract_missing_cols.sql
```diff
@@ -0,0 +1,90 @@
...
+  histogram will be returned as a missing node, rather than the subvalues within the histogram
+  (e.g. values, sum, etc.)
+- Use known_nodes if you're aware of a missing section, like ['simpleMeasurements']
+
+See here for an example usage
+https://sql.telemetry.mozilla.org/queries/64460/source
+ 
+*/
+
+CREATE TEMP FUNCTION udf_js_json_extract_missing_cols (input STRING, indicates_node ARRAY<STRING>, known_nodes ARRAY<STRING>)
```
fbertsch: This function does not handle arrays of structs well, but that's not a common use case.

udf_js/json_extract_missing_cols.sql
```diff
@@ -0,0 +1,90 @@
...
+      udf_js_json_extract_missing_cols(additional_properties, ARRAY[], ARRAY[]) AS no_args,
+      udf_js_json_extract_missing_cols(additional_properties, ARRAY['third'], ARRAY[]) AS indicates_node_arg,
+      udf_js_json_extract_missing_cols(additional_properties, ARRAY[], ARRAY['first']) AS is_node_arg
+    FROM
+      addl_properties )
+    --
+    SELECT
+      assert_sets_equals(no_args, ARRAY['`first`.`second`.`third`', '`first`.`other-second`']),
+      assert_sets_equals(indicates_node_arg, ARRAY['`first`.`second`', '`first`.`other-second`']),
+      assert_sets_equals(no_args, ARRAY['`first`'])
```
jklukas:
```suggestion
      assert_sets_equals(is_node_arg, ARRAY['`first`'])
```

templates/pipeline/structured_detailed_error_counts_v1/view.sql
```diff
@@ -0,0 +1,36 @@
+WITH error_examples AS (
+  SELECT
+    TIMESTAMP_TRUNC(submission_timestamp, HOUR) AS hour,
+    document_namespace,
+    document_type,
+    document_version,
+    error_type,
+    error_message,
+    udf_js_gunzip(ARRAY_AGG(payload)[OFFSET(0)]) AS sample_payload,
```
fbertsch: Open to alternatives that don't require aggregating all the payloads (where is `arbitrary` when you need it?)

tests/test_udfs.py
```diff
@@ -58,6 +58,27 @@
       ' but got ',
       TO_JSON_STRING(actual))),
     TRUE));
+
+CREATE TEMP FUNCTION
+  assert_sets_equals(expected ANY TYPE, actual ANY TYPE) AS (
```
jklukas: Perhaps this should be `assert_array_equals_any_order` since I think this can support multiple identical entries. I don't think this fully handles the case, though, where multiple different entries appear multiple times, since it only checks the overall array lengths match.

fbertsch: Agreed, it's not perfect. We have three options:
1. Ignore duplicate values (essentially, turn the arrays to sets) 
2. Ensure arrays in any order, with dupes, match
3. Fail if any array has dupes (contract that they are sets)

I'd be good with (2.) or (3.). We should avoid (1.) for obvious reasons.

jklukas: I'd recommend that we sort both arrays and then compare that all elements are equal by offset.

Something like:

```
WITH a AS (
  SELECT x FROM UNNEST(arr1) AS x OFFSET AS offset ORDER BY x)
), b AS (
  SELECT x FROM UNNEST(arr2) AS x OFFSET AS offset ORDER BY x)
)
select a join b using (offset)
```

templates/pipeline/structured_detailed_error_counts_v1/view.sql
jklukas: Does ANY_VALUE work here?

tests/test_udfs.py
fbertsch: I'm good with (2.). I'll update the test.

templates/pipeline/structured_detailed_error_counts_v1/view.sql
fbertsch: Aha, that's just what I was looking for, thanks!

fbertsch:
This is failing with a permissions error, which is perhaps expected.

templates/pipeline/structured_error_counts_v1/view.sql
```diff
@@ -0,0 +1,45 @@
...
+), structured_hourly_errors AS (
+  SELECT
+    hour,
+    document_namespace,
+    document_type,
+    document_version,
+    error_type,
+    ping_count + error_counts.error_count AS ping_count,
+    error_counts.error_count,
+    SAFE_DIVIDE(1.0 * error_counts.error_count, error_counts.error_count + ping_count) AS error_ratio
```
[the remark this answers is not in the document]
fbertsch: I'm not following - how does this reduce logic duplication?

fbertsch:
@jklukas I believe these views could be user-facing, since they support the structured endpoint. `debug` could work but I believe these are more than just for debugging, these are for e.g. testing, monitoring, and alerting. They could be put in a `structured-ingestion` dataset to indicate what the support?

templates/pipeline/structured_error_counts_v1/view.sql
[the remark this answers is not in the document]
jklukas:
```
     ping_count + error_counts.error_count AS ping_count,
    error_counts.error_count,
    SAFE_DIVIDE(1.0 * error_counts.error_count, error_counts.error_count + ping_count) AS error_ratio 
```

We use the literal addition `ping_count + error_counts.error_count` twice. I'd prefer to define it once and then use the named column as the denominator when we divide.

jklukas:
> @jklukas I believe these views could be user-facing, since they support the structured endpoint. `debug` could work but I believe these are more than just for debugging, these are for e.g. testing, monitoring, and alerting. They could be put in a `structured-ingestion` dataset to indicate what the support?

cc @jasonthomas, @whd, and @relud on this naming question.

I think I'm inclined to say we just put these in the `telemetry` dataset. But other options are that we could define a `payload_bytes` dataset that contains views on top of tables in `payload_bytes_*`. Or we could call that dataset `monitoring`.

A `monitoring` dataset could also serve as a home for the authorized views on top of the raw bytes tables that hide IP info.

templates/pipeline/structured_error_counts_v1/view.sql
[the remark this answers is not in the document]
fbertsch: Aha, that makes sense. I thought you were referring to the two `SAFE_DIVIDES`. I'll update.

whd:
> I think I'm inclined to say we just put these in the `telemetry` dataset. But other options are that we could define a `payload_bytes` dataset that contains views on top of tables in `payload_bytes_*`. Or we could call that dataset `monitoring`.
> 
> A `monitoring` dataset could also serve as a home for the authorized views on top of the raw bytes tables that hide IP info.

Without having considered this very carefully, the notion of "pipeline family" (e.g. structured, telemetry) is not meant to be user-facing in that it is mostly an implementation detail of the pipeline. In the somewhat related issue of https://github.com/mozilla/bigquery-etl/issues/360 I had outlined what I thought was the way we should present these errors to end users of the pipeline, namely on a per-namespace basis. In that model, we would create these kinds of views on top of the per-namespace (hence, per-dataset) errors views.

The notion of having general (i.e. pipeline-family) monitoring views seems useful, in particular for pings we receive that are so garbled that we cannot map them to a namespace. I see these views as mostly useful to data engineering and operations folks, as most end users would be more concerned with a particular namespace. Given that, I'm currently inclined to agree with :klukas here and simply put the structured error view in the `telemetry` dataset for now (for our convenience, not necessarily as a permanent home for a user-facing view), to avoid dataset proliferation without more careful consideration.

As to the dataset proliferation, all of `structured`, `payload_bytes`, `debug`, `monitoring` seem reasonable. In the case we create a `structured` dataset, we introduce some ambiguity since `telemetry` is overloaded as both a namespace and a pipeline family, while structured is only a pipeline family, and dataset prefixes have hitherto been on a per-namespace basis. Of these I would currently be most inclined toward `monitoring`.

fbertsch:
I don't recommend putting these in `telemetry` as they are structured ingestion only - we don't want to pollute the `telemetry` dataset or namespace any more.

`monitoring` seems acceptable. We will need to authorize the `monitoring` dataset for access to the `payload_bytes_*` tables.

whd:
> I don't recommend putting these in `telemetry` as they are structured ingestion only

Is there something intrinsic to these queries that makes them structured ingestion only? 

> we don't want to pollute the `telemetry` dataset or namespace any more.

What's your opinion of generating these views by namespace instead?

> `monitoring` seems acceptable. We will need to authorize the `monitoring` dataset for access to the `payload_bytes_*` tables.

If we go with a new `monitoring` dataset, it sounds like we'd have the following structure:

1. `monitoring` (shared by all pipeline families) has authorized views into `payload_bytes_*` that preclude access to sensitive fields
2. the view in this PR is built on top of those authorized views, also in the `monitoring` dataset
3. the views needed for #360 are built on top of the authorized views in (1) in either `monitoring` or per-namespace
4. we create user-facing views on top of this PR per-namespace (or we don't, and instead recommend the `monitoring` view be used by end-users, but as you can see I am trying to avoid exposing "pipeline family" in any way)

fbertsch:
> Is there something intrinsic to these queries that makes them structured ingestion only?

There is, they are reading only the `moz-fx-data-shared-prod.payload_bytes_decoded.structured_*` and similar error tables. We could expand these to include _all_ errors, but in the spirit of historical behavior I thought we may want to tackle them separately ;). We could union all telemetry and structured ingestion together here.

> What's your opinion of generating these views by namespace instead?

That could be acceptable, but I'm not sure if it'd be confusing to users to have both data and metadata contained within their namespace.

> If we go with a new `monitoring` dataset, it sounds like we'd have the following structure:
> 
>     1. `monitoring` (shared by all pipeline families) has authorized views into `payload_bytes_*` that preclude access to sensitive fields
> 
>     2. the view in this PR is built on top of those authorized views, also in the `monitoring` dataset
> 
>     3. the views needed for #360 are built on top of the authorized views in (1) in either `monitoring` or per-namespace
> 
>     4. we create user-facing views on top of this PR per-namespace (or we don't, and instead recommend the `monitoring` view be used by end-users, but as you can see I am trying to avoid exposing "pipeline family" in any way)

I think that's a good way to do it, and makes it so we only have to have one authorized view dataset (I'm assuming that's the case if we have other views in other datasets that query the authorized view in `monitoring`). That also lets us move forward with this and manage the per-namespace ones later.

whd:
> That could be acceptable, but I'm not sure if it'd be confusing to users to have both data and metadata contained within their namespace.

My assumption has been that it would be less confusing to users to be able to see their errors (and only their errors) and similar metadata as part of their own namespace/dataset's standard user-facing views. This may be a wrong assumption.

> I think that's a good way to do it, and makes it so we only have to have one authorized view dataset (I'm assuming that's the case if we have other views in other datasets that query the authorized view in monitoring). That also lets us move forward with this and manage the per-namespace ones later.

Ok, absent further opinion from @jklukas, @jasonthomas, or @relud this is how we're likely to proceed. Please file an ops bug for creating this dataset.

jklukas:
I generally like the approach we've arrived at. We'll put the authorized views in a new `monitoring` dataset and then decide later on whether we want views per-namespace. I'd agree it would generally be easier for end users to reason about per-namespace views, but it may be easier for admins to reason about per-pipeline-family views.

So, we create the admin-level view for now and we can look into generating per-namespace views later.

whd:
Aside: it occurs to me that I have some backlogged work to implement beam errors routing to BQ which will likely also end up in the `monitoring` dataset: https://github.com/mozilla/gcp-ingestion/issues/756 has some discussion on that.

fbertsch:
@jklukas we good to merge?

script/dryrun
```diff
@@ -32,6 +33,7 @@ SKIP = {
     "sql/telemetry/core_clients_last_seen_raw_v1/init.sql",
     "sql/telemetry/fxa_users_last_seen_raw_v1/init.sql",
     # Reference table not found
+    "sql/monitoring/structured_detailed_error_counts_v1/view.sql",
```
jklukas: This sounds like an error we care about. Do we understand why this needs to be exempted?

sql/monitoring/structured_detailed_error_counts_v1/view.sql
```diff
@@ -0,0 +1,83 @@
...
+      var unzipped = gunzip.decompress();
+      return binary2String(unzipped);
+    } catch (err) {
+      return binary2String(compressedData);
+    }
+"""
+OPTIONS (
+  library = "gs://moz-fx-data-circleci-tests-bigquery-etl/gunzip.min.js",
+  library = "gs://moz-fx-data-circleci-tests-bigquery-etl/atob.js"
+);
```
jklukas: We cannot use temporary UDFs in a view definition. We have to reference a persistent view instead.

templates/monitoring/structured_detailed_error_counts_v1/view.sql
```diff
@@ -0,0 +1,41 @@
...
+AS
+WITH error_examples AS (
+  SELECT
+    TIMESTAMP_TRUNC(submission_timestamp, HOUR) AS hour,
+    document_namespace,
+    document_type,
+    document_version,
+    error_type,
+    error_message,
+    udf_js_gunzip(ANY_VALUE(payload)) AS sample_payload,
```
jklukas:
```suggestion
    udf_js.gunzip(ANY_VALUE(payload)) AS sample_payload,
```

Cannot use a temporary UDF in a view.

jklukas:
We cannot merge this until we have a `monitoring` dataset in place in both shared-prod and derived-datasets.

whd:
https://github.com/mozilla-services/cloudops-infra/pull/1439 for adding this dataset.

script/dryrun
fbertsch: Right now the view doesn't have access to the `payload_bytes_decoded` dataset. We do certainly care about that, cc @whd for making this an authorized view.

fbertsch: Ah, this error in particular is actually because it references the other view. Offhand I don't see an easy way to dry-run a DAG of queries.

PR #390, opened 2019-09-27
Add min_profile_subsession_counter to clients_daily

tdsmith:
Being able to identify rows in clients_daily that correspond to a profile's first-ever ping is useful.

Will this automatically propagate to clients_last_seen once it starts being populated here?


relud:
note that this won't immediately effect clients_daily, as we are still generating it using the spark job in telemetry-batch-view

relud:
closing due to inactivity

PR #391, opened 2019-09-28
Improve pytest failure output for udf and sql tests

relud:
things this does:
- put `assert_*` udfs in separate files, so that i can use them from a shell, e.g. `(cat assert/false.sql; echo 'SELECT assert_false(TRUE)' | bq query`
- report udf test titles as `{name}#{ordinal}`, e.g. `udf_get_key#1` so we don't need to wrap exceptions to get readable output
- report sql test titles as `{dataset}.{table}:{test}` instead of just `test session`
- allow pytest to filter udf tests, e.g. `pytest udf/get_key.sql` and `pytest -k 'udf_get_key#1'`
- add `udf` pytest marker, so that udf tests can be run via `pytest -m udf`
- fix `ResourceWarning: unclosed file` from pytest-xdist
- put the pytest plugins in `bigquery_etl` so that we don't have to mess with `sys.path`


assert/array_empty.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TEMP FUNCTION
```
jklukas: Would it be a big pain to put this `assert` dir inside `tests`? It seems an odd fit for the top level.

relud: i bet i could work that out

relud: fixed

PR #395, opened 2019-10-02
Filter out some oauth client_ids

jklukas:
See https://github.com/mozilla/bigquery-etl/issues/348#issuecomment-537254498

For a sample day I checked, this filtered out 98.9% of rows, so this
is indeed going to be a nice efficiency boost for queries that hit
what used to be a very large table.


jklukas:
Once this is merged, I'll recreate the existing table to filter out existing rows with these IDs.

templates/telemetry/fxa_oauth_events_v1/query.sql
```diff
@@ -11,5 +11,10 @@ WHERE
   (jsonPayload.type LIKE '%.amplitudeEvent' OR jsonPayload.fields.op = 'amplitudeEvent')
   AND jsonPayload.fields.event_type IS NOT NULL
   AND jsonPayload.fields.user_id IS NOT NULL
+  AND JSON_EXTRACT_SCALAR(jsonPayload.fields.event_properties,
+    '$.oauth_client_id') NOT IN ( -- Per https://github.com/mozilla/bigquery-etl/issues/348
+    '3332a18d142636cb',
+    '5882386c6d801776',
+    '1b1a3e44c54fbb58')
```
jklukas: @irrationalagent - Can you give a quick explanation of the significance of these IDs so I can include as a comment in the code here?

irrationalagent: they are the oauth client ids for fennec sync, desktop sync and firefox for ios sync respectively. we filter them out (1) because of their high volume and (2) because the events are redundant with our cert_signed events.

jklukas: Added comments to document these thoughts. r?

PR #396, opened 2019-10-02
fxa_users_services tables

jklukas:
[Per design discussed in a sheet](https://docs.google.com/spreadsheets/d/1R1UgiD1_Ol-AbSyTcEgz2pnVM3tkISNV6Sch9wnv0DQ/edit#gid=1573769019), this adds new ETL for supporting FxA per user per service metrics.

### DIFF
--- a/sql/telemetry/fxa_content_auth_events_v1/view.sql
+++ b/sql/telemetry/fxa_content_auth_events_v1/view.sql
@@ -17,7 +17,7 @@ WITH
     `timestamp`,
     receiveTimestamp
   FROM
-    `moz-fx-data-derived-datasets.telemetry.fxa_auth_events_v1`
+    `moz-fx-data-derived-datasets.telemetry.fxa_content_events_v1`
   ),
   --
   auth as (
@@ -35,7 +35,7 @@ WITH
     `timestamp`,
     receiveTimestamp
   FROM
-    `moz-fx-data-derived-datasets.telemetry.fxa_content_events_v1`
+    `moz-fx-data-derived-datasets.telemetry.fxa_auth_events_v1`
   ),
   --
   unioned AS (
--- a/sql/telemetry/fxa_content_auth_events_v1/view.sql
+++ b/sql/telemetry/fxa_content_auth_events_v1/view.sql
@@ -6,6 +6,7 @@ WITH
   SELECT
     jsonPayload.logger,
     jsonPayload.fields.event_type,
+    jsonPayload.fields.app_version,
     jsonPayload.fields.os_name,
     jsonPayload.fields.os_version,
     jsonPayload.fields.country,
@@ -23,6 +24,7 @@ WITH
   SELECT
     jsonPayload.logger,
     jsonPayload.fields.event_type,
+    jsonPayload.fields.app_version,
     jsonPayload.fields.os_name,
     jsonPayload.fields.os_version,
     jsonPayload.fields.country,
--- a/sql/telemetry/fxa_users_services_daily/view.sql
+++ b/sql/telemetry/fxa_users_services_daily/view.sql
@@ -2,10 +2,10 @@ CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.telemetry.fxa_users_services_daily` AS
 SELECT
   * REPLACE(cn.code AS country),
-  ls.country AS country_name
+  fxa.country AS country_name
 FROM
-  `moz-fx-data-shared-prod.telemetry_derived.fxa_users_services_daily_v1` ls
+  `moz-fx-data-shared-prod.telemetry_derived.fxa_users_services_daily_v1` AS fxa
 LEFT JOIN
-  `moz-fx-data-shared-prod.static.country_names_v1` cn
+  `moz-fx-data-shared-prod.static.country_names_v1` AS cn
 ON
-  cn.name = ls.country
+  cn.name = fxa.country
--- a/sql/telemetry/fxa_users_services_daily/view.sql
+++ b/sql/telemetry/fxa_users_services_daily/view.sql
@@ -1,0 +1,11 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.fxa_users_services_daily` AS
+SELECT
+  * REPLACE(cn.code AS country),
+  ls.country AS country_name
+FROM
+  `moz-fx-data-shared-prod.telemetry_derived.fxa_users_services_daily_v1` ls
+LEFT JOIN
+  `moz-fx-data-shared-prod.static.country_names_v1` cn
+ON
+  cn.name = ls.country
--- a/sql/telemetry_derived/fxa_users_services_daily_v1/query.sql
+++ b/sql/telemetry_derived/fxa_users_services_daily_v1/query.sql
@@ -44,6 +44,31 @@ CREATE TEMP FUNCTION
         'fxa_reg - complete')) );
   --
 WITH
+  base AS (
+    SELECT
+      * REPLACE(
+        -- cert_signed is specific to sync, but these events do not have the
+        -- 'service' field populated, so we fill in the service name for this special case.
+        IF(service IS NULL AND event_type = 'fxa_activity - cert_signed', 'sync', service) AS service)
+    FROM
+      `moz-fx-data-shared-prod.telemetry.fxa_content_auth_oauth_events_v1`
+    WHERE
+      user_id IS NOT NULL
+      AND event_type NOT IN ( --
+      'fxa_email - bounced',
+      'fxa_email - click',
+      'fxa_email - sent',
+      'fxa_reg - password_blocked',
+      'fxa_reg - password_common',
+      'fxa_reg - password_enrolled',
+      'fxa_reg - password_missing',
+      'fxa_sms - sent',
+      'mktg - email_click',
+      'mktg - email_open',
+      'mktg - email_sent',
+      'sync - repair_success',
+      'sync - repair_triggered')
+  )
   windowed AS (
   SELECT
     DATE(`timestamp`) AS submission_date,
@@ -58,24 +83,9 @@ WITH
     udf_contains_tier1_country(ARRAY_AGG(country) OVER w1) AS seen_in_tier1_country,
     udf_contains_registration(ARRAY_AGG(event_type) OVER w1) AS registered
   FROM
-    `moz-fx-data-shared-prod.telemetry.fxa_content_auth_oauth_events_v1`
+    base
   WHERE
-    user_id IS NOT NULL
-    AND service IS NOT NULL
-    AND event_type NOT IN ( --
-      'fxa_email - bounced',
-      'fxa_email - click',
-      'fxa_email - sent',
-      'fxa_reg - password_blocked',
-      'fxa_reg - password_common',
-      'fxa_reg - password_enrolled',
-      'fxa_reg - password_missing',
-      'fxa_sms - sent',
-      'mktg - email_click',
-      'mktg - email_open',
-      'mktg - email_sent',
-      'sync - repair_success',
-      'sync - repair_triggered')
+    service IS NOT NULL
     -- Reprocess all dates by running this query with --parameter=submission_date:DATE:NULL
     AND (@submission_date IS NULL OR @submission_date = DATE(`timestamp`))
   WINDOW
--- a/sql/telemetry_derived/fxa_users_services_daily_v1/query.sql
+++ b/sql/telemetry_derived/fxa_users_services_daily_v1/query.sql
@@ -1,0 +1,104 @@
+CREATE TEMP FUNCTION
+  udf_mode_last(list ANY TYPE) AS ((
+    SELECT
+      _value
+    FROM
+      UNNEST(list) AS _value
+    WITH
+    OFFSET
+      AS
+    _offset
+    GROUP BY
+      _value
+    ORDER BY
+      COUNT(_value) DESC,
+      MAX(_offset) DESC
+    LIMIT
+      1 ));
+--
+CREATE TEMP FUNCTION
+  udf_contains_tier1_country(x ANY TYPE) AS ( --
+    EXISTS(
+    SELECT
+      country
+    FROM
+      UNNEST(x) AS country
+    WHERE
+      country IN ( --
+        'United States',
+        'France',
+        'Germany',
+        'United Kingdom',
+        'Canada')) );
+  --
+  -- This UDF is also only applicable in the context of this query.
+CREATE TEMP FUNCTION
+  udf_contains_registration(x ANY TYPE) AS ( --
+    EXISTS(
+    SELECT
+      event_type
+    FROM
+      UNNEST(x) AS event_type
+    WHERE
+      event_type IN ( --
+        'fxa_reg - complete')) );
+  --
+WITH
+  windowed AS (
+  SELECT
+    DATE(`timestamp`) AS submission_date,
+    user_id,
+    service,
+    ROW_NUMBER() OVER w1_unframed AS _n,
+    udf_mode_last(ARRAY_AGG(country) OVER w1) AS country,
+    udf_mode_last(ARRAY_AGG(language) OVER w1) AS language,
+    udf_mode_last(ARRAY_AGG(app_version) OVER w1) AS app_version,
+    udf_mode_last(ARRAY_AGG(os_name) OVER w1) AS os_name,
+    udf_mode_last(ARRAY_AGG(os_version) OVER w1) AS os_version,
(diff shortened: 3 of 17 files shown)


sql/telemetry/fxa_content_auth_events_v1/view.sql
```diff
@@ -6,6 +6,7 @@ WITH
   SELECT
     jsonPayload.logger,
     jsonPayload.fields.event_type,
+    jsonPayload.fields.app_version,
```
jklukas: We include app_version in the existing fxa_users_daily, etc. queries. Seems we might as well keep that dimension, so need to include in the views here.

sql/telemetry_derived/fxa_users_services_daily_v1/query.sql
```diff
@@ -0,0 +1,104 @@
+CREATE TEMP FUNCTION
```
jklukas: This query is basically identical to fxa_users_daily but windows by `user_id, service` rather than `user_id`.

templates/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
```diff
@@ -0,0 +1,80 @@
+CREATE TABLE
```
jklukas: I think it's going to be a bit tricky to actually make this query incremental. I think we can delay that, though. This `init.sql` query which creates the table from scratch isn't ridiculously expensive; it scans 2.1 TB, so costs ~$10. For now, my plan is to schedule this query to completely recreate the table every day. We can refactor later to make it more efficient if needed.

templates/telemetry_derived/fxa_users_services_last_seen_v1/query.sql
```diff
@@ -0,0 +1,78 @@
...
+    (_current.user_id IS NOT NULL,
+      _current,
+      _previous).* REPLACE ( --
+      udf_combine_adjacent_days_bits(_previous.days_seen_bits,
+        _current.days_seen_bits) AS days_seen_bits,
+      udf_combine_adjacent_days_bits(_previous.days_seen_in_tier1_country_bits,
+        _current.days_seen_in_tier1_country_bits) AS days_seen_in_tier1_country_bits,
+      udf_coalesce_adjacent_days_bits(_previous.days_registered_bits,
+        _current.days_registered_bits) AS days_registered_bits ),
+    _previous.user_id IS NULL AS new_observation -- used for determining "resurrected" criteria
```
jklukas: Please take an early look at the "resurrected" logic below. Basically, if we don't have a "previous" entry, it means the user hasn't been seen for at least 28 days and has fallen off the radar. We use joins below to check if each user has been seen before.

templates/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
irrationalagent: Sounds good. The logic of the query is a bit convoluted. When I have a chance I'll think about whether there is a smarter way to write it that would make it easier to "incrementalize". 

In the abstract its a table that *should* be incremental, in the sense that a given user-service combination can only have one first date, and that date can't change. But the contortions that we have to go through to get those dates makes it hard.

templates/telemetry_derived/fxa_users_services_last_seen_v1/query.sql
irrationalagent: This seems to make sense to me. One ambiguity (that we may have to live with) that will result from this is that existing users who have not had any activity since before 2019-03-01 (and therefore will not be present in      fxa_users_services_first_seen_v1) will not be counted as resurrected. I don't know how many users this will affect, but it will lead to us under-counting the true number of resurrected users.

Again, I think this is the best we can do, but it may be worth documenting.

sql/telemetry/fxa_content_auth_events_v1/view.sql
```diff
@@ -6,6 +6,7 @@ WITH
   SELECT
```
relud: both `fxa_content_auth_events_v1/view.sql` and `fxa_content_auth_oauth_events_v1/view.sql` query the wrong table the CTE for auth and content (auth queries content, and content queries auth). it doesn't impact the output, and it's not directly related to this pr, but it's something we should fix.

sql/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
```diff
@@ -0,0 +1,80 @@
...
+CLUSTER BY service, user_id AS
+WITH
+  -- use a window function to look within each USER and SERVICE for the first value of service, os, and country.
+  -- also, get the first value of flow_id for later use and create a boolean column that is true if the first instance of a service usage includes a registration.
+  first_services AS (
+  SELECT
+    user_id,
+    FIRST_VALUE(service) OVER (PARTITION BY user_id, service ORDER BY `timestamp`) AS service,
+    FIRST_VALUE(`timestamp`) OVER (PARTITION BY user_id, service ORDER BY `timestamp`) AS first_service_timestamp,
+    FIRST_VALUE(os_name) OVER (PARTITION BY user_id, service ORDER BY `timestamp`) AS first_service_os,
```
relud: I think this should either do `first_mode` aggregation or ```ORDER BY `timestamp`, os_name``` to make it idempotent for events with matching timestamps

sql/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
```diff
@@ -0,0 +1,80 @@
...
+  -- this means we have to use first_service_flow to join back on the original source table's flow_id,
+  -- and take the first occurrence of `entrypoint` within the flow that the user first appeared in the service on.
+  flows AS (
+  SELECT
+    DISTINCT s.first_service_flow,
+    FIRST_VALUE(f.entrypoint) OVER (PARTITION BY f.flow_id ORDER BY f.`timestamp`) AS first_service_entrypoint
+  FROM
+    first_services_g s
+  INNER JOIN
+    `moz-fx-data-derived-datasets.telemetry.fxa_content_auth_oauth_events_v1` f
```
relud: Table aliases throughout this PR should use `AS` explicitly, to be consistent with column aliases.
```suggestion
    `moz-fx-data-derived-datasets.telemetry.fxa_content_auth_oauth_events_v1` AS f
```

sql/telemetry/fxa_users_services_daily/view.sql
```diff
@@ -0,0 +1,11 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.fxa_users_services_daily` AS
+SELECT
+  * REPLACE(cn.code AS country),
+  ls.country AS country_name
+FROM
+  `moz-fx-data-shared-prod.telemetry_derived.fxa_users_services_daily_v1` ls
```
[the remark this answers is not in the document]
jklukas: I will use alias `fxa` instead of `ls` in both of these views.

sql/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
```diff
@@ -0,0 +1,80 @@
...
+  -- use a window function to look within each USER and SERVICE for the first value of service, os, and country.
+  -- also, get the first value of flow_id for later use and create a boolean column that is true if the first instance of a service usage includes a registration.
+  first_services AS (
+  SELECT
+    user_id,
+    FIRST_VALUE(service) OVER (PARTITION BY user_id, service ORDER BY `timestamp`) AS service,
+    FIRST_VALUE(`timestamp`) OVER (PARTITION BY user_id, service ORDER BY `timestamp`) AS first_service_timestamp,
+    FIRST_VALUE(os_name) OVER (PARTITION BY user_id, service ORDER BY `timestamp`) AS first_service_os,
+    FIRST_VALUE(country) OVER (PARTITION BY user_id, service ORDER BY `timestamp`) AS first_service_country,
+    FIRST_VALUE(flow_id) OVER (PARTITION BY user_id, service ORDER BY `timestamp`) AS first_service_flow,
```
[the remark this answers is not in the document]
irrationalagent: yes I think its what we want, because some events do not have all of these properties set.

sql/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
```diff
@@ -0,0 +1,80 @@
...
+    LOGICAL_OR(
+      CASE
+        WHEN event_type = 'fxa_reg - complete' THEN TRUE
+        ELSE FALSE
+      END
+    ) OVER (PARTITION BY user_id, service ORDER BY `timestamp`) AS did_register -- probably doesn't need to be a window function
```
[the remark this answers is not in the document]
irrationalagent: Yes, the registration event does indicate registering with FxA generally, not the service. Removing service from the partition clause would still indicate the service that the user generated the registration event on, so I think we want to go with the latter option then.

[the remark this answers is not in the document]
relud: > Removing service from the partition clause would still indicate the service that the user generated the registration event on

that's what i would expect if we *keep* service in the partition clause.

sql/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
```diff
@@ -0,0 +1,80 @@
...
+WITH
+  -- use a window function to look within each USER and SERVICE for the first value of service, os, and country.
+  -- also, get the first value of flow_id for later use and create a boolean column that is true if the first instance of a service usage includes a registration.
+  first_services AS (
+  SELECT
+    user_id,
+    FIRST_VALUE(service) OVER (PARTITION BY user_id, service ORDER BY `timestamp`) AS service,
+    FIRST_VALUE(`timestamp`) OVER (PARTITION BY user_id, service ORDER BY `timestamp`) AS first_service_timestamp,
+    FIRST_VALUE(os_name) OVER (PARTITION BY user_id, service ORDER BY `timestamp`) AS first_service_os,
+    FIRST_VALUE(country) OVER (PARTITION BY user_id, service ORDER BY `timestamp`) AS first_service_country,
```
relud: these window clauses need a frame that includes all rows, so that they are able to collect the first non-null value from rows that may come after the first row.

sql/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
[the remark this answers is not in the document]
irrationalagent: OK I think i misunderstood then, and this is the reason i originally had it in the partition clause. So then I take it you are recommending 

> or the GROUP BY clause in first_services_g should be replaced with a ROW_NUMBER() ... AS _n here and WHERE _n = 1 in first_services_g

which I'm not sure I quite follow

irrationalagent:
So I realized something about `fxa_users_services_last_seen`. The majority of recurring sync events are of event_type `fxa_activity - cert_signed` which will not have a service property in most cases (i.e. service will be null). This means that users that only generate `fxa_activity - cert_signed` on a given day will not be attributed to service = sync in this table. This will happen frequently because the only thing many users do on a given day is sync.

I was able to partially get around this while writing queries by doing an anti-join on `fxa_users_last_seen_v1`, for example: 

```sql
WITH
  services_last_seen as (
    SELECT
      submission_date,
      user_id,
      service
    FROM 
      `moz-fx-data-shared-prod.telemetry.fxa_users_services_last_seen`
    WHERE submission_date  >= date(2019,3,1)
  ),
    
  sync AS (
    SELECT
      ls.submission_date,
      ls.user_id,
      'sync' as service
    FROM 
      `moz-fx-data-shared-prod.telemetry.fxa_users_last_seen_v1` ls
    WHERE submission_date >= date(2019,3,1)
  ),
  
  last_seen_complete as (
    SELECT
      ls.submission_date,
      ls.user_id,
      ls.service
    FROM sync ls
    LEFT JOIN services_last_seen sls
    ON (ls.user_id = sls.user_id and ls.submission_date = sls.submission_date and ls.service = sls.service)
    WHERE (sls.user_id IS NULL and sls.submission_date IS NULL and sls.service IS NULL)
  ),
```
This works because `fxa_activity - cert_signed` is specific to sync, so we can assume that users generating this event are on sync. However this anti-join does not fix the fact that the resurrected variable will be inaccurate.

right now I'm unsure of the best way to fix this. one way would be to just incorporate a similar anti-join as above into the etl. i doubt that's the best way though. another way would be to hard-code service as sync for all cases it is null on the `fxa_activity - cert_signed` event, upstream in the etl. Maybe that would be better.

sql/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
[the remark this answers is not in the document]
relud: it means this change:

```diff
diff --git a/templates/telemetry_derived/fxa_users_services_first_seen_v1/init.sql b/templates/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
index e16acfb..d0478c2 100644
--- a/templates/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
+++ b/templates/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
@@ -7,6 +7,7 @@ WITH
   -- also, get the first value of flow_id for later use and create a boolean column that is true if the first instance of a service usage includes a registration.
   first_services AS (
   SELECT
+    ROW_NUMBER() OVER (PARTITION BY user_id, service ORDER BY `timestamp`) AS _n,
     user_id,
     FIRST_VALUE(service) OVER (PARTITION BY user_id, service ORDER BY `timestamp`) AS service,
     FIRST_VALUE(`timestamp`) OVER (PARTITION BY user_id, service ORDER BY `timestamp`) AS first_service_timestamp,
@@ -32,22 +33,11 @@ WITH
   -- I've verified that `date(first_service_timestamp), count(distinct user_id) where did_register = true group by 1`  matches the counts of registrations per day in amplitude.
   first_services_g AS (
   SELECT
-    user_id,
-    service,
-    first_service_timestamp,
-    first_service_os,
-    first_service_country,
-    first_service_flow,
-    LOGICAL_OR(did_register) AS did_register
+    * EXCEPT(_n)
   FROM
     first_services
-  GROUP BY
-    user_id,
-    service,
-    first_service_timestamp,
-    first_service_os,
-    first_service_country,
-    first_service_flow ),
+  WHERE
+    _n = 1),
   -- sadly, `entrypoint` is null on registration complete and login complete events.
   -- this means we have to use first_service_flow to join back on the original source table's flow_id,
   -- and take the first occurrence of `entrypoint` within the flow that the user first appeared in the service on.
```

relud:
> another way would be to hard-code service as sync for all cases it is null on the fxa_activity - cert_signed event, upstream in the etl. Maybe that would be better.

definitely sounds better

sql/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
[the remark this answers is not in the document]
jklukas: @relud's suggestion here makes sense to me, so I'm implementing that for now, but feel free to continue discussion here.

sql/telemetry_derived/fxa_users_services_daily_v1/query.sql
```diff
@@ -0,0 +1,114 @@
...
+      event_type IN ( --
+        'fxa_reg - complete')) );
+  --
+WITH
+  base AS (
+    SELECT
+      * REPLACE(
+        -- cert_signed is specific to sync, but these events do not have the
+        -- 'service' field populated, so we fill in the service name for this special case.
+        IF(service IS NULL AND event_type = 'fxa_activity - cert_signed', 'sync', service) AS service)
```
jklukas: @irrationalagent This is an attempt to address your concern about cert_signed events not being attributable to sync. This inserts 'sync' as the service at the point where we populate the daily table, and then would flow through to last_seen.

sql/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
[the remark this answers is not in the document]
irrationalagent: thanks for clarifying @relud this makes sense to me now

sql/telemetry_derived/fxa_users_services_daily_v1/query.sql
irrationalagent: yes, this looks like a good solution. if in the future we fix things on FxA's side to start populating service on this event then we can remove this (though it won't break anything if we do make the fix)

jklukas:
This is merged now. I've recreated the first_seen and daily tables based on the merged code and am now backfilling `fxa_users_services_last_seen_v1`.

irrationalagent:
yea yea thanks @jklukas ! you have no idea how useful these tables will be, not just for skyline but going forward indefinitely. cc @ksiegler1

PR #398, opened 2019-10-03
Add udf to deanonymize event tuples

sunahsuh:
Whoo, go @acmiyaguchi for getting event tuples working in BQ! This UDF renames the anonymous struct fields in event tuples into meaningful names

### DIFF
--- a/udf/deanonymize_event.sql
+++ b/udf/deanonymize_event.sql
@@ -11,25 +11,25 @@ CREATE TEMP FUNCTION
     f3_ STRING,
     f4_ STRING,
     f5_ ARRAY<STRUCT<key STRING, value STRING>>>) AS (
-    (SELECT
-      AS STRUCT tuple.f0_ AS `timestamp`,
-      tuple.f1_ AS category,
-      tuple.f2_ AS method,
-      tuple.f3_ AS object,
-      tuple.f4_ AS string_value,
-      tuple.f5_ AS map_values)
+
+    STRUCT(tuple.f0_ AS event_timestamp,
+      tuple.f1_ AS event_category,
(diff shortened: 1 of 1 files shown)


udf/deanonymize_event.sql
```diff
@@ -0,0 +1,43 @@
...
+CREATE TEMP FUNCTION
+  udf_deanonymize_event(tuple STRUCT<
+    f0_ INT64,
+    f1_ STRING,
+    f2_ STRING,
+    f3_ STRING,
+    f4_ STRING,
+    f5_ ARRAY<STRUCT<key STRING, value STRING>>>) AS (
+    (SELECT
+      AS STRUCT tuple.f0_ AS `timestamp`,
```
jklukas: Elsewhere, we have prefixed all of these field names with `event_` which has some benefits in avoiding reserved words like `timestamp` and making the purpose of the field more obvious when unnesting. Should we do so here?

Also, we should add calls to this function into the views for `main` and other pings, but that can be a separate PR.

sunahsuh: I left these as unprefixed since these fields are still encapsulated within a struct, and it seemed a bit redundant in that context. I can see the argument for the adding the prefix, though for the flattened events dataset my plan was to manually add the prefix in the new query

sunahsuh: Although tbf the javascript udf _that I wrote a few weeks ago_ that parsed out events had the prefix and that output was also a struct, so I guess I have to go discuss this with my past self

relud: this should use `STRUCT(` instead of `(SELECT AS STRUCT`, because BigQuery can only handle a limited number of expression subqueries.

PR #407, opened 2019-10-08
Fix some windowing bugs

relud:


sql/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
```diff
@@ -44,12 +45,23 @@ WITH
...
+    w1_reversed AS (
       PARTITION BY
         user_id,
         service
       ORDER BY
-        `timestamp` ) ),
+        `timestamp` DESC
+      ROWS BETWEEN
+        UNBOUNDED PRECEDING
+        AND UNBOUNDED FOLLOWING),
```
relud: this was missing a frame, preventing the window from aggregating more than one row (because the default frame is `ROWS UNBOUNDED PRECEDING`)

sql/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
```diff
@@ -44,12 +45,23 @@ WITH
...
   WINDOW
-    w1 AS (
+    -- We must provide a window with `ORDER BY timestamp DESC` so that udf_mode_last actually aggregates mode first.
+    w1_reversed AS (
       PARTITION BY
         user_id,
         service
       ORDER BY
-        `timestamp` ) ),
+        `timestamp` DESC
```
relud: this needs to be `DESC` so that `mode_last` becomes `mode_first`

jklukas:
cc @irrationalagent I'm restarting the backfill in order to pull in these changes

PR #411, opened 2019-10-08
change fxa_users_services_first_seen_v1 to account for service being null for most sync events

irrationalagent:
should have caught this before Jeff merged his PR, sorry :(. I think this means will will have to run the backfill again :(

this is similar to problem jeff addressed as part of https://github.com/mozilla/bigquery-etl/pull/396 that adds 'sync' for cert_signed events. I should have realized that we would have to do it for this table too. 

I'm not sure about the scope of the window clause but I hazarded a guess that I needed to repeat the `IF(` clause there. there may be a better way of doing it. 

we shouldn't have to worry about nulls creeping into service here because all the other `fxa_activity - *` events will have a non-null value for service. 

cc @ksiegler1

### DIFF
--- a/sql/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
+++ b/sql/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
@@ -48,7 +48,7 @@ WITH
     w1_reversed AS (
       PARTITION BY
         user_id,
-        IF(service IS NULL AND event_type = 'fxa_activity - cert_signed', 'sync', service) AS service
+        IF(service IS NULL AND event_type = 'fxa_activity - cert_signed', 'sync', service)
       ORDER BY
         `timestamp` DESC
       ROWS BETWEEN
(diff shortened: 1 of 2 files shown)


irrationalagent:
I should be more specific about why we want this change. This table is supposed to code the first date (after 2019-03-01) that we saw each user on each service they have used. Since the `fxa_activity - cert_signed` event only occurs for sync, but lacks the `service` property, as the table stands today it will incorrectly encode the first date of sync usage for some users. And some users might not show up with sync at all on this table, if they have never explicitly logged into sync but were already logged in before 2019-03-01 (this is the bigger problem).

sql/telemetry_derived/fxa_users_services_first_seen_v1/init.sql
```diff
@@ -39,17 +39,16 @@ WITH
   FROM
     `moz-fx-data-derived-datasets.telemetry.fxa_content_auth_oauth_events_v1`
   WHERE
-    ((event_type IN ('fxa_login - complete', 'fxa_reg - complete'))
+    ((event_type IN ('fxa_login - complete', 'fxa_reg - complete') AND service IS NOT NULL)
```
jklukas: The where clause doesn't have access to the redefined `service` field, so this is using the raw value of `service` and will still filter out the sync events we want to include. I'm going to push another commit to factor the redefinition of `service` to a base CTE, and then we can use this filter here.

jklukas:
I've recreated first_seen and kicked off the backfill for last_seen again. Should be fully refreshed by EOD.

irrationalagent:
Thanks (again) for all the time spent on this Jeff. much much much appreciated

PR #413, opened 2019-10-09
Add --preceding-days option to copy_deduplicate

relud:


script/copy_deduplicate
```diff
@@ -32,10 +32,15 @@ WITH
...
+    DATE(submission_timestamp) >= DATE_SUB(
+        DATE(@start_time),
+        INTERVAL @num_preceding_days DAY
+    )
     AND submission_timestamp < @end_time
   GROUP BY
-    document_id),
+    document_id
+  HAVING
+    submission_timestamp >= @start_time),
```
jklukas: It took me a while to understand this HAVING clause. It might be easier to read through this query if we name the column `MIN(submission_timestamp) AS first_submission_timestamp`, so that we're referencing `first_submission_timestamp` here. Then we'd rename it back to `submission_timestamp` in the final `SELECT`. I suppose that introduces some small risk of a name collision, though.

relud:
I copied `sample_id = 42` of `moz-fx-data-shared-prod.telemetry_live.main_v4` for `2019-09-18` and `2019-09-19` to `relud-17123.telemetry_live.main_v4`. for `2019-09-19` this resulted in 2,958,134 rows and 96,302,891,653 bytes.


I then ran `script/copy_deduplicate --only=telemetry_live.main_v4 --date=2019-09-19 --project_id=relud-17123` and `script/copy_deduplicate --only=telemetry_live.main_v4 --date=2019-09-19 --project_id=relud-17123 --preceding-days=1`. The first scanned 96,302,891,653 bytes (which exactly matches `numBytes` for the input partition), ran for 112 seconds, and removed 818 rows, while `--preceding-days=1` scanned 96,576,863,697 bytes, ran in 120 seconds, and removed 5,033 rows.

So `--preceding-days=1` resulted in ~0.28% (273,972,044 bytes) more data scanned, ~7.1% (9 seconds) longer run time, and removed 615% (4,215) more rows == output ~0.14% fewer rows.

jklukas:
The 615% increase in number of removed rows seems very high. I had expected something well under a 100% increase. Do you have any sense of why we remove more rows based on the previous day's docIds compared to the current day?

relud:
> The 615% increase in number of removed rows seems very high. I had expected something well under a 100% increase. Do you have any sense of why we remove more rows based on the previous day's docIds compared to the current day?

the dataflow deduplication is catching most of the duplicates in a rolling 24 hour window, and just outside the 24-hour window the number of duplicates reaching bigquery spikes, which this now catches and removes some of.

script/copy_deduplicate
relud: how about just adding this comment:
```suggestion
    -- HAVING applies *after* aggregation, so remove document_ids first seen before @start_time
    submission_timestamp >= @start_time),
```

PR #415, opened 2019-10-09
Remove default views

jklukas:
We have integrated generation of default views into our schema deployment
pipeline and we have decided that we want to rely on that so that new tables
are automatically presented to users as views without manual intervention in
this repo.

This PR removes all of the views whose text exactly matches what the pipeline
produces as the default. Views like sync ping and core where we've made changes
are not removed here.

Removing the views here should be a no-op as far what the schema propagation
pipeline actually publishes.


jklukas:
To generate this PR, I took the `publish_views` script and modified it a bit. The script lists stable tables, then checks whether the sql it would generate for a view matches what's already in the repo. If they match, we delete the file. If they differ, we leave the file alone.

PR #423, opened 2019-10-12
DOC:Improvements

cyberxl512:
Fix the bold statements and also a dividing line.

peace!


jklukas:
Hi @cyberxl512. Thanks for your interest in improving docs here. Could you tell use more about what appears to be broken without this change? AFAICT, the existing README renders well in GitHub markdown. Are you using a different rendering engine?

PR #426, opened 2019-10-16
Support many billing projects and dates in copy_deduplicate

relud:
still needs to be tested


script/copy_deduplicate
```diff
@@ -135,6 +139,15 @@ parser.add_argument(
...
     help="Number of days preceding --date that should be used to filter out duplicates",
 )
+parser.add_argument(
+    "--billing_projects",
+    "--billing-projects",
+    "--billing_project",
+    "--billing-project",
+    nargs="+",
+    default=[None],
+    help="One or more billing projects over which bigquery jobs should be distributed",
```
jklukas: How does this interact with the project_id param? The docs on project_id (ID of the project in which to run query jobs) no longer apply.

relud: `--project_id` was never `(ID of the project in which to run query jobs)`, rather it was `ID of the project in which to find tables`. will update accordingly

script/copy_deduplicate
```diff
@@ -135,6 +139,15 @@ parser.add_argument(
     default=0,
     help="Number of days preceding --date that should be used to filter out duplicates",
 )
+parser.add_argument(
+    "--billing_projects",
+    "--billing-projects",
+    "--billing_project",
+    "--billing-project",
+    nargs="+",
+    default=[None],
```
jklukas: Could we add a comment here indicating the significance of this default? It took me a while to realize the single None would be passed literally to `Client`. And probably also indicate in the help string what to expect if this is not set.

script/copy_deduplicate
```diff
@@ -358,47 +373,72 @@ def list_live_tables(client, pool, project_id, except_tables, only_tables):
...
-                        live_table,
-                        args.date,
-                        args.dry_run,
-                        args.slices,
-                        args.priority,
-                        args.preceding_days,
-                    )
-                    for live_table in live_tables
-                ],
+        with client_q.client() as client:
```
jklukas: These two context managers can be combined in one statement to reduce nesting. I think the syntax is:

    with ThreadPool(args.parallelism) as pool, client_q.client() as client:

jklukas: Or maybe they have different lifetimes here.

relud: they have different lifetimes

script/copy_deduplicate
```diff
@@ -358,47 +373,72 @@ def list_live_tables(client, pool, project_id, except_tables, only_tables):
...
+            ]
+
+        # preserve query_jobs order so results stay sorted by stable_table for groupby
+        results = pool.starmap(client_q.with_client, query_jobs, chunksize=1)
+        copy_jobs = [
+            (copy_join_parts, stable_table, [query_job for _, query_job in group])
             for stable_table, group in groupby(results, key=lambda result: result[0])
         ]
-        pool.starmap(copy_join_parts, copy_args)
+        pool.starmap(client_q.with_client, copy_jobs, chunksize=1)
```
jklukas: I just tried running this on one of the main_v4 backfill tables, and it looks like it ran all 100 slices across 8 projects as I had configured, but then failed with

```
  File "./heka-backfill/py-multi-copydedupe", line 447, in <module>
    main()
  File "./heka-backfill/py-multi-copydedupe", line 441, in main
    for stable_table, group in groupby(results, key=lambda result: result[0])
  File "./heka-backfill/py-multi-copydedupe", line 440, in <listcomp>
    (copy_join_parts, stable_table, [query_job for _, query_job in group])
  File "./heka-backfill/py-multi-copydedupe", line 441, in <lambda>
    for stable_table, group in groupby(results, key=lambda result: result[0])
TypeError: 'NoneType' object is not subscriptable
```

So perhaps the return value of the query jobs has changed.

relud: fixed

PR #428, opened 2019-10-16
Add views for search_derived dataset

relud:


sunahsuh:
Looks like we're dependent on mozilla-services/cloudops-infra#1494 for dry-run-sql to pass

PR #431, opened 2019-10-18
add oauth client id filter to fxa auth server etl

irrationalagent:
in the last fxa train the oauth server was merged into the auth server. previously, we were filtering some high volume events out from the oauth server:

https://github.com/mozilla/bigquery-etl/blob/51a6443d880e2d1f47a621ca4db0225c5267a5a0/templates/telemetry/fxa_oauth_events_v1/query.sql#L14-L21

im 90% sure that now these events will be coming from the *auth* server rather than *oauth* server, which means we will stop filtering them out. this pr would fix this *i think*. ~but i need to double confirm with the FxA engineers before we go ahead and merge this~,


irrationalagent:
I confirmed that we do need this filter. You can see that the filtered services started to get into yesterday's tables https://console.cloud.google.com/bigquery?sq=630180991450:cbcbfe29b06943309c5022dcd3261e69

irrationalagent:
not sure why the generated sql test is failing, i thought i did that correctly this time.

jklukas:
I'm investigating the test issue. Indeed, it looks like you did the right thing.

PR #435, opened 2019-10-18
Bug 1576712: Add experiments and total_uri_count to search_clients_daily

Ben-Wu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1586025


templates/search/search_clients_daily_v8/query.sql
```diff
@@ -29,6 +29,19 @@ CREATE TEMP FUNCTION get_search_addon_version(active_addons ANY type) AS (
   )
 );
 
+-- Take array of experiment structs and return union of experiments.key_value arrays
+-- This is a workaround for the inability to use ARRAY_CONCAT_AGG as an analytical function
+CREATE TEMP FUNCTION union_experiments(list ANY TYPE) AS (
```
Ben-Wu: This is a kinda ugly function to get around not being able to use `ARRAY_CONCAT_AGG` over a window.  It does a union of the experiments by concatenating all the array and then deduping it.  Is there an easier way to do this that I'm missing?

templates/search/search_clients_daily_v8/query.sql
```diff
@@ -29,6 +29,19 @@ CREATE TEMP FUNCTION get_search_addon_version(active_addons ANY type) AS (
...
+  (
+    SELECT udf_dedupe_array(ARRAY_CONCAT_AGG(key_value))
+    FROM (
+      SELECT 1 AS val, key_value
+      FROM UNNEST(list)
+    )
+    GROUP BY val
+  )
+);
```
relud: this can be more efficiently written as:
```sql
CREATE TEMP FUNCTION union_experiments(list ANY TYPE) AS (
  (
    SELECT
      udf_dedupe_array(ARRAY_CONCAT_AGG(key_value))
    FROM
      UNNEST(list)
  )
);
```

templates/search/search_clients_daily_v8/query.sql
relud: I would recommend this:
```sql
CREATE TEMP FUNCTION udf_map_mode_last(list_of_maps ANY TYPE) AS (
  ARRAY(
    SELECT AS STRUCT
      key,
      udf_mode_last(ARRAY_AGG(value)) AS value
    FROM
      UNNEST(list_of_maps),
      UNNEST(key_value)
    GROUP BY
      key
  )
);
...
    udf_map_mode_last(ARRAY_AGG(experiments) OVER w1) AS experiments,
```

templates/search/search_clients_daily_v8/query.sql
relud: there is an implicit `GROUP BY TRUE` here

templates/search/search_clients_daily_v8/query.sql
Ben-Wu: That makes more sense

PR #436, opened 2019-10-21
Handle null oauth client ID in fxa auth event query

jklukas:
We found a dip in event counts because we were inadvertently throwing away sync events where the oauth client ID is null, but we want to include those events.


jklukas:
We also refactor a bit.

jklukas:
Backfilling the last few days now.

PR #441, opened 2019-10-23
Bug 1399532 -  Views for Windows 10 distribution dashboard

wlach:


sql/telemetry/windows_10_aggregate_v1/view.sql
```diff
@@ -0,0 +1,68 @@
...
+              IF
+                ((windows_build_number <= 17763),
+                  '1809',
+                IF
+                  ((windows_build_number <= 18362),
+                    '1903',
+                  IF
+                    ((windows_build_number > 18362),
+                      'Insider',
+                      NULL))))))))) build_group,
```
relud: this should use a case statement instead of nested if statements

sql/telemetry/windows_10_build_distribution/view.sql
```diff
@@ -0,0 +1,37 @@
...
+    build_group,
+    max(build_number) AS oo,
+    sum(count) AS frequency
+  FROM
+    `moz-fx-data-shared-prod.telemetry.windows_10_aggregate`
+  GROUP BY
+    normalized_channel,
+    build_group
+  ORDER BY
+    build_group ASC),
```
relud: I don't think this clause does anything. if so it should be removed

PR #443, opened 2019-10-23
Support converting key-value arrays to maps during parquet export

relud:


relud:
this worked as expected for exporting a 1% sample of main summary.

Issue #421, opened 2019-10-10
Add view for desktop events


jklukas:
@sunahsuh has mentioned that we want a view for desktop events, pulling in both main ping and event pings.

As an aside, I wanted to see what it would look like to rename fields inside the `main` view itself so that event fields have nice deanonymized names. It's quite verbose and ends up needing a lot of subqueries:

```
SELECT
  * REPLACE( (
    SELECT
      AS STRUCT payload.* REPLACE( (
        SELECT
          AS STRUCT payload.processes.* REPLACE( (
            SELECT
              AS STRUCT payload.processes.parent.* REPLACE( 
              ARRAY(SELECT
                  `moz-fx-data-shared-prod.udf.deanonymize_event`(e)
                FROM
                  UNNEST(payload.processes.parent.events) e) AS events)) AS parent)) AS processes)) AS payload)
FROM
  `moz-fx-data-shared-prod.telemetry_stable.main_v4`
WHERE
  DATE(submission_timestamp) = '2019-10-05'
  AND ARRAY_LENGTH(payload.processes.parent.events) > 0
LIMIT
  10
```

But at least it's possible to express in a view, which is fun! Maybe it's still worth codifying that in `main`.

Anyway, a really nice event view should add a parsed timestamp field that adds in the subsession start time and transforms the value into a BQ timestamp type. We can record more thoughts in this issue.

@sunahsuh - Is this something you were planning to pursue, or is it up for grabs? Do we know what name this view should have?

sunahsuh:
I'm working on a view that's schema-compatible with the existing `events` dataset (and considering whether it should be materialized, which I'm inclined to say yes since clustering by `event_category` feels like enough of a benefit to incur the storage costs)

Re: adding event timestamps, I'm not sure we do much that takes advantage of absolute-ish time right now, but it's certainly easy enough to add in a view :) I'm not planning on it in this current iteration at the moment, though

sunahsuh:
Just ran a quick test aggregating a couple days of enrollment/unenrollment events on a particular experiment on the view vs the existing parquet events table:
View:
![Screen Shot 2019-10-10 at 4 43 59 PM](https://user-images.githubusercontent.com/953153/66609017-92551680-eb7d-11e9-801c-dcfde458d773.png)

Events table:
[...]

PR #445, opened 2019-10-24
Bug 1570069: Add bigquery-native, backwards compatible events table

sunahsuh:
This creates a bigquery version of the existing events table. After running this for a few days we'll verify the outputs match up with the parquet version, then modify the user-facing view to be a union of this table going forward and the parquet table for historical data.

See also #421, which has a brief discussion on why this is a materialized table vs a view over the ping tables.


sunahsuh:
Filed #446 for the CI issue, which looks unrelated

templates/telemetry_derived/bq_events/query.sql
```diff
@@ -0,0 +1,139 @@
...
+    UNNEST(payload.events.gpu) AS e
+  UNION ALL
+  SELECT
+    *,
+    e,
+    "parent" AS event_process
+  FROM
+    telemetry.event
+  CROSS JOIN
+    UNNEST(payload.events.parent) AS e
```
relud: I think it would be better to `UNNEST` like this:
```sql
SELECT
  event.*,
  e,
  event_process
FROM
  telemetry.event AS event
CROSS JOIN
  UNNEST([
    STRUCT("content" AS event_process, payload.events.content as events),
    ("dynamic", payload.events.dynamic),
    ("extension", payload.events.extension),
    ("gpu", payload.events.gpu),
    ("parent", payload.events.parent)
  ])
CROSS JOIN
  UNNEST(events) AS e
```

templates/telemetry_derived/bq_events/query.sql
```diff
@@ -0,0 +1,139 @@
...
+    payload.info.session_start_date
+  ) AS session_start_time,
+  payload.info.subsession_id AS subsession_id,
+  submission_timestamp AS timestamp,
+  udf.deanonymize_event(e).*,
+  event_process "main" AS doc_type
+FROM
+  main_events
+WHERE
+  submission_date = @submission_date
```
relud: I would find the query easier to read if the field selection and where clauses moved up to the CTEs (after converting the CTEs to not use `UNION`), so that this end section is reduced to:
```sql
SELECT
  *
FROM
  event_events
UNION ALL
SELECT
  *
FROM
  main_events
```

templates/telemetry_derived/bq_events/query.sql
sunahsuh: Oh, lovely – didn't even realize this was possible :D

PR #450, opened 2019-10-25
Bug 1570069: Break out events query into separate tables per doctype

sunahsuh:
Per discussion in slack, I'm breaking these out into separate tables so they can be downstream of their respective copy_dedupe jobs


templates/telemetry_derived/bq_main_events/query.sql
```diff
@@ -0,0 +1,45 @@
+-- Creates a backwards-compatible flattened events dataset created from gcp-ingested main pings
```
relud: these tables need version suffixes, e.g. `templates/telemetry_derived/bq_main_events_v1/query.sql`

templates/telemetry_derived/bq_main_events/init.sql
```diff
@@ -0,0 +1,33 @@
+CREATE TABLE IF NOT EXISTS
+  `moz-fx-data-shared-prod.telemetry_derived.bq_main_events` (
```
relud: these tables need version suffixes
```suggestion
  `moz-fx-data-shared-prod.telemetry_derived.bq_main_events_v1` (
```

relud: also, i feel like we shouldn't be using a `bq_` prefix here

templates/telemetry_derived/bq_event_events/init.sql
```diff
@@ -0,0 +1,33 @@
+CREATE TABLE IF NOT EXISTS
+  `moz-fx-data-shared-prod.telemetry_derived.bq_event_events` (
```
relud: `event_events` oh how i wish there were a better name for this, but alas I can't think of one v_v

sql/telemetry_derived/bq_event_events/init.sql
```diff
@@ -1,5 +1,5 @@
 CREATE TABLE IF NOT EXISTS
-  `moz-fx-data-shared-prod.telemetry_derived.bq_events` (
+  `moz-fx-data-shared-prod.telemetry_derived.bq_event_events` (
```
jklukas: This table name should have a version suffix.

jklukas: And why the bq prefix here?

sql/telemetry_derived/bq_main_events/init.sql
```diff
@@ -1,5 +1,5 @@
 CREATE TABLE IF NOT EXISTS
-  `moz-fx-data-shared-prod.telemetry_derived.bq_events` (
+  `moz-fx-data-shared-prod.telemetry_derived.bq_main_events` (
```
jklukas: I had intended that tables in telemetry_derived are for backend use only and should always be versioned. Then the views we present on top of them in `telemetry` can be unversioned. That gives us more wiggle room to completely replace what the backend tables look like while keeping the user-facing view clean and unburdened with knowing what version of the backend table it's hitting.

sql/telemetry_derived/bq_event_events/init.sql
jklukas: I should have read through @relud's review first. I'm just making extra noise here.

templates/telemetry_derived/bq_event_events/init.sql
sunahsuh: Well, the good news is that users will never see the table name

sql/telemetry_derived/bq_event_events/init.sql
sunahsuh: Re: the bq prefix, we have an existing events table from parquet. I'm open to suggestions but the intention is to differentiate between the old table and the new one

relud: the existing table is just called `events` though, so this won't collide with that anymore.

PR #451, opened 2019-10-25
Search clients last seen

fbertsch:
Ben Miroglio to review for LTV-related aggregations. Is the output what we need to properly break revenue down by-month?

@relud I've changed the testing in a few ways:
- Added coercion to/from BYTES to allow serialization of those fields (as hex strings)
- Nicer pretty printing for test errors


bigquery_etl/util.py
```diff
@@ -100,18 +113,50 @@ def load_tables(
...
+
+            job_config.schema = updated_schema
+            temp_table_name = "_" + table.name
+            destination = dataset.table(temp_table_name)
+            as_bytes = ",".join(
+                [
+                    f"FROM_HEX({f.name}) AS {f.name}"
+                    for f in table.schema
+                    if f.field_type == "BYTES"
+                ]
```
relud: If you supply hex encoded bytes in ndjson format for a bytes field bq load will already decode them from hex.

also, this would break if `f.name` is nested within an array, and probably also if it's nested within a struct.

fbertsch: This could be useful, which hex representation? My only concern is that the input/output formats might be different in that case. Whatever we do we should ensure they are the same.

relud: er, sorry, not hex, base64

templates/search/search_clients_last_seen_v1/query.sql
```diff
@@ -0,0 +1,163 @@
...
+  _derived_search_cols AS (
+    SELECT
+      CASE
+          WHEN engine IS NULL THEN 'none'
+          WHEN STARTS_WITH(engine, 'google') THEN 'google'
+          WHEN STARTS_WITH(engine, 'ddg') OR STARTS_WITH(engine, 'duckduckgo') THEN 'ddg'
+          WHEN STARTS_WITH(engine, 'bing') THEN 'bing'
+          WHEN STARTS_WITH(engine, 'yandex') THEN 'yandex'
+          ELSE 'other'
+      END AS short_engine,
```
fbertsch: Is this the proper breakdown for engine? So for e.g. google, do we want to coalesce all google searches into the same bucket, or should some google engines be treated differently than others? (if we know a certain google engine doesn't provide revenue, for example, we should perhaps exclude it here.)

templates/search/search_clients_last_seen_v1/query.sql
```diff
@@ -0,0 +1,163 @@
...
+    WHERE
+      submission_date_s3 = @submission_date
+  ),
+
+  _derived_engine_searches AS (
+    SELECT
+      STRUCT(
+        short_engine AS key,
+        STRUCT(
+          total_searches,
```
fbertsch: For each `short_engine` (so bing, google, ddg, yandex) we will get a monthly breakdown of total_searches, tagged_searches, ad_clicks, and searches_with_ads. We can consider adding other breakdowns here if they could perhaps correlate more closely with revenue.

bigquery_etl/util.py
relud: and I believe it uses a base64 character set that matches python.

templates/search/search_clients_last_seen_v1/init.sql
```diff
@@ -0,0 +1,70 @@
...
+  search_clients_last_seen_v1
+PARTITION BY
+  submission_date
+CLUSTER BY
+  sample_id,
+  client_id
+OPTIONS
+  ( require_partition_filter=TRUE) AS
+SELECT
+    CAST(NULL AS DATE) AS submission_date,
```
fbertsch: Explicitly defined the column ordering here (rather than `SELECT *`) to match what we do in `_grouped` in `query.sql`.

bigquery_etl/util.py
```diff
@@ -142,13 +190,33 @@ def load_views(bq: bigquery.Client, dataset: bigquery.Dataset, views: Dict[str,
...
+                f"Line {i+1} column {e.colno} of file {filepath}, {e.msg}"
+            )
+
+    return res
+
+
+def json_load(file_obj: Iterable[str], filepath: str) -> List[Any]:
+    """Decode json from file_obj."""
+    try:
+        return json.loads("".join(file_obj))
```
relud: this should use the json method for loading directly from file objects, and fix the parameter type of file_obj to be `io.TextIOWrapper`
```suggestion
        return json.load(file_obj)
```

bigquery_etl/util.py
```diff
@@ -142,13 +190,33 @@ def load_views(bq: bigquery.Client, dataset: bigquery.Dataset, views: Dict[str,
...
+            res.append(json.loads(line))
+        except json.JSONDecodeError as e:
+            raise NDJsonDecodeError(
+                f"Line {i+1} column {e.colno} of file {filepath}, {e.msg}"
+            )
+
+    return res
+
+
+def json_load(file_obj: Iterable[str], filepath: str) -> List[Any]:
```
relud:
```suggestion
def json_load(file_obj: TextIOWrapper, filepath: str) -> List[Any]:
```

bigquery_etl/util.py
fbertsch: That could work, and we could use `TO_BASE64` on the other end. I'll take a look

fbertsch: Hmm, on second though a base-64 encoded byte string would be very confusing to look at, especially in tests...

bigquery_etl/util.py
```diff
@@ -100,18 +113,50 @@ def load_tables(
...
+        has_bytes_fields = False
+
         if table.schema is None:
             # autodetect schema if not provided
             job_config.autodetect = True
         else:
             job_config.schema = table.schema
             # look for time_partitioning_field in provided schema
-            for field in job_config.schema:
+            for i, field in enumerate(job_config.schema):
```
relud: this change doesn't appear to be used

bigquery_etl/util.py
relud: > That could work, and we could use `TO_BASE64` on the other end. I'll take a look

why would we need to use `TO_BASE64` on the other end?

relud: > Hmm, on second though a base-64 encoded byte string would be very confusing to look at, especially in tests...

base64 is also what the bigquery console and bq tool use to represent bytes, and what is used throughout ingestion-beam tests. I think we should stick with base64

sql/search/search_clients_last_seen_v1/init.sql
```diff
@@ -0,0 +1,99 @@
...
+    ] AS engine_searches
+
+    -- Each of the below is one year of activity, as BYTES.
+    udf_zeroed_365_days_bytes() AS days_seen_bytes,
+    udf_zeroed_365_days_bytes() AS days_searched_bytes,
+    udf_zeroed_365_days_bytes() AS days_tagged_searched_bytes,
+    udf_zeroed_365_days_bytes() AS days_searched_with_ads_bytes,
+    udf_zeroed_365_days_bytes() AS days_clicked_ads_bytes,
+    udf_zeroed_365_days_bytes() AS days_created_profile_bytes
+ FROM
```
relud:
```suggestion
FROM
```

udf/zeroed_364_days_active_1_bytes.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TEMP FUNCTION
+  udf_zeroed_364_days_active_1_bytes() AS (
```
fbertsch: I am not at all married to any of these function names, but this one in particular is especially awkward...

templates/search/search_clients_last_seen_v1/query.sql
```diff
@@ -0,0 +1,164 @@
...
+      udf_mode_last(ARRAY_AGG(profile_creation_date)) AS profile_creation_date,
+      udf_mode_last(ARRAY_AGG(default_search_engine)) AS default_search_engine,
+      udf_mode_last(ARRAY_AGG(default_search_engine_data_load_path)) AS default_search_engine_data_load_path,
+      udf_mode_last(ARRAY_AGG(default_search_engine_data_submission_url)) AS default_search_engine_data_submission_url,
+      udf_mode_last(ARRAY_AGG(profile_age_in_days)) AS profile_age_in_days,
+      udf_mode_last(ARRAY_AGG(active_addons_count_mean)) AS active_addons_count_mean,
+      udf_mode_last(ARRAY_AGG(user_pref_browser_search_region)) AS user_pref_browser_search_region,
+      udf_mode_last(ARRAY_AGG(os_version)) AS os_version,
+
+      -- User activity data
```
fbertsch: Ben, are these usage aggregate appropriate for `search_clients_daily`? The `SUM` fields are especially concerning because we don't want to double count anything. What this assumes is that now rows in `search_clients_daily` share any common e.g. `tab_open_event_count_sum` values, which may not be true if a single session used multiple engines.

templates/search/search_clients_last_seen_v1/query.sql
benmiroglio: I'd opt for using the standard `engine` field without bucketing for the top-level dataset; we can always bucket accordingly for specific analyses. For example, we have a few different google codes, and this in theory can change, as could our list of partners. 

I see this dataset being useful outside of the scope of LTV as well.

templates/search/search_clients_last_seen_v1/query.sql
benmiroglio: hm, these fields have particularly weird behavior, as I was going to suggest using `FIRST()` here instead of `SUM()`, however in `search_clients_daily`, for a client-day, these values are not always the same (for the same engine):

https://console.cloud.google.com/bigquery?pli=1&project=moz-fx-data-derived-datasets&folder&organizationId=442341870013&j=bq:US:bquxjob_6da07000_16e04a3130a&page=queryresults

templates/search/search_clients_last_seen_v1/query.sql
benmiroglio: Alternatively, we could handle STARTS_WITH(engine, 'other-') as 'other' and use the standard engine name otherwise, which would reduce the cardinality. These engines never generate revenue as well, so we could opt to simply exclude `other-*` engines and keep the rest.

templates/search/search_clients_last_seen_v1/query.sql
fbertsch: Looking over the code for `search_clients_daily` I don't think there's anyway we can get guaranteed correct values for these per-client. The problem is that some of them may have occurred in the same session, but some may not have, and we don't know which are which.

benmiroglio: AFAIK, this only affects these non search-related fields:

```sql
SUM(tab_open_event_count_sum) AS tab_open_event_count_sum,
SUM(active_hours_sum) AS active_hours_sum,
SUM(subsession_hours_sum) AS subsession_hours_sum,
SUM(sessions_started_on_this_day) AS sessions_started_on_this_day
```

Couldn't we treat these aggregations separately, which would require a join on `client_id`? This is something we'd want to do for `search_clients_daily` as well I'm thinking. It would duplicate the fields for clients with multiple `engine` or `source` fields (i.e. a client with 4 active hours will report 4 active hours for every row in scd), and we'd have to aggregate using `FIRST()` for downstream analysis.  Its not pretty, but I'm not sure if the current implementation has a useful interpretation.

fbertsch: Right, we're on the same page here. Honestly my thought was to simply elide them, since by themselves they aren't useful, and aggregating them together isn't viable.

templates/search/search_clients_last_seen_v1/query.sql
fbertsch: I'm not totally sure this dataset will be useful outside the scope of LTV:
- The search counts are the total for each month over the past year, and this is actual month (so e.g. sept, oct) which means the values are covering different #s of days
- We are currently planning on running on a sample, so these will be inaccurate to some degree

bigquery_etl/util.py
```diff
@@ -142,13 +190,33 @@ def load_views(bq: bigquery.Client, dataset: bigquery.Dataset, views: Dict[str,
...
+
+
+def json_load(file_obj: Iterable[str], filepath: str) -> List[Any]:
+    """Decode json from file_obj."""
+    try:
+        return json.loads("".join(file_obj))
+    except json.JSONDecodeError as e:
+        raise JsonDecodeError(
+            f"Line {e.lineno} column {e.colno} of file {filepath}, {e.msg}"
+        )
```
relud: These should use `file_obj.name`, rather than passing it in redundantly, like so:
```python
from io import TextIOWrapper


def read(*paths: str, decoder: Optional[Callable] = None, **kwargs):
    """Read a file and apply decoder if provided."""
    with open(os.path.join(*paths), **kwargs) as f:
        return decoder(f) if decoder else f.read()


def ndjson_load(file_obj: TextIOWrapper) -> List[Any]:
    """Decode newline delimited json from file_obj."""
    try:
        return [json.loads(line) for line in file_obj]
    except json.JSONDecodeError as e:
        raise NDJsonDecodeError(
            f"Line {i+1} column {e.colno} of file {file_obj.name}, {e.msg}"
        )


def json_load(file_obj: TextIOWrapper) -> List[Any]:
    """Decode json from file_obj."""
    try:
        return json.load(file_obj)
    except json.JSONDecodeError as e:
        raise JsonDecodeError(
            f"Line {e.lineno} column {e.colno} of file {file_obj.name}, {e.msg}"
        )
```

bigquery_etl/util.py
```diff
@@ -224,3 +292,60 @@ def coerce_result(*elements: Any) -> Generator[Any, None, None]:
             yield str(element)
         else:
             yield element
+
+
+def get_differences(exp, res, path="", sep=" / "):
```
relud: I would find this significanlty easier to read if it used the full names for the variables `expected` and `result`
```suggestion
def get_differences(expected, result, path="", sep=" / "):
```

relud: why was `" / "` chosen as the path separator here, instead of `"."`? this is for testing BigQuery SQL, so field names are guaranteed to not contain `.`, and `.` is an existing standard path separator, both for BigQuery and `jq`

templates/search/search_clients_last_seen_v1/query.sql
```diff
@@ -0,0 +1,164 @@
...
+      SUM(total_searches) AS total_searches,
+      SUM(tagged_searches) AS tagged_searches,
+
+      -- Monthly search totals
+      udf_aggregate_search_map(ARRAY_AGG(engine_searches)) AS engine_searches
+
+    FROM _derived_engine_searches
+    GROUP BY
+      sample_id,
+      client_id
```
relud: this data derives from `search_clients_daily`, which does not contain multiple values per `client_id` on a given `submission_date`, so this group by expression will have no effect other than casting `sample_id` to `INT64`

udf/bool_to_bytes.sql
```diff
@@ -0,0 +1,12 @@
+/* Convert a boolean to byte array */
+
+CREATE TEMP FUNCTION
+  udf_bool_to_bytes(val BOOLEAN) AS (
+    IF(COALESCE(val, FALSE), udf_zeroed_364_days_active_1_bytes(), udf_zeroed_365_days_bytes()));
```
relud: the `COALESCE` here has no effect, because `IF` treats `FALSE` and `NULL` the same:
```suggestion
    IF(val, udf_zeroed_364_days_active_1_bytes(), udf_zeroed_365_days_bytes()));
```

udf/array_sum.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TEMP FUNCTION
+  udf_array_sum(a ARRAY<INT64>) AS ((
```
relud: this UDF is not used anywhere in this PR

udf/add_monthly_searches.sql
```diff
@@ -0,0 +1,129 @@
+/*
+*/
+
+CREATE TEMP FUNCTION
+  udf_add_monthly_searches(prev ARRAY<STRUCT<key STRING, value STRUCT<total_searches ARRAY<INT64>, tagged_searches ARRAY<INT64>, search_with_ads ARRAY<INT64>, ad_click ARRAY<INT64>>>>,
+                           curr ARRAY<STRUCT<key STRING, value STRUCT<total_searches ARRAY<INT64>, tagged_searches ARRAY<INT64>, search_with_ads ARRAY<INT64>, ad_click ARRAY<INT64>>>>,
+                           submission_date DATE) AS ((
+  WITH prev_tbl AS (
```
relud: huh, I didn't know CTEs were allowed in subquery expressions. that's awesome!

udf/add_monthly_searches.sql
```diff
@@ -0,0 +1,129 @@
...
+    SELECT *
+    FROM UNNEST(prev) AS p
```
relud: the ` AS p` here is unused:
```suggestion
    FROM UNNEST(prev)
```

templates/search/search_clients_last_seen_v1/init.sql
```diff
@@ -0,0 +1,70 @@
...
+        udf_engine_searches_struct() AS value)
+    ] AS engine_searches
+
+    -- Each of the below is one year of activity, as BYTES.
+    udf_zeroed_365_days_bytes() AS days_seen_bytes,
+    udf_zeroed_365_days_bytes() AS days_searched_bytes,
+    udf_zeroed_365_days_bytes() AS days_tagged_searched_bytes,
+    udf_zeroed_365_days_bytes() AS days_searched_with_ads_bytes,
+    udf_zeroed_365_days_bytes() AS days_clicked_ads_bytes,
+    udf_zeroed_365_days_bytes() AS days_created_profile_bytes
```
relud: these should be defined using `CAST(NULL AS BYTES)` rather than `udf_zeroed_365_days_bytes()`

templates/search/search_clients_last_seen_v1/query.sql
```diff
@@ -0,0 +1,164 @@
...
+
+  _previous AS (
+  SELECT
+    * EXCEPT (submission_date)
+  FROM
+    search_clients_last_seen_v1
+  WHERE
+    submission_date = DATE_SUB(@submission_date, INTERVAL 1 DAY)
+    -- Filter out rows from yesterday that have now fallen outside the 365-day window.
+    AND udf_combine_adjacent_days_bytes(days_seen_bytes, udf_bool_to_bytes(FALSE)) != udf_zeroed_365_days_bytes())
```
relud: This would be more concise and easier to understand (at least for me) with `NULL`:
```suggestion
    AND udf_combine_adjacent_days_bytes(days_seen_bytes, NULL) != udf_zeroed_365_days_bytes())
```

relud: actually, I take that back, I think this should use `udf_shift_bytes_one_day`
```suggestion
    AND udf_shift_bytes_one_day(days_seen_bytes) != udf_zeroed_365_days_bytes())
```

templates/search/search_clients_last_seen_v1/query.sql
relud: nevermind, i'm wrong

udf/add_monthly_searches.sql
```diff
@@ -0,0 +1,129 @@
...
+                           udf_array_drop_first_and_append(udf_zeroed_array(12), 2) AS tagged_searches,
+                           udf_array_drop_first_and_append(udf_zeroed_array(12), 3) AS search_with_ads,
+                           udf_array_drop_first_and_append(udf_zeroed_array(12), 4) AS ad_click) AS value
+)] AS curr,
+        "bing" as type
+    UNION ALL
+    SELECT NULL as curr, "null" as type
+), dates AS (
+    SELECT d AS date
+    FROM UNNEST(ARRAY ["2019-10-01", "2019-10-02"]) AS d
```
relud: these should be date literals, rather than casting them to dates later:
```suggestion
    FROM UNNEST(ARRAY [DATE "2019-10-01", DATE "2019-10-02"]) AS d
```

udf/add_monthly_searches.sql
```diff
@@ -0,0 +1,129 @@
...
+    SELECT ARRAY[STRUCT(
+                    "bing" AS key,
+                    STRUCT(udf_array_drop_first_and_append(udf_zeroed_array(12), 1) AS total_searches,
+                           udf_array_drop_first_and_append(udf_zeroed_array(12), 2) AS tagged_searches,
+                           udf_array_drop_first_and_append(udf_zeroed_array(12), 3) AS search_with_ads,
+                           udf_array_drop_first_and_append(udf_zeroed_array(12), 4) AS ad_click) AS value
+)] AS curr,
+        "bing" as type
+    UNION ALL
+    SELECT NULL as curr, "null" as type
```
relud: this is incredibly difficult to read as formatted. please run it through `script/format-sql`, for output like:
```sql
WITH previous_examples AS (
  SELECT
    ARRAY [
      STRUCT(
        "google" AS key,
        STRUCT(
          generate_array(11, 0, -1) AS total_searches,
          generate_array(12, 1, -1) AS tagged_searches,
          udf_zeroed_array(12) AS search_with_ads,
          udf_zeroed_array(12) AS ad_click
        ) AS value
      )
    ] AS prev,
    "google" AS type
  UNION ALL
  SELECT
    NULL AS prev,
    "null" AS type
),
current_examples AS (
  SELECT
    ARRAY [
      STRUCT(
        "google" AS key,
        STRUCT(
          udf_array_drop_first_and_append(udf_zeroed_array(12), 5) AS total_searches,
          udf_array_drop_first_and_append(udf_zeroed_array(12), 10) AS tagged_searches,
          udf_array_drop_first_and_append(udf_zeroed_array(12), 15) AS search_with_ads,
          udf_array_drop_first_and_append(udf_zeroed_array(12), 20) AS ad_click
        ) AS value
      )
    ] AS curr,
    "google" AS type
  UNION ALL
  SELECT
    ARRAY[
      STRUCT(
        "bing" AS key,
        STRUCT(
          udf_array_drop_first_and_append(udf_zeroed_array(12), 1) AS total_searches,
          udf_array_drop_first_and_append(udf_zeroed_array(12), 2) AS tagged_searches,
          udf_array_drop_first_and_append(udf_zeroed_array(12), 3) AS search_with_ads,
          udf_array_drop_first_and_append(udf_zeroed_array(12), 4) AS ad_click
        ) AS value
      )
    ] AS curr,
    "bing" AS type
  UNION ALL
  SELECT
    NULL AS curr,
    "null" AS type
```

udf/add_monthly_searches.sql
```diff
@@ -0,0 +1,129 @@
...
+    FROM UNNEST(ARRAY ["2019-10-01", "2019-10-02"]) AS d
+), results AS (
+    SELECT udf_add_monthly_searches(p.prev, c.curr, CAST(d.date AS date)) AS res,
+      p.type AS p_type, c.type AS c_type, d.date AS date
+    FROM previous_examples p
+    CROSS JOIN current_examples c
+    CROSS JOIN dates d
+), expected AS (
+    SELECT *
+    FROM UNNEST(ARRAY [
```
relud: for consistency with existing code, please remove the leading `ARRAY` in front of `[` throughout this PR, unless it's being used to specify a schema for elements.

udf/add_monthly_engine_searches.sql
```diff
@@ -0,0 +1,63 @@
+/*
+*/
+
+CREATE TEMP FUNCTION
+  udf_add_monthly_engine_searches(prev STRUCT<total_searches ARRAY<INT64>, tagged_searches ARRAY<INT64>, search_with_ads ARRAY<INT64>, ad_click ARRAY<INT64>>,
+                           curr STRUCT<total_searches ARRAY<INT64>, tagged_searches ARRAY<INT64>, search_with_ads ARRAY<INT64>, ad_click ARRAY<INT64>>,
+                           submission_date DATE) AS (
+  IF(EXTRACT(DAY FROM submission_date) = 1,
```
relud: please add a comment explaining this behavior

udf/bool_to_bytes.sql
```diff
@@ -0,0 +1,12 @@
+/* Convert a boolean to byte array */
+
+CREATE TEMP FUNCTION
+  udf_bool_to_bytes(val BOOLEAN) AS (
```
relud: This isn't an appropriate name for this function, I believe it's misleading given that it's in a shared namespace, especially with the description `Convert a boolean to byte array`

udf/coalesce_adjacent_days_bytes.sql
```diff
@@ -0,0 +1,28 @@
...
+Given bytes representing usage from the previous day and the current day,
+this function shifts the first argument by one day and returns either that
+value if non-zero and non-null, the current day value if non-zero and non-null,
+or else 0.
+*/
+
+
+
+CREATE TEMP FUNCTION
+  udf_coalesce_adjacent_days_bytes(prev BYTES, curr BYTES) AS (
```
relud: these functions are all named with `bytes`, presumably to match the existing `bits` functions that store 28 days of history. I think don't think this naming is sufficiently descriptive to avoid confusing users who access our library of persistent udfs.

I propose that we rename the existing functions to be named `28bits` and these functions to be named `365bits`.

cc @jklukas

udf/int_to_hex_string.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TEMP FUNCTION
+  udf_int_to_hex_string(value INT64) AS ((
+    SELECT STRING_AGG(
+        SPLIT('0123456789ABCDEF','')[OFFSET((value >> (bits*4)) & 0xF)], ''
+        ORDER BY bits DESC
+    )
+    FROM UNNEST(generate_array(0, 91)) AS bits
```
relud: an INT64 only contains 16 nibbles, not 92, also I think `nibble` is a more accurate name than `bits` here:
```suggestion
    FROM UNNEST(generate_array(0, 15)) AS nibble
```

relud: i see that this is being used to test `udf_zeroed_364_days_active_1_bytes`, which is great, but this function shouldn't return extra bytes because it would cause unnecessary confusion for people accessing our library of UDFs.

udf/vector_add.sql
```diff
@@ -0,0 +1,26 @@
...
+  udf_vector_add(a ARRAY<INT64>, b ARRAY<INT64>) AS ((
+    with a_unnested AS (
+      SELECT _a, _a_off
+      FROM UNNEST(a) AS _a WITH OFFSET _a_off
+    ), b_unnested AS (
+      SELECT _b, _b_off
+      FROM UNNEST(b) AS _b WITH OFFSET _b_off
+    )
+
+    SELECT ARRAY_AGG(COALESCE(_a + _b, _a, _b) ORDER BY COALESCE(_a_off, _b_off) ASC)
```
relud: this should use an array subquery instead of `ARRAY_AGG`

udf/zeroed_364_days_active_1_bytes.sql
relud:
```suggestion
  udf_one_as_365bits() AS (
```

udf/zeroed_365_days_bytes.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TEMP FUNCTION
+  udf_zeroed_365_days_bytes() AS (
```
relud:
```suggestion
  udf_zero_as_365bits() AS (
```

udf/engine_searches_struct.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TEMP FUNCTION  udf_engine_searches_struct() AS (
+  STRUCT(
+    udf_zeroed_array(12) AS total_searches,
```
relud: please add a comment explaining why this 12

udf/engine_searches_struct.sql
```diff
@@ -0,0 +1,8 @@
...
+    udf_zeroed_array(12) AS tagged_searches,
```
relud: these should probably be replaced with a function like `udf_12_zeroes` that returns a constant and doesn't use a subquery expression, because a fully denormalized query can contain a limited number of subquery expressions, and because the query optimizer does nice things with constants.

udf/engine_searches_struct.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TEMP FUNCTION  udf_engine_searches_struct() AS (
```
relud:
```suggestion
CREATE TEMP FUNCTION  udf_engine_searches_empty_struct() AS (
```

udf/bytesmask_365_days.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TEMP FUNCTION
+  udf_bytesmask_365_days() AS (
```
relud:
```suggestion
  udf_bitmask_365_days() AS (
```

udf/array_drop_first_and_append.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TEMP FUNCTION
+  udf_array_drop_first_and_append(arr ANY TYPE, append ANY TYPE) AS (
+    ARRAY_CONCAT(
+      ARRAY(
+        SELECT v
+        FROM UNNEST(arr) AS v WITH OFFSET off
+        WHERE off > 0
+        ORDER BY off ASC),
```
relud: this order by clause shouldn't have any effect on the result

udf/bool_to_bytes.sql
relud:
```suggestion
  udf_bool_to_365bits(val BOOLEAN) AS (
```

udf/coalesce_adjacent_days_bytes.sql
relud:
```suggestion
  udf_coalesce_adjacent_days_365bits(prev BYTES, curr BYTES) AS (
```

udf/zeroed_array_with_last_val.sql
```diff
@@ -0,0 +1,3 @@
+CREATE TEMP FUNCTION
+  udf_zeroed_array_with_last_val(val INT64)  AS (
```
relud:
```suggestion
  udf_array_11_zeroes_then(val INT64) AS (
```

templates/search/search_clients_last_seen_v1/query.sql
fbertsch: Switching to `BIT_COUNT` to save on a UDF call.

udf/coalesce_adjacent_days_bytes.sql
jklukas: I agree that it makes sense to specify the number of bits in the function names, though I dislike the lack of underscore in `28bits`. We do already use `mau28` in the `exact_mau28` table names. Somehow, the number as suffix seems more palatable than as prefix.

So I'd prefer one of:

```
udf_coalesce_adjacent_days_bits365
udf_coalesce_adjacent_days_365_bits
```

fbertsch: My only other caveat is it's different types for the two. This is using `bytes`, while the existing one use `int`. However I doubt they'll get used at all outside our queries so perhaps it doesn't matter.

fbertsch: I also prefer `*_N_bits`, so let's go with that.

udf/array_drop_first_and_append.sql
fbertsch: It should retain the original ordering of the array, sans the first element: (see note [here](https://cloud.google.com/bigquery/docs/reference/standard-sql/arrays#filtering-arrays))

udf/bytesmask_365_days.sql
```diff
@@ -0,0 +1,15 @@
...
+    CONCAT(
+        b'\x1F\xFF\xFF\xFF',
```
jklukas: Looks like we could make this less verbose as `RPAD(b'\x1F', 46, b'\xFF')` or perhaps read legibly `CONCAT(b'\x1F', REPEAT(b'\xFF', 45))`.

fbertsch: Funny, we just reached the same conclusion independently!

jklukas: I created an issue to follow up on ways we can make more maintainable bit logic: https://github.com/mozilla/bigquery-etl/issues/578

udf/new_monthly_engine_searches_struct.sql
```diff
@@ -0,0 +1,13 @@
+/*
+This struct represents the past year's worth of searches.
+Each month has its own entry, hence 12.
+*/
+
+CREATE TEMP FUNCTION  udf_new_monthly_engine_searches_struct() AS (
```
jklukas:
```suggestion
CREATE TEMP FUNCTION udf_new_monthly_engine_searches_struct() AS (
```

templates/search_derived/search_clients_last_seen_v1/init.sql
```diff
@@ -0,0 +1,70 @@
...
+    sessions_started_on_this_day,
+
+    -- Search data
+    organic,
+    sap,
+    unknown,
+    tagged_sap,
+    tagged_follow_on,
+    ad_click,
+    search_with_ads,
```
jklukas: Do we generally want to include all fields from search_clients_daily? Should we express this with a `* EXCEPT(...)` rather than listing out all fields? I assume the exceptions are a much smaller list that would be less likely to change.

templates/search_derived/search_clients_last_seen_v1/query.sql
```diff
@@ -0,0 +1,164 @@
...
+      COALESCE(organic, 0) + COALESCE(sap, 0) + COALESCE(unknown, 0) + COALESCE(tagged_sap, 0) + COALESCE(tagged_follow_on, 0) AS total_searches,
+      COALESCE(tagged_sap, 0) + COALESCE(tagged_follow_on, 0) AS tagged_searches,
+      COALESCE(ad_click, 0) AS ad_click,
+      COALESCE(search_with_ads, 0) AS search_with_ads,
+      * EXCEPT (ad_click, search_with_ads)
+    FROM
+      search_clients_daily_v8
+    WHERE
+      submission_date = @submission_date
+      and sample_id = "84"
```
jklukas: sample_id should be an integer here. And can we add a comment about why the magic number 84?

templates/search_derived/search_clients_last_seen_v1/init.sql
jklukas: Reading the main query content, I now realize we have a grouping step where we have to list explicit aggregations for each column, so it makes more sense to me now why we'd list everything in init as well.

templates/search_derived/search_clients_last_seen_v1/query.sql
```diff
@@ -0,0 +1,164 @@
...
+      -- In this raw table, we capture the history of activity over the past
+      -- 365 days for each usage criterion as an array of bytes. The
+      -- rightmost bit represents whether the user was active in the current day.
+      udf_bool_to_365_bits(TRUE) AS days_seen_bytes,
+      udf_bool_to_365_bits(total_searches > 0) AS days_searched_bytes,
+      udf_bool_to_365_bits(tagged_searches > 0) AS days_tagged_searched_bytes,
+      udf_bool_to_365_bits(search_with_ads > 0) AS days_searched_with_ads_bytes,
+      udf_bool_to_365_bits(ad_click > 0) AS days_clicked_ads_bytes,
+      udf_int_to_365_bits(
+        udf_28_bits_from_days_since_created_profile(
```
jklukas: Are we converting this to 365 bits just for consistency with the other columns? Seems a little wacky, but I respect having more consistency in later calculations.

sql/search_derived/search_clients_last_seen_v1/init.sql
```diff
@@ -0,0 +1,81 @@
...
+  )
+);
+--
+CREATE TABLE
+  search_clients_last_seen_v1
+PARTITION BY
+  submission_date
+CLUSTER BY
+  sample_id,
+  client_id
```
jklukas: I don't see any benefit to clustering on client_id. I'd be inclined to keep this consistent with historical ping tables and cluster on (channel, sample_id)

fbertsch: It may improve join performance when joining this with `clients_daily` (_if_ the optimizer uses it).

templates/search_derived/search_clients_last_seen_v1/query.sql
fbertsch: Right, no reason to have another function that does the same as `udf_28_bits_from_days_since_created_profile` IMO.

PR #455, opened 2019-10-29
Add sql init files for Glam.

emtwo:


sql/telemetry/clients_daily_histogram_aggregates_v1/init.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE IF NOT EXISTS
+  `moz-fx-data-shared-prod.telemetry.clients_daily_histogram_aggregates_v1` (
```
fbertsch: These shouldn't have the `project.dataset` references, those are implied by the directory (`telemetry`).

sql/telemetry/clients_daily_histogram_aggregates_v1/init.sql
```diff
@@ -0,0 +1,18 @@
...
+    histogram_aggregates ARRAY <STRUCT<metric STRING,
+    metric_type STRING,
+    key STRING,
+    agg_type STRING,
+    bucket_range ARRAY <STRUCT<first_bucket INT64,
+    last_bucket INT64,
+    num_buckets INT64>>,
+    value ARRAY <STRUCT<key STRING,
+    value INT64>>>>)
+PARTITION BY submission_date
```
fbertsch: We should have the `require_partition_filter=TRUE` option

sql/telemetry/clients_daily_histogram_aggregates_v1/init.sql
emtwo: @jklukas just to confirm I understand what you're saying correctly...

I should be moving all the Glam tables that are now under `templates/telemetry/` and `sql/telemetry/` to be `templates/telemetry_derived` and `sql/telemetry_derived` respectively?

And that anything under`templates/telemetry/` and `sql/telemetry` should actually just be views?

jklukas: Exactly. Waiting for a separate PR sound good.

There are some existing tables defined under `telemetry` but the intent is to clean those up. We're also in an awkward transition phase in terms of moz-fx-data-derived-datasets vs. moz-fx-data-shared-prod.

shared-prod is the home for all data moving forward, so we should definitely define these glam tables there. But, some older tables still live in derived-datasets and derived-datasets is the default project for Airflow.

So, for now, it is indeed appropriate to specify the full name `moz-fx-data-shared-prod.telemetry.clients_daily_histogram_aggregates_v1` here when creating the table. When creating the tasks in Airflow, there's also some verbose boilerplate needed to make sure the tasks hit tables in shared-prod. Feel free to tag me when Airflow scheduling comes up and I can help make sure that process goes smoothly.

And apologies that we haven't been able to get to these cleanups sooner. I expected the transition period to be pretty short, but other tasks have taken priority, so it's not a great situation here in terms of being able to get a handle on the situation from docs.

jklukas:
I just pushed a commit to move these into telemetry_derived and add the require_partition_filter. I then ran the two statements, so these tables now exist.

PR #456, opened 2019-10-29
Bug 1587874 - Add default private search engine to search datasets

Ben-Wu:
@benmiroglio This adds `default_private_search_engine` and the load path and submission url values but not the origin and name values to match `default_search_engine`.  Is this structure fine?


benmiroglio:
SGTM.

Thanks for landing this!

sql/search/search_aggregates_v8/query.sql
```diff
@@ -9,20 +9,22 @@ SELECT
...
+  SUM(tagged_follow_on) AS tagged_follow_on,
+  SUM(sap) AS sap,
+  SUM(ad_click) AS ad_click,
+  SUM(search_with_ads) AS search_with_ads,
+  SUM(unknown) AS unknown
 FROM
   search_clients_daily_v8
 WHERE
   submission_date = @submission_date
+  AND engine IS NOT NULL
```
Ben-Wu: This line isn't really related to the PR but it should be here.

PR #459, opened 2019-10-30
Move Glam tables to telemetry_derived and add views in telemetry

emtwo:


sql/telemetry/clients_daily_histogram_aggregates_v1/view.sql
```diff
@@ -0,0 +1 @@
+SELECT * FROM telemetry_derived.clients_daily_histogram_aggregates_v1
```
jklukas: It's a BQ-level restriction that views must always use fully-qualified names. And we also require that `view.sql` files explicitly contain `CREATE OR REPLACE VIEW` statements, though I would like to eventually move to this style where the file contains only the body of the query.

```suggestion
CREATE OR REPLACE VIEW
  `moz-fx-data-shared-prod.telemetry.clients_daily_histogram_aggregates`
AS SELECT 
  * 
FROM 
  `moz-fx-data-shared-prod.telemetry_derived.clients_daily_histogram_aggregates`
```

jklukas: The automated schema pipeline handles publishing views and will also publish the views to the derived-datasets project so that Redash can see it.

emtwo: I've updated these, but the `dry-run-sql` script seems to be failing. I believe it's because in the views it's looking for the tables in `moz-fx-data-shared-prod` which don't exist and unfortunately I don't have permission to create them.

Would it make more sense to reference `moz-fx-data-derived-datasets` for now instead?

emtwo: Yes, @fbertsch has reviewed the tables in earlier PRs. I had filed an issue here to create the tables from the contents of this PR: https://github.com/mozilla/bigquery-etl/pull/455. Basically only 2 of the tables have partitions and clustering that needs to be specified (for now anyways) so they needed to be explicitly created with that.

The other tables I figured could be created in the airflow dag: https://github.com/mozilla/telemetry-airflow/pull/607 - note this needs a bit of an update after this PR..

Either way, I think it doesn't hurt to make all the tables in advance.

jklukas: I will try to make this happen right now.

jklukas:
For `clients_histogram_aggregates_v1` and `clients_scalar_aggregates_v1` I got:

> - query: Unrecognized name: app_build_id at [31:5]

Otherwise, these tables are now created (empty).

jklukas:
The probe_counts tables also failed to be created because of the two other tables not being created.

emtwo:
@jklukas My mistake in https://github.com/mozilla/bigquery-etl/pull/455/files I missed `app_build_id` in those files which caused the failures. I've added them now. So unfortunately we'll need to recreate these tables.

I took a look at the tables that were created and there are 3 `clients_daily_keyed_*` tables that we wont' actually need. The only tables needed are the ones referenced by the 5 views here:
Screen Shot 2019-11-01 at 2 22 04 PM

jklukas:
Making progress. I recreated the two tables with init.sql files.

Now, I get:

```
 - query: Cannot access field first_bucket on a value with type
ARRAY<STRUCT<first_bucket INT64, last_bucket INT64, num_buckets
INT64>> at [52:20]
```

from trying to run clients_histogram_aggregates_v1.

clients_scalar_aggregates_v1 is currently running but taking a while (3 minutes so far).

jklukas:
I've now removed the `clients_daily_keyed_*` tables.

jklukas:
Also, note that Airflow now has permissions to launch query jobs in shared-prod, which will greatly simplify the Airflow configuration for these jobs, but also means that we are good to use short table names here without always having to give the fully qualified name including project.

jklukas:
Fascinating. clients_scalar_aggregates_v1 ran for 41 minutes and scanned 1.3 TB, but returned no results. I imagine it returned no results because the daily aggregates table is empty.

Anyways, that table exists now and so does clients_scalar_probe_counts_v1.

The last two are still blocked by the `Cannot access field first_bucket` error.

script/dryrun
```diff
@@ -54,26 +54,32 @@ SKIP = {
...
-    "sql/telemetry/client_scalar_probe_counts_v1/query.sql",
     "sql/telemetry/firefox_kpi_dashboard_v1/query.sql",
     "sql/telemetry/core_clients_last_seen_raw_v1/query.sql",
+    "sql/telemetry/clients_daily_scalar_aggregates_v1/view.sql",
+    "sql/telemetry/clients_daily_histogram_aggregates_v1/view.sql",
+    "sql/telemetry/clients_scalar_aggregates_v1/view.sql",
+    "sql/telemetry/clients_histogram_aggregates_v1/view.sql",
+    "sql/telemetry/client_probe_counts_v1/view.sql",
+    "sql/telemetry_derived/clients_scalar_probe_counts_v1/query.sql",
+    "sql/telemetry_derived/clients_histogram_probe_counts_v1/query.sql",
```
jklukas: These can all be removed from this USE_DERIVED_DATASETS dict. We should let them hit the default shared-prod endpoint. I just checked and it looks like they all dry run successfully there.

PR #462, opened 2019-10-30
Add view for `telemetry.origin_content_blocking`

acmiyaguchi:
This view deduplicates aggregates created by the `prio_processor` DAG in Airflow. Since the DAG inserts rows directly into BigQuery, it is possible to have duplicate rows in a day if the job is rerun. Each row corresponds to the aggregate count of a blocked origin during a day. The most recent row by timestamp is returned in the view.


sql/telemetry/origin_content_blocking/view.sql
```diff
@@ -0,0 +1,28 @@
...
+    SELECT
+      *,
+      ROW_NUMBER() OVER (
+        PARTITION BY
+          submission_date,
+          batch_id,
+          origin
+        ORDER BY
+          TIMESTAMP DESC
+      ) AS insert_rank
```
relud: how many rows does this dataset have? this method is known to have performance issues at high volume. it sounds like it's not much, but i want to clarify.

sql/telemetry/origin_content_blocking/view.sql
```diff
@@ -0,0 +1,28 @@
...
+    FROM
+      `moz-fx-data-shared-prod.telemetry_derived.origin_content_blocking`
+  )
+WHERE
+  insert_rank = 1
```
relud: this is inconsistent with this pattern's use elsewhere in the repo. I think we should:
1. rename `insert_rank` as `_n`
2. make the select outer select clause `SELECT * EXCEPT(_n)`
3. move the inner query to a CTE

sql/telemetry/origin_content_blocking/view.sql
```diff
@@ -0,0 +1,28 @@
...
+ORDER BY
+  submission_date DESC,
+  batch_id,
+  aggregate DESC
```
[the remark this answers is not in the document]
acmiyaguchi: The upside of this is that `SELECT * from telemetry.origin_content_blocking` will return a useful view of the data at first glance.

sql/telemetry/origin_content_blocking/view.sql
acmiyaguchi: At the moment, it's 3m rows, or about 20mb scanned for this view.

PR #464, opened 2019-10-31
Bug 1592848 - Add a view for monitoring missing columns in the telemetry namespace

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1592848)

We can use this view to build queries like https://sql.telemetry.mozilla.org/queries/65690/source to figure out which paths are most impactful to add to mozilla-pipeline-schemas. This view uses the most recent date and a 1% sample due to the size of `additional_properties` in the ping table. This query is ~100GB.


acmiyaguchi:
This is affected by #465.

templates/monitoring/telemetry_missing_columns_v1/view.sql
```diff
@@ -0,0 +1,32 @@
...
+  REGEXP_EXTRACT(_TABLE_SUFFIX, r "^.*_v(.*)$") AS document_version,
+  path,
+  COUNT(*) AS path_count
+FROM
+  `moz-fx-data-shared-prod.telemetry_stable.*`,
+  UNNEST(
+    udf_js.json_extract_missing_cols(
+      additional_properties,
+      ARRAY ["histogram_type"],
+      ARRAY [
```
relud: i think it's inconsistent with the use of other arrays in this repo to prefix arrays with `ARRAY` when it's not being used to specify the schema of the array's elements.

fbertsch: I'm low-key in favor of prepending the type in all cases, it's pretty standard for SQL.

sql/monitoring/telemetry_missing_columns_v1/view.sql
```diff
@@ -0,0 +1,32 @@
+CREATE
+OR REPLACE VIEW `moz-fx-data-shared-prod.monitoring.telemetry_missing_columns_v1` AS
+SELECT
+  TIMESTAMP_TRUNC(submission_timestamp, DAY) AS day,
+  'telemetry' AS document_namespace,
+  REGEXP_EXTRACT(_TABLE_SUFFIX, r "^(.*)_v.*") AS document_type,
```
fbertsch: These could be useful UDFs: `udf_extract_doctype`, `udf_extract_docversion`

sql/monitoring/telemetry_missing_columns_v1/view.sql
```diff
@@ -0,0 +1,32 @@
...
+  COUNT(*) AS path_count
+FROM
+  `moz-fx-data-shared-prod.telemetry_stable.*`,
+  UNNEST(
+    udf_js.json_extract_missing_cols(
+      additional_properties,
+      ARRAY ["histogram_type"],
+      ARRAY [
+        "activeAddons",
+        "userPrefs",
```
fbertsch: Do these still showup in the `additional_properties`?

templates/monitoring/telemetry_missing_columns_v1/view.sql
relud: that's fine, but we should use one or the other, rather than slowly introduce inconsistency.

sql/monitoring/telemetry_missing_columns_v1/view.sql
acmiyaguchi: It is for documents that aren't the main ping -- we end up having missing section in a lot of pings because the common ping format isn't being applied consistently.

See the example redash query and look for userPrefs in the top results:

["heartbeat_v4","crash_v4","event_v4","third_party_modules_v4","modules_v4","new_profile_v4","shield_study_v4","testpilot_v4","untrusted_modules_v4","system_addon_deployment_diagnostics_v4","update_v4","shield_study_addon_v4"]

sql/monitoring/telemetry_missing_columns_v1/view.sql
acmiyaguchi: I've opened #471 for this. I don't think it's possible to deploy a udf and a view at the same time.

templates/monitoring/telemetry_missing_columns_v1/view.sql
acmiyaguchi: I'm going to use the version without the ARRAY prefix, it seems like the more natural way to write this.

PR #473, opened 2019-11-01
Add skip to dry_run

relud:


jklukas:
I will merge #460 tomorrow which will remove this file, so there should be no need for this change.

PR #481, opened 2019-11-05
Unnest clients_daily

relud:
and update `clients_histogram_aggregates` to handle unnested `udf_map_sum` function


sql/telemetry/clients_histogram_aggregates_v1/query.sql
```diff
@@ -126,8 +127,7 @@ normalized_histograms AS
       key,
       agg_type,
       latest_version,
-      udf_normalized_sum(
-        udf_aggregate_map_sum(ARRAY_AGG(STRUCT<key_value ARRAY<STRUCT <key STRING, value INT64>>>(value)))) AS aggregates
+      udf_normalized_sum(udf_map_sum(ARRAY_CONCAT_AGG(value))) AS aggregates
```
relud: r? @emtwo on modifying `clients_histogram_aggregates` to handle the updated `udf_map_sum` function. This change shouldn't impact the input or output of this query.

relud:
> we will need to do some juggling to prevent this from breaking `clients_last_seen`

@jklukas do you think it would be better to:
1. select each day from `telemetry.clients_last_seen_v1` into a temp table, update the `telemetry.clients_last_seen_v1` view to point to the temp table and not unnest, then copy the temp table over `telemetry_derived.clients_last_seen_v1`, then update the view and remove the temp table
2. copy `telemetry_derived.clients_last_seen_v1` to a different name, update the view, recreate `telemetry_derived.clients_last_seen_v1` empty and unnested, and update the view to union the two tables
3. update airflow to write unnested `clients_last_seen_v1` to a different location, and update the view to union the two tables.

relud:
(and the same with `clients_daily_v6`)

jklukas:
We should discuss in real life so I can get some more context.

relud:
decision: we will write the new unnested tables to the `moz-fx-data-shared-prod:telemetry_derived` dataset, and then update the views to union those with the old nested tables in `moz-fx-data-derived-datasets:telemetry_derived`.

then later we can copy the old data into the new tables, update the views to not reference the old tables, and delete the old tables.

Timing of the last part will need to be soon for `clients_last_seen` to fix `sample_id`, but isn't urgent for `main_summary` or `clients_daily`. Non-urgent changes might benefit from waiting until CCPA so that we will have reserved slots and no loss of long-term storage.

relud:
I have dropped unnesting `clients_last_seen` from this PR because it needs to wait until we are confident enough with sql `main_summary` and `clients_daily_v6` to stop running them in spark.

PR #486, opened 2019-11-06
Add convenience function to extract count histogram values

sunahsuh:
This will make some of the longitudinal migration queries a bit more readable


udf/extract_count_histogram_value.sql
```diff
@@ -0,0 +1,26 @@
+CREATE TEMP FUNCTION
```
jklukas: Can you add some docs here at the top of the file here? It's not obvious to me why we're extracting just the `0` key.

PR #491, opened 2019-11-06
Bug 1594108 - Derive error_aggregates from live crash pings

Ben-Wu:


Ben-Wu:
Tests are coming

sql/telemetry_derived/error_aggregates/query.sql
```diff
@@ -0,0 +1,185 @@
...
+    application.display_version,
+    environment.build.build_id,
+    metadata.uri.app_name,
+    IF(metadata.uri.app_name = 'Fennec'
+       AND environment.system.os.name = 'Linux',
+       'Android',
+       environment.system.os.name) AS os,
+    environment.system.os.version AS os_version,
+    environment.build.architecture,
+    normalized_country_code AS country,
```
wlach: We probably don't strictly some of these dimensions, though I guess it doesn't hurt to throw them in. It may help with validation.

sql/telemetry_derived/error_aggregates/query.sql
```diff
@@ -0,0 +1,185 @@
...
+    normalized_channel AS channel,
+    environment.build.version,
+    application.display_version,
+    environment.build.build_id,
+    metadata.uri.app_name,
+    environment.system.os.name AS os,
+    environment.system.os.version AS os_version,
+    environment.build.architecture,
+    normalized_country_code AS country,
+    -- 0 columns to match crash ping
```
wlach: Maybe we should something like, "we get these columns from the crash pings, so initialize them to zero"

sql/telemetry_derived/error_aggregates/query.sql
```diff
@@ -0,0 +1,185 @@
...
+    0 AS startup_crash,
+    0 AS content_shutdown_crash,
+    LEAST(GREATEST(durations / 3600, 0), 25) AS usage_hours,
+    0 AS gpu_crashes,
+    0 AS plugin_crashes,
+    0 AS gmplugin_crashes
+  FROM
+    `moz-fx-data-shared-prod.telemetry_live.core_v10`
+),
+combined_crashes AS (
```
wlach: Maybe this should be called something like `combined_aggregates` instead?

sql/telemetry_derived/error_aggregates/query.sql
```diff
@@ -0,0 +1,185 @@
...
+    environment.system.os.name AS os,
+    environment.system.os.version AS os_version,
+    environment.build.architecture,
+    normalized_country_code AS country,
+    -- 0 columns to match crash ping
+    0 AS main_crash,
+    0 AS content_crash,
+    0 AS startup_crash,
+    0 AS content_shutdown_crash,
+    LEAST(GREATEST(payload.info.subsession_length / 3600, 0), 25) AS usage_hours,
```
wlach: Could you add a comment here on how we're doing this to protect against clients sending absurdly high values for usage_hours? This bug has a reference: https://bugzilla.mozilla.org/show_bug.cgi?id=1514392

sql/telemetry_derived/error_aggregates/query.sql
```diff
@@ -0,0 +1,185 @@
...
+    SELECT
+      udf_json_extract_histogram(value).sum
+    FROM
+      UNNEST(histogram_list)
+    WHERE
+      key = target_key
+    LIMIT
+      1
+  )
+);
```
wlach: I wonder if we should have some kind of udf for getting the sum of a histogram? Seems like something that would be useful in many places.

sql/telemetry_derived/error_aggregates/query.sql
```diff
@@ -0,0 +1,185 @@
...
+  udf_json_extract_int_map (input STRING) AS (ARRAY(
+    SELECT
+      STRUCT(CAST(SPLIT(entry, ':')[OFFSET(0)] AS INT64) AS key,
+             CAST(SPLIT(entry, ':')[OFFSET(1)] AS INT64) AS value)
+    FROM
+      UNNEST(SPLIT(REPLACE(TRIM(input, '{}'), '"', ''), ',')) AS entry
+    WHERE
+      LENGTH(entry) > 0 ));
+CREATE TEMP FUNCTION
+  udf_json_extract_histogram (input STRING) AS (STRUCT(
```
wlach: Is this function not already available for use? https://docs.telemetry.mozilla.org/cookbooks/bigquery.html#accessing-histograms

jklukas: Because persistent UDFs are still technically in beta, we use temporary UDFs (which are GA) in these ETL queries. You can reference names like `udf_json_extract_histogram` in your queries in the `templates` folder and the  bigquery-etl machinery will inject the temporary UDF into the generated files under `sql`. So this looks fine to me.

I bet persistent UDFs will be GA soon and we'll be able to get rid of the generation machinery altogether.

udf/round_timestamp_to_minute.sql
```diff
@@ -0,0 +1,24 @@
+/*
+
+Floor a timestamp object to the given minute interval.
+
+ */
+CREATE TEMP FUNCTION udf_round_timestamp_to_minute(timestamp_expression TIMESTAMP, minute INT64) AS (
+  TIMESTAMP_SECONDS(
+    CAST((FLOOR(UNIX_SECONDS(timestamp_expression) / (minute * 60)) * minute * 60) AS INT64)
```
jklukas: Can you add a test case for `udf_round_timestamp_to_minute(TIMESTAMP '2019-05-03 00:34:59.999999', 5)` to validate whether `UNIX_SECONDS` is potentially rounding up?

udf/histogram_get_sum.sql
```diff
@@ -0,0 +1,18 @@
...
+    SELECT
+      udf_json_extract_histogram(value).sum
+    FROM
+      UNNEST(histogram_list)
+    WHERE
+      key = target_key
+    LIMIT
+      1
+  )
+);
```
akkomar: @relud should we add tests for such simple udfs?

templates/telemetry_derived/error_aggregates/query.sql
```diff
@@ -0,0 +1,147 @@
...
+  SUM(usage_hours) AS usage_hours
+FROM
+  combined_crashes
+WHERE
+  DATE(submission_timestamp) = '2019-11-05' -- TODO: USE param
+  AND DATE_DIFF(  -- Only use builds from the last 6 months
+    CURRENT_DATE(),
+    SAFE.PARSE_DATE('%Y%m%d', SUBSTR(build_id, 0, 8)),
+    MONTH
+  ) <= 6
```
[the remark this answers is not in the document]
wlach: New data for old build ids should still be appearing in the live ping tables, I would assume? You bring up a good point though -- I think for backfilling old data into this table (if that turns out to be required), we should use the stable ping tables. We should probably go through the possible scenarios required before landing this.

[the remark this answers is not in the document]
akkomar: > New data for old build ids should still be appearing in the live ping tables, I would assume?

Yes, that's correct.

> We should probably go through the possible scenarios we need to handle before landing this.

:+1:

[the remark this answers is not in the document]
Ben-Wu: It may be possible to use a parameter like `@source_table` that can be set in airflow so that the scheduled run uses live and backfills use stable.

udf/histogram_get_sum.sql
relud: IMO if it's complex enough to be a UDF, it's complex enough to need tests.

templates/telemetry_derived/error_aggregates/query.sql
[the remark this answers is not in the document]
relud: unfortunately you can't determine tables directly using a query parameter. **BUT** you can dynamically read from predefined tables depending on the partitions being read via unions:

```sql
WITH pings AS (
  SELECT
    *
  FROM
    `moz-fx-data-shared-prod.telemetry_live.core_v10`
  WHERE
    DATE(submission_timestamp) >= DATE_SUB(CURRENT_DATE, INTERVAL 7 DAY)
  UNION ALL
  SELECT
    *
  FROM
    `moz-fx-data-shared-prod.telemetry_stable.core_v10`
  WHERE
    DATE(submission_timestamp) < DATE_SUB(CURRENT_DATE, INTERVAL 7 DAY)
)
SELECT
  *
FROM
  pings
WHERE
```

templates/telemetry_derived/error_aggregates/query.sql
```diff
@@ -0,0 +1,147 @@
+WITH crash_ping_agg AS (
```
akkomar: Nit: I'm not sure these result sets should be suffixed with `_agg` since there are no aggregations there?

Ben-Wu:
Ready for re-review now.  Added tests in which I needed to fudge the dates and build ids because the query uses `CURRENT_DATE` and I couldn't find a way to mock that.
Also now using the union table method @relud suggested: query live tables for the last day and stable tables for anything older.

udf/histogram_get_sum.sql
```diff
@@ -0,0 +1,39 @@
...
+  (
+    SELECT
+      udf_json_extract_histogram(value).sum
+    FROM
+      UNNEST(histogram_list)
+    WHERE
+      key = target_key
+    LIMIT
+      1
+  )
```
relud: this seems a bit verbose compared to `udf_json_extract_histogram(udf_get_key(keyed_histogram, key)).sum`

udf/histogram_get_sum.sql
```diff
@@ -0,0 +1,39 @@
+/*
+
+Take a list of histograms of type STRUCT<key STRING, value STRING>,
+extract the histogram of the given key, and return the sum value
+
+ */
+CREATE TEMP FUNCTION udf_histogram_get_sum(histogram_list ANY TYPE, target_key STRING) AS (
```
relud: I think this should named to indicate that it takes a keyed histogram:
```suggestion
CREATE TEMP FUNCTION udf_keyed_histogram_get_sum(keyed_histogram ANY TYPE, target_key STRING) AS (
```

sql/telemetry_derived/error_aggregates/query.sql
```diff
@@ -0,0 +1,234 @@
...
+    telemetry.core_live
+  WHERE
+    DATE_DIFF(CURRENT_DATE, DATE(submission_timestamp), DAY) <= 1
+  UNION ALL
+  SELECT
+    *
+  FROM
+    telemetry.core
+  WHERE
+    DATE_DIFF(CURRENT_DATE, DATE(submission_timestamp), DAY) > 1
```
relud: I think these are more readable, easier to reason about, and more consistent with other queries in this repo when formatted with the partition field on the left and the (effectively) constant value on the right:
```suggestion
    DATE(submission_timestamp) < DATE_SUB(CURRENT_DATE, INTERVAL 1 DAY)
```

Ben-Wu: Sure. I find both to be equally hard to reason about

udf/histogram_get_sum.sql
Ben-Wu: Yes, I overlooked `udf_get_key`

sql/telemetry_derived/error_aggregates/query.sql
```diff
@@ -0,0 +1,233 @@
...
+  SUM(gpu_crashes) AS gpu_crashes,
+  SUM(plugin_crashes) AS plugin_crashes,
+  SUM(gmplugin_crashes) AS gmplugin_crashes,
+  SUM(usage_hours) AS usage_hours
+FROM
+  combined_ping_data
+WHERE
+  DATE(submission_timestamp) = @submission_date
+  AND DATE_DIFF(  -- Only use builds from the last 6 months
+    CURRENT_DATE,
```
akkomar: I would consider using `@submission_date` here, otherwise output depends on the run date which might result in some data weirdness after backfills.

sql/telemetry_derived/error_aggregates/query.sql
```diff
@@ -0,0 +1,233 @@
...
+  SUM(startup_crash) AS startup_crashes,
+  SUM(content_shutdown_crash) AS content_shutdown_crashes,
+  SUM(gpu_crashes) AS gpu_crashes,
+  SUM(plugin_crashes) AS plugin_crashes,
+  SUM(gmplugin_crashes) AS gmplugin_crashes,
+  SUM(usage_hours) AS usage_hours
+FROM
+  combined_ping_data
+WHERE
+  DATE(submission_timestamp) = @submission_date
```
akkomar: Is this going to be scheduled on Airflow?
If so, it would need to run more often than daily. If I remember correctly @wlach wanted to get updates every hour?

Ben-Wu: Yes we'll run it every few hours in Airflow.  I don't expect any issues with running it more than once a day.

sql/telemetry_derived/error_aggregates/query.sql
Ben-Wu: Yes that's a good point

sql/telemetry/core_live/view.sql
```diff
@@ -0,0 +1,27 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.core_live`
```
relud: I think this would be better suited to living in the `telemetry_live` dataset, and I'm also interested in @jklukas opinion on the matter:
```suggestion
  `moz-fx-data-shared-prod.telemetry_live.core`
```

jklukas: Views on top of a live table are an edge case that we haven't fully grappled with. I can see an argument for putting this in `telemetry_live`, but it's forbidden by the permissions (Airflow should not have access to create tables/views in the `_live` datasets) and logically I want to keep `_live` and `_stable` as being purely owned by the pipeline.

So I think `moz-fx-data-shared-prod.telemetry.core_live` as used here seems like a reasonable option.

jklukas: Although I also might argue we should put this in `telemetry_derived` to hide it as an implementation detail rather than making it user-facing.

relud: good point, i'm on board with putting this in `telemetry_derived`:
```suggestion
  `moz-fx-data-shared-prod.telemetry_derived.core_live`
```

Ben-Wu: I don't have any strong feelings on where this should be.  `telemetry.core_live` is nice for consistency with `telemetry.core` and the other views but I understand wanting to hide it.  @relud Final decision?

sql/telemetry_derived/error_aggregates/query.sql
relud: `udf_json_extract_histogram(histogram).sum` is the same length as `udf_json_extract_histogram_sum(histogram)`, and because it's a sql UDF it's also just as efficient (@jklukas did an analysis to confirm this).

templates/telemetry_derived/error_aggregates/query.sql
```diff
@@ -0,0 +1,196 @@
...
+
+SELECT
+  DATE(submission_timestamp) AS submission_date,
+  udf_round_timestamp_to_minute(submission_timestamp, 5) AS window_start,
+  channel,
+  version,
+  display_version,
+  build_id,
+  app_name as application,
+  os as os_name,
```
relud: please capitalize the `AS` keyword

templates/telemetry_derived/error_aggregates/query.sql
```diff
@@ -0,0 +1,196 @@
...
+    os,
+    osversion AS os_version,
+    arch AS architecture,
+    normalized_country_code AS country,
+    -- 0 for values retrieved from crash pings
+    0 AS main_crash,
+    0 AS content_crash,
+    0 AS startup_crash,
+    0 AS content_shutdown_crash,
+    LEAST(GREATEST(durations / 3600, 0), 25) AS usage_hours,  -- protect against extreme values
```
relud: should we be limiting to 24 hours of usage in a day/ping? cc @wlach 
```suggestion
    LEAST(GREATEST(durations / 3600, 0), 24) AS usage_hours,  -- protect against extreme values
```
and should we be making extreme values `NULL` instead? 
```suggestion
    IF(durations/3600 BETWEEN 0 AND 24, durations/3600, NULL) AS usage_hours,  -- protect against extreme values
```

relud: I think this should select `payload.info.subsession_length` and `durations` as `usage_seconds`, and then convert to hours and protect against extreme values in the final select, to reduce code duplication

Ben-Wu: Using 24 hours makes sense.  I just copied 25 from the spark job maybe there's a reason for that @wlach knows.  Setting extreme values to `NULL` will inflate the crash rates since they won't contribute to the total usage hours but will contribute crash counts so I'm not sure that's the best way to do it.  Of course setting artificial limits is also not ideal, we just need to accept that the data is somewhat messed up.

Ben-Wu: I just ran a query on main pings and it looks like there are a bunch of subsession lengths that are 24 hours + a bit so maybe that's why.

templates/telemetry_derived/error_aggregates/query.sql
```diff
@@ -0,0 +1,196 @@
...
+    normalized_channel AS channel,
+    environment.build.version,
+    application.display_version,
+    environment.build.build_id,
+    metadata.uri.app_name,
+    IF(metadata.uri.app_name = 'Fennec'
+       AND environment.system.os.name = 'Linux',
+       'Android',
+       environment.system.os.name) AS os,
+    environment.system.os.version AS os_version,
```
relud: I think we should be using `normalized_os` and `normalized_os_version` throughout this, unless we have a specific need here. `Fennec` + `Linux` will probably need to stay though, per https://github.com/mozilla/gcp-ingestion/issues/999

Ben-Wu: Sure.  The only "need" is that this requires changing mission control to use normalized values instead

relud: the difference should be that `os = 'Darwin'` becomes `normalized_os = 'Mac'`, and unrecognized values are normalized to `Other`.

cc @wlach regarding whether mission control should needs the original values

templates/telemetry_derived/error_aggregates/query.sql
relud: > Setting extreme values to `NULL` will inflate the crash rates since they won't contribute to the total usage hours

setting extreme values to `0` has the same effect on `SUM` as setting them to `NULL`, except that if *all* the values for a group are `NULL` them `SUM` returns `NULL`

templates/telemetry_derived/error_aggregates/query.sql
relud: > changing mission control to use normalized values instead

also, we could use `IF(normalized_os = 'Mac', 'Darwin', normalized_os)`.



> the difference should be that `os = 'Darwin'` becomes `normalized_os = 'Mac'`, and unrecognized values are normalized to `Other`.

er, I oversimplified. there's quite a few differences:
- os **contains** `Linux`, `BSD`, or `SunOS` becomes `Linux`
- os **starts with** `Windows` or `WINNT` becomes `Windows`
- os **starts with** `Darwin` becomes `Mac`
- os **starts with** `iOS` or **contains** `iPhone` becomes `iOS`
- os **starts with** `Android` becomes `Android`
- anything else becomes `Other`

and if https://github.com/mozilla/gcp-ingestion/issues/1000 is merged, it will be even more complex.

templates/telemetry_derived/error_aggregates/query.sql
Ben-Wu: I was more concerned about the high values than negative.  In a quick scan I didn't find any negative values but a lot of >24 hour values.  Setting those to `NULL` instead of 24 (or 25) will decrease the summed usage hours by a lot.

relud: ah, now I understand. You're right, we shouldn't use `NULL`.

sql/telemetry/core_live/view.sql
relud: > @relud Final decision?

use `telemetry_derived`, if people outside data eng are reading live pings then the implied contract is `ymmv, but we'll try to help`, and not `this is supported`

templates/telemetry_derived/error_aggregates/query.sql
Ben-Wu: That's a bit of a mess.  It should be easier to use normalized and make the changes in mission control.

wlach: I'm fine with whatever you want to do here, Ben is actually doing the mission control changes so can presumably adjust things on that side as needed.

templates/telemetry_derived/error_aggregates/query.sql
wlach: In general going forward we should use active_ticks for getting usage information, which isn't susceptible to these weird values. The model used in missioncontrol-v2 does this, among other things.

For this dataset however, our priority should be on not doing anything too surprising from the point of view of the users of the mission control dashboard (i.e. try to keep the rate numbers the same, more or less), unless we encounter some kind of egregious error. We know for a fact that this "error aggregates" approach has a *lot* of problems -- that's why we're moving away from it! :smiley:

PR #492, opened 2019-11-07
Support bigquery export to parquet via avro

relud:


relud:
I successfully used this to handle exporting sql clients_daily in 10 minutes.

PR #494, opened 2019-11-08
Add a dataset for compatibility with "client_count"

mreid-moz:
:(


templates/telemetry/firefox_desktop_exact_mau28_by_client_count_dimensions_v1/query.sql
```diff
@@ -0,0 +1,46 @@
...
+  locale,
+  normalized_channel,
+  os,
+  os_version,
+  distribution_id,
+  COALESCE(cc.name, cls.country) AS country_name
+FROM
+  clients_last_seen_v1 cls
+LEFT JOIN
+  static.country_codes_v1 cc
```
relud: table aliases should use `AS`, for consistency

PR #499, opened 2019-11-12
Add the Windows 10 November 2019 update

rvandermeulen:
Also fixes a logic error for Insider builds (should be looking at all higher build numbers).


wlach:
Thank you for catching the Insider error! Can't believe I missed that earlier.

PR #500, opened 2019-11-12
Switch search to read from flattened main summary

Ben-Wu:
This is so we can put search in shared prod and backfill from the derived datasets main summary.


Ben-Wu:
I repurposed the pr to read from the main summary view.  Originally I had just the backfill in mind and switch over to the new main summary after but in retrospect that doesn't really make sense to do.

templates/search_derived/search_clients_daily_v8/query.sql
```diff
@@ -52,25 +35,25 @@ WITH
       END
         AS type
       FROM
-        UNNEST(search_counts.list) ),
+        UNNEST(search_counts) element ),
```
relud: Aliases need to explicitly use `AS`
```suggestion
        UNNEST(search_counts) AS element),
```

PR #501, opened 2019-11-13
Promote sql clients_daily_v6

relud:
co-dependent on https://github.com/mozilla/telemetry-airflow/pull/725


templates/telemetry/clients_daily_v6/view.sql
```diff
@@ -12,19 +12,18 @@ SELECT
     ARRAY(SELECT * FROM UNNEST(environment_settings_intl_requested_locales.list)) AS environment_settings_intl_requested_locales,
     ARRAY(SELECT * FROM UNNEST(environment_settings_intl_system_locales.list)) AS environment_settings_intl_system_locales,
     experiments.key_value AS experiments,
+    SAFE.PARSE_DATE("%F", SUBSTR(profile_creation_date, 0, 10)) AS profile_creation_date,
```
relud: This is set up so that `profile_creation_date` will be changed from a `STRING` to a `DATE` in `clients_daily` and `clients_last_seen`, and the old data is converted in views.

I need to verify that this won't cause issues with any scheduled re:dash queries.

For parquet the string format will be restored on export for backwards compatibility.

templates/telemetry/clients_last_seen_v1/view.sql
```diff
@@ -15,6 +15,7 @@ SELECT
     ARRAY(SELECT * FROM UNNEST(environment_settings_intl_requested_locales.list)) AS environment_settings_intl_requested_locales,
     ARRAY(SELECT * FROM UNNEST(environment_settings_intl_system_locales.list)) AS environment_settings_intl_system_locales,
```
relud: For parquet the nested format will be restored on export for backwards compatibility.

templates/telemetry/clients_daily_v6/view.sql
jklukas: We could instead add a `parsed_profile_creation_date` field alongside the raw value.

sql/telemetry_derived/clients_last_seen_v1/init.sql
```diff
@@ -3,22 +3,21 @@ CREATE TABLE
 PARTITION BY
   submission_date
 CLUSTER BY
-  sample_id,
-  client_id AS
+  normalized_channel,
+  sample_id
```
jklukas: 👍

sql/telemetry_derived/clients_last_seen_v1/init.sql
```diff
@@ -3,22 +3,21 @@ CREATE TABLE
...
-  ( require_partition_filter=TRUE) AS
+  (require_partition_filter=TRUE) AS
 SELECT
-  CAST(NULL AS DATE) AS submission_date,
-  0 AS days_seen_bits,
-  0 AS days_visited_5_uri_bits,
-  0 AS days_opened_dev_tools_bits,
-  0 AS days_created_profile_bits,
-  [STRUCT("" AS experiment, "" AS branch, 0 AS bits)] AS days_seen_in_experiment,
+  NULL AS days_seen_bits,
```
jklukas: The zeroes here ensured we got the right type for the field. I believe INT64 is the default type used for NULLs, but I'd rather not rely on that implicit behavior. We can make these `CAST(NULL AS INT64)` to be explicit.

templates/telemetry/clients_last_seen_v1/view.sql
```diff
@@ -28,3 +29,28 @@ SELECT
...
   `moz-fx-data-derived-datasets.telemetry_derived.clients_last_seen_v1`
+WHERE
+  submission_date < DATE '2019-11-14'
+UNION ALL
+SELECT
+  -- We cannot use UDFs in a view, so we paste the body of udf_bitpos(bits) literally here.
+  CAST(SAFE.LOG(days_seen_bits & -days_seen_bits, 2) AS INT64) AS days_since_seen,
+  CAST(SAFE.LOG(days_visited_5_uri_bits & -days_visited_5_uri_bits, 2) AS INT64) AS days_since_visited_5_uri,
+  CAST(SAFE.LOG(days_opened_dev_tools_bits & -days_opened_dev_tools_bits, 2) AS INT64) AS days_since_opened_dev_tools,
+  CAST(SAFE.LOG(days_created_profile_bits & -days_created_profile_bits, 2) AS INT64) AS days_since_created_profile,
```
jklukas: I'd prefer that we do the union in a CTE and then add these days_since_ columns on top of the CTE to avoid duplicating the logic between the two sides of the union.

PR #502, opened 2019-11-13
Update ssl_ratios_v1 to read main instead of main_summary

relud:
~As per the comment being removed:~
```
It is only possible for histogram_parent_http_pageload_is_ssl to be NULL
because telemetry.main_summary_v4 is a view that selects it from a STRUCT.
If the underlying table changes to not nest this column in a STRUCT this must become:
COUNTIF(ARRAY_LENGTH(histogram_parent_http_pageload_is_ssl) > 0) / COUNT(*) AS reporting_ratio
```
~and starting tonight main summary meets the condition `If the underlying table changes to not nest this column in a STRUCT`~
cc @chutten as data owner


tests/telemetry_derived/ssl_ratios_v1/test_aggregation/expect.ndjson
```diff
@@ -2,4 +2,4 @@
 {"country": "DE", "non_ssl_loads": 4, "os": "Windows_NT", "reporting_ratio": 1.0, "ssl_loads": 2, "submission_date": "2019-08-01"}
 {"country": "MX", "non_ssl_loads": 10, "os": "Windows_NT", "reporting_ratio": 1.0, "ssl_loads": 20, "submission_date": "2019-08-01"}
 {"country": "US", "non_ssl_loads": 20, "os": "Darwin", "reporting_ratio": 1.0, "ssl_loads": 30, "submission_date": "2019-08-01"}
-{"country": "US", "non_ssl_loads": 5, "os": "Windows_NT", "reporting_ratio": 0.4, "ssl_loads": 5, "submission_date": "2019-08-01"}
+{"country": "US", "non_ssl_loads": 5, "os": "Windows_NT", "reporting_ratio": 0.25, "ssl_loads": 5, "submission_date": "2019-08-01"}
```
jklukas: Is this value changing due to the COUNTIF change?

relud: this value is changing because the test data can no longer differentiate between an empty list and a null list, so an input row that tested that differentiation was removed below: https://github.com/mozilla/bigquery-etl/pull/502/files#diff-c56a03fdceac19ccb90ad9bd8444b607R12-L14

so the correct value went from 2 in 5 to 1 in 4

jklukas: Thank you. I read through that file several times and couldn't see what had changed.

relud:
I just realized that we can preserve existing behavior here by reading from `main` instead of `main_summary`

chutten:
_looks around, sees nothing to do. Approves._

PR #503, opened 2019-11-14
Re-add aggregated numerical values in search

Ben-Wu:
@fbertsch @relud thoughts on this approach? Would this be preferred over the alternatives that were discussed (derive search clients daily from clients daily or join the two)?


templates/search_derived/search_clients_daily_v8/query.sql
```diff
@@ -51,9 +62,24 @@ WITH
...
+    -- TODO: experiments
   FROM
-    telemetry.main_summary ),
+    telemetry.main_summary
+  WINDOW
+    w1 AS (
+      PARTITION BY
+        client_id,
+        submission_date
+    )),
```
relud: beautiful ❤️

templates/search_derived/search_clients_daily_v8/query.sql
```diff
@@ -95,6 +121,13 @@ WITH
     udf_mode_last(ARRAY_AGG(default_private_search_engine_data_load_path) OVER w1) AS default_private_search_engine_data_load_path,
     udf_mode_last(ARRAY_AGG(default_private_search_engine_data_submission_url) OVER w1) AS default_private_search_engine_data_submission_url,
     udf_mode_last(ARRAY_AGG(sample_id) OVER w1) AS sample_id,
+    udf_mode_last(ARRAY_AGG(subsession_hours_sum) OVER w1) AS subsession_hours_sum,
+    udf_mode_last(ARRAY_AGG(sessions_started_on_this_day) OVER w1) AS sessions_started_on_this_day,
+    udf_mode_last(ARRAY_AGG(active_addons_count_mean) OVER w1) AS active_addons_count_mean,
+    udf_mode_last(ARRAY_AGG(max_concurrent_tab_count_max) OVER w1) AS max_concurrent_tab_count_max,
+    udf_mode_last(ARRAY_AGG(tab_open_event_count_sum) OVER w1) AS tab_open_event_count_sum,
+    udf_mode_last(ARRAY_AGG(active_hours_sum) OVER w1) AS active_hours_sum,
+    udf_mode_last(ARRAY_AGG(total_uri_count) OVER w1) AS total_uri_count,
```
relud: all the values will be the same because these values are already aggregated by a subset of the partitions for w1, and there's no requirement to aggregate them because this is using windows, so these shouldn't be aggregated.
i.e.
```diff
-    udf_mode_last(ARRAY_AGG(subsession_hours_sum) OVER w1) AS subsession_hours_sum,
-    udf_mode_last(ARRAY_AGG(sessions_started_on_this_day) OVER w1) AS sessions_started_on_this_day,
-    udf_mode_last(ARRAY_AGG(active_addons_count_mean) OVER w1) AS active_addons_count_mean,
-    udf_mode_last(ARRAY_AGG(max_concurrent_tab_count_max) OVER w1) AS max_concurrent_tab_count_max,
-    udf_mode_last(ARRAY_AGG(tab_open_event_count_sum) OVER w1) AS tab_open_event_count_sum,
-    udf_mode_last(ARRAY_AGG(active_hours_sum) OVER w1) AS active_hours_sum,
-    udf_mode_last(ARRAY_AGG(total_uri_count) OVER w1) AS total_uri_count,
+    subsession_hours_sum,
+    sessions_started_on_this_day,
+    active_addons_count_mean,
+    max_concurrent_tab_count_max,
+    tab_open_event_count_sum,
+    active_hours_sum,
+    total_uri_count,
```

PR #519, opened 2019-11-19
Move nondesktop KPI queries to stable table DAG

jklukas:
This makes it explicit that we no longer are using imported Parquet data.
It also moves several of these tables to the shared-prod project where
we want them to live long-term.


templates/telemetry/core_clients_daily_v1/query.sql
```diff
@@ -1,23 +1,14 @@
 WITH
```
relud: is this supposed to move to `templates/telemetry_derived/core_clients_daily_v1/query.sql`?

relud: or perhaps it's that it's in both places and just wasn't removed?

relud: i like the one in `templates/telemetry_derived/core_clients_daily_v1/query.sql` better

jklukas: Yes, this looks like a mistake. This should have been replaced with a view.

jklukas: > i like the one in `templates/telemetry_derived/core_clients_daily_v1/query.sql` better

Why? You prefer explicitly listing `DATE(submission_timestamp)` instead of `submission_date` every time?

relud: idk, i can't see them both to compare anymore. probably nothing significant, i blame the comment on morning brain.

PR #524, opened 2019-11-22
Add tables to replace experiment enrollment aggregates spark streamin…

sunahsuh:
…g job

Note: build is failing because some of the views depend on the existence of new tables/views also in this PR. I'm going to submit this for review anyway, and if things look good, create tables as necessary and re-kick CI

This set of tables and views recreates the data necessary to feed the live grafana experiment enrollments dashboards. Briefly:
`telemetry_derived.experiment_enrollment_aggregates_v1` is a table based on the `events` table which aggregates normandy enrollment/unenrollment/error events over five-minute windows. The table schema matches the previous test tube database.
`telemetry_derived.events_live` is a view over the `telemetry_live.event_v4` table which is schema-compatible with the main_events and event_events tables.
`telemetry_derived.experiment_enrollment_aggregates_live` is a union view based on the materialized aggregates table and the `events_live` view. This view is parametrized by submission_date, and will be run after the conclusion of each day's aggregation query to use the materialized aggregations as much as possible (the intent is to keep costs down for the live dashboard -- current costs are ~40gbs scanned per query by the end of the day, and there are 6 graphs hitting this particular view on the dashboard)
`telemetry_derived.experiments_daily_active_clients_v1` is a materialized table of each day's population in each experiment and branch, based on `clients_daily` (while this may appear to provide modest savings, some experiments run for several months, in which case the costs add up.)


templates/telemetry_derived/experiment_enrollment_aggregates_live/view.sql
```diff
@@ -0,0 +1,136 @@
...
+    WHERE
+      event_category = 'normandy'
+      AND date(timestamp)
+  > @submission_date
+  GROUP BY
+    1,
+    2,
+    3,
+    4,
+    5
```
akkomar: Since BigQuery allows that, I would use column aliases here for clarity and to avoid errors during refactors later.

templates/telemetry_derived/experiment_enrollment_aggregates_live/view.sql
```diff
@@ -0,0 +1,136 @@
...
+      COUNTIF(event_method = 'update') AS update_count,
+      COUNTIF(event_method = 'enrollFailed') AS enroll_failed_count,
+      COUNTIF(event_method = 'unenrollFailed') AS unenroll_failed_count,
+      COUNTIF(event_method = 'updateFailed') AS update_failed_count
+    FROM
+      `moz-fx-data-shared-prod`.telemetry_derived.events_live
+    WHERE
+      event_category = 'normandy'
+      AND date(timestamp)
+  > @submission_date
```
akkomar: How about making this a single line?

templates/telemetry_derived/experiment_enrollment_aggregates_live/view.sql
```diff
@@ -0,0 +1,136 @@
...
+  *,
+  SUM(enroll_count) OVER (
+    PARTITION BY
+      experiment,
+      branch
+    ORDER BY
+      window_start
+    ROWS BETWEEN
+      UNBOUNDED PRECEDING
+      AND CURRENT ROW
```
akkomar: This window frame clause is repeated 3x below. What do you think about defining it once at the bottom and calling it by name here?

templates/telemetry_derived/experiment_enrollment_aggregates_live/view.sql
```diff
@@ -0,0 +1,136 @@
...
+  ) AS cumulative_update_count,
+  SUM(enroll_failed_count) OVER (
+    PARTITION BY
+      experiment,
+      branch
+    ORDER BY
+      window_start
```
akkomar: Can we use the same window frame here as above? (was the the intent?)
Implicitly it is `RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW` here, which I think is equivalent.

templates/telemetry_derived/experiment_enrollment_aggregates_v1/query.sql
```diff
@@ -0,0 +1,57 @@
...
+  all_events
+WHERE
+  submission_date = @submission_date
+  AND event_category = 'normandy'
+GROUP BY
+  1,
+  2,
+  3,
+  4,
+  5
```
akkomar: How about aliases here too? :)

templates/telemetry_derived/experiment_enrollment_aggregates_live/view.sql
sunahsuh: This was the built-in formatter :/ @relud is the formatter's results a suggestion or canon?

templates/telemetry_derived/experiment_enrollment_aggregates_live/view.sql
sunahsuh: Ah I didn't know that was a thing you can do!

templates/telemetry_derived/experiment_enrollment_aggregates_live/view.sql
relud: suggestion

sql/telemetry_derived/experiment_enrollment_aggregates_live/view.sql
```diff
@@ -0,0 +1,91 @@
...
+    WHERE
+      event_category = 'normandy'
+      AND date(timestamp)
+  > @submission_date
+  GROUP BY
+    `type`,
+    experiment,
+    branch,
+    window_start,
+    window_end
```
relud: this is a due to a bug in how script/format-sql is handling `>`s. It currently can't differentiate between `>` as a closing bracket for a complex data type (e.g. `ARRAY<INT64>`), and `>` as an operator (e.g. `DATE(timestamp) > @submission_date`)

this should be formatted as:

```sql
    WHERE
      event_category = 'normandy'
      AND DATE(timestamp) > @submission_date
    GROUP BY
      `type`,
      experiment,
      branch,
      window_start,
      window_end
```

sql/telemetry_derived/experiment_enrollment_aggregates_live/view.sql
```diff
@@ -0,0 +1,91 @@
...
+        INTERVAL (
+          CAST(
+            EXTRACT(
+              MINUTE
+              FROM
+                `timestamp`
+            ) / 5 AS int64
+          ) * 5
+        ) MINUTE
+      ) AS window_start,
```
relud: I think this needs a comment indicating that it is truncating timestamp to 5 minute granularity.

sql/telemetry_derived/experiment_enrollment_aggregates_live/view.sql
```diff
@@ -0,0 +1,91 @@
...
+      TIMESTAMP_ADD(
+        TIMESTAMP_TRUNC(`timestamp`, HOUR),
+        INTERVAL (
+          CAST(
+            EXTRACT(
+              MINUTE
+              FROM
+                `timestamp`
+            ) / 5 AS int64
+          ) * 5
```
relud: casting a float to an integer will round, not truncate. I believe what we should actually do here is use the integer division function [DIV](https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions#div)

```diff
-          CAST(
-            EXTRACT(
-              MINUTE
-              FROM
-                `timestamp`
-            ) / 5 AS int64
-          ) * 5
+          DIV(
+            EXTRACT(MINUTE FROM `timestamp`),
+            5
+          ) * 5
```

sql/telemetry_derived/experiment_enrollment_aggregates_live/view.sql
```diff
@@ -0,0 +1,91 @@
...
+      COUNTIF(event_method = 'update') AS update_count,
+      COUNTIF(event_method = 'enrollFailed') AS enroll_failed_count,
+      COUNTIF(event_method = 'unenrollFailed') AS unenroll_failed_count,
+      COUNTIF(event_method = 'updateFailed') AS update_failed_count
+    FROM
+      `moz-fx-data-shared-prod`.telemetry_derived.events_live
+    WHERE
+      event_category = 'normandy'
+      AND date(timestamp)
+  > @submission_date
```
relud: this causes the error: `Query parameters cannot be used inside SQL view bodies`

sql/telemetry_derived/experiment_enrollment_aggregates_live/view.sql
Ben-Wu: This is what I did for error aggregates https://github.com/mozilla/bigquery-etl/blob/master/udf/round_timestamp_to_minute.sql

sql/telemetry_derived/events_live/view.sql
```diff
@@ -0,0 +1,39 @@
...
+    normalized_app_name AS app_name,
+    metadata.uri.app_version AS app_version,
+    normalized_os AS os,
+    normalized_os_version AS os_version,
+    environment.experiments AS experiments,
+    sample_id,
+    payload.session_id AS session_id,
+    SAFE.TIMESTAMP_MILLIS(payload.process_start_timestamp) AS session_start_time,
+    payload.subsession_id AS subsession_id,
+    submission_timestamp AS `timestamp`,
```
relud: I think it would be consistent with what we have done for other tables to expose both `submission_timestamp` and `timestamp` in views, and prefer the use of `submission_timestamp`

relud: to be clear, this would mean exposing `submission_timestamp` both here and in the `telemetry.events_v1` view

sql/telemetry_derived/experiment_enrollment_aggregates_live/view.sql
```diff
@@ -0,0 +1,91 @@
+CREATE
+OR REPLACE VIEW `moz-fx-data-shared-prod`.telemetry_derived.experiment_enrollment_aggregates_live AS (
```
relud: It's inconsistent with existing view definitions to wrap the query part in parentheses
```suggestion
OR REPLACE VIEW `moz-fx-data-shared-prod`.telemetry_derived.experiment_enrollment_aggregates_live AS
```

relud: for consistency the table_id should be be quoted entirely, rather than partially:
```suggestion
OR REPLACE VIEW `moz-fx-data-shared-prod.telemetry_derived.experiment_enrollment_aggregates_live` AS (
```

sql/telemetry_derived/experiment_enrollment_aggregates_live/view.sql
relud: filed #530 to fix that

sql/telemetry_derived/events_live/view.sql
sunahsuh: Okay – I'll file a bug for that and take care of it separately

sql/telemetry_derived/experiment_enrollment_aggregates_live/view.sql
sunahsuh: 😢

sunahsuh:
Blargh, don't re-review yet, need to figure out how to differentiate between params for the python script vs the bq command

sunahsuh:
Okay, ready for re-rereview – I ended up deciding to split query generation vs query running as two different tasks via airflow's xcom mechanism, which has the added benefit of giving us a path of passing xcoms for other uses (e.g. the experiments dataset job, which polls the normandy api.)

PR #527, opened 2019-11-22
Give precedence to content_shutdown over content crashes

Ben-Wu:
A content shutdown crash does not count as a content crash


wlach:
It appears as if there is a circle failure which seems unrelated to this PR

PR #528, opened 2019-11-22
Improve format-sql for views and timestamp functions

relud:
this should fix some of the unusual formatting results in #524


relud:
dry_run failures are unrelated

PR #532, opened 2019-11-22
Make addons and addon_aggregates exactly replace spark versions

relud:


relud:
the v3 query is what we were originally planning to migrate to, and i removed it to avoid any future confusion.

PR #533, opened 2019-11-25
Address glam dag timeouts

emtwo:
There are a bunch of changes here to speed up the glam dag computations. Here is a summary:

* Both `clients_scalar_aggregates` and `clients_histogram_aggregates` are now computed by incorporating 1 new day of data at a time with existing aggregations. This is faster than processing 90 days of data each time

* Both `clients_scalar_aggregates` and `clients_histogram_aggregates` now have `init.sql` files that include partitions and clustering

* In some cases, data was being unnested and grouped prematurely and it was more efficient to group before unnesting.

* Computations for scalar buckets were redundant and resulted in a static array anyways, so I've hardcoded it for now. We're currently in discussion with Data Science to think of better bucketing strategies for scalars so this will likely be changing again.

### DIFF
--- a/sql/telemetry_derived/clients_histogram_aggregates_v1/query.sql
+++ b/sql/telemetry_derived/clients_histogram_aggregates_v1/query.sql
@@ -10,66 +10,67 @@ CREATE TEMP FUNCTION udf_map_sum(entries ANY TYPE) AS (
   )
 );
 --
-CREATE TEMP FUNCTION udf_normalized_sum (arrs ARRAY<STRUCT<key STRING, value INT64>>)
-RETURNS ARRAY<STRUCT<key STRING, value FLOAT64>> AS (
-  -- Returns the normalized sum of the input maps.
-  -- It returns the total_count[k] / SUM(total_count)
-  -- for each key k.
+CREATE TEMP FUNCTION udf_merged_user_data(old_aggs ANY TYPE, new_aggs ANY TYPE)
+  RETURNS ARRAY<STRUCT<
+    first_bucket INT64,
+    last_bucket INT64,
+    num_buckets INT64,
+    latest_version INT64,
+    metric STRING,
+    metric_type STRING,
+    key STRING,
+    agg_type STRING,
+    aggregates ARRAY<STRUCT<key STRING, value INT64>>>> AS (
   (
-    WITH total_counts AS (
-      SELECT
-        sum(a.value) AS total_count
-      FROM
-        UNNEST(arrs) AS a
-    ),
+    WITH unnested AS
+      (SELECT *
+      FROM UNNEST(old_aggs)
 
-    summed_counts AS (
-      SELECT
-        a.key AS k,
-        SUM(a.value) AS v
-      FROM
-        UNNEST(arrs) AS a
+      UNION ALL
+
+      SELECT *
+      FROM UNNEST(new_aggs)),
+
+    aggregated_data AS
+      (SELECT AS STRUCT
+        first_bucket,
+        last_bucket,
+        num_buckets,
+        latest_version,
+        metric,
+        metric_type,
+        key,
+        agg_type,
+        udf_map_sum(ARRAY_CONCAT_AGG(aggregates)) AS histogram_aggregates
+      FROM unnested
       GROUP BY
-        a.key
-    ),
+        first_bucket,
+        last_bucket,
+        num_buckets,
+        latest_version,
+        metric,
+        metric_type,
+        key,
+        agg_type)
 
-    final_values AS (
-      SELECT
-        STRUCT<key STRING, value FLOAT64>(k, 1.0 * v / total_count) AS record
-      FROM
-        summed_counts
-      CROSS JOIN
-        total_counts
-    )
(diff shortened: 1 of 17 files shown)


sql/telemetry_derived/clients_histogram_aggregates_v1/query.sql
```diff
@@ -143,23 +147,93 @@ normalized_histograms AS
...
+  LEFT JOIN latest_versions
+  ON latest_versions.channel = histogram_aggs.channel
+  WHERE app_version >= (latest_version - 2)),
+
+joined_new_old AS (
+  SELECT
+    CASE
+      WHEN old_data.client_id IS NOT NULL THEN old_data.client_id
+      ELSE new_data.client_id
+    END AS client_id,
```
jklukas: This can be conveyed more concisely as `COALESCE(old_data.client_id, new_data.client_id) AS client_id`, same for the following case statements here.

templates/telemetry_derived/clients_histogram_aggregates_v1/query.sql
```diff
@@ -131,23 +135,93 @@ normalized_histograms AS
...
+      ELSE new_data.app_build_id
+    END AS app_build_id,
+    CASE
+      WHEN old_data.channel IS NOT NULL THEN old_data.channel
+      ELSE new_data.channel
+    END AS channel,
+    old_data.histogram_aggregates AS old_aggs,
+    new_data.histogram_aggregates AS new_aggs
+  FROM clients_histogram_aggregates_new AS new_data
+  FULL OUTER JOIN clients_histogram_aggregates_old AS old_data
```
jklukas: Given that this is now an incremental query, does old_data ever age out? Is this query now effectively looking at more than 90 days of history? Or does data naturally fall out because we're only looking two versions back, so we end up dropping old data once new FF versions are released?

emtwo: Yes exactly, the data naturally falls out since we're only looking 2 versions back.

templates/telemetry_derived/clients_scalar_aggregates_v1/init.sql
```diff
@@ -0,0 +1,14 @@
...
+    app_version INT64,
+    app_build_id STRING,
+    channel STRING,
+    scalar_aggregates ARRAY <STRUCT<metric STRING,
+    metric_type STRING,
+    key STRING,
+    agg_type STRING,
+    value FLOAT64>>)
+PARTITION BY RANGE_BUCKET(app_version, GENERATE_ARRAY(30, 200, 1))
+CLUSTER BY app_version, channel, client_id
```
jklukas: If app_version is a discrete integer used for partitioning already, I don't think there's any benefit from including it as a clustering field. There may be some small performance penalty for including it.
```suggestion
CLUSTER BY channel, client_id
```

emtwo: Yes, unfortunately this is a bit strange, but I'll explain why I ended up doing this.

When I created a partition by `app_version`  and require that queries filter by partition, if I do something like `WHERE app_version >= (latest_version - 2)` (which is how the filtering is done in the query now) it complains that I'm not filtering by app version

It's only happy if I explicitly say `WHERE app_version > 50` for example. Not surprising I guess since this feature is in beta. This made me suspect that we're not getting partitioning gains by partitioning on `app_version` and that maybe we can make up for it by clustering on it as well. I haven't confirmed this theory. 

However, in response to your question above, I've confirmed performance gains overall when incorporating all the changes in this PR.

emtwo:
> IIRC some of these queries are exempted from dry runs due to taking too long. It would be worth checking whether that situation is better with the modifications and if we might be able to allow tests to dry run them again.

Dry run is looking good without the exemptions now!

templates/telemetry_derived/clients_scalar_aggregates_v1/init.sql
jklukas: We'll only ever have three app_version values in the table at a time, correct? If that's true, I don't see much benefit from partitioning on app_version. It's best to avoid a beta feature if possible, too. And it looks like we'd be able to avoid all the casting between int and string if we don't use app_version as a partitioning field?

Clustering requires partitioning, but we could introduce a dummy date-type column to be the partitioning column if needed.

> When I created a partition by app_version and require that queries filter by partition, if I do something like WHERE app_version >= (latest_version - 2) (which is how the filtering is done in the query now) it complains that I'm not filtering by app version

This may be a limitation for date partitioning too, if the value you're trying to filter on can't be known until looking up results from a subquery.

emtwo: It can have up to 5 values for `app_version` at a time since the `latest_version` is different depending on the channel. Even if we don't cluster on `app_version`, it would still be useful to be able to cluster on the other fields.

I like the dummy date-type column idea. I will file a follow-up issue to switch to that. 
Would just like to get the ball rolling on this to see how it goes, so I'll circle back to fix this.

PR #537, opened 2019-11-26
Fix new fields in main_summary.active_addons

relud:


relud:
> it would be helpful to have a description of the issue and what this PR is fixing

behavior: for `foreign_install` and `user_disabled` provided via the bigquery fields, `0` values were coerced to `TRUE` instead of `FALSE`, and for `version` provided via the bigquery field, empty string values were converted to `NULL`.

issue: for `foreign_install` and `user_disabled` integer values are provided to the javascript UDF as string values, and string `0`s are coerced to `TRUE`, while integer `0`s are converted to `FALSE` when returning a boolean. for `version` empty strings **don't** short-circuit the `||` operator, so the json value was being checked, which returns `undefined` and gets coerced to `NULL`.

how this fixes it: we define `ifnull` so that integer `0`s and empty strings **do** short-circuit, but we also define `maybeParseInt` so that `null` and `undefined` become `null`, so that when provided to `ifnull` it **doesn't** short-circuit.

PR #539, opened 2019-11-26
Bug 1599488 Force sums to be zero in aggregate_search_counts

jklukas:


relud:
should we run an update query on clients_daily to coalesce these fields to 0?

jklukas:
> should we run an update query on clients_daily to coalesce these fields to 0?

That sounds good to me. It's probably less important to make the same change in clients_last_seen. I will take a look at doing this and how much data it will scan.

PR #540, opened 2019-11-26
 Bug 1593773 - Update main_summary to use new location of FXA_CONFIGURED

relud:
includes new location for sync as well
cc @irrationalagent

### DIFF
--- a/sql/telemetry_derived/main_summary_v4/part1.sql
+++ b/sql/telemetry_derived/main_summary_v4/part1.sql
@@ -271,10 +271,10 @@ SELECT
   creation_date,
   environment.partner.distribution_id,
   DATE(submission_timestamp) AS submission_date,
-  -- See bug 1550752
-  udf_boolean_histogram_to_boolean(payload.histograms.fxa_configured) AS fxa_configured,
-  -- See bug 1232050
-  udf_boolean_histogram_to_boolean(payload.histograms.weave_configured) AS sync_configured,
+  -- See bugs 1550752 and 1593773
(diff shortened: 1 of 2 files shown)


sql/telemetry_derived/main_summary_v4/part1.sql
```diff
@@ -272,9 +272,9 @@ SELECT
   environment.partner.distribution_id,
   DATE(submission_timestamp) AS submission_date,
   -- See bug 1550752
-  udf_boolean_histogram_to_boolean(payload.histograms.fxa_configured) AS fxa_configured,
+  ifnull(environment.services.account_enabled, udf_boolean_histogram_to_boolean(payload.histograms.fxa_configured)) AS fxa_configured,
```
jklukas: Apparently IFNULL is a synonym for COALESCE. I'm more familiar with COALESCE. Any argument for preferring one over the other?

jklukas: Should we update the bug numbers to the ones that describe the new values?

relud: i like using `IFNULL` when it's only 2 fields, because it's shorter. i'm okay with preferring to never use `IFNULL`.

relud: i'm going to merge this with `IFNULL` for the moment, and file a follow-up issue to consider banning `IFNULL` in favor of `COALESCE`.

jklukas: > i like using `IFNULL` when it's only 2 fields, because it's shorter. i'm okay with preferring to never use `IFNULL`.

If it's shorter only in that the name of the function is fewer characters, I'd prefer to use COALESCE everywhere as the more general function.

PR #543, opened 2019-11-26
Bug 1593773 - Move the main summary cutover date back to 2019-10-25

relud:
draft so that this isn't merged until backfilling to 2019-10-25 finishes


relud:
clients_daily and clients_last_seen also need the backfill, because they use `sync_configured`

relud:
relevant tables have been copied from derived-datasets back at least as far as `2019-10-25`, so this PR can proceed. I will run the actual backfill for relevant dates separately.

PR #544, opened 2019-11-26
Add wau and dau to firefox_kpi_dashboard query

jklukas:


templates/telemetry/firefox_kpi_dashboard_v1/query.sql
```diff
@@ -171,27 +248,17 @@ WITH
...
-    `date`,
-    mau,
-    -- The confidence interval is chosen here somewhat arbitrarily as 0.5% of
-    -- the value; in any case, we should show a larger interval than we do for
-    -- non-imputed actuals.
-    mau - mau * 0.005 AS mau_low,
-    mau + mau * 0.005 AS mau_high
-  FROM
-    static.firefox_desktop_imputed_mau28_v1 )
+  LEFT JOIN
```
jklukas: The LEFT JOIN and COALESCE here is a more general way of including the imputed values. We have to move to this method since we now are including dau and wau, which aren't present in the imputed table.

templates/telemetry/firefox_kpi_dashboard_v1/query.sql
```diff
@@ -61,7 +63,25 @@ WITH
...
+        0)) AS mau,
+    SUM(
+    IF
+      (country IN ('US',
+          'FR',
+          'DE',
+          'GB',
+          'CA'),
+        wau,
+        0)) AS wau,
```
relud: defining a query-scope UDF `udf_country_is_tier1` would remove the need to repeat this list so much

templates/telemetry/firefox_kpi_dashboard_v1/query.sql
```diff
@@ -121,7 +163,27 @@ WITH
...
+    SUM(seen_in_tier1_country_mau) AS mau,
+    SUM(
+    IF
+      (country IN ('US',
+          'FR',
+          'DE',
+          'GB',
+          'CA'),
+        wau,
+        0)) AS wau,
```
relud: i suppose we don't have a historical definition of tier1 FxA wau that we need to match, and we're already mixing the definitions in this PR, but it seems odd to use both "seen in tier1" and "last seen in tier1" in the same row.

templates/telemetry/firefox_kpi_dashboard_v1/query.sql
relud: elegant solution

templates/telemetry/firefox_kpi_dashboard_v1/query.sql
jklukas: Agreed it's a bit weird, but we generally have agreement to use the last seen values, and I think it's good to keep the "seen in tier 1" exception as small as possible.

PR #549, opened 2019-11-27
Normalize metadata in generated views on historical ping tables

jklukas:
Merging this change will cause the changes to be used on the next deploy of the schema generation pipeline.

See https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/Jenkinsfile.bigquery.prod#L105

Closes https://github.com/mozilla/gcp-ingestion/issues/631


jklukas:
I'll hold off on merging this until next week.

jklukas:
cc @whd in case you have worries that this might break things.

whd:
LGTM

PR #553, opened 2019-11-30
Fix #550: create airflow xcom output file in dockerfile and make writ…

### LINKED ISSUE #550: Container entrypoint broken on Jenkins
Caused by https://github.com/mozilla/bigquery-etl/pull/524/files#diff-2991ff0ca3a63f2bcba3249c8c35b284R33. In the case of Jenkins, the container is [not being run as a privileged user](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/Jenkinsfile.bigquery.stage#L105) and so fails to create the directory at the root filesystem.

sunahsuh:
…eable to all users

Not sure how to test this, unfortunately, since I'm not 100% sure where this is breaking in the first place, but from the description in the issue it seems this should fix it.


script/entrypoint
```diff
@@ -32,6 +32,5 @@ elif [ "$1" = "query" ]; then
 else
     # KubernetesPodOperator will extract the contents of /airflow/xcom/return.json as an xcom
     # if the xcom_push parameter is true
-    mkdir -p /airflow/xcom && touch /airflow/xcom/return.json
     exec "$@" > /airflow/xcom/return.json
```
relud: > I'd also like to not lose error logs in the exec case i.e. use `&>`

Good point, I think we should be preserving stdout unless specifically requested. 

I think we should handle this before this conditional block via `exec`, e.g. add `if "${XCOM_PUSH:-false}"; then exec > /airflow/xcom/return.json; fi` around line 19, and remove the redirect here.
```suggestion
     exec "$@"
```

sunahsuh: Any reason we'd prefer that over using `tee`?

whd: I don't know much about the xcom stuff but as long as stdout/err are preserved on normal jenkins runs, I'm easy as to the underlying mechanism. To that end `a+w` (assuming umask is fine) and `tee` would suffice.

relud: > Any reason we'd prefer that over using `tee`?

it applies to the whole script, rather than just this line.

relud: but i'd settle for `tee`

PR #557, opened 2019-12-02
Add fxa usage criterion that filters out monitor email engagement

jklukas:
This change allows us to compare FxA MAU calculations that include or exclude the monitor email engagement events as added in https://github.com/mozilla/fxa/issues/2743

### DIFF
--- a/sql/telemetry/fxa_users_daily_v1/query.sql
+++ b/sql/telemetry/fxa_users_daily_v1/query.sql
@@ -59,7 +59,8 @@ WITH
     udf_mode_last(ARRAY_AGG(os_name) OVER w1) AS os_name,
     udf_mode_last(ARRAY_AGG(os_version) OVER w1) AS os_version,
     udf_contains_tier1_country(ARRAY_AGG(country) OVER w1) AS seen_in_tier1_country,
-    udf_contains_registration(ARRAY_AGG(event_type) OVER w1) AS registered
+    udf_contains_registration(ARRAY_AGG(event_type) OVER w1) AS registered,
+    COUNTIF(event_type != 'fxa_rp - engage') OVER w1 = 0 AS rp_engage_only
   FROM
     fxa_all_events_v1
   WHERE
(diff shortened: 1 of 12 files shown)


sql/telemetry/fxa_users_daily_v1/query.sql
```diff
@@ -59,7 +59,8 @@ WITH
     udf_mode_last(ARRAY_AGG(os_name) OVER w1) AS os_name,
     udf_mode_last(ARRAY_AGG(os_version) OVER w1) AS os_version,
     udf_contains_tier1_country(ARRAY_AGG(country) OVER w1) AS seen_in_tier1_country,
-    udf_contains_registration(ARRAY_AGG(event_type) OVER w1) AS registered
+    udf_contains_registration(ARRAY_AGG(event_type) OVER w1) AS registered,
+    COUNTIF(event_type != 'fxa_rp - engage') OVER w1 = 0 AS rp_engage_only
```
jklukas: This new `rp_engage_only` dimension is set to true when the only engagement event we saw from a user in a given day was `fxa_rp - engage`, with the idea that we use this down the line to be able to track MAU unaffected by this event type.

Does that sound reasonable, @irrationalagent ?

irrationalagent: Yes, this seems like it would work for now. However long term other Relying Parties (RPs) of FxA may want to use this event. As implemented here, it would flag all users generating this event regardless of RP. If we want to selectively flag users who generated the event via monitor, we would have to check for `service = fx-monitor` in addition. I don't know of any other RPs that have plans in the works to this right now, but FxA are setup to support them if they want to.

sql/telemetry/fxa_users_last_seen_raw_v1/init.sql
```diff
@@ -11,6 +11,7 @@ SELECT
   0 AS days_seen_bits,
   0 AS days_seen_in_tier1_country_bits,
   0 AS days_registered_bits,
+  0 AS days_seen_no_monitor_bits,
```
irrationalagent: Similar to above, RPs other than monitor may also want to use this event, so would it make to rename this to be more general `days_seen_no_rp_enage_bits` even though its less transparent (assuming we will want to continue to filter this event for all RPs)?

jklukas:
I updated to make this monitor-specific. Really this field is intended only for the next few weeks so we can compare FxA mau calculated the two different ways through the end of 2019.

PR #559, opened 2019-12-02
Bug 1543434 - Create search datasets for mobile

Ben-Wu:
Looking for feedback on this.  There's some clunkiness in handling the inconsistencies in data like normalizing the search keys and unioning the core and fenix data.


sql/search_derived/mobile_search_clients_daily_v1/query.sql
```diff
@@ -0,0 +1,168 @@
...
+    udf_mode_last(ARRAY_AGG(normalized_country_code)) AS country,
+    udf_mode_last(ARRAY_AGG(locale)) AS locale,
+    'Fenix' AS app_name,
+    udf_mode_last(ARRAY_AGG(client_info.app_display_version)) AS app_version,
+    udf_mode_last(ARRAY_AGG(normalized_channel)) AS channel,
+    udf_mode_last(ARRAY_AGG(normalized_os)) AS os,
+    udf_mode_last(ARRAY_AGG(client_info.android_sdk_version)) AS os_version,
+    udf_mode_last(ARRAY_AGG(metrics.string.search_default_engine_code)) AS default_search_engine,
+    udf_mode_last(ARRAY_AGG(metrics.string.search_default_engine_submission_url)) AS default_search_engine_submission_url,
+    CAST(NULL AS STRING) AS distribution_id,
```
fbertsch: We _should_ have distribution information in Fenix. If we don't we should file an issue, but I'm not seeing it anywhere.

sql/search_derived/mobile_search_clients_daily_v1/query.sql
```diff
@@ -0,0 +1,168 @@
...
+));
+
+-- Older versions have source.engine instead of engine.source
+-- Assume action is one of actionbar, listitem, suggestion, or quicksearch
+-- https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/data/core-ping.html#searches
+-- Return array of form [source, engine] if key is valid, empty array otherwise
+CREATE TEMP FUNCTION normalize_core_search_key(key STRING) AS (
+  CASE
+    WHEN ARRAY_LENGTH(SPLIT(key, '.')) != 2 THEN []
+    WHEN SPLIT(key, '.')[OFFSET(0)] IN UNNEST(['actionbar', 'listitem', 'suggestion', 'quicksearch'])
```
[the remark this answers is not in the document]
Ben-Wu: Oh this is normalizing to the newer way.  I just mixed it up in the comment.  The logic here is it gets reversed if the value in the first index is a source

sql/search_derived/mobile_search_clients_daily_v1/query.sql
```diff
@@ -0,0 +1,168 @@
...
+  CROSS JOIN
+    UNNEST(
+      -- Add a null search to pings that have no searches
+      IF(ARRAY_LENGTH(searches) = 0,
+         null_search(),
+         searches
+      )) AS searches
+),
+
+-- baseline has locale but not default search engine, metrics has default search engine but not locale
```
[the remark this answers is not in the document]
Ben-Wu: Sure

templates/search_derived/mobile_search_clients_daily_v1/query.sql
```diff
@@ -0,0 +1,144 @@
...
+    udf_mode_last(ARRAY_AGG(sample_id)) AS sample_id
+  FROM
+    core_flattened_searches
+  GROUP BY
+    client_id,
+    submission_date,
+    engine,
+    source,
+    app_name
+  UNION ALL
```
relud: aggregation should be done after the union instead of before it, to reduce repeated logic, unless there is a significant performance gain from doing it this way.

udf/iso8601_str_to_date.sql
```diff
@@ -0,0 +1,23 @@
+/*
+
+Take a ISO 8601 date or date and time string and return a DATE.  Return null if parse fails.
+
+Possible formats: 2019-11-04, 2019-11-04T21:15:00+00:00, 2019-11-04T21:15:00Z, 20191104T211500Z
+
+ */
+
+CREATE TEMP FUNCTION udf_iso8601_str_to_date(date_str STRING) RETURNS DATE AS (
```
relud: nit: I think this should be named `parse_iso8601_date`, to be more consistent with the name of the function it's wrapping

sql/search_derived/mobile_search_clients_daily_v1/query.sql
Ben-Wu: There are open issues in Fenix for the distribution info so I assume it's not available yet.  By "should" do you mean that this should already be done or that this is something we want to have?

fbertsch: By "should" I mean, if we don't we should. I thought it was in place but apparently not :)

Ben-Wu:
I made it so high search counts are replaced with a null search engine/source and 0 search count.  I assume this is preferred over removing the client because we would want entries for all clients.

PR #564, opened 2019-12-03
Upgrade to python 3.8

relud:
I'd rather unpin the python minor version, but it doesn't pass CI 😢


jklukas:
> I'd rather unpin the python minor version, but it doesn't pass CI 😢

Python minor version releases are infrequent and significant enough that I'd prefer to keep it explicit anyway.

PR #565, opened 2019-12-03
Reimplement experiments_v1 as SQL

relud:


templates/telemetry_derived/experiments_v1/get_experiment_list.py
```diff
@@ -0,0 +1,51 @@
...
+"""Get experiment list for a given date from the recipe server."""
+
+from argparse import ArgumentParser
+import json
+import sys
+import time
+
+import requests
+
+experiments_qualifying_actions = {"preference-experiment", "opt-out-study"}
```
sunahsuh: I have a more extensive revision to this script I'd like to submit later so we can use the v3 api and do backfills correctly, but for now this will fix bug 1587231, if you don't mind rolling this in
```suggestion
experiments_qualifying_actions = {"preference-experiment", "opt-out-study", "branched-addon-study", "multi-preference-experiment"}
```

templates/telemetry_derived/experiments_v1/get_experiment_list.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def get_experiment_list(date):
+    for recipe in requests.get(recipes_url).json():
+        arguments = recipe.get("arguments") or {}
+        branches = arguments.get("branches")
+        name = arguments.get("slug", arguments.get("name"))
+        last_updated = time.strptime(recipe.get("last_updated"), timestamp_format)
+        if (
+            "branches" in arguments
```
sunahsuh: I'm confused at this clause (this will exclude old-style addon experiments, I'm fairly certain.) I can't seem to find a corresponding clause in TBV job either. Did this get added accidentally?

templates/telemetry_derived/experiments_v1/get_experiment_list.py
```diff
@@ -0,0 +1,51 @@
...
+            and recipe.get("action") in experiments_qualifying_actions
+            and arguments.get("isHighVolume") in {None, False}
+            and name not in excluded_experiments
+            and (recipe.get("enabled") or last_updated > date)
+        ):
+            yield name
+
+
+def main():
+    print(json.dumps(list(get_experiment_list(parser.parse_args().date))))
```
sunahsuh: As far as I can tell, xcom_push does want the top level of the json result to be an object with keys

```suggestion
    print(json.dumps({"experiment_list": list(get_experiment_list(parser.parse_args().date))}))
```

relud: can confirm from testing, it just wants json, no need for it to be an object.

templates/telemetry_derived/experiments_v1/get_experiment_list.py
relud: in scala we decode this json into `case class NormandyRecipeArguments(branches: List[NormandyRecipeBranch], ...)`, which doesn't make `branches` optional, so `Try(json.extract[NormandyRecipe])` doesn't return a `Success`.

This is why we don't have `"Pioneer Enrollment"` in the current output.

If that behavior is a bug, i'm happy to remove this clause.

templates/telemetry_derived/experiments_v1/get_experiment_list.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python3
+
+"""Get experiment list for a given date from the recipe server."""
+
+from argparse import ArgumentParser
+import json
+import sys
```
relud:
```suggestion
```

PR #567, opened 2019-12-03
Fix json output used for XCOM with experiment_enrollment_aggregates_live

relud:


relud:
> For posterity: per convo w relud, `KubernetesPodOperator` will only return non-keyed XComs, so we need to return a json-formatted but non-keyed result

sorry, this was a miscommunication. `KubernetesPodOperator` will return any JSON.

`KubernetesPodOperator` doesn't set an **XCOM key**, because it pushes by returning the value from `KubernetesPodOperator.execute`

relud:
what this fixes is removing the need for `['return_value']` in `"{{" + "task_instance.xcom_pull({!r})['return_value']".format(xcom_task_id) + "}}"`, so that we can do  `"{{" + "task_instance.xcom_pull({!r})".format(xcom_task_id) + "}}"`

PR #570, opened 2019-12-04
Bug 1601139 - Add query to sample documents per doctype 

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1601139)

### DIFF
--- a/templates/monitoring/document_sample_nonprod_v1/query.sql
+++ b/templates/monitoring/document_sample_nonprod_v1/query.sql
@@ -1,0 +1,69 @@
+WITH extract_decoded AS (
+  SELECT
+    metadata.document_namespace,
+    metadata.document_type,
+    metadata.document_version,
+    '' AS error_message,
+    payload
+  FROM
+    `moz-fx-data-shar-nonprod-efed.payload_bytes_decoded.*`
+  WHERE
+    submission_timestamp
+> TIMESTAMP_SUB(current_timestamp, INTERVAL 1 day)
+),
+extract_error AS (
+  SELECT
+    document_namespace,
+    document_type,
+    document_version,
+    error_message,
+    payload
+  FROM
+    `moz-fx-data-shar-nonprod-efed.payload_bytes_error.*`
+  WHERE
+    submission_timestamp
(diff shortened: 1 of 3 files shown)


templates/monitoring/document_sample_nonprod_v1/query.sql
```diff
@@ -0,0 +1,69 @@
...
+    metadata.document_namespace,
+    metadata.document_type,
+    metadata.document_version,
+    '' AS error_message,
+    payload
+  FROM
+    `moz-fx-data-shar-nonprod-efed.payload_bytes_decoded.*`
+  WHERE
+    submission_timestamp
+> TIMESTAMP_SUB(current_timestamp, INTERVAL 1 day)
```
jklukas: This is strange formatting with the `>` completely unindented. Can we move this onto the previous line?

templates/monitoring/document_sample_nonprod_v1/query.sql
```diff
@@ -0,0 +1,69 @@
...
+AND error_type = 'ParsePayload'
+)
+SELECT
+  n_samples,
+  document_is_decoded,
+  sample.*
+FROM
+  (
```
jklukas: Can we make this another named CTE rather than a subquery? I think it would be easier to follow the logic by not having this nested here.

templates/monitoring/document_sample_nonprod_v1/query.sql
acmiyaguchi: This is a side-effect of the `script/format-sql` script. I can use the default BigQuery or Redash formatter here though.

acmiyaguchi:
The idea is that this will update a table in prod everyday. I add the nonprod suffix here because this is more of a diagnostic for creating the samples (40gb instead of 4tb per query). Once this has been running on a regular schedule and all the pieces have been set up elsewhere (airflow, gcp-ingestion, and mozilla-pipeline-schemas), we can swap to using samples derived from prod. Prod is better for getting samples of data that are at least 1000 documents large. One side-effect of sampling from a sample is that some of the document types have far less than 1000 documents.

PR #571, opened 2019-12-04
Fix divide by 0 errors.

emtwo:
This should fix a divide by 0 failure currently happening in the glam dag


emtwo:
> @emtwo would `ieee_divide` or `safe_divide` work for this case? Using the latter we could coalesce `null` to 0.

@fbertsch Thanks for pointing these out. I've updated to use `safe_divide`

emtwo:
> Awesome. I do want to ensure we're not losing information with the `coalesce`, i.e. are `NULL` values (those with `total_count = 0`, regardless of `v`) different in some way than `0` values (those with `total_count > 0`, but `v = 0`), or is finding out which rows are which still available in the data (by looking at e.g. `total-count` directly)?

Yes you could find out which it is by seeing that all values are 0 in a histogram.

PR #572, opened 2019-12-04
Bug 1591207 - Create static list of normalized os

Ben-Wu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1591207

The list of os' is stored in a udf so that it can be accessed from other udfs and a query.  It would be ideal to keep the list as a view so it can be part of the publish views but udfs can't access tables and views, and views can't use udfs.  Updates are probably infrequent but if needed we could schedule the query.

We may want to do something similar for search engines so we should come up with a solution we're happy with.

### DIFF
--- a/sql/static/os_versions/query.sql
+++ b/sql/static/os_versions/query.sql
@@ -1,0 +1,69 @@
+CREATE TEMP FUNCTION udf_get_normalized_os_list() AS (
+  ARRAY<STRUCT<name STRING, version STRING, display_family STRING, display_version STRING, release_date DATE>>[
+    ('Windows_NT', '10.0', 'Windows', '10', '2015-07-15'),
+    ('Windows_NT', '6.3', 'Windows', '8.1', '2013-08-27'),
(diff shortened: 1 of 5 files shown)


jklukas:
> The list of os' is stored in a udf so that it can be accessed from other udfs and a query. It would be ideal to keep the list as a view so it can be part of the publish views but udfs can't access tables and views, and views can't use udfs. Updates are probably infrequent but if needed we could schedule the query.

Using persistent UDFs in views is now GA, so there are some options here.

Fundamentally, though, I think a lookup table should be a lookup table rather than a UDF that returns an array, so I'm -1 on the approach here. It seems more natural to do joins like we do for country codes. Perhaps we should extend the bigquery-etl machinery for publishing views to also deploy static tables.

Ben-Wu:
I did originally want to do something to deploy static tables but I think the use case makes a udf easier to use because there are various special cases.  It could just be a matter of coming up with a table structure that works well or making a udf that doesn't need the whole list.  I can try adding something to deploy csv's as static tables

PR #510, opened 2019-11-18
Bug 1574490 - Add views for events.

mreid-moz:
These views are generated based on [the json configuration files from telemetry-streaming](https://github.com/mozilla/telemetry-streaming/tree/master/configs).

I used [this script](https://gist.github.com/mreid-moz/33837c4c511b4fb7923f67a91232beb7) to generate them.

PR #580, opened 2019-12-09
Bug 1574490 - Add Amplitude-compatible views for mobile events

akkomar:
This adds mobile views introduced in https://github.com/mozilla/bigquery-etl/pull/510


akkomar:
Yes, these are the tested views.

PR #582, opened 2019-12-09
Add script for publishing csv's as static tables

Ben-Wu:
This generalizes the bq load that country codes was using so we can add other static data tables as csv's (e.g. normalized os and search engines) and publish regularly like views.


script/publish_static
```diff
@@ -0,0 +1,100 @@
...
+
+def load_table(
+    data_file_path, schema_file_path=None, description_file_path=None, project=None
+):
+    client = bigquery.Client()
+
+    # Assume path is .../dataset/table/data.csv
+    path_split = os.path.normcase(data_file_path).split("/")
+    dataset_id = path_split[-3]
+    table_id = path_split[-2]
```
Ben-Wu: I acknowledge that the script is tightly coupled with the directory structure and this will cause problems if the structure changes in the future or if the `data.csv` is in the wrong place.  I haven't thought of a nice solution to this yet.

jklukas: This seems reasonable for now.

script/publish_static
```diff
@@ -0,0 +1,100 @@
...
+            table = client.get_table(table_ref)
+            table.description = description
+            client.update_table(table, ["description"])
+
+
+def main():
+    args = parse_args()
+
+    for root, dirs, files in os.walk(args.data_dir):
+        for file in files:
```
jklukas: `file` is a built-in in python, so best to avoid as a variable name. I'd recommend `filename` here.

templates/static/country_names_alternate.csv
```diff
@@ -1,38 +1,38 @@
...
-"Bonaire, Sint Eustatius, and Saba",BQ
-British Virgin Islands,VG
-Federated States of Micronesia,FM
-Saint Martin,MF
-São Tomé and Príncipe,ST
-Vatican City,VA
-Falkland Islands,FK
-North Korea,KP
-Cocos [Keeling] Islands,CC
-Saint Helena,SH
```
jklukas: Why does this show up as deletions? It looks like the file contents are identical save for the header line.

Ben-Wu: I have no idea.  My local history only shows a change to the first line

jklukas: 🤷‍♂ I can live with it.

PR #583, opened 2019-12-10
Add authorized views for payload_bytes raw and error

jklukas:
Supports https://github.com/mozilla/bigquery-etl/issues/360

Adds a new `--authorize` option which the schema deploy pipeline will want to enable to opt-in to publishing authorized views.


script/publish_views
```diff
@@ -30,6 +32,25 @@ def process_file(client, args, filepath):
...
+            if filepath in VIEWS_TO_AUTHORIZE:
+                view = client.get_table(target_view.replace("`", ""))
+                source_datasets = set(
+                    ".".join([t.project, t.dataset_id])
+                    for t in query_job.referenced_tables
+                )
+                for d in source_datasets:
+                    source_dataset = client.get_dataset(d)
+                    access_entries = source_dataset.access_entries
+                    access_entries.append(
```
whd: Should we check if the view is already authorized here?

script/publish_views
```diff
@@ -9,6 +9,8 @@ import os
 from google.cloud import bigquery
 import sqlparse
 
+VIEWS_TO_AUTHORIZE = ("sql/monitoring/payload_bytes_error_structured/view.sql",)
```
acmiyaguchi: Having `payload_bytes_error_telemetry` at the get-go would be nice too.

acmiyaguchi:
> We'll need to change the source for the existing views to `monitoring.payload_*` along with this change

The existing views won't break, but they will need to be updated so we can shared them through STMO.

script/publish_views
jklukas: That was an oversight. Adding all 4 of these new views to the list.

PR #586, opened 2019-12-10
Revert "Add authorized views for payload_bytes raw and error"

jklukas:
Reverts mozilla/bigquery-etl#583

I merged before getting @whd's feedback. Sounds like we need to discuss a bit more.


acmiyaguchi:
From what I understand, the terraform configuration will effectively de-authorize these views, so the PR in itself shouldn't be problematic other than not behaving as expected.

jklukas:
The important thing is that we modified some of the monitoring views to hit these new views, so those views will break on next schema deploy if we do nothing.

PR #588, opened 2019-12-11
Break up histogram processing and update probe counts view

emtwo:
This PR includes 2 changes:

1. The histogram probe counts query has been broken up into two pieces as I've done with the scalar probe counts. The query was starting to run long on airflow and this change gives significant performance wins based on my testing so far.

2. I've updated the probe counts view to include percentiles. This is beneficial to the Glam app server so it only needs to query one place for all the data instead of 3.

The associated airflow PR is here: https://github.com/mozilla/telemetry-airflow/pull/799

### DIFF
--- a/sql/telemetry/client_probe_counts_v1/view.sql
+++ b/sql/telemetry/client_probe_counts_v1/view.sql
@@ -1,9 +1,29 @@
 CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.telemetry.client_probe_counts`
-AS SELECT
-  *
-FROM
-  `moz-fx-data-shared-prod.telemetry_derived.client_probe_counts_v1`
+AS
+
+WITH all_counts AS (
+  SELECT
+    *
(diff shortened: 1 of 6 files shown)


sql/telemetry/client_probe_counts_v1/view.sql
```diff
@@ -1,9 +1,29 @@
 CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.telemetry.client_probe_counts`
-AS SELECT
-  *
-FROM
-  `moz-fx-data-shared-prod.telemetry_derived.client_probe_counts_v1`
+AS
+
+WITH all_counts AS (
```
fbertsch: +1, this seems sensible

emtwo:
> LGTM. Did you change `bucket_counts` when moving to the separate query? It's hard to tell from this diff.

Nope, no changes. It's just no longer a CTE

PR #579, opened 2019-12-06
Add draft query for attitudes daily

benmiroglio:
Adding a draft query here to get `attitudes_daily` into big query.

@sunahsuh r?

Some notes:
* [Here](https://bugzilla.mozilla.org/show_bug.cgi?id=1583947) is the tracking bug.
* I created a prototype of this in BQ: https://console.cloud.google.com/bigquery?sq=630180991450:6ac5dfcf88ad48eab9c7e56a2aaf9b29 which uses a temporarily test table analysis.survey_gizmo_test_data that was uploaded from a csv after being generated in this notebook https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/234967/command/234975
* We need a permanent table in lieu of the prototype table `analysis.survey_gizmo_test_data`
* This follows the convention of using the `@submission_date` parameter I've seen in other queries.
* `$ script/format-sql templates/telemetry_derived/attitudes_daily_v1/query.sql` threw the following error: 

```
/Users/bmiroglio/src/bigquery-etl/script/format-sql:126
 const modified = args.flatMap((arg) => {
                        ^

TypeError: args.flatMap is not a function
    at Object.<anonymous> (/Users/bmiroglio/src/bigquery-etl/script/format-sql:126:25)
    at Module._compile (module.js:649:30)
    at Object.Module._extensions..js (module.js:660:10)
    at Module.load (module.js:561:32)
    at tryModuleLoad (module.js:501:12)
    at Function.Module._load (module.js:493:3)
    at Function.Module.runMain (module.js:690:10)
    at startup (bootstrap_node.js:194:16)
    at bootstrap_node.js:666:3
```

PR #589, opened 2019-12-12
Bug 1583947: Add surveygizmo import job for attitudes daily

sunahsuh:
Based on the notebook here: https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/149025/command/178484

Downloads a day's data from SurveyGizmo and imports the data into BigQuery. When I tested this, one day generated ~2500 rows.

The imported data is then used in #579 to correlate responses to telemetry from that client that day.

@benmiroglio FYI I found a bug re: the previous script and timezones – SurveyGizmo expects and returns all times in Eastern, which doesn't match up with UTC which we use in `clients_daily` and everywhere else. Lmk if I shouldn't be correcting for that here. (And please review the script overall.)


templates/external/surveygizmo_daily_attitudes/import_responses.py
```diff
@@ -0,0 +1,119 @@
...
+  return dt.datetime.strftime(dt.date.fromisoformat(date_string) + dt.timedelta(days=1), '%Y-%m-%d')
+
+
+def format_response (s, date):
+    # Takes a single user's responses and returns a list of dictionaries, one per answer, formatted to corresponding bigquery fields.
+    # `survey_data` is a dict with question ID as the key and question/response details as the value
+    # e.g. survey_data: {'25': {'id': 25, 'type': 'RADIO', 'question': 'I trust Firefox to help me with my online privacy',
+    # 'section_id': 2, 'answer': 'Agree', 'answer_id': 10066, 'shown': True}}
+    resp = list(s.get("survey_data", {}).values())
+    ret = []
```
sunahsuh:
```suggestion
```

templates/external/surveygizmo_daily_attitudes/import_responses.py
```diff
@@ -0,0 +1,118 @@
...
+  "I trust Firefox to help me with my online privacy": "trust_firefox",
+  "All the sites I’ve visited recently have worked; none of them seem broken": "sites_work",
+  "The internet is open and accessible to all": "internet_accessible",
+  "Using the internet helped me meet my goals today": "met_goals"
+}
+
+
+def utc_date_to_eastern_string(date_string):
+  # Takes in a YYYY-MM-DD date string and returns the equivalent in eastern time
+  # of midnight UTC of that date
```
jklukas: Might as well make these docstrings rather than comments to be more pythonic.

jklukas: Might as well make these docstrings rather than comments to be more pythonic.

templates/external/surveygizmo_daily_attitudes/import_responses.py
```diff
@@ -0,0 +1,118 @@
...
+        resp = requests.get(url + f"&page={page}")
+        resp.raise_for_status()
+        ret = ret + construct_data(resp.json(), date_string)
+    return ret
+
+
+def insert_to_bq(data, table):
+    client = bigquery.Client()
+    print(f"Inserting {len(data)} rows into bigquery")
+    job = client.load_table_from_json(data, table)
```
jklukas: We should make this idempotent. To do so, we should specify the particular date partition as the destination, so something like `{table}${date.replace('-', '')}` to get something like `moz-fx-data-shared-prod.external.survey_gizmo_daily_attitudes$20191201`. And then we need to provide a LoadJobConfig that sets the write_disposition to overwrite.

You should be able to copy almost verbatim from https://github.com/mozilla/forecasting/blob/master/simpleprophet/simpleprophet/output.py#L97-L107 and https://github.com/mozilla/forecasting/blob/master/simpleprophet/simpleprophet/pipeline.py#L56-L57

sql/external/surveygizmo_daily_attitudes/init.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE IF NOT EXISTS
+  `moz-fx-data-shared-prod.external.survey_gizmo_daily_attitudes` (
```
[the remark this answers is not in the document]
sunahsuh: No, doesn't exist, and requires someone from Ops, I think, to take it on – how about `third_party` instead?

templates/external/surveygizmo_daily_attitudes/import_responses.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
sunahsuh: Both relud and I have scripts running from the `template` directory these days – for `experiment_enrollment_aggregates_live` and for `experiments_v1`. I'm fine with moving them but if so we should reorganize the scripts directory a bit, probably

templates/external/surveygizmo_daily_attitudes/import_responses.py
```diff
@@ -0,0 +1,118 @@
...
+import re
+import requests
+from time import sleep
+
+
+parser = ArgumentParser(description=__doc__)
+parser.add_argument("--date", required=True)
+parser.add_argument("--survey_id", required=True)
+parser.add_argument("--sg_api_token", required=True)
+parser.add_argument("--sg_api_secret", required=True)
```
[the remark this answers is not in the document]
sunahsuh: Yes, but we use this pattern for, say, all the leanplum jobs, unfortunately, so I've chalked this up as a losing battle. It's also in plaintext in the databricks notebook. I gave some thought to subclassing the GKEPodOperator with one that takes in an extra arg that's a list of airflow variables that can go in env variables, and that could be a way forward if there's impetus for changing how this works.

templates/external/surveygizmo_daily_attitudes/import_responses.py
```diff
@@ -0,0 +1,118 @@
...
+        # Shield ID is sent as a hidden field with question name "Shield ID"
+        shield_id = [r.get("answer") for r in resp if r.get('question') == 'Shield ID'][0]
+    except IndexError:
+        shield_id = None
+
+    return [{
+        'shield_id': shield_id,
+        'date': date,
+        'question': r.get('question'),
+        'question_key': q_fields.get(r.get('question')),
```
sunahsuh:
```suggestion
        'question_key': Q_FIELDS.get(r.get('question')),
```

sql/external/surveygizmo_daily_attitudes/init.sql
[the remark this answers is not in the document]
jklukas: We could put it in `telemetry_derived`, which is already home to various non-telemetry data, and delay the decision on a more proper home.`third_party` seems like a reasonable choice if we provision a new dataset, or perhaps `imported`.

templates/external/surveygizmo_daily_attitudes/import_responses.py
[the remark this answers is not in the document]
jklukas: If there's existing precedent, then let's keep the script here for the time being and delay reorganizing to a separate issue.

templates/external/surveygizmo_daily_attitudes/import_responses.py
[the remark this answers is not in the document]
jklukas: Thanks for the context. Not worth blocking on this.

sunahsuh:
@jklukas I still need a green check to merge this in – the CI failure is from the unrelated search_clients_daily issue

templates/telemetry_derived/surveygizmo_daily_attitudes/import_responses.py
```diff
@@ -0,0 +1,121 @@
...
+
+    for page in range(2, total_pages+1):
+        print("fetching page {}".format(page))
+        resp = requests.get(url + f"&page={page}")
+        resp.raise_for_status()
+        ret = ret + construct_data(resp.json(), date_string)
+    return ret
+
+
+def insert_to_bq(data, table, date, write_disposition=bigquery.job.WriteDisposition.WRITE_APPEND):
```
jklukas:
```suggestion
def insert_to_bq(data, table, date, write_disposition=bigquery.job.WriteDisposition.WRITE_TRUNCATE):
```

Given we do the manipulation within this function to specify append a partition decorator to the table name, WRITE_TRUNCATE seems like what we would almost always want here.

sunahsuh: Ah thanks for catching that!

PR #590, opened 2019-12-12
Bug 1591207 - Add normalized os static table

Ben-Wu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1591207

Enables normalizing os versions.  Usage looks like this:
```sql
...
FROM
  `moz-fx-data-shared-prod.telemetry_stable.main_v4`
LEFT JOIN
  `static.normalized_os_version` norm_os
ON
  (environment.system.os.name = norm_os.os_name
  AND REGEXP_CONTAINS(environment.system.os.version, norm_os.os_version))
```
A bit awkward because of the `REGEXP_CONTAINS` but should be fine as a first pass.


templates/static/normalized_os_version/data.csv
```diff
@@ -0,0 +1,51 @@
+os_name,os_version,normalized_os_name,normalized_os_version,release_date
```
jklukas:
```suggestion
os_name,os_version_pattern,normalized_os_name,normalized_os_version,release_date
```

The fact that the os_version is a regex seems important and potentially unexpected. Somehow naming the column such that it's more obviously a regex could help discoverability. `os_version_regexp` would be another reasonable choice, especially since that matches REGEXP_CONTAINS.

Ben-Wu: Yes `os_version_regexp` makes the most sense to me

PR #596, opened 2019-12-13
Use normalize engine udf in all search queries

Ben-Wu:
Use `udf_strict_normalize_search_engine` in all the search queries.  Spoke with bmiroglio again and we probably don't actually need the non-strict function.  I think he misunderstood and thought the normalized column would replace the normal engine.  Should we keep the non-strict option or just use one to avoid confusion?

@fbertsch The normalized values are different from the ones previously in search clients last seen so you may want to backfill.


templates/search_derived/search_clients_last_seen_v1/query.sql
```diff
@@ -1,14 +1,7 @@
...
     SELECT
-      CASE -- we only care about certain normalized engines
-          WHEN engine IS NULL THEN 'none'
-          WHEN STARTS_WITH(engine, 'google') THEN 'google'
-          WHEN STARTS_WITH(engine, 'ddg') OR STARTS_WITH(engine, 'duckduckgo') THEN 'ddg'
-          WHEN STARTS_WITH(engine, 'bing') THEN 'bing'
-          WHEN STARTS_WITH(engine, 'yandex') THEN 'yandex'
-          ELSE 'other'
-      END AS short_engine,
+      udf_strict_normalize_search_engine(engine) AS short_engine,
```
Ben-Wu: We can get normalized directly from scd if we backfill it but this should do for now.

sql/search_derived/mobile_search_clients_daily_v1/query.sql
```diff
@@ -181,5 +206,6 @@ GROUP BY
   submission_date,
   client_id,
   engine,
+  normalized_engine,
```
fbertsch: Surely we don't need to group by both engine and normalized_engine? How does Ben feel about grouping by just the latter?

Ben-Wu: We need to group by `engine` for granularity but for `normalized_engine` engine I don't think it would matter since each engine would have one normalized engine value.  I just found grouping by `normalized_engine` to be the simplest way to do it since otherwise we need to aggregate it or make this a CTE and write another small select statement.

e.g.
```sql
SELECT *, udf_strict_normalize_search_engine(engine) AS normalized_engine
FROM scd
```

Although after writing this out I actually prefer this way because it removes the repeated logic of filtering out high search count and getting the engine from the normalized search key.

PR #603, opened 2019-12-16
Bug 1598002 - add submission_timestamp_min to clients_daily

relud:
when this is merged, the column will need to be added to `clients_daily` and `clients_last_seen` to prevent errors in airflow


relud:
CI failure is unrelated

relud:
> when this is merged, the column will need to be added to `clients_daily` and `clients_last_seen` to prevent errors in airflow

columns added, downstream views confirmed to be working

PR #604, opened 2019-12-16
Remove non-strict search engine normalization

Ben-Wu:
and remove strict from function name to avoid confusion on which one to use.  Probably should have done this in the previous PR.  bmiroglio agreed this should be done


Ben-Wu:
dry run failure is happening on master too.  I'm still trying to find a fix

PR #606, opened 2019-12-17
Point search clients daily to main summary table (not view)

Ben-Wu:
Main summary view became "too complex".  This is likely a short term solution since we can't backfill far with this.


relud:
ah, thanks for finding the issue with this. the complexity was caused by a combination of #601 and #602

relud:
> This is likely a short term solution since we can't backfill far with this

for CCPA compliance work the prod main summary table will be backfilled, so it shouldn't be a problem long term

PR #610, opened 2019-12-18
Bug: publish_views ignored some view.sql files

### LINKED ISSUE #600: Update experiments view
It looks like the underlying table transitioned to shared-prod, but the view wasn’t updated to do a union. That should be fixed.

jklukas:
Closes #600

We improve the logic for parsing view definition files, and we also now error out when we encounter a view.sql file that doesn't match our parsing rather than silently skipping.


jklukas:
We also normalize some of the view definition files as part of this PR, although the improved parsing logic is now more robust to be able to handle these variants.

PR #611, opened 2019-12-18
RFM View for LTV

fbertsch:
Frequency: Number of days of said activity during the period
Recency: Number of days between first and last activity
T: Number of days since first instance of said activity

### DIFF
--- a/templates/search/search_rfm/view.sql
+++ b/templates/search/search_rfm/view.sql
@@ -1,37 +1,12 @@
 CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.search.search_rfm` AS
 SELECT
-  STRUCT(
-    udf.bits_to_days_seen_int(days_seen_bytes) AS frequency,
-    udf.bits_to_first_seen_int(days_seen_bytes) AS T,
-    udf.bits_to_first_seen_int(days_seen_bytes)
-        - udf.bits_to_last_seen_int(days_seen_bytes) AS recency
-  ) AS days_seen,
-  STRUCT(
-    udf.bits_to_days_seen_int(days_searched_bytes) AS frequency,
-    udf.bits_to_first_seen_int(days_searched_bytes) AS T,
-    udf.bits_to_first_seen_int(days_searched_bytes)
-        - udf.bits_to_last_seen_int(days_searched_bytes) AS recency
-  ) AS days_searched,
-  STRUCT(
-    udf.bits_to_days_seen_int(days_tagged_searched_bytes) AS frequency,
-    udf.bits_to_first_seen_int(days_tagged_searched_bytes) AS T,
-    udf.bits_to_first_seen_int(days_tagged_searched_bytes)
-        - udf.bits_to_last_seen_int(days_tagged_searched_bytes) AS recency
-  ) AS days_tagged_searched,
-  STRUCT(
-    udf.bits_to_days_seen_int(days_searched_with_ads_bytes) AS frequency,
-    udf.bits_to_first_seen_int(days_searched_with_ads_bytes) AS T,
-    udf.bits_to_first_seen_int(days_searched_with_ads_bytes)
-        - udf.bits_to_last_seen_int(days_searched_with_ads_bytes) AS recency
-  ) AS days_searched_with_ads,
-  STRUCT(
-    udf.bits_to_days_seen_int(days_clicked_ads_bytes) AS frequency,
-    udf.bits_to_first_seen_int(days_clicked_ads_bytes) AS T,
-    udf.bits_to_first_seen_int(days_clicked_ads_bytes)
-        - udf.bits_to_last_seen_int(days_clicked_ads_bytes) AS recency
-  ) AS days_clicked_ads,
-  udf.bits_to_first_seen_int(days_created_profile_bytes) AS days_since_created_profile,
+  `moz-fx-data-shared-prod`.udf.days_seen_bytes_to_rfm(days_seen_bytes) AS days_seen,
+  `moz-fx-data-shared-prod`.udf.days_seen_bytes_to_rfm(days_searched_bytes) AS days_searched,
+  `moz-fx-data-shared-prod`.udf.days_seen_bytes_to_rfm(days_tagged_searched_bytes) AS days_tagged_searched,
+  `moz-fx-data-shared-prod`.udf.days_seen_bytes_to_rfm(days_searched_with_ads_bytes) AS days_searched_with_ads,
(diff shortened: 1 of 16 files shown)


fbertsch:
Need to rename the UDF in the smoot queries.

jklukas:
Can you define RFM and LTV, or point me to some documentation for more context?

fbertsch:
@jklukas sure, some docs are here: https://lifetimes.readthedocs.io/en/latest/Quickstart.html#the-shape-of-your-data

LTV = Lifetime Value, a project we're working on
RFM = recency, frequency, monetary

udf/pos_of_trailing_set_bit.sql
```diff
@@ -10,12 +10,12 @@ see https://stackoverflow.com/a/42747608/1260237
 */
 
 CREATE TEMP FUNCTION
-  udf_bitpos( bits INT64 ) AS ( CAST(SAFE.LOG(bits & -bits, 2) AS INT64));
+  udf_pos_of_trailing_set_bit( bits INT64 ) AS ( CAST(SAFE.LOG(bits & -bits, 2) AS INT64));
```
jklukas: For naming, I think I chose `bitpos` originally in analogy to BigQuery's STRPOS function, although the direction and the indexing are different.

`pos_of_trailing_set_bit` is a comprehensive name, but I would like it to be shorter if possible. What do you think of `bitpos_rightmost` or `bitpos_trailing`? Or does `bitpos` seem to obtuse in the first place?

udf/bits_to_last_seen_int.sql
```diff
@@ -0,0 +1,41 @@
+/*
+Given a BYTE, return the number of days since the client was
```
jklukas:
```suggestion
Given a single-entry BYTES array, return the number of days since the client was
```

Is the above what you mean by a BYTE?

udf/bits_to_last_seen_int.sql
```diff
@@ -0,0 +1,41 @@
...
+Otherwise the results are 0-indexed, meaning \x01 will return 0.
+
+Tests showed this being 5-10x faster than the simpler alternative:
+CREATE TEMP FUNCTION
+  udf_bits_to_last_seen_int(b BYTES) AS ((
+    SELECT MIN(n)
+    FROM UNNEST(GENERATE_ARRAY(0, 364)) AS n
+    WHERE BIT_COUNT(SUBSTR(b >> n, -1) & b'\x01') > 0));
+
+See also: bits_to_last_seen_int.sql
```
jklukas:
```suggestion
See also: bits_to_first_seen_int.sql
```

udf/pos_of_trailing_set_bit.sql
fbertsch: Analogy to `STRPOS` I understand, but imho `bitpos` is confusing to see, since it doesn't clarify that it's finding set bits.

I _believe_ this will be a largely internally-used function, so I'm okay with a larger, comprehensive name. If you see a need for it outside of these few views/UDFs then we should shorten it.

udf/bits_to_last_seen_int.sql
```diff
@@ -0,0 +1,41 @@
...
+*/
+
+CREATE TEMP FUNCTION
+  udf_bits_to_last_seen_int(b BYTES) AS ((
```
jklukas: How about `udf_bits_to_days_since_seen`? That would feel more consistent to me with existing terminology. We have days_seen_bits which are used to calculate days_since_X columns in clients_last_seen.

`last_seen_int` sounds to me like we're returning the integer which was last seen, rather than returning an integer representing the last seen occurrence of something.

udf/bits_to_last_seen_int.sql
```diff
@@ -0,0 +1,41 @@
...
+
+If no bits are set, returns NULL, indicating we don't know.
+Otherwise the results are 0-indexed, meaning \x01 will return 0.
+
+Tests showed this being 5-10x faster than the simpler alternative:
+CREATE TEMP FUNCTION
+  udf_bits_to_last_seen_int(b BYTES) AS ((
+    SELECT MIN(n)
+    FROM UNNEST(GENERATE_ARRAY(0, 364)) AS n
+    WHERE BIT_COUNT(SUBSTR(b >> n, -1) & b'\x01') > 0));
```
jklukas: I really appreciate the documentation here of alternate approaches and reasoning behind the implementation.

udf/bits_to_days_seen_int.sql
```diff
@@ -0,0 +1,9 @@
+/*
+Given a BYTE, get the number of days the user was seen.
+
+NULL input returns NULL output.
+*/
+
+CREATE TEMP FUNCTION
+  udf_bits_to_days_seen_int(b BYTES) AS (
```
jklukas: I would leave off the `_int` suffix as it seems unnecessary.

templates/search/search_rfm/view.sql
```diff
@@ -0,0 +1,45 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.search.search_rfm` AS
+SELECT
+  STRUCT(
+    udf.bits_to_days_seen_int(days_seen_bytes) AS frequency,
+    udf.bits_to_first_seen_int(days_seen_bytes) AS T,
+    udf.bits_to_first_seen_int(days_seen_bytes)
+        - udf.bits_to_last_seen_int(days_seen_bytes) AS recency
```
jklukas: Should we factor out this logic to a UDF like `udf.days_seen_bytes_to_rfm(days_seen_bytes)`?

udf/bits_to_last_seen_int.sql
fbertsch: Agreed, it is confusing. How about `udf_bits_to_days_since_last_seen` and `udf_bits_to_days_since_first_seen`

jklukas: I would argue that the standard we have is `days_since_<action>` and that `seen` is the default action that means most recent occurrence rather than `last_seen`.

So `udf_bits_to_days_since_first_seen` sounds good to me, but I'd prefer `udf_bits_to_days_since_seen` over `udf_bits_to_days_since_last_seen`

udf/bits_to_last_seen_int.sql
fbertsch: No, I do mean `BYTES` though!

fbertsch: I'm specifically leaving out the word `array` since that may confused people, since arrays are a different data structure.

PR #612, opened 2019-12-18
Fix monthly_searches to account for null

fbertsch:
Now I'm remembering why I used `"none"` instead of `null`! A full outer join with two `NULL` keys doesn't match and returns two separate rows.


fbertsch:
Somethings going wrong with the `generate_sql` script. I'll debug later tonight

udf/add_monthly_searches.sql
```diff
@@ -48,7 +49,10 @@ WITH previous_examples AS (
                    udf_zeroed_array(12) AS ad_click) AS value
         )] AS prev, "google" as type
     UNION ALL
-    SELECT NULL AS prev, "null" as type
+    SELECT [ STRUCT(
+            CAST(NULL AS STRING) AS key,
+            udf_new_monthly_engine_searches_struct() AS value
+        ) ] AS prev, "null" as type
```
Ben-Wu: Actually looking back this looks kind of messy.
I think this is easier to read:
```sql
 SELECT 
  [STRUCT(
    CAST(NULL AS STRING) AS key,
    udf_new_monthly_engine_searches_struct() AS value
  )] AS prev, 
  "null" as type
```

PR #614, opened 2019-12-18
Add attitudes_daily_v1 init statement

sunahsuh:
I forgot to add an `init.sql` statement for this table, which is clients_daily plus some stuff from surveygizmo. Yesterday's run produced ~2000 rows so it's not much of a duplication of clients_daily, and the query takes long enough that I think it makes sense to materialize (also I think they want to use this for a GUD graph?)


sunahsuh:
Redirecting review to @jklukas since it's work hours for him now and he has more context

PR #620, opened 2019-12-19
Add simpleprophet_forecasts view

jklukas:
cc @jmccrosky


jklukas:
Tables already exist and the queries to populate them are already scheduled, so this is good to go.

relud:
> Tables already exist and the queries to populate them are already scheduled, so this is good to go.

whoops, i typoed a command and thought they weren't there. sweet.

PR #625, opened 2019-12-20
Include all Rocket events in Amplitude view

akkomar:


templates/telemetry/rocket_android_events_v1/view.sql
```diff
@@ -233,47 +234,7 @@ WHERE
...
-      UNION ALL SELECT 'show_keyboard' AS key, `moz-fx-data-derived-datasets.udf.get_key`(event_map_values, 'show_keyboard') AS value
-      UNION ALL SELECT 'url_counts' AS key, `moz-fx-data-derived-datasets.udf.get_key`(event_map_values, 'url_counts') AS value
-      UNION ALL SELECT 'impression' AS key, `moz-fx-data-derived-datasets.udf.get_key`(event_map_values, 'impression') AS value
-      UNION ALL SELECT 'loadtime' AS key, `moz-fx-data-derived-datasets.udf.get_key`(event_map_values, 'loadtime') AS value
-      UNION ALL SELECT 'audience_name' AS key, `moz-fx-data-derived-datasets.udf.get_key`(event_map_values, 'audience_name') AS value
-      UNION ALL SELECT 'finished' AS key, `moz-fx-data-derived-datasets.udf.get_key`(event_map_values, 'finished') AS value
-      UNION ALL SELECT 'task' AS key, `moz-fx-data-derived-datasets.udf.get_key`(event_map_values, 'task') AS value
-      UNION ALL SELECT 'item_name' AS key, `moz-fx-data-derived-datasets.udf.get_key`(event_map_values, 'item_name') AS value
+      SELECT 'event_value' AS key, event_value AS value
```
akkomar: * generic `event_value` is added here for new events
* event properties that were explicitly extracted here are removed because they're all automatically pulld into `event_props_1` couple of lines above

cnevinc:
@akkomar  thank you so much for this PR!
I'll need the BI team in Taipei to confirm this before it's merged.
Thanks again!

templates/telemetry/rocket_android_events_v1/view.sql
cnevinc: I don't understand this. Can you elaborate more?

cnevinc:
for the user attribute, how can we do the mapping?

templates/telemetry/rocket_android_events_v1/view.sql
akkomar: >     * generic `event_value` is added here for new events

Previously `event_value` was used for distinguishing and naming events (big `all_events` section above). We are keeping this for old events. But we'll also pull `event_value` to a parameter which is for new event types, in order to enable filtering in Amplitude.

>     * event properties that were explicitly extracted here are removed because they're all automatically pulled into `event_props_1` couple of lines above

This is implementation detail - since whole `event_map_values` map is inserted into event_parameters, these explicit extract statements (e.g. `UNION ALL SELECT 'item_name' AS key, `moz-fx-data-derived-datasets.udf.get_key`(event_map_values, 'item_name') AS value`) are not necessary.

akkomar:
> for the user attribute, how can we do the mapping?

User properties are assembled here: https://github.com/mozilla/bigquery-etl/blob/4e53197f0899fc074199f32bd4b6046bb7b9a7af/sql/telemetry/rocket_android_events_v1/view.sql#L240-L257 (and we can add new, custom ones there).

akkomar:
Closing this - per discussion with @cnevinc we prefer to explicitly select events for this export.

PR #635, opened 2019-12-27
Add script for self serve deletion

relud:


bigquery_etl/shredder/config.py
```diff
@@ -0,0 +1,239 @@
...
+        """Fully qualified table id in standard sql format."""
+        return f"{self.project}.{self.table}"
+
+
+@dataclass(frozen=True)
+class DeleteTarget:
+    """Data class for deletion request target."""
+
+    table: str
+    id_: str
```
jklukas: Why is this suffixed with an underscore? Is the intention that it will never be read by the user and is thus considered private?

bigquery_etl/shredder/config.py
```diff
@@ -0,0 +1,239 @@
...
+    client_id_target(
+        table="telemetry_stable.main_v4",
+        cluster_conditions=tuple(
+            (condition, needs_clustering)
+            for condition, needs_clustering in chain(
+                {
+                    f"sample_id = {sample_id} AND normalized_channel = 'release'": False
+                    for sample_id in range(100)
+                }.items(),
+                [("normalized_channel != 'release'", True)],
```
jklukas: We may want to be careful about whether `normalized_channel` is ever `NULL`, which IIRC, won't be caught by any of these cases. Even if `normalized_channel` _should_ never be null, it probably will be at some point.

bigquery_etl/shredder/config.py
```diff
@@ -0,0 +1,239 @@
...
+        return f"{self.project}.{self.table}"
+
+
+@dataclass(frozen=True)
+class DeleteTarget:
+    """Data class for deletion request target."""
+
+    table: str
+    id_: str
+    cluster_conditions: Optional[Tuple[Tuple[str, bool]]] = None
```
jklukas: Could you document either in a comment here or in the docstring for the class what this field does? I think I'm inferring the purpose from how it's used, but it's not obvious.

Or maybe it would be good to have a separate class for the main ping clustering case. Not sure whether inheritance is possible with data classes.

bigquery_etl/shredder/config.py
```diff
@@ -0,0 +1,239 @@
...
+    client_id_target(table="telemetry_stable.update_v4"): CLIENT_ID_SRC,
+    client_id_target(table="telemetry_stable.voice_v4"): CLIENT_ID_SRC,
+    # activity stream
+    impression_id_target(
+        table="activity_stream_stable.impression_stats_v1"
+    ): IMPRESSION_ID_SRC,
+    impression_id_target(
+        table="activity_stream_stable.spoc_fills_v1"
+    ): IMPRESSION_ID_SRC,
+    # TODO check if this table's impression_id is from activity stream
```
jklukas: It is.

bigquery_etl/shredder/config.py
```diff
@@ -0,0 +1,239 @@
...
+    impression_id_target(
+        table="messaging_system_stable.undesired_events_v1"
+    ): IMPRESSION_ID_SRC,
+}
+
+UNSUPPORTED = [
```
jklukas: Why are these unsupported? Is this a temporary list while details get sorted out?

bigquery_etl/shredder/delete.py
```diff
@@ -0,0 +1,337 @@
...
+FROM
+  `{{target.sql_table_id}}`
+{WHERE_CLAUSE.strip()}
+  AND {{cluster_condition}}
+"""
+
+
+async def gather_all(futures):
+    """Gather results, raise last exception, and log other exceptions."""
+    results = await asyncio.gather(*futures, return_exceptions=True)
```
jklukas: Note that I am only passingly familiar with asyncio in python, so not going to be able to provide much validation of the task handling here.

bigquery_etl/shredder/delete.py
```diff
@@ -0,0 +1,337 @@
...
+        for query in queries
+    )
+    # copy results into place if necessary
+    if target.cluster_conditions:
+        sources = [
+            f"{t.project}.{t.dataset_id}.{t.table_id}"
+            for t in [job.destination for job in jobs]
+        ]
+        dest = f"{target.sql_table_id}${partition_id}"
+        # update clustering fields if necessary before copying into place
```
jklukas: This could use more explanation in the code. Am I correct in understanding that this is a performance optimization where we create temp tables without clustering in order to make writes faster, but then have to update the clustering definition in order to satisfy the BQ API's requirement that source and destination tables have the same clustering definition when we `bq cp`?

If that's true, then I worry that our result is going to be essentially unclustered.

bigquery_etl/util/table_filter.py
```diff
@@ -0,0 +1,52 @@
+"""TODO."""
```
jklukas: Make sure you address TODO items.

bigquery_etl/util/temp_table.py
```diff
@@ -0,0 +1,43 @@
+"""Generate temporary tables."""
```
jklukas: It's really nice to have some of these items that existed in previous scripts factored out into more maintainable and better-documented modules.

bigquery_etl/shredder/delete.py
relud: > Am I correct in understanding that this is a performance optimization where we create temp tables without clustering in order to make writes faster, but then have to update the clustering definition in order to satisfy the BQ API's requirement that source and destination tables have the same clustering definition when we `bq cp`?

yes.

> If that's true, then I worry that our result is going to be essentially unclustered.

I tested it, and the result is effectively clustered.

relud:
ci failure is due to #705 and unrelated to this PR

PR #637, opened 2019-12-30
Bug 1606171 - Add is_default_browser to search clients daily and search aggregates

wlach:


wlach:
I think this should work? Probably not worth backfilling for this alone, but I imagine something similar will come up in the near future.

wlach:
> LGTM. You need to manually add the `is_default_browser` column to both the tables in bigquery because the schema doesn't auto update. You can do it through the bigquery console.

Interesting, I'm not actually sure how to do this!

> For backfill it should be possible to recreate search clients daily from a copy of itself joining on `is_default_browser` values from `main_summary`. This way we don't need to scan main summary and it would cost hundreds instead of thousands. I can look more into that when I'm back next week because we'll probably want to do that somewhat often.

Sounds good, let's chat about this when you return. I don't think there is any rush on merging this.

wlach:
Update: discussed with @Ben-Wu, we should add this field to search_aggregates as well before landing this. gonna do that

wlach:
@Ben-Wu ready for another look! good catch on the view

wlach:
I'll let you take care of merging this and doing the backfill, if that's ok!

Ben-Wu:
Sure I'll merge this now and hopefully backfill next week if I can get experiments added

PR #639, opened 2020-01-02
Bug 1606687 - Add fxa_configured to clients_daily

relud:


relud:
I checked the table and this is the only field that's present in the schema but missing from the query.

PR #640, opened 2020-01-03
Rewrite script/format_sql in python

relud:


relud:
> Is the primary motivation here to make this tool more accessible, since the rest of the tooling is python-based?

my motivations, in descending order:

1. correctly differentiate operators from angle brackets used in types, which requires a more stateful tokenizer than the js lib provides.
2. provide fine-grained control of formatter behavior so that we can get it working well enough to require it
3. stop requiring node, since the rest of the tooling is python-based

> Does this run reasonably quickly compared to the js lib we were using?

yes

> Are the results identical?

no. This is intentionally modified to:

- correctly identify angle brackets and signed number literals
- better preserve comment location
- inline bracket blocks if they don't exceed the max line length

bigquery_etl/format_sql/formatter.py
```diff
@@ -0,0 +1,189 @@
...
+                    yield from pending
+                    skip_lines += pending_lines
+                    if line.can_start_inline_block:
+                        pending_lines = 0
+                        pending = []
+                    else:
+                        break
+
+
+def reformat(query):
```
acmiyaguchi: If this module were to be put under test, a callback argument for the format type would make this a useful entrypoint.

bigquery_etl/format_sql/formatter.py
```diff
@@ -0,0 +1,189 @@
...
+    def tokens(self):
+        """Get a list of all the tokens in this line."""
+        if self.indent_token is None:
+            return self.inline_tokens
+        else:
+            return [self.indent_token] + self.inline_tokens
+
+
+def inline_block_format(tokens, max_inline_length=100):
+    """Extend simple_format with inline blocks.
```
acmiyaguchi: What is an inline block?

acmiyaguchi:
The output is pretty appealing, here are some cases that I'd like to highlight:

In `templates/monitoring/document_sample_nonprod_v1/query.sql`, the line wrapping behavior is much easier to read than the BigQuery console formatter. 
```diff
-      ARRAY_AGG(
-        extracted
-        ORDER BY
-          RAND()
-        LIMIT
-          1000
-      ) samples
+      ARRAY_AGG(extracted ORDER BY RAND() LIMIT 1000) samples
```

In `templates/org_mozilla_fenix_derived/clients_daily_v1/query.sql` has output I hadn't considered when looking at the newline keywords:
```diff
-    ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING),
+        ROWS BETWEEN
+          UNBOUNDED PRECEDING
+          AND UNBOUNDED FOLLOWING
+      ),
```

In `templates/telemetry/fxa_users_daily_v1/query.sql`, no more `--` hacks to get around the redash formatter:

```diff
-CREATE TEMP FUNCTION
-  udf_contains_tier1_country(x ANY TYPE) AS ( --
-    EXISTS(
+CREATE TEMP FUNCTION udf_contains_tier1_country(
+  x ANY TYPE
+) AS ( --
+  EXISTS(
```

`templates/telemetry_derived/clients_daily_histogram_aggregates_v1/query.sql`
```diff
-    ARRAY<STRUCT<
-      metric STRING,
-      value STRING
-    >>[
+      ARRAY<STRUCT<metric STRING, value STRING>>[
```
---

There are a couple of quirks, but only a few of them break the script as far as I can tell.

In `templates/monitoring/telemetry_missing_columns_v1/view.sql`, a space is added to a regex literal (syntax error).
```diff
-  REGEXP_EXTRACT(_TABLE_SUFFIX, r"^(.*)_v.*") AS document_type,
+  REGEXP_EXTRACT(_TABLE_SUFFIX, r "^(.*)_v.*") AS document_type,
```

In `templates/org_mozilla_fenix/clients_last_seen/view.sql`, a space is added after `-` in a negative literal:
```diff
-    CAST(SAFE.LOG(days_seen_bits & -days_seen_bits, 2) AS INT64) AS days_since_seen,
+      CAST(SAFE.LOG(days_seen_bits & - days_seen_bits, 2) AS INT64) AS days_since_seen,
```

In `templates/search/mobile_search_aggregates/view.sql`, it seems natural to put the create or replace clause on a single line. 
```diff
-CREATE OR REPLACE VIEW
-  `moz-fx-data-shared-prod.search.mobile_search_aggregates`
-AS
+CREATE
+OR REPLACE VIEW `moz-fx-data-shared-prod.search.mobile_search_aggregates` AS
```

In `templates/search_derived/mobile_search_clients_daily_v1/query.sql`, the `END` keyword comes directly after the final statement.
```diff
+CREATE TEMP FUNCTION normalize_fenix_search_key(key STRING) AS (
+  (
+    CASE
     WHEN ARRAY_LENGTH(SPLIT(key, '_')) = 2 THEN SPLIT(key, '_')
     WHEN ARRAY_LENGTH(SPLIT(key, '.')) = 2 THEN SPLIT(key, '.')
-    ELSE []
-  END
-));
-
+    ELSE[] END
+  )
+);
```

In general, comments will retain their level of indentation as an artifact of formatting.
```diff
+  CAST(NULL AS DATE) AS submission_date,
     -- Grouping columns
-    client_id,
```

In `templates/telemetry/fennec_ios_events_v1/view.sql`, the formatting makes the script more difficult to read, but this might be due to the 100 character column limit.

In `templates/telemetry/fxa_all_events_v1/view.sql`, LANGUAGE gets caught as a keyword:
```diff
-      jsonPayload.fields.language,
+      jsonPayload.fields.
+      LANGUAGE,
```

In `templates/telemetry_derived/main_summary_v4/part1.sql`, UPDATE gets caught as a keyword:
```diff
-  environment.settings.update.channel AS update_channel,
+  environment.settings.
+UPDATE
+  .channel AS update_channel,
```

VALUES is also caught as a keyword, where a space is added:
```diff
-      UNNEST(udf_json_extract_histogram(payload.histograms.ssl_handshake_result).values)
+      UNNEST(udf_json_extract_histogram(payload.histograms.ssl_handshake_result). VALUES)
```

In `templates/telemetry_derived/clients_histogram_aggregates_v1/init.sql`, it might be nice to add a newline if it doesn't exist, like `black`. It looks like removing the newline may be intended?
```diff
+  client_id
\ No newline at end of file
```

In `templates/telemetry_derived/clients_histogram_aggregates_v1/query.sql`, the original version looks less wonky:
```diff
-      aggregates ARRAY<STRUCT<key STRING, value INT64>>>(
-        first_bucket,
-        last_bucket,
-        num_buckets,
-        latest_version,
-        metric,
-        metric_type,
-        key,
-        agg_type,
-        aggregates
-  )) AS histogram_aggregates
+          aggregates ARRAY<STRUCT<key STRING, value INT64>>
+        >(
+          first_bucket,
+          last_bucket,
+          num_buckets,
+          latest_version,
+          metric,
+          metric_type,
+          key,
+          agg_type,
+          aggregates
+        )
+      ) AS histogram_aggregates
```

In `templates/telemetry_derived/clients_scalar_aggregates_v1/init.sql`, there's a missing space between the backtick and open parenthesis.

```diff
-  `moz-fx-data-shared-prod.telemetry_derived.clients_scalar_aggregates_v1` (
+  `moz-fx-data-shared-prod.telemetry_derived.clients_scalar_aggregates_v1`(
```

In `templates/telemetry_derived/firefox_desktop_exact_mau28_by_client_count_dimensions_v1/query.sql`, there's a missing space between `AND` and a subexpression:
```diff
-  AND (@submission_date IS NULL OR @submission_date = submission_date)
+  AND(@submission_date IS NULL OR @submission_date = submission_date)
```

In `templates/telemetry_derived/fxa_users_services_first_seen_v1/init.sql`, a comment gets rewrapped due to a use of backticks (leading to a likely syntax error)
```diff
-  -- I've verified that `date(first_service_timestamp), count(distinct user_id) where did_register = true group by 1`  matches the counts of registrations per day in amplitude.
-  first_services_g AS (
+  -- I've verified that `date(first_service_timestamp), count(distinct user_id) where did_register = true group by 1`
+  matches the counts OF registrations per day IN amplitude.first_services_g AS (
```

In `templates/telemetry_derived/fxa_users_services_last_seen_v1/init.sql`, the AS statement looks like it should be on its own line:

```diff
 CLUSTER BY
   service,
-  user_id
-AS
+  user_id AS
 SELECT
   CAST(NULL AS DATE) AS submission_date,
```

bigquery_etl/format_sql/formatter.py
relud: I'm open to suggestions on wording, I came up with this because the first docstring line has to fit within 88 characters.

> What is an inline block?

a bracket block, i.e. anything wrapped in brackets, that is on a single line instead of multiple lines.


for example, reducing:
```sql
IF(
  condition,
  value_if_true,
  value_if_false
)
```
to
```sql
IF(condition, value_if_true, value_if_false)
```
but only when it doesn't exceed `max_line_length`.

relud:
> One thing that could improve this PR is a sample query for before/after comparison or a codification of the behavior in a test.

I'll add an allowlist of queries to enforce formatting on, and commit those changes.

> There are a couple of quirks, but only a few of them break the script as far as I can tell.
> 
> In `templates/monitoring/telemetry_missing_columns_v1/view.sql`, a space is added to a regex literal (syntax error).
> 
> ```diff
> -  REGEXP_EXTRACT(_TABLE_SUFFIX, r"^(.*)_v.*") AS document_type,
> +  REGEXP_EXTRACT(_TABLE_SUFFIX, r "^(.*)_v.*") AS document_type,
> ```

Fixed by tokenizing literals before identifiers.

> In `templates/org_mozilla_fenix/clients_last_seen/view.sql`, a space is added after `-` in a negative literal:
> 
> ```diff
> -    CAST(SAFE.LOG(days_seen_bits & -days_seen_bits, 2) AS INT64) AS days_since_seen,
> +      CAST(SAFE.LOG(days_seen_bits & - days_seen_bits, 2) AS INT64) AS days_since_seen,
> ```

Fixed, and in doing so eliminated the need for the tokenizer to use context to include leading signs in number literals.

> In `templates/search/mobile_search_aggregates/view.sql`, it seems natural to put the create or replace clause on a single line.
> 
> ```diff
> -CREATE OR REPLACE VIEW
> -  `moz-fx-data-shared-prod.search.mobile_search_aggregates`
> -AS
> +CREATE
> +OR REPLACE VIEW `moz-fx-data-shared-prod.search.mobile_search_aggregates` AS
> ```

Fixed. This was due to missing entries for create view DDL in `TOP_LEVEL_KEYWORDS`

> In `templates/search_derived/mobile_search_clients_daily_v1/query.sql`, the `END` keyword comes directly after the final statement.
> 
> ```diff
> +CREATE TEMP FUNCTION normalize_fenix_search_key(key STRING) AS (
> +  (
> +    CASE
>      WHEN ARRAY_LENGTH(SPLIT(key, '_')) = 2 THEN SPLIT(key, '_')
>      WHEN ARRAY_LENGTH(SPLIT(key, '.')) = 2 THEN SPLIT(key, '.')
> -    ELSE []
> -  END
> -));
> -
> +    ELSE[] END
> +  )
> +);
> ```

Fixed by adding `END` to `NEWLINE_KEYWORDS`

> In general, comments will retain their level of indentation as an artifact of formatting.
> 
> ```diff
> +  CAST(NULL AS DATE) AS submission_date,
>      -- Grouping columns
> -    client_id,
> ```

I will break this out into a separate issue, because it's going to be difficult to fix.

> In `templates/telemetry/fennec_ios_events_v1/view.sql`, the formatting makes the script more difficult to read, but this might be due to the 100 character column limit.
> 
> In `templates/telemetry/fxa_all_events_v1/view.sql`, LANGUAGE gets caught as a keyword:
> 
> ```diff
> -      jsonPayload.fields.language,
> +      jsonPayload.fields.
> +      LANGUAGE,
> ```
> 
> In `templates/telemetry_derived/main_summary_v4/part1.sql`, UPDATE gets caught as a keyword:
> 
> ```diff
> -  environment.settings.update.channel AS update_channel,
> +  environment.settings.
> +UPDATE
> +  .channel AS update_channel,
> ```
> 
> VALUES is also caught as a keyword, where a space is added:
> 
> ```diff
> -      UNNEST(udf_json_extract_histogram(payload.histograms.ssl_handshake_result).values)
> +      UNNEST(udf_json_extract_histogram(payload.histograms.ssl_handshake_result). VALUES)
> ```

fixed for the field access case (following `.` operator), but not for the alias case (following `AS`), because `AS` isn't *always* indicating an alias, and I don't want to complicate the tokenizer with the amount of state needed to differentiate this.

> In `templates/telemetry_derived/clients_histogram_aggregates_v1/init.sql`, it might be nice to add a newline if it doesn't exist, like `black`. It looks like removing the newline may be intended?
> 
> ```diff
> +  client_id
> \ No newline at end of file
> ```

fixed

> In `templates/telemetry_derived/clients_histogram_aggregates_v1/query.sql`, the original version looks less wonky:
> 
> ```diff
> -      aggregates ARRAY<STRUCT<key STRING, value INT64>>>(
> -        first_bucket,
> -        last_bucket,
> -        num_buckets,
> -        latest_version,
> -        metric,
> -        metric_type,
> -        key,
> -        agg_type,
> -        aggregates
> -  )) AS histogram_aggregates
> +          aggregates ARRAY<STRUCT<key STRING, value INT64>>
> +        >(
> +          first_bucket,
> +          last_bucket,
> +          num_buckets,
> +          latest_version,
> +          metric,
> +          metric_type,
> +          key,
> +          agg_type,
> +          aggregates
> +        )
> +      ) AS histogram_aggregates
> ```

this isn't something i can fix

> In `templates/telemetry_derived/clients_scalar_aggregates_v1/init.sql`, there's a missing space between the backtick and open parenthesis.
> 
> ```diff
> -  `moz-fx-data-shared-prod.telemetry_derived.clients_scalar_aggregates_v1` (
> +  `moz-fx-data-shared-prod.telemetry_derived.clients_scalar_aggregates_v1`(
> ```

I think there shouldn't be a space here.

> In `templates/telemetry_derived/firefox_desktop_exact_mau28_by_client_count_dimensions_v1/query.sql`, there's a missing space between `AND` and a subexpression:
> 
> ```diff
> -  AND (@submission_date IS NULL OR @submission_date = submission_date)
> +  AND(@submission_date IS NULL OR @submission_date = submission_date)
> ```

I'll deal with this in a follow-up PR. I need to think about how best to identify that these parentheses don't indicate that `AND` is a function.

> In `templates/telemetry_derived/fxa_users_services_first_seen_v1/init.sql`, a comment gets rewrapped due to a use of backticks (leading to a likely syntax error)
> 
> ```diff
> -  -- I've verified that `date(first_service_timestamp), count(distinct user_id) where did_register = true group by 1`  matches the counts of registrations per day in amplitude.
> -  first_services_g AS (
> +  -- I've verified that `date(first_service_timestamp), count(distinct user_id) where did_register = true group by 1`
> +  matches the counts OF registrations per day IN amplitude.first_services_g AS (
> ```

I'll fix this on monday.

> In `templates/telemetry_derived/fxa_users_services_last_seen_v1/init.sql`, the AS statement looks like it should be on its own line:
> 
> ```diff
>  CLUSTER BY
>    service,
> -  user_id
> -AS
> +  user_id AS
>  SELECT
>    CAST(NULL AS DATE) AS submission_date,
> ```

`AS` has lots of meanings, so for simplicity it currently isn't treated as a top level keyword. I'll file a follow-up issue to deal with `AS`.

bigquery_etl/format_sql/formatter.py
acmiyaguchi: The wording is fine, this example would clarifies the wording quite a bit. It would be useful to add below the next sentence.

relud:
> > In `templates/telemetry_derived/firefox_desktop_exact_mau28_by_client_count_dimensions_v1/query.sql`, there's a missing space between `AND` and a subexpression:
> > ```diff
> > -  AND (@submission_date IS NULL OR @submission_date = submission_date)
> > +  AND(@submission_date IS NULL OR @submission_date = submission_date)
> > ```
> 
> I'll deal with this in a follow-up PR. I need to think about how best to identify that these parentheses don't indicate that `AND` is a function.

Fixed this by making `NewlineKeyword` inherit from `SpaceBeforeBracketKeyword`

> > In `templates/telemetry_derived/fxa_users_services_first_seen_v1/init.sql`, a comment gets rewrapped due to a use of backticks (leading to a likely syntax error)
> > ```diff
> > -  -- I've verified that `date(first_service_timestamp), count(distinct user_id) where did_register = true group by 1`  matches the counts of registrations per day in amplitude.
> > -  first_services_g AS (
> > +  -- I've verified that `date(first_service_timestamp), count(distinct user_id) where did_register = true group by 1`
> > +  matches the counts OF registrations per day IN amplitude.first_services_g AS (
> > ```
> 
> I'll fix this on monday.

fixed incorrect comment regex.

> > I'll add an allowlist of queries to enforce formatting on, and commit those changes.
> 
> This would be great. At some point formatting will be required for all new queries?

yes, but I want to do that in a separate PR.

templates/telemetry_derived/main_summary_v4/part1.sql
```diff
@@ -222,13 +230,14 @@ SELECT
       udf_json_extract_histogram(value).sum AS `count`
     FROM
       UNNEST(payload.keyed_histograms.search_counts),
-      UNNEST([REPLACE(key, 'in-content.', 'in-content:')]) AS _key,
+      UNNEST([REPLACE (key, 'in-content.', 'in-content:')]) AS _key,
```
relud: this is incorrectly spaced

relud:
ready for another review. everything should be addressed now, except two cases:

> In `templates/telemetry_derived/clients_histogram_aggregates_v1/query.sql`, the original version looks less wonky:
> 
> ```diff
> -      aggregates ARRAY<STRUCT<key STRING, value INT64>>>(
> -        first_bucket,
> -        last_bucket,
> -        num_buckets,
> -        latest_version,
> -        metric,
> -        metric_type,
> -        key,
> -        agg_type,
> -        aggregates
> -  )) AS histogram_aggregates
> +          aggregates ARRAY<STRUCT<key STRING, value INT64>>
> +        >(
> +          first_bucket,
> +          last_bucket,
> +          num_buckets,
> +          latest_version,
> +          metric,
> +          metric_type,
> +          key,
> +          agg_type,
> +          aggregates
> +        )
> +      ) AS histogram_aggregates
> ```

I don't think I can programmatically achieve the original version, but it can be wrapped in `-- format:off` and `-- format:on` to preserve custom formatting.

> In `templates/telemetry_derived/clients_scalar_aggregates_v1/init.sql`, there's a missing space between the backtick and open parenthesis.
> 
> ```diff
> -  `moz-fx-data-shared-prod.telemetry_derived.clients_scalar_aggregates_v1` (
> +  `moz-fx-data-shared-prod.telemetry_derived.clients_scalar_aggregates_v1`(
> ```

I believe this is valid without a space, and it is easier not to use one here because it means we don't need to differentiate between a udf call and a table definition.

templates/monitoring/telemetry_missing_columns_v1/view.sql
```diff
@@ -1,6 +1,7 @@
 CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.monitoring.telemetry_missing_columns_v1`
-AS SELECT
+AS
+SELECT
```
acmiyaguchi: :+1:

templates/telemetry/fxa_all_events_v1/view.sql
```diff
@@ -1,74 +1,82 @@
...
-  )
-
-SELECT * FROM fxa_auth_events
+fxa_auth_bounce_events AS (
+  SELECT
+    timestamp AS submission_timestamp,
+    jsonPayload.fields.user_id,
+    CAST(
+      NULL AS STRING
+    ) AS country,  -- No country field in auth_bounces
```
acmiyaguchi: Are comments counting toward the line limit here?

relud: comments prevent a line from being merged, regardless of length, so the comment would needs to go on a separate line

PR #641, opened 2020-01-03
Add parsed start and end times to views on top of Glean schemas

jklukas:
Closes https://github.com/mozilla/gcp-ingestion/issues/633

This PR introduces no change to non-Glean views other than some whitespace difference. For Glean pings, though, it adds a UDF call. Here's the result for fenix's baseline ping:

```
CREATE OR REPLACE VIEW
  `moz-fx-data-shared-prod.org_mozilla_fenix.baseline`
AS SELECT
  * REPLACE(
    `moz-fx-data-shared-prod.udf.normalize_metadata`(metadata) AS metadata,
    `moz-fx-data-shared-prod.udf.normalize_ping_info`(ping_info) AS ping_info)
FROM
  `moz-fx-data-shared-prod.org_mozilla_fenix_stable.baseline_v1`
```

### DIFF
--- a/script/generate_views
+++ b/script/generate_views
@@ -152,8 +152,8 @@ def create_views_if_not_exist(client, views, exclude, sql_dir):
                 "`moz-fx-data-shared-prod.udf.normalize_metadata`(metadata)"
                 " AS metadata"
             ]
-            if "ping_info" in (f.name for f in table.schema):
-                # If we have ping_info, this must be a Glean schema.
+            # Add special replacements for Glean pings.
+            if table.labels.get("schema_id", None) == "glean_ping_1":
                 replacements += [
                     "`moz-fx-data-shared-prod.udf.normalize_glean_ping_info`(ping_info)"
                     " AS ping_info"
--- a/script/generate_views
+++ b/script/generate_views
@@ -28,7 +28,7 @@ DEFAULT_EXCLUDE = r"*_raw"
 VIEW_QUERY_TEMPLATE = """
 SELECT
   * REPLACE(
-    `moz-fx-data-shared-prod.udf.normalize_metadata`(metadata) AS metadata)
+    {replacements})
 FROM
   `{target}`
 """
@@ -72,7 +72,7 @@ def main():
     client = bigquery.Client()
     views = get_views(client, args.patterns)
 
-    create_views_if_not_exist(views, args.exclude, args.sql_dir)
+    create_views_if_not_exist(client, views, args.exclude, args.sql_dir)
 
 
 def get_views(client, patterns):
@@ -118,7 +118,7 @@ def get_views(client, patterns):
     return views
 
 
-def create_views_if_not_exist(views, exclude, sql_dir):
+def create_views_if_not_exist(client, views, exclude, sql_dir):
     # create views unless a local file for creating the view exists
     for view, tables in views.items():
         if any(fnmatchcase(pattern, view) for pattern in exclude):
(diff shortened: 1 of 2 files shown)


jklukas:
cc @chutten @mdboom

script/generate_views
```diff
@@ -140,13 +140,23 @@ def create_views_if_not_exist(views, exclude, sql_dir):
...
         view_dataset = dataset.rsplit("_", 1)[0]
         full_view_id = ".".join([project, view_dataset, viewname])
         target_file = os.path.join(sql_dir, view_dataset, viewname, "view.sql")
-        full_sql = f"CREATE OR REPLACE VIEW\n  `{full_view_id}`\nAS {view_query}\n"
 
         if not os.path.exists(target_file):
+            replacements = [
+                "`moz-fx-data-shared-prod.udf.normalize_metadata`(metadata) AS metadata",
+            ]
+            table = client.get_table(target)
```
jklukas: We make this API call inside the loop so that the script can avoid making extra calls when target files already exist on disk.

script/generate_views
```diff
@@ -140,13 +140,23 @@ def create_views_if_not_exist(views, exclude, sql_dir):
...
+            if 'ping_info' in (f.name for f in table.schema):
+                # If we have ping_info, this must be a Glean schema.
```
jklukas: This inference of Glean schemas is that part of this that seems potentially dangerous. It seems unlikely that non-Glean pings would have a top-level `ping_info` field, but a miss here would cause the production schema pipeline to error out. I'm open to suggestions on better ways to determine "glean-ness", but this seems like a pragmatic choice for the time being.

mdboom: Maybe we could put something in the table's metadata?  We know when we generate the schema for the table whether it's glean or not, and maybe we could add something then?

script/generate_views
mdboom: Maybe add a comment about that?

script/generate_views
relud: > I'm open to suggestions on better ways to determine "glean-ness", but this seems like a pragmatic choice for the time being.

glean datasets start with `org_mozilla_`, which doesn't require a `client.get_table` call to determine.

jklukas: > glean datasets start with org_mozilla_, which doesn't require a client.get_table call to determine.

I think the `org_mozilla_` prefix is an artifact of all Glean projects so far being Android projects. I don't know that we're guaranteed to have that prefix for all glean datasets in the future.

And in practice, it looks like the `client.get_table` call is returning quite quickly, so I think it's fine to introduce this API call here, as we already rely on the BQ API to list tables.

jklukas: > Maybe we could put something in the table's metadata? We know when we generate the schema for the table whether it's glean or not, and maybe we could add something then?

I agree this sounds like the right thing to do longer-term. I'll look into how feasible it would be to do so sooner rather than later.

jklukas: [I've put together a PR](https://github.com/mozilla-services/cloudops-infra/pull/1723) to add a `schema_id` label when deploying schemas. For Glean schemas, this would always be set to "moz://mozilla.org/schemas/glean/ping/1", and we would test for existence of such a label here.

whd: This change has been deployed to production so the label should be available, with the label value normalized [as discussed in that PR](https://github.com/mozilla-services/cloudops-infra/pull/1723#issuecomment-571730689]).

jklukas: We now use labels to determine this.

PR #649, opened 2020-01-08
Inject null for negative session_length and event timestamps

jklukas:
As of https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/474
we will be allowing negative values for session lengths and event timestamps.
We provide some safety by modifying the user-facing view for main pings to null
out these negative values (since large negative values may cause significant
skew in aggregations).

We don't modify the user-facing view for event pings, but rather assume users
will be applying the deanonymize_events UDF and handle nulling negative values
there.

See bugs
https://bugzilla.mozilla.org/show_bug.cgi?id=1592012
and
https://bugzilla.mozilla.org/show_bug.cgi?id=1602521

### DIFF
--- a/script/generate_views
+++ b/script/generate_views
@@ -153,10 +153,16 @@ def create_views_if_not_exist(client, views, exclude, sql_dir):
                 " AS metadata"
             ]
             # Add special replacements for Glean pings.
-            if table.labels.get("schema_id", None) == "glean_ping_1":
+            schema_id = table.labels.get("schema_id", None)
+            if schema_id == "glean_ping_1":
                 replacements += [
                     "`moz-fx-data-shared-prod.udf.normalize_glean_ping_info`(ping_info)"
                     " AS ping_info"
+                ]
+            elif schema_id in ("main_ping_1", "main_ping_4"):
+                replacements += [
+                    "`moz-fx-data-shared-prod.udf.normalize_main_payload`(payload)"
+                    " AS payload"
                 ]
             replacements = ",\n    ".join(replacements)
             view_query = VIEW_QUERY_TEMPLATE.format(
--- a/sql/telemetry/core/view.sql
+++ b/sql/telemetry/core/view.sql
@@ -21,6 +21,7 @@ WITH unioned AS (
   SELECT * FROM `moz-fx-data-shared-prod.telemetry_stable.core_v10`)
   --
 SELECT
-  * REPLACE(`moz-fx-data-shared-prod.udf.normalize_metadata`(metadata) AS metadata)
+  * REPLACE(
+    `moz-fx-data-shared-prod.udf.normalize_metadata`(metadata) AS metadata)
 FROM
   unioned
--- a/sql/telemetry/sync/view.sql
+++ b/sql/telemetry/sync/view.sql
@@ -1,4 +1,7 @@
 CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.telemetry.sync`
-AS SELECT * FROM
+AS SELECT
+  * REPLACE(
+    `moz-fx-data-shared-prod.udf.normalize_metadata`(metadata) AS metadata)
+FROM
   `moz-fx-data-shared-prod.telemetry_stable.sync_v4`
--- a/sql/telemetry_derived/main_summary_v4/part1.sql
+++ b/sql/telemetry_derived/main_summary_v4/part1.sql
@@ -720,7 +720,7 @@ SELECT
     NULL AS room_share
   ) AS loop_activity_counter
 FROM
-  `moz-fx-data-shared-prod.telemetry_stable.main_v4`
+  `moz-fx-data-shared-prod.telemetry.main`
 WHERE
   DATE(submission_timestamp) = @submission_date
   AND normalized_app_name = 'Firefox'
--- a/udf/normalize_main_payload.sql
+++ b/udf/normalize_main_payload.sql
@@ -4,33 +4,31 @@ Accepts a pipeline metadata struct as input and returns a modified struct that
 includes a few parsed or normalized variants of the input metadata fields.
 
 */
-
-CREATE TEMP FUNCTION
-  udf_normalize_main_payload(payload ANY TYPE) AS ((
-    SELECT
-      AS STRUCT payload.* REPLACE ( (
-        SELECT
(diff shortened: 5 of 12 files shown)


script/generate_views
```diff
@@ -153,11 +153,17 @@ def create_views_if_not_exist(client, views, exclude, sql_dir):
...
             ]
             # Add special replacements for Glean pings.
-            if table.labels.get("schema_id", None) == "glean_ping_1":
+            schema_id = table.labels.get("schema_id", None)
+            if schema_id == "glean_ping_1":
                 replacements += [
                     "`moz-fx-data-shared-prod.udf.normalize_glean_ping_info`(ping_info)"
                     " AS ping_info"
                 ]
+            elif schema_id in ("main_ping_1", "main_ping_4"):
```
jklukas: Once https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/479 makes it to prod, the schema_id will have a 4 rather than a 1.

sql/telemetry/core/view.sql
```diff
@@ -21,6 +21,7 @@ WITH unioned AS (
   SELECT * FROM `moz-fx-data-shared-prod.telemetry_stable.core_v10`)
   --
 SELECT
-  * REPLACE(`moz-fx-data-shared-prod.udf.normalize_metadata`(metadata) AS metadata)
+  * REPLACE(
+    `moz-fx-data-shared-prod.udf.normalize_metadata`(metadata) AS metadata)
```
jklukas: This is just making whitespace consistent with generated views.

sql/telemetry/sync/view.sql
```diff
@@ -1,4 +1,7 @@
 CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.telemetry.sync`
-AS SELECT * FROM
+AS SELECT
+  * REPLACE(
+    `moz-fx-data-shared-prod.udf.normalize_metadata`(metadata) AS metadata)
```
jklukas: We had missed applying this change to the sync ping view (which we actually commit to the repo here due to the extra v5 version of this table).

sql/telemetry_derived/main_summary_v4/part1.sql
```diff
@@ -720,7 +720,7 @@ SELECT
     NULL AS room_share
   ) AS loop_activity_counter
 FROM
-  `moz-fx-data-shared-prod.telemetry_stable.main_v4`
+  `moz-fx-data-shared-prod.telemetry.main`
```
jklukas: Changing main_summary to hit the user-facing view rather than the stable table so that it gets the nulled sessionLengths.

Looking especially for @relud's okay on this one.

wlach:
Should we also zero out session lengths that are excessively large? This can just as easily throw things off, and is something we protect for in various derived datasets already:

https://github.com/mozilla/bigquery-etl/blob/master/sql/telemetry_derived/error_aggregates/query.sql#L214

jklukas:
> Should we also zero out session lengths that are excessively large?

That's not a bad idea. Do we know what a reasonable upper bound for sessionLength should be, though?

wlach:
> > Should we also zero out session lengths that are excessively large?
> 
> That's not a bad idea. Do we know what a reasonable upper bound for sessionLength should be, though?

I'd wager @chutten or @badboy would know

jklukas:
> lgtm, but we should keep an eye out to see if main summary becomes too complex when the updated main ping view is deployed

Yeah, I will plan to merge this in the morning and then kick off a rerun of the previous day's main_summary job to verify. If it breaks, I'll revert the change.

chutten:
> > > Should we also zero out session lengths that are excessively large?
> > 
> > 
> > That's not a bad idea. Do we know what a reasonable upper bound for sessionLength should be, though?
> 
> I'd wager @chutten or @badboy would know

Session lengths could arguably be as large as the number of seconds since a build with that buildid has been around. A kiosk running release Firefox 50 build 20161115 could have a session over 3 years long by now if it was exceedingly lucky, but anyone running release 72 build 20200107 can't have had it for more than a day.

So, `( submission_date - application.build_id (as a date) + 1 ) * 86400` maybe?

jklukas:
> So, `( submission_date - application.build_id (as a date) + 1 ) * 86400` maybe?

At least for the scope of this PR, I'd prefer to avoid adding that much complexity to this field. I think I worry in particular about relying on parsing build_id.

udf/normalize_main_payload.sql
```diff
@@ -0,0 +1,36 @@
+/*
```
acmiyaguchi: I'd recommend enforcing formatting on this udf, since it looks likely that this will be modified in the future.

chutten:
Taking a different tack, then, let's say a year in seconds. So that's 525600 minutes (thank you RENT) * 60s/min = 31536000s.

Either it's a bad session length, or it's a legitimate session length from a build that's over a year old. Either way we probably don't care to have it in a user-facing view.

wlach:
> Taking a different tack, then, let's say a year in seconds. So that's 525600 minutes (thank you RENT) * 60s/min = 31536000s.
> 
> Either it's a bad session length, or it's a legitimate session length from a build that's over a year old. Either way we probably don't care to have it in a user-facing view.

This seems reasonable to me. I didn't know any of this, I wonder if we should flesh out the section on time in "common analysis" gotchas with some of this information:

https://github.com/mozilla/bigquery-etl/pull/649#issuecomment-572201091

udf/normalize_main_payload.sql
jklukas: I will run the formatter on it. Do we have a way now of marking certain files for enforcing formatting?

jklukas:
> Taking a different tack, then, let's say a year in seconds. So that's 525600 minutes (thank you RENT) * 60s/min = 31536000s.

I've added 31536000 as an upper bound for session_length.

udf/normalize_main_payload.sql
relud: adding a file to this list will enforce formatting: https://github.com/mozilla/bigquery-etl/blob/master/.circleci/config.yml#L52-L60

jklukas: Done

jklukas:
The main_summary queries ran successfully with these changes.

PR #650, opened 2020-01-08
Add tile_id_types view Bug 1604643

jasonthomas:
Background information can be found at https://bugzilla.mozilla.org/show_bug.cgi?id=1604643

r?


jasonthomas:
Thanks! I've updated the view to use `activity_stream` dataset instead.

PR #651, opened 2020-01-08
Add a UDF to extract schema validation path from error_message

acmiyaguchi:
This refactors the schema validation path extraction in `monitoring.schema_error_counts_v1` into a UDF so it can be used else where.


udf/extract_schema_validation_path.sql
```diff
@@ -0,0 +1,27 @@
...
+
+Return a path derived from an error message in `payload_bytes_error`
+
+ */
+CREATE TEMP FUNCTION udf_extract_schema_validation_path(error_message STRING)
+RETURNS STRING AS (
+  IF(
+    STARTS_WITH(error_message, "org.everit.json.schema.ValidationException"),
+    TRIM(SPLIT(error_message, ":")[OFFSET(1)]),
+    ""
```
jklukas: Why not null here rather than an empty string?

acmiyaguchi: I was getting an error with the tests that complained about not being able to compare `INT64` and `STRING`. 

```diff
diff --git a/udf/extract_schema_validation_path.sql b/udf/extract_schema_validation_path.sql
index f3a146a..d461b92 100644
--- a/udf/extract_schema_validation_path.sql
+++ b/udf/extract_schema_validation_path.sql
@@ -8,13 +8,13 @@ RETURNS STRING AS (
   IF(
     STARTS_WITH(error_message, "org.everit.json.schema.ValidationException"),
     TRIM(SPLIT(error_message, ":")[OFFSET(1)]),
-    ""
+    NULL
   )
 );
 -- Tests
 SELECT
   assert_equals(
-    "",
+    NULL,
     udf_extract_schema_validation_path(
       "com.mozilla.telemetry.decoder.Deduplicate$DuplicateIdException: A message with this documentId has already been successfully processed."
     )
```

```
pytest -k udf/extract_schema
```

leads to 


```
400 Invalid function assert_equals
No matching signature for operator = for argument types: INT64, STRING. Supported signatures: ANY = ANY
Analysis of function Templated_SQL_Function:assert_equals failed at [18:3]
```

jklukas: You can use `assert_null` for this case.

You could also use `CAST(NULL AS STRING)`. It defaults to INT64 as type for null when it's ambiguous.

acmiyaguchi:
Adding a path field would still require another round of aggregation, since a path can have many different error messages, since the message includes the specific values that were hit in the validation error.

`schema_error_counts` can be derived from `structured_detailed_error_counts`, with the addition of the other namespaces via `payload_bytes_error.*`, but it does increase the size of the data scanned from about 250GB to 1.5TB due to the inclusion of a sample payload. I think `schema_error_counts` is still useful at this point since it covers all namespaces and can be extended to a longer time period.

acmiyaguchi:
@fbertsch Agreed, I filed #665 for this.

Issue #653, opened 2020-01-08
CI failing on activity_stream/tile_id_types/view.sql


sunahsuh:
I had an unrelated PR just fail the dry-run test with:
`Access Denied: Dataset pocket-tiles:pocket_tiles_data: User does not have bigquery.datasets.update permission for dataset pocket-tiles:pocket_tiles_data.` [1]

Looks like the corresponding file was merged in 3 hours ago and passed dry-run. Did something change permissions-wise in the meantime?

cc @acmiyaguchi @jasonthomas 

[1] https://circleci.com/gh/mozilla/bigquery-etl/8226

acmiyaguchi:
From what I can tell, I don't have `bigquery.jobs.list` permissions on the `pocket-tiles` project, so it seems reasonable that the dry-run would fail. I don't know how the test passed in #650 if it actually applied the views within the test project.

jklukas:
I'm not sure how CI succeeded on the original PR, but this is an authorized view and for now I think it makes sense to basically keep it out of the automated machinery entirely. I will make a PR to that effect so that we don't attempt to publish this view with schema deploys.

jasonthomas:
The circleci and dry run service accounts have data viewer access to the dataset in `pocket-tiles` project. I had added those permissions after tests in #650 failed. 

I am not sure why bigquery.datasets.update permissions is required here but it might be related to authorized views. It could be that we just need to change it to `CREATE VIEW IF NOT EXISTS` instead of `CREATE OR REPLACE VIEW`.

jklukas:
> It could be that we just need to change it to `CREATE VIEW IF NOT EXISTS` instead of `CREATE OR REPLACE VIEW`.

It looks like it still generates an error if you try to run `CREATE VIEW IF NOT EXISTS`:

> Access Denied: Table pocket-tiles:pocket_tiles_data.tile_id_types: User does not have permission to query table pocket-tiles:pocket_tiles_data.tile_id_types.

I'm adding a list of authorized views to skip in the view publishing script: #655

PR #655, opened 2020-01-09
Add list of authorized views to exempt from publishing

jklukas:
Related to CI failures addressed in https://github.com/mozilla/bigquery-etl/issues/653

### DIFF
--- a/script/publish_views
+++ b/script/publish_views
@@ -11,7 +11,13 @@ from google.cloud import bigquery
 import sqlparse
 
 
+AUTHORIZED_VIEWS_TO_SKIP = ("sql/activity_stream/tile_id_types/view.sql",)
+
+
 def process_file(client, args, filepath):
+    if filepath in AUTHORIZED_VIEWS_TO_SKIP:
+        print(f"Skipping authorized view definition {filepath}")
+        return
     with open(filepath) as f:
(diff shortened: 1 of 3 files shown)


whd:
This doesn't appear to properly exclude the view in the configuration used by Jenkins, which runs something like this:

```
cd /app
cp -R sql /tmp/sql
./script/generate_views moz-fx-data-shared-prod:*_stable.* --sql-dir /tmp/sql
./script/publish_views --target-project moz-fx-data-shared-prod /tmp/sql
```

`Skipping authorized view definition` is not printed in the output and the job fails when trying to publish the view.

When talking to :jason it seemed ideal to have Jenkins be able to create these views, but view authorization requires `datasets.update` on the source dataset (which has been discussed as the original CI issue) which we're not likely to always have for all external-facing projects. We could add it in this case since :jason is an owner of the pocket tiles project, but it appears we should in fact exclude views such as this that can't/shouldn't be deployed automatically.

This is separate from the terraform-related issues around authorized views for datasets we do own and manage that have been backed out (which is currently blocked on some policy decisions in addition to the technical ones). Even if we did add `datasets.update` to our Jenkins user for the source dataset of this view, I believe we would need to add the authorization logic ala #583 to properly publish it.

PR #663, opened 2020-01-10
Add view on impression_stats data

jklukas:
Based on a query provided by @kirill-demtchouk

### DIFF
--- a/templates/activity_stream/impression_stats_flat/view.sql
+++ b/templates/activity_stream/impression_stats_flat/view.sql
@@ -1,0 +1,13 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.activity_stream.impression_stats_flat`
+AS
+SELECT
+  tile_id,
+  tile_id_types.type AS tile_type,
+  stats.* EXCEPT (tile_id)
+FROM
+  `moz-fx-data-shared-prod.activity_stream_derived.impression_stats_flat_v1` AS stats
+LEFT JOIN
+  `moz-fx-data-shared-prod.activity_stream.tile_id_types` AS tile_id_types
+USING
+  (tile_id)
--- a/templates/activity_stream_derived/impression_stats_flat_v1/query.sql
+++ b/templates/activity_stream_derived/impression_stats_flat_v1/query.sql
@@ -1,0 +1,45 @@
+SELECT
+  -- truncate timestamp to seconds
+  TIMESTAMP_TRUNC(submission_timestamp, SECOND) AS submission_timestamp,
+  impression_id AS client_id, -- client_id renamed to impression_id in GCP
+  flattened_tiles.id AS tile_id,
+  addon_version,
+  COUNT(loaded) AS loaded,
+  COUNTIF(click IS NULL AND block IS NULL AND pocket IS NULL AND loaded IS NULL) AS impressions,
+  COUNT(click) AS clicks,
+  COUNT(block) AS blocked,
+  COUNT(pocket) AS pocketed,
+  -- the 3x1 layout has a bug where we need to use the position of each element
+  -- in the tiles array instead of the actual pos field
+  IFNULL(flattened_tiles.pos, alt_pos) AS position,
+  page,
+  source,
+  locale,
+  normalized_country_code AS country_code,
+  version,
+  user_prefs,
+  release_channel,
+  IFNULL(shield_id, 'n/a') AS shield_id,
+  sample_id
+FROM
+  activity_stream_stable.impression_stats_v1
+CROSS JOIN
+  UNNEST(tiles) AS flattened_tiles
+WITH OFFSET AS alt_pos
+WHERE
+  (@submission_date IS NULL OR @submission_date = DATE(submission_timestamp))
+GROUP BY
+  submission_timestamp,
+  impression_id,
+  tile_id,
+  addon_version,
+  position,
+  page,
+  source,
+  locale,
+  country_code,
+  version,
+  user_prefs,
+  release_channel,
+  shield_id,
+  sample_id
--- a/sql/activity_stream/impression_stats_flat/view.sql
+++ b/sql/activity_stream/impression_stats_flat/view.sql
@@ -1,0 +1,13 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.activity_stream.impression_stats_flat`
+AS
+SELECT
+  tile_id,
+  tile_id_types.type AS tile_type,
+  stats.* EXCEPT (tile_id)
+FROM
+  `moz-fx-data-shared-prod.activity_stream_derived.impression_stats_flat_v1` AS stats
+LEFT JOIN
+  `moz-fx-data-shared-prod.activity_stream.tile_id_types` AS tile_id_types
+USING
+  (tile_id)
--- a/sql/activity_stream_derived/impression_stats_flat_v1/query.sql
+++ b/sql/activity_stream_derived/impression_stats_flat_v1/query.sql
@@ -1,0 +1,45 @@
+SELECT
+  -- truncate timestamp to seconds
+  TIMESTAMP_TRUNC(submission_timestamp, SECOND) AS submission_timestamp,
+  impression_id AS client_id, -- client_id renamed to impression_id in GCP
+  flattened_tiles.id AS tile_id,
+  addon_version,
+  COUNT(loaded) AS loaded,
+  COUNTIF(click IS NULL AND block IS NULL AND pocket IS NULL AND loaded IS NULL) AS impressions,
+  COUNT(click) AS clicks,
+  COUNT(block) AS blocked,
+  COUNT(pocket) AS pocketed,
+  -- the 3x1 layout has a bug where we need to use the position of each element
+  -- in the tiles array instead of the actual pos field
+  IFNULL(flattened_tiles.pos, alt_pos) AS position,
+  page,
+  source,
+  locale,
+  normalized_country_code AS country_code,
+  version,
+  user_prefs,
+  release_channel,
+  IFNULL(shield_id, 'n/a') AS shield_id,
+  sample_id
+FROM
+  activity_stream_stable.impression_stats_v1
+CROSS JOIN
+  UNNEST(tiles) AS flattened_tiles
+WITH OFFSET AS alt_pos
+WHERE
+  (@submission_date IS NULL OR @submission_date = DATE(submission_timestamp))
+GROUP BY
+  submission_timestamp,
+  impression_id,
+  tile_id,
+  addon_version,
+  position,
+  page,
+  source,
+  locale,
+  country_code,
+  version,
+  user_prefs,
+  release_channel,
+  shield_id,
+  sample_id


templates/activity_stream/tile_impressions/view.sql
```diff
@@ -0,0 +1,82 @@
...
+    *
+  FROM
+    `moz-fx-data-shared-prod.activity_stream_stable.impression_stats_v1`
+  UNION ALL
+  SELECT
+    *
+  FROM
+    `moz-fx-data-shared-prod.activity_stream_stable.impression_stats_v1`
+  WHERE
+    DATE(submission_timestamp) = CURRENT_DATE()
```
jklukas: We could provide separate views for historical and live data, or we can union the two tables like this, which will likely be more convenient if it performs well.

templates/activity_stream/tile_impressions/view.sql
```diff
@@ -0,0 +1,82 @@
...
+    *
+  FROM
+    unioned
+  WHERE
+    -- make sure data is valid/non-empty
+    ARRAY_LENGTH(tiles) >= 1
+)
+SELECT
+  -- truncate timestamp to seconds
+  TIMESTAMP_TRUNC(submission_timestamp, SECOND) AS submission_timestamp,
```
jklukas: BigQuery needs to be able to push down filters on `submission_timestamp` to the underlying tables to limit the number of day partitions we scan. The base query used `UNIX_SECONDS` here which yields an integer; BigQuery is not able to infer partitions across that type transformation, but it _is_ able to infer partitions through the `TIMESTAMP_TRUNC` function.

templates/activity_stream/tile_impressions/view.sql
```diff
@@ -0,0 +1,82 @@
...
+  impression_id AS client_id, -- client_id renamed to impression_id in GCP
+  flattened_tiles.id AS tile_id,
+  addon_version,
+  SUM(CASE WHEN loaded IS NOT NULL THEN 1 ELSE 0 END) AS loaded,
```
jklukas: I believe these `SUM(CASE WHEN x IS NOT NULL THEN 1 ELSE 0 END)` aggregations are equivalent to `COUNT(x)`; see https://cloud.google.com/bigquery/docs/reference/standard-sql/aggregate_functions#count

templates/activity_stream/tile_impressions/view.sql
```diff
@@ -0,0 +1,82 @@
...
+  source,
+  locale,
+  normalized_country_code AS country_code,
+  version,
+  user_prefs,
+  release_channel,
+  CASE
+  WHEN shield_id IS NULL THEN 'n/a'
+  ELSE shield_id
+  END AS shield_id,
```
jklukas: Can this be `IFNULL(shield_id, 'n/a') AS shield_id`?

jklukas:
I've published this draft view, but it looks like it does not perform very well. For example:

```
select * from `moz-fx-data-shared-prod.activity_stream.tile_impressions`
where date(submission_timestamp) = '2020-01-09'
AND sample_id = 0
limit 100
```

That query hits only a single day of data, ideally it would hit only a 1% sample by using sample_id = 0, and it is limited to return only 100 rows, but the query takes almost 2 minutes to run. I think the aggregation that's happening here essentially requires BQ to scan everything, do a bunch of repartitioning and sorting on the entire day partition regardless of filters or limit clause.

Can you explain more about the intention of this aggregation? If the purpose of aggregating by second is deduplication, then we likely can avoid aggregation altogether since the historical ping tables are guaranteed to be fully deduplicated within each day partition based on document_id. This view would be able to take advantage of clustering and limit clauses if we can remove the aggregation.

Otherwise, we probably need to turn this query into a scheduled job that runs nightly and materializes results to a table in order to get reasonable performance for user queries.

jklukas:
Note that if I limit to a very small time range, the view gives near instantaneous performance. The following returns in 2 seconds:

```
select * from `moz-fx-data-shared-prod.activity_stream.tile_impressions`
where submission_timestamp between '2020-01-09 12:00:00' and '2020-01-09 12:00:10'
AND sample_id = 0
limit 100
```

So, I think performance worsens geometrically with the size of the time range considered. I don't think BQ is able to take advantage of the fact that it can aggregate individual seconds independently.

templates/activity_stream/tile_impressions/view.sql
kirill-demtchouk: That would be awesome if performance is not a constraint!

templates/activity_stream/tile_impressions/view.sql
kirill-demtchouk: Amazing, thank you for the change.

templates/activity_stream/tile_impressions/view.sql
kirill-demtchouk: I am ok with this change, thanks for pointing out!

templates/activity_stream/tile_impressions/view.sql
kirill-demtchouk: I am ok with this change, thanks for pointing out!

kirill-demtchouk:
> I've published this draft view, but it looks like it does not perform very well. For example:
> 
> ```
> select * from `moz-fx-data-shared-prod.activity_stream.tile_impressions`
> where date(submission_timestamp) = '2020-01-09'
> AND sample_id = 0
> limit 100
> ```
> 
> That query hits only a single day of data, ideally it would hit only a 1% sample by using sample_id = 0, and it is limited to return only 100 rows, but the query takes almost 2 minutes to run. I think the aggregation that's happening here essentially requires BQ to scan everything, do a bunch of repartitioning and sorting on the entire day partition regardless of filters or limit clause.
> 
> Can you explain more about the intention of this aggregation? If the purpose of aggregating by second is deduplication, then we likely can avoid aggregation altogether since the historical ping tables are guaranteed to be fully deduplicated within each day partition based on document_id. This view would be able to take advantage of clustering and limit clauses if we can remove the aggregation.
> 
> Otherwise, we probably need to turn this query into a scheduled job that runs nightly and materializes results to a table in order to get reasonable performance for user queries.

@jklukas Responding to your question here. The main purpose of this aggregation is to detect bad click data. We've had instances in the past where the same `tile_id` was clicked by the same `impression_id`/`client_id` within the same second many times.

Through chatting with our data & learning team that also uses the tile data, we agreed on an informal threshold of <3 clicks per tile ID within one second by the same browser instance (i.e. allowing for single and double clicks but not anything else).

I was defaulting to the same level of aggregation that the old `tiles` Redshift ETL was doing (one record per second per client ID per tile ID) in order to detect the bad clicks. Are you saying this de-duping shouldn't be a concern in GCP?

jklukas:
> I was defaulting to the same level of aggregation that the old tiles Redshift ETL was doing (one record per second per client ID per tile ID) in order to detect the bad clicks. Are you saying this de-duping shouldn't be a concern in GCP?

If each click is generating a separate payload with a distinct document_id (I assume this is the case), then the pipeline's deduping isn't going to help. We'd still see one row per click in the `stable` table.

Thanks for the explanation here. If this aggregated view is the primary way that people will want to look at this data, I definitely think it's worth making it a scheduled query and materializing the results nightly.

kirill-demtchouk:
> And also, what would you want to name this? I picked `tile_impressions` somewhat out of thin air.

Would `impression_stats_flat` work?

kirill-demtchouk:
> > I was defaulting to the same level of aggregation that the old tiles Redshift ETL was doing (one record per second per client ID per tile ID) in order to detect the bad clicks. Are you saying this de-duping shouldn't be a concern in GCP?
> 
> If each click is generating a separate payload with a distinct document_id (I assume this is the case), then the pipeline's deduping isn't going to help. We'd still see one row per click in the `stable` table.
> 
> Thanks for the explanation here. If this aggregated view is the primary way that people will want to look at this data, I definitely think it's worth making it a scheduled query and materializing the results nightly.

No problem! How would I go about doing that? If it's costly from a storage/processing perspective on GCP, I can just try using the view for now!

jklukas:
> impression_stats_flat

Sounds fine.

> How would I go about doing that? 

Basically we define this query as a `query.sql` file instead of `view.sql` in this PR and we put it under an `activity_stream_derived` directory. Then we need to make a PR to Airflow to schedule it. The process is documented in https://docs.telemetry.mozilla.org/cookbooks/bigquery-airflow.html

I can go ahead and make those changes here for this go around. Sound good?

> If it's costly from a storage/processing perspective on GCP, I can just try using the view for now!

There are a lot of rows in this table, but the overall data size is pretty small. The total size right now is ~8 TB, which is about the same as a single day of main ping data. So, I think it's not unreasonable to duplicate it in this way if it provides major usability benefits.

kirill-demtchouk:
@jklukas If you want to go ahead and make the change from a view to a schedule query, I would be grateful!

jklukas:
I've update this PR to be a scheduled query and I'm running a backfill right now. You should be able to hit `activity_stream.impression_stats_flat` and see dates in early 2019-07, with more filling in over the next few hours. Please take a look and let me know if anything seems amiss. If not, I'll go ahead with the PR for scheduling.

kirill-demtchouk:
I took a look at the initial data! No complaints @jklukas . You can proceed with the backfill. Thank you!

jklukas:
Prepping schedule in https://github.com/mozilla/telemetry-airflow/pull/832

templates/activity_stream/impression_stats_flat/view.sql
```diff
@@ -0,0 +1,13 @@
...
+  `moz-fx-data-shared-prod.activity_stream.impression_stats_flat`
+AS
+SELECT
+  tile_id,
+  tile_id_types.type AS tile_type,
+  stats.* EXCEPT(tile_id)
+FROM
+  `moz-fx-data-shared-prod.activity_stream_derived.impression_stats_flat_v1` AS stats
+LEFT JOIN
+  `moz-fx-data-shared-prod.activity_stream.tile_id_types` AS tile_id_types
```
jklukas: My understanding is that the `tile_id_types` view exists in order to enrich the impression_stats data and we can do the join as part of this view with negligible impact since `tile_id_types` is so small.

@kirill-demtchouk, does this change look good to you? It makes a `tile_type` column available when querying `activity_stream.impression_stats_flat`.

kirill-demtchouk: @jklukas Looks good to me. Good call!

templates/activity_stream_derived/impression_stats_flat_v1/query.sql
```diff
@@ -0,0 +1,60 @@
...
+),
+impression_data AS (
+  SELECT
+    *
+  FROM
+    base
+  WHERE
+    -- make sure data is valid/non-empty
+    ARRAY_LENGTH(tiles) >= 1
+)
```
relud: these CTEs seem unnecessary. the `ARRAY_LENGTH(tiles) >= 1` condition is redundant with `UNNEST(impression_data.tiles)` and the `submission_date` filter should work fine on the final query.

jklukas:
This is now scheduled and backfilled. Enjoy querying `activity_stream.impression_stats_flat`!

PR #670, opened 2020-01-13
Add fields to clustered event tables

akkomar:
This adds more fields that are used in Devtools Amplitude events view.

Should be merged after https://github.com/mozilla/telemetry-airflow/pull/833.


sql/telemetry_derived/event_events_v1/init.sql
```diff
@@ -26,7 +26,15 @@ CREATE TABLE IF NOT EXISTS
...
-    event_process STRING
+    event_process STRING,
+    build_id STRING,
+    build_architecture STRING,
+    profile_creation_date FLOAT64,
+    is_default_browser BOOLEAN,
+    attribution_source STRING,
+    system_is_wow64 BOOLEAN,
+    system_memory_mb FLOAT64,
+    city STRING
```
sunahsuh: Logically I feel like it makes sense to move these up before the event-specific fields where the rest of the ping metadata is (maybe after `os_version`?)

jklukas: Practically, these fields will end up at the end of the schema regardless of how they're defined here. When we update a table to include new fields, they can only be appended at the end (or for nested fields, appended at the end of the struct they're a part of). If we were to recreate the table, we could have any order we want, and perhaps it's a good thing to define the init.sql to match that ideal state?

sunahsuh: Sad news 😿 I think if that's the case init.sql and the view statement should match

akkomar: > (...) define the init.sql to match that ideal state?

Yep, I think this is a good idea since we'll need to rewrite this anyway.

akkomar: > (...) we'll need to rewrite this anyway.

On second thought, I'm not sure that's the case. @jklukas what's the best way for backfilling these tables?
I assumed we'd create their new versions in separate dataset, backfill them and then move to production dataset, although I'm not sure the last part is feasible in BigQuery.

sunahsuh: FWIW the user-facing view is the unioned view,`events_v1` anyway – we can codify the preferred order there

jklukas: >  I assumed we'd create their new versions in separate dataset, backfill them and then move to production dataset, although I'm not sure the last part is feasible in BigQuery.

If you're planning to completely recreate the tables in order to backfill these fields, then this plan sounds fine. You can't "move" a table, but you can copy it from one place to another relatively instantaneously. So it's feasible to create the new tables in a temporary location, backfill them, and then do a fairly quick succession of operations to delete the old tables and copy the new versions into place. I believe you can actually do that atomically now in a script so that users would never see a time where the table doesn't exist.

jklukas: I had assumed we were going to live without the fields for older days, and just backfill from 2019-12-04 (which we're doing anyway for sessionLength).

akkomar: > backfill from 2019-12-04 (which we're doing anyway for sessionLength).

Good point, I didn't realized that. Then let's keep new fields after the existing ones both in queries and init scripts for consistency. As Sunah suggested, we'll use view for codifying a nicer order.

Issue #672, opened 2020-01-13
Filter min/max/num_buckets using probe info service for glam


emtwo:
```
SELECT payload.histograms.memory_total
FROM `moz-fx-data-shared-prod.telemetry_stable.main_v4`
WHERE DATE(submission_timestamp) = '2020-01-08'
  AND normalized_os = 'Windows'
  AND application.build_id = '20191202093317'
  AND normalized_channel = 'release'
  AND SPLIT(application.version, '.')[OFFSET(0)] = '71'
  AND client_id IS NOT NULL
  AND JSON_EXTRACT(payload.histograms.memory_total, "$.bucket_count") = '813824896'
```

This query shows an example of data with a bogus bucket count of `813824896`.

We should be filtering out data with incorrect bucket counts and/or using probe info service to check what the correct bucket counts are.

emtwo:
resolved

PR #673, opened 2020-01-13
Glam bug fixes

emtwo:
This PR contains 2 bug fixes:

1. There are new out of memory errors triggered by buckets again. This time its due to the number of buckets being too high. This query shows us a row that triggers this issue:

```
SELECT payload.histograms.memory_total
FROM `moz-fx-data-shared-prod.telemetry_stable.main_v4`
WHERE DATE(submission_timestamp) = '2020-01-08'
  AND normalized_os = 'Windows'
  AND application.build_id = '20191202093317'
  AND normalized_channel = 'release'
  AND SPLIT(application.version, '.')[OFFSET(0)] = '71'
  AND client_id IS NOT NULL
  AND JSON_EXTRACT(payload.histograms.memory_total, "$.bucket_count") = '813824896'
```

It looks like `813824896` is a bogus number of buckets. So I looked at the probe info service and it seems the max number of buckets is actually `10000` which explains the change I've made in the code. However, this is a temporary solution as it does not get rid of bogus rows of data. As a follow up, I've filed an issue here to handle the issue by looking at the probe info service: https://github.com/mozilla/bigquery-etl/issues/672

2. Unexpectedly `normalized_os` in main ping has some values that are null. However, these get mixed up with `NULL` being a representation of *all* OS's. This results in duplicate rows where `os = NULL`. As such, when os is null, it should not be included as the name of an OS.

### DIFF
--- a/sql/telemetry_derived/clients_histogram_probe_counts_v1/query.sql
+++ b/sql/telemetry_derived/clients_histogram_probe_counts_v1/query.sql
@@ -8,7 +8,7 @@ LANGUAGE js AS
     current = 1;
   } // If starting from 0, the second bucket should be 1 rather than 0
   let retArray = [0, current];
-  for (let bucketIndex = 2; bucketIndex < Math.min(nBuckets, max); bucketIndex++) {
+  for (let bucketIndex = 2; bucketIndex < Math.min(nBuckets, max, 10000); bucketIndex++) {
     let logCurrent = Math.log(current);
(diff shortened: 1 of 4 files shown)


emtwo:
> For `os` being null, I want to be clear about what this behavior is doing. This means those rows will be included for all other aggregates (e.g. when selecting "all OSs"), but there will _not_ be an option to only select `NULL` os. Is that right?

yes correct

PR #679, opened 2020-01-15
Bug 1592930 Filter out histograms in fenix metrics ping from Glean SDK<19

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1592930


script/generate_views
```diff
@@ -159,6 +159,14 @@ def create_views_if_not_exist(client, views, exclude, sql_dir):
                     "`moz-fx-data-shared-prod.udf.normalize_glean_ping_info`(ping_info)"
                     " AS ping_info"
                 ]
+                if table.dataset_id.startswith(
```
fbertsch: Why this instead of explicitly defining the view?

jklukas: Because we have this machinery now, and if we make general updates to the glean views in the future, they'll be included automatically in the fenix metrics views this way. If we explicitly define the view files, we'll need to remember to apply the same changes in the future.

If this seems too magical, though, explicitly defining the views is definitely an option.

fbertsch: That makes sense, seems that this machinery could use some standardization though. Maybe in a centralized file that just has e.g. namespace, document, query

jklukas: > That makes sense, seems that this machinery could use some standardization though. Maybe in a centralized file that just has e.g. namespace, document, query

We can consider better standardization for the future, but for the time being, I don't know that we understand the constraints well enough. By keeping this in python for the moment, we have the flexibility to define a condition like this that will apply to all fenix apps, even if additional channels are defined (such as beta which we have so far been ignoring).

I've split out the problem of standardizing this process to https://github.com/mozilla/bigquery-etl/issues/691

fbertsch: Great. I meant to approve when I put that other comment, LGTM

PR #681, opened 2020-01-16
Bug 1609666 Use SAFE_CAST in udf.json_extract_int_map

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1609666

I am running main_summary with this change right now to verify whether it fixes the exception we saw in last night's runs.


relud:
main summary succeeded

Issue #680, opened 2020-01-16
Table descriptions as metadata in BigQuery


scholtzan:
Related to [bugzilla#1603536](https://bugzilla.mozilla.org/show_bug.cgi?id=1603536) it might be useful to allow adding descriptions to tables that will then be set as [the table's description on BigQuery](https://cloud.google.com/bigquery/docs/managing-tables#updating_a_tables_description).

I'm not sure what the best approach here is. One way might be doing something similar to how tests for UDFs are defined currently. So having, for example, a `-- Description: ... table description ...` comment inside the template file and updating the description for the table on BigQuery.

I'm open for any other suggestions.

jklukas:
There was a proposal a while ago for adding `metadata.yaml` files in the directories alongside `query.sql` files. The metadata file would be a great place for table descriptions, clustering definitions, etc. We could also put info about dependencies and scheduling in there if we moved forward with the idea of automating creation of a DAG from bigquery-etl queries.

scholtzan:
Sounds related to https://github.com/mozilla/bigquery-etl/issues/296

scholtzan:
I guess that is the proposal: https://docs.google.com/document/d/1zJtRTiiBxpE6faf-HNL6IVTXIFHXvceKCxWXYM0wf9w

jklukas:
> I guess that is the proposal: https://docs.google.com/document/d/1zJtRTiiBxpE6faf-HNL6IVTXIFHXvceKCxWXYM0wf9w

Indeed!

PR #684, opened 2020-01-17
Table metadata

scholtzan:
This allows to define metadata for tables in `metadata.yaml` files which can be added to directories in `/template` that contain some `.sql` file. The metadata file gets parsed and used to update table metadata in BigQuery.

This addresses https://github.com/mozilla/bigquery-etl/issues/680


script/publish_metadata
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python3
```
scholtzan: I'm not sure if having a script for updating the metadata is the right way? For now that would mean metadata would require manual execution of this script. 

Or should metadata get updated every time corresponding queries are executed via Airflow? Or as a separate step after that?

script/publish_metadata
```diff
@@ -0,0 +1,79 @@
...
+    try:
+        logging.basicConfig(level=args.log_level, format="%(levelname)s %(message)s")
+    except ValueError as e:
+        parser.error(f"argument --log-level: {e}")
+
+    for target in args.target:
+        if os.path.isdir(target):
+            for root, dirs, files in os.walk(target):
+                if METADATA_FILE in files:
+                    dataset = root.split("/")[-2]
```
jklukas: It would be technically safer to use `os.path.split` here, but this is likely fine.

script/generate_sql
```diff
@@ -66,6 +69,12 @@ def main():
                 with open(os.path.join(basename, filename), "a+") as output_file:
                     output_file.write(full_text)
 
+            if METADATA_FILE in files:
+                shutil.copyfile(
+                    os.path.join(root, METADATA_FILE),
+                    os.path.join(basename, METADATA_FILE),
+                )
```
relud: I think we shouldn't copy these to `sql/`, and instead just run `script/publish_metadata` against `templates/`

templates/telemetry_derived/clients_daily_v6/metadata.yaml
```diff
@@ -0,0 +1,11 @@
+friendly_name: Clients Daily
+description: >
+  A daily aggregate of each Firefox client,
+  partitioned by day
+owners:
+  - relud@mozilla.com
+labels:
+  application: firefox
+  refresh: daily
+  version: 6
```
relud: this label is redundant with the table name
```suggestion
```

templates/telemetry_derived/clients_daily_v6/metadata.yaml
```diff
@@ -0,0 +1,11 @@
+friendly_name: Clients Daily
+description: >
+  A daily aggregate of each Firefox client,
+  partitioned by day
+owners:
+  - relud@mozilla.com
```
relud: this should use my canonical mozilla email, instead of the alias:
```suggestion
  - dthorn@mozilla.com
```

templates/telemetry_derived/clients_daily_v6/metadata.yaml
```diff
@@ -0,0 +1,11 @@
+friendly_name: Clients Daily
+description: >
+  A daily aggregate of each Firefox client,
```
jklukas:
```suggestion
  A daily aggregate of main pings from each Firefox client,
```

templates/telemetry_derived/clients_daily_v6/metadata.yaml
```diff
@@ -0,0 +1,11 @@
...
+  partitioned by day
+owners:
+  - relud@mozilla.com
+labels:
+  application: firefox
+  refresh: daily
+  version: 6
+  incremental: true
```
jklukas: BQ labels don't natually suit boolean values; IIUC this will be a string value "true". Maybe `schedule: daily` would be more extensible than `incremental: true`. I need to read the proposal again to see how much the particular labels were discussed.

templates/telemetry_derived/clients_daily_v6/metadata.yaml
relud:
```suggestion
  A daily aggregate of main pings from each Firefox desktop client,
```

script/publish_metadata
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python3
+
+"""Update metadata of BigQuery tables."""
```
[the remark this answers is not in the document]
scholtzan: Yes, it works for views too.

templates/telemetry_derived/clients_daily_v6/metadata.yaml
scholtzan: True. As far as I can tell, the specific labels aren't really discussed in the proposal. I just copied the example for `clients_daily` from there for testing.

scholtzan:
Ready for another look

templates/telemetry_derived/clients_daily_v6/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: Clients Daily
+description: >
+  A daily aggregate of main pings from each Firefox desktop client,
+  partitioned by day
+owners:
+  - dthorn@mozilla.com
+labels:
+  application: firefox
+  refresh: daily
+  schedule: daily
```
relud: we probably only need one of these, i prefer `schedule`

Issue #439, opened 2019-10-22
Write UDF for extracting parts of the `uri` field in payload_bytes_error and payload_bytes_raw


sunahsuh:
Would be a nice convenience function when investigating issues

scholtzan:
https://github.com/mozilla/bigquery-etl/pull/634

PR #687, opened 2020-01-20
Add UDF for extracting payload URI

scholtzan:
This addresses https://github.com/mozilla/bigquery-etl/issues/439

### DIFF
--- a/udf/extract_payload_uri.sql
+++ b/udf/extract_payload_uri.sql
@@ -1,0 +1,81 @@
+/*
+
+Extract all parts of the `uri` field used in payload tables. 
+
+*/
+CREATE TEMP FUNCTION 
+  udf_extract_payload_uri(uri STRING) 
+  RETURNS STRUCT<
+    prefix STRING,
+    document_namespace STRING,
+    document_id STRING,
+    document_type STRING,
+    browser STRING,
+    version STRING,
+    channel STRING,
+    build_id STRING>
+  AS ((
+    SELECT
+      STRUCT(
+        SPLIT(uri, "/")[
+          OFFSET(1)
+        ] AS prefix,
+        SPLIT(uri, "/")[
+          OFFSET(2)
+        ] AS document_namespace,
+        SPLIT(uri, "/")[
+          OFFSET(3)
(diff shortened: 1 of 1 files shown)


udf/extract_payload_uri.sql
```diff
@@ -0,0 +1,80 @@
+/*
+
+Extract all parts of the `uri` field used in payload tables. 
+
+*/
+CREATE TEMP FUNCTION udf_extract_payload_uri(uri STRING)
```
jklukas: It looks like we're moving towards a suite of uri parsing functions, so we should make sure they're named similarly for better discoverability. In particular, we have `parse_desktop_telemetry_uri` which may be a subset of the functionality here.

We need different functions for telemetry vs. structured; see https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java#L56-L63

acmiyaguchi: The edge spec is also found in the data docs, for reference: https://docs.telemetry.mozilla.org/concepts/pipeline/http_edge_spec.html

udf/extract_payload_uri.sql
```diff
@@ -0,0 +1,80 @@
...
+        SPLIT(uri, "/")[OFFSET(8)] AS build_id
+      )
+  )
+);
+-- Tests
+SELECT
+  assert_equals(
+    "submit",
+    udf_extract_payload_uri(
+      "/submit/telemetry/9ea03c76-1d1e-4a12-b1db-6c93d7b0d052/core/Fennec/46.0.1/release/20160502161457"
```
acmiyaguchi: Tests in the future might benefit from using the scripting functionality, in particular the [`DECLARE` expression](https://cloud.google.com/bigquery/docs/reference/standard-sql/scripting#declare):

```
DECLARE test_uri STRING DEFAULT "/submit/telemetry/9ea03c76-1d1e-4a12-b1db-6c93d7b0d052/core/Fennec/46.0.1/release/20160502161457";
```

Issue #469, opened 2019-10-31
test the speed of calculating crc32 in a stored procedure


relud:
[stored procedures](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_procedure) allow recursion, unlike udfs, so we might be able to implement a better version of [udf_safe_crc32_uuid](https://github.com/mozilla/bigquery-etl/blob/master/udf/safe_crc32_uuid.sql) that isn't limited to a precise input length.

relud:
fixed by #689

PR #689, opened 2020-01-20
Stored procedure for calculating CRC-32 hash of arbitrary length

scholtzan:
Addresses https://github.com/mozilla/bigquery-etl/issues/469

Not sure if we want to merge this. I tested the speed of this stored procedure and it's about 2 to 3 times slower than the UDF. 
Run time stored procedure: ~1.2s
Run time UDF `safe_crc32_uuid`: ~0.4s

But on the positive side, this implementation is not limited to a certain length. We would need to change/add some logic to publish stored procedures though.


relud:
> Not sure if we want to merge this.

I think that we should include this as an example of a stored procedure, even if it's slower than the sql and js versions.

> I tested the speed of this stored procedure and it's about 2 to 3 times slower than the UDF.
> Run time stored procedure: ~1.2s
> Run time UDF `safe_crc32_uuid`: ~0.4s

Those timings look like it's for a single input. We should test with a larger input, e.g. calculating sample_id for a whole day of clients_daily. Also, it would be good to include timing for `udf_js_sample_id`, which I believe @jklukas determined runs as fast as `udf_safe_crc32_uuid`.

relud:
> But on the positive side, this implementation is not limited to a certain length.

actually, from [the scripting docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/scripting#call)

    The maximum depth of procedure calls is 50 frames.

scholtzan:
> `The maximum depth of procedure calls is 50 frames.`

Will that actually become an issue here? I was hoping that using an iterative algorithm rather than a recursive one would avoid problems like these.

relud:
> > `The maximum depth of procedure calls is 50 frames.`
> 
> Will that actually become an issue here? I was hoping that using an iterative algorithm rather than a recursive one would avoid problems like these.

oh, good point, i got mixed up

stored_procedures/safe_crc32_uuid.sql
```diff
@@ -0,0 +1,89 @@
...
+SET
+  k = TO_CODE_POINTS(SUBSTR(value, i, 1))[SAFE_OFFSET(0)];
+SET
+  crc = (crc >> 8) ^ crc_table[OFFSET(((crc & 0xFF) ^ k))];
+SET
+  i = i + 1;
+END WHILE;
+SET
+  crc = crc ^ 0xFFFFFFFF;
+END
```
relud: pretty sure this needs a semicolon for ending the procedure definition
```suggestion
END;
```

stored_procedures/safe_crc32_uuid.sql
```diff
@@ -0,0 +1,89 @@
+/*
+Calculate the CRC-32 hash of a UUID using a stored procedure.
+
+Based on https://stackoverflow.com/a/18639999/1260237
+See https://en.wikipedia.org/wiki/Cyclic_redundancy_check
+*/
+CREATE
+OR REPLACE PROCEDURE udf_safe_crc32_uuid(value BYTES, OUT crc INT64) BEGIN DECLARE len INT64;
```
relud: please format with the updates in #693, and add this file to `SKIP` in [script/format_sql](https://github.com/mozilla/bigquery-etl/blob/master/script/format_sql#L17)

scholtzan:
I ran some more tests with a larger input (2020-01-19 of clients_daily):
Processing with `udf_safe_crc32_uuid` and `udf_js_sample_id` takes less than a minute.
Using the stored procedure, processing 100 rows takes more than 3 minutes. Having more rows as input results in a "resources exceeded" error.

I moved the stored procedure into a separate directory `stored_procedures/` for now, in case we want to keep it around as an example.

stored_procedures/safe_crc32_uuid.sql
```diff
@@ -0,0 +1,89 @@
...
+  crc = crc ^ 0xFFFFFFFF;
+END
+-- Tests
+DECLARE crc1 INT64;
+DECLARE crc2 INT64;
+CALL udf_safe_crc32_uuid(b"51baf8b4-75d1-3648-b96d-809569b89a12", crc1);
+CALL udf_safe_crc32_uuid(b"0", crc2);
+SELECT
+  assert_equals(308953907, crc1),
+  assert_equals(4108050209, crc2);
```
relud: Please file a follow-up issue to add support for running this via a pytest. We're also going to need some way to delimit tests, so that we can have separate tests with some that are `xfail`.

stored_procedures/safe_crc32_uuid.sql
```diff
@@ -0,0 +1,89 @@
...
+SET
+  i = 1;
+WHILE len >= i DO
+SET
+  k = TO_CODE_POINTS(SUBSTR(value, i, 1))[SAFE_OFFSET(0)];
+SET
+  crc = (crc >> 8) ^ crc_table[OFFSET(((crc & 0xFF) ^ k))];
+SET
+  i = i + 1;
+END WHILE;
```
relud: This can be done with fewer set operations per loop by calling `TO_CODE_POINTS` once on the whole input, which might improve performance, e.g.:
```sql
DECLARE code_points ARRAY<INT64>;
SET code_points = TO_CODE_POINTS(value);
WHILE
  len >= i
DO
  SET crc = (crc >> 8) ^ crc_table[OFFSET(((crc & 0xFF) ^ code_points[OFFSET(i)]))];
  SET i = i + 1;
END WHILE;
```

stored_procedures/safe_crc32_uuid.sql
relud: Could you add a another test for an input longer than 50 characters?

scholtzan: Added one with 60 characters

stored_procedures/safe_crc32_uuid.sql
scholtzan: It's a little faster but not significantly.

PR #692, opened 2020-01-21
Syntax error in publish_views for fenix

jklukas:
Follow up to #679


script/generate_views
```diff
@@ -164,7 +164,7 @@ def create_views_if_not_exist(client, views, exclude, sql_dir):
                 ) and table.table_id.startswith("metrics"):
                     replacements += [
                         "`moz-fx-data-shared-prod.udf.normalize_fenix_metrics`"
-                        "(client_info.telemetry_sdk_build, metrics.metrics)"
+                        "(client_info.telemetry_sdk_build, metrics)"
```
jklukas: Because the query here ends up referencing table `metrics_v1` rather than `metrics`, we don't have the naming ambiguity and it's not appropriate to use `metrics.metrics`.

PR #693, opened 2020-01-21
Correctly format scripting keywords

relud:
adds tests for `script/format_sql` as directories that contain `expect.sql`, and optionally `input.sql` if it differs from `expect.sql`

to avoid complex handling of `THEN`, `CASE` formatting has changed from
```
CASE
WHEN ... THEN ...
END
```
to 
```sql
WHEN
  ...
THEN
  ...
```


relud:
This also adds blank lines between statements in formatted files

sql/telemetry/fxa_users_daily_v1/query.sql
```diff
@@ -38,11 +36,9 @@ CREATE TEMP FUNCTION udf_contains_tier1_country(
       )
   )
 );
-  --
```
acmiyaguchi: This statement looks like it was generated by `prepend_udf_usage_definitions`.

sql/telemetry/fxa_users_daily_v1/query.sql
```diff
@@ -38,11 +36,9 @@ CREATE TEMP FUNCTION udf_contains_tier1_country(
...
-  -- This UDF is also only applicable in the context of this query.
-CREATE TEMP FUNCTION udf_contains_registration(
-  x ANY TYPE
-) AS ( --
```
acmiyaguchi: But was this this `--` also generated by the same function? The new formatting looks much better.

relud: i manually fixed up this formatting

sql/telemetry/fxa_users_daily_v1/query.sql
relud: yeah, it was

Issue #497, opened 2019-11-11
Please help add org_mozilla_vrbrowser.clients_last_seen schema


daoshengmu:
Firefox Reality project have enabled Glean baseline pings, so we would like to have `org_mozilla_vrbrowser.clients_last_seen` to query our DAU/WAU/MAU. Please help add it. Thanks!

jklukas:
This would be very similar logic to existing fenix clients_last_seen and clients_daily. There are, however, some fenix-specific metrics that are pulled in via the clients_daily query: https://github.com/mozilla/bigquery-etl/blob/master/templates/org_mozilla_fenix_derived/clients_daily_v1/query.sql#L62-L64

Are there are vrbrowser-specific metrics that you'd need to have pulled in for this? Or we could do just the info from baseline pings.

daoshengmu:
I think there is only only specific metric we need for vrbrowser, `metrics.string.distribution_channel_name`.

jklukas:
We're making some updates to how we do glean ETL in prep for Fenix launch, and I think we should pick up this org_mozilla_vrbrowser ETL while we're at it, so this should happen in coming weeks.

We'll basically want to duplicate `templates/org_mozilla_fenix_derived/clients_daily` and clients_last_seen, and modify to consider only the single `metrics.string.distribution_channel_name` metric. We also should change to use a LEFT JOIN from baseline pings to metrics pings rather than the union we currently use. Going forward, we should be considering a user active in a given day only if they sent a baseline ping. We do not want to count a user active based on a metrics ping alone.

jklukas:
Once we have the clients_last_seen view set up, we can add that view into `telemetry.nondesktop_clients_last_seen`, which will get it flowing to the smoot_* tables that power GUD.

scholtzan:
https://github.com/mozilla/bigquery-etl/pull/700 added `org_mozilla_vrbrowser.clients_last_seen`

PR #700, opened 2020-01-22
Add org_mozilla_vrbrowser.clients_last_seen and org_mozilla_vrbrowser…

scholtzan:
Addresses: #497

### DIFF
--- a/sql/org_mozilla_vrbrowser_derived/clients_daily_v1/query.sql
+++ b/sql/org_mozilla_vrbrowser_derived/clients_daily_v1/query.sql
@@ -1,169 +1,28 @@
-CREATE TEMP FUNCTION udf_geo_struct(
-  country STRING,
-  city STRING,
-  geo_subdivision1 STRING,
-  geo_subdivision2 STRING
-) AS ( IF(
-    country IS NULL
-    OR country = '??',
-    NULL,
-    STRUCT(
-      country,
-      NULLIF(city, '??') AS city,
-      NULLIF(geo_subdivision1, '??') AS geo_subdivision1,
-      NULLIF(geo_subdivision2, '??') AS geo_subdivision2
-    )
-  )
-);
-
-CREATE TEMP FUNCTION
-  udf_glean_timespan_seconds(timespan STRUCT<time_unit STRING, value INT64>)
-  RETURNS INT64 AS (
-    CAST(
-      FLOOR(
-        CASE timespan.time_unit
-          WHEN 'nanosecond' THEN timespan.value / 1000 / 1000 / 1000
-          WHEN 'microsecond' THEN timespan.value / 1000 / 1000
-          WHEN 'millisecond' THEN timespan.value / 1000
-          WHEN 'second' THEN timespan.value
-          WHEN 'minute' THEN timespan.value * 60
-          WHEN 'hour' THEN timespan.value * 60 * 60
-          WHEN 'day' THEN timespan.value * 60 * 60 * 24
-        END )
-      AS INT64));
-
-CREATE TEMP FUNCTION udf_json_mode_last(list ANY TYPE) AS (
-  (
-    SELECT
-      ANY_VALUE(_value)
-    FROM
-      UNNEST(list) AS _value
-    WITH OFFSET AS _offset
-    GROUP BY
-      TO_JSON_STRING(_value)
-    ORDER BY
-      COUNT(_value) DESC,
-      MAX(_offset) DESC
-    LIMIT
-      1
-  )
-);
-
-CREATE TEMP FUNCTION
-  udf_mode_last(list ANY TYPE) AS ((
-    SELECT
-      _value
-    FROM
-      UNNEST(list) AS _value
-    WITH
-    OFFSET
-      AS
-    _offset
-    GROUP BY
-      _value
-    ORDER BY
-      COUNT(_value) DESC,
-      MAX(_offset) DESC
-    LIMIT
-      1 ));
-
-WITH unioned AS (
-  SELECT
-    submission_timestamp,
-    document_id,
-    client_info,
-    sample_id,
-    metadata,
-    normalized_channel,
-    metrics AS baseline_metrics,
-    NULL AS metrics
-  FROM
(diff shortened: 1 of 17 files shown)


sql/org_mozilla_vrbrowser_derived/clients_daily_v1/query.sql
```diff
@@ -0,0 +1,169 @@
...
+    client_info.client_id IS NOT NULL
+),
+  --
+windowed AS (
+  SELECT
+    submission_date,
+    client_id,
+    sample_id,
+    ROW_NUMBER() OVER w1_unframed AS _n,
+    'Firefox Reality' AS app_name,
```
scholtzan: Is there a list with these names somewhere?

sql/org_mozilla_vrbrowser_derived/clients_daily_v1/query.sql
```diff
@@ -0,0 +1,169 @@
...
+    submission_timestamp,
+    document_id,
+    client_info,
+    sample_id,
+    metadata,
+    normalized_channel,
+    NULL AS baseline_metrics,
+    metrics
+  FROM
+    org_mozilla_vrbrowser_stable.metrics_v1
```
scholtzan: I looked into using `LEFT JOIN`, the best I could come up with but also doesn't quite work was:
```diff
   FROM
     org_mozilla_vrbrowser_stable.baseline_v1
   UNION ALL
+  -- Consider users only as active if they sent a baseline ping.
   SELECT
-    submission_timestamp,
-    document_id,
-    client_info,
-    sample_id,
-    metadata,
-    normalized_channel,
+    metrics.submission_timestamp,
+    metrics.document_id,
+    metrics.client_info,
+    metrics.sample_id,
+    metrics.metadata,
+    metrics.normalized_channel,
     NULL AS baseline_metrics,
-    metrics
+    metrics.metrics
   FROM
-    org_mozilla_vrbrowser_stable.metrics_v1
+    org_mozilla_vrbrowser_stable.baseline_v1 AS baseline
+  LEFT JOIN
+    org_mozilla_vrbrowser_stable.metrics_v1 AS metrics
+  ON
+    baseline.client_info.client_id = metrics.client_info.client_id
+  WHERE
+    baseline.client_info.client_id IS NOT NULL
 ),
   --
 base AS (
```

jklukas: Good point. I think we'll need to do separate aggregations on the baseline and metrics pings and then do the join between the two CTEs that each have only one row per client.

This also touches on a broader issue that we'd like to be able to support new glean applications automatically for GUD, etc. The baseline ping is standard for all apps and is intended as the source of being able to calculate retention, etc. If we need to aggregate baseline and metrics separately anyway to do the join here, maybe the best way forward here is to define a completely generic `baseline_daily` query and then individual apps can define custom `metrics_daily` queries that can be joined in as needed.

So, if it seems like doing separate aggregations and then joining is a straight-forward solution here, let's do that for now. Otherwise, we can table this for the moment and discuss the more general way forward for glean aggregations during Berlin and include @fbertsch in the discussion.

sql/org_mozilla_vrbrowser_derived/clients_daily_v1/query.sql
jklukas: There's a partial list in https://docs.telemetry.mozilla.org/concepts/choosing_a_dataset_mobile.html#products-overview but it's kind of a judgment call right now.

We should definitely have a more coherent and well-documented story for this as part of generalizing derived tables for Glean apps.

sql/org_mozilla_vrbrowser_derived/clients_daily_v1/query.sql
scholtzan: I we do separate aggregations would we simply use dimensions related to `client_info` from `baseline_daily`? That behaviour (and potentially event results?) would be different from using `udf_mode_last` on both baseline and metrics pings.

jklukas: We would used `udf_mode_last` over baseline pings for the fields that exist in baseline pings, and we would use `udf_mode_last` over metrics pings for any metrics-ping-specific fields.

Yes, that could lead to us choosing a different value for a dimension compared to using `udf_mode_last` over baseline and metrics pings. But it seems conceptually reasonable to me to accept the value we see from the baseline ping. Baseline pings become the main source of truth, and metrics pings are considered only for enriching the data with dimensions that exist as metrics.

scholtzan:
I added `baseline_daily` and `metrics_daily` and updated `clients_daily` and `clients_last_seen`. If that seems reasonable I'll also add `metadata.yaml` files for the queries.

scholtzan:
I added metadata files. I think the table descriptions could use some improvements

templates/org_mozilla_vrbrowser_derived/metrics_daily_v1/metadata.yaml
```diff
@@ -0,0 +1,9 @@
+friendly_name: VR Browser Metrics Daily
+description: >
+  A daily aggregate of metrics pings from each Firefox Reality client,
+  partitioned by day
+owners:
+  - ascholtz@mozilla.com
```
scholtzan: not sure if that should be me?

scholtzan:
> We'll need to run the queries manually to create the tables in order to get the dry-run test passing. Let me know if you'd like to coordinate on that.

I'd be doing that some time today. We could coordinate over zoom whenever you have time today.

templates/org_mozilla_vrbrowser_derived/metrics_daily_v1/metadata.yaml
jklukas: If you feel comfortable with being listed as owner, go for it. But feel free to mark `jklukas@mozilla.com` as the owner, or as an additional owner.

scholtzan: good idea, I added you as owner

PR #721, opened 2020-02-03
Add compressed smoot new_profiles table

jklukas:
Follow-up to #719


jklukas:
Needs to be scheduled in https://github.com/mozilla/telemetry-airflow/pull/848

Issue #706, opened 2020-01-23
Remove templates/


relud:
keeping `templates/` separate from `sql/` was initially a mitigation until persistent UDFs became available. We have a variety of options now for going back to just having `sql/`, including:
- use persistent UDFs
  - would need to replace persistent UDFs with temporary UDFs when running tests, so that UDF changes are properly tested
- modify `script/generate_sql` to work on files in-place
  - could obscure the difference between UDFs defined in `udf*/*.sql` and UDFs currently defined in `templates/*/query.sql`
- create a new script that prepends UDF definitions at run-time
  - could benefit from a bot to automatically comment on PRs with a list of queries impacted by modified UDFs, or something like that

jklukas:
I think we may want to build more support in bigquery-etl for generating queries/views such that maybe continuing to have a `templates` directory is a good thing for the long term. I'm imagining that `generate_sql` could evolve to run a series of modules. These could include:

- Inject temporary UDF definitions (as already exists)
- Run custom scripts for query generation (such as the existing `templates/telemetry/longitudinal.sql.py` and other python scripts that we rely on folks running manually)
- Generate daily and last_seen aggregate queries for all Glean products
- Generate Airflow DAGs based on `metadata.yaml` files

I'm seeing a lot of potential flexibility and value from code generation and I really like being able to see the generated results in PRs alongside changes to the templates.

relud:
discussed with Ben-Wu, jklukas, sunahsuh, and scholtzan in person and determined we should:

> - use persistent UDFs

Issue #319, opened 2019-08-28
Add a CI job to publish UDFs


jklukas:
We got part of the way on #272, but now we need to circle back and get this CI job built and reliable.

scholtzan:
Done in https://github.com/mozilla-services/cloudops-infra/commit/7d0cd4696c1c95bc258a77f575e304420879d195

PR #723, opened 2020-02-03
Remove templates/ and use persistent UDFs

scholtzan:
This addresses https://github.com/mozilla/bigquery-etl/issues/706 and https://github.com/mozilla/bigquery-etl/issues/319

I'm not sure if the CI job for publishing the persistent UDFs will actually run since it'll need GCP credentials.

This PR consists of quite a few changes, although most of those are formatting changes or file deletions. If it somehow makes sense I can try to split this up into smaller pieces to make reviewing this easier.

I'll update documentation in a follow-up PR and try changing `main_summary` to use persistent UDFs.


bigquery_etl/parse_udf.py
```diff
@@ -34,32 +34,54 @@ class RawUdf:
...
+            text = f.read()
+
+        sql = sqlparse.format(text, strip_comments=True)
+        statements = [s for s in sqlparse.split(sql) if s.strip()]
+
         prod_name = basename.replace(".sql", "")
-        internal_name = os.path.basename(dirpath) + "_" + prod_name
+        persistent_name = os.path.basename(dirpath) + "." + prod_name
+        temp_name = os.path.basename(dirpath) + "_" + prod_name
+        internal_name = None
```
scholtzan: This is to both support temporary and persistent UDFs. (Temporary UDFs, such as `assert_false`, are for example still used for tests)

sql/monitoring/structured_detailed_error_counts_v1/view.sql
```diff
@@ -9,7 +9,7 @@ WITH error_examples AS (
     document_version,
     error_type,
     error_message,
-    `moz-fx-data-shared-prod.udf_js.gunzip`(ANY_VALUE(payload)) AS sample_payload,
+    `moz-fx-data-shared-prod.udf.js.gunzip`(ANY_VALUE(payload)) AS sample_payload,
```
scholtzan: I thought it'd be best to use `<udf|udf_js>.<udf_name>` naming to distinguish persistent UDFs from temporary UDFs which are named something like `udf_<udf_name>`. However, it's easy to go back to the previous naming if that's better.

scholtzan: Most of the changes that follow are renaming UDFs, changes from running the formatting script and deletions due to removing `templates/`

sql/telemetry_derived/main_summary_v4/part1.sql
```diff
@@ -125,13 +125,11 @@ function ifnull(value1, value2) {
   }
   return value2;
 }
-
```
scholtzan: Using persistent UDFs in `part1` and `part2` of `main_summary` resulted in failed tests because the UDFs could not be found. I'll look into this later, for now all the UDFs used in those 2 queries are temporary UDFs.

.circleci/config.yml
```diff
@@ -25,29 +25,27 @@ jobs:
...
-            rm -rf sql/
-            venv/bin/python script/generate_sql
-            echo "Inconsistencies between templates and generated SQL:"
-            git ls-files --other --modified -x sql/*
-            test `git ls-files --other --modified -x sql/* | wc -l` = 0
-  verify-format-sql:
+          name: Verify that SQL is correctly formatted
+          # check all directories owned by git except sql
+          command: script/format_sql --check $(git ls-tree -d HEAD --name-only | grep -vx sql)
+  publish-persistent-udfs:
```
scholtzan: This is failing, not sure what the best way for this is.

jklukas: We handle view publishing as part of the Jenkins job that does schema deploys (see https://github.com/mozilla-services/cloudops-infra/blob/9a7b3eb1b2ee7a3f068a7bb3cda626cc19dd1f72/projects/data-shared/Jenkinsfile.bigquery.prod#L100-L110).

I think we should publish UDFs at that same point, right before we publish views.

We'll still have the problem that PR authors will likely need to manually publish changes in order to get the dry run CI task passing, which is currently the case with PRs that add new queries.

bigquery_etl/parse_udf.py
```diff
@@ -34,32 +34,54 @@ class RawUdf:
     def from_file(filepath):
         """Read in a RawUdf from a SQL file on disk."""
         dirpath, basename = os.path.split(filepath)
+
+        with open(filepath) as f:
+            text = f.read()
+
+        sql = sqlparse.format(text, strip_comments=True)
+        statements = [s for s in sqlparse.split(sql) if s.strip()]
```
jklukas: This may still be vulnerable to unexpected newlines or multiple spaces in the source. In publish_views, for example, we have to do the following in order to catch edge cases:

    " ".join(tokens[0].normalized.split()) == "CREATE OR REPLACE"

sql/monitoring/structured_detailed_error_counts_v1/view.sql
jklukas: I'm not understanding this comment. `moz-fx-data-shared-prod.udf.js.gunzip` looks like an invalid name containing 3 periods rather than 4. Is this a typo?

scholtzan: Yeah, that was a typo

scholtzan:
I created a PR for adding publishing persistent UDFs in Jenkins: https://github.com/mozilla-services/cloudops-infra/pull/1799

script/publish_persistent_udfs
```diff
@@ -33,14 +33,10 @@ parser.add_argument(
...
-    default="udf",
-    help="The name of the dataset the persistent UDFs will be stored in.",
-)
-parser.add_argument(
-    "--udf-dir",
-    default="udf/",
-    help="The directory where declarations of temporary UDFs are stored.",
+    "--udf-dirs",
+    nargs="+",
+    default=["udf/", "udf_js/"],
```
scholtzan: It might be handy to specify multiple directories with UDFs at once.

jklukas: Do we lose the ability to specify the target dataset? By making this change, we assume that the dataset is named the same as the directory?

I might be okay with that, but we should update the help string to make this more clear. I do like the idea that running the script without arguments handles everything we want to do during schema deploys.

scholtzan: Yeah, the target datasets would be the same as the names of those directories. I updated the help string to make this clearer.

script/publish_persistent_udfs
```diff
@@ -33,14 +33,14 @@ parser.add_argument(
...
+    help="""
+    The directories where declarations of UDFs are stored. 
+    The names of these directories serve also specify the target dataset
```
jklukas:
```suggestion
    The names of these directories serve also to specify the target dataset
```

script/publish_persistent_udfs
```diff
@@ -33,14 +33,14 @@ parser.add_argument(
...
-    default="udf/",
-    help="The directory where declarations of temporary UDFs are stored.",
+    "--udf-dirs",
+    nargs="+",
+    default=["udf/", "udf_js/"],
+    help="""
+    The directories where declarations of UDFs are stored.
+    The names of these directories serve also to specify the target datasets
+    the UDFs get published in.
+    """,
```
jklukas: When help needs to span multiple lines, we've used style like the below elsewhere:

```
        help=(
            "If specified, create views in the target project rather than"
            " the project specified in the file"
        ),
```

That explicitly controls whitespace. It's not clear to me that argparse is going to correctly strip whitespace from a multiline string literal.

sql/telemetry_derived/main_summary_v4/part2.sql
```diff
@@ -1,17 +1,3 @@
...
-        SAFE_CAST(SPLIT(entry, ':')[OFFSET(1)] AS INT64) AS value
-      )
-    FROM
-      UNNEST(SPLIT(REPLACE(TRIM(input, '{}'), '"', ''), ',')) AS entry
-    WHERE
-      LENGTH(entry) > 0
-  )
-);
-
 CREATE TEMP FUNCTION udf_main_summary_scalars(processes ANY TYPE) AS (
```
scholtzan: What is the reason for skipping `udf/main_summary_scalars.sql` in the `publish_persistent_udf` script? Or can that UDF just get published and then having this temporary function here could get removed?

scholtzan:
Should be ready for another look.

sql/telemetry_derived/main_summary_v4/part2.sql
relud: I believe the function was too large to publish

bigquery_etl/parse_udf.py
```diff
@@ -34,32 +34,55 @@ class RawUdf:
...
```
relud: `if s.strip()` doesn't do anything here
```suggestion
        statements = [s for s in sqlparse.split(sql)]
```

bigquery_etl/parse_udf.py
```diff
@@ -34,32 +34,55 @@ class RawUdf:
...
+            normalized_statement = " ".join(s.lower().split())
+            if normalized_statement.startswith("create or replace function"):
+                definitions.append(s)
+                if persistent_name in normalized_statement:
+                    internal_name = persistent_name
+
+            elif normalized_statement.startswith("create temp function"):
+                definitions.append(s)
+                if temp_name in normalized_statement:
+                    internal_name = temp_name
```
relud: this should use an else block to populate `tests`, so that `"create or replace function"` and `"create temp function"` don't need to be repeated below.
```suggestion
                    internal_name = temp_name
            else:
                tests.append(s)
```

bigquery_etl/parse_udf.py
```diff
@@ -34,32 +34,55 @@ class RawUdf:
...
 
         for name in (prod_name, internal_name):
+            print(name)
```
relud: did you mean to remove this?
```suggestion
```

bigquery_etl/parse_udf.py
```diff
@@ -34,32 +34,55 @@ class RawUdf:
...
+        # find usages of both persistent and temporary UDFs
+        dependencies = re.findall(PERSISTENT_UDF_RE, "\n".join(definitions))
+        dependencies = list(map(lambda t: ".".join(t), dependencies))
+        dependencies.extend(re.findall(TEMP_UDF_RE, "\n".join(definitions)))
+
         if internal_name not in dependencies:
             raise ValueError(
-                f"Expected a temporary UDF named {internal_name} "
-                f"to be defined in {filepath}"
+                f"Expected a UDF named {internal_name} " f"to be defined in {filepath}"
```
relud: The only way for `internal_name` to not be in dependencies is if it is `None`, so this should reference `persistent_name` and `temporary_name`:
```suggestion
                f"Expected a UDF named {persistent_name} or {temporary_name} "
                f"to be defined in {filepath}"
```

bigquery_etl/parse_udf.py
```diff
@@ -34,32 +34,55 @@ class RawUdf:
...
+            and not s.lower().startswith("create temp function")
         ]
-        dependencies = re.findall(UDF_RE, "\n".join(definitions))
+
+        # find usages of both persistent and temporary UDFs
+        dependencies = re.findall(PERSISTENT_UDF_RE, "\n".join(definitions))
+        dependencies = list(map(lambda t: ".".join(t), dependencies))
+        dependencies.extend(re.findall(TEMP_UDF_RE, "\n".join(definitions)))
+
         if internal_name not in dependencies:
```
relud:
```suggestion
        if internal_name is None:
```

bigquery_etl/parse_udf.py
```diff
@@ -34,32 +34,55 @@ class RawUdf:
...
+            s
+            for s in statements
+            if not s.lower().startswith("create or replace function")
+            and not s.lower().startswith("create temp function")
         ]
-        dependencies = re.findall(UDF_RE, "\n".join(definitions))
+
+        # find usages of both persistent and temporary UDFs
+        dependencies = re.findall(PERSISTENT_UDF_RE, "\n".join(definitions))
+        dependencies = list(map(lambda t: ".".join(t), dependencies))
```
relud: I think this is more readable as a list comprehension:
```suggestion
        dependencies = [".".join(t) for t in dependencies]
```

bigquery_etl/parse_udf.py
```diff
@@ -155,12 +178,32 @@ def udf_usage_definitions(text, raw_udfs=None):
...
+    Persistent UDFs need to be rewritten as temporary UDF so that changes
+    can be tested.
+    """
+    tests_full_sql = []
+    for test in raw_udf.tests:
+        test_sql = prepend_udf_usage_definitions(test, raw_udfs)
+        test_sql = sub_persistent_udfs_as_temp(test_sql)
+        test_sql = test_sql.replace(
+            "CREATE OR REPLACE FUNCTION", "CREATE TEMP FUNCTION"
+        )
```
relud: this should be case insensitive, match any whitespace, and cover all variations of persistent udf definitions (`CREATE FUNCTION`, `CREATE OR REPLACE FUNCTION`, `CREATE FUNCTION IF NOT EXISTS`).

for example:
```python
UDF_DEFINITION_KEYWORD = re.compile(r"CREATE\s+(OR\s+REPLACE\s+)?FUNCTION(\s+IF\s+NOT\s+EXISTS)?", re.IGNORECASE)

...

        test_sql = UDF_DEFINITION_KEYWORD.sub("CREATE TEMP FUNCTION", test_sql)
```

bigquery_etl/pytest_plugin/sql.py
```diff
@@ -107,11 +107,12 @@ def runtest(self):
...
 
         # rewrite all udfs as temporary
-        temp_udfs = parse_udf.sub_persisent_udfs_as_temp(query)
+        temp_udfs = parse_udf.sub_persistent_udfs_as_temp(query)
         if temp_udfs != query:
-            query = temp_udfs
             # prepend udf definitions
             query = parse_udf.prepend_udf_usage_definitions(query)
+            query = parse_udf.sub_persistent_udfs_as_temp(query)
+            query = query.replace("CREATE OR REPLACE FUNCTION", "CREATE TEMP FUNCTION")
```
relud: we should extract this logic to a function that we also call from `udf_tests_sql`, rather than duplicating it

script/format_sql
```diff
@@ -16,144 +16,6 @@ from bigquery_etl.format_sql.formatter import reformat  # noqa E402
...
-    "templates/telemetry_derived/histogram_percentiles_v1/query.sql",
-    "templates/telemetry_derived/latest_versions/query.sql",
-    "templates/telemetry_derived/main_events_v1/init.sql",
-    "templates/telemetry_derived/main_events_v1/query.sql",
-    "templates/telemetry_derived/scalar_percentiles_v1/query.sql",
-    "templates/telemetry_derived/smoot_usage_desktop_v2/query.sql",
-    "templates/telemetry_derived/smoot_usage_fxa_v2/query.sql",
-    "templates/telemetry_derived/smoot_usage_new_profiles_v2/query.sql",
-    "templates/telemetry_derived/smoot_usage_nondesktop_v2/query.sql",
-    "templates/telemetry_derived/surveygizmo_daily_attitudes/init.sql",
```
relud: I think we should either `s/templates/sql/` here, or remove these exceptions in a separate PR, rather include formatting changes in this PR.

sql/org_mozilla_fenix/clients_last_seen/view.sql
```diff
@@ -3,7 +3,7 @@ CREATE OR REPLACE VIEW
 AS
 WITH with_days_since AS (
   SELECT
-    -- We cannot use UDFs in a view, so we paste the body of udf_bitpos(bits) literally here.
+    -- We cannot use UDFs in a view, so we paste the body of udf.bitpos(bits) literally here.
```
relud: we actually can, but `udf_bitpos` seems to have disappeared prior to this PR.

relud: cc @jklukas

jklukas: Indeed. We renamed a number of these bit and byte manipulation functions. `bitpos` is now `pos_of_trailing_set_bit`.

script/format_sql
scholtzan: I changed it to ignoring the SQL files instead and did undo all the previous formatting of those files for this PR.

scholtzan:
Some tests related to UDF tests are failing, I'll look into those tomorrow.

PR #724, opened 2020-02-04
Bug 1610983 - Add clients daily scalar aggregates for GLAM in Fenix

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1610983)

This PR modifies the `clients_daily_scalar_aggregates_v1.sql.py` script for use with the Glean tables for `org-mozilla-fenix`. This only supports a subset of the metrics, but hopefully it'll be useful enough to feed into the next step of the pipeline.

This can be used on any of the pings sent by Glean, so I've written two scripts: one generates the relevant sql, and the other runs the job into a single table. 

The [glam-fenix-dev project](https://console.cloud.google.com/bigquery?project=glam-fenix-dev&organizationId=442341870013&p=glam-fenix-dev&d=fenix_scalar_test&page=dataset) has the `glam_etl_fenix_test` dataset with the relevant table.

The easiest way to test this out is to run the sql generation script.

```bash
$ python3 -m bigquery_etl.glam.glean_scalar_aggregates \
    --agg-type keyed_scalars \
    --table-id org_mozilla_fenix_stable.activation_v1 \
    | pbcopy

# paste into bigquery console
```

If you have access to the `glam-fenix-dev` project, then run the following commands:

```bash
$ script/glam/generate_fenix_sql
$ script/glam/run_fenix_sql
```


acmiyaguchi:
@emtwo This is a good spot for initial review. I still need to figure out some of the details, but it works for the most part.

acmiyaguchi:
I'm familiar enough to incorporate here, but I'm cautious to use it as the source of truth since there is some lag until metrics in the probe info service propagate into the BigQuery schemas. There doesn't seem to be anything too tricky outside of navigating the api to get all the relevant information.

The [mozilla-schema-generator](https://github.com/mozilla/mozilla-schema-generator/blob/master/mozilla_schema_generator/glean_ping.py) takes care of generating schemas with all of the calculated products and dependencies. The normalized `.[].app_id` key in the `/glean/repositories` endpoint maps the Glean product to a namespace in the schema repository.

Is there an example of information beyond the `metric_type` that would need to be taken into account?

PR #725, opened 2020-02-04
Parallelize scripts/publish_views

jklukas:
Per request of @whd.

### DIFF
--- a/script/publish_views
+++ b/script/publish_views
@@ -95,18 +95,18 @@ def main():
         parser.error(f"argument --log-level: {e}")
 
     worker_entrypoint = partial(process_file, client, args)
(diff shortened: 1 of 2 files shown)


script/publish_views
```diff
@@ -84,13 +94,24 @@ def main():
     except ValueError as e:
         parser.error(f"argument --log-level: {e}")
 
+    worker_entrypoint = partial(process_file, client, args)
+    sqlfiles = []
```
scholtzan: Maybe `sql_files`?

jklukas: This is matching the naming used in `dryrun`. I will update the name in both places.

PR #727, opened 2020-02-07
Add template for clients_scalar_aggregates

acmiyaguchi:
I thought it might be useful to play with templates because there is a lot of repetition going on in the queries. The gleanified versions of these queries are only going to be moderately different. For example, there is no `process` in the `user_data_attributes`, but there is a `ping_type` in the `attributes`. 

This is tested using the following command, after using `scripts/format_sql`.

```bash
% diff <(python3 -m bigquery_etl.glam.scalar_aggregates_incremental) <(cat templates/telemetry_derived/clients_scalar_aggregates_v1/query.sql)
1d0
< -- generated by: python3 -m bigquery_etl.glam.scalar_aggregates_incremental
```


bigquery_etl/glam/templates/clients_scalar_aggregates_v1.sql
```diff
@@ -0,0 +1,118 @@
...
+  FROM
+    filtered_new AS new_data
+  FULL OUTER JOIN
+    filtered_old AS old_data
+  ON
+    new_data.client_id = old_data.client_id
+    AND new_data.os = old_data.os
+    AND CAST(new_data.app_version AS INT64) = old_data.app_version
+    AND new_data.app_build_id = old_data.app_build_id
+    AND new_data.channel = old_data.channel
```
acmiyaguchi: This section can probably with jinja and just the `attributes` data.

acmiyaguchi:
I've added in the sections needed to make this generic, and the resulting diff looks like:

```diff
% diff <(cat templates/telemetry_derived/clients_scalar_aggregates_v1/query.sql) <(python3 -m bigquery_etl.glam.scalar_aggregates_incremental) 
125c125,126
<     *
---
>     * EXCEPT (app_version),
>     CAST(app_version, INT64) AS app_version
155,159c156,160
<     client_id,
<     os,
<     app_version,
<     app_build_id,
<     scalar_aggs.channel AS channel,
---
>     scalar_aggs.client_id,
>     scalar_aggs.os,
>     scalar_aggs.app_version,
>     scalar_aggs.app_build_id,
>     scalar_aggs.channel,
171c172
<     latest_versions.channel = scalar_aggs.channel
---
>     (channel)
173c174
<     CAST(app_version AS INT64) >= (latest_version - 2)
---
>     app_version >= (latest_version - 2)
230,234c231,235
<     client_id,
<     os,
<     app_version,
<     app_build_id,
<     scalar_aggs.channel AS channel,
---
>     scalar_aggs.client_id,
>     scalar_aggs.os,
>     scalar_aggs.app_version,
>     scalar_aggs.app_build_id,
>     scalar_aggs.channel,
241c242
<     latest_versions.channel = scalar_aggs.channel
---
>     (channel)
249c250
<     COALESCE(old_data.app_version, CAST(new_data.app_version AS INT64)) AS app_version,
---
>     COALESCE(old_data.app_version, new_data.app_version) AS app_version,
259,263c260
<     new_data.client_id = old_data.client_id
<     AND new_data.os = old_data.os
<     AND CAST(new_data.app_version AS INT64) = old_data.app_version
<     AND new_data.app_build_id = old_data.app_build_id
<     AND new_data.channel = old_data.channel
---
>     (client_id, os, app_version, app_build_id, channel)
```

I'm actually not sure if this PR would make it less or more difficult to maintain this script. Let me know what your thoughts are. I'm not too attached to this code (and understand it quite a bit more now), and I can do something simpler that's glean specific.

scholtzan:
FYI: https://github.com/mozilla/bigquery-etl/pull/723 just got merged, so `templates/` got removed and persistent UDFs are the way to go from now on.

acmiyaguchi:
> FYI: #723 just got merged, so `templates/` got removed and persistent UDFs are the way to go from now on.

Thanks for the heads up, I'll make sure to generate directly into the `sql` directory.

README.md
```diff
@@ -186,7 +186,25 @@ Contributing
...
+# Add the dependency to constraints.txt e.g. jinja2. The purpose of constraints.txt
+# is to avoid cluttering the real dependencies requirements.txt.
+hashin jinja2==2.11.1 -r constraints.txt
+
+# Compile constraints to obtain hashes for dependencies
+pip-compile --generate-hashes constraints.txt
+
+# Finally ass dependency to requirements.txt
+hashin jinja2==2.11.1
+```
```
acmiyaguchi: @relud Does this piece of documentation sound?

bigquery_etl/glam/templates/clients_scalar_aggregates_v1.sql
```diff
@@ -0,0 +1,118 @@
...
+
+WITH filtered_date_channel AS (
+  SELECT
+    *
+  FROM
+    clients_daily_scalar_aggregates_v1
+  WHERE
+    submission_date = @submission_date
+),
+filtered_aggregates AS (
```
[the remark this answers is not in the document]
acmiyaguchi: Okay, if this is based on experience. I've heard one way or the other that subqueries can improve or degrade performance of the query planner.  In theory though, it shouldn't matter.

acmiyaguchi:
> My main concern is that there is divergence between desktop and glean/fenix data processing and it might not always make sense for them to share templates. For example, right now, `clients_histogram_aggregates` will completely change for desktop to address the shuffle failures. But this change might be overkill for fenix/glean data.

This is a fair point, because not every query should be shared between the two ping types. However, my goal is to reuse the existing work that you have done, with some refactoring. For `clients_scalar_aggregates`, it looks like almost all of it (from the udf to the overall logic and data processing flow) can be reused for the Fenix data by changing the variables defined in this PR like `attributes` and `user_data_type` as per https://github.com/mozilla/bigquery-etl/pull/727#issue-372182029.

> What do you think about starting with the templates for glean and then later applying them to desktop if they fit? If they don't fit, then desktop will need its own templates.

I think the approach of starting from a copy of the desktop version and refactoring for Fenix has been working well. This refactoring might be unique to this query, but I'll make sure to change my approach if it doesn't fit.

README.md
relud: unfortunately no. The correct steps are:
```bash
# create and activate a python virtual environment
python3 -m venv venv/
. venv/bin/activate

# If not installed:
pip install hashin

# Add the dependency to requirements.txt e.g. Jinja2.
hashin Jinja2==2.11.1

# Use pip install to check for missing meta-dependencies
pip install --require-hashes -r requirements.txt

# Add hashes for missing meta-dependencies until pip install succeeds
hashin MarkupSafe -r constraints.txt

# deactivate the python virtual environment
deactivate
```

note that requirements go in *either* `constraints.txt` or `requirements.txt`, but not both.

constraints.txt
```diff
@@ -1,45 +1,252 @@
...
-zipp==2.1.0 --hash=sha256:ccc94ed0909b58ffe34430ea5451f07bc0c76467d7081619a454bf5c98b89e28 --hash=sha256:feae2f18633c32fc71f2de629bfb3bd3c9325cd4419642b1f1da42ee488d9b98
+#
+# This file is autogenerated by pip-compile
+# To update, run:
+#
+#    pip-compile --generate-hashes constraints.txt
+#
+apipkg==1.5 \
+    --hash=sha256:37228cda29411948b422fae072f57e31d3396d2ee1c9783775980ee9c9990af6 \
+    --hash=sha256:58587dd4dc3daefad0487f6d9ae32b4542b185e1c36db6993290e7c41ca2b47c
```
relud: this is cool, and we should probably switch to using `pip-compile` in a separate PR, but for now `constraints.txt` should not be rewritten to add one dependency.

README.md
acmiyaguchi: Okay, thanks. It looks like this command was what I was missing when running into missing transitive dependencies:

```
# Use pip install to check for missing meta-dependencies
pip install --require-hashes -r requirements.txt
```

relud: filed #736 to start using `pip-compile`, so that this is easier in the future.

PR #730, opened 2020-02-10
Add Jinja-based templating to `format_sql`

acmiyaguchi:
Jinja is a templating language for Python that has useful constructs for building templated SQL. I've been using it in #727, which has let me be expressive without excessive repetition. We use also use templates in `telemetry-airflow`/

In this PR, I propose adopting two constructs from Jinja: [expressions and statements](https://jinja.palletsprojects.com/en/2.11.x/templates/). An expression is surrounded by double braces `{{ ... }}`. A statement is a control flow operator, that is denoted by `{% ... %}` start and end markers.

It would be useful to format SQL that uses templates by treating expressions as an identifier, and by treating statements sections that cannot be formatted. 

The `template_expression_example` is implemented, but I haven't implemented `template_statement_example`.


acmiyaguchi:
@relud f?

Do you have particular thoughts about adding support for formatting outside of BigQuery syntax? This would be a useful extension to have for some of the code I've been adding to `bigquery_etl.glam`, where I use `bigquery_etl.format_sql.reformat` as the last step for generating sql. 

Do you also have a sense for whether statements could be done as a regex, or if it requires something more complex? There are various statements such as `include`, `for/endfor`, and `if/endif`.

relud:
> Do you have particular thoughts about adding support for formatting outside of BigQuery syntax?

I think it should stick to SQL. I would be difficult to format the jinja template in a way that would be readable and also result in a template that outputs correctly formatted SQL.

> This would be a useful extension to have for some of the code I've been adding to `bigquery_etl.glam`, where I use `bigquery_etl.format_sql.reformat` as the last step for generating sql.

The format_sql should be applied after the template is compiled, so I don't understand how jinja support would help this?

> Do you also have a sense for whether statements could be done as a regex, or if it requires something more complex?

I don't think jinja is a regular language and I think it would be better to determine if the jinja library can be used to tokenize a template, rather than rolling our own tokenizer for that.

acmiyaguchi:
> The format_sql should be applied after the template is compiled, so I don't understand how jinja support would help this?

I do run `bigquery_etl.format_sql.reformat` in the relevant `__main__` sections, so this applies only to the templated sql. It's been *very* handy. 

The `bigquery_etl/glam/templates/*.sql` files need to be added to the `SKIP` section in `script/format_sql`. Adding support for ignoring or properly handling template sections so they remain valid after as pass through the formatter and formatted in sections when applicable e.g. https://github.com/mozilla/bigquery-etl/blob/437b34d2cf87a3d1c1c88ac668d3f457c09d2de3/bigquery_etl/glam/templates/udf_merged_user_data.sql#L1-L7 would be a nice to have.

> > Do you also have a sense for whether statements could be done as a regex, or if it requires something more complex?
> 
> I don't think jinja is a regular language and I think it would be better to determine if the jinja library can be used to tokenize a template, rather than rolling our own tokenizer for that.

I've been thinking about a simple tokenizer that would help ignore the statement blocks. I'm thinking something like the following after writing the tests mulling over the problem. The goal is to effectively blocks wrapped in `{% .. %}` into `--format:{off|on}` sections. Thinking out loud, I think it would be something like:

```python
# TemplateStatement

statement_expression = ["include", ...]
statement_block = [("for", "endfor"), ("if", "endif"), ...]

expression_pattern = r"..."  # for each statement_expression
block_pattern  = "|".join([
    # but it would have to handle nested statements in balanced way...
    fr"{%.*{start_block}.*%}.*{%.*{end_block}.*%}"
    for start_block, end_block in statement_block
])

pattern = re.compile(fr"{expression_pattern}|{block_pattern}")
```

Jinja does provide a low level api for its [lexer](https://jinja.palletsprojects.com/en/2.11.x/api/#jinja2.Environment.lex) and [parser](https://jinja.palletsprojects.com/en/2.11.x/api/#jinja2.Environment.parse), but it seems like a fair amount of grokking to get a sense of how it would fit together.

relud:
in that case it seems like adding jinja's `{{}}` and `{% %}` blocks as comment-types would be a good path forward, and that seems like a fine fit for supporting in `format_sql`. As long as we don't do anything too fancy in jinja, e.g. escaping jinja brackets: `{{ "{{" }} "meta_template!" {{ "}}" }}`, then we should be able to do it with regexes.

tests/format_sql/template_statement_example/expect.sql
```diff
@@ -0,0 +1,13 @@
+{% include "udf.sql" %}
+SELECT
+  {% for column in column_list %}
+    {{ column }},
```
robhudson: This would leave a trailing comma which wouldn't be valid SQL, correct?

You could use Jinja's `join` filter: `{{ column_list|join(", ") }}`.

acmiyaguchi: This is true, although this is a (semi-) contrived example for unit testing.

relud: Also: BigQuery allows trailing commas in the `SELECT` section

bigquery_etl/format_sql/tokenizer.py
```diff
@@ -433,6 +457,8 @@ class FieldAccessOperator(Operator):
     AngleBracketKeyword,
     SpaceBeforeBracketKeyword,
     ReservedKeyword,
+    TemplateStatement,
+    TemplateExpression,
```
relud:
```suggestion
    JinjaExpression,
```

bigquery_etl/format_sql/tokenizer.py
```diff
@@ -433,6 +457,8 @@ class FieldAccessOperator(Operator):
     AngleBracketKeyword,
     SpaceBeforeBracketKeyword,
     ReservedKeyword,
+    TemplateStatement,
```
relud:
```suggestion
    JinjaStatementStart,
    JinjaStatementEnd,
    JinjaStatement,
```

bigquery_etl/format_sql/formatter.py
```diff
@@ -20,6 +20,8 @@
     ReservedKeyword,
     SpaceBeforeBracketKeyword,
     StatementSeparator,
+    TemplateStatement,
```
relud:
```suggestion
```

bigquery_etl/format_sql/formatter.py
```diff
@@ -20,6 +20,8 @@
...
+    TemplateExpression,
```
relud:
```suggestion
```

acmiyaguchi:
@relud Thanks for the suggestions! This almost does the right thing with some minor adjustments, but it looks like the `BlockKeyword` concept is tied to `ReservedKeyword`, which results in the following diff:

```diff
E       {% include "udf.sql" %}
E       SELECT
E     -   {% for column in column_list %}
E     +   {% FOR COLUMN IN COLUMN_LIST %}
E           {{ column }},
E     -   {% endfor %}
E     +   {% ENDFOR %}
E       FROM
E         dataset.table
E       WHERE
E     -   {% if condition %}
E     +   {% IF CONDITION %}
E           {{ column }} = {{ value * 2 }}
E     -   {% else %}
E     +   {% ELSE %}
E           {{ column }} = {{ value * 3 }}
E     -   {% endif %}
E     +   {% ENDIF %}
```

I think some minor refactoring could decouple the concept of blocking and reserved keywords.

bigquery_etl/format_sql/formatter.py
```diff
@@ -5,6 +5,9 @@
 
 from .tokenizer import (
     AliasSeparator,
+    BlockToken,
+    BlockStart,
+    BlockEnd,
     BlockKeyword,
```
relud:
```suggestion
```

bigquery_etl/format_sql/formatter.py
```diff
@@ -5,6 +5,9 @@
...
     BlockStartKeyword,
```
relud:
```suggestion
```

bigquery_etl/format_sql/formatter.py
```diff
@@ -5,6 +5,9 @@
...
     BlockEndKeyword,
```
relud:
```suggestion
```

bigquery_etl/format_sql/tokenizer.py
```diff
@@ -420,6 +436,50 @@ class FieldAccessOperator(Operator):
...
+class JinjaStatementMiddle(BlockMiddle):
+    """Template control flow.
+
+        -- Jinja2, trailing comma
+        SELECT
+            {% for column in column_list %}
+                {{ column }} AS prefixed_{{ column }},
+            {% endfor %}
+        FROM
+            {{ table }}
```
relud: these 7 lines need to be unindented one level

acmiyaguchi:
For kicks, I put together a script to visualize the inheritance tree of tokens (which is to say, the `Block` classes I introduced make things complicated).

```python
#!/usr/bin/env python3
data = []
with open("bigquery_etl/format_sql/tokenizer.py") as fp:
    for line in fp.readlines():
        if "class" in line and ":" in line:
            data.append(line)

print("graph LR")
lines = []
for line in data:
    name = line.lstrip("class ").strip().rstrip(":")
    if "(" in name:
        name, deps = name.rstrip(")").split("(")
        for dep in deps.split(","):
            lines.append(f"{dep.strip()} --> {name}")
    else:
        lines.append(name)

for line in sorted(lines):
    print(line)

```

Before:

[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiZ3JhcGggTFJcbkJsb2NrRW5kS2V5d29yZCAtLT4gQmxvY2tNaWRkbGVLZXl3b3JkXG5CbG9ja0tleXdvcmQgLS0-IEJsb2NrRW5kS2V5d29yZFxuQmxvY2tLZXl3b3JkIC0tPiBCbG9ja1N0YXJ0S2V5d29yZFxuQmxvY2tTdGFydEtleXdvcmQgLS0-IEJsb2NrTWlkZGxlS2V5d29yZFxuQ29tbWVudCAtLT4gQmxvY2tDb21tZW50XG5Db21tZW50IC0tPiBMaW5lQ29tbWVudFxuTmV3bGluZUtleXdvcmQgLS0-IFRvcExldmVsS2V5d29yZFxuT3BlcmF0b3IgLS0-IEZpZWxkQWNjZXNzT3BlcmF0b3JcblJlc2VydmVkS2V5d29yZCAtLT4gQW5nbGVCcmFja2V0S2V5d29yZFxuUmVzZXJ2ZWRLZXl3b3JkIC0tPiBCbG9ja0tleXdvcmRcblJlc2VydmVkS2V5d29yZCAtLT4gU3BhY2VCZWZvcmVCcmFja2V0S2V5d29yZFxuU3BhY2VCZWZvcmVCcmFja2V0S2V5d29yZCAtLT4gQWxpYXNTZXBhcmF0b3JcblNwYWNlQmVmb3JlQnJhY2tldEtleXdvcmQgLS0-IE5ld2xpbmVLZXl3b3JkXG5Ub2tlblxuVG9rZW4gLS0-IENsb3NpbmdCcmFja2V0XG5Ub2tlbiAtLT4gQ29tbWVudFxuVG9rZW4gLS0-IEV4cHJlc3Npb25TZXBhcmF0b3JcblRva2VuIC0tPiBJZGVudGlmaWVyXG5Ub2tlbiAtLT4gTGl0ZXJhbFxuVG9rZW4gLS0-IE1heWJlQ2xvc2luZ0FuZ2xlQnJhY2tldFxuVG9rZW4gLS0-IE1heWJlT3BlbmluZ0FuZ2xlQnJhY2tldFxuVG9rZW4gLS0-IE9wZW5pbmdCcmFja2V0XG5Ub2tlbiAtLT4gT3BlcmF0b3JcblRva2VuIC0tPiBSZXNlcnZlZEtleXdvcmRcblRva2VuIC0tPiBTdGF0ZW1lbnRTZXBhcmF0b3JcblRva2VuIC0tPiBXaGl0ZXNwYWNlIiwibWVybWFpZCI6eyJ0aGVtZSI6ImRlZmF1bHQifSwidXBkYXRlRWRpdG9yIjpmYWxzZX0)

After:

[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiZ3JhcGggTFJcbkJsb2NrRW5kIC0tPiBCbG9ja0VuZEtleXdvcmRcbkJsb2NrRW5kIC0tPiBKaW5qYVN0YXRlbWVudEVuZFxuQmxvY2tFbmRLZXl3b3JkIC0tPiBCbG9ja01pZGRsZUtleXdvcmRcbkJsb2NrU3RhcnQgLS0-IEJsb2NrU3RhcnRLZXl3b3JkXG5CbG9ja1N0YXJ0IC0tPiBKaW5qYVN0YXRlbWVudFN0YXJ0XG5CbG9ja1N0YXJ0S2V5d29yZCAtLT4gQmxvY2tNaWRkbGVLZXl3b3JkXG5CbG9ja1Rva2VuIC0tPiBCbG9ja0VuZFxuQmxvY2tUb2tlbiAtLT4gQmxvY2tTdGFydFxuQ29tbWVudCAtLT4gQmxvY2tDb21tZW50XG5Db21tZW50IC0tPiBMaW5lQ29tbWVudFxuSWRlbnRpZmllciAtLT4gSmluamFFeHByZXNzaW9uXG5KaW5qYVN0YXRlbWVudEVuZCAtLT4gSmluamFTdGF0ZW1lbnRNaWRkbGVcbkppbmphU3RhdGVtZW50U3RhcnQgLS0-IEppbmphU3RhdGVtZW50TWlkZGxlXG5OZXdsaW5lS2V5d29yZCAtLT4gVG9wTGV2ZWxLZXl3b3JkXG5PcGVyYXRvciAtLT4gRmllbGRBY2Nlc3NPcGVyYXRvclxuUmVzZXJ2ZWRLZXl3b3JkIC0tPiBBbmdsZUJyYWNrZXRLZXl3b3JkXG5SZXNlcnZlZEtleXdvcmQgLS0-IEJsb2NrRW5kS2V5d29yZFxuUmVzZXJ2ZWRLZXl3b3JkIC0tPiBCbG9ja1N0YXJ0S2V5d29yZFxuUmVzZXJ2ZWRLZXl3b3JkIC0tPiBTcGFjZUJlZm9yZUJyYWNrZXRLZXl3b3JkXG5TcGFjZUJlZm9yZUJyYWNrZXRLZXl3b3JkIC0tPiBBbGlhc1NlcGFyYXRvclxuU3BhY2VCZWZvcmVCcmFja2V0S2V5d29yZCAtLT4gTmV3bGluZUtleXdvcmRcblRva2VuXG5Ub2tlbiAtLT4gQmxvY2tUb2tlblxuVG9rZW4gLS0-IENsb3NpbmdCcmFja2V0XG5Ub2tlbiAtLT4gQ29tbWVudFxuVG9rZW4gLS0-IEV4cHJlc3Npb25TZXBhcmF0b3JcblRva2VuIC0tPiBJZGVudGlmaWVyXG5Ub2tlbiAtLT4gSmluamFTdGF0ZW1lbnRMaW5lXG5Ub2tlbiAtLT4gTGl0ZXJhbFxuVG9rZW4gLS0-IE1heWJlQ2xvc2luZ0FuZ2xlQnJhY2tldFxuVG9rZW4gLS0-IE1heWJlT3BlbmluZ0FuZ2xlQnJhY2tldFxuVG9rZW4gLS0-IE9wZW5pbmdCcmFja2V0XG5Ub2tlbiAtLT4gT3BlcmF0b3JcblRva2VuIC0tPiBSZXNlcnZlZEtleXdvcmRcblRva2VuIC0tPiBTdGF0ZW1lbnRTZXBhcmF0b3JcblRva2VuIC0tPiBXaGl0ZXNwYWNlIiwibWVybWFpZCI6eyJ0aGVtZSI6ImRlZmF1bHQifSwidXBkYXRlRWRpdG9yIjpmYWxzZX0)

acmiyaguchi:
I enabled this on the few examples that exist within the codebase, it looks like there's still a little more cleanup necessary:

```diff
diff --git a/bigquery_etl/glam/templates/clients_scalar_aggregates_v1.sql b/bigquery_etl/glam/templates/clients_scalar_aggregates_v1.sql
index 1fb49c7..78180ba 100644
--- a/bigquery_etl/glam/templates/clients_scalar_aggregates_v1.sql
+++ b/bigquery_etl/glam/templates/clients_scalar_aggregates_v1.sql
@@ -1,6 +1,4 @@
-{{ header }}
-{% include "udf_merged_user_data.sql" %}
-
+{{ header }} {% include "udf_merged_user_data.sql" %}
 WITH filtered_date_channel AS (
   SELECT
     {{ extract_select_clause }}
@@ -33,8 +31,7 @@ version_filtered_new AS (
     agg_type,
-    clients_scalar_aggregates_v1 AS scalar_aggs
-  {{ join_filter }}
+    clients_scalar_aggregates_v1 AS scalar_aggs {{ join_filter }}
 ),
 joined_new_old AS (
   SELECT
     {% for attribute in attributes_list %}
-      COALESCE(old_data.{{attribute}}, new_data.{{attribute}}) as {{attribute}},
-    {% endfor %}
+      COALESCE(old_data.{{attribute}}, new_data.{{attribute}}) as {{attribute}}, {% endfor %}
     old_data.scalar_aggregates AS old_aggs,
     new_data.scalar_aggregates AS new_aggs
   FROM
```

acmiyaguchi:
I'm going to wait until there are more examples in the repository before moving forward with this. I have some open PRs that will provide more coverage for formatting, like #735 and #743.

bigquery_etl/format_sql/tokenizer.py
```diff
@@ -433,6 +487,11 @@ class FieldAccessOperator(Operator):
     AngleBracketKeyword,
     SpaceBeforeBracketKeyword,
     ReservedKeyword,
+    JinjaStatementLine,
+    JinjaStatementMiddle,
+    JinjaStatementStart,
+    JinjaStatementEnd,
+    JinjaExpression,
```
relud: these are getting complicated. I think we should move these to separate token priority list `BIGQUERY_JINJA_TOKEN_PRIORITY`, and require an option to enable jinja support when formatting.

jklukas:
@relud - Do you think this is still worth pursuing?

relud:
not really

PR #731, opened 2020-02-11
Fenix ETL based on baseline pings for activity

jklukas:
This is intended to replace the existing ETL in `org_mozilla_fenix` datasets that references both baseline and metrics pings. I will save that for a follow-up PR, though, where I tear down the old definitions and hook up the new definitions to downstream ETL for GUD and KPI reporting.


sql/org_mozilla_firefox_derived/baseline_daily_v1/query.sql
```diff
@@ -0,0 +1,126 @@
...
+  UNION ALL
+  SELECT
+    extract_fields(baseline).*,
+    'beta' AS normalized_channel
+  FROM
+    org_mozilla_firefox_beta.baseline AS baseline
+  UNION ALL
+  SELECT
+    extract_fields(baseline).*,
+    'nightly' AS normalized_channel
```
jklukas: Maybe a different channel name here?

sql/org_mozilla_firefox_derived/clients_first_seen_v1/init.sql
```diff
@@ -0,0 +1,100 @@
...
+    base
+  WHERE
+    submission_timestamp > '2010-01-01'
+  GROUP BY
+    client_id
+)
+--
+SELECT
+  per_client_id.client_id,
+  per_client_id.first_seen_date,
```
jklukas: Maybe call this fenix_first_seen_date

sql/org_mozilla_firefox_derived/baseline_daily_v1/query.sql
jklukas: Perhaps 'fennec nightly' or 'aurora nightly'

sql/org_mozilla_firefox/baseline_daily/view.sql
```diff
@@ -0,0 +1,7 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.org_mozilla_firefox.baseline_daily`
+AS
+SELECT
+  *
+FROM
+  `moz-fx-data-shared-prod.org_mozilla_firefox_derived.baseline_daily_v1`
```
jklukas: The ETL here is across all channels of Fenix; I chose to put them in `org_mozilla_firefox` as the production app name, even though they hit datasets for a variety of distinct app names for the different channels. @fbertsch expressed some concern about this, but it seems like a reasonable compromise to avoid the extra overhead of defining these queries separately for each of the 5 app names involved and scheduling all of the queries.

sql/org_mozilla_firefox_derived/baseline_daily_v1/query.sql
```diff
@@ -0,0 +1,152 @@
...
+SELECT
+  wnd.submission_date,
+  cfs.fenix_first_seen_date,
+  cfs.fennec_first_seen_date,
+  wnd.* EXCEPT (_n, submission_date)
+  -- We need to do some post-processing on these arrays; they may contain nulls,
+  -- but BigQuery will throw an error if it tries to materialize an array with
+  -- null elements.
+  REPLACE(
+    ARRAY(SELECT DISTINCT d FROM UNNEST(start_dates) d WHERE d IS NOT NULL) AS start_dates,
```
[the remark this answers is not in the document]
jklukas: @relud Do you have an opinion on whether it would be better to save this is a bit pattern in baseline_daily rather than an array of dates? We'd be able to use a `BIT_OR` rather than an ARRAY_AGG, and it saves us this extra step of having to filter out null values.

sql/org_mozilla_firefox_derived/baseline_daily_v1/query.sql
```diff
@@ -0,0 +1,152 @@
...
+    ROW_NUMBER() OVER w1_unframed AS _n,
+    --
+    -- Take the earliest first_run_date if ambiguous.
+    MIN(first_run_date) OVER w1 AS first_run_date,
+    --
+    -- Sums over distinct baseline pings.
+    SUM(IF(duration BETWEEN 0 AND 100000, duration, 0)) OVER w1 AS durations,
+    --
+    -- Maintain lists of unique client activity dates for usage criteria based on client timestamps.
+    ARRAY_AGG(DATE(TIMESTAMP_SUB(end_time, INTERVAL duration SECOND))) OVER w1 AS start_dates,
```
relud: should this use `IF(duration BETWEEN 0 AND 100000, duration, NULL)` for both of these cases, so that it drops start dates from to oversize durations?

sql/org_mozilla_firefox_derived/baseline_daily_v1/query.sql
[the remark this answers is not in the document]
relud: those seem like good reasons to use a bit pattern.

Is it known who or what will be using this field, and how?

Would the pattern would be relative to submission_date?

Would we be exposing the bit pattern as a date array in a view?

I'm fine with it, but I'm also a bad judge of how confusing bit patterns are for consumers.

sql/org_mozilla_firefox_derived/baseline_daily_v1/query.sql
```diff
@@ -0,0 +1,169 @@
...
+    -- Sums over distinct baseline pings.
+    SUM(IF(duration BETWEEN 0 AND 100000, duration, 0)) OVER w1 AS durations,
+    --
+    -- Bit patterns capturing activity dates relative to the submission date.
+    BIT_OR(
+      1 << IF(session_start_date_offset BETWEEN 0 AND 27, session_start_date_offset, NULL)
+    ) OVER w1 AS days_seen_session_start_bits,
+    BIT_OR(
+      1 << IF(session_end_date_offset BETWEEN 0 AND 27, session_end_date_offset, NULL)
+    ) OVER w1 AS days_seen_session_end_bits,
```
jklukas: Went ahead and created bit patterns for session start and session end dates instead of date arrays. We can revisit this decision in the future.

udf/combine_adjacent_days_28_bits.sql
```diff
@@ -1,4 +1,18 @@
...
+/*
+
+Combines two bit patterns. The first pattern represents activity over a 28-day
+period ending "yesterday". The second pattern represents activity as observed
+today (usually just 0 or 1). We shift the bits in the first pattern by one to
+set the new baseline as "today", then perform a bitwise OR of the two patterns.
+
+*/
+CREATE OR REPLACE FUNCTION udf.combine_adjacent_days_28_bits(prev INT64, curr INT64) AS (
+  udf.shift_28_bits_one_day(prev) | IFNULL(curr, 0)
```
jklukas: Updated this combine_adjacent_days_28_bits function to use `|` instead of `+`. They end up doing the same thing in the simple case where `curr` is either 0 or 1, but the logical OR `|` is the correct thing to use now that we can have `curr` values that reflect activity on previous days.

PR #732, opened 2020-02-11
Break up clients_histogram_aggregates in Glam into smaller chunks.

emtwo:
`clients_histogram_aggregates` continues to give shuffle errors. This is an attempt to resolve it. I've tested it for 3 days of data and it hasn't failed yet.

Basically, motivated by: https://cloud.google.com/bigquery/docs/best-practices-performance-compute, it seems beneficial to break up computation into multiple intermediate materialized tables. I also used a partition on `channel` to make it possible to process one channel at a time to further break up computations. Since you cannot have partitions based on strings, I mapped each channel to an integer, `channel_enum`.

dry run will fail as the `channel_enum` field needs to be added to `clients_histogram_aggregates_v1`


emtwo:
@jklukas I wanted to see if you had any other thoughts for optimizations and/or your perspective on this. I really was hoping for a more elegant solution than just breaking it up, but this is what seemed to work....

sql/telemetry_derived/clients_histogram_aggregates_merged_v1/query.sql
```diff
@@ -0,0 +1,24 @@
...
+  COALESCE(old_data.os, new_data.os) AS os,
+  COALESCE(old_data.app_version, CAST(new_data.app_version AS INT64)) AS app_version,
+  COALESCE(old_data.app_build_id, new_data.app_build_id) AS app_build_id,
+  COALESCE(old_data.channel, new_data.channel) AS channel,
+  COALESCE(old_data.channel_enum, new_data.channel_enum) AS channel_enum,
+  old_data.histogram_aggregates AS old_aggs,
+  new_data.histogram_aggregates AS new_aggs
+FROM old_data
+FULL OUTER JOIN new_data
+  ON new_data.join_key = old_data.join_key
```
jklukas: Instead of the two CTEs, we could express this same join as:

```
FROM clients_histogram_aggregates_old_v1 AS old_data
FULL OUTER JOIN clients_histogram_aggregates_new_v1 AS new_data
ON new_data.join_key = old_data.join_key
WHERE old_data.channel_enum = @channel_enum
AND new_data.channel_enum = @channel_enum
```

It may allow BigQuery to better optimize the query plan, but no guarantees it will make any difference.

emtwo:
> If breaking it up by channel seems to help, do you think it would make sense to partition the data by client sample_id? There would be less skew for each of the parameterized queries, since the sample_ids are evenly distributed.

I really like this idea! I think if this fails again, we can try partitioning by `sample_id`. I was concerned about the skew as well but there weren't many options for fields to partition by, given what currently existed in that query. We would need to pull in the `sample_id` field starting at the `clients_daily_histogram_aggregates` stage.

> Additionally, what is the retention period for these queries? Will they be replaced in full each day, or some other mechanism to drop them to avoid the extra cost of storage?

These ones are replaced in full each day, only keeping the 3 most recent versions per channel. So in that sense, data is dropped.

acmiyaguchi:
> > If breaking it up by channel seems to help, do you think it would make sense to partition the data by client sample_id? There would be less skew for each of the parameterized queries, since the sample_ids are evenly distributed.
> 
> I really like this idea! I think if this fails again, we can try partitioning by `sample_id`. I was concerned about the skew as well but there weren't many options for fields to partition by, given what currently existed in that query. We would need to pull in the `sample_id` field starting at the `clients_daily_histogram_aggregates` stage.

This isn't necessarily the case, since `sample_id` can be derived using [`udf.safe_sample_id`](https://github.com/mozilla/bigquery-etl/blob/master/udf/safe_sample_id.sql) i.e. `crc32(client_id) % 100`. But yes, it's probably better to propagate the pre-computed sample id.

> > Additionally, what is the retention period for these queries? Will they be replaced in full each day, or some other mechanism to drop them to avoid the extra cost of storage?
> 
> These ones are replaced in full each day, only keeping the 3 most recent versions per channel. So in that sense, data is dropped.

:+1:

PR #734, opened 2020-02-11
Add glam views

robhudson:
Each view is per channel, intended to be extracted to CSV for import
into Glam's Postgresql db.


robhudson:
Assuming the CSV extract code is in the airflow repo, it could be written in Python and then the glam specific bits here could go away if that's important? It would be faster if we didn't have to write to a temp table before extracting but I'm not sure that's important. Open to discussion.

robhudson:
Closing in favor of another approach.

PR #735, opened 2020-02-12
Add scalar aggregates from clients daily scalar aggregates for Fenix

acmiyaguchi:
This depends on #727.

This PR adds `sql/glam/fenix_clients_scalar_aggregates_v1`. This is an incremental dataset that aggregates new dates into a single dataset.

This has been tested with a single day of data in `glam-fenix-dev:glam_etl`. However, more thought will need to be put into testing the aggregation (probably left to another PR).


bigquery_etl/glam/glean_scalar_aggregates.py
```diff
@@ -67,7 +67,12 @@ def generate_sql(
             DATE(submission_timestamp) AS submission_date,
             client_info.client_id,
             REPLACE(ping_info.ping_type, "_", "-") as ping_type,
-            SPLIT(client_info.app_display_version, '.')[OFFSET(0)] AS app_version,
+            COALESCE(
+                SAFE_CAST(
+                    SPLIT(client_info.app_display_version, '.')[OFFSET(0)]
+                    AS INT64
+                ),
+            0) AS app_version,
```
acmiyaguchi: There are pings with names like `Nightly 191010 06:01` that need to be filtered out. This logic is expecting semantic versioning. 
https://sql.telemetry.mozilla.org/queries/68262/source

bigquery_etl/glam/scalar_aggregates_incremental.py
```diff
@@ -1,68 +1,221 @@
...
+            "channel",
+        ],
+        attributes_type=["STRING", "STRING", "STRING", "INT64", "STRING", "STRING"],
+        extract_select_clause="*",
+        # no filtering on channel
+        join_filter="",
+        # TODO: is this the appropriate partitioning of the table?
+        partition_clause="""
+            PARTITION BY
+                RANGE_BUCKET(app_version, GENERATE_ARRAY(0, 100, 1))
```
acmiyaguchi: I'm not sure what appropriate partitioning for the table is. Does it need partitioning and clustering at all?

sql/glam_etl/fenix_clients_daily_scalars_aggregates_migration_v1/query.sql
```diff
@@ -940,6 +943,252 @@ aggregated AS (
         'sum',
         sum(CAST(metrics.counter.migration_open_tabs_success_reason AS INT64))
       ),
+      (
```
acmiyaguchi: The following section comes from updates to the migration ping.

sql/telemetry_derived/clients_scalar_aggregates_v1/query.sql
```diff
@@ -125,7 +125,7 @@ WITH filtered_date_channel AS (
     * EXCEPT (app_version),
     CAST(app_version AS INT64) AS app_version
   FROM
-    clients_daily_scalar_aggregates_v1
+    telemetry_derived.clients_daily_scalar_aggregates_v1
```
acmiyaguchi: This was necessary to make the init and query generic.

bigquery_etl/glam/scalar_aggregates_incremental.py
```diff
@@ -1,68 +1,221 @@
...
+            "ping_type",
+            "os",
+            "app_version",
+            "app_build_id",
+            "channel",
+        ],
+        attributes_type=["STRING", "STRING", "STRING", "INT64", "STRING", "STRING"],
+        extract_select_clause="*",
+        # no filtering on channel
+        join_filter="",
```
emtwo: What is the reason for having an empty join filter here?

bigquery_etl/glam/scalar_aggregates_incremental.py
emtwo: > I'm not sure what appropriate partitioning for the table is. Does it need partitioning and clustering at all?

The partitioning by version is useful because the join filter in desktop looks at `app_version >= (latest_version - 2)`.

bigquery_etl/glam/scalar_aggregates_incremental.py
acmiyaguchi: I wasn't sure that the versioning scheme is the same, since Fenix doesn't follow the release train for Firefox. But, it seems okay to just look at the last two versions based on how long 2/3 have been around based on some poking around.

emtwo: If it looks reasonable to you to treat Fenix versioning in the same way, I think it makes for a good filter.

acmiyaguchi: Okay, I've added the query for `fenix_latest_versions_v1`. Ironically, this is the most expensive query out of all of them, coming in at 2GB.

acmiyaguchi:
Updated this PR with a `fenix_latest_versions_v1` table, and use it as a version filter for the aggregates. The test failure is unrelated to this PR.

PR #736, opened 2020-02-12
Use pip-compile instead of constraints.txt

relud:
This makes it so that adding new dependencies doesn't have a manual process of tracking down dependencies.

If I remember correctly dependabot doesn't break requirements across multiple lines, and I don't know if it will preserve end-of-line comments, not to mention having it edit `requirements.txt` goes against `This file is autogenerated by pip-compile`. @acmiyaguchi do you know if dependabot can handle `requirements.in`?


acmiyaguchi:
It looks so, based on their blog post: ["Dependabot now supports pip-compile"](https://dependabot.com/blog/dependabot-now-supports-pip-compile/).

Dockerfile
```diff
@@ -2,13 +2,13 @@ ARG PYTHON_VERSION=3.8
...
-COPY requirements.txt constraints.txt ./
-RUN apt-get update -qqy && apt-get install -qqy gcc libc-dev && \
-    pip install -r requirements.txt
+RUN apt-get update -qqy && apt-get install -qqy gcc libc-dev
+COPY requirements.txt ./
+RUN pip install -r requirements.txt
 
 FROM python:${PYTHON_VERSION}-slim
-# add bash for entrypoing and python2 for google-cloud-sdk
-RUN apt-get update -qqy && apt-get install -qqy bash python
```
relud: `google-cloud-sdk` now supports python 3

PR #743, opened 2020-02-13
End-to-end Fenix scalar aggregates

acmiyaguchi:
With the generate/run bash scripts and templating logic figured out, this PR will finish up the scalar aggregates pipeline for Fenix data. 

```bash
% bq ls glam-fenix-dev:glam_etl
                  tableId                    Type    Labels   Time Partitioning          Clustered Fields          
 ------------------------------------------ ------- -------- ------------------- --------------------------------- 
  fenix_clients_daily_scalar_aggregates_v1   TABLE                                                                 
  fenix_clients_scalar_aggregates_v1         TABLE                                app_version, channel, client_id  
  fenix_clients_scalar_bucket_counts_v1      TABLE                                                                 
  fenix_clients_scalar_probe_counts_v1       TABLE                                                                 
  fenix_scalar_percentiles_v1                TABLE
```


bigquery_etl/glam/templates/metric_bucketing_v1.udf.sql
```diff
@@ -0,0 +1,359 @@
+CREATE TEMP FUNCTION udf_bucket(val FLOAT64)
+RETURNS FLOAT64 AS (
+  -- Bucket `value` into a histogram with min_bucket, max_bucket and num_buckets
+  (
+    SELECT
+      max(CAST(bucket AS INT64))
+    FROM
+      UNNEST(
```
acmiyaguchi: turn off formatting

bigquery_etl/glam/templates/metric_bucketing_v1.sql
```diff
@@ -0,0 +1,52 @@
...
+    {{ attributes }},
+    {{ aggregate_attributes }},
+    agg_type,
+    SAFE_CAST(udf_bucket(SAFE_CAST(value AS FLOAT64)) AS STRING) AS bucket
+  FROM
+    {{ source_table }}
+  CROSS JOIN
+    UNNEST(scalar_aggregates)
+  WHERE
+    metric_type in ({{ scalar_metric_types }})
```
acmiyaguchi: this should exclude the boolean type

script/glam/run_fenix_sql
```diff
@@ -82,9 +100,12 @@ function main {
...
     gcloud config set project $project
-    # force delete the dataset
-    bq rm -r -f $dataset
-    bq mk $dataset
+
+    if $reset; then
+        # force delete the dataset
+        bq rm -r -f $dataset
+        bq mk $dataset
+    fi
```
acmiyaguchi: this will fail if run for the first time and the dataset does not exist

bigquery_etl/glam/glean_metric_counts.py
```diff
@@ -0,0 +1,69 @@
...
+    """Render the main query."""
+    env = Environment(loader=PackageLoader("bigquery_etl", "glam/templates"))
+    sql = env.get_template("metric_counts_v1.sql")
+
+    # include attributes included and excluded from grouping set
+    attribute_combinations = []
+    for subset_size in reversed(range(len(attributes) + 1)):
+        for grouping in combinations(attributes, subset_size):
+            hidden = list(sorted(set(attributes) - set(grouping)))
+            attribute_combinations.append((grouping, hidden))
```
acmiyaguchi: This generates more lines than anticipated, what is the difference in logic from the handwritten one? @emtwo

bigquery_etl/glam/probe_counts.py
```diff
@@ -0,0 +1,108 @@
...
+
+    # only compute a shallow set of attributes to avoid excess complexity
+    max_combinations = 3
+    attribute_combinations = []
+    for subset_size in reversed(range(max_combinations + 1)):
+        for grouping in combinations(attributes, subset_size):
+            select_expr = []
+            for attribute in attributes:
+                select_expr.append((attribute, attribute in grouping))
+            attribute_combinations.append(select_expr)
```
emtwo: In the Glam front-end, both `channel` and `app_version` must be specified, so it's not necessary to have the combinations where these are `NULL`, especially if it can speed things up. The only reason to have them in the dataset is if someone wants to look at them outside of the Glam front-end. The desktop telemetry glam ETL doesn't need these either.

bigquery_etl/glam/templates/probe_counts_v1.sql
```diff
@@ -0,0 +1,49 @@
...
+        CASE
+        WHEN
+            metric_type IN ({{ scalar_metric_types }})
+        THEN
+            udf_fill_buckets(
+                udf_dedupe_map_sum(ARRAY_AGG(STRUCT<key STRING, value FLOAT64>(bucket, count))),
+                udf_get_buckets()
+            )
+        WHEN
+            metric_type in ('boolean', 'keyed-scalar-boolean')
```
emtwo: I guess this has no effect for now since you haven't added booleans yet, but they will later become a variable

bigquery_etl/glam/templates/bucket_counts_v1.udf.sql
```diff
@@ -0,0 +1,359 @@
...
+        THEN
+          value
+        ELSE
+          0
+        END
+        AS bool_false
+      FROM
+        UNNEST(scalar_aggs)
+      WHERE
+        metric_type IN ("boolean", "keyed-scalar-boolean")
```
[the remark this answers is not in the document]
acmiyaguchi: `boolean` is applicable, but `keyed-scalar-boolean` gets treated as a no-op. I just left it alone as part of refactoring.

bigquery_etl/glam/probe_counts.py
acmiyaguchi: I was curious about this, I noticed that I generated more lines than the original, but I couldn't figure out the pattern.

PR #746, opened 2020-02-14
Add metadata.yaml to ssl_ratios query

scholtzan:
This adds `metadata.yaml` for the `ssl_ratios_v1` query. This will be the first table being part of the public dataset.

### DIFF
--- a/sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
+++ b/sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
@@ -1,6 +1,7 @@
 friendly_name: SSL Ratios
 description: >
-  Percentages of page loads Firefox users have performed that were conducted over SSL.
+  Percentages of page loads Firefox users have performed that were 
+  conducted over SSL broken down by country.
 owners:
   - chutten@mozilla.com
 labels:
--- a/sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
+++ b/sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
@@ -6,6 +6,6 @@ owners:
   - chutten@mozilla.com
 labels:
   application: firefox
-  schedule: daily
+  incremental: true
   public_json: true
   public_bigquery: true
--- a/sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
+++ b/sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
@@ -7,5 +7,6 @@ owners:
 labels:
   application: firefox
   incremental: true
+  schedule: daily
   public_json: true
   public_bigquery: true
--- a/sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
+++ b/sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
@@ -1,0 +1,10 @@
+friendly_name: SSL Ratios
+description: >
+  Percentages of page loads Firefox users have performed that were conducted over SSL.
+owners:
+  - chutten@mozilla.com
+labels:
+  application: firefox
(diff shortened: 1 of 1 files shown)


sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: SSL Ratios
+description: >
+  Percentages of page loads Firefox users have performed that were conducted over SSL.
+owners:
+  - chutten@mozilla.com
+labels:
+  application: firefox
+  schedule: daily
+  public_json: true
```
scholtzan: I'll update the existing script for publishing metadata to BigQuery in a separate PR so that key-value pairs with boolean values get published as tags.

sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: SSL Ratios
+description: >
+  Percentages of page loads Firefox users have performed that were conducted over SSL.
+owners:
+  - chutten@mozilla.com
```
scholtzan: I'm actually not sure about this, but I saw that @chutten is the [owner of the dag in telemetry-airflow](https://github.com/mozilla/telemetry-airflow/blob/831fe84a36347f440ede4f5a90e0bf83d4fa1e1e/dags/ssl_ratios.py#L12). But correct me if it should be someone else.

chutten: If anyone would like to volunteer themselves, I'm happy to take myself out of the way. But I'm still probably the one people should come to with questions about this dataset, so it's fair. : )

sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: SSL Ratios
+description: >
+  Percentages of page loads Firefox users have performed that were conducted over SSL.
```
fbertsch: I might include something about it being broken down by country.

sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
```diff
@@ -0,0 +1,10 @@
...
+owners:
+  - chutten@mozilla.com
+labels:
```
fbertsch: Needs: `incremental` and `version`

scholtzan: There was some discussion in https://github.com/mozilla/bigquery-etl/pull/684 about using [`schedule:daily` instead of `incremental`](https://github.com/mozilla/bigquery-etl/pull/684#discussion_r368087046) and [the `version` label is redundant with the table name](https://github.com/mozilla/bigquery-etl/pull/684#discussion_r368081032)

jklukas: I specifically requested that we use `schedule:daily` rather than `incremental:true` because a string "true" seemed like an odd fit. But given that @scholtzan has now added support for key-less tags, it might be reasonable to call this use `incremental` as a tag rather than `schedule:daily`.

scholtzan: Okay, I changed it back to the `incremental` label

fbertsch: I think there's a bit of confusion here - something can be scheduled daily but not be incremental. We should probably include both.

scholtzan:
This should be good to go now

PR #754, opened 2020-02-19
Glam: Build ID filtering and more sample_id partition usage

emtwo:
Build ID Filtering:
Essentially keeping only the top two build IDs each day for each channel based on client count. 

Sample ID Partitioning:
I preemptively added more of this here since those other queries were also processing 30T+. We also get the added bonus that some of the processing can run in parallel.

### DIFF
--- a/script/dryrun
+++ b/script/dryrun
@@ -68,6 +68,8 @@ SKIP = {
     "sql/telemetry_derived/latest_versions/query.sql",
     # Query parameter not found
     "sql/telemetry_derived/experiments_v1/query.sql",
+    "sql/telemetry_derived/clients_histogram_aggregates_v1/query.sql",
+    "sql/telemetry_derived/clients_histogram_aggregates_old_v1/query.sql",
     "sql/telemetry_derived/clients_histogram_aggregates_merged_v1/query.sql",
     # Dataset moz-fx-data-shared-prod:glam_etl was not found
     "sql/glam_etl/fenix_clients_scalar_aggregates_v1/query.sql",
--- a/sql/telemetry_derived/clients_histogram_aggregates_old_v1/query.sql
+++ b/sql/telemetry_derived/clients_histogram_aggregates_old_v1/query.sql
@@ -1,3 +1,9 @@
+WITH clients_histogram_aggregates_partition AS
+  (SELECT *
+  FROM clients_histogram_aggregates_v1
+  WHERE sample_id >= @min_sample_id
+    AND sample_id <= @max_sample_id)
+
 SELECT
   sample_id,
   client_id,
@@ -7,7 +13,7 @@ SELECT
   hist_aggs.channel AS channel,
   CONCAT(client_id, os, app_version, app_build_id, hist_aggs.channel) AS join_key,
   histogram_aggregates
-FROM clients_histogram_aggregates_v1 AS hist_aggs
+FROM clients_histogram_aggregates_partition AS hist_aggs
 LEFT JOIN latest_versions
 ON latest_versions.channel = hist_aggs.channel
 WHERE app_version >= (latest_version - 2)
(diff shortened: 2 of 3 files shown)


sql/telemetry_derived/glam_valid_builds_v1/query.sql
```diff
@@ -0,0 +1,14 @@
+WITH build_id_user_counts AS (
```
acmiyaguchi: A docstring comment would be nice: `-- Calculate the top two build ids per day over the last year.`

sql/telemetry_derived/glam_valid_builds_v1/query.sql
```diff
@@ -0,0 +1,14 @@
...
+  SELECT channel, SAFE.PARSE_DATE('%Y%m%d', SAFE.SUBSTR(app_build_id, 0, 8)) AS date, app_build_id, COUNT(DISTINCT client_id) AS client_count,
+  FROM clients_histogram_aggregates_v1
+  GROUP BY 1, 2, 3),
+
+ranked AS (
+  SELECT DENSE_RANK() OVER (PARTITION BY channel, date order by client_count DESC) AS rank, *
+  FROM build_id_user_counts
+  WHERE date > DATE_SUB(CURRENT_DATE, INTERVAL 365 DAY))
+
+SELECT channel, ARRAY_AGG(app_build_id) AS build_ids
```
acmiyaguchi: Why nest the query instead of leaving it flat?

sql/telemetry_derived/clients_daily_keyed_histogram_aggregates_v1/query.sql
```diff
@@ -85,16 +85,28 @@ WITH filtered AS (
...
+      submission_date,
+      os,
+      app_version,
+      app_build_id,
+      payload,
+      COALESCE(glam_valid_builds_v1.channel, channel_filtered.channel) AS channel
+    FROM channel_filtered
+    LEFT JOIN glam_valid_builds_v1
+      USING (channel)
+    WHERE app_build_id IN UNNEST(build_ids)),
```
acmiyaguchi: If this were a flat dataset, it would just be the inner join of the `channel_filter` and `glam_valid_builds_v1` using `channel` and `build_id`.

sql/telemetry_derived/glam_valid_builds_v1/query.sql
acmiyaguchi: Do you know how much data this query is using, from one of your test runs?

emtwo: ~27GB

emtwo:
> Do you know how many build ids this is filtering out? A before/after number would be great to build an understanding of performance.

| Channel | Old Build ID Count | New Build ID Count  |
|-|-|-|
|Nightly|277|130|
|Beta|128|79|
|Release|486|197|

Although this looks like a huge improvement, it only slightly reduced the side of `clients_histogram_aggregates` when I filtered out invalid builds

| | Pre-Filter | Post-Filter |
|-|-|-|
|Row Count|461,194,610|460,479,164|
|Terabytes|35.43 TB|35.35 TB|

I suspect this is because most users generating data are using valid builds, so removing the invalid builds doesn't actually remove too much data.

acmiyaguchi:
> Although this looks like a huge improvement, it only slightly reduced the side of `clients_histogram_aggregates` when I filtered out invalid builds
> 	Pre-Filter 	Post-Filter
> Row Count 	461,194,610 	460,479,164
> Terabytes 	35.43 TB 	35.35 TB

Hmm, looks like only a difference of 1m rows, which is not all that much considering everything else. It may have more of an effect if it were used before the large join done in `clients_histogram_aggregates` to filter out invalid/old builds in the old aggregates, but I don't think it would prevent the shuffling errors.

It seems like it may be more fruitful to increase the number of subqueries (from 4 partitions to 20 partitions). Airflow will also handle the concurrency for the jobs, it shouldn't run more than 5-10 jobs at a time.

emtwo:
> I don't think it would prevent the shuffling errors.

I agree, but it also has the benefit of making less filtering work for the front-end which is helpful

> It seems like it may be more fruitful to increase the number of subqueries (from 4 partitions to 20 partitions). Airflow will also handle the concurrency for the jobs, it shouldn't run more than 5-10 jobs at a time.

Yes, this is worth a try as well

PR #756, opened 2020-02-20
Add static normal_distribution table

jklukas:
Per request of @kirill-demtchouk


sql/static/normal_distribution/data.csv
```diff
@@ -0,0 +1,6002 @@
+"score","value"
```
Ben-Wu: Doesn't matter in this case because `schema.json` exists but I believe if it didn't it would include the quotes in the field name which bigquery wouldn't allow anyway.  They should be removed in case anyone looks at this as an example.

PR #758, opened 2020-02-21
Bug 1572873 Add tables for powering AMO stats dashboard

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1572873

These two tables provide all the raw data needed to replicate the existing AMO stats page with the following caveat:

- We do not provide downloads broken down by source

### DIFF
--- a/sql/telemetry_derived/amo_stats_dau_v1/query.sql
+++ b/sql/telemetry_derived/amo_stats_dau_v1/query.sql
@@ -1,0 +1,189 @@
+/*
+
+Daily user statistics to power AMO stats pages. See bug 1572873.
+
+Each row in this table represents a particular addon on a particular day
+and provides all the information needed to populate the various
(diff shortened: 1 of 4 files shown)


sql/telemetry_derived/amo_stats_dau_v1/query.sql
```diff
@@ -0,0 +1,189 @@
...
+  WHERE
+    ARRAY_LENGTH(active_addons) > 0
+    AND submission_date = @submission_date
+),
+--
+unnested AS (
+  SELECT
+    * EXCEPT (active_addons, version, os),
+    version AS addon_version,
+    os AS app_os,
```
mreid-moz: curious: why rename this here rather than in the `cd` expression?

jklukas: That sounds reasonable. Will change.

PR #767, opened 2020-02-26
Simplify scalar aggregates for glam-fenix etl

acmiyaguchi:
This simplifies the scalar aggregates so they are computed in a single query. There are not as many metrics as the main ping, so the query complexity shouldn't spike too badly.

I've refactored the query generator to use a separate template. I've tested that these queries work in the glam-fenix-dev project via the `run_fenix_sql` script.


bigquery_etl/glam/templates/clients_daily_scalar_aggregates_v1.sql
```diff
@@ -0,0 +1,100 @@
...
+    {{ attributes }},
+    ARRAY<STRUCT<metric STRING, metric_type STRING, key STRING, agg_type STRING, value FLOAT64>>[
+        {{ unlabeled_metrics }}
+    ] as scalar_aggregates
+  FROM
+    extracted
+  GROUP BY
+    {{ attributes }}
+),
+grouped_labeled_metrics AS (
```
acmiyaguchi: Note: If the existing separation between scalars and keyed-scalars were to be emulated, it could be done using`{% if is_labeled %} ... {% endif %} sections with a variable on whether to include certain sub-expressions.

PR #768, opened 2020-02-26
Fixing some glam issues

emtwo:
This PR addresses 3 issues, laid out in the individual commit messages.


sql/telemetry_derived/scalar_percentiles_v1/query.sql
```diff
@@ -35,6 +35,7 @@ percentiles AS (
     APPROX_QUANTILES(value, 100)  AS aggregates
   FROM
     flat_clients_scalar_aggregates
+  WHERE os IS NOT NULL
```
acmiyaguchi: Why doesn't this need to be done for other attributes, like `app_build_id`?

emtwo: From looking at the data, `app_build_id` always has a value, but `os` does not and that was the specific source of duplication issues.

PR #771, opened 2020-02-28
Machinery to make queries public

scholtzan:
This is a first step towards publicly available datasets. This PR is more of a prototype and open for discussion. It adds support for writing query results into a public dataset. 

How it works:

* When passing a query into `entrypoint` (e.g. when called by an Airflow task) instead of running `bq` CLI to execute the query, the new script `run_query` is executed. This script will check if some metadata is associated with the query. In cases where the metadata indicates that query results should be public the results will be written to the public dataset.
* The script `publish_public_data_views` will generate and publish views for queries that are as public to the corresponding internal dataset. This script should be [executed alongside `publish_views`](https://github.com/mozilla-services/cloudops-infra/blob/9a7b3eb1b2ee7a3f068a7bb3cda626cc19dd1f72/projects/data-shared/Jenkinsfile.bigquery.prod#L100-L110)

Currently, making multipart query results public is not supported, but adding support in `run_multipart_query` should be relatively straight forward.

Right now, existing tables that get marked as public would need to be copied to the public dataset manually before running these scripts. 

The first dataset that will be made public this way will be [`ssl_ratios_v1`](https://github.com/mozilla/bigquery-etl/tree/master/sql/telemetry_derived/ssl_ratios_v1).


script/dryrun
```diff
@@ -26,6 +26,9 @@ DERIVED_DATASETS_DRY_RUN_URL = (
     "https://us-central1-moz-fx-data-derived-datasets.cloudfunctions.net/"
     "bigquery-etl-dryrun"
 )
+PUBLIC_DATASETS_DRY_RUN_URL = (
+    "https://us-central1-mozilla-public-data-prod.cloudfunctions.net/bigquery-etl-dryrun"
```
scholtzan: This is just a placeholder for now and it needs to be verified if this is actually the URL once the project is created

script/run_query
```diff
@@ -0,0 +1,62 @@
...
+            # a view to the public table in the internal dataset is created when CI runs
+            public_project_id_sub = rf"\1{args.public_project_id}"
+            query_arguments = PROJECT_ID_ARG_RE.sub(public_project_id_sub, query_arguments)
+    except yaml.YAMLError as e:
+        print(e)
+    except FileNotFoundError as e:
+        print("Metadata file does not exist: {}", e)
+
+    # run the query as shell command so that passed parameters can used as is
+    subprocess.call(["bq {} < {}".format(query_arguments, query_file)], shell=True)
```
scholtzan: Running the query using `bq` command-line tool allows to just pass all the query parameters and arguments through

script/run_query
```diff
@@ -0,0 +1,62 @@
...
+    path, _ = os.path.split(query_file)
+    metadata_file = os.path.join(path, METADATA_FILE)
+
+    try:
+        metadata = Metadata.from_file(metadata_file)
+        if metadata.is_public_bigquery(args.query_file):
+            # change the project ID to write results to the public dataset
+            # a view to the public table in the internal dataset is created when CI runs
+            public_project_id_sub = rf"\1{args.public_project_id}"
+            query_arguments = PROJECT_ID_ARG_RE.sub(public_project_id_sub, query_arguments)
```
scholtzan: The BigQuery project Id is passed as a `--project_id` from Airflow. To publish data in the project with public datasets this parameter gets replaced with the name of the public project

bigquery_etl/parse_metadata.py
```diff
@@ -0,0 +1,78 @@
+"""
+Parsing of metadata yaml files.
+"""
+
+import re
+import yaml
+
+
+class Metadata:
```
scholtzan: Accessing metadata is used in several places. Having a separate class with all the logic for parsing the metadata files reduced code duplications

fbertsch:
This is very exciting! I'll find time to review this on Monday.

script/run_query
```diff
@@ -0,0 +1,62 @@
...
+    except yaml.YAMLError as e:
+        print(e)
```
jklukas:
```suggestion
        print(e)
        sys.exit(1)
```

Seems like we should exit with an error status in this case.

script/run_query
jklukas: It looks like [argparse.parse_known_args](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.parse_known_args) exists to handle this sort of situation. That should allow us to avoid having a `--arguments` arg, and it should allow us to avoid `shell=True` in this invocation.

script/dryrun
jklukas: Do we need a separate dry run endpoint for public tables? I would imagine we could still use the shared-prod version to access public data.

scholtzan: Looking a bit more into this, it should be possible, however it would require to adjust the `bigquery-etl-dryrun` Cloud Function to allow specifying different projects. The public data queries need to get executed in the public data project.

script/dryrun
```diff
@@ -90,20 +97,32 @@ USE_DERIVED_DATASETS = {
...
         url = DRY_RUN_URL
+
     try:
+        request_data = {"dataset": basename(dirname(dirname(sqlfile))), "query": sql}
+
+        try:
+            # specify project ID for queries in public project
+            metadata = Metadata.of_sql_file(sqlfile)
+            if metadata.is_public_bigquery():
+                request_data["project"] = public_project
```
scholtzan: This option is currently ignored by the Cloud Function.

jklukas: I think the question here is which project will Airflow be running the query in. I was assuming that Airflow would be running the query jobs in the `shared-prod` project, with the destination table being in the public project. If that's the case, then I'd expect we can dry run the query using the existing cloud function in shared-prod.

scholtzan: True. 

I think right now the Cloud Function just dry runs the query file in the derived dataset without having an option to specify a different destination table.

scholtzan: Oh wait, for dryrun that should not matter

jklukas: Yeah, for dry run, there is no destination table. So it's a matter of whether the cloud function has access to the referenced tables, and that the project the query is running in matches the project assumed by the table references in the query.

script/run_query
scholtzan: Instead of replacing the `project_id` the `destination_table` needs to be set to the one that is in the public dataset.

scholtzan:
Alright, this is ready for review

script/run_query
```diff
@@ -0,0 +1,84 @@
...
+                )
+                query_arguments.append(
+                    "--destination_table={}".format(destination_table)
+                )
+                use_public_table = True
+    except yaml.YAMLError as e:
+        print(e)
+        sys.exit(1)
+    except FileNotFoundError as e:
+        print("Metadata file does not exist: {}", e)
```
jklukas: A metadata file is currently optional, so this condition is expected, but sounds like an error. Maybe this could be `INFO: No metadata.yaml found in {directory}`.

script/run_query
```diff
@@ -0,0 +1,84 @@
...
+
+    if not use_public_table:
+        # destination table was parsed by argparse, however if it wasn't modified to
+        # point to a public table it needs to be passed as parameter for the query
+        query_arguments.append("--destination_table={}".format(args.destination_table))
+
+    with open(query_file) as query_stream:
+        # run the query as shell command so that passed parameters can be used as is
+        sql = query_stream.read()
+        print(query_arguments)
```
jklukas: Is this print intentional?

script/run_query
```diff
@@ -0,0 +1,84 @@
...
+        # dataset ID was parsed by argparse but needs to be passed as parameter
+        # when running the query
+        query_arguments.append("--dataset_id={}".format(args.dataset_id))
+
+    use_public_table = False
+
+    try:
+        metadata = Metadata.of_sql_file(query_file)
+        if metadata.is_public_bigquery():
+            # change the destination table to write results to the public dataset
```
jklukas:
```suggestion
            # change the destination table to write results to the public dataset;
```

script/run_query
```diff
@@ -0,0 +1,84 @@
...
+            # a view to the public table in the internal dataset is created
+            # when CI runs
+            if args.dataset_id is not None and args.destination_table is not None:
```
jklukas: I'm worried this logic makes too many assumptions. It's possible to run `bq query` with no dataset specified, and with destionation_table set to something like `my_dataset.my_table`. Perhaps we should throw an error here if the options look like they don't match this particular format.

script/run_query
```diff
@@ -0,0 +1,84 @@
...
+                    "--destination_table={}".format(destination_table)
+                )
+                use_public_table = True
+    except yaml.YAMLError as e:
+        print(e)
+        sys.exit(1)
+    except FileNotFoundError as e:
+        print("Metadata file does not exist: {}", e)
+
+    if not use_public_table:
```
jklukas:
```suggestion
    if not use_public_table and args.destination_table is not None:
```

script/run_query
scholtzan: whoops

script/run_query
scholtzan: I added a regex for the destination table for more validation

script/entrypoint
```diff
@@ -27,8 +27,12 @@ elif [ "$1" = "query" ]; then
     # Alias
     #     query [options] FILE
     # to
-    #     bq query [options] < FILE
-    exec bq "${@:1:$#-1}" < "${@: -1}"
+    #     bq query [options] "FILE"
```
jklukas: This is no longer strictly true. Perhaps this comment can be:

```
    # For an invocation like:
    #     query [options] FILE
    # we dispatch to a script that inspects metadata and emulates the following call:
    #     bq query [options] < FILE 
```

script/run_query
```diff
@@ -0,0 +1,95 @@
...
+                query_arguments.append(
+                    "--destination_table={}".format(destination_table)
+                )
+                use_public_table = True
+            else:
+                print(
+                    "ERROR: Cannot run public dataset query. Parameters"
+                    " --destination_table=<table without dataset ID> and"
+                    " --dataset_id=<dataset> required"
+                )
```
jklukas: This looks great.

bigquery_etl/parse_metadata.py
```diff
@@ -0,0 +1,87 @@
...
+        friendly_name = None
+        description = None
+        labels = {}
+
+        with open(metadata_file, "r") as yaml_stream:
+            try:
+                metadata = yaml.safe_load(yaml_stream)
+
+                if "friendly_name" in metadata:
+                    friendly_name = metadata["friendly_name"]
```
jklukas:
```suggestion
                friendly_name = metadata.get("friendly_name", None)
```

This is a more compact notation that avoids repeating the same string literal twice. It also means you can avoid explicitly initializing the variable outside this context.

Can also be applied to description below.

scholtzan:
Before actually merging this the `ssl_ratios_v1` table needs to be copied to the public dataset. 

Also `publish_public_data_views` script needs to get set up to generate and publish views for queries that are as public.

Not sure if it makes sense to mention (especially for reviewers) that if tables get marked as public they need to be moved to the public dataset? Otherwise query results won't get stored anywhere.

scholtzan:
https://github.com/mozilla-services/cloudops-infra/pull/1919 adds support for calling  `publish_public_data_views`

bigquery_etl/parse_metadata.py
```diff
@@ -0,0 +1,84 @@
...
+
+    @staticmethod
+    def is_valid_label(label):
+        """
+        Check if a label has the right format.
+
+        Only hyphens (-), underscores (_), lowercase characters, and
+        numbers are allowed. International characters are allowed.
+        """
+        return re.match(r"[\w\d_-]+", label)
```
fbertsch: Should this check for length as well?

script/run_query
```diff
@@ -0,0 +1,95 @@
...
+                destination_table = "{}:{}.{}".format(
+                    args.public_project_id, args.dataset_id, args.destination_table
+                )
+                query_arguments.append(
+                    "--destination_table={}".format(destination_table)
+                )
+                use_public_table = True
+            else:
+                print(
+                    "ERROR: Cannot run public dataset query. Parameters"
```
fbertsch: I think this error can be moved up a level (not just for public queries). When do we have any queries running without both a `destination_table` and `dataset_id` specified?

script/publish_public_data_views
```diff
@@ -0,0 +1,108 @@
...
+)
+parser.add_argument("--target-project", help="Create views in the target project")
+parser.add_argument(
+    "--public-project",
+    default="mozilla-public-data",
+    help=("Project with publicly available data"),
+)
+parser.add_argument(
+    "--dry_run",
+    "--dry-run",
```
fbertsch: 💯

script/run_query
scholtzan: Currently it would be possible to run queries without specifying these parameters. So I'm not sure if we want to potentially break compatibility. However, I'm not aware of any queries not specifying those

scholtzan:
I updated the `publish_public_data_views` script to use BigQuery as the source of truth (as we discussed in the meeting today and mentioned in https://github.com/mozilla-services/cloudops-infra/pull/1919#pullrequestreview-370683356). So the script will retrieve all tables from the public dataset and create views in a target project for those tables. It is similar to `generate_views`.

@jklukas could you give that script another look?

scholtzan:
There seems to be an issue with permissions for performing a dryrun: `Access Denied: Table mozilla-public-data:telemetry_derived.ssl_ratios_v1: User does not have permission to query table mozilla-public-data:telemetry_derived.ssl_ratios_v1`

scholtzan:
Fixed the issue by setting `BigQuery Data Viewer` permissions for `allUsers`. This will also make the table publicly accessible.

PR #774, opened 2020-03-03
Get histogram buckets from probe info service (#672)

emtwo:
This addresses the case where some rows have values for min/max/num_buckets that are incorrect.


sql/telemetry_derived/clients_daily_histogram_aggregates_v1.sql.py
```diff
@@ -423,16 +438,40 @@ def get_histogram_probes(histogram_type):
...
+            if 'nightly' not in data[key]["history"]:
+                channel = 'beta'
+
+                if 'beta' not in data[key]["history"]:
+                    channel = 'release'
+
+            data_details = data[key]["history"][channel][0]["details"]
+            probe = key.replace("histogram/", "").replace(".", "_").lower()
+            bucket_details[probe] = {
+                "n_buckets": int(eval(str(data_details["n_buckets"]))),
```
acmiyaguchi: Is `eval` necessary to cast a string to an integer? 

```python
>>> int("5")
5
```

You may also consider using [`ast.literal_eval`](https://docs.python.org/3/library/ast.html#ast.literal_eval), which will evaluate literals and not expressions.

emtwo: Unfortunately, some probes, (e.g. `POPUP_NOTIFICATION_MAINACTION_TRIGGERED_MS`) have values in the probe info service like `80 * 25` for the value of `n_buckets` etc. So they do need to be evaluated as expressions.

Perhaps can add a comment

emtwo:
> Does this require changes in the `clients_histogram_aggregates` job to use the bucket ranges from this job?

No because the bucket ranges were being pulled in from main ping initially so they were already in `clients_daily_*`. I've just swapped them out to use the ranges from probe info service instead.

> It might be preferable to denormalize the data here and to make bucket range a separate table that gets joined during the `histogram_bucket_counts` job, where the bucket ranges are actually formed. The histograms get unnested, and they could be joined with a table derived from the probe dictionary that contains the appropriate join key (probe name) and the first bucket, last bucket, and num_bucket information needed for generating the final range.

I initially aimed to create this with as minimal changes as possible because I thought it was blocking MVP, but it's not. We can do it this way and will need to go back and make changes across the entire pipeline so that buckets are not passed down from the start.

sql/telemetry_derived/clients_daily_histogram_aggregates_v1.sql.py
```diff
@@ -1,11 +1,13 @@
 #!/usr/bin/env python3
 """clients_daily_histogram_aggregates query generator."""
 import sys
+import gzip
 import json
 import argparse
 import textwrap
 import subprocess
 import urllib.request
+from bigquery_etl.format_sql.formatter import reformat
```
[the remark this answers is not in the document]
emtwo: I had `PYTHONPATH` set to the repo root when I ran this locally. But good point, this generalizes better

acmiyaguchi:
> > Does this require changes in the `clients_histogram_aggregates` job to use the bucket ranges from this job?
> 
> No because the bucket ranges were being pulled in from main ping initially so they were already in `clients_daily_*`. I've just swapped them out to use the ranges from probe info service instead.

Ah, I didn't look close enough to see that it was replacing `udf_get_bucket_range`.

> I initially aimed to create this with as minimal changes as possible because I thought it was blocking MVP, but it's not. We can do it this way and will need to go back and make changes across the entire pipeline so that buckets are not passed down from the start.

I'm all for incremental change. Since this doesn't actually change the schema, I think this is the right approach. I may go for outlined approach for the Glean aggregates though, since the custom distribution metadata is not reported within the ping itself if I understand correctly.

sql/telemetry_derived/clients_daily_histogram_aggregates_v1.sql.py
acmiyaguchi: Leaving a comment about this specific case would be great.

PR #778, opened 2020-03-04
Bug fixes for fenix clients_last_seen query

jklukas:
When doing a `FULL JOIN`, rows that do not have a match will show nulls for
the values of all columns of the missing table, but the table itself will not
pass an `IS NULL` check.

This PR modifies the logic to check `table.client_id IS NOT NULL` instead, and
also makes sure we filter out rows that have not seen a baseline ping in the
relevant window.

With these changes, we should be guaranteed to have at least one populated
day in `baseline.days_seen_bits` and non-null `normalized_channel`.
I am backfilling and validating now.


sql/org_mozilla_firefox_derived/clients_last_seen_v1/query.sql
```diff
@@ -27,23 +27,17 @@ cls_today AS (
     client_id,
     sample_id,
     baseline_today.normalized_channel,
-    IF(
-      baseline_today IS NULL,
```
jklukas: I believe now that this comparison always returned false.

jklukas:
This is fully backfilled and I've validated that the new version looks more reasonable.

sql/org_mozilla_firefox_derived/clients_last_seen_v1/query.sql
```diff
@@ -18,32 +18,24 @@ WITH cls_yesterday AS (
     clients_last_seen_v1 AS cls
   WHERE
     submission_date = DATE_SUB(@submission_date, INTERVAL 1 DAY)
--- Filter out rows from yesterday that have now fallen outside the 28-day window.
-    AND udf.shift_28_bits_one_day(baseline.days_seen_bits) > 0
```
jklukas: This check moves to the end of the query so that we can filter out rows that only have metrics ping information.

PR #782, opened 2020-03-05
Materialize search RFM

fbertsch:


fbertsch:
Straightforward materialization of the existing view, simply for easier DS access from spark.

sql/search/search_rfm/query.sql
```diff
@@ -1,6 +1,3 @@
-CREATE OR REPLACE VIEW
-  `moz-fx-data-shared-prod.search.search_rfm`
-AS
 SELECT
   `moz-fx-data-shared-prod.udf.days_seen_bytes_to_rfm`(days_seen_bytes) AS days_seen,
```
relud: This query should use relative UDF and table references, because fully qualified references are only needed for views:
```suggestion
  udf.days_seen_bytes_to_rfm(days_seen_bytes) AS days_seen,
```

sql/search/search_rfm/query.sql
```diff
@@ -24,3 +21,5 @@ SELECT
   )
 FROM
   `moz-fx-data-shared-prod.search_derived.search_clients_last_seen_v1`
```
relud:
```suggestion
  search_clients_last_seen_v1
```

sql/search/search_rfm/init.sql
```diff
@@ -0,0 +1,28 @@
+CREATE OR REPLACE TABLE `moz-fx-data-shared-prod.search.search_rfm`
```
relud: `init.sql` is only needed when query.sql depends on past results, so this file should be removed

script/dryrun
```diff
@@ -39,7 +39,8 @@ SKIP = {
     "sql/telemetry/fxa_oauth_events_v1/query.sql",
     "sql/search_derived/search_clients_last_seen_v1/init.sql",
     "sql/search_derived/search_clients_last_seen_v1/query.sql",
-    "sql/search/search_rfm/view.sql",
+    "sql/search/search_rfm/init.sql",
+    "sql/search/search_rfm/query.sql",
```
relud: This should move to search_derived, add a version, and be exposed by a `sql/search/search_rfm/view.sql`

```suggestion
    "sql/search_derived/search_rfm_v1/query.sql",
```

Issue #785, opened 2020-03-05
FxA oauth log tables no longer exist


jklukas:
The [fxa_oauth_events query](https://github.com/mozilla/bigquery-etl/blob/master/sql/telemetry/fxa_oauth_events_v1/query.sql) failed today.

[Logs for the airflow task](https://workflow.telemetry.mozilla.org/log?task_id=fxa_oauth_events&dag_id=fxa_events&execution_date=2020-03-04T10%3A00%3A00%2B00%3A00) show:

> fxa_prod_logs.docker_fxa_oauth_20* does not match any table

Has there been a reconfiguration on the FxA backend? cc @irrationalagent

irrationalagent:
it appears that the oauth tables have gone away now and all their events should be living in the auth server tables eg `fxa_prod_logs.docker_fxa_auth_20*` so it is probably possible to kill this job.

[i checked the auth table real quick](https://sql.telemetry.mozilla.org/queries/68958/source) and the expected oauth-based events (token_created and token_checked) are there in expected quantities. 

see also jbuck's comment here https://bugzilla.mozilla.org/show_bug.cgi?id=1620321

irrationalagent:
though im realizing there are jobs that are downstream of this table as well. they probably need to be modified to not draw from this table too... such as `telemetry.fxa_content_auth_oauth_events_v1`

jklukas:
> though im realizing there are jobs that are downstream of this table as well. they probably need to be modified to not draw from this table too... such as `telemetry.fxa_content_auth_oauth_events_v1`

I think those views should stay as-is for now. They reference the imported tables; the imported `oauth` table still exists and will stay put in case we need to backfill downstream tables for time periods where events we care about were flowing to `oauth`.

PR #786, opened 2020-03-05
Deprecate fxa_oauth_events query

jklukas:
See #785


jklukas:
> lgtm, perhaps as a follow-up we should start a GRAVEYARD.md

I'm all for having a GRAVEYARD.md. For this particular case, though, I think it makes sense to keep the query around as long as the table itself still exists. I think it will be confusing to have random tables sitting in derived datasets that don't correspond to a query. We can graveyard in the case that we delete an underlying table.

PR #788, opened 2020-03-07
Add shredder support for integer range partitioning

relud:


bigquery_etl/shredder/delete.py
```diff
@@ -188,7 +189,7 @@ def get_task_id(target, partition_id):
     def create_job(client):
         query = dedent(
             f"""
-            DELETE
+            {"SELECT * FROM" if dry_run else "DELETE"}
```
relud: dry_run fails with `User does not have bigquery.tables.updateData permission for table` with a `DELETE` statement, but works and returns the correct total bytes processed with read-only permission using a `SELECT * FROM`

bigquery_etl/shredder/delete.py
```diff
@@ -367,7 +401,6 @@ def list_partitions(client, target):
                     max_single_dml_bytes=args.max_single_dml_bytes,
                     state_table=args.state_table,
                     states=states,
-                    expiration_timestamp=expiration_timestamp,
```
relud: this was leftover from removing clustering support

bigquery_etl/shredder/delete.py
```diff
@@ -223,30 +224,76 @@ def create_job(client):
...
-    for field in table.schema:
-        if field.name != table.time_partitioning.field:
-            continue
-        if field.field_type == "TIMESTAMP":
-            return f"DATE({field.name})"
-        return field.name
+    if table.range_partitioning:
+        return table.range_partitioning.field
+    if table.time_partitioning:
+        return next(
```
jklukas: Am I understanding correctly the intention here is that `next` returns the first item from the expression? It would feel more idiomatic to use a list comprehension and take the first element via `[0]`.

bigquery_etl/shredder/delete.py
```diff
@@ -223,30 +224,76 @@ def create_job(client):
...
 
-def list_partitions(client, target):
-    """List the partition ids and corresponding dates in a table."""
+def get_partition(table, partition_expr, end_date, id_=None) -> Optional[Partition]:
+    """Return a Partition for id_ unless it is a date on or after end_date."""
+    if id_ is None:
+        if table.time_partitioning:
+            return Partition(f"{partition_expr} < '{end_date}'")
+        return Partition("TRUE")
+    if id_ == "__NULL__":
```
jklukas: Where does this magic `"__NULL__"` value come from? Can this be a named constant with a descriptive name?

bigquery_etl/shredder/delete.py
```diff
@@ -223,30 +224,76 @@ def create_job(client):
...
+        return Partition(f"{partition_expr} IS NULL", id_)
+    if table.time_partitioning:
+        date = datetime.strptime(id_, "%Y%m%d").date()
+        if date < end_date:
+            return Partition(f"{partition_expr} = '{date}'", id_)
+        return None
+    if table.range_partitioning:
+        if id_ == "__UNPARTITIONED__":
```
jklukas: Same with this magic value.

bigquery_etl/shredder/delete.py
```diff
@@ -223,30 +224,76 @@ def create_job(client):
...
-            return f"DATE({field.name})"
-        return field.name
+    if table.range_partitioning:
+        return table.range_partitioning.field
+    if table.time_partitioning:
+        return next(
+            f"DATE({field.name})" if field.field_type == "TIMESTAMP" else field.name
+            for field in table.schema
+            if field.name == table.time_partitioning.field
+        )
```
jklukas:
```suggestion
        return f"CAST({field.name} AS DATE)".format(field=table.time_partitioning.field)
```

I believe this form is tolerant of both TIMESTAMP and DATE without having to determine the underlying type, but we should check that it is still usable for partition selection.

bigquery_etl/shredder/delete.py
relud: `next` is more performant, because it stops after the iterator returns one thing. I would argue that it's more idiomatic to use `next`:
```
Help on built-in function next in module builtins:

next(...)
    next(iterator[, default])

    Return the next item from the iterator. If default is given and the iterator
    is exhausted, it is returned instead of raising StopIteration.
```

bigquery_etl/shredder/delete.py
relud: > I believe this form is tolerant of both TIMESTAMP and DATE without having to determine the underlying type, but we should check that it is still usable for partition selection.

ooh, nice. i'll check and use if so.

relud: 🎉 cast works!

bigquery_etl/shredder/delete.py
relud: irrelevant, because we don't need to iterate, we can use: `f"CAST({table.time_partitioning.field} AS DATE)"`

bigquery_etl/shredder/delete.py
relud: `__NULL__` and `__UNPARTITIONED__` can be observed by creating a range partitioned table with values that are null and outside the configured range, and creating a time partitioned table with values that are null, and then listing the `partition_id`s with legacy SQL.

I became aware of this because it was causing errors because `__NULL__` isn't a valid date

bigquery_etl/shredder/delete.py
```diff
@@ -27,6 +31,13 @@
     help="Do not make changes, only log actions that would be taken; "
     "use with --log-level=DEBUG to log query contents",
 )
+parser.add_argument(
+    "--read_only",
+    "--read-only",
+    action="store_true",
+    help="Use SELECT * FROM instead of DELETE with dry run queries to prevent errors "
+    "due to read-only permissions being insufficient to dry run DELETE dml",
+)
```
relud: re-r? @jklukas this was added after you reviewed, the change is isolated to commit [a362c82](https://github.com/mozilla/bigquery-etl/pull/788/commits/a362c82af02858c79ebeedd602c372194fb4f1d0)

PR #789, opened 2020-03-07
Support using --only and --except together in shredder

relud:


bigquery_etl/util/table_filter.py
```diff
@@ -41,14 +40,22 @@ def glob_predicate(table: str, pattern: re.Pattern, arg: str) -> bool:
...
+            (args.only_tables, "only"),
+            (args.except_tables, "except"),
+        ]
+        if patterns
+    ]
+    if len(predicates) > 1:
+        return lambda table: all(predicate(table) for predicate in predicates)
+    if len(predicates) == 1:
+        return predicates[0]
     return lambda table: True
```
jklukas:
```suggestion
    return lambda table: all(predicate(table) for predicate in predicates)
```

I don' think the if conditions are necessary here, as the `all` solution will be valid even down to the empty list case `all([]) == True`.

PR #791, opened 2020-03-09
Add mode_last_retain_nulls

fbertsch:
When aggregating columns, it may be surprising for some people
to have the most frequent non-null column be returned for
`udf_mode_last`. This introduces a new function which will
return null if that is the most frequent value.

### DIFF
--- a/udf/mode_last_retain_nulls.sql
+++ b/udf/mode_last_retain_nulls.sql
@@ -1,0 +1,30 @@
+/*
+
+Returns the most frequently occuring element in an array.
+
+In the case of multiple values tied for the highest count, it returns the value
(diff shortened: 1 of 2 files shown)


udf/mode_last_retain_nulls.sql
```diff
@@ -0,0 +1,30 @@
+/*
+
+Returns the most frequently occuring element in an array.
+
+In the case of multiple values tied for the highest count, it returns the value
+that appears latest in the array. Nulls are retained.
```
jklukas: Can we add something along the lines of "see also: udf.mode_last which ignores null values" here, and then an equivalent callout in `udf.mode_last` for discovering this function?

PR #792, opened 2020-03-09
Move views to point at shared-prod data with fixed submission_date

jklukas:
Part of https://bugzilla.mozilla.org/show_bug.cgi?id=1612541


sql/telemetry/sync_flat_summary_v1/view.sql
```diff
@@ -6,4 +6,4 @@ SELECT
   submission_date_s3 AS submission_date,
   * REPLACE (ARRAY(SELECT * FROM UNNEST(devices.list)) AS devices)
 FROM
-  `moz-fx-data-derived-datasets.telemetry_derived.sync_flat_summary_v1`
+  `moz-fx-data-shared-prod.telemetry_derived.sync_flat_summary_v1`
```
jklukas: This table is large such that we want to avoid query-populating it, so it continues to have a field named `submission_date_s3`.

PR #793, opened 2020-03-10
Bug 1620663 - Add active clients derived tables for Messaging System

ncloudioj:
This adds various derived tables to calculate the daily, weekly, and monthly active clients for `messaging_system.[cfr/onboarding/snippets]`. Will file another PR to enable them in Airflow.

r? @jklukas

PR #794, opened 2020-03-10
Messaging system ETL following clients_last_seen methodology

jklukas:
Proposed as an alternative to https://github.com/mozilla/bigquery-etl/pull/793

In this formulation, MAU can be recovered via:

```
SELECT
  submission_date,
  release_channel,
  COUNTIF(days_since_seen < 1) AS dau,
  COUNTIF(days_since_seen < 7) AS wau,
  COUNTIF(days_since_seen < 28) AS mau,
FROM
  `moz-fx-data-shared-prod.messaging_system.cfr_users_last_seen`
WHERE
  submission_date > '2020-01-01'
GROUP BY
  1,
  2
```

### DIFF
--- a/sql/messaging_system_derived/cfr_users_daily_v1/query.sql
+++ b/sql/messaging_system_derived/cfr_users_daily_v1/query.sql
@@ -1,0 +1,46 @@
+WITH windowed AS (
+  SELECT
+    DATE(submission_timestamp) AS submission_date,
+    impression_id,
+    client_id,
+    ROW_NUMBER() OVER w1_unframed AS _n,
+    --
+    -- For all dimensions, we use the mode of observed values in the day.
+    udf.mode_last(ARRAY_AGG(release_channel) OVER w1) AS release_channel,
+    udf.mode_last(ARRAY_AGG(addon_version) OVER w1) AS addon_version,
+    udf.mode_last(ARRAY_AGG(metadata.geo.country) OVER w1) AS country,
+    udf.mode_last(ARRAY_AGG(version) OVER w1) AS version,
+  FROM
+    messaging_system.cfr
+  WHERE
+    -- Reprocess all dates by running this query with --parameter=submission_date:DATE:NULL
+    (@submission_date IS NULL OR @submission_date = DATE(submission_timestamp))
+  WINDOW
+    w1 AS (
+      PARTITION BY
+        sample_id,
+        client_id,
+        DATE(submission_timestamp)
+      ORDER BY
+        submission_timestamp
+      ROWS BETWEEN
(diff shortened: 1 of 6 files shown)


sql/messaging_system_derived/cfr_users_daily_v1/query.sql
```diff
@@ -0,0 +1,46 @@
...
+    DATE(submission_timestamp) AS submission_date,
+    impression_id,
+    client_id,
+    ROW_NUMBER() OVER w1_unframed AS _n,
+    --
+    -- For all dimensions, we use the mode of observed values in the day.
+    udf.mode_last(ARRAY_AGG(release_channel) OVER w1) AS release_channel,
+    udf.mode_last(ARRAY_AGG(addon_version) OVER w1) AS addon_version,
+    udf.mode_last(ARRAY_AGG(metadata.geo.country) OVER w1) AS country,
+    udf.mode_last(ARRAY_AGG(version) OVER w1) AS version,
```
jklukas: I picked out the dimensions that seem to be more-or-less stable per client to include here.

sql/messaging_system_derived/cfr_users_last_seen_v1/query.sql
```diff
@@ -0,0 +1,42 @@
+WITH _current AS (
+  SELECT
+-- In this raw table, we capture the history of activity over the past
+-- 28 days as a single 64-bit integer. The rightmost bit represents
+-- whether the user was active in the current day.
+    CAST(TRUE AS INT64) AS days_seen_bits,
+    * EXCEPT (submission_date)
+  FROM
+    cfr_users_daily_v1
```
ncloudioj: I guess this should be the raw table, i.e. `messaging_system_stable.cfr_v1` or `cfr_v1` for short here?

ncloudioj: Oh, I see. You created a view called `cfr_users_daily` above.

ncloudioj: Hmm, I am still a bit confused here about the name. `cfr_users_daily_v1` should be the table name, correct?

jklukas: See https://github.com/mozilla/bigquery-etl#views

> Views should generally not be named with a version suffix; a view represents a stable interface for users and whenever possible should maintain compatibility with existing queries; if the view logic cannot be adapted to changes in underlying tables, breaking changes must be communicated to fx-data-dev@mozilla.org

So the raw table has a version, but the view does not.

PR #795, opened 2020-03-10
clients_daily_* generated queries need a couple of additional arguments.

emtwo:
There was a failure due to using `xcom_push` without a json format, and I noticed these args will be needed


emtwo:
https://github.com/mozilla/telemetry-airflow/pull/885 is the relevant airflow PR.

note the failure is unrelated to this pr

PR #803, opened 2020-03-10
Bug 1616895 Move impression_stats_flat into bi dataset

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1616895


jklukas:
Dry run will fail here until the dataset exists.

PR #804, opened 2020-03-10
Sort shredder jobs by partition_id then table_id

### LINKED ISSUE #797: Sort shredder jobs by partition_id then table_id
Jobs are currently handled in order of when the table appears in `DELETE_TARGETS`, then `partition_id` in the order returned by the query that lists them.

This should simplify testing, by reducing the wait time needed to validate that recently submitted data is successfully deleted.

This should also reduce the likelihood of negative interactions between shredder and backfills, because shredder will spend most of each month handling data that is older than typical backfills.

### LINKED ISSUE #799: Shredder doesn't stop or wait for jobs to complete if updating state table fails
This results in submitting jobs very quickly and failing with `Too many DML statements outstanding against table`

relud:
fixes #797
fixes #799

This should fix an issue where shredder isn't fully utilizing the available reserved slots in bigquery-batch-prod because DELETE jobs against the same table get automatically delayed after the first two to avoid metadata update conflicts, even when --priority=INTERACTIVE

### DIFF
--- a/bigquery_etl/shredder/delete.py
+++ b/bigquery_etl/shredder/delete.py
@@ -1,13 +1,14 @@
 """Delete user data from long term storage."""
 
 from argparse import ArgumentParser
-from concurrent.futures import ThreadPoolExecutor
+from collections import defaultdict
 from dataclasses import dataclass, replace
 from datetime import datetime, timedelta
 from functools import partial
+from multiprocessing.pool import ThreadPool
+from operator import attrgetter
 from textwrap import dedent
-from typing import Optional
-import asyncio
+from typing import Callable, Iterable, Optional
 import logging
 import warnings
 
@@ -181,12 +182,9 @@ def get_task_id(target, partition_id):
     return task_id
 
 
-async def delete_from_partition(
-    executor,
-    client_q,
+def delete_from_partition(
     dry_run,
     partition_condition,
-    partition_id,
     priority,
     read_only,
     source,
@@ -194,7 +192,7 @@ async def delete_from_partition(
     target,
     **wait_for_job_kwargs,
 ):
-    """Process deletion requests for partitions of a target table."""
+    """Return callable to handle deletion requests for partitions of a target table."""
     # noqa: D202
 
     def create_job(client):
@@ -220,17 +218,9 @@ async def delete_from_partition(
             query, bigquery.QueryJobConfig(dry_run=dry_run, priority=priority)
         )
 
-    job = await client_q.async_with_client(
-        executor,
-        partial(
-            wait_for_job,
-            create_job=create_job,
-            dry_run=dry_run,
-            task_id=get_task_id(target, partition_id),
-            **wait_for_job_kwargs,
-        ),
+    return partial(
+        wait_for_job, create_job=create_job, dry_run=dry_run, **wait_for_job_kwargs
     )
-    return job.total_bytes_processed
 
 
 def get_partition_expr(table):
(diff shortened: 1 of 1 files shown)


bigquery_etl/shredder/delete.py
```diff
@@ -306,46 +294,52 @@ def list_partitions(client, table, partition_expr, end_date, max_single_dml_byte
...
-                    end_date=end_date,
-                    **kwargs,
-                )
-                for partition in list_partitions(
-                    client, table, partition_expr, end_date, max_single_dml_bytes
-                )
-            ]
+    for partition in list_partitions(
+        client, table, partition_expr, end_date, max_single_dml_bytes
+    ):
```
acmiyaguchi: +1 on finding removing the limit before I could comment on it.

bigquery_etl/shredder/delete.py
```diff
@@ -392,36 +387,52 @@ def list_partitions(client, table, partition_expr, end_date, max_single_dml_byte
...
+    # https://docs.python.org/3/howto/sorting.html#sort-stability-and-complex-sorts
+    tasks.sort(key=lambda task: sql_table_id(task.table))
+    tasks.sort(key=attrgetter("partition_sort_key"), reverse=True)
+    with ThreadPool(args.parallelism) as pool:
+        results = pool.map(
+            client_q.with_client, (task.func for task in tasks), chunksize=1
+        )
+    bytes_processed_by_table = defaultdict(lambda: 0)
+    for i, job in enumerate(results):
+        bytes_processed_by_table[tasks[i].table] += job.total_bytes_processed
```
acmiyaguchi: Curious about the choice of `defaultdict` instead of just setting the value in a dictionary directly.

bigquery_etl/shredder/delete.py
```diff
@@ -392,36 +387,52 @@ def list_partitions(client, table, partition_expr, end_date, max_single_dml_byte
...
-    if not results:
+    ]
+    if not tasks:
         logging.error("No tables selected")
         parser.exit(1)
-    bytes_processed, bytes_deleted = map(sum, zip(*results))
+    # ORDER BY partition_sort_key DESC, sql_table_id ASC
+    # https://docs.python.org/3/howto/sorting.html#sort-stability-and-complex-sorts
+    tasks.sort(key=lambda task: sql_table_id(task.table))
+    tasks.sort(key=attrgetter("partition_sort_key"), reverse=True)
```
acmiyaguchi: Am I understanding this correctly?

```
table_v1, 20200301
table_v1, 20200302
table_v2, 20200301
table_v2, 20200302
```

Would be sorted as:

```
table_v1, 20200302
table_v2, 20200302
table_v1, 20200301
table_v2, 20200301
```

The threadpool assigns tasks sequentially based on the ordering of the list, so DELETEs from the same table are unlikely to be scheduled at the same time, therefore increasing utilization of slots. 

---

Have your initial tests shown the expected increase in utilization?

bigquery_etl/shredder/delete.py
relud: I find it more readable than:
```python
bytes_processed_by_table[tasks[i].table] = bytes_processed_by_table.get(tasks[i].table, 0) + job.total_bytes_processed
```
or
```python
if tasks[i].table not in bytes_processed_by_table:
    bytes_processed_by_table[tasks[i].table] = 0
bytes_processed_by_table[tasks[i].table] += job.total_bytes_processed

bigquery_etl/shredder/delete.py
relud: > Am I understanding this correctly?

yes

> Have your initial tests shown the expected increase in utilization?

yes

bigquery_etl/shredder/delete.py
acmiyaguchi: Ah right, there are multiple partitions per table.

Issue #807, opened 2020-03-11
GLAM queries fail with "Argument list too long"


scholtzan:
Some GLAM queries such as `clients_daily_histogram_aggregates` fail with `OSError: [Errno 7] Argument list too long: 'bq'`. This is related to the new wrapper for running queries.

See: https://workflow.telemetry.mozilla.org/log?execution_date=2020-03-10T00%3A00%3A00%2B00%3A00&dag_id=glam&task_id=clients_daily_histogram_aggregates

Also potentially related to https://github.com/mozilla/bigquery-etl/issues/806

cc @emtwo

jklukas:
I think the problem here is that we used to do `bq <args> < path_to_query` but we now do `bq <args> query_text`. For long query text, we exceed the size that bash is able to handle.

We need to pipe the contents of the file into the invocation.

scholtzan:
Indeed. https://github.com/mozilla/bigquery-etl/pull/808 fixes this issue

Issue #806, opened 2020-03-11
Event queries fail with RecursionError


jklukas:
From [logs in Airflow](https://workflow.telemetry.mozilla.org/log?task_id=bq_main_events&dag_id=main_summary&execution_date=2020-03-10T01%3A00%3A00%2B00%3A00) it looks like main_events and event_events queries failed last night, even though the tasks show success in Airflow.

The error is:

> RecursionError: maximum recursion depth exceeded

And it looks like it's coming from the `bq` CLI. This may have something to do with the new wrapper we have for queries in the `entrypoint` script. We need to figure out how to get these queries running, and also how to make sure the script passes through the correct exit code.

cc @relud @scholtzan

jklukas:
We may want to look into using [os.execv](https://docs.python.org/3/library/os.html#os.execv) or similar to actually replace the python process with the script it's invoking.

scholtzan:
[subprocess.check_call](https://docs.python.org/2/library/subprocess.html#subprocess.check_call) might also be an option here.

jklukas:
It looks like the os.execv option would be pretty ugly due to the need to pipe in the query contents: https://stackoverflow.com/a/24514542/1260237

jklukas:
I was able to successfully rerun the task!

PR #808, opened 2020-03-11
Use subprocess.check_call when running queries

scholtzan:
[`subprocess.check_call`](https://docs.python.org/2/library/subprocess.html#subprocess.check_call) will wait for command to complete. If the return code was zero then return, otherwise raise `CalledProcessError`

https://github.com/mozilla/bigquery-etl/issues/806 https://github.com/mozilla/bigquery-etl/issues/807

### DIFF
--- a/script/run_query
+++ b/script/run_query
@@ -87,8 +87,7 @@ def main():
 
     with open(query_file) as query_stream:
         # run the query as shell command so that passed parameters can be used as is
-        sql = query_stream.read()
-        subprocess.call(["bq"] + query_arguments + [sql])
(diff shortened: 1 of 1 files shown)


script/run_query
```diff
@@ -87,8 +87,7 @@ def main():
 
     with open(query_file) as query_stream:
         # run the query as shell command so that passed parameters can be used as is
-        sql = query_stream.read()
-        subprocess.call(["bq"] + query_arguments + [sql])
+        subprocess.check_call(["bq"] + query_arguments, stdin=query_stream)
```
scholtzan: Using `stdin` solves https://github.com/mozilla/bigquery-etl/issues/807

PR #809, opened 2020-03-11
Update glam extract queries to be per channel

robhudson:


sql/telemetry_derived/glam_client_probe_counts_beta_extract_v1/query.sql
```diff
@@ -11,7 +11,7 @@ WITH deduped AS (
     FROM
         `moz-fx-data-shared-prod.telemetry.client_probe_counts`
     WHERE
-        channel IN ("nightly", "beta", "release")
+        channel="beta"
```
emtwo: In order to be less redundant, you could add in a query param like is done here: https://github.com/mozilla/bigquery-etl/blob/master/sql/telemetry_derived/clients_histogram_aggregates_merged_v1/query.sql#L10

Then pass it in from airflow like so: https://github.com/mozilla/telemetry-airflow/blob/master/dags/glam_subdags/general.py#L27 to the extracts tasks. They would need to have a `sql_file_path` arg so they all point to the same file to be used.

robhudson: Great! Thanks for pointing that out. This is now updated.

Matching change in airflow: https://github.com/mozilla/telemetry-airflow/pull/896

PR #810, opened 2020-03-11
Use standard SQL when running queries

scholtzan:
`bq` uses legacy SQL by default (see https://cloud.google.com/bigquery/docs/reference/bq-cli-reference). This causes some of the SQL queries that exceed 250,000 characters to fail running. 

Use standard SQL instead (which has 1MB limit and is also used for dryruns)

cc @emtwo

### DIFF
--- a/script/run_query
+++ b/script/run_query
@@ -36,14 +36,21 @@ parser.add_argument(
 )
 parser.add_argument("--dataset_id", help="Destination dataset")
 parser.add_argument("--query_file", help="File path to query to be executed")
+parser.add_argument(
+    "--use_legacy_sql",
+    help="Enable use of legacy SQL; Standard SQL by default",
+    default=False,
+)
 
 
 def main():
     args, query_arguments = parser.parse_known_args()
     query_file = args.query_file
 
-    # bq uses legacy SQL by default, force use of standard SQL
-    query_arguments.append("--use_legacy_sql=false")
+    # bq uses legacy SQL by default, force use of standard SQL if no flag is set
+    query_arguments.append(
+        "--use_legacy_sql={}".format(str(args.use_legacy_sql).lower())
+    )
 
     if args.dataset_id is not None:
         # dataset ID was parsed by argparse but needs to be passed as parameter
--- a/script/run_query
+++ b/script/run_query
@@ -42,6 +42,9 @@ def main():
     args, query_arguments = parser.parse_known_args()
     query_file = args.query_file
 
+    # bq uses legacy SQL by default, force use of standard SQL
+    query_arguments.append("--use_legacy_sql=false")
+
     if args.dataset_id is not None:
         # dataset ID was parsed by argparse but needs to be passed as parameter
         # when running the query


scholtzan:
I'm surprised this didn't cause any problems previously 🤔

jklukas:
I suspect there's already something in place, or that there was something in place that was setting this parameter. I'm doing some digging.

jklukas:
This is handled in https://github.com/mozilla/bigquery-etl/blob/master/.bigqueryrc

scholtzan:
I see. Seems like `run_query` currently ignores this file

jklukas:
It's copied to `/root` in the Dockerfile: https://github.com/mozilla/bigquery-etl/blob/f3e11b42554edf47173db7e7bc9dcb06e5649ea9/Dockerfile#L16

I think it must be in effect within `run_query` or else all queries would be failing. Is there a specific case where we thing legacy SQL is being used?

scholtzan:
Yes, there are some GLAM queries failing, like ` clients_daily_histogram_aggregates`.

If I do a dryrun locally:
`./script/entrypoint query --destination_table=clients_daily_histogram_aggregates_v1$20200310 --dataset_id=telemetry_derived --project_id=moz-fx-data-shared-prod --dry_run --parameter=submission_date:DATE:2020-03-10 sql/telemetry_derived/clients_daily_histogram_aggregates_v1/query.sql
`

it throws

```
BigQuery error in query operation: The query is too large. The maximum legacy SQL query length is 256.000K characters, including comments and white space characters.
```

jklukas:
That makes sense that outside the docker container, `entrypoint` may not do the right thing. I'm reticent to inject `--use_legacy_sql` into the query arguments in case the user sends in that parameter. We could check for its existence in `query_args` and only inject it if it doesn't already exist.

Another option is to inject `#standardSQL\n` at the top of the query text: https://cloud.google.com/bigquery/docs/reference/standard-sql/enabling-standard-sql#sql-prefix

scholtzan:
Checking for the query argument and setting it if it doesn't exists is probably the more elegant solution here. At least that would avoid changing all the queries affected and would be more compatible with what was in place previously

script/run_query
```diff
@@ -36,12 +36,22 @@ parser.add_argument(
...
+)
 
 
 def main():
     args, query_arguments = parser.parse_known_args()
     query_file = args.query_file
 
+    # bq uses legacy SQL by default, force use of standard SQL if no flag is set
+    query_arguments.append(
+        "--use_legacy_sql={}".format(str(args.use_legacy_sql).lower())
```
scholtzan: I believe it only allows `true` or `false` so we need to lowercase the Python bool values

jklukas: I believe it cannot not take an argument. You can specify `--use_legacy_sql` to ensure legacy SQL is used or `--nouse_legacy_sql` to ensure Standard SQL is used.

scholtzan: Looks like both `--nouse_legacy_sql` and `--use_legacy_sql=false` is supported. We use the latter here: https://github.com/mozilla/bigquery-etl/blob/107bc37cbd96712b4ee4f92ec06e9b2322e7e69b/.bigqueryrc#L5

scholtzan:
This seems no longer to be an issue. All the GLAM tasks succeeded in Airflow.

PR #812, opened 2020-03-12
Stop shredder immediately when tasks fail

relud:
When an exception is raised by a task submitted to `ThreadPool`, the pool will refuse to accept more tasks, but continue until the current tasks are done. With `pool.map` all tasks are submitted immediately, resulting in all tasks executing even a task throws an exception.

This avoids that behavior by using `pool.imap` to submit tasks lazily, resulting in only waiting until currently running tasks are done when a task throws an exception.


relud:
After reflecting on what this means for shredder execution, I think it's better to leave this unmerged, to minimize the number of retries that may be needed due to transient failures.

PR #813, opened 2020-03-12
Add exact_mau28 tables/views for Messaging System

ncloudioj:
r? @jklukas 

This is a follow up to precomputer the mau/dau/wau for those `user_last_seen` tables defined in Messaging System.


jklukas:
This looks good at first glance. I'll plan to create these tables tomorrow to get the dry run working and then approve.

ncloudioj:
Sounds good to me. Thanks!

I've checked the derived datasets that we added yesterday, they all look solid. This would be the last piece for the mau/dau/wau work.

PR #814, opened 2020-03-12
Remove the dedupe logic from glam extracts

robhudson:


ANich:
This was done in https://github.com/mozilla/bigquery-etl/pull/957/

PR #817, opened 2020-03-13
Update publish_metadata to write metadata to public tables

scholtzan:
I updated the `publish_metadata` to retrieve tables from BigQuery and update their metadata. This is similar to what is already done in `publish_public_data_views`.

I'm wondering if it would make sense to add this script to the docker container to be executed alongside [the other scripts](https://github.com/mozilla-services/cloudops-infra/blob/9f4bc5d4fe6c08677f75f7e4da0ca5e575f1653e/projects/data-shared/Jenkinsfile.bigquery.prod#L124)?


scholtzan:
That would make sense

PR #818, opened 2020-03-13
Use stable tables rather than live

jklukas:
These two cases look like mistakes where we should be referencing stable rather than live tables.

### DIFF
--- a/sql/monitoring/telemetry_missing_columns_v2/view.sql
+++ b/sql/monitoring/telemetry_missing_columns_v2/view.sql
@@ -17,7 +17,7 @@ extracted AS (
     `moz-fx-data-shared-prod`.udf.extract_document_version(_TABLE_SUFFIX) AS document_version,
     additional_properties
   FROM
-    `moz-fx-data-shared-prod.telemetry_live.*`
+    `moz-fx-data-shared-prod.telemetry_stable.*`
   WHERE
     -- only observe full days of data
     (
--- a/sql/telemetry/fenix_events_v1/view.sql
+++ b/sql/telemetry/fenix_events_v1/view.sql
@@ -29,6 +29,6 @@ SELECT
       STRUCT(client_info.architecture AS arch)
     ) AS user_properties
 FROM
-    `moz-fx-data-shared-prod.org_mozilla_fenix_live.events_v1`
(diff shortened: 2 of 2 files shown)


sql/monitoring/telemetry_missing_columns_v2/view.sql
```diff
@@ -17,7 +17,7 @@ extracted AS (
     `moz-fx-data-shared-prod`.udf.extract_document_version(_TABLE_SUFFIX) AS document_version,
     additional_properties
   FROM
-    `moz-fx-data-shared-prod.telemetry_live.*`
+    `moz-fx-data-shared-prod.telemetry_stable.*`
```
jklukas: @acmiyaguchi From my reading of this query, we don't ever read data from the current day. Is there any reason not to hit stable rather than live tables?

sql/telemetry/fenix_events_v1/view.sql
```diff
@@ -29,6 +29,6 @@ SELECT
       STRUCT(client_info.architecture AS arch)
     ) AS user_properties
 FROM
-    `moz-fx-data-shared-prod.org_mozilla_fenix_live.events_v1`
+    `moz-fx-data-shared-prod.org_mozilla_fenix_stable.events_v1`
```
jklukas: @fbertsch This view is used for amplitude event export. I'm guessing the reference to live table here is either a mistake or a historical artifact. Any objection to switching to stable tables for this?

In particular, I believe the amplitude export job checks to make sure the count is non-zero from this query before running. That check seems meaningless on top of live tables.

sql/monitoring/telemetry_missing_columns_v2/view.sql
acmiyaguchi: This was because this query was developed in `moz-fx-data-shar-nonprod-efed`, where the stable tables do not exist. It should be fine to use the stable table and use the most recent date.

sql/telemetry/fenix_events_v1/view.sql
fbertsch: Yup, that seems correct to me. This probably existed for the same reason as the other one; that there were no stable tables.

PR #819, opened 2020-03-13
Add addons_daily sql

benmiroglio:
Getting the ball rolling for recreating the `addons_daily` dataset in GCP.

I created a base query, filtering to a @submission_date, and the corresponding view. Query review / guidance for next steps to get this scheduled would be much appreciated.


sql/telemetry_derived/addons_daily_v1/query.sql
```diff
@@ -0,0 +1,235 @@
...
+    days_since_seen < 7
+  GROUP BY
+    1
+),
+maut AS (
+  SELECT
+    aa.addon_id,
+    COUNT(DISTINCT client_id) AS mau
+  FROM
+    addons_expanded
```
jklukas:
```suggestion
    addons_expanded
  WHERE
    days_since_seen < 28
```

This is still correct without a WHERE clause, but I like having the 28 in here to be explicit about how we define mau.

jklukas:
I don't have previous experience with addons_daily, but this query looks to be nicely following best practices for this repo.

This is pulling in data from search_clients_daily in the `search` dataset which has restricted access. Should this also be under the `search` dataset? I assume this is replacing an existing job that populates a table outside of `search`, though.

benmiroglio:
@jklukas Thanks for the review. This ideally will populate a table outside of the `search` dataset. We are likely going to deprecate this dataset as we reevaluate the need for restricted access. In the meantime, however, I can add the appropriate user (whatever CI is using + Airflow) to the search group to get around that error. Let me know what you think!

sql/telemetry_derived/addons_daily_v1/query.sql
benmiroglio: Agreed!

jklukas:
I created this table and have backfilled 2020 dates. Please take a look and make sure the content seems correct.

The dry run is going to fail on this due to the access control on search. You'll need to add this to the skip list in `scripts/dry_run`.

If you make that change and the table looks good to you, I'll approve, merge, and finish the backfill back to 2016.

benmiroglio:
Thanks Jeff!

I'll plan to do some toy analysis of that dataset and solicit DS feedback before giving the OK to merge / backfill in case there's anything critically missing or broken. I'm out next week at a work week so this may happen week after next.

benmiroglio:
@jklukas 

I added two fields to the dataset. Would you mind replacing the current table with the updated one for a few days as one last sanity check before we do the big backfill?

jklukas:
With the current version, I'm getting an error:

> - query: Unrecognized name: default_search_engine at [142:23]

benmiroglio:
My mistake, missed that in my copy-paste from prototyping in GCP console. Should work now.

jklukas:
Query is now working and 2020 backfill is running. Should finish within an hour.

jklukas:
> Query is now working and 2020 backfill is running. Should finish within an hour.

Turns out it only took 10 minutes.

benmiroglio:
Things look good in the new dataset. I'd say we're ready to merge/backfill. Thanks for all your help @jklukas !

jklukas:
Backfilling now.

jklukas:
Backfill is complete.

PR #820, opened 2020-03-14
Add `build_id` column to `events_v1` view

mnoorenberghe:
This is useful for checking correlations of new data added in the middle of a release cycle vs. an existing event from earlier build IDs.

For example, I had an existing event e2 and added event e1 mid-cycle and now I want to confirm that e2 events are a subset of e1 events on builds that have e1.

Since both sources already have this column it seems reasonable to expose it on the view I think.

I haven't tested this patch but it seems straightforward.


mnoorenberghe:
Hmm… I was a bit too quick updating the query and though the UNION was only of two tables/views, not three so I missed the `moz-fx-data-shared-prod.telemetry_derived.events_v1` one… I can't find the source for that table in this repo so I don't know what to do about it… any suggestions? Is this only old data so we could perhaps put `NULL` into the column?

jklukas:
> Hmm… I was a bit too quick updating the query and though the UNION was only of two tables/views, not three, so I missed the `moz-fx-data-shared-prod.telemetry_derived.events_v1` one… I can't find the source for that table in this repo so I don't know what to do about it… any suggestions? Is this only old data so we could perhaps put `NULL` into the column?

`telemetry_derived.events_v1` is indeed "old" data populated from a previous version of this logic, likely it's data imported from Parquet before we transitioned to the main_event and event_event queries.

Using `CAST(NULL AS STRING) AS build_id` seems like the a reasonable option here.

jklukas:
Adding @sunahsuh for review based on her expertise with these event views.

mnoorenberghe:
> I'm not quite sure why I had those columns in the main_events and event_events tables but didn't expose them in the view – this looks great to me though, and I'll see about adding the rest of the columns that aren't exposed.

Hi @sunahsuh, does that mean you will add to this PR or will you do that in a separate PR? (I don't have merge privileges)

jklukas:
I'm rerunning tests now and will go ahead and merge. These changes should be applied to prod in the next day or two.

sunahsuh:
Whoops, sorry, habit -- we tend to let authors merge in their own PRs on the team and I forgot contributors from outside can't do that

mnoorenberghe:
No worries, that's why I asked :) Thank you both for your help!

mnoorenberghe:
> These changes will be deployed to prod in the next day or two.

Sorry to bother you but did this get deployed yet? I'm not seeing the column on STMO.

jklukas:
We have been experiencing unexpected issues with pipeline stability recently which have delayed publishing schema changes.

I just manually republished this view, though, so please try again. STMO may cache the schema, so may not be immediately available in the schema browser there.

mnoorenberghe:
Thank you. It's available now!

PR #824, opened 2020-03-16
Move intermediate old + merged queries into clients_histogram_aggregates

emtwo:
This causes less data to be read overall since it doesn't need to be stored in multiple intermediate locations to be read.


emtwo:
for airflow: https://github.com/mozilla/telemetry-airflow/pull/905

Issue #60, opened 2019-04-03
Migrate DesktopDau and DesktopActiveDau to clients_last_seen


relud:
these are probably already covered by firefox_desktop_exact_mau28 and growth dashboard, but this should include disconnecting the old datasets where possible.

PR #825, opened 2020-03-16
Add a segment for regular users.

felixlawrence:
This is a straw man proposal. Before accepting the PR:

1. With feedback from Data Science, the precise definition of the segment might be tweaked.
2. We should decide whether to include the version (v1) in the column name, column values, both, or neither. (The goal is to iterate on this definition rather than lock us into something suboptimal).
3. We should decide whether to extend `days_seen_bits` and friends to include 29 days - so that we can use 28 days of history as context for the current day (day 29).
4. My code should probably be rewritten following whatever style and naming conventions are applicable to this project!

This is related to [GUD issue #60](https://github.com/mozilla/GUD/issues/60)

### DIFF
--- a/sql/telemetry/clients_last_seen_v1/view.sql
+++ b/sql/telemetry/clients_last_seen_v1/view.sql
@@ -13,6 +13,36 @@ SELECT
   CAST(
     SAFE.LOG(days_created_profile_bits & -days_created_profile_bits, 2) AS INT64
   ) AS days_since_created_profile,
+  CASE
+    CAST(
+      BIT_COUNT(
+        days_visited_5_uri_bits & `moz-fx-data-shared-prod.udf.bitmask_range`(2, 6)
+      ) >= 1 AS INT64
+    ) + CAST(
+      BIT_COUNT(
+        days_visited_5_uri_bits & `moz-fx-data-shared-prod.udf.bitmask_range`(8, 7)
+      ) >= 2 AS INT64
+    ) + CAST(
+      BIT_COUNT(
+        days_visited_5_uri_bits & `moz-fx-data-shared-prod.udf.bitmask_range`(15, 7)
+      ) >= 2 AS INT64
+    ) + CAST(
+      BIT_COUNT(
+        days_visited_5_uri_bits & `moz-fx-data-shared-prod.udf.bitmask_range`(22, 7)
+      ) >= 2 AS INT64
+    )
+  WHEN
+    4
+  THEN
+    'regular_v1'
+  WHEN
+    0
+  THEN
+    'new_irregular_v1'
+  ELSE
+    'semi_regular_v1'
+  END
+  AS segment_regular_users_v1,
   * EXCEPT (
     active_experiment_id,
     scalar_parent_dom_contentprocess_troubled_due_to_memory_sum,
--- a/sql/telemetry/clients_last_seen_v1/view.sql
+++ b/sql/telemetry/clients_last_seen_v1/view.sql
@@ -15,17 +15,17 @@ SELECT
   ) AS days_since_created_profile,
   CASE
   WHEN
-    BIT_COUNT(days_visited_5_uri_bits) & 0x0FFFFFFE = 0
+    BIT_COUNT(days_visited_5_uri_bits & 0x0FFFFFFE) = 0
   THEN
     'new_irregular_users_v1'
   WHEN
-    BIT_COUNT(days_visited_5_uri_bits) & 0x0FFFFFFE
+    BIT_COUNT(days_visited_5_uri_bits & 0x0FFFFFFE)
     BETWEEN 1
     AND 7
   THEN
     'semi_regular_users_v1'
   WHEN
-    BIT_COUNT(days_visited_5_uri_bits) & 0x0FFFFFFE
+    BIT_COUNT(days_visited_5_uri_bits & 0x0FFFFFFE)
     BETWEEN 8
     AND 27
   THEN
--- a/sql/telemetry/clients_last_seen_v1/view.sql
+++ b/sql/telemetry/clients_last_seen_v1/view.sql
@@ -34,15 +34,15 @@ SELECT
   WHEN
     4
   THEN
-    'regular_v1'
+    'regular_users_v1'
   WHEN
     0
   THEN
-    'new_irregular_v1'
+    'new_irregular_users_v1'
   ELSE
-    'semi_regular_v1'
+    'semi_regular_users_v1'
   END
-  AS segment_regular_users_v1,
+  AS segment_usage_regularity_v1,
   * EXCEPT (
     active_experiment_id,
     scalar_parent_dom_contentprocess_troubled_due_to_memory_sum,
--- a/sql/telemetry/clients_last_seen_v1/view.sql
+++ b/sql/telemetry/clients_last_seen_v1/view.sql
@@ -17,21 +17,21 @@ SELECT
   WHEN
     BIT_COUNT(days_visited_5_uri_bits & 0x0FFFFFFE) = 0
   THEN
-    'new_irregular_users_v1'
+    'new_irregular_users_v2'
   WHEN
     BIT_COUNT(days_visited_5_uri_bits & 0x0FFFFFFE)
     BETWEEN 1
     AND 7
   THEN
-    'semi_regular_users_v1'
+    'semi_regular_users_v2'
   WHEN
     BIT_COUNT(days_visited_5_uri_bits & 0x0FFFFFFE)
     BETWEEN 8
     AND 27
   THEN
-    'regular_users_v1'
+    'regular_users_v2'
   END
-  AS segment_usage_regularity_v1,
+  AS segment_usage_regularity_v2,
   * EXCEPT (
     active_experiment_id,
     scalar_parent_dom_contentprocess_troubled_due_to_memory_sum,
--- a/sql/telemetry/clients_last_seen_v1/view.sql
+++ b/sql/telemetry/clients_last_seen_v1/view.sql
@@ -14,33 +14,22 @@ SELECT
     SAFE.LOG(days_created_profile_bits & -days_created_profile_bits, 2) AS INT64
   ) AS days_since_created_profile,
   CASE
-    CAST(
-      BIT_COUNT(
-        days_visited_5_uri_bits & `moz-fx-data-shared-prod.udf.bitmask_range`(2, 6)
-      ) >= 1 AS INT64
-    ) + CAST(
-      BIT_COUNT(
-        days_visited_5_uri_bits & `moz-fx-data-shared-prod.udf.bitmask_range`(8, 7)
-      ) >= 2 AS INT64
-    ) + CAST(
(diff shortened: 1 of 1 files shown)


felixlawrence:
I found the README and tried running SQL formatter, but got a `ModuleNotFoundError: No module named 'dataclasses'`. In this venv, after a `pip install -r requirements.in` I can successfully `import mypy.plugins.dataclasses`, but still can't `import dataclasses`.

jklukas:
> I found the README and tried running SQL formatter, but got a `ModuleNotFoundError: No module named 'dataclasses'`. In this venv, after a `pip install -r requirements.in` I can successfully `import mypy.plugins.dataclasses`, but still can't `import dataclasses`.

You're likely on an older python version. The Docker container (which is how this runs in Airflow) uses python 3.8. You probably need at least python 3.7. And this should be documented if it currently is not.

felixlawrence:
Yeah, that's correct, I'm on 3.6. Will sort that out in the coming days - but I hope that shouldn't be a blocker for the feedback this PR needs most (a general tick of approval that this is a sensible thing to put into the ETL and in this way, and responses to my points 2-4).

udf/num_weeks_gte_2_from_bits.sql
```diff
@@ -0,0 +1,12 @@
+CREATE OR REPLACE FUNCTION
+  udf.num_weeks_gte_n_from_bits(b BYTES, i INT64) AS (
+    CAST(
+        BIT_COUNT(b & udf.bitmask_range(1, 7)) >= i AS INT64
+    ) + CAST(
+        BIT_COUNT(b & udf.bitmask_range(8, 7)) >= i AS INT64
+    ) + CAST(
+        BIT_COUNT(b & udf.bitmask_range(15, 7)) >= i AS INT64
+    ) + CAST(
+        BIT_COUNT(b & udf.bitmask_range(22, 7)) >= i AS INT64
```
jklukas: The 28 bits we have are generally thought to represent the current day (day 0) backward in time to day 27. Week 0 is days 0 through 6, etc. so that Week 3 is days 20 through 27.

So I would expect the offsets here to be 0, 7, 14, and 21 rather than 1, 8, 15, and 22.

jklukas: That is how we've been thinking about retention for GUD.

jklukas:
> Yeah, that's correct, I'm on 3.6. Will sort that out in the coming days - but I hope that shouldn't be a blocker for the feedback this PR needs most (a general tick of approval that this is a sensible thing to put into the ETL and in this way, and responses to my points 2-4).

Indeed this is not a blocker for giving feedback on this PR. This is blocked on me having headspace to think through it. I've glanced over it and what's happening here makes sense, but I need to do some thinking about whether there's a different way to express this and what the way forward would be to enable iteration on ideas like this without requiring a huge amount of bespoke operational work.

udf/num_weeks_gte_2_from_bits.sql
felixlawrence: When we partition clients into segments to study their data for some day, that day's data _shouldn't_ be used to determine which segment the client falls into. So if we want to look at today's DAU among "regular users v1", and today is day 0, then looking backwards we need to look at days 1 through 7, etc. Otherwise our segments can become trivial - "users who are highly engaged tend to be highly engaged" vs "users who have been highly engaged tend to stay highly engaged".

So, either in the analysis or the ETL we need to look one day backward from the day of interest.

I _think_ it's best to do this here in the ETL, because people using `clients_last_seen` to compute DAU would get the right behaviour from the naive query. People wanting to compute WAU or MAU, and people wanting to use other tables, would need to `LEFT JOIN clients_last_seen` and query the row from the correct `submission_date`, the date that the client first appears in their analysis.

Happy to discuss this point further via other channels if it's not clear here; it's an important subtlety and I would love to be surprised by a solution cleaner than what I've proposed.

felixlawrence:
[This is the document](https://docs.google.com/document/d/1eI7dXB6iYl_8gQ0wXdPiilLDe4bzAShdgigrDCqThvg/) I've circulated to the DS team; a PR for DTMO will be forthcoming, and wider socialisation would probably follow the segment successfully getting into the ETL, or at least mozanalysis.

udf/num_weeks_gte_2_from_bits.sql
felixlawrence: Wait, I'm wrong - `udf.bitmask_range(22, 7)` goes off the back of the array; we do the "look one day backward" by using `_previous.days_visited_5_uri_bits` rather than `_current.days_visited_5_uri_bits`. I'll update it.

jklukas:
Note that I've been focused mostly on operational work the last few days in response to emergent pipeline issues, which has prevented me from giving more feedback here. Here's hoping for next week.

udf/num_weeks_gte_2_from_bits.sql
felixlawrence: ...ah, I was originally right. `udf.bitmask_range` internally subtracts 1 from its input, raises an error when I try `udf.bitmask_range(0, 7)` and has a test case `assert_equals(1, udf.bitmask_range(1, 1))`

jklukas: Sorry for the confusion there. Indeed, this function uses ordinal indices to match the SUBSTR function and I had forgotten that detail.

jklukas:
The code style here looks reasonable. No need for significant changes there beyond running the `script/format_sql` formatter.

I left more commentary on the definition of windows here. If we adopt a definition that relies on the 28 bits already codified in clients_last_seen, then you can grab this segmentation at query time rather than having to build it into clients_last_seen and this PR would be unnecessary. There would still need to be more discussion about how to fit this into GUD.

felixlawrence:
I agree that GUD is a separate conversation; let's have it in [GUD issue #60](https://github.com/mozilla/GUD/issues/60) once things are further along.

OK, I'm going to split out the UDF into a separate PR, so that we can merge it separately (immediately improving the ease of using this segment, without being blocked by the schema change which presumably needs some backfill) and discuss here whether we add a segment column to `clients_last_seen`.

I think we _should_ add the column to clients last seen. My primary reason is that this formalises the segment definition in the ETL so decreases the likelihood of having different implementations of it.

Additionally, I do want to define the segment using days -1 to -28 (inclusive) as is effectively done in this pull request (by passing `_previous.days_visited_5_uri_bits` rather than `_current.days_visited_5_uri_bits`): the segment for a client on a given `submission_date` really shouldn't be defined using behavioural data from that `submission_date`; the better alternative would be to look at days -1 to -27 inclusive, but this adds a weirdness to the definition (a 6 day week!) that I would prefer to avoid.

Does this sound convincing? Is there anything you'd like to happen before I make a PR to add the column?

jklukas:
> I think we _should_ add the column to clients last seen. My primary reason is that this formalises the segment definition in the ETL so decreases the likelihood of having different implementations of it.

If we don't include day -28, we can still achieve this without needing to backfill the table by adding the field in the user-facing view rather than in the query. This allows us to refine the definition or add an additional definition instantaneously, though I expect you also care about the field being available in Spark, which this doesn't solve.

> Additionally, I do want to define the segment using days -1 to -28 (inclusive) as is effectively done in this pull request (by passing `_previous.days_visited_5_uri_bits` rather than `_current.days_visited_5_uri_bits`): the segment for a client on a given `submission_date` really shouldn't be defined using behavioural data from that `submission_date`; the better alternative would be to look at days -1 to -27 inclusive, but this adds a weirdness to the definition (a 6 day week!) that I would prefer to avoid.
> 
> Does this sound convincing? Is there anything you'd like to happen before I make a PR to add the column?

I do understand the argument that `submission_date` should be excluded, but I think the weirdness of a 6-day week first week is a preferable to having segmentation windows be one day out of line with the boundaries for WAU and DAU.

felixlawrence:
OK, I need to work out a few things and get back to you:
1. Does the segment become notably less powerful with a 6-day first week?
2. Does the segment become problematically complicated to explain to people if it has a 6-day first week?
3. How are we going to define MAU with regards to segments (does a user vanish from MAU once they no longer qualify to be in that segment - this is a discussion I'll be having with a few interested DSes), and what are the table columns that are most useful for this, which bleeds into the next point
4. Is it OK to say that the behavioural segment columns in `clients_last_seen` represent the client's segment _at the end of_ the day, and ask DSes to join with `submission_date` lagged by one day and `COALESCE` to treat nulls as "new_irregular_v1".

felixlawrence:
> 1. Does the segment become notably less powerful with a 6-day first week?

Not really - if we do this then there are about 1% fewer "regular users" and 2% more "new/irregular users". So this option is on the table. Will work through the other points (which are likely to take a few days of discussion with other DSes)

felixlawrence:
OK, if we tweak the segment to best use 27 days of data, then the definition becomes boutique enough that I don't think the UDF can pretend to be general - i.e. if we make a UDF then its sole purpose would be to accept `days_visited_5_uri_bits` and spit out the name of a segment.

Given that, is it worth creating a UDF, or should I just update this PR to stick the following code directly into the view for `clients_last_seen`:

```CASE
      CAST(
        BIT_COUNT(
          seg_cls.days_visited_5_uri_bits & `moz-fx-data-shared-prod`.udf.bitmask_range(2, 6)
        ) >= 1 AS INT64
      ) + CAST(
        BIT_COUNT(
          seg_cls.days_visited_5_uri_bits & `moz-fx-data-shared-prod`.udf.bitmask_range(8, 7)
        ) >= 2 AS INT64
      ) + CAST(
        BIT_COUNT(
          seg_cls.days_visited_5_uri_bits & `moz-fx-data-shared-prod`.udf.bitmask_range(15, 7)
        ) >= 2 AS INT64
      ) + CAST(
        BIT_COUNT(
          seg_cls.days_visited_5_uri_bits & `moz-fx-data-shared-prod`.udf.bitmask_range(22, 7)
        ) >= 2 AS INT64
      )
WHEN 4 THEN 'regular_v1'
WHEN 0 THEN 'new_irregular_v1'
ELSE 'semi_regular_v1'
END AS segment_regular_users_v1
```

jklukas:
Throwing the verbose definition into the `clients_last_seen` view sounds like a good plan here.

felixlawrence:
This PR is still blocked on positive review of the definitions and names from the DS team. That process is happening on and around [the firefox-data-docs PR](https://github.com/mozilla/firefox-data-docs/pull/428). I'll update this PR as ready for review once that piece is in place.

In the meantime, please let me know if I added the column in the right place in the right view, etc.

jklukas:
> In the meantime, please let me know if I added the column in the right place in the right view, etc.

Looks great. I'm happy to approve and merge whenever DS team gives the go-ahead.

felixlawrence:
We've settled on this segment definition, and they've gotten the tick from the DS team over [on DTMO](https://github.com/mozilla/firefox-data-docs/pull/428).

Let me know if I should click the "Update branch" button, or locally rebase on master then force push, or leave it to someone else.

felixlawrence:
Obsoleted by the v3 definitions.

PR #826, opened 2020-03-17
Use streaming inserts for shredder state

### LINKED ISSUE #815: Use streaming inserts instead of sql for shredder state updates

relud:
fixes #815


bigquery_etl/shredder/delete.py
```diff
@@ -354,39 +345,46 @@ def main():
...
-                ).strip()
-            ).result()
-        }
+        state_table_exists = False
+        try:
+            client.get_table(args.state_table)
+            state_table_exists = True
+        except NotFound:
+            if not args.dry_run:
+                client.create_table(
```
relud: this fixes dry run when using credentials that don't have permissions needed to create the state table

PR #827, opened 2020-03-17
Glam histogram aggregates table is partitioned by submission_date and clustered by client_id.

emtwo:
The main point here is to be able to query the previous day's `clients_histogram_aggregates` in intermediate steps. But it's also nice to keep a history of what this table looked like in the past rather than replacing it entirely on every run.

### DIFF
--- a/sql/telemetry_derived/clients_histogram_aggregates_v1/init.sql
+++ b/sql/telemetry_derived/clients_histogram_aggregates_v1/init.sql
@@ -1,5 +1,6 @@
 CREATE TABLE IF NOT EXISTS
   `moz-fx-data-shared-prod.telemetry_derived.clients_histogram_aggregates_v1` (
+    submission_date DATE,
     sample_id INT64,
     client_id STRING,
     os STRING,
@@ -19,5 +20,9 @@ CREATE TABLE IF NOT EXISTS
       aggregates ARRAY<STRUCT<key STRING, value INT64>>
     >>
 )
-PARTITION BY RANGE_BUCKET(sample_id, GENERATE_ARRAY(0, 100, 1))
-CLUSTER BY app_version, channel, client_id
+PARTITION BY submission_date
+CLUSTER BY sample_id, app_version, channel
+OPTIONS(
+  require_partition_filter=true,
+  partition_expiration_days=3
+);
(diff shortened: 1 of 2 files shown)


emtwo:
associated airflow PR: https://github.com/mozilla/telemetry-airflow/pull/907

emtwo:
Note: I tested that the partitioning + clustering continues to read as much data as it would if it were just partitioned

emtwo:
> How long of a history do you have in mind for this table? The`clients_histogram_aggregates_v1` table is currently ~50TB, which may be okay for a few copies of the table. Storing every submission date may not bring much value vs the cost of storing the data, which currently stands at [$0.02/GB per month](https://cloud.google.com/bigquery/pricing).

Thanks for pointing this out. I don't think we need to keep it for long. We could set a partition expiration for 3 days, let's say. I do really think this will be useful though for both debugging newly computed data and to avoid losing data by overwriting this table with incorrect data due to a bug, for example.

acmiyaguchi:
+1 on having a backup of the table. `mozaggregator` works in a similar way, where there are snapshots of the last 3 days for recovery situations. Is there a programmatic way to set the table expiration date, or would the mechanism for dropping old partitions need to be added to the airflow DAG?

emtwo:
https://cloud.google.com/bigquery/docs/managing-partitioned-tables#partition-expiration

We just need to set it once on the table and BQ handles it, looks like

acmiyaguchi:
It might be a good idea to set this in the query directly, to make sure the option is captured in code. According to the linked docs, something like this?

```sql
 ALTER TABLE telemetry_derived.clients_histogram_aggregates_v1
 SET OPTIONS (
    partition_expiration_days=3
 )
 ```

**Comment by emtwo** [comment]
@acmiyaguchi added it to the `init.sql` file

File: sql/telemetry_derived/clients_histogram_aggregates_v1/init.sql [hunk 1]
```diff
@@ -19,5 +20,12 @@ CREATE TABLE IF NOT EXISTS
       aggregates ARRAY<STRUCT<key STRING, value INT64>>
     >>
 )
-PARTITION BY RANGE_BUCKET(sample_id, GENERATE_ARRAY(0, 100, 1))
-CLUSTER BY app_version, channel, client_id
+PARTITION BY submission_date
+CLUSTER BY sample_id, app_version, channel
+OPTIONS(require_partition_filter=true);
```
acmiyaguchi: If a new table is going to be created, the `partition_expiration_days=3` option can probably be moved into this section.

emtwo: Yes, I'll be creating a new table, so I'll move this into the existing options.

PR #828, opened 2020-03-17
Standardize common script arguments

relud:


jklukas:
I totally missed this review request. Looking now.

bigquery_etl/util/standard_args.py
```diff
@@ -0,0 +1,179 @@
...
+        help="One or more billing projects over which bigquery jobs should be "
+        "distributed",
+    )
+
+
+def add_dry_run(parser, debug_log_queries=True):
+    """Add argument for dry run."""
+    add_argument(
+        parser,
+        "-n",
```
jklukas: Why does `-n` mean dry run?

relud: This was chosen because of the `-n` flag in `gsutil rsync`:
```
  -n             Causes rsync to run in "dry run" mode, i.e., just outputting
                 what would be copied or deleted without actually doing any
                 copying/deleting.
```

I think it could be short for `noop`, but mostly it's because `-d` seemed like it should be reserved for `--date`.

I'd also be find removing the short flag from this, if you would prefer.

jklukas: I'd prefer to remove the short flag. I was expecting it to be related to the `-n` option in `bq query` that limits the number of returned rows.

PR #829, opened 2020-03-17
Add histogram percentiles for Fenix data into GLAM

acmiyaguchi:
This implemented `fenix_histogram_percentiles_v1`. The table seems to work, but I have no idea if the reported aggregates are valid or not. I'll need to figure out how to generate tests for this, because I'm confident of the correctness.


udf_js/glean_percentile.sql
```diff
@@ -0,0 +1,51 @@
...
+      acc += normalized[i];
+      index = i;
+      if (acc >= percentile / 100) {
+          break;
+      }
+  }
+
+  // NOTE: we do not perform geometric or linear interpolation, but this would
+  // be the place to implement it.
+  return histogram[index].key;
```
acmiyaguchi: @mdboom I decided to rewrite the function to calculate percentiles without interpolation of the histogram values. I was wondering if you had any opinions on this? Note that this can be changed at a later time, since this is near the end of the transformation pipeline.

mdboom: I think this is fine.  One of the design decisions was to make sure the buckets are never so wide as to lose a lot of meaningful resolution.

However, I feel like this is fundamentally a DS question.  Corey Howgelund worked on the design of these distributions in the past.

acmiyaguchi: Thank you, I ended up filing a [Jira Issue [DS-430]](https://jira.mozilla.com/browse/DS-430) to figure out the best way to handle this.

acmiyaguchi:
I've filed an bug for implementing a weighted quantile function (https://bugzilla.mozilla.org/show_bug.cgi?id=1624492).

acmiyaguchi:
I filed an issue for the unrelated CI failure in #852

Issue #831, opened 2020-03-17
Variable assignment inside conditional under histogram_percentiles_v1


acmiyaguchi:
https://github.com/mozilla/bigquery-etl/blob/029b06e342d36312c928054824634b5317b41dd6/sql/telemetry_derived/histogram_percentiles_v1/query.sql#L36

I was reading through the source for `udf_percentile`. The line above looks like a typo, which may be producing incorrect values for linear histogram percentiles.

@emtwo

emtwo:
Oh good catch! For reference, this was the original function that was reused: https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/v2/telemetry.js#L90

PR #832, opened 2020-03-18
Refactor udf_percentile in telemetry histogram percentiles as a persistent udf 

acmiyaguchi:
The user defined function for calculating histogram percentiles is complex enough to warrant tests. I have been adapting this function for calculating percentiles for #829, but I'm not sure how to determine whether the results look sane. By moving the reference udf into it's own module, it can be unit-tested. This should also determine whether #831 is actually an issue.

I've decided to leave the logic as is, and document a few test cases with what I expect the values to be and what the actual value is for a few medians.


acmiyaguchi:
@relud Is this the right way to factor out a javascript udf from the sql source, or does this have to be done in two separate PRs to make dryrun succeed?

relud:
> @relud Is this the right way to factor out a javascript udf from the sql source

yes

> does this have to be done in two separate PRs to make dryrun succeed?

to make the dry run succeed without multiple PRs: wait for this to get approved, manually publish the UDF, manually re-run the failed dry-run-sql in CI

tests/assert/approx_equals.sql
```diff
@@ -0,0 +1,20 @@
...
+        'Expected ',
+        TO_JSON_STRING(ROUND(expected, 2)),
+        ' ± ',
+        TO_JSON_STRING(tolerance),
+        ' but got ',
+        TO_JSON_STRING(ROUND(actual, 2))
+      )
+    )
+  )
+);
```
relud:
```suggestion
);

-- Tests
SELECT
  assert_approx_equals(1, 1, 0),
  assert_approx_equals(1, 2, 1);
  
#xfail
SELECT
  assert_approx_equals(1,2,0.9);
```

acmiyaguchi: Ah, I should have figured that you could write tests for the test helpers. TIL about the `xfail` directive in the testing framework, that'll come in handy :+1:.

udf_js/telemetry_percentile.sql
```diff
@@ -0,0 +1,113 @@
...
+  }
+  percentileBucketIndex--;
+
+  //Undo the last loop iteration where we overshot
+  hitsAtPercentileInBar += histogram[percentileBucketIndex].value;
+
+  // The ratio of the hits in the percentile to the hits in the bar containing it.
+  // How far we are inside the bar?
+  let ratioInBar = hitsAtPercentileInBar / histogram[percentileBucketIndex].value;
+  if (type === "histogram-exponential") {
```
emtwo: thank you!

PR #833, opened 2020-03-18
Add submission_date filter where needed.

emtwo:
The filter was missing from two queries and also the date param did not need to be cast to date


emtwo:
associated airflow pr: https://github.com/mozilla/telemetry-airflow/pull/909

emtwo:
> r+
> 
> I didn't catch that downstream jobs would be affected. It'd be nice if there were a better way to test the dag before running it in prod.

I agree. I could add a script that mimics the runs airflow would do but on a small subset of the data.

Issue #834, opened 2020-03-18
copy_deduplicate failing due to temporary beam table


jklukas:
This problem happened once yesterday and now twice today. See [Airflow log for one of the failed tasks](https://workflow.telemetry.mozilla.org/log?task_id=copy_deduplicate_all&dag_id=copy_deduplicate&execution_date=2020-03-17T01%3A00%3A00%2B00%3A00).

The `copy_deduplicate_all` task fails trying to process a table named like:

    moz-fx-data-shared-prod.telemetry_live.beam_load_sink0031115053449eb3b40_d5cf2ea9e0d24625bc9c5091104cec83_cd7a2b84d40165d52a9286ed58fdbdc9_00001_00026

I think this problem is related to issues we've been having with the Beam sink getting stuck which started a few days ago. See in particular https://bugzilla.mozilla.org/show_bug.cgi?id=1623000#c4

I think we should add some logic in copy_deduplicate to ignore these temporary tables, though it's not entirely clear to me how best to go about it. We could filter based on length of the table name, or look for the exact prefix 'beam_load_sink'.

jklukas:
cc @whd @scholtzan @relud

jklukas:
I've merged a basic fix for this, but I'm open to better solutions.

Also, note that while yesterday these tables were transient, we have a whole bunch of them that are sitting around long-term as of today. Likely related to a Dataflow job failing to make progress.

PR #835, opened 2020-03-18
Filter out temporary tables created by the Beam BigQuery sink job

jklukas:
Closes https://github.com/mozilla/bigquery-etl/issues/834


jklukas:
The dry-run failure is due to `clients_histogram_aggregates_v1` and is unrelated to these changes.

PR #839, opened 2020-03-18
Publish query results as JSON

scholtzan:
I started looking into making query results available as JSON. 

When a query gets executed and the corresponding metadata file has `public_json=true` set, results will need to get uploaded as JSON to Cloud Storage. 
For queries that are not `incremental` the entire destination table gets exported (for example `.../telemetry/ssl_ratios/v1/files/...`). For queries that are marked as `incremental` in their metadata, only query results for the current submission_date need to get uploaded (for example to a destination like `.../telemetry/clients_daily/v1/files/2019-07-01/`).

Since only entire tables can be exported, for incremental queries results are written into a temporary table.


script/entrypoint
```diff
@@ -32,7 +32,11 @@ elif [ "$1" = "query" ]; then
     # For queries that have a corresponding metadata file with public_bigquery set to
     # True, the destination_table option will be mapped to the destination table in
     # the public dataset.
-    exec script/run_query --query_file="${@: -1}" ${@:1:$#-1}
+    python script/run_query --query_file="${@: -1}" ${@:1:$#-1}
+
+    # For queries with public_json set in the metadata, the results will be
+    # exported as JSON to Cloud storage.
+    exec script/publish_public_data_json --query_file="${@: -1}" ${@:1:$#-1}
```
scholtzan: This could be problematic, I think. It would be nice if first the script that runs the query and then the script that checks and publishes results as JSON could be executed. 
It would be possible to move all of the logic to `run_query`, but since that is quite a bit of code it might get harder to maintain. Having a wrapper script that calls both scripts would probably also work. I'm not sure what the best way here is.

script/publish_public_data_json
```diff
@@ -0,0 +1,144 @@
...
+from google.cloud import bigquery
+
+# sys.path needs to be modified to enable package imports from parent
+# and sibling directories. Also see:
+# https://stackoverflow.com/questions/6323860/sibling-package-imports/23542795#23542795
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+from bigquery_etl.parse_metadata import Metadata  # noqa E402
+
+
+DEFAULT_BUCKET = "ascholtz-dev-test"  # todo
```
scholtzan: Currently, the bucket for public data doesn't exist. I'll open a ticket for this.

script/publish_public_data_json
```diff
@@ -0,0 +1,144 @@
...
+METADATA_FILE = "metadata.yaml"
+SUBMISSION_DATE_RE = re.compile(r"^submission_date:DATE:(\d\d\d\d-\d\d-\d\d)$")
```
scholtzan: This is the format of `submission_timestamp` I found in Airflow.

script/publish_public_data_json
```diff
@@ -0,0 +1,155 @@
...
+    else:
+        # if date exists, then query is incremental and newest results are exported
+        destination_uri += f"{date}/*.json"
+
+    table_ref = client.get_table(result_table)
+
+    job_config = bigquery.ExtractJobConfig()
+    job_config.destination_format = "NEWLINE_DELIMITED_JSON"
+    extract_job = client.extract_table(
+        table_ref, destination_uri, location="US", job_config=job_config
```
scholtzan: I'm guessing location will always be "US"?

script/publish_public_data_json
```diff
@@ -0,0 +1,155 @@
...
+import os
+import sys
+import re
+
+from google.cloud import bigquery
+
+# sys.path needs to be modified to enable package imports from parent
+# and sibling directories. Also see:
+# https://stackoverflow.com/questions/6323860/sibling-package-imports/23542795#23542795
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
fbertsch: Maybe it's time to reconsider the directory structure.

script/publish_public_data_json
```diff
@@ -0,0 +1,155 @@
...
+        # if date is not set, then export the entire table and overwrite existing JSON
+        destination_uri += "*.json"
+    else:
+        # if date exists, then query is incremental and newest results are exported
+        destination_uri += f"{date}/*.json"
+
+    table_ref = client.get_table(result_table)
+
+    job_config = bigquery.ExtractJobConfig()
+    job_config.destination_format = "NEWLINE_DELIMITED_JSON"
```
fbertsch: The spec specifies we use JSON, not ndjson. The reason for this is it's easier for non-engineer types to use, and is easily viewable in the browser. Unfortunately that means we'll need to pull these files down and transform them.

script/publish_public_data_json
```diff
@@ -0,0 +1,155 @@
...
+parser.add_argument("--parameter", help="Query parameters, such as submission_date")
+parser.add_argument("--query_file", help="File path to query to be executed")
+
+
+def publish_table_as_json(
+    bucket, client, dataset, table, version, result_table, date=None
+):
+    """Export the `result_table` data as JSON to Cloud Storage."""
+    # "*" makes sure that files larger than 1GB get split up into multiple JSON files
+    destination_uri = f"gs://{bucket}/{dataset}/{table}/{version}/files/"
```
fbertsch: This needs to be: `/api/{api_version}/tables/{dataset}/{table}/{version}/files/`

We may want to remove the table version, and have it simply be e.g. `ssl_ratios_v1`, similar to our directory layout.

script/publish_public_data_json
```diff
@@ -0,0 +1,155 @@
...
+    client = bigquery.Client(args.project_id)
+    date_search = re.search(SUBMISSION_DATE_RE, args.parameter)
+
+    if date_search:
+        date = date_search.group(1)
+
+    (dataset, table, version) = dataset_table_version_from_file(args.query_file)
+
+    if metadata.is_incremental():
+        if date is not None:
```
fbertsch: nit: can we check for `date is None`, error out, and then have everything following be un-indented?

script/entrypoint
jklukas: When we get to the point of having DAGs generated within this repo, it would indeed to nice to separate out the steps within the DAG. For the time being, it looks like the approach here is attempting to satisfy the spirit of "all you need to do to publish a query is to set public to true in the metadata". If we continue to see that as a requirement, then we have to handle it all within this single invocation of `entrypoint`.

I'm in favor of the approach captured here, though I dislike the asymmetry of invoking the second script via `exec` and not the first one. `exec` is nice for the simple case where the entrypoint is dispatching to a single script, but I think we should drop it on both invocations here.

script/publish_public_data_json
jklukas: Long-term, I think this repo is going to need a significant refactor where we pull all the python logic inside the `bigquery_etl` module and the scripts become light entrypoints into that codebase. But that's definitely outside the scope of this PR.

fbertsch:
There's also the issue of testing. Anthony and I have had success testing Python scripts, even ones that are doing convoluted work with APIs. See some examples [here](https://github.com/mozilla/mozilla-schema-generator/blob/master/tests/test_scripts.py) and [here](https://github.com/mozilla/leanplum-data-export/tree/master/tests).

script/entrypoint
fbertsch: Unfortunately we're not at the point right now where it's ever just "specify in the metadata", since this still needs to get added to Airflow anyways. I also think there's a halfway point between "generate dags in this repo" and "run everything as a script block". At it's simplest we could specify  this as a new util that first runs queries and then publishes. Later we could merge that back into generated DAGs.

scholtzan: Yeah, it might be more convenient to call this script from Airflow. I'll look into adding it as a util which gets executed after the query.

scholtzan: I added something in Airflow which might work for this: https://github.com/mozilla/telemetry-airflow/pull/911

scholtzan:
I added some tests which execute the `publish_public_data_json` script and check if data is written to Cloud Storage. Running these tests in the CI causes some issues due to missing permissions. Not sure what the best way forward is...

tests/publish_public_data_json/test_publish_public_data_json.py
```diff
@@ -0,0 +1,168 @@
...
+import subprocess
+
+from google.cloud import bigquery
+from google.cloud import storage
+from google.api_core.exceptions import NotFound
+
+
+class TestPublishJson(object):
+    test_bucket = "moz-fx-data-stage-bigquery-etl"
+    project_id = "moz-fx-data-shar-nonprod-efed"
```
scholtzan: Do we have a project for running tests?

fbertsch: We do not. It usually makes sense to mock the API and check that files are written correctly using [patch](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch).

I do like the idea of running tests in a test project. It should be acceptable from a privacy perspective if everything we write there comes from our tests.

script/publish_public_data_json
```diff
@@ -0,0 +1,191 @@
...
+            args.api_version,
+            dataset,
+            table,
+            version,
+            temp_result_table,
+            date,
+        )
+
+        # remove the temporary table after it has been exported
+        client.delete_table(temp_result_table)
```
fbertsch: If `publich_table_as_json` fails, we'll end up the tmp table still available in bq. One alternative could be to wrap the temp table pieces into an object with `__enter__` for setup and `__exit__` for teardown, using `with` here.

script/publish_public_data_json
```diff
@@ -0,0 +1,191 @@
...
+
+def write_results_to_temp_table(client, project, query_file, date):
+    """
+    Write the results of the query to a temporary table and return the table
+    name.
+    """
+    (dataset, table, version) = dataset_table_version_from_file(query_file)
+
+    table_date = date.replace("-", "")
+    temp_table = f"{project}.{dataset}.{table}_{version}_{table_date}_temp"
```
fbertsch: Temp table should be written to the `tmp` dataset.

tests/publish_public_data_json/test_publish_public_data_json.py
```diff
@@ -0,0 +1,168 @@
...
+        blob_len = 0
+
+        for blob in blobs:
+            content = json.loads(blob.download_as_string().decode("utf-8").strip())
+            blob_len += 1
+            assert len(content) == 3
+
+        assert blob_len == 1
+
+    # todo: test splitting of files larger than 1GB
```
fbertsch: 😬

script/publish_public_data_json
```diff
@@ -0,0 +1,191 @@
...
+
+
+def gcp_convert_ndjson_to_json(bucket, gcp_path):
+    """Converts ndjson files on GCP to json files."""
+    storage_client = storage.Client()
+
+    blobs = storage_client.list_blobs(bucket, prefix=gcp_path)
+
+    for blob in blobs:
+        content = blob.download_as_string().decode("utf-8").strip()
```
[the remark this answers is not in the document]
jklukas: The GCS API has everything we need to stream the contents in and stream contents out to a new object without pulling down the entire object to memory, so I'm hoping we can avoid needing high-mem instances.

I have not done this in the python API, but it should be possible. In particular, we should be able to replace `\n` in the input stream with `,\n`, add a `[` character at the beginning and end of the stream.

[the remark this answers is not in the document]
scholtzan: Alright, I'll look into it.

tests/publish_public_data_json/test_publish_public_data_json.py
scholtzan: Mocking seems to work well when testing separate methods. I had some troubles testing the individual methods in the script since the script itself cannot be imported as module.

script/publish_public_data_json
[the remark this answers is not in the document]
scholtzan: Thinking about this a bit more, wouldn't it also be easier for consumers to read ndjson instead of json? That way streaming of json objects would be easier?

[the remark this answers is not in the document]
fbertsch: I understand that thought process. This format was agreed on across across the data team because we expect consumers will not be data engineers but researchers and data scientists, who are not usually exposed to ndjson.

tests/publish_public_data_json/test_publish_public_data_json.py
fbertsch: Hmm, another reason to refactor some of this. It could be worth it to approach this script as an entry into the `bigquery_etl` package, and the tests run pieces there.

bigquery_etl/publish_json.py
```diff
@@ -0,0 +1,148 @@
...
+        bucket = self.storage_client.bucket(self.target_bucket)
+
+        for blob in blobs:
+            blob_path = f"gs://{self.target_bucket}/{blob.name}"
+            tmp_blob = bucket.blob(blob.name + ".tmp")
+            tmp_blob_name = blob_path + ".tmp"
+
+            # stream from GCS
+            with smart_open.open(blob_path) as fin:
+                with smart_open.open(tmp_blob_name, "w") as fout:
```
scholtzan: The [smart_open](https://github.com/RaRe-Technologies/smart_open) library has support for streaming from and to files in GCS. Doing this using this library was much easier than reproducing the streaming stuff.

conftest.py
```diff
@@ -8,3 +11,18 @@
...
+
+def pytest_collection_modifyitems(config, items):
+    keywordexpr = config.option.keyword
+    markexpr = config.option.markexpr
+    if keywordexpr or markexpr:
+        return
+
+    skip_publish_json_script = pytest.mark.skip(
+        reason='publish_json_script marker not selected'
+    )
```
scholtzan: I decided to add a custom marker to prevent running the tests that test the `publish_public_data_json` script in the CI. To run those tests locally that require access to GCP: `pytest -m publish_json_script`

script/entrypoint
```diff
@@ -28,11 +28,11 @@ elif [ "$1" = "query" ]; then
...
-    #
-    # For queries that have a corresponding metadata file with public_bigquery set to
-    # True, the destination_table option will be mapped to the destination table in
-    # the public dataset.
-    exec script/run_query --query_file="${@: -1}" ${@:1:$#-1}
+    python script/run_query --query_file="${@: -1}" ${@:1:$#-1}
+
+    # For queries with public_json set in the metadata, the results will be
+    # exported as JSON to Cloud storage.
+    python script/publish_public_data_json --query_file="${@: -1}" ${@:1:$#-1}
```
scholtzan: See https://github.com/mozilla/bigquery-etl/pull/839#discussion_r395073609

scholtzan:
This is ready for another look.

Content from GCS is now streamed when converting ndjson to json. I also made changes to tests to mock the GCS API to run these tests in CI.

conftest.py
fbertsch: I like the custom marker but don't like that it doesn't run in CI. We could unintentionally run into a situation where tests fail but CI doesn't catch it. Can we add integration testing with proper service accounts in CI?

bigquery_etl/publish_json.py
fbertsch: Good choice, we used smart_open on AWS and it's a nice lib.

bigquery_etl/publish_json.py
```diff
@@ -0,0 +1,148 @@
...
+                    fout.write("[\n")
+
+                    first_line = True
+
+                    for line in fin:
```
fbertsch: nit: simpler to enumerate and use index to find the first line.

bigquery_etl/publish_json.py
```diff
@@ -0,0 +1,148 @@
...
+                        if not first_line:
+                            fout.write(",\n")
+
+                        first_line = False
+                        fout.write(line.replace("\n", ""))
+
+                    fout.write("]")
+
+            bucket.rename_blob(tmp_blob, blob.name)
```
fbertsch: This may end up a bit complicated but we _should_ gzip these files. Reading through a bit it looks like [smart_open can handle that](https://github.com/RaRe-Technologies/smart_open#supported-compression-formats), but we may need to [set the `Content-Encoding`](https://cloud.google.com/storage/docs/transcoding#content-type_vs_content-encoding).

bigquery_etl/publish_json.py
```diff
@@ -0,0 +1,148 @@
...
+        extract_job.result()
+
+        self._gcp_convert_ndjson_to_json(prefix)
+
+    def _gcp_convert_ndjson_to_json(self, gcp_path):
+        """Convert ndjson files on GCP to json files."""
+        blobs = self.storage_client.list_blobs(self.target_bucket, prefix=gcp_path)
+        bucket = self.storage_client.bucket(self.target_bucket)
+
+        for blob in blobs:
```
fbertsch: We need to ensure this is an atomic operation, such that if we fail at any point during any of these blobs, it looks as if no data is processed. In that case we should backup the data to a saved location, write the data to a staging location, overwrite the prod data with that staging data, then delete the backup data. With this scenario only time we may see an atomicity fail is during overwriting the prod data with stage data, so we would have to check for that kind of failure and either write the backups (in case of non-incremental queries) or delete the data entirely for that day (in case of incremental queries). Let me know if you want to chat about this.

bigquery_etl/publish_json.py
```diff
@@ -0,0 +1,148 @@
...
+import smart_open
+
+import os
+import sys
+import re
+
+# sys.path needs to be modified to enable package imports from parent
+# and sibling directories. Also see:
+# https://stackoverflow.com/questions/6323860/sibling-package-imports/23542795#23542795
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
fbertsch: We can't change this now that we've moved to the `bigquery_etl` dir?

jklukas:
I glanced over this again and it's looking really good. I'll plan to do a more thorough review tomorrow.

conftest.py
scholtzan: I'm all for it. We could have a GCP project specifically intended for running integration tests that doesn't have access to any actual data. I guess setting up such a project will require ops to do it.

jklukas: I set up a sandbox account for gcp-ingestion's CI tests without ops involvement. We could do the same here, and we can perhaps follow up with ops about whether they want to own these projects instead. We probably don't want to block this work on ops availability to provision the things. Feel free to ask me questions about how gcp-ingestion's CI tests and creds are set up.

bigquery_etl/publish_json.py
```diff
@@ -0,0 +1,166 @@
...
+    def _gcp_convert_ndjson_to_json(self, gcs_path):
+        """Convert ndjson files on GCP to json files."""
+        blobs = self.storage_client.list_blobs(
+            self.target_bucket, prefix=self.stage_gcs_path
+        )
+        bucket = self.storage_client.bucket(self.target_bucket)
+
+        for blob in blobs:
+            blob_path = f"gs://{self.target_bucket}/{blob.name}"
+            tmp_blob_name = blob_path + ".tmp.gz"
```
scholtzan: By setting the `.gz` file ending smart_open automatically gzips the content.

bigquery_etl/publish_json.py
```diff
@@ -0,0 +1,166 @@
...
+        tmp_blobs = self.storage_client.list_blobs(
+            self.target_bucket, prefix=self.stage_gcs_path
+        )
+
+        for tmp_blob in tmp_blobs:
+            # only copy gzipped files to target directory
+            if "tmp.gz" in tmp_blob.name:
+                # remove .tmp from the final file name
+                file_name = tmp_blob.name.split("/")[-1].replace("tmp.", "")
+                bucket.rename_blob(tmp_blob, gcs_path + file_name)
```
scholtzan: It seems like each file has to be moved separately. I tried to rename the entire directory but it didn't work (seems like directories aren't valid blobs)

conftest.py
scholtzan: Sounds good. I'll take a look tomorrow.

.circleci/config.yml
```diff
@@ -43,6 +43,30 @@ jobs:
...
+    - &pytest_integration_test
+      run:
+        name: PyTest Integration Test
+        # Google's client libraries will check for GOOGLE_APPLICATION_CREDENTIALS
+        # and use a file in that location for credentials if present;
+        # See https://cloud.google.com/docs/authentication/production
+        command: |
+          export GOOGLE_APPLICATION_CREDENTIALS="/tmp/gcp.json"
+          echo "$GCLOUD_SERVICE_KEY" > "$GOOGLE_APPLICATION_CREDENTIALS"
+          PATH="venv/bin:$PATH" script/entrypoint -m integration
```
scholtzan: I took the gcp-ingestion setup for integration tests as an example for this

conftest.py
scholtzan: I set it up for CI.

scholtzan:
Talking to Jesse, he is interested in using this for publishing data related to the covid work. In addition to publishing the data in BigQuery.

scholtzan:
This should be ready for another look

bigquery_etl/publish_json.py
```diff
@@ -0,0 +1,166 @@
...
+            blob_path = f"gs://{self.target_bucket}/{blob.name}"
+            tmp_blob_name = blob_path + ".tmp.gz"
+
+            # stream from GCS
+            with smart_open.open(blob_path) as fin:
+                with smart_open.open(tmp_blob_name, "w") as fout:
+                    fout.write("[\n")
+
+                    for i, line in enumerate(fin):
+                        if i > 0:
```
jklukas:
```suggestion
                        # skip the first line, which is a header
                        if i > 0:
```

bigquery_etl/publish_json.py
```diff
@@ -0,0 +1,166 @@
...
+            prefix += f"{self.date}/"
+
+        table_ref = self.client.get_table(result_table)
+
+        job_config = bigquery.ExtractJobConfig()
+        job_config.destination_format = "NEWLINE_DELIMITED_JSON"
+
+        # "*" makes sure that files larger than 1GB get split up into JSON files
+        # files are written to a stage directory first
+        destination_uri = f"gs://{self.target_bucket}/" + self.stage_gcs_path + "*.json"
```
fbertsch:
```suggestion
        destination_uri = f"gs://{self.target_bucket}/" + self.stage_gcs_path + "*.ndjson"
```

bigquery_etl/publish_json.py
fbertsch: We'd need to `s/ndjson/json` here with the above change.

bigquery_etl/publish_json.py
fbertsch: There is still a failure scenario here where we end up with incomplete data in the final bucket. It might be worth it to add some logging in here that will transmit to Airflow so if something happens it's easy to find out on which object.

PR #842, opened 2020-03-19
Construct non-null join key for cfr_users_last_seen query

jklukas:
Closes https://github.com/mozilla/bigquery-etl/issues/841

cc @ncloudioj


ncloudioj:
@jklukas LGTM. Thanks for the fix!

jklukas:
Backfilling right now and the MAU number is looking much more reasonable. Should be done in ~1 hour.

PR #843, opened 2020-03-20
Add incline exec dash submission_date based query

fbertsch:


sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
```diff
@@ -0,0 +1,141 @@
+CREATE TEMP FUNCTION bucket_manufacturer(manufacturer STRING) AS (
+  IF(
+    manufacturer IN UNNEST(['samsung', 'huawei', 'xiaomi', 'lge', 'motorola', 'sony', 'google', 'oppo', 'oneplus']),
```
jklukas:
```suggestion
    manufacturer IN ('samsung', 'huawei', 'xiaomi', 'lge', 'motorola', 'sony', 'google', 'oppo', 'oneplus'),
```

I expect this simpler syntax will work.

sql/org_mozilla_firefox/migrated_clients_v1/view.sql
```diff
@@ -0,0 +1,65 @@
+CREATE OR REPLACE VIEW `moz-fx-data-shared-prod.org_mozilla_firefox.migrated_clients_v1` AS
```
jklukas: We should remove the _v1 from the view name unless there's a compelling reason to version it per https://github.com/mozilla/bigquery-etl#views

sql/org_mozilla_firefox/migrated_clients_v1/view.sql
```diff
@@ -0,0 +1,65 @@
...
+    normalized_channel
+)
+
+SELECT
+  *
+FROM
+  clients
+WHERE
+  fenix_client_id != 'c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0'
+  AND fennec_client_id != 'c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0'
```
jklukas: This view looks expensive enough that it should be materialized to a table. We're likely fine for current ping volumes, but this could explode with release data.

The logic here seems similar to the `clients_first_seen_v1` table (which has not actually been scheduled) which additionally makes sure that a given client_id appears only once in the entire history of the table by depending on past days.

sql/org_mozilla_firefox/migrated_clients_v1/view.sql
```diff
@@ -0,0 +1,65 @@
...
+    submission_date,
+    normalized_channel,
+    `moz-fx-data-shared-prod.udf.mode_last`(ARRAY_AGG(fennec_client_id IGNORE NULLS)) AS fennec_client_id,
+    COUNT(*) AS migration_ping_count
+  FROM
+    migrations
+  GROUP BY
+    fenix_client_id,
+    submission_date,
+    normalized_channel
```
jklukas: I think we should avoid grouping by channel and instead use mode_last. It's unlikely we'll see the same id show up in two different apps, but stranger things have happened.

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
```diff
@@ -0,0 +1,141 @@
...
+    `moz-fx-data-shared-prod.udf.active_n_weeks_ago`(days_created_profile_bits,
+      1) AS new_last_week
+  FROM
+    `moz-fx-data-shared-prod.telemetry_derived.core_clients_last_seen_v1` clients_last_seen
+  LEFT JOIN
+    `moz-fx-data-shared-prod.org_mozilla_firefox.migrated_clients_v1` migrated_clients
+    ON clients_last_seen.client_id = migrated_clients.fennec_client_id
+  WHERE
+    AND clients_last_seen.submission_date = @submission_date
+    AND migrated_clients.submission_date >= @submission_date
```
jklukas: Is this supposed to be `<=`? What is the purpose of this filter?

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
```diff
@@ -0,0 +1,141 @@
...
+    country IN UNNEST(['US', 'CA', 'DE', 'IN', 'FR', 'CN', 'IR', 'BR', 'IE', 'GB', 'ID']),
+    [country, 'Tier 1'],
+    ['Other'])
+);
+
+WITH
+  fennec_client_info AS (
+  SELECT
+    clients_last_seen.submission_date AS date,
+    IF(migrated_clients.fenix_client_id IS NOT NULL, 'Yes', 'No') AS is_migrated,
```
jklukas: Can this be a boolean?

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
```diff
@@ -0,0 +1,141 @@
...
+    baseline.device_manufacturer AS manufacturer,
+    baseline.country,
+    `moz-fx-data-shared-prod.udf.active_n_weeks_ago`(baseline.days_seen_bits,
+      0) AS active_this_week,
+    `moz-fx-data-shared-prod.udf.active_n_weeks_ago`(baseline.days_seen_bits,
+      1) AS active_last_week,
+    DATE_DIFF(clients_last_seen.submission_date, baseline.first_run_date, DAY) BETWEEN 0 AND 6  AS new_this_week,
+    DATE_DIFF(clients_last_seen.submission_date, baseline.first_run_date, DAY) BETWEEN 7 AND 13 AS new_last_week
+  FROM
+    `moz-fx-data-shared-prod.org_mozilla_firefox_derived.clients_last_seen_v1` clients_last_seen
```
jklukas: I'd recommend using the `org_mozilla_firefox.clients_last_seen` view instead, which will avoid you having to descend into `baseline`.

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
```diff
@@ -0,0 +1,141 @@
...
+  CROSS JOIN
+    UNNEST(['Overall', NULL]) AS is_migrated_group
+  CROSS JOIN
+    UNNEST(['Overall', NULL]) AS channel_group
+  CROSS JOIN
+    UNNEST(['Overall', NULL]) AS manufacturer_group
+  CROSS JOIN
+    UNNEST(['Overall', NULL]) AS country_group
+  CROSS JOIN
+    UNNEST(bucket_country(country)) AS bucketed_country
```
jklukas: I don't understand what these joins for defining groups are doing. Can we add some commentary?

sql/org_mozilla_firefox_derived/migrated_clients_v1/query.sql
```diff
@@ -0,0 +1,42 @@
...
+    ) AS fennec_client_id,
+    COUNT(*) AS migration_ping_count
+  FROM
+    migrations
+  GROUP BY
+    fenix_client_id,
+    submission_date,
+    normalized_channel
+)
+SELECT
```
fbertsch: Note that we are taking the _latest_ value found from the migration ping, to account for migration retries and migrations that span multiple submission_dates. This will require overwriting the entire table every day, and a slight change to the bigquery_etl airflow operator.

sql/org_mozilla_firefox_derived/migrated_clients_v1/metadata.yaml
```diff
@@ -0,0 +1,15 @@
...
+  channels. Every client will have at most one entry, and if a 
+  Fennec client is not present then it indicates that they have
+  not migrated. Fennec Client ID, channel, and submission_date
+  are taken from the _last_ value seen for that client.
+owners:
+  - frank@mozilla.com
+labels:
+  application: fenix
+  schedule: daily
+  incremental: true
```
fbertsch: I'm wondering if this is indeed incremental.

sql/org_mozilla_firefox/migration/view.sql
```diff
@@ -0,0 +1,23 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.org_mozilla_firefox.migration`
```
fbertsch: This will overwrite the current migration ping table for Fennec release. That seems fine for us, but open to alternatives.

jklukas: I would argue this view definition is dangerous. There's no guarantee that these schemas will stay in sync across the different app IDs. In particular, it wasn't possible to make a simple view on top of baseline pings that unions apps because there are a few fields that aren't present everywhere.

jklukas: I think we should verbosely do a union with just the fields we need within the migrated_clients query rather than making this a view that unions them.

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
jklukas: The descriptions here are very helpful.

sql/org_mozilla_firefox_derived/migrated_clients_v1/init.sql
```diff
@@ -0,0 +1,46 @@
...
+    UNNEST(metrics.labeled_string.migration_migration_versions) versions
+  WHERE
+    submission_date
+    BETWEEN '2020-01-01'
+    AND current_date
+),
+clients AS (
+  SELECT
+    fenix_client_id,
+    MAX(submission_date) AS submission_date,
```
jklukas:
```suggestion
    MIN(submission_date) AS submission_date,
```

We need to take the earliest migration date in order to have logical consistency between this and the incremental query.

sql/org_mozilla_firefox_derived/migrated_clients_v1/query.sql
jklukas: If we're going to recreate the entire table every day, I'd argue for having just the `init.sql`, making it a CREATE OR REPLACE statement, and removing this incremental file. This is what we do for `fxa_users_services_first_seen.sql` and the Airflow operator already supports that pattern.

sql/org_mozilla_firefox_derived/migrated_clients_v1/metadata.yaml
jklukas: If we're rewriting it every day, then I'd say it's not incremental.

sql/org_mozilla_firefox/migration/view.sql
fbertsch: This ping should remain consistent for two reasons:
1. It's [defined](https://probeinfo.telemetry.mozilla.org/glean/support-migration/pings) in a shared lib
2. Each of these app ids [shares the same repository url](https://probeinfo.telemetry.mozilla.org/glean/repositories)

sql/org_mozilla_firefox_derived/migrated_clients_v1/query.sql
fbertsch: The caveat is suddenly we'd be reading more data. I'd prefer to keep it using existing results, and updating with the new data it reads. That should keep the query smaller and faster.

sql/org_mozilla_firefox/migration/view.sql
fbertsch: I do think this is a useful view to have for all migrations. I don't see this being a heavily used table aside from me.

sql/org_mozilla_firefox_derived/migrated_clients_v1/metadata.yaml
fbertsch: Hm, but we use the results from the previous day, and increment the entire table based on those.

sql/org_mozilla_firefox/migration/view.sql
jklukas: AFAIK, the reasons you give for consistency of migration ping should also apply to baseline ping. How are the cases different?

sql/org_mozilla_firefox_derived/migrated_clients_v1/query.sql
jklukas: Do we have a sense of how large the migration table could grow to be? If the total size is under a few TB and run time is no more than a few minutes, I'd argue the simplicity of completely recalculating is worth it.

jklukas: But don't get slowed down by this. I'm interested in getting this stably into prod rather than making it perfect.

sql/org_mozilla_firefox/migration/view.sql
fbertsch: They do apply for the baseline ping. I made a quick comparison here: https://gist.github.com/fbertsch/9fb62a9d22b45d269555eab3a9f80729, and they _should_ be the same between those apps.

fbertsch: Okay, I compared their actual schemas in BQ and they also seem to match using https://gist.github.com/fbertsch/2378a39de9b4cc485d5d9151999f9546.

```
python compare_schemas.py --project moz-fx-data-shared-prod org_mozilla_firefox.baseline org_mozilla_firefox_beta.baseline org_mozilla_fenix.baseline org_mozilla_fennec_aurora.baseline org_mozilla_fenix_nightly.baseline

moz-fx-data-shared-prod.org_mozilla_firefox.baseline and moz-fx-data-shared-prod.org_mozilla_firefox_beta.baseline match schemas
moz-fx-data-shared-prod.org_mozilla_firefox.baseline and moz-fx-data-shared-prod.org_mozilla_fenix.baseline match schemas
moz-fx-data-shared-prod.org_mozilla_firefox.baseline and moz-fx-data-shared-prod.org_mozilla_fennec_aurora.baseline match schemas
moz-fx-data-shared-prod.org_mozilla_firefox.baseline and moz-fx-data-shared-prod.org_mozilla_fenix_nightly.baseline match schemas
```

jklukas: In practice, the baseline schemas are not compatible. If you attempt the following today, you get an error about incompatible types:

```
select * from org_mozilla_fenix_stable.baseline_v1
union all
select * from org_mozilla_firefox_stable.baseline_v1
union all
select * from org_mozilla_firefox_beta_stable.baseline_v1
union all
select * from org_mozilla_fennec_aurora_stable.baseline_v1
union all
select * from org_mozilla_fenix_nightly_stable.baseline_v1
```

It does happen to work for `migration` currently. I won't block this PR on this issue, but be aware that it's technical debt that could cause this ETL to fail unexpectedly in the future.

fbertsch: Right, it looks like the structs must not be ordered the same way. This will require close inspection at the pieces of probe-scraper, MSG, and jsonschema-transpiler to figure out where the issue lies for baseline.

fbertsch: I opened a [bug 1624862](https://bugzilla.mozilla.org/show_bug.cgi?id=1624862) since this clearly seems like an issue. Thanks for bringing it up!

fbertsch:
@jklukas went ahead and removed the query update for migrated_clients_v1. If workload becomes an issue we can re-implement it.

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
```diff
@@ -0,0 +1,170 @@
...
+    is_migrated,
+    app_name,
+    channel,
+    manufacturer,
+    country
+)
+SELECT
+  * EXCEPT (established_churned, new_churned),
+  -1 * established_churned AS established_churned,
+  -1 * new_churned AS new_churned,
```
jklukas: Need comments here to explain the `-1`

PR #844, opened 2020-03-23
update rocket telemetry on Mar. 23rd

Conjuror:


sql/telemetry/rocket_android_events_v1/view.sql
```diff
@@ -24,7 +24,7 @@ FROM
 SELECT
     submission_timestamp,
     client_id AS device_id,
-    (created + COALESCE(SAFE_CAST(`moz-fx-data-shared-prod.udf.get_key`(event_map_values, 'session_id') AS INT64), 0)) AS session_id,
+    (created + COALESCE(SAFE_CAST(`moz-fx-data-derived-datasets.udf.get_key`(event_map_values, 'session_id') AS INT64), 0)) AS session_id,
```
jklukas: All UDF references should point to the `shared-prod` now rather than `derived-datasets`. The `derived-datasets` project is being deprecated and will eventually be removed. See https://bugzilla.mozilla.org/show_bug.cgi?id=1612543

It looks like this PR makes no other changes, so likely this can be closed?

Conjuror:
Thanks Jeff. You are right about it. I'll close the issue. Thanks for clarifying.

PR #849, opened 2020-03-24
Add UDF used to define proposed segments

felixlawrence:
I'm proposing a [trio of segments](https://github.com/mozilla/firefox-data-docs/pull/428), including "regular users v1". It is currently possible to query these segments (see the firefox-data-docs PR, or the [mozanalysis PR](https://github.com/mozilla/mozanalysis/pull/76)), but it is not very user friendly.

If we add this UDF then it becomes much easier to compute the quantities in terms of which these segments are defined.

In [PR 825](https://github.com/mozilla/bigquery-etl/pull/825), I will separately propose to add a column to `clients_last_seen` that contains each user's segment.

### DIFF
--- a/udf/num_weeks_gte_n_from_bits.sql
+++ b/udf/num_weeks_gte_n_from_bits.sql
@@ -1,0 +1,13 @@
+/*
+
+Returns the number of weeks with >=n nonzero bits, given an integer representing
+a bit array for the last 4 weeks.
+
+*/
+CREATE OR REPLACE FUNCTION udf.num_weeks_gte_n_from_bits(b BYTES, n INT64) AS (
+  CAST(BIT_COUNT(b & udf.bitmask_range(1, 7)) >= n AS INT64) + CAST(
+    BIT_COUNT(b & udf.bitmask_range(8, 7)) >= n AS INT64
+  ) + CAST(BIT_COUNT(b & udf.bitmask_range(15, 7)) >= n AS INT64) + CAST(
+    BIT_COUNT(b & udf.bitmask_range(22, 7)) >= n AS INT64
+  )
+);


udf/num_weeks_gte_n_from_bits.sql
```diff
@@ -0,0 +1,13 @@
+/*
+
+Returns the number of weeks with >=n nonzero bits, given an integer representing
+a bit array for the last 4 weeks.
+
+*/
+CREATE OR REPLACE FUNCTION udf.num_weeks_gte_n_from_bits(b BYTES, n INT64) AS (
```
jklukas:
```suggestion
CREATE OR REPLACE FUNCTION udf.num_weeks_gte_n_from_bits(b INT64, n INT64) AS (
```

The 28-day bit patterns in clients_last_seen are integers rather than BYTES.

udf/num_weeks_gte_n_from_bits.sql
```diff
@@ -0,0 +1,13 @@
...
+  CAST(BIT_COUNT(b & udf.bitmask_range(1, 7)) >= n AS INT64) + CAST(
+    BIT_COUNT(b & udf.bitmask_range(8, 7)) >= n AS INT64
+  ) + CAST(BIT_COUNT(b & udf.bitmask_range(15, 7)) >= n AS INT64) + CAST(
+    BIT_COUNT(b & udf.bitmask_range(22, 7)) >= n AS INT64
+  )
+);
```
jklukas:
```suggestion
);

-- Tests

SELECT
  assert_equals(1, udf.num_weeks_gte_n_from_bits(1 << 1 + 1 << 9 + 1 << 10, 2))
```

We should add at least one test case, which would have caught the type mismatch above.

felixlawrence:
Holding off on this, pending an investigation of the potential change to the segment definition proposed in #825

felixlawrence:
The current proposed segment definition [has changed](https://github.com/mozilla/bigquery-etl/pull/825#issuecomment-605355290) such that this UDF is no longer required.

PR #850, opened 2020-03-24
ASN aggregates

scholtzan:
This implements the ASN aggregates dataset as described in: https://bugzilla.mozilla.org/show_bug.cgi?id=1615269

Currently the `static.geoip2_isp_blocks_ipv4` is not available in prod, so we will need to wait for it before merging. But I think it is ready for a look.

After that I will schedule it to run in Airflow.


sql/telemetry_derived/asn_aggregates_v1/query.sql
```diff
@@ -0,0 +1,136 @@
...
+    ) AS prefix,
+    autonomous_system_number
+  FROM
+    (
+      SELECT
+        cidr_range(network) AS ip_range,
+        autonomous_system_number,
+        network
+      FROM
+        `static.geoip2_isp_blocks_ipv4`
```
scholtzan: This table doesn't exist yet in prod and might also have a different name once it gets created. But I tested it in a sandbox project of mine.

sql/telemetry_derived/asn_aggregates_v1/query.sql
```diff
@@ -0,0 +1,136 @@
...
+        udf.parse_desktop_telemetry_uri(uri).document_id AS document_id
+      FROM
+        `moz-fx-data-shared-prod.payload_bytes_raw.telemetry`
+    ) AS payload_bytes_raw
+  ON
+    payload_bytes_raw.document_id = main_summary.document_id
+  WHERE
+    submission_date = @submission_date
+    AND DATE(payload_bytes_raw.submission_timestamp) = @submission_date
+),
```
tdsmith: Cool, I see how this works.

fwiw the thing I actually want to count comes from events, not main pings -- I want to count the number of events with category `doh` and object `heuristics` by whether or not the map value `canary` is `disable_doh`.

sql/telemetry_derived/asn_aggregates_v1/query.sql
```diff
@@ -0,0 +1,136 @@
...
+    submission_date,
+    client_id,
+    trr,
+    autonomous_system_number,
+  FROM
+    main_summary_with_ip
+  JOIN
+    asn_ip_address_range AS a
+  ON
+    a.prefix = NET.IP_TRUNC(ip_address, 16)
```
tdsmith: Why do you truncate to the /16?

sql/telemetry_derived/asn_aggregates_v1/query.sql
tdsmith: The `telemetry_derived.events` table has a `document_id` field and should support this similarly.

tdsmith:
Is the Maxmind database guaranteed to not have overlapping network ranges?

sql/telemetry_derived/asn_aggregates_v1/query.sql
scholtzan: I ran into some performance issues with joining network ranges so I was trying to do some pre-filtering. However, I found a blog post with an even better way to do all of this: https://towardsdatascience.com/geolocation-with-bigquery-de-identify-76-million-ip-addresses-in-20-seconds-e9e652480bd2

I updated the query based on what is described there, and now it takes only a few seconds to run.

sql/telemetry_derived/asn_aggregates_v1/query.sql
```diff
@@ -0,0 +1,124 @@
...
+),
+events_with_doh AS (
+  -- Get event data with DoH information.
+  SELECT
+    submission_date,
+    document_id,
+    client_id,
+    event_object,
+    event_category,
+    udf.get_key(event_map_values, 'canary') AS doh,
```
tdsmith: nit: can we call this `canary`?

scholtzan:
> Is the Maxmind database guaranteed to not have overlapping network ranges?

In theory there shouldn't be overlapping ranges (otherwise it would mean different network operators assign the same IP address to multiple clients, no?). But I couldn't find any explicit guarantees.

tdsmith:
The link between AS numbers and networks is ephemeral and there's no authoritative view of the routing table; network operators announce which prefixes they route, and then those announcements can be filtered or aggregated into "supernetworks" by upstream providers, so it's all a bit of a hash. :D

tdsmith:
I think I decided it doesn't matter, though; if a client ends up matching multiple networks that's fine

sql/telemetry_derived/asn_aggregates_v1/query.sql
```diff
@@ -0,0 +1,129 @@
...
+  """;
+
+--
+WITH asn_ip_address_range AS (
+  SELECT
+    NET.SAFE_IP_FROM_STRING(SPLIT(network, "/")[OFFSET(0)]) AS network_ip,
+    CAST(SPLIT(network, "/")[OFFSET(1)] AS INT64) AS mask,
+    autonomous_system_number
+  FROM
+    `moz-fx-data-shared-prod.static.isp_blocks_ipv4_20200407`
```
scholtzan, line 43: We got a temporary table with the ISP data in shared-prod. In the future, this table will be updated weekly and probably renamed to `isp_blocks_ipv4`.

PR #851, opened 2020-03-25
Cleanup GLAM etl queries for Fenix

acmiyaguchi:
This PR does some refactoring, such as renaming variables for consistency and removing all unused code pathways (options meant for telemetry). Most of the code generation has gone into the `bigquery_etl.glam.generate` module, which reduces the number of files by a bit. Finally, I've added the final `fenix_view_client_probe_counts_v1`, which will be used to ingest into GLAM.

This PR should make it easier to set up new products with these scripts in the next few iterations.


script/glam/generate_fenix_sql
```diff
@@ -45,113 +45,10 @@ function write_clients_daily_aggregates {
...
-    latest_versions
-
-    write_scalar_bucket_counts
-    write_scalar_probe_counts
-    write_scalar_percentiles
-
-    write_histogram_bucket_counts
-    write_histogram_probe_counts
-    write_histogram_percentiles
+    python -m bigquery_etl.glam.generate
```
emtwo, line 53: `python3`?

bigquery_etl/glam/generate.py
```diff
@@ -0,0 +1,161 @@
...
+            "scalar_percentiles_v1",
+            **models.scalar_percentiles(
+                source_table="glam_etl.fenix_clients_scalar_aggregates_v1"
+            ),
+        ),
+        view("view_clients_daily_scalar_aggregates_v1"),
+        view("view_clients_daily_histogram_aggregates_v1"),
+        table("histogram_percentiles_v1"),
+        view("view_client_probe_counts_v1"),
+    ]
```
emtwo, line 157: out of curiousity, why are the square brackets needed?

acmiyaguchi: It's actually not necessary, it's just collecting the results in a list before throwing it away. I was thinking of using this for building up a mermaid diagram in the future.

PR #854, opened 2020-03-25
Add OR REPLACE to migrated_clients_v1

relud:


jklukas:
Thanks for fixing. This will also allow the Airflow task to actually succeed tomorrow! 🍾

PR #859, opened 2020-03-27
Usage anomalies

scholtzan:
Query for publishing the deviations dataset to BigQuery and JSON
cc @jmccrosky

### DIFF
--- a/sql/analysis/deviations_v1/metadata.yaml
+++ b/sql/analysis/deviations_v1/metadata.yaml
@@ -1,10 +1,9 @@
-friendly_name: Daily Active Users Deviations
+friendly_name: Deviations
 description: >
-  Deviation of DAU (Daily Active Users) from forecast.
+  Deviation of different metrics from forecast.
 owners:
   - jmccrosky@mozilla.com
 labels:
-  incremental: true
   schedule: daily
   public_json: true
   public_bigquery: true
--- a/sql/analysis/deviations_v1/metadata.yaml
+++ b/sql/analysis/deviations_v1/metadata.yaml
@@ -1,0 +1,10 @@
+friendly_name: Daily Active Users Deviations
+description: >
+  Deviation of DAU (Daily Active Users) from forecast.
+owners:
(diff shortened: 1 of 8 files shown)


sql/analysis/deviations_v1/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: Deviations
+description: >
+  Deviation of different metrics from forecast.
+owners:
+  - jmccrosky@mozilla.com
+labels:
+  incremental: true
+  schedule: daily
+  public_json: true
+  public_bigquery: true
```
scholtzan: This is already works

sql/analysis/deviations_v1/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: Deviations
+description: >
+  Deviation of different metrics from forecast.
+owners:
+  - jmccrosky@mozilla.com
+labels:
+  incremental: true
+  schedule: daily
+  public_json: true
```
scholtzan: This will work once https://github.com/mozilla/bigquery-etl/pull/839 is merged

sql/analysis/deviations_v1/query.sql
```diff
@@ -0,0 +1,10 @@
+SELECT
+  date,
+  ci_deviation,
+  deviation,
+  geography,
+  metric
+FROM
+  `moz-fx-data-derived-datasets.analysis.deviations`
```
scholtzan: @jklukas I saw this bug about [decomissioning derived datasets](https://bugzilla.mozilla.org/show_bug.cgi?id=1612543). Does that mean that this data should rather be written somewhere else?

jklukas: Yes, we should write it to `moz-fx-data-shared-prod.analysis.deviations`

scholtzan:
I moved it to `telemetry_derived`. Does that seems right?

PR #863, opened 2020-03-31
Bug 1607229 - Add authorized view to buildhub2 table

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1607229)

This creates a view for the official build metadata for Firefox, which is managed in a separate project. The view needs to exist before it can be authorized.


sql/telemetry/buildhub2/view.sql
```diff
@@ -0,0 +1,10 @@
+-- This is an authorized view that allows us to read specific data from
```
jklukas: We should also add this to the skip list in `script/publish_views` do prevent it from being published in the schema pipeline.

jklukas: It will need to be in the dry run skiplist too, as I expect the dry run service's permissions will never allow this DDL statement to run successfully.

PR #864, opened 2020-03-31
Bug 1601281 addon_names query

jklukas:
As originally proposed in https://bugzilla.mozilla.org/show_bug.cgi?id=1601281#c2


sql/telemetry_derived/addon_names_v1/query.sql
```diff
@@ -0,0 +1,31 @@
+WITH grouped AS (
+  SELECT
+    addons.key AS addon_id,
+    addons.value.name AS addon_name,
+    COUNT(*) AS occurrences
```
relud: This should probably be:
```suggestion
    COUNT(DISTINCT client_id) AS occurrences
```

sql/telemetry/addons_daily/view.sql
```diff
@@ -4,4 +4,4 @@ AS
 SELECT
   *
 FROM
-  `moz-fx-data-shared-prod.telemetry_derived.addons_daily_v1`
\ No newline at end of file
+  `moz-fx-data-shared-prod.telemetry_derived.addons_daily_v1`
```
jklukas: Unrelated change that was picked up by the formatter.

PR #866, opened 2020-04-01
Add review links to public datasets and make deviations non-incremental

scholtzan:
The `deviations_v1` and `ssl_ratios` queries should be marked as non-incremental. That way, all data will get exported to one file instead of 1 file per day. 

As a next future step, metadata about existing public datasets will be published to GCS. This metadata should also contain a link to the data review.

### DIFF
--- a/sql/telemetry_derived/deviations_v1/metadata.yaml
+++ b/sql/telemetry_derived/deviations_v1/metadata.yaml
@@ -8,4 +8,4 @@ labels:
   schedule: daily
   public_json: true
   public_bigquery: true
-  review_link: https://bugzilla.mozilla.org/show_bug.cgi?id=1624528
+  review_bug: 1624528
--- a/sql/telemetry_derived/deviations_v1/metadata.yaml
+++ b/sql/telemetry_derived/deviations_v1/metadata.yaml
@@ -4,7 +4,8 @@ description: >
 owners:
   - jmccrosky@mozilla.com
 labels:
-  incremental: true
+  incremental: false
   schedule: daily
   public_json: true
   public_bigquery: true
+  review_link: https://bugzilla.mozilla.org/show_bug.cgi?id=1624528
--- a/sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
+++ b/sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
@@ -6,8 +6,9 @@ owners:
   - chutten@mozilla.com
 labels:
   application: firefox
+  # Note: the query is incremental but exported data is treated as non-incremental
   incremental: false
   schedule: daily
   public_json: true
   public_bigquery: true
-  review_link: https://bugzilla.mozilla.org/show_bug.cgi?id=1414839
+  review_bug: 1414839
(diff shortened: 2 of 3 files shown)


sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
```diff
@@ -6,7 +6,8 @@ owners:
   - chutten@mozilla.com
 labels:
   application: firefox
-  incremental: true
+  incremental: false
```
scholtzan, line 9: `incremental` is a bit confusing here since the query itself is incremental (it produces new data every day and appends it to the table), but it should be treated as non-incremental. That probably needs some further discussion when we get to implementing low-friction scheduled queries.

sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
```diff
@@ -6,7 +6,8 @@ owners:
...
   schedule: daily
   public_json: true
   public_bigquery: true
+  review_link: https://bugzilla.mozilla.org/show_bug.cgi?id=1414839
```
jklukas, line 13: URLs contain characters that aren't allowed for label keys or values. We could have a simple bug number here instead.

sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
jklukas: We should definitely note this gotcha as a comment here.

sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
jklukas: Perhaps we call it `review_bug`

scholtzan: I guess that is okay for now. Though is it possible in the future that data reviews will be posted on somewhere other than Bugzilla?

jklukas: Indeed that may happen in the future, but it seems like it would be pretty easy to refactor the labels at that point. I'm open to suggestions on a more adaptable format, but don't want to block your work here on it.

scholtzan: I think its a good option for now. The code itself should be easy to refactor if/when the use case arises.

sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
fbertsch: This should get added into the README at https://github.com/mozilla/bigquery-etl#query-metadata.

sql/telemetry_derived/deviations_v1/metadata.yaml
```diff
@@ -4,7 +4,8 @@ description: >
 owners:
   - jmccrosky@mozilla.com
 labels:
-  incremental: true
+  incremental: false
   schedule: daily
   public_json: true
   public_bigquery: true
+  review_link: https://bugzilla.mozilla.org/show_bug.cgi?id=1624528
```
fbertsch, line 11:
```suggestion
  review_bug: 1624528
```

sql/telemetry_derived/ssl_ratios_v1/metadata.yaml
fbertsch: Hmm. Let's open an issue for how we want to solve this. I believe that we want a separate label for how the data export is handled; e.g.
```
incremental: true
incremental_export: false
```

PR #868, opened 2020-04-01
Add build_id column to telemetry_derived.events_live

mnoorenberghe:
I used the name `build_id` to keep it consistent with `telemetry.events_v1` so it's easy to switch queries between the two views.


mnoorenberghe:
CC @sunahsuh

mnoorenberghe:
Thanks! Note that I don't have merge permission so if someone can merge it for me that would be great :)

Issue #865, opened 2020-04-01
Large number of files when exporting partitioned tables to JSON


scholtzan:
For exporting data as JSON to GCS, `extract_table` is currently used. When exporting a partitioned table, data gets exported into multiple separate files based on the partition. E.g. in the case of `ssl_ratios`, a new file is created for each date. This can result in a very large number of files for non-incremental queries.

I think it would probably be best, when converting the files from ndjson to json to merge them. Still, 1GB can be kept as a maximum in terms of file size.

PR #869, opened 2020-04-01
Fix: large number of files when exporting tables to JSON 

scholtzan:
This fixes https://github.com/mozilla/bigquery-etl/issues/865

For exporting data as JSON to GCS, `extract_table` is currently used. When exporting a table, data gets exported into a lot of separate files. E.g. in the case of `ssl_ratios`, a new file is created for each date. This can result in a very large number of files.

First I suspected that this is only the case for partitioned tables, but I ran some tests and it is also the case for tables that are not partitioned

The fix here: when converting ndjson to json the output will be written to one file until the output file size limit of 1 GB is reached. If the limit is exceeded a new output file is created.

### DIFF
--- a/bigquery_etl/publish_json.py
+++ b/bigquery_etl/publish_json.py
@@ -12,7 +12,7 @@ from bigquery_etl.parse_metadata import Metadata
 METADATA_FILE = "metadata.yaml"
 SUBMISSION_DATE_RE = re.compile(r"^submission_date:DATE:(\d\d\d\d-\d\d-\d\d)$")
 QUERY_FILE_RE = re.compile(r"^.*/([a-zA-Z0-9_]+)/([a-zA-Z0-9_]+)_(v[0-9]+)/query\.sql$")
-
+MAX_JSON_SIZE = 1 * 1024 * 1024 * 1024  # 1 GB as max. size of exported JSON files
 
 logging.basicConfig(
     level=logging.DEBUG, format="%(asctime)s: %(levelname)s: %(message)s"
@@ -64,8 +64,8 @@ class JsonPublisher:
 
     def __del__(self):
         """Delete temporary artifacts."""
-        # if self.temp_table:
-        #     self.client.delete_table(self.temp_table)
+        if self.temp_table:
+            self.client.delete_table(self.temp_table)
 
         self._clear_stage_directory()
 
@@ -90,12 +90,11 @@ class JsonPublisher:
             # if it is an incremental query, then the query result needs to be
             # written to a temporary table to get exported as JSON
             self._write_results_to_temp_table()
+            self._publish_table_as_json(self.temp_table)
         else:
-            # for non-incremental queries, the destination table is needs to be
-            # written to a temporary non-partioned table which gets exported
-            self._export_table_to_temp_table()
-
-        self._publish_table_as_json(self.temp_table)
+            # for non-incremental queries, the entire destination table is exported
+            result_table = f"{self.dataset}.{self.table}_{self.version}"
+            self._publish_table_as_json(result_table)
 
     def _publish_table_as_json(self, result_table):
         """Export the `result_table` data as JSON to Cloud Storage."""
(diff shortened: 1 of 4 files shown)


bigquery_etl/publish_json.py
```diff
@@ -12,7 +12,7 @@
 METADATA_FILE = "metadata.yaml"
 SUBMISSION_DATE_RE = re.compile(r"^submission_date:DATE:(\d\d\d\d-\d\d-\d\d)$")
 QUERY_FILE_RE = re.compile(r"^.*/([a-zA-Z0-9_]+)/([a-zA-Z0-9_]+)_(v[0-9]+)/query\.sql$")
-
+MAX_JSON_SIZE = 1 * 1024 * 1024 * 1024  # 1 GB as max. size of exported JSON files
```
scholtzan, line 15: For now the limit is 1GB, but we can also increase/decrease it

bigquery_etl/publish_json.py
```diff
@@ -133,26 +133,57 @@ def _gcp_convert_ndjson_to_json(self, gcs_path):
...
-
-                    fout.write("]")
+                for line in fin:
+                    # check if JSON output file reached limit
+                    if output_file is None or output_size >= MAX_JSON_SIZE:
+                        if output_file is not None:
+                            output_file.write("]")
+                            output_file.close()
+
+                        if output_file_counter >= 1000000000000:
```
fbertsch, line 157: I doubt we'll run into this, but this is off-by-one; we'll have already written the file with this file counter.

bigquery_etl/publish_json.py
```diff
@@ -133,26 +133,57 @@ def _gcp_convert_ndjson_to_json(self, gcs_path):
...
+                        tmp_blob_name = "/".join(blob.name.split("/")[:-1])
+                        tmp_blob_name += (
+                            "/" + str(output_file_counter).zfill(12) + ".json.tmp.gz"
+                        )
+                        tmp_blob_path = f"gs://{self.target_bucket}/{tmp_blob_name}"
+
+                        logging.info(f"""Write {blob_path} to {tmp_blob_path}""")
+
+                        output_file = smart_open.open(tmp_blob_path, "w")
+                        output_file.write("[\n")
```
fbertsch, line 172: We could alleviate some size constraints by eliding the newlines. JSON viewers in browsers should be an acceptable alternative.

bigquery_etl/publish_json.py
```diff
@@ -46,7 +46,7 @@ def __init__(
 
         self.metadata = Metadata.of_sql_file(self.query_file)
 
-        if self.parameter:
+        if self.metadata.is_incremental() and self.parameter:
```
fbertsch, line 49: This line needs a comment now.

bigquery_etl/publish_json.py
jklukas: Can we make this a named constant? It's not obvious to me what this literal means.

jklukas: And apparently we can use underscores in numeric literals since python 3.6: https://www.python.org/dev/peps/pep-0515/

So this can be `1_000_000_000_000`.

scholtzan:
I ran into some issues with the Covid data which was stored in a table that wasn't partitioned, but still it got split up into a separate file for each date.

tests/publish_public_data_json/test_publish_public_data_json_script.py
```diff
@@ -46,7 +46,10 @@ def setup(self):
         try:
             self.client.get_table(self.non_incremental_table)
         except NotFound:
-            job_config = bigquery.QueryJobConfig(destination=self.non_incremental_table)
+            date_partition = bigquery.table.TimePartitioning(field="d")
```
scholtzan, line 49: Having a partition here, results in exporting a ndjson file for each date.

bigquery_etl/publish_json.py
fbertsch: In retrospect this should be a smaller literal. If we hit 10k files we're probably having a problem and are ready to bail.

bigquery_etl/publish_json.py
```diff
@@ -12,7 +12,9 @@
 METADATA_FILE = "metadata.yaml"
 SUBMISSION_DATE_RE = re.compile(r"^submission_date:DATE:(\d\d\d\d-\d\d-\d\d)$")
 QUERY_FILE_RE = re.compile(r"^.*/([a-zA-Z0-9_]+)/([a-zA-Z0-9_]+)_(v[0-9]+)/query\.sql$")
-
+MAX_JSON_SIZE = 1 * 1024 * 1024 * 1024  # 1 GB as max. size of exported JSON files
+# maximum number of JSON output files, output file names are only up to 12 characters
+MAX_FILE_COUNT = 999_999_999_999
```
fbertsch, line 17: Thinking about it, should we make this smaller? If we hit 10k files something wrong has probably happened and we should bail.

bigquery_etl/publish_json.py
```diff
@@ -133,26 +137,57 @@ def _gcp_convert_ndjson_to_json(self, gcs_path):
...
+                        )
+                        tmp_blob_path = f"gs://{self.target_bucket}/{tmp_blob_name}"
+
+                        logging.info(f"""Write {blob_path} to {tmp_blob_path}""")
+
+                        output_file = smart_open.open(tmp_blob_path, "w")
+                        output_file.write("[")
+                        first_line = True
+                        output_file_counter += 1
+                        output_size = 3
```
fbertsch, line 179:
```suggestion
                        output_size = 1
```

bigquery_etl/publish_json.py
```diff
@@ -133,26 +137,57 @@ def _gcp_convert_ndjson_to_json(self, gcs_path):
...
+
+                        if output_file_counter >= MAX_FILE_COUNT:
+                            logging.error(
+                                "Maximum number of JSON output files reached."
+                            )
+                            sys.exit(1)
+
+                        tmp_blob_name = "/".join(blob.name.split("/")[:-1])
+                        tmp_blob_name += (
+                            "/" + str(output_file_counter).zfill(12) + ".json.tmp.gz"
```
fbertsch, line 169: `12` should also be a constant, in line with `max_file_count`

bigquery_etl/publish_json.py
```diff
@@ -133,26 +137,57 @@ def _gcp_convert_ndjson_to_json(self, gcs_path):
...
+                        output_file.write("[")
+                        first_line = True
+                        output_file_counter += 1
+                        output_size = 3
+
+                    # skip the first line, it has no preceding json object
+                    if not first_line:
+                        output_file.write(",")
+
+                    output_file.write(line.replace("\n", ""))
```
fbertsch, line 185: Do these files have trailing newlines? If not, we could remove the `first_line` handling and replace newlines with commas.

scholtzan: Yeah, they have trailing newlines.

PR #870, opened 2020-04-02
Fenix exports for GLAM

acmiyaguchi:
This adds the last few tables necessary for extracting probe and user counts with deduplication.

```bash
$ gsutil ls gs://glam-fenix-dev/                              
gs://glam-fenix-dev/glam-extract-fenix-000000000000.csv
gs://glam-fenix-dev/glam-extract-fenix-counts.csv
```


script/glam/run_glam_sql
```diff
@@ -63,6 +63,7 @@ function run_query {
         --parameter="submission_date:DATE:$SUBMISSION_DATE" \
         --parameter="min_sample_id:INT64:$min_sample_id" \
         --parameter="max_sample_id:INT64:$max_sample_id" \
+        "$(if $time_partition; then echo --time_partitioning_type="DAY"; fi)" \
```
acmiyaguchi, line 66: This is necessary when running with `RESET=true`.

script/glam/export_csv
```diff
@@ -0,0 +1,22 @@
...
+# To be run in glam via:
+#    docker-compose run server ./manage.py import_glean_aggs
+
+set -e
+
+src_project=${SRC_PROJECT:-glam-fenix-dev}
+# TODO: glam-dev-bespoke-nonprod-dataops-mozgcp-net
+dst_project=${DST_PROJECT:-$src_project}
+dataset=${DATASET:-glam_etl_dev}
+product=${PRODUCT:-fenix}   # TODO: set default to org_mozilla_fenix
```
acmiyaguchi, line 11: I'm going to leave this for another PR, where the prefix will change from `fenix` to `org_mozilla_fenix`.

script/glam/export_csv
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+# To be run in glam via:
+#    docker-compose run server ./manage.py import_glean_aggs
+
+set -e
+
+src_project=${SRC_PROJECT:-glam-fenix-dev}
+# TODO: glam-dev-bespoke-nonprod-dataops-mozgcp-net
```
acmiyaguchi, line 8: I don't have permissions to write into this bucket, so I'll modify the glam import script accordingly.

bigquery_etl/glam/templates/client_probe_counts_extract_v1.sql
```diff
@@ -0,0 +1,46 @@
+-- TODO: Remove deduping when dupes are fixed.
+WITH deduped AS (
```
robhudson, line 2: The de-duping was something we had problems with for desktop aggregations and which should be fixed now. Does this also have the issue with Glean on Fenix? If not, I'd remove this and see how the imports work. The imports use upserts which rely on there being a unique row per the dimensions.

script/glam/export_csv
```diff
@@ -0,0 +1,22 @@
...
+bucket="gs://${dst_project}"
+gsutil ls "$bucket" > /dev/null
+
+bq extract --destination_format CSV \
+    "${src_project}:${dataset}.${product}_client_probe_counts_extract_v1" \
+    "$bucket/glam-extract-${product}-*.csv"
+
+bq extract --destination_format CSV \
+    "${src_project}:${dataset}.${product}_user_counts_extract_v1" \
+    "$bucket/glam-extract-${product}-counts.csv"
```
robhudson, line 22: This is nice and compact. Is this the way we should approach this for desktop also, instead of in airflow?

robhudson, line 22: Does this default to printing headers in the CSVs?

bigquery_etl/glam/templates/client_probe_counts_extract_v1.sql
acmiyaguchi: I haven't tried it without this. Do you know what the original reason for duplicates was?

script/glam/export_csv
acmiyaguchi: It's certainly easier than in airflow for a development workflow. I'm not 100% confident on whether this should take precedent over airflow native constructs though.

acmiyaguchi: Yes this does, I've added the appropriate flags to stop printing the header.

emtwo:
Oh, somehow this PR slipped my radar. I will take a look today or early tomorrow. But it looks fine at a quick glance.

bigquery_etl/glam/models.py
```diff
@@ -177,3 +177,28 @@ def scalar_percentiles(**kwargs):
...
+    attribute_combinations = []
+    for subset_size in reversed(range(max_combinations + 1)):
+        for grouping in combinations(attributes, subset_size):
+            # channel and app_version are required in the GLAM frontend
+            if "channel" not in grouping or "app_version" not in grouping:
+                continue
+            select_expr = []
+            for attribute in attributes:
+                select_expr.append((attribute, attribute in grouping))
+            attribute_combinations.append(select_expr)
```
emtwo, line 198: nit: why not factor out this snippet of code for combinations into a separate function? Seems redundant in 2 other locations. Though while we're here, I can't remember why we don't have the check for `channel` and `app_version` to be in `grouping` for `scalar_percentiles?`

bigquery_etl/glam/templates/client_probe_counts_extract_v1.sql
emtwo: Two main reasons:
1. Some fields, e.g. os, had values of `NULL`, this interfered with our use of os = `NULL` to include all OSes and resulted in some rows appearing twice.
2. Buckets in the data for a given metric had different values so when grouping included buckets, some rows would appear more than once.

Both of these have since been resolved for desktop (except for the user counts query which does not exclude null OSes yet).

I suspect Glean data is more trustworthy and should not unexpectedly have incorrect/null values for fields but it's worth double checking before removing the deduping here.

bigquery_etl/glam/templates/client_probe_counts_extract_v1.sql
```diff
@@ -0,0 +1,47 @@
...
+                agg_type
+            ORDER BY
+                total_users DESC
+        ) AS rank
+    FROM
+        `{{ dataset }}.{{ prefix }}_view_client_probe_counts_v1`
+    WHERE
+        channel IS NOT NULL
+        AND app_version IS NOT NULL
+        AND total_users >= 100
```
emtwo, line 25: Even if the deduping is removed, these filters are still needed.

bigquery_etl/glam/templates/client_probe_counts_extract_v1.sql
```diff
@@ -0,0 +1,47 @@
...
+)
+
+SELECT
+    channel,
```
emtwo, line 29: @robhudson I know you had numeric representations for desktop's channel and wondering if you had thoughts on whether they are needed here?

robhudson: Good call out, thanks.

I think for Glean I will leave out the numeric representations. The data is smaller and it also seems like it is in fluctuation, at least with the Fenix and Fennec transition, and having that as constants seems like not such a great idea. So this is good as-is.

bigquery_etl/glam/models.py
acmiyaguchi: I'll factor this out -- these combinations should be the same for all of the queries that compute the data cube? I think `scalar_percentiles` was probably an oversight.

bigquery_etl/glam/templates/client_probe_counts_extract_v1.sql
acmiyaguchi: I removed the deduping and verified that the resulting aggregates violate the uniqueness constraint in the GLAM backend. I'll have to dig into this a bit more in the future.

acmiyaguchi:
I've update this PR with a few more changes. First, the extract tables are prefixed instead of suffixed (e.g. `fenix_extract_probe_counts_v1`). One of the tables was converted into a table (`fenix_view_user_counts_v1`). Finally, the views were added to `publish_views` correctly, so it won't break on deploy.

PR #872, opened 2020-04-02
Incline fix

fbertsch:
- Limit manufacturers
- add can/ cant migrate
- make is_migrated a bool
- fix fenix normalized channel
- Add previous year's established returning retention delta
- add cumulative migration count
- add previous values (from last week) for topline values


sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
```diff
@@ -1,37 +1,29 @@
...
-    ),
-    manufacturer,
-    'Other'
-  )
+  IF(LOWER(manufacturer) IN ('samsung', 'huawei'), LOWER(manufacturer), 'Other')
 );
 
 CREATE TEMP FUNCTION bucket_country(country STRING) AS (
   IF(
     country IN ('US', 'CA', 'DE', 'IN', 'FR', 'CN', 'IR', 'BR', 'IE', 'GB', 'ID'),
```
jklukas, line 7: The established definition of tier 1 used in 2019 KPIs is (US, CA, DE, FR, GB). IT looks like we're including additional countries as tier 1 here.

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
```diff
@@ -1,37 +1,29 @@
...
+    [country, 'tier-1'],
+    ['non-tier-1']
   )
 );
 
 WITH fennec_client_info AS (
   SELECT
     clients_last_seen.submission_date AS date,
-    IF(migrated_clients.fenix_client_id IS NOT NULL, 'Yes', 'No') AS is_migrated,
+    migrated_clients.fenix_client_id IS NOT NULL is_migrated,
```
jklukas, line 16:
```suggestion
    migrated_clients.fenix_client_id IS NOT NULL AS is_migrated,
```

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
```diff
@@ -1,37 +1,29 @@
...
+    migrated_clients.submission_date = clients_last_seen.submission_date AS migrated_today,
     app_name,
     clients_last_seen.normalized_channel AS channel,
     SPLIT(device, '-')[OFFSET(0)] AS manufacturer,
-    country,
+    clients_last_seen.country,
+    COALESCE(
+      CAST(REGEXP_EXTRACT(metadata_app_version, r"^[0-9]+") AS INT64) >= 68
```
jklukas, line 23:
```suggestion
      SAFE_CAST(REGEXP_EXTRACT(metadata_app_version, r"^[0-9]+") AS INT64) >= 68
```

fbertsch:
For reference, there were a few specific problems, highlighted especially by the Fenix migrated clients retention plot (https://moment.data.firefox.com/, "Fennec to Fenix Migration 1-Week Retention Plot").

Active Current and Active Previous look very strange for that population: https://sql.telemetry.mozilla.org/queries/69767/source#175860

The proposed changes seem to fix that: https://sql.telemetry.mozilla.org/queries/69766/source#175857

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
```diff
@@ -41,23 +33,43 @@ WITH fennec_client_info AS (
...
   WHERE
-    clients_last_seen.submission_date = @submission_date
-    AND (
-      migrated_clients.submission_date = @submission_date
-      OR migrated_clients.submission_date IS NULL)
+    clients_last_seen.submission_date IN (
+      @submission_date,
+      DATE_SUB(@submission_date, INTERVAL 1 YEAR),
+      DATE_SUB(DATE_SUB(@submission_date, INTERVAL 1 YEAR), INTERVAL 1 WEEK)
+    )
```
jklukas, line 46: This condition definitely could use some comments. Do we really mean to include (1 year + 1 week)? Or is this supposed to be today, 1 week ago, and 1 year ago?

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
```diff
@@ -41,23 +33,43 @@ WITH fennec_client_info AS (
   LEFT JOIN
     `moz-fx-data-shared-prod.org_mozilla_firefox.migrated_clients` migrated_clients
```
jklukas, line 34:
```suggestion
    `moz-fx-data-shared-prod.org_mozilla_firefox.migrated_clients` AS migrated_clients
```

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
```diff
@@ -153,23 +166,111 @@ counts AS (
...
+    with_retention a
+  INNER JOIN
+    with_retention b
+  ON
+    a.date = DATE_ADD(b.date, INTERVAL 7 DAY)
+    AND a.is_migrated = b.is_migrated
+    AND a.app_name = b.app_name
+    AND a.channel = b.channel
+    AND a.manufacturer = b.manufacturer
+    AND a.country = b.country
```
jklukas, line 226: country could very well be different for the same user on different observations. We may not want to include that in the join condition here.

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
```diff
@@ -153,23 +166,111 @@ counts AS (
...
+    a.channel,
+    a.manufacturer,
+    a.country,
+    a.established_returning_retention_rate - b.established_returning_retention_rate AS established_returning_retention_delta
+  FROM
+    with_retention a
+  INNER JOIN
+    with_retention b
+  ON
+    a.date = DATE_ADD(b.date, INTERVAL 7 DAY)
```
jklukas, line 221: I think we use DATE_SUB everwhere else, so I'd prefer to see this also expressed that way. Also, we should use `1 WEEK` over `7 DAY` consistently.

sql/org_mozilla_firefox_derived/migrated_clients_v1/init.sql
```diff
@@ -3,40 +3,37 @@ CREATE OR REPLACE TABLE
...
   SELECT
     fenix_client_id,
     MAX(submission_date) AS submission_date,
-    `moz-fx-data-shared-prod.udf.mode_last`(
-      ARRAY_AGG(normalized_channel IGNORE NULLS)
-    ) AS normalized_channel,
-    `moz-fx-data-shared-prod.udf.mode_last`(
-      ARRAY_AGG(fennec_client_id IGNORE NULLS)
-    ) AS fennec_client_id,
+    `moz-fx-data-shared-prod.udf.mode_last`( ARRAY_AGG(normalized_channel IGNORE NULLS) ) AS normalized_channel,
```
jklukas, line 28: It would be preferred to not use the project prefix here. That's only needed in view definitions.

sql/org_mozilla_firefox_derived/incline_executive_v1/init.sql
```diff
@@ -0,0 +1,48 @@
...
+    CAST(NULL AS FLOAT64) AS perc_of_active_established_returning,
+    CAST(NULL AS FLOAT64) AS perc_of_active_new_returning,
+    CAST(NULL AS FLOAT64) AS quick_ratio,
+    CAST(NULL AS FLOAT64) AS retention_rate_previous,
+    CAST(NULL AS FLOAT64) AS quick_ratio_previous,
+    CAST(NULL AS INT64) AS can_migrate_previous,
+    CAST(NULL AS FLOAT64) AS established_returning_retention_delta,
+    CAST(NULL AS FLOAT64) AS established_returning_retention_delta_previous,
+    CAST(NULL AS INT64) AS cumulative_migration_count
+)
```
jklukas, line 42: Any particular reason for having a WITH statement here rather than flattening this to a simple SELECT?

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
```diff
@@ -153,23 +166,111 @@ counts AS (
...
+  WHERE
+    date = @submission_date
+),
+last_week AS (
+  SELECT
+    * EXCEPT (date)
+  FROM
+    `moz-fx-data-shared-prod.org_mozilla_firefox_derived.incline_executive_v1`
+  WHERE
+    date = DATE_SUB(@submission_date, INTERVAL 1 WEEK)
```
jklukas, line 205: Are we able to use this same technique for the `last_year` piece? If we can simplify this query by backfilling this table for a year, that seems worth doing.

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
jklukas: Can we completely avoid this logic if we backfill this table to 2019-01-01?

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
fbertsch: Hmm, this came from the incline team meeting. Let's check in with Emily.

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
fbertsch: We can, but I would much prefer not to. We are still iterating on this and a year+ backfill would slow us down prohibitively.

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
fbertsch: `with_retention` is not broken down by client, but by app, is_migrated, channel, manufacturer, and country. We have to include it for the final join below.

sql/org_mozilla_firefox_derived/incline_executive_v1/init.sql
fbertsch: None!

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
jklukas: This was marked as resolved, but I don't see any change. Was this intentional?

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
jklukas: This was marked as resolved, but I don't see any change. Was this intentional?

Issue #867, opened 2020-04-01
JSON export incremental configuration 


scholtzan:
As a follow-up to https://github.com/mozilla/bigquery-etl/pull/866#discussion_r401814091

There are queries, such as `ssl_ratios` which are `incremental` (it produces new data every day and appends it to the table), but it should be treated as non-incremental when exporting the data as JSON to GCS.

Having separate labels is probably the right option here: 

```
incremental: true
incremental_export: false
```

PR #873, opened 2020-04-02
Add incremental_export option

scholtzan:
This addresses https://github.com/mozilla/bigquery-etl/issues/867

I think having a `incremental_export` option is the best way to go here.

### DIFF
--- a/bigquery_etl/parse_metadata.py
+++ b/bigquery_etl/parse_metadata.py
@@ -94,3 +94,7 @@ class Metadata:
     def is_incremental(self):
         """Return true if the incremental flag is set."""
         return "incremental" in self.labels
+
+    def is_incremental_export(self):
+        """Return true if the incremental_export flag is set."""
+        return "incremental_export" in self.labels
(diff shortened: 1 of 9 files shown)


tests/publish_public_data_json/test_sql/test/incremental_query_non_incremental_export_v1/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: Test table for a non-incremental query
+description: >
+  Test table for a non-incremental query and incremental export
+owners:
+  - ascholtz@mozilla.com
+labels:
+  schedule: daily
+  public_json: true
+  incremental: true
+  incremental_export: false
```
scholtzan, line 10: This is to test the behavior that is also expected from `ssl_ratios` and `deviations`: incremental query but non-incremental export

bigquery_etl/parse_metadata.py
```diff
@@ -94,3 +94,7 @@ def is_public_json(self):
     def is_incremental(self):
         """Return true if the incremental flag is set."""
         return "incremental" in self.labels
+
+    def is_incremental_export(self):
+        """Return true if the incremental_export flag is set."""
+        return "incremental_export" in self.labels
```
fbertsch, line 100: I am not going to lie, the fact that `true` / `false` is reported as a tag with `""` made this a very confusing read-through. We should consider moving those labels to tags themselves, either present or not, so the metadata.yaml is exactly what is being used here, and in BQ.

scholtzan: I see. I opened https://github.com/mozilla/bigquery-etl/issues/878 for now and will address that later.

PR #881, opened 2020-04-03
Desktop Glam Sampling

emtwo:
Note: tests were not updated and are expected to fail.

@acmiyaguchi I wanted to get your opinion on sampling in this way before I actually run it for multiple sample sizes.

It looks like for a given `sample_id` the distribution of users in each channel+os combination is roughly the same so looking at 1 sample id for release + windows is roughly like looking at a 1% sample of it.


sql/telemetry_derived/clients_daily_histogram_aggregates_v1.sql.py
```diff
@@ -111,6 +111,26 @@ def generate_sql(opts, additional_queries, windowed_clause, select_clause, json_
...
                 AND client_id IS NOT NULL),
 
+        sampled_data AS (
+          SELECT
+            *
+          FROM
+            filtered
+          WHERE
+            channel IN ("nightly", "beta")
+            OR (channel = "release" AND os != "Windows")
```
acmiyaguchi, line 121: You can probably avoid the `union all` section by adding another OR statement here.

sql/telemetry_derived/clients_daily_histogram_aggregates_v1/query.sql
```diff
@@ -93,9 +93,29 @@ WITH filtered AS (
...
+    OR (channel = "release" AND os != "Windows")
+  UNION ALL
+  SELECT
+    *
+  FROM
+    filtered
+  WHERE
+    channel = 'release'
+    AND os = 'Windows'
+    AND sample_id >= @min_sample_id
```
acmiyaguchi, line 112: The airflow jobs will need to be modified to provide default for these new variables.

emtwo: yep. will write that up when we make a decision on size.

emtwo:
> I'm not sure this has been considered, but `clients_*_aggregates` tables will need to be filtered on these attributes since there are clients that will need to be purged when sampling actually takes effect.

Yes, I'm planning on running a query 1 time to do that purge so it doesn't need to be included in these changes but the sampling would pretty much look the same as what's here.

> When you do the actual measurements on various fractions of sample id, will you be running in a separate project?

Yes, I've been using `glam-desktop-dev` to do testing.

emtwo:
Relevant Airflow PR: https://github.com/mozilla/telemetry-airflow/pull/942

tests/telemetry_derived/clients_histogram_probe_counts_v1/test_aggregation/expect.ndjson
```diff
@@ -0,0 +1,6 @@
...
+{"agg_type":"histogram","aggregates":[{"key":"0","value":0.2},{"key":"1","value":0},{"key":"2","value":0.1},{"key":"3","value":0},{"key":"5","value":0.2},{"key":"9","value":0.3},{"key":"16","value":0},{"key":"29","value":0},{"key":"54","value":0.2},{"key":"100","value":0}],"app_build_id":"second","app_version":75,"channel":"nightly","client_agg_type":"summed_histogram","key":"","metric":"test","metric_type":"histogram-exponential","process":"parent","total_users":1}
+{"agg_type":"histogram","aggregates":[{"key":"0","value":0.4},{"key":"1","value":0},{"key":"2","value":0.2},{"key":"3","value":0},{"key":"5","value":0.4},{"key":"9","value":0.6},{"key":"16","value":0},{"key":"29","value":0},{"key":"54","value":0.4},{"key":"100","value":0}],"app_version":75,"channel":"nightly","client_agg_type":"summed_histogram","key":"","metric":"test","metric_type":"histogram-exponential","os":"first","process":"parent","total_users":2}
```
emtwo: note that `total_users` is 2 here since there is no `app_build_id` and the input yaml is for a single client appearing once for each `app_build_id` they submitted

tests/telemetry_derived/clients_histogram_probe_counts_v1/test_aggregation/expect.ndjson
```diff
@@ -1,6 +1,3 @@
...
-{"agg_type":"histogram","aggregates":[{"key":"0","value":0.4},{"key":"1","value":0},{"key":"2","value":0.2},{"key":"3","value":0},{"key":"5","value":0.4},{"key":"9","value":0.6},{"key":"16","value":0},{"key":"29","value":0},{"key":"54","value":0.4},{"key":"100","value":0}],"app_version":75,"channel":"nightly","client_agg_type":"summed_histogram","key":"","metric":"test","metric_type":"histogram-exponential","process":"parent","total_users":2}
\ No newline at end of file
+{"agg_type":"histogram","aggregates":[{"key":"0","value":0.2},{"key":"1","value":0},{"key":"2","value":0.1},{"key":"3","value":0},{"key":"5","value":0.2},{"key":"9","value":0.3},{"key":"16","value":0},{"key":"29","value":0},{"key":"54","value":0.2},{"key":"100","value":0}],"app_version":75,"channel":"nightly","client_agg_type":"summed_histogram","key":"","metric":"test","metric_type":"histogram-exponential","os":"first","process":"parent","total_users":1}
```
emtwo: Now note that a single user with 2 different `app_build_id`s would have been aggregated and counted as one user in `clients_histogram_bucket_counts` so by the time we get to probe counts, there are no longer 2 users when `app_build_id` is null

emtwo:
@acmiyaguchi this PR is ready for another review pass.

emtwo:
@saptarshiguha mentioning you for review on the relevant computation parts of the PR:

`clients_histogram_bucket_counts_v1` , `clients_histogram_probe_counts_v1`, `clients_scalar_bucket_counts_v1` , and `clients_scalar_probe_counts_v1` for fudging the numbers for the 10% sampling and for the Dirichlet density. Note that we haven't talked so far about scalars so I didn't apply any of the density discussion there, just fudging

sql/telemetry_derived/clients_histogram_probe_counts_v1/query.sql
```diff
@@ -90,14 +70,15 @@ RETURNS ARRAY<STRUCT<key STRING, value FLOAT64>> AS (
...
-CREATE TEMP FUNCTION udf_fill_buckets(input_map ARRAY<STRUCT<key STRING, value FLOAT64>>, buckets ARRAY<STRING>)
+CREATE TEMP FUNCTION udf_fill_buckets(input_map ARRAY<STRUCT<key STRING, value FLOAT64>>, buckets ARRAY<STRING>, total_users INT64)
 RETURNS ARRAY<STRUCT<key STRING, value FLOAT64>> AS (
   -- Given a MAP `input_map`, fill in any missing keys with value `0.0`
   (
     WITH total_counts AS (
       SELECT
         key,
-        COALESCE(e.value, 0.0) AS value
+        -- Dirichlet distribution density for each bucket in a histogram
```
acmiyaguchi, line 80: It'd be useful to cite the document for computing the densities here.

sql/telemetry_derived/clients_histogram_bucket_counts_v1/query.sql
```diff
@@ -40,61 +41,166 @@ RETURNS ARRAY<STRUCT<key STRING, value FLOAT64>> AS (
...
+    agg_type,
+    aggregates,
+    os = 'Windows'and channel = 'release' AS sampled
+  FROM
+    clients_histogram_aggregates_v1
+  CROSS JOIN UNNEST(histogram_aggregates)
+  WHERE submission_date = @submission_date
+    AND first_bucket IS NOT NULL),
+
+aggregated_histograms AS
```
acmiyaguchi, line 68: Does this just move the data cubing computation from probe counts into this query to prevent double counting? Is it possible to collapse the queries together now that the query is processing less data?

sql/telemetry_derived/clients_histogram_bucket_counts_v1/query.sql
```diff
@@ -125,14 +227,15 @@ SELECT
...
   metric,
   metric_type,
-  key,
+  normalized_histograms.key AS key,
   process,
   agg_type,
   STRUCT<key STRING, value FLOAT64>(
-    CAST(bucket AS STRING),
+    CAST(aggregates.key AS STRING),
     1.0 * SUM(value)
```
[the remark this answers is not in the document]
emtwo: yes. `value` is a unique field name here so it didn't need the qualifier like `key` did. But I'll add this in to be more explicit.

sql/telemetry_derived/clients_histogram_bucket_counts_v1/query.sql
emtwo: Yes, the data cubing is moved so it is done per-client first to avoid double counting.

Unfortunately, it doesn't seem possible to collapse the queries. Initially I tried running with a 1% sample and it worked out. However, with the 10% sample, there are shuffle errors.

In fact, doing the data cubing per client for 10% seemed to give shuffle errors even without merging both bucket count + probe count queries. So I have made some changes that I will be pushing shortly to break up the processing into quarters like with histogram aggregates.

emtwo: Also worth noting that data cubing by client increased the amount of processing and therefore the cost by about $100 for the day I tested on.

acmiyaguchi: If you break up the query into smaller pieces, remember that the release windows clients are all taken from consecutive sample ids, so partitions will be uneven unless its adjusted e.g. `sample_id % x == 0`

And where does the extra 20TB of data come from?

sql/telemetry_derived/clients_histogram_bucket_counts_v1/query.sql
```diff
@@ -1,4 +1,4 @@
-CREATE TEMP FUNCTION udf_normalized_sum (arrs ARRAY<STRUCT<key STRING, value INT64>>)
+CREATE TEMP FUNCTION udf_normalized_sum (arrs ARRAY<STRUCT<key STRING, value INT64>>, sampled BOOL)
```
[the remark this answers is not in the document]
emtwo: We do the latter - earlier in the pipeline (not included in this PR), when new data comes in for a user, their histograms for the same metric are grouped together here. `udf.map_sum()` will sum their respective values for a bucket

sql/telemetry_derived/clients_histogram_probe_counts_v1/query.sql
```diff
@@ -90,14 +70,15 @@ RETURNS ARRAY<STRUCT<key STRING, value FLOAT64>> AS (
...
+        SAFE_DIVIDE(COALESCE(e.value, 0.0) + SAFE_DIVIDE(1, ARRAY_LENGTH(buckets)), total_users + 1) AS value
```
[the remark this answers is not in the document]
emtwo: Yes, `clients_histogram_bucket_counts_v1` runs before `clients_histogram_probe_counts_v1`.

So the data has already passed through a multiplication of * 10 here https://github.com/mozilla/bigquery-etl/blob/a049d327ae6fc2be375311e3a1a41da5a6dedac1/sql/telemetry_derived/clients_histogram_bucket_counts_v1/query.sql#L29 

So after `udf_normalized_sum()` is called, all users that were sampled have 10x their value and this gets carried down into all following calculations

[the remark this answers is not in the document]
emtwo: yes correct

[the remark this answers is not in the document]
saptarshiguha: But in this line, the `v/total_count * If(if sampled, 10,1)`  only multiplies the proportions by 10. Shouldn't we also take into client  that this profile has been factored 10 times (so `total_users` takes into account this client has been weighted 10 times). 
For example, if we have 10 windows users with sum of p equal to say 2.4, then after weighting we have 2.4*10 and 10*10 `total_users`

sql/telemetry_derived/clients_histogram_bucket_counts_v1/query.sql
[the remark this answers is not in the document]
saptarshiguha: Got it!

sql/telemetry_derived/clients_histogram_bucket_counts_v1/query.sql
emtwo: Here is what I tested:
The `clients_histogram_aggregates` table had ~15.5T of data based on a 10% sample of windows + release.
a) I ran `histogram_bucket_counts` as it appears now in this PR and that consistently gave a shuffle error.
b) I added a filter for `histogram_bucket_counts` between 0-24 inclusive sample_id  and this ran and processed 15T
c) I then separately ran just the first subquery (without the data cubing) and that processed 6T.

So even though both queries b) and c) should be reading the same amount of data from `clients_histogram_aggregates`, they had very different amounts of data processed. I could only speculate that this was due to internally the `filtered_data` subquery being run multiple times perhaps. Do you have any other intuitions for what this could be?

Based on this, I decided to have the `filtered_data` subquery as its own materialized table.
Now I ran the materialized table for partitioned on sample_id for *all*  sample_ids. This should process ~15.5T (the size of `clients_histogram_aggregates`) - It seemed to have processed 16.44T. I then ran the rest of `histogram_bucket_counts` and `histogram_probe_counts` as one query for each of the quarters of data and this processed a total of 22.46T (again larger than the 16.44T I would have expected, but less dramatic than 6T -> 15T. and maybe a similar issue where the table is being read multiple times with the data cubing).

Overall this cost 22.46T + 16.44T = 38.9T.
If we could just do the first read we wouldn't need that extra 22.46T to be processed.

To confirm that this was cheaper than not having the intermediate materialized table, I also ran `histogram_bucket_counts` with sample_ids between 50-74 inclusive, these would not have included samples. This was 11.7T.
So the cost without the materialized table would be 11.7T * 3 + 15T = 50.1T. So the materialized table saves ~11T.

That's a lot of text but if this happens to give you any money-saving ideas, let me know!

Also, good idea doing %x to keep it even.

emtwo: One more thing, I think the data cubing was a lot less costly before because by the time the data made it into `clients_histogram_bucket_counts` it was around 50GB! So even if that exploded with cubing, it wasn't too crazy.

acmiyaguchi: Based on the [execution details page](https://console.cloud.google.com/bigquery?project=glam-desktop-dev&j=bq:US:bquxjob_5991e116_171849d0c1f&page=queryresults) for the job, we can confirm that the filtered_data subquery is reading from the source table several times (3 times to be exact).

It might be useful to cross join twice on two smaller tables that only contain a null value for `os` and `app_build_id`. I'm not sure what the exact syntax would be off the top of my head, but something like

```sql
SELECT
  ...
FROM
   filtered_data
CROSS JOIN (SELECT null as os)
USING (os)
CROSS JOIN (SELECT null as app_build_id)
USING (app_build_id)
```

sql/telemetry_derived/clients_histogram_probe_counts_v1/query.sql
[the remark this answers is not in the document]
emtwo: > For example, if we have 10 windows users with sum of p equal to say 2.4, then after weighting we have 2.4 * 10 and 10 * 10 `total_users`

This should be included implicitly. Here is hopefully a more clear walk-through with an example.

Given the [example data here](https://console.cloud.google.com/bigquery?project=glam-desktop-dev&j=bq:US:bquxjob_306433d9_17199217d12&page=queryresults), we start with 2 users who are each part of a sample taken for Windows + release.

The output of the intermediate `normalized_histograms` table is:
Screen Shot 2020-04-20 at 3 46 51 PM
each users normalized values were multiplied by 10 so if you sum all of their values the total is 10

The output of the intermediate `bucket_counts` table is:
Screen Shot 2020-04-20 at 3 49 36 PM
the buckets for both users are grouped/summed together from the previous step. If all values of `record.value` are summed, we should get 20 users.

So to compute `total_users` in the last step, we sum `record.value` and see that `total_users` is 20 (which is 2 * 10 as expected).
Screen Shot 2020-04-20 at 3 49 58 PM

Hopefully this clarifies what is happening here but please let me know if something is still missing.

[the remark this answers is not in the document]
saptarshiguha: Ah yes, this is implicitly carried over as you describe. Thanks for the detailed explanation.

sql/telemetry_derived/clients_histogram_bucket_counts_v1/query.sql
emtwo: I'm not 100% sure what your intention was with this example since `CROSS JOIN` doesn't seem to work with `USING` syntax. I tried your suggestion that you mentioned in slack with unioning the datasets before the group by [here](https://console.cloud.google.com/bigquery?project=glam-desktop-dev&j=bq:US:bquxjob_33f796c0_17198e1523a&page=queryresults). Unfortunately it seemed to have the same number of bytes processed as [here](https://console.cloud.google.com/bigquery?project=glam-desktop-dev&j=bq:US:bquxjob_7672ddad_17198ac1dd1&page=queryresults) where there are multiple group by clauses.

script/glam/run_glam_sql
```diff
@@ -100,25 +104,62 @@ function run_view {
...
+    local PARTITION_SIZE=$((NUM_SAMPLE_IDS / NUM_PARTITIONS))
+    local table_name=$1
+    local partitioned_result=${2:-true}
+
+    run_query $table_name true 0 19
+    for partition in $(seq 1 $((NUM_PARTITIONS - 1))); do
+        local min_param=$((partition * PARTITION_SIZE))
+        local max_param=$((min_param + PARTITION_SIZE - 1))
+        run_query $table_name $partitioned_result $min_param $max_param \
+            $table_name "--append_table --noreplace"
```
Ben-Wu, line 119: Isn't `--noreplace` only for `bq load`?  I don't think it does anything for `bq query`

sql/telemetry_derived/clients_histogram_bucket_counts_v1/query.sql
emtwo: Very cool that seems to have worked! Thanks for this idea!! And now I've updated the PR to get rid of `histogram_bucket_counts`. We can probably get rid of bucket counts for scalars too but this PR is getting hefty and I want to get it out the door

acmiyaguchi: Do you mind sharing the processed/slot time/shuffled/spilled stats (or links to the specific execution) for the baseline vs updated versions of the query when you get a chance? I'm curious how it turned out.

emtwo: Yea sure.

[Without static combos](https://console.cloud.google.com/bigquery?project=glam-desktop-dev&j=bq:US:bquxjob_7672ddad_17198ac1dd1&page=queryresults): 482.3gb processed.
[With static combos](https://console.cloud.google.com/bigquery?project=glam-desktop-dev&organizationId=442341870013&j=bq:US:bquxjob_aee9f96_171a816853d&page=queryresults): 120.6gb processed

PR #882, opened 2020-04-04
Public data - Publish metadata for navigating the HTTP endpoint

scholtzan:
This adds a script which publishes metadata files to GCS:
* `all_datasets.json`: contains information about all available datasets and how to navigate to them
* table metadata: contains a list of available files for a specific table

For a full specification see: https://docs.google.com/document/d/1zJtRTiiBxpE6faf-HNL6IVTXIFHXvceKCxWXYM0wf9w/edit# 

I extended the `entrypoint` script so that it will be possible to run publishing this metadata from Airflow.


bigquery_etl/public_data/publish_gcs_metadata.py
```diff
@@ -0,0 +1,202 @@
...
+        metadata_json["friendly_name"] = self.metadata.friendly_name
+        metadata_json["description"] = self.metadata.description
+        metadata_json["incremental"] = self.metadata.is_incremental()
+        metadata_json["incremental_export"] = self.metadata.is_incremental_export()
+
+        if self.metadata.review_bug() is not None:
+            metadata_json["review_link"] = REVIEW_LINK + self.metadata.review_bug()
+
+        metadata_json["files_uri"] = self.files_uri
+        # todo: add last updated
```
scholtzan, line 81: I'll open an issue for publishing information about the last time table data was updated

script/entrypoint
```diff
@@ -33,6 +33,12 @@ elif [ "$1" = "query" ]; then
     # For queries with public_json set in the metadata, the results will be
     # exported as JSON to Cloud storage.
     python script/publish_public_data_json --query_file="${@: -1}" ${@:1:$#-1}
+elif [ "$1" = "publish_gcs_metadata" ]; then
+    # For an invocation like:
+    #     publish_gcs_metadata [options]
+    # we dispatch to a script that publishes metadata files to GCS for datasets
+    # that have previsously been made public on GCS.
+    exec script/publish_public_data_gcs_metadata ${@:2}
```
scholtzan, line 41: This will be needed for Airflow

tests/public_data/test_sql/test/incremental_query_v1/metadata.yaml
```diff
@@ -1,6 +1,5 @@
-friendly_name: Test table for an incremental query
```
scholtzan, line 1: Turns out that writing it this way will result in a newline character when parsing using `yaml`

tests/public_data/test_publish_gcs_metadata.py
```diff
@@ -0,0 +1,232 @@
...
+from unittest.mock import call, Mock, MagicMock
+
+import bigquery_etl.public_data.publish_gcs_metadata as pgm
+
+
+class TestPublishGcsMetadata(object):
+    test_bucket = "test-bucket"
+    project_id = "test-project-id"
+    api_version = "v1"
+    endpoint = "http://test.endpoint.mozilla.com/"
```
jklukas, line 14: We should probably use https:// consistently even in tests.

bigquery_etl/public_data/publish_gcs_metadata.py
```diff
@@ -0,0 +1,202 @@
...
+
+        output_file = f"gs://{args.target_bucket}/all_datasets.json"
+        publish_all_datasets_metadata(gcs_table_metadata, output_file)
+        publish_table_metadata(gcs_table_metadata, args.target_bucket)
+    else:
+        print(
+            """
+            Invalid target: {}, target must be a directory with
+            structure /<dataset>/<table>/metadata.yaml.
+            """.format(
```
jklukas, line 195: This multiline string will have be printed with indentation we don't want. We may want to change this to several single line string literals.

tests/public_data/all_datasets.json
```diff
@@ -0,0 +1 @@
+{"test": {"non_incremental_query": {"v1": {"friendly_name": "Test table for a non-incremental query", "description": "Test table for a non-incremental query", "incremental": false, "incremental_export": false, "review_link": "https://bugzilla.mozilla.org/show_bug.cgi?id=1999999", "files_uri": "http://test.endpoint.mozilla.com/api/v1/tables/test/non_incremental_query/v1/files"}}, "incremental_query": {"v1": {"friendly_name": "Test table for an incremental query", "description": "Test table for an incremental query", "incremental": true, "incremental_export": true, "files_uri": "http://test.endpoint.mozilla.com/api/v1/tables/test/incremental_query/v1/files"}}}}
```
jklukas, line 1: For these metadata files, the overall size is going to be fairly small, so we may want to do a multiline pretty-print to increase usability if the user happens to be looking at this without their browser parsing the json for them.

script/entrypoint
jklukas: We can call the full path to the script (`script/publish_public_data_gcs_metadata`) from Airflow. That's what we do for copy_deduplicate, and I think we should stick to that rather than adding additional logic here for new cases.

PR #883, opened 2020-04-04
Fix improper client counting from multiple joins

fbertsch:
In this query we are using CROSS JOIN UNNEST to create
extra groups. This is effective only if you are grouping
on those groups, or else those clients end up double counted.

However, we were doing another CROSS JOIN UNNEST to create
more country groups. What this resulted in was extra counting
of the client for every country group they were in. This actually
made non-broken out countries unaffected, but doubled the counts
of broken-out countries (since they have two groups), and ~30%
increase in Overall counts (from the broken out countries).

The fix here is to remove the extra CROSS JOIN UNNEST for country
bucketing, and instead append that array to the country_group
break-out.


fbertsch:
I have confirmed this works as intended by running a similar, modified query where instead of e.g. `COUNT(*)` and `COUNTIF`, I use `COUNT(DISTINCT IF(..., client_id, NULL))`, which will give correct results even if the bucketing results in extra rows.

sql/org_mozilla_firefox_derived/incline_executive_v1/query.sql
```diff
@@ -150,28 +150,20 @@ counts AS (
...
   -- sets for each one of these fields. They create
-  -- a row with 'Overall' for value and NULL for value
-  -- for each row, which allows us to count that row
-  -- for both fields. When we COALESCE the null above,
-  -- we end up grouping on both that rows natural value,
-  -- and 'Overall', which contains all rows.
+  -- a row with 'Overall' and that rows value
   CROSS JOIN
-    UNNEST(['Overall', NULL]) AS is_migrated_group
+    UNNEST(['Overall', IF(is_migrated, 'Yes', 'No')]) AS is_migrated_group
```
fbertsch, line 160: I also removed the confusing `COALESCE` and `NULL` pieces, they're unnecessary.

PR #887, opened 2020-04-07
Use baseline_clients_last_seen for Fenix

jklukas:
This implements the [proposed Fenix KPI reporting changes](https://docs.google.com/document/d/1Ym4eZyS0WngEP6WdwJjmCoxtoQbJSvORxlQwZpuSV2I/edit#heading=h.69hvvg35j8un) in GUD.

PR #888, opened 2020-04-07
Count DOH events for each user only once in ASN aggregates

scholtzan:
This should fix the issue that we count only whether clients report enable/disable DOH events, instead of the number of reported events.

### DIFF
--- a/sql/telemetry_derived/asn_aggregates_v1/query.sql
+++ b/sql/telemetry_derived/asn_aggregates_v1/query.sql
@@ -84,7 +84,7 @@ events_with_ip AS (
 events_with_asn AS (
   -- Lookup ASNs for IP addresses.
   SELECT
-    submission_date,
+    DISTINCT submission_date,
     client_id,
     canary,
     event_category,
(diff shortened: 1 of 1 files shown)


sql/telemetry_derived/asn_aggregates_v1/query.sql
```diff
@@ -105,6 +105,13 @@ events_with_asn AS (
     asn_ip_address_range
   USING
     (network_ip, mask)
+  GROUP BY
+    submission_date,
+    client_id,
+    canary,
+    event_category,
+    event_object,
+    autonomous_system_number
```
scholtzan, line 114: Grouping them here should ensure that for each client only up to one doh event will be aggregated in the final part of the query.

scholtzan: This way `n_clients` would still refer to the total number of distinct client IDs part of a specific ASN

tdsmith: It's a little more concise to write SELECT DISTINCT at line 86.

scholtzan:
CI is currently failing due to https://github.com/mozilla/bigquery-etl/pull/887

PR #892, opened 2020-04-09
Bug 1624492 - Use weighted quantile function for glean percentiles in GLAM

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1624492)

This copies reuses the [`weightedQuantile` function in GLAM](https://github.com/mozilla/glam/blob/e6d345d0e61df23549ece416e2d41e800980de41/src/utils/stats.js).


udf_js/weighted_quantile.sql
```diff
@@ -0,0 +1,85 @@
...
+  assert_equals(
+    1,
+    udf_js.weighted_quantile(
+      50.0,
+      ARRAY<STRUCT<key STRING, value FLOAT64>>[("0", 3), ("1", 3), ("2", 3)],
+      "timing_distribution"
+    )
+  ),;
+
+#xfail
```
Ben-Wu, line 79: What's the purpose of this case?

udf_js/lib/wtdstats.js
```diff
@@ -0,0 +1,75 @@
...
+  // returns yOut by mapping x -> xOut, finding the index, then applying to y
+  // constant (nearest-neighbors) interpolation with f=1 (always right-sided)
+  // find ind in x where xOut values should below, then look at y[ind]
+  // assume x & y are pre-sorted and match.
+  const xInds = xOut.map((xi) => x.findIndex((xx) => xx > xi));
+  const low = Math.min(...x);
+  const high = Math.max(...x);
+  return xInds.map((i) => {
+    // left & right censor extreme values.
+    if (i === -1 && xOut[i] <= low) return y[0];
```
Ben-Wu, line 46: I know this is just directly from glam but this function looks a little sketchy and is hard to understand.  I think the `xInds.map` logic is wrong.  If `i === -1`, then `xOut[-1]` is undefined and so these clauses are always false and y[i] is always returned which means y[-1] can be returned.  This is what's causing the `NaN` when weights are 1.  I think there's some other logic elsewhere that's causing the difference between weights 1 and 3 but I'm not sure where.  Probably something to talk to Hamilton about and maybe it works fine in the cases we need to use it in but I have trouble trusting this code.

udf_js/weighted_quantile.sql
acmiyaguchi: This asserts that the percentile is within the range [0, 100].

udf_js/weighted_quantile.sql
```diff
@@ -0,0 +1,85 @@
...
+  -- skew right
+  assert_equals(
+    2,
+    udf_js.weighted_quantile(
+      50.0,
+      ARRAY<STRUCT<key STRING, value FLOAT64>>[("0", 10), ("1", 10), ("2", 100)],
+      "timing_distribution"
+    )
+  ),
+    -- Bug when all weights are less than 2. This test fails with a `400
```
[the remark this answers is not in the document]
acmiyaguchi: Yes, it seems to be that there are bugs from the few tests that I've added here. The idea is that the stats code would be factored out of GLAM so it can be reused here too, which would fix some of the bugs here. The UDF wouldn't be located within the repository, but rather an external library that could be referenced directly in gcs or uploaded to the udf lib bucket with some minor modifications.

udf_js/lib/wtdstats.js
```diff
@@ -0,0 +1,75 @@
+// https://github.com/mozilla/glam/blob/e6d345d0e61df23549ece416e2d41e800980de41/src/utils/stats.js
+(function() {
+function sum(vs) { return vs.reduce((a, b) => a + b, 0); }
+
+ function ascending(a, b) {
```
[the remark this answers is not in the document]
acmiyaguchi: It's a leading space (since I generally trim trailing whitespace before committing things nowadays).

udf_js/lib/wtdstats.js
acmiyaguchi: It might be worthwhile to wait on this PR then until more testing is done on the underlying code.

Ben-Wu: That would be ideal as long as this doesn't block anything that's very time-sensitive.

hamilton: yes, I wouldn't probably move forward with this PR until we actually write [tests of my port](https://github.com/mozilla/glam/issues/351) & more thoroughly review the code. Moving away from the old `histogram.js` implementation is going to be a pretty big shift, so it is worth doing it right this time around.

irrationalagent:
I do have a use-case for this currently as well.

Issue #861, opened 2020-03-30
Canonical mapping of mobile product names


jklukas:
See https://github.com/mozilla/firefox-data-docs/pull/431#discussion_r400368399

> I have heard many people voice a need for a map between different ways of labeling products. It gets very confusing across stakeholders.

We could create a `static` table that gives a canonical mapping of low-level `app_name`/`channel`/`app_id` to high-level application name, but it's a nuanced problem. The needed field differs between core pings and Glean-based apps, etc.

acmiyaguchi:
I believe `mozfun.norm.fenix_app_info` has resolved most of the issues here (https://github.com/mozilla/bigquery-etl/blob/master/sql/mozfun/norm/fenix_app_info/metadata.yaml), which is used most recently in this PR https://github.com/mozilla/bigquery-etl/pull/1380

PR #894, opened 2020-04-09
Add additional Glean apps to GUD tables

jklukas:
We will be adding some of these to the frontend via https://github.com/mozilla/GUD/pull/77

### DIFF
--- a/script/format_sql
+++ b/script/format_sql
@@ -67,7 +67,6 @@ SKIP = {
     "sql/telemetry/fxa_users_services_first_seen/view.sql",
     "sql/telemetry/fxa_users_services_last_seen/view.sql",
     "sql/telemetry/lockwise_mobile_events_v1/view.sql",
-    "sql/telemetry/nondesktop_clients_last_seen_v1/view.sql",
     "sql/telemetry/origin_content_blocking/view.sql",
     "sql/telemetry/rocket_android_events_v1/view.sql",
     "sql/telemetry/simpleprophet_forecasts/view.sql",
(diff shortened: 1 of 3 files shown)


script/format_sql
```diff
@@ -67,7 +67,6 @@ SKIP = {
     "sql/telemetry/fxa_users_services_first_seen/view.sql",
     "sql/telemetry/fxa_users_services_last_seen/view.sql",
     "sql/telemetry/lockwise_mobile_events_v1/view.sql",
-    "sql/telemetry/nondesktop_clients_last_seen_v1/view.sql",
```
jklukas, line 70: We allow the auto-formatter to run on this file now, so there are some whitespace changes below.

sql/telemetry/nondesktop_clients_last_seen_v1/view.sql
```diff
@@ -1,37 +1,70 @@
...
-    * REPLACE('nightly' AS normalized_channel),
+    * REPLACE ('nightly' AS normalized_channel),
     'Fenix' AS app_name,
   FROM
     `moz-fx-data-shared-prod.org_mozilla_fennec_aurora.baseline_clients_last_seen`
+  -- Other apps that send Glean telemetry
+  UNION ALL
+  SELECT
+    *,
+    'Lockwise Baseline' AS app_name,
```
acmiyaguchi, line 41: Why suffix these with `Baseline` instead of following the convention with the Fenix apps?

jklukas: Fenix is the only Glean-based app that is included in KPIs. I needed a way of making the rest of these identifiable with a common prefix or suffix so we can filter them out in the smoot query for the Any Mobile Activity aggregation.

jklukas: We need a more coherent story around how we present app names as captured in https://github.com/mozilla/bigquery-etl/issues/861

PR #895, opened 2020-04-09
Remove legacy per-app Glean usage ETL

jklukas:
The incline dashboard, KPIs, and GUD are now referencing the generic glean ETL
(baseline_clients_daily and baseline_clients_last_seen), so these other
tables should no longer be needed.

I also need to remove these from Airflow and I will need to send out a message
to fx-data-dev to advertise this breaking change, but I doubt there's much
usage so far of these legacy tables since they have not been well documented
or advertised; they existed primarily to serve GUD, Incline, and KPIs.


jklukas:
>     1. How have you confirmed that no one is using the metrics derived fields?

I don't have proof of this, but I will send out an announcement with at least a few days for folks to report uses before I merge this and delete things.

>     2. Will we need to backfill with these schema-incompatible changes?

The generic glean ETL is all new tables `baseline_clients_daily` and `baseline_clients_last_seen`. We are simply deleting tables here, not altering any schemas.

jklukas:
This was done in https://github.com/mozilla/bigquery-etl/pull/925 instead.

Issue #896, opened 2020-04-09
generate_views: TypeError: 're.Match' object is not iterable


whd:
Example invocation (affects both stage and prod): `script/generate_views moz-fx-data-shar-nonprod-efed:*_stable.* --sql-dir /tmp/sql`

Example traceback:

    Traceback (most recent call last):
      File "/usr/lib64/python3.7/runpy.py", line 193, in _run_module_as_main
        "__main__", mod_spec)
      File "/usr/lib64/python3.7/runpy.py", line 85, in _run_code
        exec(code, run_globals)
      File "/home/wes/source/bigquery-etl/bigquery_etl/generate_views.py", line 159, in <module>
        main()
      File "/home/wes/source/bigquery-etl/bigquery_etl/generate_views.py", line 80, in main
        create_views_if_not_exist(client, views, args.exclude, args.sql_dir)
      File "/home/wes/source/bigquery-etl/bigquery_etl/generate_views.py", line 100, in create_views_if_not_exist
        for table in tables
      File "/home/wes/source/bigquery-etl/bigquery_etl/generate_views.py", line 101, in <genexpr>
        for match in VERSION_RE.search(table)
    TypeError: 're.Match' object is not iterable

This appears to be a regression and not specific to any stage or prod state. If I use strings for table/views like the following it fails the same way:

    #!/usr/bin/env python
    
    import re
    
    view = 'moz-fx-data-shared-prod.activity_stream_stable.events'
    tables = ['moz-fx-data-shared-prod.activity_stream_stable.events_v1']
    VERSION_RE = re.compile(r"_v([0-9]+)$")
    
    print(view)
    print(tables)
    version = max(
        int(match.group()[2:])
        for table in tables
        for match in VERSION_RE.search(table)
        if match is not None
    )
    

@scholtzan is going to investigate.

PR #898, opened 2020-04-09
Fix generate_views version parsing

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/896

`re.search` doesn't return an iterable, because of that an error is thrown in the list comprehension for determining the table version.

### DIFF
--- a/bigquery_etl/generate_views.py
+++ b/bigquery_etl/generate_views.py
@@ -93,10 +93,9 @@ def create_views_if_not_exist(client, views, exclude, sql_dir):
             continue
 
         version = max(
-            int(match.group()[2:])
+            int(VERSION_RE.search(table).group()[2:])
             for table in tables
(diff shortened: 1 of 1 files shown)


bigquery_etl/generate_views.py
```diff
@@ -93,10 +93,9 @@ def create_views_if_not_exist(client, views, exclude, sql_dir):
             continue
 
         version = max(
-            int(match.group()[2:])
+            int(VERSION_RE.search(table).group()[2:])
             for table in tables
-            for match in VERSION_RE.search(table)
-            if match is not None
+            if VERSION_RE.search(table) is not None
```
relud, line 98: Instead of calling `VERSION_RE.search` twice I think this should do `for match in [VERSION_RE.search(table)]`

PR #900, opened 2020-04-13
Bug 1629099 - Create a 1-Week Retention View

relud:


sql/telemetry/retention_1_week/view.sql
```diff
@@ -0,0 +1,13 @@
+CREATE OR REPLACE VIEW
+  `retention`
```
jklukas, line 2: This name should match the filename. And perhaps we should make this name desktop-specific? Perhaps `desktop_retention_1_week`?

Or, would we potentially be better off with `clients_last_seen_1_week_retention` or `clients_last_seen_day_13` or `clients_last_seen_active_last_week` where we do `SELECT * REPLACE(DATE_SUB(submission_date, INTERVAL 13 DAY) AS submission_date)` and add the `retained` column.

relud: `desktop_retention_1_week` sounds best of those options to me.

jmccrosky:
This looks good.  One thing I'll mention is that it's a little mentally jarring to me that I can filter to a particular "submission_date" and probably be looking at clients that were note active on that date.  Maybe we just call it "date" or even "week_0_beginning_date"?

I don't feel strongly though.

jklukas:
> This looks good. One thing I'll mention is that it's a little mentally jarring to me that I can filter to a particular "submission_date" and probably be looking at clients that were note active on that date. Maybe we just call it "date" or even "week_0_beginning_date"?
> 
> I don't feel strongly though.

I agree that `date` would be appropriate here.

sql/telemetry/desktop_retention_1_week/view.sql
```diff
@@ -0,0 +1,13 @@
+CREATE OR REPLACE VIEW
+  `desktop_retention_1_week`
+AS
+SELECT
+  client_id,
+  DATE_SUB(submission_date, INTERVAL 13 DAY) AS submission_date,
```
jklukas, line 6:
```suggestion
  DATE_SUB(submission_date, INTERVAL 13 DAY) AS `date`,
```

Issue #886, opened 2020-04-06
Add last_updated to GCS dataset metadata


scholtzan:
The metadata for `all_datasets.json` should contain a `last_updated` entry for each public table. This will link to a document that contains the timestamp of when the table data was last updated on GCS.

PR #902, opened 2020-04-13
Add last_updated to GCS datasets metadata

scholtzan:
This fixes https://github.com/mozilla/bigquery-etl/issues/886

This writes the timestamp of when the data of a dataset on GCS was last updated to GCS. In the `all_datasets.json` metadata file, for each dataset a `last_updated` entry got added which links to the corresponding `last_updated` file for this dataset.

### DIFF
--- a/bigquery_etl/public_data/publish_gcs_metadata.py
+++ b/bigquery_etl/public_data/publish_gcs_metadata.py
@@ -16,7 +16,7 @@ from ..util import standard_args
 
 DEFAULT_BUCKET = "mozilla-public-data-http"
 DEFAULT_API_VERSION = "v1"
-DEFAULT_ENDPOINT = "http.public-data.prod.dataops.mozgcp.net/"
+DEFAULT_ENDPOINT = "https://public-data.telemetry.mozilla.org/"
 REVIEW_LINK = "https://bugzilla.mozilla.org/show_bug.cgi?id="
 GCS_FILE_PATH_RE = re.compile(
     r"api/(?P<api_version>.+)/tables/(?P<dataset>.+)/(?P<table>.+)/(?P<version>.+)/"
@@ -80,6 +80,7 @@ class GcsTableMetadata:
         if self.metadata.review_bug() is not None:
             metadata_json["review_link"] = REVIEW_LINK + self.metadata.review_bug()
 
+        metadata_json["files_metadata"] = self.files_uri + "/metadata.json"
         metadata_json["files_uri"] = self.files_uri
         metadata_json["last_updated"] = self.last_updated_uri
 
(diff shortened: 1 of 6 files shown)


bigquery_etl/public_data/publish_gcs_metadata.py
```diff
@@ -164,6 +167,16 @@ def publish_table_metadata(table_metadata, bucket):
             fout.write(json.dumps(metadata.files_metadata_to_json(), indent=4))
 
 
+def publish_last_modified(table_metadata, bucket):
```
jklukas, line 170: We're using a mix of `last_updated` and `last_modified`. My preference would be for `last_modified` as that seems more familiar, but the important thing is to be consistent everywhere we use the term.

bigquery_etl/public_data/publish_gcs_metadata.py
```diff
@@ -51,21 +51,25 @@
...
+        (self.dataset, self.table, self.version) = dataset_table_version_from_gcs_blob(
+            self.blobs[0]
         )
         self.metadata = Metadata.of_table(
             self.dataset, self.table, self.version, target_dir
         )
 
+        self.last_updated_path = self.blobs[0].name.split("files")[0] + "last_updated"
+        self.last_updated_uri = endpoint + self.last_updated_path
+        self.last_updated = min(self.blobs, key=lambda b: b.updated).updated
```
jklukas, line 71:
```suggestion
        self.last_updated = min(b.updated for b in blobs)
```

Would this be equivalent to what you have written here?

bigquery_etl/public_data/publish_gcs_metadata.py
```diff
@@ -78,7 +82,7 @@ def table_metadata_to_json(self):
             metadata_json["review_link"] = REVIEW_LINK + self.metadata.review_bug()
 
         metadata_json["files_uri"] = self.files_uri
-        # todo: add last updated
+        metadata_json["last_updated"] = self.last_updated_uri
```
jklukas, line 85: Should we call the key `last_updated_uri`? I would naively expect a field called `last_updated` to be a timestamp string.

scholtzan:
@fbertsch In the public data specification, the `last_updated` timestamp is described to be stored in a separate file. The metadata file containing information about all available datasets links to these files. Would it make more sense to instead directly write the `last_updated` timestamp to the metadata file? 

Basically, change it to:

```diff
dataset_name:
  table_name:
    v1:
      review_link: <link to public data review>
      friendly_name: “A friendly name”
      description: “A description of the data contained”
      incremental: “true”
-     last_updated: >
-        https://public-data.telemetry.mozilla.org/api/v1/tables
-        /dataset_name/table_name/v1/last_updated
+     last_updated: "2020-04-23 11:23:02"
      files_uri: >
        https://public-data.telemetry.mozilla.org/api/v1/tables
        /dataset_name/table_name/v1/files
```

scholtzan:
Alright, that makes sense. I made some changes so that `last_updated` gets published to GCS when the data actually gets exported

PR #904, opened 2020-04-13
Get histogram type from probeinfo service (fixes #703)

### LINKED ISSUE #703: Distinguish enumerated from linear histograms in Glam
Basically these two types of histograms are labelled as linear because they contain numeric keys.

However, some of them are considered enumerated, and the only way to tell is by looking at the description in probe dictionary (https://probeinfo.telemetry.mozilla.org/firefox/all/main/all_probes). Sometimes there is explicit text labeling it `enumerated` and sometimes you just have to read the description and make a judgement.

So given that the enumerated ones are a small number, we can make a list and label them correctly so that they can be handled/graphed correctly by the front end.

Ben-Wu:
Use probe info service to get histogram type instead of using the type value stored in the main ping.

Here are some results on a 1% sample showing how histogram types would be affected.  Screenshots because redash isn't working for me right now.

histograms:
hist

keyed histograms:
keyed_hist


emtwo:
We will need to run a one-time query to migrate all the existing metric types assigned to each row in `clients_histogram_aggregates_v1` to the new metric type that comes from the probe info service. 

We need to do this since the `metric_type` is a field that we group rows on and this would cause duplicate rows when the new types came in.

I'd say let's wait for https://github.com/mozilla/bigquery-etl/pull/881 to merge before doing this so there is less data to scan and update in that case.

emtwo:
@Ben-Wu #881 has merged so we can get this one in soon. I'm just wondering if you can use the probe info service to create the one-time query I described above that would update `clients_histogram_aggregates_v1`? We can include it as part of the review for this PR.

Ben-Wu:
Yes I can do that

Ben-Wu:
Will filtering out invalid build ids https://github.com/mozilla/bigquery-etl/issues/876 also need to be done retroactively?  Since this requires scanning all of `clients_histogram_aggregates_v1` it will take some time so it might make sense to do both of these together if needed.

emtwo:
> Will filtering out invalid build ids #876 also need to be done retroactively? Since this requires scanning all of `clients_histogram_aggregates_v1` it will take some time so it might make sense to do both of these together if needed.

Yes! Good catch, the same type of thing will be needed for build ids.

Ben-Wu:
Now rebased and I have the query to update here https://gist.github.com/Ben-Wu/1479738680104d61b09fc0bb73d6ed5b

I did some validation for the total number of histograms before, after, and with the two full joined on all the columns except metric type to make sure none of the other values changed.  Here's a query but it's probably missing some context: https://sql.telemetry.mozilla.org/queries/70882/source

emtwo:
The glam dag is stable and up to date now so feel free to run the migration query and merge this when you are ready. Probably best not to do it on a Friday in case there are failures over the weekend for any reason. Let me know if you need anything else for this one.

Ben-Wu:
Both `clients_histogram_aggregates` and `clients_scalar_aggregates` have been backfilled with this change and the build id filter.  Backups are in `telemetry_derived` as `clients_histogram_aggregates_v1_backup_200514` and `clients_scalar_aggregates_v1_backup_200514`

PR #909, opened 2020-04-15
Run publish_public_data_json as bash script in entrypoint

scholtzan:
Fixes failing Airflow jobs


scholtzan:
I'll write some tests for this since it's not the first time something like this happened :/

PR #910, opened 2020-04-15
Tests for metadata parsing

scholtzan:
Tests for metadata parsing and refactoring of existing public data tests


bigquery_etl/parse_metadata.py
```diff
@@ -29,7 +29,13 @@ def is_valid_label(label):
         63 characters, and cannot be empty. Values can be empty, and have
         a maximum length of 63 characters.
         """
-        return re.match(r"[\w\d_-]+", label) and len(label) <= 63
+        return (
+            len(label) > 0
+            and re.match(r"[\w\d_-]+", label)
+            and len(label) <= 63
+            and (label.isdigit() or label.islower())
```
jklukas, line 36: I don't understand this line. I would expect label could be a mix of digits and lowercase letters, but this seems to enforce that it's either all digits or all letters,

tests/public_data/test_publish_public_data_json_script.py
```diff
@@ -125,6 +146,8 @@ def test_script_non_incremental_query(self):
 
     @pytest.mark.dependency(name="test_script_non_incremental_export")
     def test_script_non_incremental_export(self):
+        print(self.incremental_non_incremental_export_sql_path)
```
jklukas, line 149: Should this line be removed?

tests/data/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: "Test metadata file"
+owners:
+  - ascholtz@mozilla.com
+labels:
+  schedule: daily
+  public_json: true
+  incremental: true
+  incremental_export: true
+  invalid.label: foo
+  invalid_value: Fo.
```
jklukas, line 10: We should add test cases with mixed numbers and letters

bigquery_etl/parse_metadata.py
scholtzan: The problem here I was running into was that `"123".islower() == False` while `"123a".islower() == True`. 
I guess using `label.lower() == label` covers these cases better though.

bigquery_etl/parse_metadata.py
```diff
@@ -29,7 +29,13 @@ def is_valid_label(label):
         63 characters, and cannot be empty. Values can be empty, and have
         a maximum length of 63 characters.
         """
-        return re.match(r"[\w\d_-]+", label) and len(label) <= 63
+        return (
+            len(label) > 0
+            and re.match(r"[\w\d_-]+", label)
+            and len(label) <= 63
+            and label.lower() == label
```
jklukas, line 36: It might be better to build this into the regex. `re.fullmatch(r"[0-9a-z-_]{1,63}")` would handle all of these checks, I believe. It would disallow international characters, but it feels safer to disallow those anyway (and note as much in the docstring).

scholtzan: Yeah, the simplification is probably worth not allowing international characters

Issue #750, opened 2020-02-15
Document command to run all CI checks locally


acmiyaguchi:
Currently, checks for python formatting and linting are documented only in the CI configuration file. It would be useful to run linting checks locally to avoid running into CircleCI failures on pushes to a branch.

Something along the lines of `circleci local execute` would be useful to have in the README.

scholtzan:
I added documentation in https://github.com/mozilla/bigquery-etl/pull/914 since I had to figure this out for setting up integration tests.

dexter030199:
ff

PR #914, opened 2020-04-16
CI build step for integration tests

scholtzan:
When running integration tests it currently fails with: 

```
script/entrypoint: line 25: exec: pytest: not found
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,6 @@ jobs:
           name: Verify that BigQuery validates each query
           command: script/dryrun
   integration:
-    <<: *build
     docker:
     - image: python:3.8
     steps:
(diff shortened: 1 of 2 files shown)


scholtzan:
Is there a good way to test this locally?

jklukas:
> Is there a good way to test this locally?

You can try using the CircleCI CLI: https://circleci.com/docs/2.0/local-cli/

It's been a while since I've used it. It was a pretty nice experience, but had some holes when creds, etc. are involved.

scholtzan:
I guess that's also a good opportunity to tackle https://github.com/mozilla/bigquery-etl/issues/750

.circleci/config.yml
```diff
@@ -57,6 +57,12 @@ jobs:
             echo "Cannot pass creds to forked PRs, so marking this step successful"
             circleci step halt
           fi
+    - &build
+      run:
+        name: Build
+        command: |
+          python3.8 -m venv venv/
+          venv/bin/pip install --upgrade -r requirements.txt
```
jklukas, line 65: The `&build` anchor is already defined earlier in this file, so we should be able to inject the contents here. I'm actually confused about the `<<: *build` line earlier in the `integration` job configuration, as that looks to be injecting content from the anchor, but at the wrong level here. I may have missed that in an earlier review. I can't remember exactly what the right yaml syntax should be here.

scholtzan: Figured it out.

scholtzan:
I figured out how to run CircleCI successfully locally and added documentation.

PR #915, opened 2020-04-17
Add the Windows 10 May 2020 Update to Windows 10 Aggregates

rvandermeulen:
Microsoft will be rolling out the May 2020 update (version 2004) soon. Update the query to break on the final build number for it, 19041.


rvandermeulen:
One note, when I previously did this (#499), I had to make the changes in the now-gone templates/ directory as well. I didn't see any obvious other place this change was needed, but I'm happy to revise if I missed something.

jklukas:
> One note, when I previously did this (#499), I had to make the changes in the now-gone templates/ directory as well. I didn't see any obvious other place this change was needed, but I'm happy to revise if I missed something.

`templates/` has been removed and you're doing the right thing making changes under `sql/` directly.

jklukas:
This should get deployed to prod in the next few days.

PR #917, opened 2020-04-17
Update public data section in README

scholtzan:
Updated the public data section now that the final public domain is set up


scholtzan:
👍 Currently adding docs to docs.tmo for it

PR #918, opened 2020-04-20
Lower-friction scheduled queries config parsing

scholtzan:
This is the first step for providing support for lower-friction scheduled queries. I think it makes sense to break adding that support into multiple PR to make reviewing a little easier.

This PR implements the parsing of the `dags.yaml` file and parsing of scheduling information in `metadata.yaml` files. The parsed config information is used to create `Dag` and `Task` objects which represent Airflow DAGs and tasks. Eventually, these will be used to generate the Airflow code for scheduling queries.

If this looks like it's on the right track, I'll also implement more tests.


tests/query_scheduling/test_generate_airflow_dags.py
```diff
@@ -0,0 +1,28 @@
...
+        assert dags.dag_by_name("bqetl_core") is not None
+
+        events_dag = dags.dag_by_name("bqetl_events")
+        assert len(events_dag.tasks) == 2
+        assert events_dag.tasks[0].args["depends_on_past"] is False
+        assert events_dag.tasks[1].args["depends_on_past"] is False
+
+        core_dag = dags.dag_by_name("bqetl_core")
+        assert len(core_dag.tasks) == 1
+        assert core_dag.tasks[0].args["depends_on_past"]
```
scholtzan, line 28: More tests to come. I just wanted to make sure that I'm on the right track before adding them.

bigquery_etl/query_scheduling/dag.py
```diff
@@ -0,0 +1,63 @@
...
+            }
+        }
+        """
+        if len(d.keys()) != 1:
+            raise DagParseException(f"Invalid DAG name in {d}.")
+
+        name = list(d.keys())[0]
+
+        if "schedule_interval" not in d[name]:
+            raise DagParseException(f"schedule_interval missing in {d}.")
```
jklukas, line 54: It's looking like we have a good bit of validation code here. We might want to consider defining JSON schemas for the expected structure and rely on validating against the schema rather than writing our own validation code. We'd write the json schemas in yaml files like is done for Glean. We can write up a separate issue for that if you think it's worth pursuing.

bigquery_etl/query_scheduling/dags.py
```diff
@@ -0,0 +1,70 @@
+"""Represents a collection of configured Airflow DAGs."""
+
+from itertools import groupby
+import yaml
+
+from bigquery_etl.query_scheduling.dag import Dag
+
+
+class Dags:
```
jklukas, line 9: I'd prefer to call this something a bit more verbose, like `DagList` since it's difficult to tell the difference between `Dag` and `Dags` at a glance.

bigquery_etl/query_scheduling/dags.py
```diff
@@ -0,0 +1,70 @@
...
+
+        return None
+
+    def with_tasks(self, tasks):
+        """Assign tasks to their corresponding DAGs."""
+        for dag_name, tasks in groupby(tasks, lambda t: t.dag_name):
+            dag = self.dag_by_name(dag_name)
+
+            if dag is None:
+                print(
```
jklukas, line 58: Should we raise an exception instead? Or perhaps set a flag so that we can return a failure state later?

bigquery_etl/query_scheduling/generate_airflow_dags.py
```diff
@@ -0,0 +1,78 @@
...
+                query_file = os.path.join(root, QUERY_FILE)
+
+                try:
+                    task = Task.of_query(query_file)
+                    tasks.append(task)
+                except FileNotFoundError:
+                    # query has no metadata.yaml file; skip
+                    pass
+                except UnscheduledTask:
+                    logging.info(
```
jklukas, line 48: If we say "warning" in the message, this should be a `warn`-level message. Also, I would expect that the logging configuration would handle prefixing the output with "WARN".

jklukas: Thinking more about this, I think this should probably be silent about unscheduled tasks, at least for the early phases of this project. I expect it will take a while for us to transition queries to this framework, so initially most queries will not have scheduling metadata.

bigquery_etl/query_scheduling/dag.py
scholtzan: This might be a good idea, especially as the config files get more complex. I'll create a separate issue for this.

PR #922, opened 2020-04-21
Address Amplitude feedback on sync_send_tab_events

jklukas:
Amplitude folks indicated that `time` must always be populated and they cannot have duplicate `device_id` fields.


sql/telemetry_derived/sync_send_tab_events_v1/view.sql
```diff
@@ -46,8 +43,14 @@ SELECT
     [device_id, event_category, event_method, event_object, server_time, flow_id],
     '-'
   ) AS insert_id,
-  time,
-  TIMESTAMP_MILLIS(time) AS timestamp,
```
jklukas, line 50: We remove `timestamp` as it is redundant and Amplitude folks indicate `time` in epoch millis is the canonical format.

PR #923, opened 2020-04-21
Add docs for running integration tests

scholtzan:
Describes how to run integration tests.


tests/README.md
```diff
@@ -15,6 +15,11 @@ venv/bin/pytest --black --docstyle --flake8 --mypy-ignore-missing-imports -n 4
 
 # use -k to selectively run a set of tests that matches the expression `udf`
 venv/bin/pytest -k udf
+
+# run integration tests with 4 workers in parallel
+gcloud auth application-default login # or set GOOGLE_APPLICATION_CREDENTIALS
+export GOOGLE_PROJECT_ID="bigquery-etl-integration-test"
+venv/bin/pytest -m integration -n 4
```
jklukas, line 22:
```suggestion
./venv/bin/pytest -m integration -n 4
```

We should change all of these examples of running scripts by path to use the `./` prefix as it should be more generally applicable to different shells.

jklukas: Feel free to ignore this suggestion as what you have here fits with existing style.

scholtzan: I'll leave it as is for now to stay consistent with the rest of the docs.

PR #928, opened 2020-04-24
Fix search view reference

whd:
Ignored in dryrun due to being a search view, and failed on deploy like so:

    google.api_core.exceptions.BadRequest: 400 GET Unrecognized name: days_clicked_ad_bytes; Did you mean days_clicked_ads_bytes? at [14:5]

Side note: https://bugzilla.mozilla.org/show_bug.cgi?id=1623180 will likely open up access to this dataset at some point.


whd:
For posterity, when deploying this again I saw what looks like a novel transient timing dependency (which resolved on another redeploy):

   google.api_core.exceptions.BadRequest: 400 Invalid snapshot time 1587697166918 for table moz-fx-data-shared-prod:search.search_clients_last_seen_v1. Cannot read before 1587697166966

Those timestamps are within 50ms of each other.

PR #932, opened 2020-04-27
Lower-friction scheduled queries dependencies

scholtzan:
Adds support for determining task dependencies that are defined in bigquery-etl.

The next step will be determining dependencies that are defined in telemetry-airflow.

### DIFF
--- a/bigquery_etl/query_scheduling/task.py
+++ b/bigquery_etl/query_scheduling/task.py
@@ -1,6 +1,7 @@
 """Represents a scheduled Airflow task."""
 
 import re
+from google.cloud import bigquery
 
 from bigquery_etl.metadata.parse_metadata import Metadata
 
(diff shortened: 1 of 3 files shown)


tests/query_scheduling/test_task.py
```diff
@@ -87,3 +90,180 @@ def test_task_instantiation(self):
...
+
+        result = task.get_dependencies(bigquery_client, dags)
+
+        tables = [f"{t.dataset}__{t.table}__{t.version}" for t in result]
+
+        assert f"{temporary_dataset}__table1__v1" in tables
+        assert f"{temporary_dataset}__table2__v1" in tables
+
+    @pytest.mark.integration
+    def test_task_get_view_dependencies(
```
scholtzan, line 164: I thought there might be issues when a view is part of a query. But running the test case for this scenario shows  that the tables that are referenced in that view are returned. Which is what we want.

bigquery_etl/query_scheduling/task.py
```diff
@@ -75,12 +76,36 @@ def of_query(cls, query_file):
...
         return cls(query_file, metadata)
 
-    def get_dependencies(self):
+    def _get_referenced_tables(self, client):
+        """
+        Perform a dry_run to get tables the query depends on.
+
+        Queries that reference more than 50 tables will not have a complete list
+        of dependencies. See https://cloud.google.com/bigquery/docs/reference/
+        rest/v2/Job#JobStatistics2.FIELDS.referenced_tables
```
jklukas, line 85: Can we emit a warning if we see 50 tables in the output here?

bigquery_etl/query_scheduling/task.py
```diff
@@ -75,12 +76,36 @@ def of_query(cls, query_file):
         metadata = Metadata.of_sql_file(query_file)
         return cls(query_file, metadata)
 
-    def get_dependencies(self):
+    def _get_referenced_tables(self, client):
```
jklukas, line 79: We might want to memoize this using a `functools.cached_property`decorator. If this has to be invoked multiple times, it could slow things down.

PR #934, opened 2020-04-28
Create a query to export aggregated data for showcasing

Dexterp37:
the Italian outage in May 2020.

See [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1633681).


sql/internet_outages/italy_covid19_outage_v1/query.sql
```diff
@@ -0,0 +1,193 @@
...
+        COUNTIF(eUndefined > 0) as num_clients_eUndefined,
+        COUNTIF(eTimeOut > 0) as num_clients_eTimeOut,
+        COUNTIF(eAbort > 0) as num_clients_eAbort,
+        COUNTIF(eUnreachable > 0) as num_clients_eUnreachable,
+        COUNTIF(eTerminated > 0) as num_clients_eTerminated,
+        COUNTIF(eChannelOpen > 0) as num_clients_eChannelOpen,
+    FROM health_data_sample
+    GROUP BY date
+    HAVING
+      COUNT(*) > MIN_NUMBER_CLIENTS
```
Dexterp37, line 65: Note: this is not reporting days for which we have less than 5000 samples. Users of the "health" ping can report multiple errors per ping.

sql/internet_outages/italy_covid19_outage_v1/query.sql
```diff
@@ -0,0 +1,193 @@
...
+), final_health_data AS (
+    SELECT
+        h.date,
+        (num_clients_eUndefined / DAUs.client_count) as proportion_undefined,
```
Dexterp37, line 69: Another option to make aggregates even safer is to round any ratio < 0.01 (1% of the reference user-base) to 0.

sql/internet_outages/italy_covid19_outage_v1/metadata.yaml
```diff
@@ -0,0 +1,13 @@
+friendly_name: Italy's mid-pandemic internet outage (May 2020)
+description: >-
+  This dataset contains data from the Firefox "health"
+  and "main" pings, aggregated by day.
+owners:
+  - aplacitelli@mozilla.com
```
Dexterp37, line 6: Note to self: Saptarshi Guah should be owning this as well.

sql/internet_outages/italy_covid19_outage_v1/query.sql
Dexterp37: This didn't turn out to be required at data-review time.

sql/internet_outages/italy_covid19_outage_v1/query.sql
```diff
@@ -0,0 +1,193 @@
+DECLARE country_code STRING DEFAULT "IT";
+DECLARE fromdate DATE DEFAULT '2020-03-01'; -- '2019-12-01'
```
Dexterp37, line 2: Note: fixup the dates.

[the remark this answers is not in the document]
Dexterp37: Right, sorry for that!

sql/internet_outages/italy_covid19_outage_v1/query.sql
```diff
@@ -0,0 +1,193 @@
...
+    WHERE
+        submission_date >= fromdate
+        AND submission_date <= todate
+        AND country = country_code
+    GROUP BY 1
+),
+-- Compute aggregates for the health data.
+health_data_sample AS (
+    SELECT 
+        DATE(SAFE_CAST(creation_date as TIMESTAMP)) as date,
```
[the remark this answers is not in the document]
Dexterp37: The `submission_date` in the WHERE clause is required because of the partitioning of the dataset. It's also handy for us, since filtering by `submission_date` actually makes this dataset reproducible: if we only filtered by creation date, then we'd have to add some "slack time" since pings submitted later on might have a creation date way earlier than that. This is especially true for outages.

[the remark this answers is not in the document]
Dexterp37: Also, added fun: `telemetry.clients_daily` doesn't allow to filter by `creation_date`, which makes normalizing a bit weird, but generally ok.

sql/internet_outages/italy_covid19_outage_v1/metadata.yaml
```diff
@@ -0,0 +1,32 @@
...
+  As with any observational data, there are many caveats and interpretation must be done carefully. Below is a list of issues we have considered, but it is not exhaustive.
+    - Firefox users are not representative of the general population in their region.
+    - Users can experience multiple types of failures and so the proportions are not summable. For example, if 2.4% of clients had a timeout and 2.6% of clients had eUnreachable that doesn't necessarily mean that 5.0% of clients had a timeout or a eUnreachable
+    - Geographical data is based on IPGeo databases. These databases are imperfect, so some activity may be attributed to the wrong location. Further, proxy and VPN usage can create geo-attribution errors.
+owners:
+  - aplacitelli@mozilla.com
+  - sguha@mozilla.com
+labels:
+  incremental: true
+  schedule: daily
```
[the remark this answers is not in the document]
Dexterp37: No, I don't really want this to be scheduled. I left this here to raise it as a flag at review time (sorry, forgot to do so, but good thing you noticed :-D ). [Documentation](https://github.com/mozilla/bigquery-etl#query-metadata) is a bit unclear with respect to this: how do I say "I don't want this to be scheduled"?  Should I simply leave the field out?

sql/internet_outages/italy_covid19_outage_v1/query.sql
```diff
@@ -0,0 +1,193 @@
...
+    SELECT
+        submission_date as date,
+        COUNT(*) as client_count
+    FROM
+        telemetry.clients_daily
+    WHERE
+        submission_date >= fromdate
+        AND submission_date <= todate
+        AND country = country_code
+    GROUP BY 1
```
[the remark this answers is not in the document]
Dexterp37: For this specific dataset, doesn't change anything because we're always over the `MIN_NUMBER_CLIENTS`, but you're right, it's better to add it to be formally correct.

sql/internet_outages/italy_covid19_outage_v1/metadata.yaml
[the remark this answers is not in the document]
Dexterp37: Ok, I now see that @scholtzan hinted at `schedule: once` in #963 - is this something TBD or something that I can already put in here?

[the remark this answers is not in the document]
scholtzan: The query scheduling is TBD. You can remove the `schedule` label for now.

scholtzan:
I wonder if we want to create a separate dataset for this query, or if it should be part of `static`, `analysis` or maybe `telemetry_derived` instead.

sql/internet_outages/italy_covid19_outage_v1/query.sql
```diff
@@ -0,0 +1,192 @@
+DECLARE country_code STRING DEFAULT "IT";
```
jklukas, line 1: Given this query is to be run once, I'd argue it should be called `init.sql` rather than `query.sql` and it should explicitly create the destination table via a CREATE TABLE AS query.

BQ scripts are incompatible with destination tables, so this script has to handle putting data where we want it. We can't tell the job to insert the output of the final query into a destination.

jklukas:
> I wonder if we want to create a separate dataset for this query, or if it should be part of static, analysis or maybe telemetry_derived instead.

I don't think it's worth the extra hassle of provisioning a new dataset for this. I think `telemetry_derived` would be a reasonable home for it.

Dexterp37:
> 
> 
> > I wonder if we want to create a separate dataset for this query, or if it should be part of static, analysis or maybe telemetry_derived instead.
> 
> I don't think it's worth the extra hassle of provisioning a new dataset for this. I think `telemetry_derived` would be a reasonable home for it.

@jklukas / @scholtzan what does that mean for me? Is there anything I need to change here?

sql/internet_outages/italy_covid19_outage_v1/query.sql
```diff
@@ -0,0 +1,192 @@
...
+    FROM health_data_aggregates AS h
+    INNER JOIN DAUs ON DAUs.date = h.date
+),
+-- Compute aggregates for histograms coming from the health ping.
+histogram_data_sample AS (
+  SELECT
+    client_id,
+    document_id,
+    DATE(SAFE_CAST(creation_date as TIMESTAMP)) as time_slot,
+    payload.info.subsession_length  as subsession_length,
```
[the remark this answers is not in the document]
Dexterp37: Good catch!

sql/internet_outages/italy_covid19_outage_v1/query.sql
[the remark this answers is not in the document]
Dexterp37: In general, it would not help us: we're interested in when the outage was _detected_ (i.e. when the ping was created) versus when the outage was _reported_ (i.e. submission time). This would basically shift the "spikes" for certain classes of outages to dates in the future.

For this specific case it doesn't make much difference, but I'd like to keep this code as close as possible to what we use for other cases. Unless you have a strong opinion on this, of course!

jklukas:
> What does that mean for me? Is there anything I need to change here?

The destination dataset and table name are determined by the directory structure. Could you move `sql/internet_outages/italy_covid19_outage_v1` to `sql/telemetry_derived/italy_covid19_outage_v1`?

sql/internet_outages/italy_covid19_outage_v1/query.sql
Dexterp37: > 
> 
> > What does that mean for me? Is there anything I need to change here?
> 
> The destination dataset and table name are determined by the directory structure. Could you move `sql/internet_outages/italy_covid19_outage_v1` to `sql/telemetry_derived/italy_covid19_outage_v1`?

Done!

sql/telemetry_derived/italy_covid19_outage_v1/query.sql
```diff
@@ -0,0 +1,292 @@
+DECLARE country_code STRING DEFAULT "IT";
+
+DECLARE fromdate DATE DEFAULT '2020-01-01';
+
+DECLARE todate DATE DEFAULT '2020-03-31';
+
+-- The minimum size of a bucket for it to be published.
+DECLARE MIN_NUMBER_CLIENTS INT64 DEFAULT 5000;
```
jklukas, line 8: Having these DECLARE statements turns this from a simple query into a BQ script. Our tooling for publishing public data is not set up to handle that, since script jobs cannot be run with a destination table set.

I think it would be best to turn this into a simple query by removing these DECLAREs and instead inserting the literal values into the query, even though that means duplicating the literal values.

As BQ scripting matures, I hope we will be able to support this syntax in bigquery-etl in the future.

jklukas:
I haven't closely looked through the query logic, so I defer to other reviewers on that. But as long as we remove the DECLAREs, this looks good to me technically now 👍

sql/telemetry_derived/italy_covid19_outage_v1/query.sql
```diff
@@ -0,0 +1,284 @@
...
+  INNER JOIN
+    DAUs
+  ON
+    DAUs.date = h.date
+),
+-- Compute aggregates for histograms coming from the health ping.
+histogram_data_sample AS (
+  SELECT
+    client_id,
+    document_id,
```
mreid-moz, line 118: This field appears to be unused, please remove it

Dexterp37: Cheers, thanks for spotting! Done!

jklukas:
> Query looks good to me now. The CI failure appears unrelated to this, @scholtzan can you take a look?

I've hit the "rebase" button. These problems are due to changes in some underlying tables. Being in sync with master should get dry runs working.

PR #935, opened 2020-04-28
Add experiments to impression_stats_flat

jklukas:
cc @kirill-demtchouk

### DIFF
--- a/sql/activity_stream_bi/impression_stats_flat_v1/query.sql
+++ b/sql/activity_stream_bi/impression_stats_flat_v1/query.sql
@@ -20,6 +20,7 @@ SELECT
   user_prefs,
   release_channel,
   IFNULL(shield_id, 'n/a') AS shield_id,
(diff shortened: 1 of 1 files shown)


sql/activity_stream_bi/impression_stats_flat_v1/query.sql
```diff
@@ -20,6 +20,7 @@ SELECT
   user_prefs,
   release_channel,
   IFNULL(shield_id, 'n/a') AS shield_id,
+  ANY_VALUE(experiments),
```
jklukas, line 23: I'm assuming here that all the rows we get will have identical values for `experiments`, since we're aggregating across per-tile rows all associated with the same `impression_stats` ping.

jklukas:
Went ahead and merged based on an okay from @kirill-demtchouk. I also ran this query with field addition to create the new field.

Retroactive review would still be welcome.

PR #937, opened 2020-04-29
Limit metrics published in deviations query

scholtzan:
To ensure we are not accidentally publishing data, only approved `metric`s from `deviations` are published. If there are more metrics in the future, those can be added after there has been a data review.

cc @jmccrosky

### DIFF
--- a/sql/telemetry_derived/deviations_v1/query.sql
+++ b/sql/telemetry_derived/deviations_v1/query.sql
@@ -7,5 +7,7 @@ SELECT
 FROM
   `moz-fx-data-shared-prod.analysis.deviations`
 WHERE
+  -- We explicitly enumerate allowed metric types here so that we do not automatically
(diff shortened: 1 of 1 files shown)


sql/telemetry_derived/deviations_v1/query.sql
```diff
@@ -6,4 +6,6 @@ SELECT
   metric
 FROM
   `moz-fx-data-shared-prod.analysis.deviations`
-WHERE date = @submission_date
+WHERE
+  metric IN ('desktop_dau', 'mean_active_hours_per_client')
```
jklukas, line 10:
```suggestion
  -- We explicitly enumerate allowed metric types here so that we do not automatically
  -- publish new metric types publicly without review.
  metric IN ('desktop_dau', 'mean_active_hours_per_client')
```

Issue #939, opened 2020-04-29
CI step checking public data review bugs


scholtzan:
As a first step to make sure data is not accidentally published, we should add a CI step that checks a `review_bug` exists for queries that are `public_json` or `public_bigquery`.

At the same time we should also update the PR review template for reviewers and add a checklist item for them to check if a data review has been performed.

PR #940, opened 2020-04-29
Validate metadata

scholtzan:
Talking to Mark, we decided it would make sense to have CI check that public queries have a review bug associated in their `metadata.yaml` file. This serves as a first step to insure that data doesn't accidentally get published and also provides reviewers with a reference that they should check when reviewing new queries that are to be public.

https://github.com/mozilla/bigquery-etl/issues/939

cc @mreid-moz

### DIFF
--- a/bigquery_etl/metadata/validate_metadata.py
+++ b/bigquery_etl/metadata/validate_metadata.py
@@ -23,7 +23,7 @@ def validate_public_data(metadata, path):
     is_valid = True
 
     if metadata.is_public_bigquery() or metadata.is_public_json():
-        if metadata.review_bug() is None:
+        if not metadata.review_bug():
             logging.error(f"Missing review bug for public data: {path}")
             is_valid = False
 
@@ -49,7 +49,7 @@ def main():
                     path = os.path.join(root, *dirs, file)
                     metadata = Metadata.from_file(path)
 
-                    if validate_public_data(metadata, path) is False:
+                    if not validate_public_data(metadata, path):
                         failed = True
 
                     # todo more validation
--- a/bigquery_etl/metadata/validate_metadata.py
+++ b/bigquery_etl/metadata/validate_metadata.py
@@ -1,4 +1,4 @@
-"""Validate metadata files"""
+"""Validate metadata files."""
 
 from argparse import ArgumentParser
 import logging
@@ -6,7 +6,7 @@ import os
 import sys
 
 from .parse_metadata import Metadata
-from .util import standard_args
+from ..util import standard_args
 
 DEFAULT_DIR = "sql/"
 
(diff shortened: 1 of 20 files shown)


script/run_query
```diff
@@ -52,6 +52,10 @@ def main():
     try:
         metadata = Metadata.of_sql_file(query_file)
         if metadata.is_public_bigquery():
+            if metadata.review_bug() is None:
```
scholtzan, line 55: We should probably prevent publishing altogether if there is not review bug

bigquery_etl/public_data/publish_gcs_metadata.py
```diff
@@ -10,8 +10,8 @@
 from google.cloud import storage
 from itertools import groupby
 
-from ..parse_metadata import Metadata
-from ..util import standard_args
+from bigquery_etl.metadata.parse_metadata import Metadata
```
scholtzan, line 13: I thought it would make sense to move all the metadata scripts (since there are now 3 of them) to a new `metadata` directory to keep things cleaner.

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -0,0 +1,66 @@
...
+    failed = False
+
+    if os.path.isdir(args.target):
+        for root, dirs, files in os.walk(args.target):
+            for file in files:
+                if Metadata.is_metadata_file(file):
+                    path = os.path.join(root, *dirs, file)
+                    metadata = Metadata.from_file(path)
+
+                    if validate_public_data(metadata, path) is False:
```
jklukas, line 52:
```suggestion
                    if not validate_public_data(metadata, path):
```

Checking `is False` seems unidiomatic in Python.

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -0,0 +1,66 @@
...
+)
+standard_args.add_log_level(parser)
+
+
+def validate_public_data(metadata, path):
+    """Check if the metadata for public data queries is valid."""
+    is_valid = True
+
+    if metadata.is_public_bigquery() or metadata.is_public_json():
+        if metadata.review_bug() is None:
```
jklukas, line 26:
```suggestion
        if not metadata.review_bug():
```

With this change, we'll catch empty lists, etc. which should also be flagged as error states.

bigquery_etl/public_data/publish_json.py
```diff
@@ -302,6 +302,10 @@ def main():
     if not metadata.is_public_json():
         return
 
+    if metadata.review_bug() is None:
```
jklukas, line 305: Would it be possible to call `validate_metadata` here  instead of duplicating some of the validation logic? Or perhaps refactor `validate_metadata` a bit so that we have a common function to call? I want to make sure we perform the same set of checks when running `publish_json` as we do when we run `validate_metadata`.

PR #941, opened 2020-04-29
Bug 1632635 - FxA Amplitude export for active events

fbertsch:
Some more work to do, but this gives us a skeleton of what we'll be exporting to Amplitude.

Leif, you can take a look at the contents of this export:
```
SELECT *
FROM `moz-fx-data-shared-prod.analysis.frank_fxa_amplitude_export_v1_dev`
```

That table has 1000 rows, ordered by client.


sql/fxa_derived/fxa_amplitude_export_v1/query.sql
```diff
@@ -0,0 +1,100 @@
...
+    _TABLE_SUFFIX = FORMAT_DATE('%g%m%d', '2020-04-20')
+    AND jsonPayload.fields.event_type IN ( 'fxa_activity - cert_signed',
+      'fxa_activity - access_token_checked',
+      'fxa_activity - access_token_created' )
+    AND jsonPayload.fields.user_id IS NOT NULL ),
+  grouped_by_user AS (
+  SELECT
+    -- to prevent weirdness from timestamp field, use provided
+    -- submission date parameter as timestamp
+    CAST('2020-04-20' AS DATETIME) AS submission_timestamp,
```
fbertsch, line 17: I don't believe we care about the actual timestamp of the event, except for the day it occurs. Let me know if you can think of use cases otherwise.

sql/fxa_derived/fxa_amplitude_export_v1/query.sql
```diff
@@ -0,0 +1,100 @@
+WITH
+  base_events AS(
+  SELECT
+    *
+  FROM
+    `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_auth_20*`
+  WHERE
+    _TABLE_SUFFIX = FORMAT_DATE('%g%m%d', '2020-04-20')
+    AND jsonPayload.fields.event_type IN ( 'fxa_activity - cert_signed',
```
fbertsch, line 9: These are the 3 event types we were interested in reporting.

sql/fxa_derived/fxa_amplitude_export_v1/query.sql
```diff
@@ -0,0 +1,100 @@
...
+    `moz-fx-data-shared-prod`.udf.mode_last(ARRAY_AGG(jsonPayload.fields.LANGUAGE)) AS LANGUAGE,
+    -- Event properties, arrays
+    ARRAY_AGG(DISTINCT JSON_EXTRACT_SCALAR(jsonPayload.fields.event_properties,
+        "$.service") IGNORE NULLS) AS services,
+    ARRAY_AGG(DISTINCT JSON_EXTRACT_SCALAR(jsonPayload.fields.event_properties,
+        "$.oauth_client_id") IGNORE NULLS) AS oauth_client_ids,
+    -- User properties, arrays
+    ARRAY_AGG(DISTINCT JSON_EXTRACT_SCALAR(jsonPayload.fields.user_properties,
+        "$['$append'].fxa_services_used") IGNORE NULLS) AS fxa_services_used,
+    ARRAY_AGG(DISTINCT jsonPayload.fields.os_name IGNORE NULLS) AS os_used_day,
```
fbertsch, line 34: Leif, maybe we could get the FxA team to pull e.g. `os_used_week` and `os_used_month`

sql/fxa_derived/fxa_amplitude_export_v1/query.sql
```diff
@@ -0,0 +1,100 @@
...
+    MAX(JSON_EXTRACT_SCALAR(jsonPayload.fields.user_properties,
+        "$.sync_active_devices_day")) AS sync_active_devices_day,
+    MAX(JSON_EXTRACT_SCALAR(jsonPayload.fields.user_properties,
+        "$.sync_active_devices_week")) AS sync_active_devices_week,
+    MAX(JSON_EXTRACT_SCALAR(jsonPayload.fields.user_properties,
+        "$.sync_active_devices_month")) AS sync_active_devices_month,
+    `moz-fx-data-shared-prod`.udf.mode_last(ARRAY_AGG(JSON_EXTRACT_SCALAR(jsonPayload.fields.user_properties,
+          "$.ua_version") IGNORE NULLS)) AS ua_version,
+    `moz-fx-data-shared-prod`.udf.mode_last(ARRAY_AGG(JSON_EXTRACT_SCALAR(jsonPayload.fields.user_properties,
+          "$.ua_version") IGNORE NULLS)) AS us_browser,
```
irrationalagent, line 47: ua_browser

sql/fxa_derived/fxa_amplitude_export_v1/query.sql
irrationalagent: I'm a bit ignorant about the mechanics here, but if we use the same timestamp for all these rollup events then will we see a giant spike in event counts within amplitude at the same timestamp? I'm wondering if its worth using either mode_first or mode_last here so that the timestamp reflects the first or last event that we logged per user. I think I would lean towards mode_first. Would that cause you issues?

sql/fxa_derived/fxa_amplitude_export_v1/query.sql
irrationalagent: I think Alex has talked about wanting that in the past. I'm not sure we would block this on getting it though? I can ask him.

sql/fxa_derived/fxa_amplitude_export_v1/query.sql
```diff
@@ -0,0 +1,100 @@
...
+    AND jsonPayload.fields.event_type IN ( 'fxa_activity - cert_signed',
+      'fxa_activity - access_token_checked',
+      'fxa_activity - access_token_created' )
+    AND jsonPayload.fields.user_id IS NOT NULL ),
+  grouped_by_user AS (
+  SELECT
+    -- to prevent weirdness from timestamp field, use provided
+    -- submission date parameter as timestamp
+    CAST(@submission_date AS DATETIME) AS submission_timestamp,
+    -- TODO: JS udf with HMAC
```
jklukas, line 18: I had been assuming we'd need to create a custom python script to do the HMACing outside of BQ because there's no clear path to making the key available inside BQ.

Do you have a vision in mind for how this JS UDF would work and how we make the key available without spilling the key contents into query logs, etc?

sql/fxa_derived/fxa_amplitude_export_v1/query.sql
```diff
@@ -0,0 +1,100 @@
...
+    TO_HEX(SHA256(jsonPayload.fields.user_id)) AS user_id,
+    MIN(CONCAT(insertId, '-user')) AS insert_id,
+    CAST(@submission_date AS DATETIME) AS timestamp,
+    -- Amplitude properties, scalars
+    `moz-fx-data-shared-prod`.udf.mode_last(ARRAY_AGG(jsonPayload.fields.region)) AS region,
+    `moz-fx-data-shared-prod`.udf.mode_last(ARRAY_AGG(jsonPayload.fields.country)) AS country,
+    `moz-fx-data-shared-prod`.udf.mode_last(ARRAY_AGG(jsonPayload.fields.LANGUAGE)) AS LANGUAGE,
```
jklukas, line 25:
```suggestion
    `moz-fx-data-shared-prod`.udf.mode_last(ARRAY_AGG(jsonPayload.fields.`language`)) AS `language`,
```

sql/fxa_derived/fxa_amplitude_export_v1/query.sql
jklukas: It may be possible to make a small JS lib containing the HMAC key and store that in a GCS blob that it only accessible by the Airflow service account. We could then define a JS UDF that declares that lib in GCS as a dependency. I think that would avoid the actual content of the key ever appearing in logs, and it would allow only the Airflow user access to the secret.

I'm also not confident, though, that this will be able to scale. We have encountered various resource exceeded errors when attempting to use a JS function over a large number of rows.

fbertsch: I imagine it's simpler to create a locked-down dataset and edit [access control to those query logs](https://cloud.google.com/logging/docs/access-control). Some details of this may need to be worked out, but that would alleviate having to create big work-arounds for this key, and allow us to use secrets more easily in the future as well.

sql/fxa_derived/fxa_amplitude_export_v1/query.sql
fbertsch: Nope, but we could always add it later. Just a thought.

sql/fxa_derived/fxa_amplitude_export_v1/query.sql
fbertsch: Opened up https://bugzilla.mozilla.org/show_bug.cgi?id=1634179 to discuss this with ops.

udf/hmac_sha256.sql
```diff
@@ -0,0 +1,42 @@
+CREATE OR REPLACE FUNCTION
+  udf.hmac_sha256(key BYTES,
```
fbertsch, line 2: Default sql formatting is pretty awful here....

sql/fxa_derived/fxa_amplitude_export_v1/query.sql
fbertsch: So, it would cause a spike for that timestamp, but I don't think I've seen an Amplitude chart that is any more granular than day. We can do something like that, but I'd need to verify first that these `timestamp` field is guaranteed to be on that day.

udf/hmac_sha256.sql
```diff
@@ -0,0 +1,51 @@
+CREATE OR REPLACE FUNCTION udf.hmac_sha256(key BYTES, message BYTES) AS (
```
fbertsch, line 1: I have validated this against the `crypto` JS library they were using [here](https://github.com/mozilla/fxa-amplitude-send/blob/89bfaef20a1d978fce2dccddc0155699a17ac172/pubsub/utils.js#L12), just to ensure everything matches.

sql/fxa_derived/fxa_amplitude_export_v1/query.sql
irrationalagent: FxA does look at sub-day charts on occasion, sometimes to understand what happens when a new version of the server code rolls out. Of course, in this case they would still have to wait a day to see any changes, so maybe that's no longer a use case for these events. Not trying to make more work, unless I can think of a better reason not to do it or I hear differently from Alex I'm ok moving forward like this.

fbertsch: It really shouldn't be a problem. But this is a weird event so to mark it as a specific timestamp that is distributed across the day for clients seems off as well, since that's not _actually_ correct. Really up to you all.

irrationalagent: Agreed its weird, the user would have to understand for these events that the timestamp means "user was active on this time and potentially after, but we don't know". I asked alex if he has a use case but I suspect we can move forward with a generic timestamp.

udf/hmac_sha256.sql
jklukas: Well, this is a game changer. I did not expect we'd be able to implement this as SQL. Can you add a reference here for where this algorithm comes from?

fbertsch: Luckily `BYTES` support is pretty strong in BQ.

bigquery_etl/parse_udf.py
```diff
@@ -41,12 +41,31 @@ def from_file(filepath):
         with open(filepath) as f:
             text = f.read()
 
+        name = basename.replace(".sql", "")
```
fbertsch, line 44: @relud, might be good for you to review this section. I may have manged `parse_udf` in order to generate some SQL UDF tests.

bigquery_etl/parse_udf.py
```diff
@@ -41,12 +41,31 @@ def from_file(filepath):
...
+        try:
+            udf = RawUdf.from_text(text, dataset, name)
+        except ValueError as e:
+            raise ValueError(str(e) + f" in {filepath}")
+
+        udf.filepath = filepath
+        return udf
+
+    @staticmethod
+    def from_text(text, dataset, name, is_defined=True):
```
relud, line 56: this should specify `filepath` as an argument, rather than setting it afterward:
```suggestion
        try:
            return RawUdf.from_text(text, dataset, name, filepath)
        except ValueError as e:
            raise ValueError(str(e) + f" in {filepath}")

    @staticmethod
    def from_text(text, dataset, name, filepath=None, is_defined=True):
```

bigquery_etl/parse_udf.py
```diff
@@ -68,27 +87,28 @@ def from_file(filepath):
...
+                raise ValueError(
+                    f"Expected a UDF named {persistent_name} or {temp_name} "
+                    f"to be defined"
+                )
+            dependencies.remove(internal_name)
 
         return RawUdf(
             internal_name,
-            filepath,
+            None,
```
relud, line 111:
```suggestion
            filepath,
```

tests/validation/hmac_sha256.py
```diff
@@ -0,0 +1,47 @@
...
+        "TO_HEX(SUBSTR("
+        "udf.hmac_sha256("
+        "FROM_HEX('{Key}'),"
+        "FROM_HEX('{Msg}')),"
+        "1,"
+        "{Tlen})));"
+    )
+    test_sql_stmnts = [test_sql_fixture.format(**test_case) for test_case in test_cases]
+
+    return RawUdf.from_text("\n".join(test_sql_stmnts), "udf", "hmac_sha256", False)
```
relud, line 34:
```suggestion
    return RawUdf.from_text("\n".join(test_sql_stmnts), "udf", "hmac_sha256", is_defined=False)
```

sql/fxa_derived/fxa_amplitude_export_v1/query.sql
```diff
@@ -0,0 +1,138 @@
+WITH base_events
```
jklukas, line 1: The relevant key is now avaiable to Airflow in BQ with Jason's work in https://bugzilla.mozilla.org/show_bug.cgi?id=1634179#c15

```suggestion
WITH hmac_key AS (
  SELECT
    AEAD.DECRYPT_BYTES(
      (SELECT keyset FROM `moz-fx-dataops-secrets.airflow_query_keys.fxa_prod`),
      (SELECT ciphertext FROM `moz-fx-data-shared-prod.firefox_accounts_derived.encrypted_keys_v1`), key_id) AS value 
  FROM
    `moz-fx-data-shared-prod.firefox_accounts_derived.encrypted_keys_v1`
  WHERE
    key_id = 'fxa_hmac_prod'),
base_events
```

sql/fxa_derived/fxa_amplitude_export_v1/query.sql
```diff
@@ -0,0 +1,138 @@
...
+      )
+      AND jsonPayload.fields.user_id IS NOT NULL
+  ),
+  grouped_by_user AS (
+    SELECT
+    -- to prevent weirdness from timestamp field, use provided
+    -- submission date parameter as timestamp
+      CAST(@submission_date AS DATETIME) AS submission_timestamp,
+      TO_HEX(
+        udf.hmac_sha256(CAST(@fxa_hmac AS BYTES), CAST(jsonPayload.fields.user_id AS BYTES))
```
jklukas, line 23:
```suggestion
        udf.hmac_sha256((SELECT * FROM hmac_key), CAST(jsonPayload.fields.user_id AS BYTES))
```

fbertsch:
Still planning on adding tests. There are some edge cases to cover.

sql/fxa_derived/fxa_amplitude_export_v1/query.sql
```diff
@@ -0,0 +1,128 @@
...
+      ) AS oauth_client_ids,
+    -- User properties, arrays
+      ARRAY_AGG(
+        DISTINCT JSON_EXTRACT_SCALAR(
+          jsonPayload.fields.user_properties,
+          "$['$append'].fxa_services_used"
+        ) IGNORE NULLS
+      ) AS fxa_services_used,
+      ARRAY_AGG(DISTINCT jsonPayload.fields.os_name IGNORE NULLS) AS os_used_day,
+      CAST([] AS ARRAY<STRUCT<arr ARRAY<STRING>>>) AS os_used_week,
```
fbertsch: This change adds `os_used_week` and `os_used_month`

sql/fxa_derived/fxa_amplitude_export_v1/query.sql
```diff
@@ -0,0 +1,128 @@
...
+    FROM
+      fxa_derived.fxa_amplitude_export_v1
+    WHERE
+      DATE(submission_timestamp) = DATE_SUB(@submission_date, INTERVAL 1 DAY)
+      AND udf.shift_28_bits_one_day(days_seen_bits) > 0
+  )
+
+SELECT
+  CAST(@submission_date AS TIMESTAMP) AS submission_timestamp,
+  _current.* REPLACE (
```
fbertsch, line 116: Note that if there are no values for that user for that day, then all of the fields of `_current` will be empty. The view will filter those rows out, but we can continue to join with that data for future days.

tests/assert/nested_array_equals.sql
```diff
@@ -0,0 +1,23 @@
+/*
+This assumes a convention where a nested array is actually a STRUCT with an `arr` field.
```
jklukas, line 2: In generating BQ schemas from JSONSchema, we have a convention of transforming nested arrays where we create a struct with a field called `list` rather `arr`. I would like to stick with that convention and assume the field name is `list`.

See, for example, https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/transform/PubsubMessageToObjectNode.java#L312

sql/fxa_derived/fxa_amplitude_export_v1/query.sql
```diff
@@ -0,0 +1,138 @@
...
+      _previous.days_seen_bits,
+      _current.days_seen_bits
+    ) AS days_seen_bits,
+    udf.evicting_queue_append(
+      COALESCE(_previous.os_used_week, _current.os_used_week),
+      7,
+      STRUCT(COALESCE(_current.os_used_day, CAST([] AS ARRAY<STRING>)) AS arr)
+    ) AS os_used_week,
+    udf.evicting_queue_append(
+      COALESCE(_previous.os_used_month, _current.os_used_week),
```
jklukas, line 128:
```suggestion
      COALESCE(_previous.os_used_month, _current.os_used_month),
```

sql/fxa_derived/fxa_amplitude_export_v1/query.sql
jklukas: The os_used_week and os_used_month additions here greatly increase the complexity of the logic, and it's not clear to me what these mean. Can you add some code comments to explain the intention of maintaining these arrays?

jklukas: From what I can gather so far about these arrays of OS names, I think this would be better served not as nested arrays but instead using bit patterns. This would allow more reuse of existing UDFs and reduce the need for new abstractions. We would have a single field like:

    os_usage ARRAY<STRUCT<os STRING, days_used_bits INT64>>

Then, you would be able to derive the `os_used_week` and `os_used_month` arrays from that set of bit patterns.

fbertsch:
This now uses the bit functions for `os_used_day`, `os_used_week`, and `os_used_month`.

udf/active_values_from_days_seen_map.sql
```diff
@@ -0,0 +1,58 @@
...
+  start_offset INT64,
+  n_bits INT64
+) AS (
+  ARRAY(
+    SELECT
+      DISTINCT key
+    FROM
+      UNNEST(days_seen_bits_map)
+    WHERE
+        -- TODO: Use udf.bits28_active_in_range when it's available
```
fbertsch, line 20: Once we merge those proposed UDFs, we can replace this line.

sql/firefox_accounts_derived/fxa_amplitude_export_v1/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: FxA Amplitude Export
+description: >
+  Derived from FxA logs, this table contains active events and user property
+  updates for FxA users. This table is export to Amplitude.
```
jklukas, line 4:
```suggestion
  updates for FxA users. A view of this table is exported to Amplitude.
```

sql/firefox_accounts/fxa_amplitude_export_v1/view.sql
```diff
@@ -0,0 +1,125 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.firefox_accounts.fxa_amplitude_export_v1`
```
jklukas, line 2: Either this view should be "hidden" underneath `firefox_accounts_derived` or it should not have a version suffix.

jklukas:
r+wc

PR #942, opened 2020-04-29
[Bug 1632857] Set content-type for public data to json and encoding to gzip

scholtzan:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1632857

cc @wlach

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,13 +209,13 @@ labels:
   `metadata.yaml`
   - Data will be accessible under https://public-data.telemetry.mozilla.org
     - A list of all available datasets is published under https://public-data.telemetry.mozilla.org/all-datasets.json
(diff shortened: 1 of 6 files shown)


scholtzan:
@fbertsch Would it make sense to remove the `.gz` from the published json files or should we keep that?

wlach:
> @fbertsch Would it make sense to remove the `.gz` from the published json files or should we keep that?
> 
> If files have a `.json` ending then they will be displayed in the browser, otherwise there'd be a download prompt. This might cause issues with very large files though.

Browsers can handle pretty large JSON files these days, I think even 1GB should not be a big problem.

scholtzan:
I tried to open the ssl_ratios dataset which is relatively small, in Firefox it takes a while to load but seems fine. If I open it in Chrome it show the "Site unresponsive" popup after a while and renders only the first couple 100 lines.

scholtzan:
I'll ask Jesse to update the links in his blog post and will keep the existing `json.gz` files around until this happens to prevent links from breaking.

PR #946, opened 2020-05-01
Bug 1632245 - Add new fenix apps to mobile search

Ben-Wu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1632245

I added a script and template file to create a union the different fenix apps so that if we need to add a field we can just add it once in the template.  We can't `UNION ALL` a `SELECT * FROM` each app because they have different schemas.

Here's a query checking if the new results match the existing results https://sql.telemetry.mozilla.org/queries/70614/source
Mostly match but a few counts are +/- by a few that I would guess is either from rows being added after the original etl or shredder.


sql/search_derived/mobile_search_clients_daily_v1/mobile_search_clients_daily.sql.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
+"""
+Generate mobile search clients_daily query by creating a combined CTE for
+metrics and baseline for all Firefox Android apps, then print query to stdout
+"""
+import os
+
+APP_CHANNEL_TUPLES = [
```
fbertsch, line 8: https://github.com/mozilla/bigquery-etl/issues/947 may be relevant here.

tests/search_derived/mobile_search_clients_daily_v1/test_overactive_filter/org_mozilla_firefox.metrics.yaml
```diff
@@ -0,0 +1,2 @@
+---
```
fbertsch, line 1: What's going on with these files?

sql/search_derived/mobile_search_clients_daily_v1/mobile_search_clients_daily.sql.py
Ben-Wu: That udf does look like it would simplify things here.  I'll watch for that change

tests/search_derived/mobile_search_clients_daily_v1/test_overactive_filter/org_mozilla_firefox.metrics.yaml
Ben-Wu: These are just empty arrays so that the test script will create empty tables.  Otherwise it will error with dataset/table not found

Ben-Wu:
I do agree about the multiple ways for query generation.  I didn't know we were already using jinja although in this case it works the same as jinja templating except jinja would be a little more explicit.

fbertsch:
Let's tackle it later with https://github.com/mozilla/bigquery-etl/issues/953

sql/search_derived/mobile_search_clients_daily_v1/mobile_search_clients_daily.sql.py
fbertsch: Feel free to create the UDF!

PR #948, opened 2020-05-01
Bit pattern UDFs (bits28_*)

jklukas:
This is the implementation of the new interface documented in https://github.com/mozilla/firefox-data-docs/pull/462

These two PRs are co-evolving drafts.

### DIFF
--- a/udf/bits28_from_string.sql
+++ b/udf/bits28_from_string.sql
@@ -1,0 +1,31 @@
+/*
+
+Convert a string representing individual bits into an INT64.
+
+Implementation based on https://stackoverflow.com/a/51600210/1260237
+
+See detailed docs for the bits28 suite of functions:
+https://docs.telemetry.mozilla.org/cookbooks/clients_last_seen_bits.html#udf-reference
+
+*/
+CREATE OR REPLACE FUNCTION udf.bits28_from_string(s STRING) AS (
+  IF(
+    REGEXP_CONTAINS(s, r"^[01]{1,28}$"),
+    (
+      SELECT
+        SUM(CAST(c AS INT64) << (LENGTH(s) - 1 - bit))
+      FROM
+        UNNEST(SPLIT(s, '')) AS c
+        WITH OFFSET bit
+    ),
+    ERROR(FORMAT("bits28_from_string expects a string of up to 28 0's and 1's but got: %s", s))
+  )
+);
+
+-- Tests
+SELECT
+  assert_equals(1, udf.bits28_from_string('1')),
+  assert_equals(1, udf.bits28_from_string('01')),
+  assert_equals(1, udf.bits28_from_string('0000000000000000000000000001')),
+  assert_equals(2, udf.bits28_from_string('10')),
+  assert_equals(5, udf.bits28_from_string('101'));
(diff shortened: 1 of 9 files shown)


udf/bits28_range.sql
```diff
@@ -0,0 +1,40 @@
+/*
+
+Return an INT64 representing a range of bits from a source bit pattern.
+
+See detailed docs for the bits28 suite of functions:
+https://docs.telemetry.mozilla.org/cookbooks/normandy_events.html#udf-reference
```
fbertsch, line 6: This function might be especially confusing to people. Offhand it's not clear: is this "just the activity, starting at start_offset, and extending n_bits", or is it "what bits$n_bits would be, start at start_offset"? If that isn't clear, it's the difference between not shifting the results to the end, and shifting them (as we actually do below).

fbertsch: For this function, it might be nice if we can link to documentation about it explicitly, not just a general reference to the bits28 cookbook.

udf/bits28_retention.sql
```diff
@@ -0,0 +1,58 @@
+/*
+
+
+Return a nested struct providing numerator and denominator fields for
```
fbertsch, line 4: Is this providing a numerator and denominator? It may be better to specify this as e.g. `Return a nested struct providing activity information as a boolean for the standard ...`.

udf/bits28_to_dates.sql
```diff
@@ -0,0 +1,21 @@
+/*
+
+Convert a bit pattern into an array of the dates is represents.
+
+See detailed docs for the bits28 suite of functions:
+https://docs.telemetry.mozilla.org/cookbooks/normandy_events.html#udf-reference
+
+*/
+CREATE OR REPLACE FUNCTION udf.bits28_to_dates(bits INT64, end_date DATE)
```
fbertsch, line 9: s/`end_date`/`submission_date`? It could be _not_ `submission_date` if the user shifted the bits, but that seems rare.

udf/pos_of_trailing_set_bit.sql
```diff
@@ -1,5 +1,7 @@
 /*
 
+Identical to bits28_days_since_seen.
```
fbertsch, line 3: Ah, it is noted! Still seems odd to have two identical functions. Maybe in the next few weeks we can deprecate this one.

udf/bits28_from_string.sql
```diff
@@ -0,0 +1,19 @@
...
+
+Convert a string representing individual bits into an INT64.
+
+Implementation based on https://stackoverflow.com/a/51600210/1260237
+
+See detailed docs for the bits28 suite of functions:
+https://docs.telemetry.mozilla.org/cookbooks/normandy_events.html#udf-reference
+
+*/
+CREATE OR REPLACE FUNCTION udf.bits28_from_string(s STRING) AS (
```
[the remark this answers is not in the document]
jklukas: Added check.

fbertsch:
@jklukas we can also update the `active_values_from_days_seen_map` to use the bits28 function

jklukas:
> we can also update the active_values_from_days_seen_map to use the bits28 function

Added.

PR #949, opened 2020-05-01
[Bug 1633754] Add queries for anomaly detection pipeline

scholtzan:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1633754

The anomaly detection pipeline in https://github.com/mozilla/forecasting/pull/27 depends on these tables.

I'll have to do a backfill once this gets approved since the anomaly detection pipeline depends on all the past data.

cc @jmccrosky

### DIFF
--- a/sql/telemetry_derived/clients_daily_active_hours_by_country_aggregates_v1/query.sql
+++ b/sql/telemetry_derived/clients_daily_active_hours_by_country_aggregates_v1/query.sql
@@ -1,0 +1,42 @@
+WITH geo_t AS (
+  SELECT
+    client_id,
+    geo
+  FROM
+    (
+      SELECT
+        client_id,
+        country AS geo,
+        ROW_NUMBER() OVER (PARTITION BY client_id ORDER BY submission_date DESC) AS rn
+      FROM
+        `moz-fx-data-shared-prod.telemetry.clients_daily` AS cd_t
+      WHERE
+        submission_date = @submission_date
+    )
+  WHERE
+    rn = 1
+)
+SELECT
+  submission_date AS date,
+  AVG(active_hours_sum) AS value,
+  COUNT(client_id) AS dau,
+  geo_t.geo AS geo
+FROM
+  (
+    SELECT
+      active_hours_sum,
+      client_id,
+      submission_date,
+    FROM
+      `moz-fx-data-shared-prod.telemetry.clients_daily`
+    WHERE
+      submission_date = @submission_date
+      AND attribution.content IS NOT NULL
+  )
+INNER JOIN
+  geo_t
+USING
+  (client_id)
+GROUP BY
+  geo,
+  submission_date
--- a/sql/telemetry_derived/clients_daily_active_hours_by_top_city_aggregates_v1/metadata.yaml
+++ b/sql/telemetry_derived/clients_daily_active_hours_by_top_city_aggregates_v1/metadata.yaml
@@ -1,0 +1,10 @@
+friendly_name: Clients daily active hours by top cities aggregates
+description: >-
+  Mean active hours of daily active clients by city, partitioned by day and
+  restricted to the top 1000 cities according to DAU on an arbitrary day.
+owners:
+  - jmccrosky@mozilla.com
+  - ascholtz@mozilla.com
+labels:
+  incremental: true
+  schedule: daily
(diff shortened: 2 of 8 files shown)


sql/telemetry_derived/clients_daily_active_hours_by_top_city_aggregates_v1/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: Clients daily active hours by top cities aggregates
```
scholtzan, line 1: Those query names are getting pretty long :/

sql/telemetry_derived/clients_daily_active_hours_by_country_aggregates_v1/query.sql
```diff
@@ -0,0 +1,42 @@
+WITH geo_t AS (
+  SELECT
+    client_id,
+    geo
+  FROM
+    (
+      SELECT
+        client_id,
+        country AS geo,
+        ROW_NUMBER() OVER (PARTITION BY client_id ORDER BY submission_date DESC) AS rn
```
jmccrosky, line 10: I don't entirely understand the machinery here, but I think this will cause a problem.  The geo for a profile in the pipeline needs to be fixed over time, as updated to the IPGeo database cause users to "teleport" between locations sometimes, introducing major anomalies at times.

Given that we're considering moving to a "clients_first_seen" notion of profile creation, it might be worth basing this on a such a table that would capture the basic dimensions (including geo) of each profiles first (ever) ping.

sql/telemetry_derived/clients_daily_active_hours_by_top_city_aggregates_v1/query.sql
```diff
@@ -0,0 +1,65 @@
+WITH top_cities_t AS (
+  SELECT
+    COUNT(client_id) AS dau,
+    CONCAT(country, ":", geo_subdivision1, ":", geo_subdivision2, ":", city) AS geo
+  FROM
+    `moz-fx-data-shared-prod.telemetry.clients_daily`
+  WHERE
+    submission_date = "2020-03-01"
```
jmccrosky, line 8: Just flagging that this does need to be some fixed date to keep the set of cities constant.  No action needed though, unless you want to add a comment.

sql/telemetry_derived/clients_daily_active_hours_by_country_aggregates_v1/query.sql
jklukas: I'm tempted to use this as motivation to go ahead and create a clients_first_seen table. We can probably make that happen this week if it seems like the right thing to do for this pipeline.;

scholtzan: What about `moz-fx-data-shared-prod.telemetry_derived.clients_first_seen_v6`?

sql/telemetry_derived/clients_daily_active_hours_by_top_city_aggregates_v1/metadata.yaml
jklukas: Unless we expect these to have direct application outside of this anomaly detection pipeline, we could name these all with an `anomdtct_` prefix to match the library name. So this could be maybe `anmdtct_top_city_aggregates_v1`.

scholtzan: These tables won't be necessary if the models get cached in BigQuery and then used for doing predictions on new data. Which is probably the most efficient way to go.

sql/telemetry_derived/clients_daily_active_hours_by_country_aggregates_v1/query.sql
jklukas: I just created that today 😆 

I was testing out a possible design for the query to serve this need. I'll try to turn that into a PR right now.

jklukas: See https://github.com/mozilla/bigquery-etl/pull/952

scholtzan:
This is not needed anymore as we'll use `clients_first_seen` instead

PR #952, opened 2020-05-04
clients_first_seen

jklukas:
Draft of a new table that may be useful in
https://github.com/mozilla/bigquery-etl/pull/949

If we have no immediate use, we can table this for now.

### DIFF
--- a/sql/telemetry/clients_first_seen/metadata.yaml
+++ b/sql/telemetry/clients_first_seen/metadata.yaml
@@ -1,0 +1,9 @@
+friendly_name: Clients first seen
+description: >
+  Identical in schema to clients_daily except that each client appears only
+  once over all time; includes fields first_seen_date and second_seen_date
+  for determining user "activation"
+owners:
+  - jklukas@mozilla.com
+labels:
+  application: firefox
--- a/sql/telemetry/clients_first_seen/metadata.yaml
+++ b/sql/telemetry/clients_first_seen/metadata.yaml
@@ -1,5 +1,5 @@
 friendly_name: Clients first seen
-description: >
+description: >-
   Identical in schema to clients_daily except that each client appears only
   once over all time; includes fields first_seen_date and second_seen_date
   for determining user "activation"
--- a/sql/telemetry_derived/clients_first_seen_v1/init.sql
+++ b/sql/telemetry_derived/clients_first_seen_v1/init.sql
@@ -1,27 +1,7 @@
-CREATE TEMPORARY TABLE clients_first_seen_dates
-PARTITION BY
-  first_seen_date
-AS
-WITH base AS (
-  SELECT
-    client_id,
-    ARRAY_AGG(submission_date) AS dates_seen,
-  FROM
-    clients_daily_v6
-  WHERE
-    submission_date >= DATE('2010-01-01')
-  GROUP BY
-    client_id
-)
-SELECT
-  client_id,
-  IF(ARRAY_LENGTH(dates_seen) > 0, dates_seen[OFFSET(0)], NULL) AS first_seen_date,
-  IF(ARRAY_LENGTH(dates_seen) > 1, dates_seen[OFFSET(1)], NULL) AS second_seen_date,
-FROM
-  base;
+DECLARE sample_id_iter INT64 DEFAULT 0;
 
-CREATE TABLE IF NOT EXISTS
-  clients_first_seen_v1
+CREATE OR REPLACE TABLE
+  telemetry_derived.clients_first_seen_v1
 PARTITION BY
   first_seen_date
 CLUSTER BY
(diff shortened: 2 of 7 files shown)


sql/telemetry_derived/clients_first_seen_v1/init.sql
```diff
@@ -0,0 +1,43 @@
...
+PARTITION BY
+  first_seen_date
+CLUSTER BY
+  normalized_channel,
+  sample_id
+AS
+SELECT
+  cfs.first_seen_date,
+  cfs.second_seen_date,
+  cd.* EXCEPT (submission_date)
```
jklukas, line 34: This captures the entire `clients_daily` row for the first time a client is seen.

sql/telemetry_derived/clients_first_seen_v1/query.sql
```diff
@@ -0,0 +1,25 @@
+WITH today AS (
```
jklukas, line 1: If we productionize this, I would add some information at the top here.

This query is weird in that we'd be rewriting the entire table every day rather than a single partition. Most rows will stay the same, but some will get second_seen_date filled in.

I believe the full size of this table will only be ~2 TB, so rewriting that amount of data every day seems reasonable. The `init.sql` query will have to scan the full history of clients_daily, so several hundred TB. It's expensive in dollars, but it appears to run quite quickly. It took ~1 minute when run for a single sample_id.

scholtzan:
I'm wondering if it makes sense to add additional fields, like geo information (city, country) to `clients_first_seen`. The queries in #949 and the related pipeline need to determine the geo information when the client was seen first. Only having the first-seen date would still require an expensive query to determine the geo information from `clients_daily`.

jklukas:
> I'm wondering if it makes sense to add additional fields, like geo information (city, country) to `clients_first_seen`

The design here does capture geo info as Jesse requested. It contains all clients_daily fields for the first observation of the client.

sql/telemetry_derived/clients_first_seen_v1/init.sql
scholtzan: Oh cool, I missed that 👍

sql/telemetry_derived/clients_first_seen_v1/query.sql
```diff
@@ -0,0 +1,25 @@
...
+  SELECT
+    NULL AS first_seen_date,
+    NULL AS second_seen_date,
+    * EXCEPT (submission_date)
+  FROM
+    clients_daily_v1
```
scholtzan, line 7: Shouldn't this be v6?

```suggestion
    clients_daily_v6
```

scholtzan:
This would be super useful to have for https://github.com/mozilla/forecasting/pull/27

jmccrosky:
This is super cool :)  I don't understand why it's necessary to read all of clients_daily each day though.  In theory, don't we just need to read the most recent day each time?  I can imagine joining the most recent day of clients_daily to the entire clients_first_seen and producing the new clients_first_seen from that?

jmccrosky:
Or wait, that's what you're doing, isn't it.  You just mean we have to read all of clients_daily once!  Cool :)

sql/telemetry_derived/clients_first_seen_v1/query.sql
jklukas: I think v6 was a poor initial choice on my part. Although clients_daily is v6, this will be the first version of clients_first_seen. Similarly, clients_last_seen is v1.

jklukas:
I am currently running the `init.sql` query to create the table. Dry runs will fail until that completes, but otherwise this is ready for review.

jklukas:
The `init.sql` failed with "Quota exceeded: Your project exceeded quota for total shuffle size limit". I'm now trying this as a script that iterates through sample_ids.

sql/telemetry/clients_first_seen/metadata.yaml
```diff
@@ -0,0 +1,9 @@
+friendly_name: Clients first seen
+description: >
+  Identical in schema to clients_daily except that each client appears only
+  once over all time; includes fields first_seen_date and second_seen_date
+  for determining user "activation"
+owners:
+  - jklukas@mozilla.com
+labels:
+  application: firefox
```
scholtzan, line 9: Should we also add?

```
  incremental: true
  schedule: daily
```

sql/telemetry_derived/clients_first_seen_v1/init.sql
```diff
@@ -0,0 +1,43 @@
+CREATE TEMPORARY TABLE clients_first_seen_dates
+PARTITION BY
+  first_seen_date
+AS
+WITH base AS (
+  SELECT
+    client_id,
+    ARRAY_AGG(submission_date) AS dates_seen,
+  FROM
+    clients_daily_v6
```
scholtzan: It looks like for most queries in this repository the dataset is usually specified. So maybe `telemetry_derived.clients_daily_v6`?

sql/telemetry_derived/clients_first_seen_v1/query.sql
scholtzan: Right but in this case we'd need to query `clients_daily_v6`? `clients_daily_v1` doesn't exist.

sql/telemetry_derived/clients_first_seen_v1/init.sql
scholtzan: Looks like it's been fixed.

sql/telemetry_derived/clients_first_seen_v1/query.sql
```diff
@@ -0,0 +1,64 @@
...
+    second_seen_date
+  )
+FROM
+  clients_first_seen_v1 AS cfs
+FULL JOIN
+  today
+USING
+  (client_id)
+WHERE
+  cd.submission_date = @submission_date
```
scholtzan, line 64: This `WHERE` can be removed, this should already be accounted for in `today`

jklukas:
`init.sql` has been running for an hour and has populated up through sample_id 36. So should take about 2 more hours for this to be fully available.

PR #956, opened 2020-05-05
Bug 1635529 - Histogram UDFs

fbertsch:


fbertsch:
Here's an example giving a histogram of client-buckets for a given percentile: https://sql.telemetry.mozilla.org/queries/70811/source#178135

emtwo:
Glam needs to compute histogram percentiles for both desktop and glean data. Currently they both use this: https://github.com/mozilla/bigquery-etl/blob/master/udf_js/glean_percentile.sql

I am wondering if there is overlap here and whether we should unify on one way to compute percentiles everywhere.

wlach:
@emtwo thanks for commenting! After chatting with @acmiyaguchi and @fbertsch earlier this week, I think this is indeed exactly what we should do. One thing that we also need, however, is some kind of normalized histogram aggregation step in addition to something that calculates the  percentiles. Do we have one of those already?

emtwo:
It depends how you want to normalize. In glam desktop we have this: https://github.com/mozilla/bigquery-etl/blob/master/sql/telemetry_derived/clients_histogram_probe_counts_v1/query.sql#L1

Each client's histogram is normalized to sum to 1. Note there is also a piece in that udf that adjusts for sampling that we do on users and that would need to be taken out if we wanted to reuse this function.

fbertsch:
@wlach @emtwo I've added a simpler normalization function. I'm not sure we need to necessarily use these same functions in GLAM, since these functions are all used on the histogram struct, which GLAM may not be passing around (for performance reasons). These are optimized for ease of use.

fbertsch:
I've got an example that replicates GLAM using these functions: https://sql.telemetry.mozilla.org/queries/71127/source#178778

fbertsch:
Note that uses `submission_date` rather than `build_id`. We'll have to change that.

wlach:
Oh, also, the tests seem to failing.

wlach:
> The linked redash query (71127) isn't necessarily replicating GLAM in terms of technique, as far as I know. The query merges all of the pings to compute a density per build id, instead of aggregating per client then per build id.

I think you might be misreading the query (I found this confusing at first too). I think it does in fact replicate GLAM's behaviour as you and @emtwo have described it. The key part is the per_client_day subquery, around line 92 which groups by `build_id` and `client_id`. The normalization step then will be applied to all unique build_id/client_id combinations, ensuring that things work as you describe. We confirmed that this gives the same results as GLAM in [this slack conversation](https://mozilla.slack.com/archives/GE83ZMSAW/p1589915345452900).

acmiyaguchi:
I see where I missed it, it's grouping by client id but not selecting it.

PR #957, opened 2020-05-05
Update glam extract query to transpose percentiles and histograms

robhudson:


sql/telemetry_derived/glam_client_probe_counts_extract_v1/query.sql
```diff
@@ -1,49 +1,37 @@
...
-        AND total_users > 1000
+CREATE TEMP FUNCTION udf_js_flatten(
+    histogram ARRAY<STRUCT<key STRING, value FLOAT64>>
 )
+RETURNS STRING
+LANGUAGE js
+AS
+'''
+    let data = histogram.map(function(r) {
+        var obj = {};
```
emtwo, line 9: nit: `let obj = {}`

sql/telemetry_derived/glam_client_probe_counts_extract_v1/query.sql
```diff
@@ -1,49 +1,37 @@
...
 FROM
-    deduped
+    `moz-fx-data-shared-prod.telemetry.client_probe_counts`
 WHERE
-    rank = 1
+    channel = @channel
+    AND app_version IS NOT NULL
+    AND total_users > 1000
+GROUP BY
+    channel, app_version, app_build_id, os, metric, metric_type, key, process, client_agg_type, total_users
```
emtwo, line 37: nit: newline

sql/telemetry_derived/glam_client_probe_counts_extract_v1/query.sql
```diff
@@ -1,49 +1,37 @@
...
+    process,
     metric,
     -- BigQuery has some null unicode characters which Postgresql doesn't like, so we remove those here.
     -- Also limit string length to 200 to match column length.
     SUBSTR(REPLACE(key, r"\x00", ""), 0, 200) AS key,
     client_agg_type,
     metric_type,
     total_users,
-    TO_JSON_STRING(aggregates) AS aggregates
+    MAX(IF(agg_type="histogram", udf_js_flatten(aggregates), NULL)) AS histogram,
```
emtwo, line 28: I assume the `MAX()` is so we return the non-null value which is the correct value for `histogram`? Maybe just a comment so it's clear.

robhudson: A comment would be good, I agree. I'm using this as if it's a `COALESCE` but b/c of the `GROUP BY` I need an aggregation function to include this column here. I'll make a note of it.

PR #959, opened 2020-05-07
Filter out invalid build ids for glam (fixes #876)

### LINKED ISSUE #876: Filter glam desktop by valid build id's using the build id dataset in BQ

Ben-Wu:
Setting histogram type changes as base since I plan to merge that first and then backfill with both updated histogram type and filtered build id.


emtwo:
Mind adding a gist with your query to update `clients_scalar_aggregates_v1` and `clients_histogram_aggregates_v1` filtering out build_ids like you did on the histogram type PR?

Ben-Wu:
I updated the existing gist with the query for scalar aggregates and adding the build id filter to the histogram query https://gist.github.com/Ben-Wu/1479738680104d61b09fc0bb73d6ed5b
You can look at the latest revision

I validated the histogram query on a 1% sample that it does what it's supposed to do.  I'll validate again against the backup when it's done.

PR #960, opened 2020-05-07
Bug 1636231 - Remove background events from ios Amplitude export

fbertsch:
The formatting caused a big change, so I split that into a separate commit. There are no changes in that commit other than reformatting.


fbertsch:
Thanks @jklukas!

PR #961, opened 2020-05-07
Update deviations query for Airflow anomdtct pipeline usage

scholtzan:
The anomaly detection pipeline scheduled via Airflow in https://github.com/mozilla/telemetry-airflow/pull/977 writes data to `telemetry_derived.deviations_v1`. In order for the data to get published we need to run this query. What is a bit unusual is that it will read the partition of the date to be processed and replace it at the same time.

### DIFF
--- a/script/dryrun
+++ b/script/dryrun
@@ -45,7 +45,6 @@ SKIP = {
     "sql/search/search_rfm/view.sql",
     "sql/search/search_clients_last_seen_v1/view.sql",
     "sql/search/search_clients_last_seen/view.sql",
-    "sql/telemetry_derived/deviations_v1/query.sql",
     "sql/firefox_accounts_derived/fxa_amplitude_export_v1/query.sql",
     # Already exists (and lacks an "OR REPLACE" clause)
     "sql/org_mozilla_firefox_derived/clients_first_seen_v1/init.sql",
--- a/sql/telemetry_derived/deviations_v1/query.sql
+++ b/sql/telemetry_derived/deviations_v1/query.sql
@@ -5,7 +5,7 @@ SELECT
   geography,
   metric
 FROM
-  `moz-fx-data-shared-prod.analysis.deviations`
+  `moz-fx-data-shared-prod.telemetry_derived.deviations_v1`
 WHERE
   -- We explicitly enumerate allowed metric types here so that we do not automatically
   -- publish new metric types publicly without review.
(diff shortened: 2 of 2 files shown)


sql/telemetry_derived/deviations_v1/query.sql
```diff
@@ -5,7 +5,7 @@ SELECT
   geography,
   metric
 FROM
-  `moz-fx-data-shared-prod.analysis.deviations`
+  `moz-fx-data-shared-prod.telemetry_derived.deviations_v1`
```
jklukas, line 8: This query is referencing the destination table. Do we need to add a new table name? What's the flow now?

scholtzan: The flow now is `clients_first_seen` gets populated -> the anomaly detection pipeline get triggered and forecasts data for a specific date -> data gets written to `deviations_v1` -> this query gets executed to publish data to public BigQuery and JSON endpoint

scholtzan: I guess it would be possible to let the anomdtct pipeline write results to a different table as it was before. But we'd end up having 2 tables with exactly the same data. Running this query with replacing the data worked for me when I tried it locally.

jklukas, line 8:
```suggestion
  deviations_v1
```

Per style guidelines, we should not fully qualify the table name here.

jklukas: This is making sense to me now. But it's still a very weird situation. This query file should include commentary to indicate this table is populated from an external job, and then this query exists only to ensure we aren't accidentally publishing data we didn't intend.

jklukas: Actually, maybe this doesn't make sense. The query puts the data in the public project. Don't we have an assumption that the pipeline will create a view named `moz-fx-data-shared-prod.telemetry_derived.deviations_v1` that points to the public table?

scholtzan: Oh right, so I guess we'll want to let the anomdtct pipeline write data into a different table?

jklukas: Yes, I think we'll need to provision a different table. Perhaps `deviations_anomdtct_v1`.

script/dryrun
```diff
@@ -45,7 +45,6 @@ SKIP = {
     "sql/search/search_rfm/view.sql",
     "sql/search/search_clients_last_seen_v1/view.sql",
     "sql/search/search_clients_last_seen/view.sql",
-    "sql/telemetry_derived/deviations_v1/query.sql",
```
scholtzan, line 48: The dry_run now also works

PR #970, opened 2020-05-08
Update json_extract_missing_cols to get missing columns from arrays

scholtzan:
The missing columns dashboard doesn't show missing columns in arrays. I updated the missing_cols UDF to also extract those columns. Key paths with arrays would be represented like 

`first`.`array`.[...].`nested-array-element`.`nested` 

with [...] indicating that the element is an array.

### DIFF
--- a/udf_js/json_extract_missing_cols.sql
+++ b/udf_js/json_extract_missing_cols.sql
@@ -86,11 +86,11 @@ WITH
           "other-second": "value",
           "array": [
             {
-              "array-element": "value",
-              "another-array-element": "value"
+              "duplicate-array-element": "value",
+              "unique-array-element": "value"
             },
             {
-              "array-element": "value", 
+              "duplicate-array-element": "value", 
               "nested-array-element": {"nested": "value"}
             }
           ]
@@ -108,11 +108,11 @@ WITH
     --
     SELECT
       assert_array_equals_any_order(no_args, ARRAY['`first`.`second`.`third`', '`first`.`other-second`',
-        '`first`.`array`.[...].`nested-array-element`.`nested`', '`first`.`array`.[...].`array-element`', 
-        '`first`.`array`.[...].`another-array-element`']),
+        '`first`.`array`.[...].`nested-array-element`.`nested`', '`first`.`array`.[...].`duplicate-array-element`', 
+        '`first`.`array`.[...].`unique-array-element`']),
       assert_array_equals_any_order(indicates_node_arg, ARRAY['`first`.`second`', '`first`.`other-second`', 
-        '`first`.`array`.[...].`nested-array-element`.`nested`', '`first`.`array`.[...].`array-element`',
-        '`first`.`array`.[...].`another-array-element`']),
+        '`first`.`array`.[...].`nested-array-element`.`nested`', '`first`.`array`.[...].`duplicate-array-element`',
+        '`first`.`array`.[...].`unique-array-element`']),
       assert_array_equals_any_order(is_node_arg, ARRAY['`first`'])
   FROM
     extracted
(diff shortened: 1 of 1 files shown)


udf_js/json_extract_missing_cols.sql
```diff
@@ -37,43 +37,59 @@ LANGUAGE js AS """
...
 """;
 
 -- Tests
 
 WITH
   addl_properties AS (
     SELECT
-      '{"first": {"second": {"third": "hello world"}, "other-second": "value"}}' AS additional_properties ),
+      '''
+      {"first": {"second": {"third": "hello world"}, "other-second": "value",
```
fbertsch, line 88: Can we format this to be more readable?

udf_js/json_extract_missing_cols.sql
```diff
@@ -37,43 +37,59 @@ LANGUAGE js AS """
...
-
-            if(!is_node){
-                Object.keys(next_val).map(k =>
-                    remaining_paths.push(next_keypath.concat([k]))
-                );
+            if (Array.isArray(next_val)) {
+              next_val.forEach(function (item, index) {
+                remaining_paths.push(next_keypath.concat([index]))
+              });
+            } else {
```
fbertsch, line 47: We can move `is_node` to the else statement

udf_js/json_extract_missing_cols.sql
```diff
@@ -37,43 +37,59 @@ LANGUAGE js AS """
...
-
-            if(keys_indicating_node.length > 0){
-                is_node = true;
-            }
-
-            if(!is_node){
-                Object.keys(next_val).map(k =>
-                    remaining_paths.push(next_keypath.concat([k]))
-                );
+            if (Array.isArray(next_val)) {
```
fbertsch, line 43: Looks like we could use the existing codepath for appending to `remaining_paths` here:

```
>> var a = ["a", "b", "c"];
>> Object.keys(a)
Array(3) [ "0", "1", "2" ]
>> a[0]
"a"
>> a["0"]
"a"
```

udf_js/json_extract_missing_cols.sql
scholtzan: Moving it into the else statement would skip dictionaries that have non-object members since in those cases the `next_val` type would not be `object`.

udf_js/json_extract_missing_cols.sql
scholtzan: Oh nice!

udf_js/json_extract_missing_cols.sql
```diff
@@ -61,19 +61,42 @@ LANGUAGE js AS """
...
+      '''
+      {
+        "first": {
+          "second": {
+            "third": "hello world"
+          }, 
+          "other-second": "value",
+          "array": [
+            {
+              "array-element": "value",
```
fbertsch, line 89:
```suggestion
              "duplicate-array-element": "value",
```

udf_js/json_extract_missing_cols.sql
```diff
@@ -61,19 +61,42 @@ LANGUAGE js AS """
...
+              "another-array-element": "value"
+            },
+            {
+              "array-element": "value",
```
fbertsch, line 93:
```suggestion
              "duplicate-array-element": "value", 
```

udf_js/json_extract_missing_cols.sql
```diff
@@ -61,19 +61,42 @@ LANGUAGE js AS """
...
+      {
+        "first": {
+          "second": {
+            "third": "hello world"
+          }, 
+          "other-second": "value",
+          "array": [
+            {
+              "array-element": "value",
+              "another-array-element": "value"
```
fbertsch, line 90:
```suggestion
              "unique-array-element": "value"
```

PR #971, opened 2020-05-08
Add script to generate desktop glam queries

### LINKED ISSUE #760: Glam desktop should pick up new probes nightly
There needs to be a task that regenerates the queries for `clients_daily_*` before they are run daily so that they include the latest probes.

Ben-Wu:
(fixes #760)

Separate script for generating query to match fenix pattern.  So both generate and run can be run through airflow like `bash script/glam/generate_desktop_sql && script/glam/run_glam_sql`.

A follow-up to this could be to simplify tests by having some `local` argument in the generate scripts that generate the queries using constant set of probes so we don't need to update the schemas every time.  We can commit that to the repo since we don't need to keep the real queries here anymore.

### DIFF
--- a/script/glam/generate_and_run_desktop_sql
+++ b/script/glam/generate_and_run_desktop_sql
@@ -1,0 +1,57 @@
+#!/bin/bash
+
+# import run_query procedure from run_glam_sql without running the script
+IMPORT=true source script/glam/run_glam_sql
+
+
+function generate_and_run_sql {
+    local probe_type=$1
+    local windows_release_sample_percent=$2
+    local script_type
+    local query_name
+    local destination_table
+    local write_disposition
+
+    echo "generating sql for ${probe_type}"
+
+    case $probe_type in
+        histogram | keyed_histogram)
+            script_type="histogram"
+            ;;
+        scalar | keyed_scalar | keyed_boolean)
+            script_type="scalar"
+            ;;
+        *)
+            echo "probe probe_type must be one of { histogram | keyed_histogram | scalar | keyed_scalar | keyed_boolean }"
+            exit 1
+            ;;
+    esac
+
+    python3 "sql/telemetry_derived/clients_daily_${script_type}_aggregates_v1.sql.py" \
+        "--agg-type=${probe_type}s" \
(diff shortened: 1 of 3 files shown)


Ben-Wu:
After some airflow testing I ran into this issue https://github.com/mozilla/telemetry-airflow/issues/844 because the queries take too long to execute without writing to stdout.  Workaround is to not write logs to airflow but I'm also not finding any logs in gcp.  As it is now, it's not really usable since there's no visibility into what's happening.

We should try to find a workaround/fix but another thing to consider is instead of running the entire dag in a script, we keep the existing dag with all the `bigquery_etl_query` and just replace the `clients_daily_` query tasks with a generate and run query script that doesn't log.  That way it's possible to debug without logs.

emtwo:
> We should try to find a workaround/fix but another thing to consider is instead of running the entire dag in a script, we keep the existing dag with all the `bigquery_etl_query` and just replace the `clients_daily_` query tasks with a generate and run query script that doesn't log. That way it's possible to debug without logs.

I think my initial thought on this was to only replace the `clients_daily_*` with the generate/run commands, at least for now. So the downstream tasks in the glam dag would remain the same and depend on a new task that generates/runs `clients_daily_*`.

Have you tried this approach yet? Did it seem to have the same logging issue?

Ben-Wu:
I'll try only replacing the clients daily queries later but I expect it to still have the same issue.  That would still be an improvement though.

Ben-Wu:
Created https://github.com/mozilla/telemetry-airflow/pull/992 which uses these changes.  Added a script that generates and runs the clients daily queries so the rest of the dag is retained.

For the logging bug, the scalar queries have around 4-5 minutes of headroom before they crash (It looks like the tasks crash at around 9-10 minutes with no logging) so we can allow logs for those.  But the histogram queries go over so we would need to disable logs for those.  I'll copy this to the airflow pr too.

Ben-Wu:
I'll merge this Monday because I don't want to spend time fixing things tomorrow if things break.

Ben-Wu:
Just rebased and tested again.  I'll merge now

PR #972, opened 2020-05-11
Bug 1632635 - Make FxA Amplitude export PDT-based

fbertsch:
The FxA Amplitude project uses PDT-based days. The previous approach of using UTC-based days and casting to PDT-based is wrong, since the day offsets are actually different for the two timezones.

This approach assumes we run the airflow DAG at EOD-PDT, rather than EOD-UTC.


fbertsch:
This depends on first merging https://github.com/mozilla/telemetry-airflow/pull/981.

From there, we can re-run the entire history, which is one month. The schema hasn't changed so there is no need to re-created the table.

PR #973, opened 2020-05-11
Implement fxa_amplitude_user_ids table

jklukas:
And we refactor the sync send tab view to use this new table, pulling it under the `firefox_accounts` namespace.

### DIFF
--- a/sql/firefox_accounts_derived/fxa_amplitude_user_ids_v1/query.sql
+++ b/sql/firefox_accounts_derived/fxa_amplitude_user_ids_v1/query.sql
@@ -1,0 +1,25 @@
+WITH hmac_key AS (
+  SELECT
+    AEAD.DECRYPT_BYTES(
+      (SELECT keyset FROM `moz-fx-dataops-secrets.airflow_query_keys.fxa_prod`),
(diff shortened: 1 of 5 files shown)


sql/firefox_accounts_derived/fxa_amplitude_user_ids_v1/query.sql
```diff
@@ -0,0 +1,25 @@
...
+  ) AS user_id
+FROM
+  `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_auth_20*`
+WHERE
+  _TABLE_SUFFIX = FORMAT_DATE('%g%m%d', @submission_date)
+UNION DISTINCT
+SELECT
+  user_id
+FROM
+  fxa_amplitude_user_ids_v1
```
fbertsch, line 25: Looks like we're missing an `init.sql` here.

jklukas: Added.

jklukas:
Added init.sql. Ready for review again.

Issue #967, opened 2020-05-08
Public data metadata improvements


scholtzan:
Currently, the `all-dataset.json` metadata file has a `files` entry for each dataset. The link currently doesn't work and should probably either be removed or we should set up file listing for directories in GCS.

Additionally `metadata.json` files should probably not be stored in the `files/` directory as they will get listed as available files.

PR #977, opened 2020-05-12
Write files metadata to /files

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/967

Currently the `all-datasets.json` has for each dataset two files entries:

```
"files_metadata": "https://public-data.telemetry.mozilla.org/api/v1/tables/telemetry_derived/ssl_ratios/v1/files/metadata.json",
"files_uri": "https://public-data.telemetry.mozilla.org/api/v1/tables/telemetry_derived/ssl_ratios/v1/files",
```

Opening `files_uri` currently results in an error which caused some confusion. Instead we should write the list of available files to a `files` file: 

```
"files": "https://public-data.telemetry.mozilla.org/api/v1/tables/telemetry_derived/ssl_ratios/v1/files",
```

Then we also don't have two entries and everything becomes hopefully less confusing.

If that change makes sense then I'll also update the docs in https://docs.telemetry.mozilla.org/public_data/index.html

### DIFF
--- a/bigquery_etl/public_data/publish_gcs_metadata.py
+++ b/bigquery_etl/public_data/publish_gcs_metadata.py
@@ -80,7 +80,7 @@ class GcsTableMetadata:
         if self.metadata.review_bug() is not None:
             metadata_json["review_link"] = REVIEW_LINK + self.metadata.review_bug()
 
-        metadata_json["files"] = self.files_uri
+        metadata_json["files_uri"] = self.files_uri
         metadata_json["last_updated"] = self.last_updated_uri
 
         return metadata_json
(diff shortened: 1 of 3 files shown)


scholtzan:
Right, this PR makes sure that everything works as described in the proposal. In order for the files to get listed when accessing `v1/tables/telemetry_derived/ssl_ratios/v1/files` there needs to be a `files` file in the `v1/` directory. This is currently not the case. Currently the list of available files is in `telemetry_derived/ssl_ratios/v1/files/metadata.json`.

fbertsch:
@scholtzan I see. I think `files` and `files_uri` are both fine names for the key, but it should point to a valid file that lists the available data files. Pointing to the `metadata.json` file is good, but may be a bit more confusing than simply `/files`.

fbertsch:
There is some precedent around "hitting the directory for an API lists the next level of files available", which is how we arrived at `/files` as both the parent of the data files and a file itself to list them.

scholtzan:
Yeah, I think `/files` is more intuitive.

PR #978, opened 2020-05-12
Add timestamp to fxa export

fbertsch:
I forgot the manufactured timestamp! Thank god for views.

### DIFF
--- a/sql/firefox_accounts/fxa_amplitude_export/view.sql
+++ b/sql/firefox_accounts/fxa_amplitude_export/view.sql
@@ -4,7 +4,6 @@ AS
 WITH active_users AS (
   SELECT
     TIMESTAMP(submission_date_pacific, "America/Los_Angeles") AS submission_timestamp,
-    TIMESTAMP(submission_date_pacific, "America/Los_Angeles") AS timestamp,
     `moz-fx-data-shared-prod`.udf.active_values_from_days_seen_map(
       os_used_month,
       0,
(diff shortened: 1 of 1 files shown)


sql/firefox_accounts/fxa_amplitude_export/view.sql
```diff
@@ -4,6 +4,7 @@ AS
 WITH active_users AS (
   SELECT
     TIMESTAMP(submission_date_pacific, "America/Los_Angeles") AS submission_timestamp,
+    TIMESTAMP(submission_date_pacific, "America/Los_Angeles") AS timestamp,
```
jklukas, line 7: Is this field particular to the configuration for this export?

In my discussion with Amplitude, they indicated that the preferred format was a `time` field given as milliseconds from unix epoch:

https://github.com/mozilla/bigquery-etl/blob/43a96d1d01c70c2c1aac15607e5924627f887154/sql/firefox_accounts/sync_send_tab_export/view.sql#L48-L54

If the string timestamp given here is already working, though, I won't block on optimizing this.

fbertsch: You make a good point. The original developer, Kurt, added a special transformation for this timestamp field - the `vacuum_config.json` file has `"time": ["iso_time_to_ms", "timestamp"],`, which does the transformation you mentioned.

fbertsch: Generally I prefer this way to keep the dataset consistent with how we store data internally, with the caveat that there's more transformations during ingestion.

PR #983, opened 2020-05-13
Change FxA label to "fxa" since only lowercase is allowed

jklukas:


jklukas:
The `validate-metadata` CI task doesn't fail, but it does emit error messages about the label "FxA".

PR #979, opened 2020-05-12
Add query for nondesktop day 2-7 activation metric to support weekly SC reporting

gkabbz:
@jklukas Adding in this query which should look pretty familiar to you. Also figuring out the process for adding queries to bq-etl so please be nitpicky if there are things that aren't exactly the way they should. I'm gonna add to the documentation to the rest of the DS team on how to do this so it saves DE time.

PR #984, opened 2020-05-13
Bug fix: Make cohort_date a derived field

jklukas:
Fix-up to #979 

I forgot during review that the telemetry-airflow machinery is going to replace partitions based on `submission_date`, and I had created this table with partitions based on `cohort_date`, causing a mismatch.

We should have the underlying table retain `submission_date`, and then we can present the shifted `cohort_date` as part of the user-facing view.

cc @gkabbz


jklukas:
> There is also currently a firefox_nondesktop_day_2_7_activation_v2 table in moz-fx-data-shared-prod. I'm guessing it was created just for testing purposes?

Yes, I was backfilling into that to create the new structure. I've now moved that to v1 and will remove the v2.

PR #985, opened 2020-05-13
Bug 1610925 - Add activity stream tables to shredder

relud:
fixes [bug  1610925](https://bugzilla.mozilla.org/show_bug.cgi?id=1610925)

### DIFF
--- a/bigquery_etl/shredder/config.py
+++ b/bigquery_etl/shredder/config.py
@@ -44,12 +44,12 @@ class DeleteTarget:
     @property
     def table_id(self):
         """Table Id."""
-        return self.table.split(".", 1)[-1]
+        return self.table.partition(".")[2]
 
     @property
     def dataset_id(self):
         """Dataset Id."""
-        return self.table.split(".", 1)[0]
+        return self.table.partition(".")[0]
 
 
 CLIENT_ID = "client_id"
@@ -61,6 +61,7 @@ SHIELD_ID = "shield_id"
 ECOSYSTEM_CLIENT_ID = "payload.ecosystem_client_id"
 PIONEER_ID = "payload.pioneer_id"
 ID = "id"
+CFR_ID = f"COALESCE({CLIENT_ID}, {IMPRESSION_ID})"
 
 DESKTOP_SRC = DeleteSource(
     table="telemetry_stable.deletion_request_v4", field=CLIENT_ID
@@ -68,6 +69,12 @@ DESKTOP_SRC = DeleteSource(
 IMPRESSION_SRC = DeleteSource(
     table="telemetry_stable.deletion_request_v4",
     field="payload.scalars.parent.deletion_request_impression_id",
+)
+CFR_SRC = DeleteSource(
+    # inject sql via f"`{sql_table_id(source)}`" to select client_id and impression_id
+    table="telemetry_stable.deletion_request_v4`,"
+    f" UNNEST([{CLIENT_ID}, {IMPRESSION_SRC.field}]) AS `_",
+    field="_",
 )
 FENIX_NIGHTLY_SRC = DeleteSource(
     table="org_mozilla_fenix_nightly_stable.deletion_request_v1", field=GLEAN_CLIENT_ID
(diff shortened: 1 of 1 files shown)


bigquery_etl/shredder/config.py
```diff
@@ -272,6 +272,24 @@ def dataset_id(self):
...
+    client_id_target(table='messaging_system_derived.onboarding_users_daily_v1'): DESKTOP_SRC,
+    client_id_target(table='messaging_system_derived.onboarding_users_last_seen_v1'): DESKTOP_SRC,
+    client_id_target(table='messaging_system_derived.snippets_users_daily_v1'): DESKTOP_SRC,
+    client_id_target(table='messaging_system_derived.snippets_users_last_seen_v1'): DESKTOP_SRC,
+    impression_id_target(table="activity_stream_stable.impression_stats_v1"): IMPRESSION_SRC,
+    impression_id_target(table="activity_stream_stable.spoc_fills_v1"): IMPRESSION_SRC,
+    impression_id_target(table="messaging_system_stable.undesired_events_v1"):IMPRESSION_SRC,
+    impression_id_target(table='messaging_system_derived.cfr_users_daily_v1'): IMPRESSION_SRC,
+    impression_id_target(table='messaging_system_derived.cfr_users_last_seen_v1'): IMPRESSION_SRC,
+    impression_id_target(table='messaging_system_stable.personalization_experiment_v1'): IMPRESSION_SRC,
```
relud, line 292: @ncloudioj can you confirm that `client_id`/`impression_id` in these tables that map to `client_id`/`payload.scalars.parent.deletion_request_impression_id` in `telemetry_stable.deletion_request_v4`?

ncloudioj: @relud LGTM except for one special case.

The `message_system_stable.cfr_v1` table has both `client_id` and `impressoin_id` in its schema. We're collecting `impresson_id` in the release channel, and `client_id` in the pre-release channels as well as the CFR experiment audience in the release. They never show up together in a single ping though (except the `deletion-request` ping).

So we also want to add it to the `impression_id_target`.

bigquery_etl/shredder/config.py
```diff
@@ -272,6 +272,44 @@ def dataset_id(self):
...
+    impression_id_target(
+        table="activity_stream_stable.impression_stats_v1"
+    ): IMPRESSION_SRC,
+    impression_id_target(table="activity_stream_stable.spoc_fills_v1"): IMPRESSION_SRC,
+    impression_id_target(
+        table="messaging_system_stable.undesired_events_v1"
+    ): IMPRESSION_SRC,
+    impression_id_target(
+        table="messaging_system_derived.cfr_users_daily_v1"
+    ): IMPRESSION_SRC,
```
jklukas, line 306:
```suggestion
    impression_id_target(
        table="messaging_system_stable.cfr_v1"
    ): IMPRESSION_SRC,
    impression_id_target(
        table="messaging_system_derived.cfr_users_daily_v1"
    ): IMPRESSION_SRC,
```

I think we also need to list cfr_v1 for both clientId and impression_id

relud: i need to update this so that a target is correctly handled with more than one source

bigquery_etl/shredder/config.py
relud: > They never show up together in the same ping though (except the `deletion-request` ping).

Thank you for this bit of information, it was v helpful in getting this working.

bigquery_etl/shredder/config.py
```diff
@@ -69,6 +70,12 @@ def dataset_id(self):
     table="telemetry_stable.deletion_request_v4",
     field="payload.scalars.parent.deletion_request_impression_id",
 )
+CFR_SRC = DeleteSource(
+    # inject sql via f"`{sql_table_id(source)}`" to select client_id and impression_id
+    table="telemetry_stable.deletion_request_v4`,"
+    f" UNNEST([{CLIENT_ID}, {IMPRESSION_SRC.field}]) AS `_",
+    field="_",
+)
```
relud, line 78: I did this instead of repeating cfr tables with multiple sources, so that we don't have to scan the tables twice, because repeating a table would probably cause issues attempting to update the same partition simultaneously, especially because jobs are sorted by table and partition.

relud: @jklukas hit me up if you don't like this, and we can discuss alternative solutions.

jklukas: This is clever. It's indeed rather surprising, but probably not worth refactoring unless we find other tables that also have this problem.

PR #912, opened 2020-04-15
Change naming convention of sql/glam_etl

acmiyaguchi:
This implements the first two changes from #897 to use the `{product}__` prefix for tables. This should make it easier to add new products.

```
% bq ls glam-fenix-dev:glam_etl_dev
                                 tableId                                   Type    Labels   Time Partitioning          Clustered Fields          
 ------------------------------------------------------------------------ ------- -------- ------------------- --------------------------------- 
  org_mozilla_fenix__clients_daily_histogram_aggregates_metrics_v1         TABLE            DAY                                                  
  org_mozilla_fenix__clients_daily_scalar_aggregates_activation_v1         TABLE            DAY                                                  
  org_mozilla_fenix__clients_daily_scalar_aggregates_baseline_v1           TABLE            DAY                                                  
  org_mozilla_fenix__clients_daily_scalar_aggregates_bookmarks_sync_v1     TABLE            DAY                                                  
  org_mozilla_fenix__clients_daily_scalar_aggregates_deletion_request_v1   TABLE            DAY                                                  
  org_mozilla_fenix__clients_daily_scalar_aggregates_events_v1             TABLE            DAY                                                  
  org_mozilla_fenix__clients_daily_scalar_aggregates_history_sync_v1       TABLE            DAY                                                  
  org_mozilla_fenix__clients_daily_scalar_aggregates_installation_v1       TABLE            DAY                                                  
  org_mozilla_fenix__clients_daily_scalar_aggregates_logins_sync_v1        TABLE            DAY                                                  
  org_mozilla_fenix__clients_daily_scalar_aggregates_metrics_v1            TABLE            DAY                                                  
  org_mozilla_fenix__clients_daily_scalar_aggregates_migration_v1          TABLE            DAY                                                  
  org_mozilla_fenix__clients_daily_scalar_aggregates_sync_v1               TABLE            DAY                                                  
  org_mozilla_fenix__clients_histogram_aggregates_v1                       TABLE                                app_version, channel, client_id  
  org_mozilla_fenix__clients_scalar_aggregates_v1                          TABLE                                app_version, channel, client_id  
  org_mozilla_fenix__extract_probe_counts_v1                               TABLE                                                                 
  org_mozilla_fenix__extract_user_counts_v1                                TABLE                                                                 
  org_mozilla_fenix__histogram_bucket_counts_v1                            TABLE                                                                 
  org_mozilla_fenix__histogram_percentiles_v1                              TABLE                                                                 
[...]

PR #986, opened 2020-05-18
Update glam-fenix etl with static combinations

acmiyaguchi:
This depends on #912. 

TODO: compare number of rows between heads of previous and current PR.

### DIFF
--- a/bigquery_etl/glam/models.py
+++ b/bigquery_etl/glam/models.py
@@ -128,12 +128,12 @@ def clients_histogram_bucket_counts(**kwargs):
 def probe_counts(**kwargs):
     """Variables for probe counts."""
     attributes = ["ping_type", "os", "app_version", "app_build_id", "channel"]
-
+    fixed_attributes = ["app_version", "channel"]
+    cubed_attributes = [x for x in attributes if x not in fixed_attributes]
     return dict(
         attributes=attributes,
-        attribute_combinations=compute_datacube_groupings(
-            attributes, ["app_version", "channel"]
-        ),
+        cubed_attributes=cubed_attributes,
+        attribute_combinations=compute_datacube_groupings(cubed_attributes),
         aggregate_attributes="""
             metric,
             metric_type,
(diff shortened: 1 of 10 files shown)


acmiyaguchi:
@emtwo I've updated the queries to take advantage of the static combination trick that was used in the desktop histogram probe counts query. Do you think it's prudent to also merge the bucket counts and probe counts queries here to avoid double counting?

emtwo:
Thanks for these updates! 

> Do you think it's prudent to also merge the bucket counts and probe counts queries here to avoid double counting?

I don't think merging bucket counts and probe counts is necessary, just the data cubing by client should be sufficient to address the double counting.

I do want to point out this issue: https://github.com/mozilla/bigquery-etl/issues/987
in which I describe why I'm thinking about bringing back `clients_histogram_bucket_counts_v1`.

This PR and #912 are on my radar for the week.

acmiyaguchi:
Thanks, issue #987 makes sense after reviewing the related PR. I'll keep it in mind as I work through this.

emtwo:
After taking a closer look here, I think I may have misunderstood your question earlier. In order to address the double counting issue, we need to apply the data cubing/static combos per `client_id`. That way, if a client has several rows with different `app_build_id`, they are counted only once when we group over all build ids. In this PR, the static combos would have still been applied after `client_id` was excluded in the grouping in `clients_histogram_bucket_counts`, right?

acmiyaguchi:
> After taking a closer look here, I think I may have misunderstood your question earlier. In order to address the double counting issue, we need to apply the data cubing/static combos per `client_id`. That way, if a client has several rows with different `app_build_id`, they are counted only once when we group over all build ids. In this PR, the static combos would have still been applied after `client_id` was excluded in the grouping in `clients_histogram_bucket_counts`, right?

Yes, this actually doesn't handle double counting of any kind. I've added #998 to add client_id as an attribute to cube over in bucket counts. This also changes the name of the bucket count tables to include the `clients_` prefix, since it now does include client ids again.

emtwo:
> Yes, this actually doesn't handle double counting of any kind. I've added #998 to add client_id as an attribute to cube over in bucket counts. This also changes the name of the bucket count tables to include the `clients_` prefix, since it now does include client ids again.

Got it, ok great. I look this over again tomorrow as well as #998. Thanks!

acmiyaguchi:
Closed in favor of #1002

PR #996, opened 2020-05-21
Bug 1577858 - Create a view for events to ingest into Amplitude for M…

ncloudioj:
…essaging System Onboarding

This follows the [fenix_events_v1](https://github.com/mozilla/bigquery-etl/blob/master/sql/telemetry/fenix_events_v1/view.sql) to create a view for the data ingestion into Amplitude.

Notes:
* The daily events volume of this table is about 1.5M for all channels 
*  Both `device_manufacturer` and `device_model` are set to NULL as they are not applicable in this case
* A utility function `kv_array_append_to_json_string` was added based on `kv_array_to_json_string` to facilitate some JSON processing. It's now defined as `TEMP` as this is the only use case here.

r? @fbertsch


ncloudioj:
Oops, looks like we can't define the TEMP function in a view definition. Will move it to `udf` as it could be useful in the future.

udf/kv_array_append_to_json_string.sql
```diff
@@ -0,0 +1,35 @@
...
+  udf_kv_array_append_to_json_string('{}', [STRUCT("baz" AS key, "boo" AS value)])
+
+  '{"baz": "boo"}'
+
+*/
+CREATE OR REPLACE FUNCTION udf.kv_array_append_to_json_string(input STRING, arr ANY TYPE) AS (
+  CONCAT(
+    RTRIM(input, "}"),
+    IF(input = "{}", "", ","),
+    TRIM(`moz-fx-data-shared-prod.udf.kv_array_to_json_string`(arr), "{")
```
jklukas, line 19:
```suggestion
    TRIM(udf.kv_array_to_json_string(arr), "{")
```

Within the UDF definitions, we do some substitutions in testing, so prefer non-fully-qualified UDF refs. View definitions, on the other hand, are required by BQ to use fully-qualified references, so you're doing the right thing there.

udf/kv_array_append_to_json_string.sql
```diff
@@ -0,0 +1,35 @@
...
+
+  udf_kv_array_append_to_json_string('{}', [STRUCT("baz" AS key, "boo" AS value)])
+
+  '{"baz": "boo"}'
+
+*/
+CREATE OR REPLACE FUNCTION udf.kv_array_append_to_json_string(input STRING, arr ANY TYPE) AS (
+  CONCAT(
+    RTRIM(input, "}"),
+    IF(input = "{}", "", ","),
```
jklukas, line 18:
```suggestion
    IF(input IS NULL OR input = "{}", "{", RTRIM(input, "}") || ",")
```

We could check for the null case here, which would remove the need for an `IFNULL` check in the view.

ncloudioj:
Hey @fbertsch, could you take a look at this, plz?

sql/messaging_system/onboarding_events_amplitude/view.sql
```diff
@@ -0,0 +1,39 @@
...
+  submission_timestamp,
+  client_id AS device_id,
+  document_id AS insert_id,
+  event AS event_type,
+  submission_timestamp AS timestamp,
+  version AS app_version,
+  metadata.user_agent.os AS platform,
+  metadata.user_agent.os AS os_name,
+  normalized_os_version AS os_version,
+  NULL AS device_manufacturer,
```
fbertsch, line 14: We can drop these if they're null.

ncloudioj:
@fbertsch Thanks!

Let me know if there is anything more needed from our end.

PR #998, opened 2020-05-21
Fix double counting in glam-fenix etl

acmiyaguchi:
This should fix double counting in the glam fenix etl by grouping by client id in the bucket counts job, re: https://github.com/mozilla/bigquery-etl/pull/986#issuecomment-631583965


bigquery_etl/glam/generate.py
```diff
@@ -122,25 +122,28 @@ def main():
             **models.clients_histogram_aggregates(parameterize=True),
         ),
         table(
-            "scalar_bucket_counts_v1",
-            **models.scalar_bucket_counts(
+            "clients_scalar_bucket_counts_v1",
+            **models.clients_scalar_bucket_counts(
```
acmiyaguchi, line 126: This reverts the name of the table since the query is now granular at the client id level.

acmiyaguchi: Actually, it looks like I misunderstood this. Client ids should not show up in the resulting table, but only during cubing.

acmiyaguchi:
This doesn't seem to fix double counting properly. Closing in favor of a new branch that closely mirrors the desktop changes.

PR #1000, opened 2020-05-21
Support extraction of histogram in compact encoding

jklukas:
Per the [Compact String Encoding for Histograms Proposal](https://docs.google.com/document/d/1k_ji_1DB6htgtXnPpMpa7gX0klm-DGV5NMY7KkvVB00/edit#)


udf/json_extract_histogram.sql
```diff
@@ -1,53 +1,4 @@
...
-  assert_array_equals([1, 100], `range`),
-  assert_array_equals([STRUCT(0 AS key, 12434 AS value),
-                       STRUCT(1 AS key, 297 AS value),
-                       STRUCT(13 AS key, 8 AS value)],
-                      `values`)
-FROM
-  extracted
+-- Legacy wrapper around a function moved to mozfun.
+CREATE OR REPLACE FUNCTION udf.json_extract_histogram(input STRING) AS (
+  mozfun.hist.extract(input)
```
jklukas, line 3: I may need to split out these refactorings to a separate PR, since we can't deploy mozfun.hist without merging and letting Airflow create the dataset.

mozfun/hist/extract/udf.sql
```diff
@@ -0,0 +1,174 @@
...
+      4 AS histogram_type,
+      CAST(SPLIT(input, ',')[OFFSET(0)] AS INT64) AS `sum`,
+      [1, 2] AS `range`,
+      [
+        STRUCT(0 AS key, CAST(SPLIT(input, ',')[OFFSET(0)] AS INT64) AS value) STRUCT(
+          1 AS key,
+          0 AS INT64
+        ) AS value
+      )
+    ] AS `values`
```
scholtzan, line 73:
```suggestion
      [
        STRUCT(0 AS key, CAST(SPLIT(input, ',')[OFFSET(0)] AS INT64) AS value),
        STRUCT(
          1 AS key,
          0 AS value
        )
    ] AS `values`
```

Might need re-formatting.

udf/json_extract_histogram.sql
scholtzan: It might be nice for our team to have permissions to add/delete datasets (probably also UDFs) for this project.

PR #1002, opened 2020-05-22
[glam-etl] Use static combinations and avoid double counting

acmiyaguchi:
This supercedes #986 and #998.


sql/glam_etl/org_mozilla_fenix__histogram_bucket_counts_v1/query.sql
```diff
@@ -80,7 +80,41 @@ RETURNS ARRAY<
...
+    table.* EXCEPT (ping_type, os, app_build_id),
+    COALESCE(combo.ping_type, table.ping_type) AS ping_type,
+    COALESCE(combo.os, table.os) AS os,
+    COALESCE(combo.app_build_id, table.app_build_id) AS app_build_id
+  FROM
+    glam_etl.org_mozilla_fenix__clients_histogram_aggregates_v1 table
+  CROSS JOIN
+    static_combos combo
+),
+normalized_histograms AS (
```
emtwo, line 117: Before applying normalization, there needs to be a `group by` including `client_id` so that the clients that have multiple rows for `*` build_id are grouped into 1 row to represent them and that is the row that gets normalized

sql/glam_etl/org_mozilla_fenix__scalar_bucket_counts_v1/query.sql
```diff
@@ -399,6 +399,39 @@ booleans_and_scalars AS (
...
+all_combos AS (
+  SELECT
+    table.* EXCEPT (ping_type, os, app_build_id),
+    COALESCE(combo.ping_type, table.ping_type) AS ping_type,
+    COALESCE(combo.os, table.os) AS os,
+    COALESCE(combo.app_build_id, table.app_build_id) AS app_build_id
+  FROM
+    booleans_and_scalars table
+  CROSS JOIN
+    static_combos combo
```
emtwo, line 434: Same thing as for histogram bucket counts, we need to re-aggregate by client_id after thus. This is actually wrong the way it's done for desktop right now and I'm in the process of fixing it.

sql/glam_etl/org_mozilla_fenix__histogram_bucket_counts_v1/query.sql
acmiyaguchi: This is to catch cases where a null build id is being reported?

acmiyaguchi: My understanding is that the source table (clients histogram aggregates) should only have a single row per dimension. Null build ids should be invalid (though I notice there is no filter in the base queries for clients daily). Do you have an example where a single row could have multiple `*` in the cross joined table?

acmiyaguchi:
These last two commits should address the review comments. Let me know if this is the right approach.

sql/glam_etl/org_mozilla_fenix__histogram_bucket_counts_v1/query.sql
emtwo: Sorry, my explanation wasn't very good. You handled it correctly with your new changes. I just meant that a single client could have multiple builds so when we group over all build ids they need to be deduped which you have done here.

PR #1003, opened 2020-05-22
Add ISP data to main_summary and clients_daily

scholtzan:
I guess it makes sense to add ISP data to main_summary and clients_daily?

See https://bugzilla.mozilla.org/show_bug.cgi?id=1505491


sql/telemetry_derived/main_summary_v4/part1.sql
```diff
@@ -7,6 +7,8 @@ SELECT
   normalized_os_version,
   metadata.geo.country,
   metadata.geo.city,
+  metadata.isp.name AS isp_name,
+  metadata.isp.organization AS isp_organization,
   metadata.geo.subdivision1 AS geo_subdivision1,
   metadata.geo.subdivision2 AS geo_subdivision2,
```
jklukas, line 13:
```suggestion
  metadata.geo.subdivision1 AS geo_subdivision1,
  metadata.geo.subdivision2 AS geo_subdivision2,
  metadata.isp.name AS isp_name,
  metadata.isp.organization AS isp_organization,
```

I'd prefer to keep all geo grouped together.

relud:
> main_summary caveats about choosing part1 vs. part2

part1 is the right choice, part2 is just a long list of histograms and scalars.

> whether we'll run into issues with field addition. I think Airflow won't be allowed to add fields automatically. We may need to manually add the relevant fields before merging.

yeah, the fields need to be added manually to main summary, clients daily, and clients last seen

scholtzan:
I added the fields to main summary, clients daily, clients last seen and clients first seen. And also updated the tests.

PR #1004, opened 2020-05-22
Desktop Glam scalar bug fixes and enhancements.

emtwo:
There are several changes here but my main motivation was that `total_users` was different for scalars between the histogram and percentile rows ([here for example](https://console.cloud.google.com/bigquery?project=glam-desktop-dev&j=bq:US:bquxjob_2a9716ff_1723d8555ac&page=queryresults)). So I noticed some other issues and here's a summary:

* Using the more efficient "static combos" method for data cubing
* The bucket for a scalar should be applied after a client has been re-aggregated across the static combo dimensions
* Merged `clients_scalar_bucket_counts` and `scalar_percentiles` into `clients_scalar_probe_counts`. This ran successfully and it's cheaper than reading from `clients_scalar_aggregates` separately for `scalar_percentiles`
* Factored out `udf.merge_scalar_user_data` which is used in `clients_scalar_aggregates` as well
* This goes with this airflow PR: https://github.com/mozilla/telemetry-airflow/pull/1000


udf/merge_scalar_user_data.sql
```diff
@@ -0,0 +1,105 @@
+CREATE OR REPLACE FUNCTION udf.merge_scalar_user_data(
```
acmiyaguchi, line 1: I suggest a docstring and a few tests as self documentation on how to use this function.

acmiyaguchi:
Does this same process have to be repeated for histogram probes, too?

emtwo:
> Does this same process have to be repeated for histogram probes, too?

I believe histograms should be covered because here: https://github.com/mozilla/bigquery-etl/blob/master/sql/telemetry_derived/clients_histogram_bucket_counts_v1/query.sql#L104 we do a `map_sum` over all values for a client before normalization.

emtwo:
Unfortunately, I had to bring back `scalar_percentiles`. I realized when I broke up probe counts into partitions by `sample_id`, this had the same issue that I fixed for histograms with duplicate rows for a set of dimensions. The last aggregation step for both percentiles and probe counts relied on different intermediate tables and the earliest intermediate table they shared was `user_aggregates`. This table was going to be large enough that it's not very helpful as an intermediate in terms of cost savings, so I reverted to having separate scalar percentiles again.

udf/merge_scalar_user_data.sql
emtwo: added! thank you

acmiyaguchi: Looks great.

PR #1005, opened 2020-05-22
Validate generated DAGs in CI

scholtzan:
Pulls in telemetry-airflow and runs the `test-environment` step with the DAGs generated in bigquery-etl

Depends on https://github.com/mozilla/bigquery-etl/pull/980


.circleci/config.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       - run:
           name: Verify that SQL is correctly formatted
           # check all directories owned by git except sql
-          command: script/format_sql --check $(git ls-tree -d HEAD --name-only)
+          command: script/format_sql --check $(git ls-tree -d HEAD --name-only | grep -vx sql)
```
jklukas, line 37: Is this an unintentional reversion?

scholtzan: oops, good catch

PR #1006, opened 2020-05-22
Add query to get processes found for each probe in glam

### LINKED ISSUE #955: Add a task in the desktop glam dag to create a table mapping probe -> processes

Ben-Wu:
(fixes #955)

Results: https://sql.telemetry.mozilla.org/queries/71370/source

This query is simple and fast enough that it could just exist as a view.  That way we don't need to maintain another table and airflow task.  I don't think it would matter to the frontend since it only takes a couple seconds to run and I assume the results are cached.  Thoughts on view vs query?

### DIFF
--- a/sql/telemetry/clients_probe_processes/view.sql
+++ b/sql/telemetry/clients_probe_processes/view.sql
@@ -1,3 +1,6 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.client_probe_processes`
+AS
 SELECT
   metric,
   os,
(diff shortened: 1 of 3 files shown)


emtwo:
> This query is simple and fast enough that it could just exist as a view. That way we don't need to maintain another table and airflow task. I don't think it would matter to the frontend since it only takes a couple seconds to run and I assume the results are cached. Thoughts on view vs query?

Yes, good idea, let's make it a view.

Ben-Wu:
Made it into a view and updated the example https://sql.telemetry.mozilla.org/queries/71370/source

PR #1007, opened 2020-05-22
Add regular_users_v3 and new_or_resurrected_v3.

felixlawrence:
These segments/user states got a DS review [when added to DTMO](https://github.com/mozilla/firefox-data-docs/pull/466).

Replaces #825.

### DIFF
--- a/sql/telemetry/clients_last_seen_v1/view.sql
+++ b/sql/telemetry/clients_last_seen_v1/view.sql
@@ -13,6 +13,8 @@ SELECT
   CAST(
     SAFE.LOG(days_created_profile_bits & -days_created_profile_bits, 2) AS INT64
   ) AS days_since_created_profile,
+  BIT_COUNT(days_seen_bits & 0x0FFFFFFE) >= 14 AS regular_users_v3,
+  days_seen_bits & 0x0FFFFFFE = 0 AS new_or_resurrected_v3,
   * EXCEPT (
     active_experiment_id,
     scalar_parent_dom_contentprocess_troubled_due_to_memory_sum,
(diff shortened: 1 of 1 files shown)


sql/telemetry/clients_last_seen_v1/view.sql
```diff
@@ -13,6 +13,8 @@ SELECT
   CAST(
     SAFE.LOG(days_created_profile_bits & -days_created_profile_bits, 2) AS INT64
   ) AS days_since_created_profile,
+  BIT_COUNT(days_seen_bits & 0x0FFFFFFE) >= 14 AS regular_users_v3,
+  days_seen_bits & 0x0FFFFFFE = 0 AS new_or_resurrected_v3,
```
jklukas, line 17:
```suggestion
  -- Segment definitions; see https://docs.telemetry.mozilla.org/concepts/segments.html
  -- 0x0FFFFFFE is a bitmask that accepts the previous 27 days, excluding the current day (rightmost bit)
  BIT_COUNT(days_seen_bits & 0x0FFFFFFE) >= 14 AS regular_users_v3,
  days_seen_bits & 0x0FFFFFFE = 0 AS new_or_resurrected_v3,
```

jklukas: Can we name these with `is_` prefixes, so `is_regular_user_v3` and `is_new_or_resurrected_v3`? Or perhaps `segment_regular_users_v3`? I think the `regular_users_v3` name sounds like it's a count rather than a boolean.

felixlawrence: Sure. Yes it would have been good to use a `segment_*` namespace for these, but in the medium term I've been told we need to start referring to them as "user states" rather than "segments", and `user_state_regular_user_v3` feels weird and verbose to me. So `is_regular_user_v3` it is!

jklukas:
This is now merged and published.

PR #1008, opened 2020-05-22
Fix SQL format validation step in CI

scholtzan:
CI currently does not check formatting for files in `sql/`. I believe this might be a remnant from when queries used to be defined in `templates/`.


sql/search_derived/mobile_search_clients_daily_v1/fenix_metrics.template.sql
```diff
@@ -1,14 +1,14 @@
 -- baseline for {namespace} ({app_name} {channel})
-baseline_{namespace} AS (
+baseline_ { namespace } AS (
```
acmiyaguchi, line 2: You may want to ignore all of the templated files, since the formatting script doesn't handle brackets in this context well.

scholtzan: Good catch! I added the template files to the `SKIP` list

PR #1010, opened 2020-05-26
Add noreplace to glam query

Ben-Wu:
also remove unused variable


Ben-Wu:
Ran the dag locally and results seem good https://sql.telemetry.mozilla.org/queries/71443/source

PR #1011, opened 2020-05-27
[glam-etl] Use transpose logic for fenix extracts 

acmiyaguchi:


bigquery_etl/glam/templates/extract_probe_counts_v1.sql
```diff
@@ -1,48 +1,42 @@
...
 
 SELECT
     channel,
     app_version as version,
-    COALESCE(ping_type, "*") as ping_type,
-    COALESCE(os, "*") AS os,
-    COALESCE(app_build_id, "*") AS build_id,
+    ping_type,
+    os,
+    app_build_id as build_id,
```
robhudson, line 20: Do these already contain the `*` instead of `NULL`?

bigquery_etl/glam/templates/extract_probe_counts_v1.sql
```diff
@@ -1,48 +1,42 @@
...
     metric_type,
     -- BigQuery has some null unicode characters which Postgresql doesn't like,
     -- so we remove those here. Also limit string length to 200 to match column
     -- length.
     SUBSTR(REPLACE(key, r"\x00", ""), 0, 200) AS metric_key,
     client_agg_type,
-    agg_type,
-    total_users,
-    TO_JSON_STRING(aggregates) AS data
+    MAX(total_users) as total_users,
```
robhudson, line 28: FYI: I had `total_users` in the `GROUP BY` but there was a bug for desktop where these values were different, so I moved it here. This is just a temporary fix until desktop ETL is fixed. You probably don't need this.

bigquery_etl/glam/templates/extract_probe_counts_v1.sql
acmiyaguchi: Yes (see [the static combination logic](https://github.com/mozilla/bigquery-etl/pull/1002/files#diff-041939f9ac853ab7f51fbe7cdead1466R445-R497)). If there are nulls in the data, it's a bug that should be resolved in the underlying view instead of patching up here.

bigquery_etl/glam/templates/extract_probe_counts_v1.sql
emtwo: Yea. Thanks for pointing this out. The fix is coming here: https://github.com/mozilla/bigquery-etl/pull/1004

PR #1013, opened 2020-05-27
Lower-friction scheduled queries: external telemetry-airflow dependencies

scholtzan:
This PR depends on https://github.com/mozilla/bigquery-etl/pull/980

This PR adds support to automatically determine telemetry-airflow dependencies. I'm not sure if we actually want to add this support or if users should only be able to specify such dependencies manually?

As a next step, we should support manual configuration of telemetry-airflow dependencies.


bigquery_etl/query_scheduling/telemetry_airflow.py
```diff
@@ -0,0 +1,121 @@
...
+        )
+        table = re.findall(
+            r"destination_table[\n\r\s]*=[\n\r\s]*['\"](?P<tab>[^'\"]*)['\"][\n\r\s]*,",
+            fn,
+        )
+
+        dataset_id = re.findall(
+            r"dataset_id[\n\r\s]*=[\n\r\s]*['\"](?P<dataset>[^'\"]*)['\"][\n\r\s]*,",
+            fn,
+        )
```
scholtzan, line 88: I decided to use regex to determine dependencies in telemetry-airflow. I looked into using `airflow` directly for parsing DAGs, however there are a lot of additional dependencies that need to be installed, some of the causing conflicts with library versions used here. Plus parsing the DAGs wasn't as fast as just using a regex.

It's probably best to add tests to make sure the regex approach is working correctly.

dags/bqetl_addons_daily.py
```diff
@@ -0,0 +1,50 @@
...
+        wait_for_search_clients_daily_bigquery
+    )
+
+    wait_for_clients_last_seen = ExternalTaskSensor(
+        task_id="wait_for_clients_last_seen",
+        external_dag_id="main_summary",
+        external_task_id="clients_last_seen",
+    )
+
+    telemetry_derived__addons_daily__v1.set_upstream(wait_for_clients_last_seen)
```
scholtzan, line 50: This looks pretty much identical to https://github.com/mozilla/telemetry-airflow/blob/master/dags/addons_daily.py

dags/bqetl_kpi_dashboard.py
```diff
@@ -54,6 +54,64 @@
...
+
+    wait_for_firefox_nondesktop_exact_mau28 = ExternalTaskSensor(
+        task_id="wait_for_firefox_nondesktop_exact_mau28",
+        external_dag_id="copy_deduplicate",
+        external_task_id="firefox_nondesktop_exact_mau28",
+    )
+
+    telemetry__firefox_kpi_dashboard__v1.set_upstream(
+        wait_for_firefox_nondesktop_exact_mau28
+    )
```
scholtzan, line 117: Looking at https://github.com/mozilla/telemetry-airflow/blob/master/dags/kpi_dashboard.py, there are a lot more dependencies here which might not be strictly necessary but shouldn't do any harm either?

dags/bqetl_ssl_ratios.py
```diff
@@ -0,0 +1,32 @@
...
+    telemetry_derived__ssl_ratios__v1 = bigquery_etl_query(
+        task_id="telemetry_derived__ssl_ratios__v1",
+        destination_table="ssl_ratios_v1",
+        dataset_id="telemetry_derived",
+        project_id="moz-fx-data-shared-prod",
+        owner="chutten@mozilla.com",
+        email=["chutten@mozilla.com"],
+        depends_on_past=False,
+        dag=dag,
+    )
```
scholtzan, line 32: This is missing a dependency. In a future PR I'll add support for manually specifying telemetry-airflow dependencies for tasks.

scholtzan:
This is ready for a look. If this looks like something we would want to support, I'd go ahead and add more tests.

scholtzan:
I agree that there is too much added complexity here. Plus users might find the magic that is going on here confusing (Like "Where are all these external dependencies coming from?", "How can I remove them?!")

I think for now, having an option to manually specify telemetry-airflow dependencies is more useful. And ideally over time most of the dependencies that this approach detects in telemetry-airflow will move to bigquery-etl anyway.

scholtzan:
This PR had some exploratory value, but shouldn't get merged

PR #1015, opened 2020-05-28
Move all fxa derived tables to shared-prod firefox_accounts_derived

jklukas:
Supports the effort to move data out of derived-datasets.

I've confirmed with DS folks that there are very few direct use cases for the
intermediate tables, so tearing down existing user-facing views should not be
disruptive. I will send out an email to data users before merging this change.

For review:

- I am not going to run `format_sql` until right before merging, since it will complicate review, so expect the formatting CI test to fail
- I have included bash scripts for the table movement that needs to happen; I have run the `copy_tables.sh` so that dry runs should be able to work. I will make sure to backfill additional days when this all gets merged
- I have published some of the views to allow dry runs to succeed
- It's likely helpful to read https://github.com/mozilla/telemetry-airflow/pull/1004 first, since it's the simpler of the two PRs


scholtzan:
For the failing pytest case to succeed you'll need to rename the directories: `tests/telemetry/fxa_users_daily_v1/` → `tests/firefox_accounts_derived/fxa_users_daily_v1/`

PR #1018, opened 2020-05-28
Parallelize run_glam_sql script for glean data

acmiyaguchi:
I noticed that backfilling data takes longer than I expected. This is because there is a lot of unnecessary work being done when running the jobs sequentially. I've parallelized the jobs where it makes sense, and added two options. BACKFILL_ONLY will run the queries that build up the clients aggregate state. EXPORT_ONLY will take the current clients aggregate state and generate the downstream probes tables. 

```bash
# before
RESET=true DATASET=glam_etl_dev_perf script/glam/run_glam_sql 
16.35s user 9.30s system 8% cpu 5:02.60 total

# after
RESET=true DATASET=glam_etl_dev_perf script/glam/run_glam_sql 
20.09s user 14.62s system 15% cpu 3:46.25 total

BACKFILL_ONLY=true RESET=true DATASET=glam_etl_dev_perf script/glam/run_glam_sql  
16.07s user 12.06s system 34% cpu 1:20.94 total

EXPORT_ONLY=true DATASET=glam_etl_dev_perf script/glam/run_glam_sql  
5.88s user 3.69s system 6% cpu 2:31.78 total
```

Here's a backfill script that I threw together: https://gist.github.com/acmiyaguchi/c6b7f7ccb803ff29c3fcecfe19dc3b51

### DIFF
--- a/script/glam/run_glam_sql
+++ b/script/glam/run_glam_sql
@@ -76,7 +76,7 @@ function run_init {
     local destination_table=$1
     local init="sql/$PROD_DATASET/$destination_table/init.sql"
     # run if needed
-    if ! bq show --quiet "${DATASET}.${destination_table}"; then
+    if ! bq show "${DATASET}.${destination_table}" &> /dev/null; then
         echo "running $init"
         local tmp
         tmp=$(mktemp)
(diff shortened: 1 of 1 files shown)


script/glam/run_glam_sql
```diff
@@ -187,42 +187,59 @@ function run_desktop_sql {
...
     local start_stage=${START_STAGE:-0}
+    local reset=${RESET:-false}
+    local backfill_only=${BACKFILL_ONLY:-false}
+    local export_only=${EXPORT_ONLY:-false}
+
+    if [[ $backfill_only = true && $export_only = true ]]; then
+        echo "BACKFILL_ONLY and EXPORT_ONLY are mutually exclusive"
+        exit 1
+    fi
+    if [[ $reset = true && $export_only = true ]]; then
```
Ben-Wu, line 198: I think it would be better to put this check before the `bq rm -r -f` so that if you wanted to just export but accidentally left reset true, you wouldn't delete everything.

PR #1020, opened 2020-05-29
Lower-friction scheduled queries: Handle public data export

scholtzan:
Adding support for automatically generating tasks for exporting public data to GCS.

I think having a separate DAG for the public data export makes most sense. Alternatively, export tasks could also be created inside the DAG for the original query, but it might be better to keep running queries separate from exporting data.

Once support for lower-friction scheduled queries lands in telemetry-airflow, the `entrypoint` script can be updated and calling the export script can be removed from there.


scholtzan:
This is ready for a look to see if the concept generally makes sense. Still needs tests.

jklukas:
This looks like a reasonable approach. 

I think we need to be careful and consistent about language, though. IIUC, this PR is about publishing of public JSON files to GCS and having that all happen in a single DAG. For queries that are published to tables in the public BQ project, that still happens in other DAGs, correct?

Can we make it clear in the PR name and DAG name that we're talking about exporting JSON data?

scholtzan:
This is ready for review.

PR #1021, opened 2020-06-01
Generate requirements with python 3.8

relud:
and enforce via CI

python dependencies are specific to the version of python installed

cc @jklukas this is why a dep is removed in https://github.com/mozilla/bigquery-etl/pull/893#discussion_r433275095

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2
 jobs:
   build:
-    docker:
+    docker: &docker
     - image: python:3.8
     steps:
     - checkout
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -18,7 +18,8 @@ jobs:
         name: Build
         command: |
           python3.8 -m venv venv/
-          venv/bin/pip install --upgrade -r requirements.txt
+          venv/bin/pip install pip-tools
+          venv/bin/pip-sync
```
relud, line 22: we should be using `pip-sync` instead of `pip install`, so that dependencies can be both installed and uninstalled here.

PR #1022, opened 2020-06-01
Changes in scalar bucketing.

emtwo:
A new method for scalar bucketing.

The bucketing is done by:
1. Taking the max and min value for every probe and finding the log base 2 value of those
2. Assigning 100 buckets by splitting the difference of the logged max and min linearly.
3. The created buckets are then exponentiated to bring them back to scale.


emtwo:
Got a thumbs up from Saptarshi and Hamilton on this approach so it's ready for review.

sql/telemetry_derived/clients_scalar_probe_counts_v1/query.sql
```diff
@@ -115,6 +121,25 @@ WITH flat_clients_scalar_aggregates AS (
     os = 'Windows' and channel = 'release' AS sampled,
   FROM clients_scalar_aggregates_v1),
 
+logged_min_max AS (
+  SELECT
+    metric,
+    key,
+    MIN(value) min,
+    MAX(value) max,
```
acmiyaguchi, line 129: min and max aren't being used anywhere from what I can tell

sql/telemetry_derived/clients_scalar_probe_counts_v1/query.sql
```diff
@@ -115,6 +121,25 @@ WITH flat_clients_scalar_aggregates AS (
     os = 'Windows' and channel = 'release' AS sampled,
   FROM clients_scalar_aggregates_v1),
 
+logged_min_max AS (
```
acmiyaguchi, line 124: nit:  `log_min_max`, logged doesn't seem to right when talking taking the log of something.

PR #1024, opened 2020-06-02
Fix second_seen_date bug in clients_first_seen

jklukas:
With the previous logic, we were always filling in second_seen_date where it was null. We should only do so if we see a ping for the given client on the current day.

cc @jmccrosky

### DIFF
--- a/sql/telemetry_derived/clients_first_seen_v1/query.sql
+++ b/sql/telemetry_derived/clients_first_seen_v1/query.sql
@@ -42,6 +42,12 @@ SELECT
       NULL
     WHEN
       previous.second_seen_date IS NULL
+      AND today.client_id IS NULL
+    THEN
+      NULL
+    WHEN
+      previous.second_seen_date IS NULL
+      AND today.client_id IS NOT NULL
     THEN
       @submission_date
     ELSE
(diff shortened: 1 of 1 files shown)


sql/telemetry_derived/clients_first_seen_v1/query.sql
```diff
@@ -42,6 +42,12 @@ SELECT
       NULL
     WHEN
       previous.second_seen_date IS NULL
+      AND today.client_id IS NULL
+    THEN
+      NULL
+    WHEN
+      previous.second_seen_date IS NULL
+      AND today.client_id IS NOT NULL
```
relud, line 50: This is fine to merge as-is, but I think it would be easier to understand what this is doing as an IF statement:
```sql
IF(
  previous.first_seen_date IS NOT NULL
  AND previous.second_seen_date IS NULL
  AND today.client_id IS NOT NULL,
  @submission_date,
  previous.second_seen_date
) AS second_seen_date
```

jklukas: I also updated the first_seen_date computation to be an IF statement for consistency.

PR #1025, opened 2020-06-02
WIP - fission in glam

Ben-Wu:
`clients_histogram_bucket_counts_fission_v1` fails due to shuffle size limit because of the join in the `clients_histogram_aggregates_fission` view.

Scalar queries work fine


Ben-Wu:
Closing because it sounds like data science has another way to do this

PR #1027, opened 2020-06-03
Bug 1633746 Move amo stats queries to amo_prod and amo_dev

jklukas:
Using new datasets set up in https://github.com/mozilla-services/cloudops-infra/pull/2198


jklukas:
I am using copying the DAG generated here to telemetry-airflow in https://github.com/mozilla/telemetry-airflow/pull/1009

dags/bqetl_amo_stats.py
```diff
@@ -0,0 +1,91 @@
...
+    )
+
+    amo_prod__amo_stats_dau__v1.set_upstream(wait_for_main_summary_clients_daily)
+
+    wait_for_main_summary_clients_daily = ExternalTaskSensor(
+        task_id="wait_for_main_summary_clients_daily",
+        external_dag_id="main_summary",
+        external_task_id="clients_daily",
+        dag=dag,
+    )
```
jklukas, line 89: The wait_for_main_summary_clients_daily task is duplicated because two separate queries define this dependency. Opened #1028 to track. I manually removed this block in mozilla/telemetry-airflow#1009

jklukas: It works!

PR #1030, opened 2020-06-03
Use snake_case() and associated tests.

emtwo:
I noticed that some probes were not showing up in Glam like `firstStartup.statusCode` and `firstStartup.elapsed`.  It was because these probes in probe info service are camelCase but in main ping are snake_case and we do an intersection of main ping + probe info service to decide which probes to have in Glam.

I noticed an existing function to do this conversion to snake case and I wanted to reuse it.


bigquery_etl/util/common.py
```diff
@@ -0,0 +1,8 @@
+"""Generic utility functions."""
+import re
+
+
+def convert_camel_case(name):
+    """Convert camel case to snake case."""
+    s1 = re.sub("(.)([A-Z][a-z]+)", r"\1_\2", name)
+    return re.sub("([a-z0-9])([A-Z])", r"\1_\2", s1).lower()
```
acmiyaguchi, line 8: [This is the logic](https://github.com/acmiyaguchi/test-casing/blob/master/src/main.py) that we use in [gcp-ingestion](https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/util/SnakeCase.java) and [jsonschema-transpiler](https://github.com/mozilla/jsonschema-transpiler/blob/master/src/casing.rs) for snake_casing. 

```suggestion
# Search for all camelCase situations in reverse with arbitrary lookaheads.
REV_WORD_BOUND_PAT = re.compile(
    r"""
    \b                                  # standard word boundary
    |(?<=[a-z][A-Z])(?=\d*[A-Z])        # A7Aa -> A7|Aa boundary
    |(?<=[a-z][A-Z])(?=\d*[a-z])        # a7Aa -> a7|Aa boundary
    |(?<=[A-Z])(?=\d*[a-z])             # a7A -> a7|A boundary
    """,
    re.VERBOSE,
)


def snake_case(line: str) -> str:
    """Convert a string into a snake_cased string."""
    # replace non-alphanumeric characters with spaces in the reversed line
    subbed = re.sub(r"[^\w]|_", " ", line[::-1])
    # apply the regex on the reversed string
    words = REV_WORD_BOUND_PAT.split(subbed)
    # filter spaces between words and snake_case and reverse again
    return "_".join([w.lower() for w in words if w.strip()])[::-1]
```

The regex searches for breaks in a camelCased line which are split and joined with underscores. The particular implementation doesn't matter too much though, the important bit is the test suite consisting of [`alphanum_3`, `mps-diff-integration`, and `word_4`](https://github.com/acmiyaguchi/test-casing/tree/master/test-cases) consistent behavior. Could you hook up these tests?

udf/main_summary_scalars.sql.py
```diff
@@ -39,7 +37,7 @@ def collect_probes(probes, schema_fields):
         for p in record_in_processes:
             collection[p].append(
                 (
-                    convert_camel_case(probe["name"]).replace(".", "_"),
+                    snake_case(probe["name"]).replace(".", "_"),
```
acmiyaguchi, line 40: @relud is this replacement safe for main_summary?

relud: should be fine

PR #1031, opened 2020-06-03
Add version to error_aggregates query

scholtzan:
Tooling for generating Airflow DAGs expects queries to have a version. 
To automatically generate the `error_aggregates` Airflow DAG, I added a version to that query.

To ensure nothing breaks, it probably makes sense to create a `error_aggregates` view in `telemetry_derived` to ensure nothing breaks.

Alternatively, having a version could also be made optional. However, since it's recommended practice it might be worth just adding a version to that query. 

See https://github.com/mozilla/telemetry-airflow/pull/1011 for the generated DAG.


scholtzan:
Yeah, I added the `Generated via https://github.com/mozilla/bigquery-etl/blob/master/bigquery_etl/query_scheduling/generate_airflow_dags.py` comment to generated tasks but forgot to update integration tests.

scholtzan:
I copied all data from `error_aggregates` to `error_aggregates_v1`. And also created a view.

PR #1032, opened 2020-06-03
Compare DAG diffs in CI

scholtzan:
It's probably useful to make sure in CI that generated DAGs in the `dags/` directory match with the DAGs that the scripts generate. This way cases where users might manually make changes to the generated DAGs would get flagged.


scholtzan:
Mh, CI would need access to shared-prod for generating the DAGs.

PR #1033, opened 2020-06-04
Add LTV and revenue join query

fbertsch:


fbertsch:
@benmiroglio this is ready for review

udf/parquet_array_sum.sql
```diff
@@ -0,0 +1,11 @@
+/*
+Sum an array from a paquet-derived field.
+These lists of an `element` that represents the field.
```
fbertsch, line 3:
```suggestion
These are lists of an `element` that contain the field value.
```

udf/parquet_array_sum.sql
```diff
@@ -0,0 +1,11 @@
+/*
+Sum an array from a paquet-derived field.
```
fbertsch, line 2:
```suggestion
Sum an array from a parquet-derived field.
```

PR #1034, opened 2020-06-04
Update deviations DAG

scholtzan:
I moved the `anomdtct` pipeline to its own DAG in https://github.com/mozilla/telemetry-airflow/pull/1013


scholtzan:
Depends on https://github.com/mozilla/bigquery-etl/pull/1035 for adding task arguments support

scholtzan:
bigqueryrc sets `--replace=true`, so argument support is not required for this specific query.

PR #1035, opened 2020-06-04
Lower-friction scheduled queries: Task arguments

scholtzan:
Some tasks (see https://github.com/mozilla/telemetry-airflow/pull/1013) required arguments. This adds support for specifying these arguments manually.

### DIFF
--- a/bigquery_etl/query_scheduling/templates/airflow_dag.j2
+++ b/bigquery_etl/query_scheduling/templates/airflow_dag.j2
@@ -30,7 +30,7 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
         {%+ endif -%}
         depends_on_past={{ task.depends_on_past }},
         {%+ if task.arguments | length > 0 -%}
-        arguments={{ task.arguments | format_as_tuple }},
+        arguments={{ task.arguments }},
         {%+ endif -%}
         dag=dag,
     )
--- a/bigquery_etl/query_scheduling/templates/airflow_dag.j2
+++ b/bigquery_etl/query_scheduling/templates/airflow_dag.j2
@@ -29,6 +29,9 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
         date_partition_parameter={{ task.date_partition_parameter | format_optional_string }},
         {%+ endif -%}
         depends_on_past={{ task.depends_on_past }},
+        {%+ if task.arguments | length > 0 -%}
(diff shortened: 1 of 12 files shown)


tests/data/test_sql/test/incremental_query_v1/metadata.yaml
```diff
@@ -11,3 +11,4 @@ labels:
 scheduling:
   dag_name: "bqetl_events"
   depends_on_past: false
+  arguments: ["--append"]
```
scholtzan, line 14: I was considering supporting only a subset of these arguments (based on the queries in telemetry-airflow it seems only a few of these options are used). That way these arguments could be provided without the `--`. But not sure if it's worth doing that.

scholtzan: As it's now all `bq query` flags are supported by default.

bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -29,6 +29,9 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
         date_partition_parameter={{ task.date_partition_parameter | format_optional_string }},
         {%+ endif -%}
         depends_on_past={{ task.depends_on_past }},
+        {%+ if task.arguments | length > 0 -%}
+        arguments={{ task.arguments | format_as_tuple }},
```
jklukas, line 33: Why do we need to pass through format_as_tuple here? We use lists elsewhere.

scholtzan: `bigquery_etl_query` expects a tuple parameter here: https://github.com/mozilla/telemetry-airflow/blob/bb956639405d5ebc8a38a69dac79c689a7d34f6c/dags/utils/gcp.py#L467

jklukas: > `bigquery_etl_query` expects a tuple parameter here: https://github.com/mozilla/telemetry-airflow/blob/bb956639405d5ebc8a38a69dac79c689a7d34f6c/dags/utils/gcp.py#L467

I'd argue that's poor documentation. It looks like we actually just call `list(arguments)` on it later. Any iterable should be allowed.

I'm fine either way. You can simplify by letting this be a list or you can keep format_as_tuple to match the stated contract for the function.

scholtzan: It's always good to keep things simple. List it is.

PR #1036, opened 2020-06-04
Bug 1621703 - Hide deprecated client_id field in default_browser view

jklukas:
Also see https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/557

cc @bytesized


bytesized:
Thanks for this!

PR #1038, opened 2020-06-05
Lower-friction scheduled queries: VR browser DAG

scholtzan:
This migrates VR browser related queries to a separate DAG.

### DIFF
--- a/bigquery_etl/query_scheduling/templates/public_data_json_airflow_dag.j2
+++ b/bigquery_etl/query_scheduling/templates/public_data_json_airflow_dag.j2
@@ -40,6 +40,7 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
         task_id='wait_for_{{ dependency.task_name }}',
         external_dag_id='{{ dependency.dag_name }}',
         external_task_id='{{ dependency.task_name }}',
+        check_existence=True,
     )
         
     {{ task.task_name }}.set_upstream(wait_for_{{ dependency.task_name }})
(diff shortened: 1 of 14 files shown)


bigquery_etl/query_scheduling/templates/public_data_json_airflow_dag.j2
```diff
@@ -40,6 +40,7 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
         task_id='wait_for_{{ dependency.task_name }}',
         external_dag_id='{{ dependency.dag_name }}',
         external_task_id='{{ dependency.task_name }}',
+        check_existence=True,
```
scholtzan, line 43: Added this check to make sure tasks fail if referenced tasks/DAGs do not exist instead of poking for a long time.

sql/org_mozilla_vrbrowser_derived/clients_last_seen_v1/metadata.yaml
```diff
@@ -8,3 +8,5 @@ owners:
 labels:
   application: firefox-reality
   schedule: daily
+scheduling:
+  dag_name: bqetl_vrbrowser
```
jklukas, line 12: Needs depends_on_past=True

sql/org_mozilla_vrbrowser_derived/metrics_daily_v1/metadata.yaml
```diff
@@ -8,3 +8,9 @@ owners:
 labels:
   application: firefox-reality
   schedule: daily
+scheduling:
+  dag_name: bqetl_vrbrowser
+  depends_on:
+    - dag_name: copy_deduplicate
+      task_id: copy_deduplicate_all
+
```
jklukas: Is there something weird going on with whitespace in a few of these files?

PR #1040, opened 2020-06-05
bqetl_core DAG

scholtzan:
Generate DAG for core client queries.


dags/bqetl_core.py
```diff
@@ -0,0 +1,59 @@
...
+
+with DAG("bqetl_core", default_args=default_args, schedule_interval="0 1 * * *") as dag:
+
+    telemetry_derived__core_clients_daily__v1 = bigquery_etl_query(
+        task_id="telemetry_derived__core_clients_daily__v1",
+        destination_table="core_clients_daily_v1",
+        dataset_id="telemetry_derived",
+        project_id="moz-fx-data-shared-prod",
+        owner="jklukas@mozilla.com",
+        email=["jklukas@mozilla.com"],
```
jklukas, line 27: Does `email` get automatically populated based on `owner` in the `metadata.yaml` for the query? It seems we should respect the `email` set in `default_args` instead. Or perhaps merge owner in with the `email` set in default_args. Probably out of scope here.

scholtzan: Currently, `email` for tasks is a set of the owner email plus the emails set in the tasks `scheduling` config. We could think of changing that and use the emails set in the DAG config instead.

PR #1044, opened 2020-06-08
Move firefox_nondesktop_exact_mau28_raw_v1 to telemetry_derived

scholtzan:
Generated DAG for https://github.com/mozilla/telemetry-airflow/pull/1018

Turns out there is already a view `telemetry.firefox_nondesktop_exact_mau28_v1` with DAUs, WAUs and MAUs for tier 1 countries in shared-prod. So leaving the name as is might cause the least amount of confusion.


jklukas:
> Turns out there is already a view telemetry.firefox_nondesktop_exact_mau28_v1 with DAUs, WAUs and MAUs for tier 1 countries in shared-prod. So leaving the name as is might cause the least amount of confusion.

Sorry for not explaining the full context here. Before we had the `_derived` datasets for holding derived tables, `telemetry` was a mix of both tables and views and we started using the `_raw` suffix in the case where we wanted a non-user-facing table in a different format from how we'd present it in a view. Later, once we added the `_derived` datasets, we no longer needed the `_raw` suffix in table names; you would have a table in `_derived` with the raw format, and then a view of the same name in `telemetry` with fields presented as convenient for users.

I want to make sure we stay consistent with the new naming convention as we move things into shared-prod. We can maintain a `telemetry.firefox_nondesktop_exact_mau28_raw_v1` view that's a simple `SELECT *` for the time being in order to ensure we don't break compatibility with anybody who happened to be querying the raw table directly (although I doubt that's happening).

PR #1045, opened 2020-06-08
Add `is_weekday_regular_v1` and `is_allweek_regular_v1`.

felixlawrence:
Add two segments/user states to `clients_last_seen`. Documentation and DS review for these is in a [DTMO PR](https://github.com/mozilla/firefox-data-docs/pull/493).


felixlawrence:
We shouldn't merge this until the DTMO PR is approved (i.e. until these segments get r+ from a DS), but we can get this ready in the meantime - e.g. if there are more straightforward ways to do the date arithmetic

jklukas:
This is pretty readable. I have no suggestions for better alternatives, except perhaps factoring out the logic to a UDF, so that we don't have to repeat all the logic between the weekday and allweek definitions.

felixlawrence:
> factoring out the logic to a UDF

My understanding (based on a comment higher up in the file) was that we can't use UDFs when defining views?

The DTMO PR is r+, so this can be merged once this is also r+.

jklukas:
> My understanding (based on a comment higher up in the file) was that we can't use UDFs when defining views?

That is now outdated. It's true that views can't have _temporary udfs_, but persistent UDFs are allowed.

felixlawrence:
OK cool! Shall I write a separate PR to get it into a UDF?

jklukas:
> OK cool! Shall I write a separate PR to get it into a UDF?

I'm okay with it as-is. But I would be happy to review and merge a new PR that refactors to a persistent UDF.

jklukas:
The updated view is now published.

PR #1047, opened 2020-06-09
Add a view that shows shredder progress

relud:
Adds --task-table to script/shredder_delete


sql/shredder_state/progress/view.sql
```diff
@@ -0,0 +1,147 @@
...
+    SUM(bytes_total) AS bytes_total,
+    SUM(tasks_complete) AS tasks_complete,
+    SUM(tasks_total) AS tasks_total,
+  FROM
+    progress_by_target
+  CROSS JOIN
+    avg_bytes_per_slot_ms
+)
+SELECT
+  (slot_ms_remaining + slot_ms) / slot_ms * 100 AS days_percent_complete,
```
acmiyaguchi, line 140: What's the reasoning behind `(slot_ms_remaining + slot_ms) / slot_ms` being the fraction of the jobs complete?

sql/shredder_state/progress/view.sql
```diff
@@ -0,0 +1,147 @@
...
+  JOIN
+    jobs
+  USING
+    (project_id, job_id)
+  WHERE
+    creation_time > TIMESTAMP_SUB(CURRENT_TIMESTAMP, INTERVAL 28 DAY)
+    AND state = "DONE"
+    AND error_result IS NULL
+),
+progress_by_table AS (
```
acmiyaguchi, line 85: Is this the wrong label?

```suggestion
progress_by_target AS (
```

sql/shredder_state/progress/view.sql
```diff
@@ -0,0 +1,147 @@
...
+  FROM
+    `moz-fx-data-shredder.shredder_state.tasks`
+  JOIN
+    max_end_date
+  USING
+    (end_date)
+  GROUP BY
+    task_id
+),
+jobs AS (
```
acmiyaguchi, line 56: Perhaps a link to https://cloud.google.com/bigquery/docs/information-schema-jobs around here, or the particular reference you used to develop this query.

relud: that's the only reference i used.

sql/shredder_state/progress/view.sql
relud: oh, it's upside down. it should be `slot_ms / (slot_ms_remaining + slot_ms)`. yes, this is completion %, based on the amount of time it is expected to take to finish.

relud:
> Is my understanding of the query correct? Once per run, both state and task information are recorded. The shredder state maintains the list of running jobs, while the task table keeps an estimate of the size of the job. The shredder job state is joined with the job information metadata to get the current state of the job, and the task table is used to determine the percentage of completion.

Yes, except for `Once per run ... state ... information [is] recorded`. State information is only recorded right after each job is submitted to BigQuery. That's why information metadata is needed to filter down to only successfully complete jobs.

PR #1048, opened 2020-06-09
Bug 1628708 FxA Activity and Flow metrics

jklukas:


sql/firefox_accounts_derived/fxa_log_auth_events_v1/query.sql
```diff
@@ -0,0 +1,35 @@
...
+  jsonPayload.fields.utm_content,
+  jsonPayload.fields.utm_medium,
+  jsonPayload.fields.utm_term
+FROM
+  `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_auth_20*`
+WHERE
+  NOT (
+    REGEXP_CONTAINS(jsonPayload.fields.event, r"^email\.(\w+)\.bounced$")
+    OR REGEXP_CONTAINS(jsonPayload.fields.event, r"^email\.(\w+)\.sent$")
+    OR REGEXP_CONTAINS(jsonPayload.fields.event, r"^flow\.complete\.(\w+)$")
```
jklukas, line 22: Kimmy - can you comment on why we exclude these patterns?

sql/firefox_accounts_derived/fxa_log_auth_events_v1/query.sql
```diff
@@ -0,0 +1,35 @@
...
+    'account.confirmed',
+    'account.login',
+    'account.login.blocked',
+    'account.login.confirmedUnblockCode',
+    'account.reset',
+    'account.signed',
+    'account.created',
+    'account.verified',
+    'sms.installFirefox.sent'
+  )
```
jklukas, line 34: Kimmy - can you comment on why we exclude these patterns?

sql/firefox_accounts_derived/fxa_log_content_other_events_v1/query.sql
```diff
@@ -0,0 +1,71 @@
...
+  jsonPayload.country,
+  jsonPayload.entrypoint,
+  jsonPayload.flow_id,
+  jsonPayload.region,
+  jsonPayload.service,
+  jsonPayload.utm_campaign,
+  jsonPayload.utm_content,
+  jsonPayload.utm_medium,
+  jsonPayload.utm_source,
+  -- timestamp, -- This is a static value in 1970 for all records
```
jklukas, line 15: Does it seem reasonable to leave off `timestamp` completely, considering it doesn't seem to be meaningful?

sql/firefox_accounts_derived/fxa_log_content_other_events_v1/query.sql
```diff
@@ -0,0 +1,71 @@
...
+  receiveTimestamp
```
jklukas, line 16: I might argue for renaming `receiveTimestamp` to `timestamp` here since I believe it's the value that logically _should_ be flowing to `timestamp`.

sql/firefox_accounts_derived/fxa_log_content_other_events_v1/query.sql
```diff
@@ -0,0 +1,71 @@
...
+    OR REGEXP_CONTAINS(jsonPayload.event, r"^settings\.clients\.disconnect\.submit\.([a-z]+)$")
+    OR REGEXP_CONTAINS(jsonPayload.event, r"^([\w-]+).verification.clicked$")
+    OR REGEXP_CONTAINS(jsonPayload.event, r"^flow\.signin-totp-code\.success$")
+    OR REGEXP_CONTAINS(jsonPayload.event, r"^flow\.(support)\.success$")
+    OR REGEXP_CONTAINS(jsonPayload.event, r"^flow\.(support)\.view$")
+    OR REGEXP_CONTAINS(jsonPayload.event, r"^flow\.(support)\.fail$")
+    OR REGEXP_CONTAINS(jsonPayload.event, r"^flow\.([\w-]+)\.view$")
+  )
+  AND jsonPayload.event NOT IN (
+    'flow.reset-password.submit', --regex
```
jklukas, line 52: What are these `-- regex` comments about?

jklukas:
cc @ksiegler1 @irrationalagent

sql/firefox_accounts_derived/fxa_log_auth_events_v1/query.sql
ksiegler1: These regexp contain the mappings that are used to exclude flow/auth events that are also Amplitude events. These are excluded to reduce redundancy in the events tables.

sql/firefox_accounts_derived/fxa_log_auth_events_v1/query.sql
ksiegler1: Similar to the regexp mappings, these are explicit matches that are excluded from the table to reduce redundancy in what's stored here and in Amplitude.

sql/firefox_accounts_derived/fxa_log_content_other_events_v1/query.sql
ksiegler1: Sounds good to me.

sql/firefox_accounts_derived/fxa_log_content_other_events_v1/query.sql
ksiegler1: Probably should have removed them, sorry about that. I just noted which events explicitly mapped are also caught by regexp in the same query. So having them caught twice isn't necessary, but I left them anyways because that's how it's laid out in the code with the mappings.

jklukas: I think the comments are helpful, then. I will plan to keep or expand them.

sql/firefox_accounts_derived/fxa_log_content_other_events_v1/query.sql
ksiegler1: I agree with that.

sql/firefox_accounts_derived/fxa_log_auth_events_v1/query.sql
```diff
@@ -0,0 +1,35 @@
+SELECT
+  jsonPayload.fields.event,
+  jsonPayload.fields.user_id,
```
jklukas, line 3: `user_id` is always null in `docker_fxa_auth_*`, so we should probably drop this field.

sql/firefox_accounts_derived/fxa_log_content_other_events_v1/query.sql
jklukas: > that's how it's laid out in the code with the mappings

Can you point me to the code you're referencing? It would be good to link that in a comment in the query text here.

sql/firefox_accounts_derived/fxa_log_auth_events_v1/query.sql
```diff
@@ -0,0 +1,35 @@
+SELECT
+  timestamp,
+  jsonPayload.fields.event,
```
jklukas, line 3: I removed `user_id` since it was always null.

sql/firefox_accounts_derived/fxa_log_auth_events_v1/query.sql
jklukas: Added a comment to capture this discussion.

sql/firefox_accounts_derived/fxa_log_content_performance_events_v1/query.sql
```diff
@@ -0,0 +1,26 @@
...
+    jsonPayload.utm_source,
+  FROM
+    `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_content_19700101`
+)
+SELECT
+  *
+FROM
+  base
+WHERE
+  REGEXP_CONTAINS(event, r"^flow\.performance.[\w-]+")
```
jklukas, line 25: Not that I'm planning to make these table partitioned on `DATE(timestamp)` and clustered on `event`, so it will be efficient to filter to just a single event type.

Would it make sense to collapse performance events and "other" events into a single table? I could also provide a view that filters to just performance events if the point here is to deliver different tables for different audiences.

sql/firefox_accounts_derived/fxa_log_auth_events_v1/query.sql
```diff
@@ -0,0 +1,35 @@
...
+  jsonPayload.fields.flow_id,
+  jsonPayload.fields.entrypoint,
+  jsonPayload.fields.device_id,
```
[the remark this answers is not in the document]
jklukas: Looks like we actually remove device_id, so I'll do that here as well.

sql/firefox_accounts_derived/fxa_log_auth_events_v1/query.sql
```diff
@@ -0,0 +1,35 @@
...
+      'account.login.blocked',
+      'account.login.confirmedUnblockCode',
+      'account.reset',
+      'account.signed',
+      'account.created',
+      'account.verified',
+      'sms.installFirefox.sent'
+    )
+  )
+  AND _TABLE_SUFFIX = FORMAT_DATE('%g%m%d', @submission_date)
```
scholtzan, line 35: RemindMe! 80 years ;)

PR #1050, opened 2020-06-09
Rename firefox_nondesktop_exact_mau28_raw_v1 to firefox_nondesktop_exact_mau28_v1

scholtzan:
Follow-up to https://github.com/mozilla/bigquery-etl/pull/1044#issuecomment-640956674


jklukas:
There should be some view updates needed here too.

scholtzan:
The view definitions reference the views in derived-datasets. Those will need to be (manually) updated though (since they reference shared-prod) before merging this and the updated Airflow DAGs.

jklukas:
I'm having trouble reasoning through the views and how they'll be updated here. I think we should update the `nondesktop_*` views as part of this PR to all reference shared-prod. We publish these views to both derived-datasets and shared-prod, but they should all reference shared-prod now since that's where the data lives.

scholtzan:
Updated the views to reference tables/views in shared-prod for those that already exist there.
I copied data from `moz-fx-data-shared-prod.telemetry_derived.firefox_nondesktop_exact_mau28_raw_v1` over to `moz-fx-data-shared-prod.telemetry_derived.firefox_nondesktop_exact_mau28_v1` and updated views in derived-datasets.

PR #1055, opened 2020-06-09
Add user_id to amplitude events that only had device_id

relud:


fbertsch:
@relud, I'm waiting for this review until we hear back from Amplitude in the latest email.

fbertsch:
Update, we're meeting with Amplitude on Thursday about this.

relud:
we found a different solution to this

PR #1056, opened 2020-06-10
Bug 1643683 Add Fenix DAU query for AMO stats

jklukas:


sql/amo_prod/amo_stats_fenix_dau_v1/query.sql
```diff
@@ -0,0 +1,181 @@
+WITH cd AS (
+  SELECT
+    DATE(submission_timestamp) AS submission_date,
+    client_info.client_id,
+    array_concat_agg(metrics.string_list.addons_enabled_addons) AS active_addons,
+    udf.mode_last(array_agg(client_info.app_build)) AS app_version,
+    udf.mode_last(array_agg(normalized_country_code)) AS country,
+    udf.mode_last(array_agg(client_info.locale)) AS locale,
+    udf.mode_last(array_agg(normalized_os)) AS app_os,
```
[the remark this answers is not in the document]
willdurand: What will be the OS for Fenix? `Android`?

sql/amo_prod/amo_stats_fenix_dau_v1/query.sql
```diff
@@ -0,0 +1,181 @@
...
+  FROM
+    org_mozilla_fenix_nightly_stable.metrics_v1
```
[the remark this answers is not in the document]
willdurand: What are Fenix app_ids?

willdurand:
Looks good to me. From what I can see here and also based on [your comment in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1643683), these changes should be backward compatible, right?

sql/amo_prod/amo_stats_fenix_dau_v1/query.sql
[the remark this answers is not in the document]
jklukas: Correct, os will be "Android".

sql/amo_prod/fenix_addons_by_client_v1/query.sql
```diff
@@ -0,0 +1,82 @@
...
+  UNION ALL
+  SELECT
+    get_fields(m4).*
+  FROM
+    org_mozilla_firefox.metrics AS m4
+  UNION ALL
+  SELECT
+    get_fields(m5).*
+  FROM
+    org_mozilla_firefox_beta.metrics AS m5
```
jklukas, line 45: > What are Fenix app_ids?

They're roughly equivalent to the concept of "channel" in desktop Firefox. The new Glean telemetry library sends data to different BQ tables based on the application ID in the build, which also determines how it's listed in the Play Store. What we call "Fenix" internally corresponds to 5 distinct apps in the Play store: Firefox Preview, and Firefox Preview Nightly, and then the three channels for users migrated from Fennec (nightly, beta, and release).

The union here is how we pull data from all channels. The data we show for desktop also includes all release channels.

sql/amo_prod/amo_stats_fenix_dau_v1/query.sql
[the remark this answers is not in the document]
jklukas: > What are Fenix app_ids?

See comment on the union below. I actually need to delete this file as it's no longer relevant with the refactored logic.

jklukas:
> Looks good to me. From what I can see here and also based on your comment in the bug, these changes should be backward compatible, right?

Yes, that's the intention. We maintain all the existing fields, and simply add one new field for the per-fenix-build breakdown. We can merge these changes and backfill without breaking the AMO stats page. Then you can proceed to add support for displaying data encoded in the new dau_by_fenix_build field.

sql/amo_dev/amo_stats_dau/view.sql
```diff
@@ -26,6 +26,13 @@ SELECT
     FROM
       UNNEST(dau_by_app_version)
   ) AS dau_by_app_version,
+  ARRAY(
+    SELECT AS STRUCT
+      IFNULL(key, 'Unknown') AS key,
+      value
+    FROM
+      UNNEST(dau_by_app_version)
```
scholtzan, line 34:
```suggestion
      UNNEST(dau_by_fenix_build)
```

sql/amo_prod/amo_stats_dau/view.sql
```diff
@@ -26,6 +26,13 @@ SELECT
     FROM
       UNNEST(dau_by_app_version)
   ) AS dau_by_app_version,
+  ARRAY(
+    SELECT AS STRUCT
+      IFNULL(key, 'Unknown') AS key,
+      value
+    FROM
+      UNNEST(dau_by_app_version)
```
scholtzan, line 34:
```suggestion
      UNNEST(dau_by_fenix_build)
```

sql/amo_prod/fenix_addons_by_client_v1/query.sql
```diff
@@ -0,0 +1,75 @@
+CREATE TEMP function get_fields(m ANY TYPE) AS (
```
scholtzan, line 1:
```suggestion
CREATE TEMP FUNCTION get_fields(m ANY TYPE) AS (
```

Surprising that the formatting check didn't complain.

sql/amo_prod/fenix_addons_by_client_v1/query.sql
```diff
@@ -0,0 +1,75 @@
...
+  FROM
+    org_mozilla_firefox_beta.metrics AS m5
+),
+per_client AS (
+  SELECT
+    DATE(submission_timestamp) AS submission_date,
+    client_id,
+    sample_id,
+    array_concat_agg(addons_enabled_addons) AS addons,
+    udf.mode_last(array_agg(app_build)) AS app_version,
```
scholtzan, line 46: Just to double check, this will use the app build ID while for `desktop_addons_by_client_v1` it uses the version.

jklukas: Yes, it's my understanding that `app_build` is the Glean/Fenix equivalent of app_version. We put it in a field named `app_version` for schema compatibility with the desktop table.

jklukas, line 46:
```suggestion
    -- app_build is the Fenix equivalent of app_version; we use app_version as the name for
    -- compatibility with the desktop table schema.
    udf.mode_last(array_agg(app_build)) AS app_version,
```

willdurand: @jklukas looking at some production data, `app_build` seems to be the build ID (like `2015752523`), which is kinda mysterious for end users. [This patch](https://hg.mozilla.org/integration/autoland/rev/46949083a8c8) shows a `app_display_version` field and I am wondering if this could be something we could used instead?

I haven't tried to query any data yet, so that's just a question.

jklukas: That's definitely worth investigating. If `app_display_version` is a more useful quantity for you, I'm happy to use that instead.

willdurand: Let's revisit later (when Fenix will be GA).

PR #1062, opened 2020-06-10
Move normalized search engine to views instead of tables

Ben-Wu:
Related bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1644520

Put the normalized value in the view so that we can change the normalized mappings without needing a backfill.  The value in the table will just be null.  `search_clients_last_seen` will be backfilled in the future.

Also removed the search table exclusion in dryrun since the search dataset isn't restricted anymore.


Ben-Wu:
I don't think removing a nullable column from the query should cause issues regardless of options.  Adding a column does depend on options though

PR #1064, opened 2020-06-11
mozfun setup

scholtzan:
This PR sets up the `mozfun` project with a couple of UDFs to be published. It also adds a script for publishing UDFs to `mozfun` by reusing some of the existing machinery for publishing persistent UDFs.

The structure of how UDFs are stored in the `mozfun` directory is different from the way they are stored in `udf` and `udf_js`. I think ideally, in the future `mozfun` and `udf`/`udf_js` should be structured in the same way: UDFs are stored in separate directories named after the UDF name containing a `udf.sql` file and optionally `metadata.yaml` file. That way we could consolidate the machinery for publishing UDFs and have just a single script.


mozfun/map/get_key/udf.sql
```diff
@@ -0,0 +1,14 @@
...
+of structs instead, and this function provides map-like access to such fields.
+
+*/
+CREATE OR REPLACE FUNCTION map.get_key(map ANY TYPE, k ANY TYPE) AS (
+  (SELECT key_value.value FROM UNNEST(map) AS key_value WHERE key_value.key = k LIMIT 1)
+);
+
+-- Tests
+SELECT
+  assert_equals(12, map.get_key([STRUCT('foo' AS key, 42 AS value), ('bar', 12)], 'bar'));
```
scholtzan, line 14: Having this UDF would be immediately useful for pensieve

bigquery_etl/parse_udf.py
```diff
@@ -95,6 +113,16 @@ def from_text(text, dataset, name, filepath=None, is_defined=True):
...
 
+        if filepath:
+            # for public UDFs dependencies can live in arbitrary dataset;
+            # we can check if some known dependency is part of the UDF
+            # definition instead
+            _, basename = os.path.split(filepath)
+            if basename == "udf.sql":
+                for udf in MOZFUN_UDFS:
+                    if udf in "\n".join(definitions):
+                        dependencies.append(udf)
```
scholtzan, line 124: In mozfun UDFs don't necessarily live in either `udf` or `udf_js` datasets but in some arbitrary dataset. The regexes we used before to determine dependencies therefore won't work. 

Instead, we can check if a UDF that is defined in the `mozfun/` directory is a dependency. 

One small disadvantage of this is that UDFs that are not undefined stay undetected. However, when deploying UDFs with undefined dependencies, BigQuery will throw an error.

bigquery_etl/parse_udf.py
```diff
@@ -14,20 +14,32 @@
 
 
 UDF_DIRS = ("udf", "udf_js")
+MOZFUN_DIR = ("mozfun",)
```
scholtzan, line 17: Once we consolidate how UDFs are defined in mozfun and in udf/udf_js, this script can get simplified and refactored.

PR #1066, opened 2020-06-11
Create bqetl_activity_stream DAG

scholtzan:
Scheduling `impression_stats_flat_v1` query so that it can moved out of `copy_deduplicate`.

Generates https://github.com/mozilla/telemetry-airflow/pull/1030

### DIFF
--- a/sql/activity_stream_bi/impression_stats_flat_v1/metadata.yaml
+++ b/sql/activity_stream_bi/impression_stats_flat_v1/metadata.yaml
@@ -1,0 +1,12 @@
+friendly_name: Impression Stats Flat
+description: Flattened impression stats.
+owners:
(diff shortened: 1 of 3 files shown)


sql/activity_stream_bi/impression_stats_flat_v1/metadata.yaml
```diff
@@ -0,0 +1,12 @@
+friendly_name: Impression Stats Flat
+description: Flattened impression stats.
```
scholtzan, line 2: Not sure about this description...

jklukas, line 2:
```suggestion
description: Unnested representation of tile impression statistics
```

PR #1070, opened 2020-06-12
Bug 1633918 View into CDP tables for email click export to Amplitude

jklukas:


sql/firefox_accounts/fxa_amplitude_email_clicks/view.sql
```diff
@@ -0,0 +1,220 @@
...
+    ) AS value
+  FROM
+    `moz-fx-data-shared-prod.firefox_accounts_derived.encrypted_keys_v1`
+  WHERE
+    key_id = 'fxa_hmac_prod'
+),
+sendjobs_numbered AS (
+  SELECT
+    *,
+    ROW_NUMBER() OVER (PARTITION BY SendID ORDER BY snapshot_date DESC) AS _n
```
fbertsch, line 19: Are there docs or context for these tables? This looks to me like this query would use all of history every time, but I'm not sure how bad that is here.

sql/firefox_accounts/fxa_amplitude_email_clicks/view.sql
```diff
@@ -0,0 +1,220 @@
...
+    * EXCEPT (_n)
+  FROM
+    sendjobs_numbered
+  WHERE
+    _n = 1
+),
+customers_numbered AS (
+  SELECT
+    *,
+    ROW_NUMBER() OVER (PARTITION BY email ORDER BY fxa_ts DESC) AS _n
```
fbertsch, line 34: Same question as above.

sql/firefox_accounts/fxa_amplitude_email_clicks/view.sql
```diff
@@ -0,0 +1,220 @@
...
+  FROM
+    `mozilla-cdp-prod.fxa.customer_record`
+),
+customers_deduped AS (
+  SELECT
+    * EXCEPT (_n),
+    IF(
+      net.public_suffix(email) IS NULL,
+      NULL,
+      rtrim(
```
fbertsch, line 44: A comment on what this line is doing would be useful.

sql/firefox_accounts/fxa_amplitude_email_clicks/view.sql
```diff
@@ -0,0 +1,220 @@
...
+              ),
+              STRUCT(
+                'email_region' AS key,
+                REGEXP_EXTRACT(
+                  sendjobs.EmailName,
+                  r"(?:[A-Za-z]+)_([A-Z]+)_[A-Z]*_[0-9]{4}_[A-Z]+_(?:[A-Z]+|DESK[_ ][A-Z]+)_(?:.+?)_(?:ALL|[A-Z]{2})_(?:[A-Z]{2,4})_(?:[A-Z-]+)(?:_[A-Za-z0-9]*)?"
+                ) AS value
+              ),
+              STRUCT(
+                'email_focus' AS key,
```
fbertsch, line 153: This one is new?

sql/firefox_accounts/fxa_amplitude_email_clicks/view.sql
```diff
@@ -0,0 +1,220 @@
...
+              STRUCT('email_alias' AS key, clicks.alias AS value),
+              STRUCT('email_type' AS key, clicks.EventType AS value),
+              -- sendjobs.EmailName is parsed via regex; each regex differs only
+              -- in which group is doing the capturing (no ?: prefix);
+              -- see original regex in https://github.com/mozilla/fxa-amplitude-send/blob/89bfaef20a1d978fce2dccddc0155699a17ac172/marketing.js#L109
+              STRUCT(
+                'email_sender' AS key,
+                REGEXP_EXTRACT(
+                  sendjobs.EmailName,
+                  r"([A-Za-z]+)_(?:[A-Z]+)_[A-Z]*_[0-9]{4}_[A-Z]+_(?:[A-Z]+|DESK[_ ][A-Z]+)_(?:.+?)_(?:ALL|[A-Z]{2})_(?:[A-Z]{2,4})_(?:[A-Z-]+)(?:_[A-Za-z0-9]*)?"
```
fbertsch, line 142: We could use regex_replace for this: `SPLIT(REGEX_REPLACE(regex_string, "\\1,\\2,..."))` which would allow us to have this regex string just once.

sql/firefox_accounts/fxa_amplitude_email_clicks/view.sql
```diff
@@ -0,0 +1,220 @@
...
+    email_providers_over_10k.email_provider,
+  FROM
+    customers_deduped
+  LEFT JOIN
+    email_providers_over_10k
+  USING
+    (email_provider)
+)
+SELECT
+  UNIX_MILLIS(clicks.EventDate) AS `time`,
```
fbertsch, line 82: The amplitude operator looks for a `submission_timestamp` field, in order to just pull in the most recent events. If that's the `time` field here then we can add add it also as `submission_timestamp`.

sql/firefox_accounts/fxa_amplitude_email_clicks/view.sql
jklukas: Access is fairly locked down to these tables and to the documentation around them (there's a google doc). The tables are quite small, and yes this is windowing over all history in the table. I will add some comments to explain that context.

sql/firefox_accounts/fxa_amplitude_email_clicks/view.sql
jklukas: Yes, I'm including it here for completeness, but it's a judgment call. I could be convinced to drop and keep it parallel with the pipeline it's replacing.

sql/firefox_accounts/fxa_amplitude_email_clicks/view.sql
```diff
@@ -0,0 +1,190 @@
...
+See https://bugzilla.mozilla.org/show_bug.cgi?id=1633918
+
+*/
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.firefox_accounts.fxa_amplitude_email_clicks`
+AS
+WITH hmac_key AS (
+  SELECT
+    AEAD.DECRYPT_BYTES(
+      (SELECT keyset FROM `moz-fx-dataops-secrets.airflow_query_keys.fxa_prod`),
```
jklukas, line 12: I'm realizing now we most likely won't be able to publish this as a view due to permissions. Airflow is able to query this table in dataops-secrets and the tables in the CDP project, but we have no guarantee that Jenkins will have those same permissions. I believe view publishing will fail without the user creating the view having access to the target tables.

jklukas: I think I'll exempt this from publishing, and then I'll have the Airflow DAG handle running this query file each time before querying it.

jklukas: But that won't work because the `bigquery-etl` stuff executes in kube with different permissions than the operators we use in the Amplitude export DAGs...

PR #1071, opened 2020-06-12
DAG for addon queries

scholtzan:
Continuing the migration of queries to generate Airflow DAGs.

Generates https://github.com/mozilla/telemetry-airflow/pull/1036

---

@benmiroglio For some more context: we added a mechanism that allows to generate Airflow DAGs for queries based on metadata specified in metadata.yaml files. The mechanism also automatically determines query dependencies and adds those to the generated DAG. In the near future, telemetry-airflow will automatically pick up DAGs that are generated in this repository making. That way it won't be necessary to manually write DAGs in telemetry-airflow anymore.

Currently, we are manually migrating queries to make sure nothing breaks before switching to a fully automated approach.


sql/telemetry_derived/addon_aggregates_v2/metadata.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - bmiroglio@mozilla.com
+labels:
+  application: addons
+  schedule: daily
+  incremental: true
+scheduling:
+  dag_name: bqetl_addons
+  depends_on:
+    - dag_name: main_summary
+      task_id: copy_deduplicate_main_ping
```
jklukas, line 15: I think it would be worth at this point adding support in the dag generation machinery for identifying dependencies on stable tables and injecting these dependencies on copy_dedupe tasks automatically, so that we can avoid `depends_on` entries. We're definitely going to accidentally miss adding `depends_on` at some point.

To be clear, this is a suggested follow-up. I opened https://github.com/mozilla/bigquery-etl/issues/1089 for this.

PR #1073, opened 2020-06-13
Typo in fxa_log_content query

jklukas:


jklukas:
Temporarily suspended merge restrictions so that I could merge this and avoid a bunch of failure emails over the weekend.

PR #1079, opened 2020-06-16
Remove search aggregate union with v6

Ben-Wu:
Search aggregates is currently being unioned with the old v6 data for the armag.addon period.  The changes in #1062 broke the view silently because the columns were in different order.  The existing union is unreliable because columns must be in the same order.  To make that guarantee we we would have to explicitly select every column in both `SELECT`s.  Instead, we can just backfill the current table using the v6 table and not have to worry about it.


Ben-Wu:
Thanks.  Backfill is done

PR #1082, opened 2020-06-18
Add shredder script to forward deletion requests to amplitude

relud:


relud:
> Would be good to clarify how you expect the api and secret key to be handled - I could see them as either part of the deploy of the container or as Airflow variables.

Definitely as Airflow variables, like all the other credentials.

PR #1084, opened 2020-06-22
Remove queries that write to derived-datasets.telemetry

jklukas:
The kpi_dashboard query is out of date; the 2020 dashboard is implemented in
Databricks and performs a modified version of the query logic.
We remove this table completely and we will send out a fx-data-dev email
to that effect.

Otherwise, the desktop exact mau table was the last scheduled query writing
to the telemetry dataset in the derived-datasets project.


jklukas:
I'll need to copy the tables into place and make sure the dry runs succeed before merging and committing to the change.

PR #1086, opened 2020-06-23
DAGs for client queries

scholtzan:
Depends on https://github.com/mozilla/bigquery-etl/pull/1071

Generates https://github.com/mozilla/telemetry-airflow/pull/1047

This PR adds 3 DAGs: `bqetl_clients`, `bqetl_desktop` and `bqetl_devtools` which allow to move queries out of the `main_summary` DAG.


dags.yaml
```diff
@@ -116,6 +116,33 @@ bqetl_addons:
     retries: 2
     retry_delay: 30m
     
+bqetl_clients:
```
jklukas, line 119: I don't think "clients" is specific enough here. I'd be inclined to call this `bqetl_desktop_clients_daily` or simply `bqetl_clients_daily` and then the following DAG would be `bqetl_desktop_exact_mau`.

I think it's reasonable to include clients_last_seen under the "clients_daily" label since it is a direct extension of daily aggregations per client.

I also feel like perhaps we should keep the exact_mau queries together with the clients_daily DAG. I feel like the exact_mau tables are simply optimized views on top of clients_last_seen, and since KPIs are served from exact_mau tables it feels like a nice level of safety that reprocessing early stages would automatically trigger reprocessing of these downstream tasks in the same DAG.

jklukas: As I write this, I also am realizing that these exact_mau tables are perfect fits for BQ's new [materialized view](https://cloud.google.com/bigquery/docs/materialized-views-intro) feature since they are purely COUNT and SUM aggregations on top of a base table. I will write up a separate issue for that.

jklukas: See https://github.com/mozilla/bigquery-etl/issues/1090

scholtzan: That sounds reasonable. I moved the clients_daily and exact_mau queries to the DAG `bqetl_clients_daily` for now.

scholtzan:
I'll create a follow-up PR for the changes required for https://github.com/mozilla/telemetry-airflow/pull/1047#pullrequestreview-436874178. This will hopefully make reviewing easier.

PR #1094, opened 2020-06-24
bqetl_main_summary DAG

scholtzan:
Adds support for scheduling multipart queries and moves queries out of the `main_summary` DAG.

Generates https://github.com/mozilla/telemetry-airflow/pull/1047

### DIFF
--- a/bigquery_etl/query_scheduling/task.py
+++ b/bigquery_etl/query_scheduling/task.py
@@ -2,6 +2,8 @@
 
 import attr
 import cattr
+import glob
+import os
 import re
 import logging
 from google.cloud import bigquery
@@ -18,7 +20,9 @@ from bigquery_etl.query_scheduling.utils import (
 
 
 AIRFLOW_TASK_TEMPLATE = "airflow_task.j2"
-QUERY_FILE_RE = re.compile(r"^.*/([a-zA-Z0-9_]+)/([a-zA-Z0-9_]+)_(v[0-9]+)/query\.sql$")
+QUERY_FILE_RE = re.compile(
+    r"^.*/([a-zA-Z0-9_]+)/([a-zA-Z0-9_]+)_(v[0-9]+)/(?:query\.sql|part1\.sql)$"
+)
 DEFAULT_PROJECT = "moz-fx-data-shared-prod"
 
 
@@ -95,6 +99,8 @@ class Task:
     depends_on: List[TaskRef] = attr.ib([])
     arguments: List[str] = attr.ib([])
     parameters: List[str] = attr.ib([])
+    multipart: bool = attr.ib(False)
+    sql_file_path: Optional[str] = None
 
     @owner.validator
     def validate_owner(self, attribute, value):
@@ -149,7 +155,7 @@ class Task:
         else:
             raise ValueError(
                 "query_file must be a path with format:"
-                " ../<dataset>/<table>_<version>/query.sql"
+                " ../<dataset>/<table>_<version>/(query.sql|part1.sql)"
                 f" but is {self.query_file}"
             )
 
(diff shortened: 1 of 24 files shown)


dags/bqetl_experiments.py
```diff
@@ -0,0 +1,45 @@
...
+    "email": ["telemetry-alerts@mozilla.com", "ssuh@mozilla.com"],
+    "depends_on_past": False,
+    "retry_delay": datetime.timedelta(seconds=1800),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
+}
+
+with DAG(
+    "bqetl_experiments", default_args=default_args, schedule_interval="0 1 * * *"
```
scholtzan, line 20: I'm not sure if `bqetl_experiments` is a too general of a name for this DAG. Currently, it has only 1 task, but it's possible that more experiment related queries get added in the future.

sql/amo_prod/desktop_addons_by_client_v1/metadata.yaml
```diff
@@ -12,4 +12,4 @@ scheduling:
   dag_name: bqetl_amo_stats
   depends_on:
     - task_id: copy_deduplicate_main_ping
-      dag_name: main_summary
+      dag_name: copy_deduplicate
```
scholtzan, line 15: Moved to `copy_deduplicate` in https://github.com/mozilla/telemetry-airflow/pull/1047

dags/bqetl_experiments.py
jklukas: Maybe `bqetl_experiments_daily`?

tests/query_scheduling/test_task.py
```diff
@@ -397,6 +414,66 @@ def test_task_get_multiple_dependencies(
         assert f"{temporary_dataset}__table1__v1" in tables
         assert f"{temporary_dataset}__table2__v1" in tables
 
+    @pytest.mark.integration
+    def test_muultipart_task_get_dependencies(
```
jklukas, line 418:
```suggestion
    def test_multipart_task_get_dependencies(
```

dags/bqetl_main_summary.py
```diff
@@ -0,0 +1,54 @@
...
+        destination_table="main_summary_v4",
+        dataset_id="telemetry_derived",
+        project_id="moz-fx-data-shared-prod",
+        owner="dthorn@mozilla.com",
+        email=["dthorn@mozilla.com"],
+        start_date=datetime.datetime(2019, 10, 25, 0, 0),
+        date_partition_parameter="submission_date",
+        depends_on_past=False,
+        mulitpart=True,
+        sql_file_path="sql/telemetry_derived/main_summary_v4",
```
jklukas, line 39: The existing task this is replacing has a trailing `/` here. Are we certain the query running script handles this gracefully?

bigquery_etl/query_scheduling/task.py
```diff
@@ -219,23 +239,33 @@ def _get_referenced_tables(self, client):
...
+                    )
+
+                for t in referenced_tables:
+                    table_names.add((t.dataset_id, t.table_id))
+
+        # the order of table dependencies changes between requests
+        # sort to maintain same order between DAG generation runs
+        sorted_table_names = list(table_names)
+        sorted_table_names.sort()
+        return sorted_table_names
```
relud, line 268:
```suggestion
        return sorted(table_names)
```

dags/bqetl_main_summary.py
relud: this value is passed to `os.path.join` and `os.walk`, so losing the trailing `/` is fine.

scholtzan:
I've looked a bit into the alerting problems. I found some open issues related to the problem: https://issues.apache.org/jira/browse/AIRFLOW-5071 and possibly https://issues.apache.org/jira/browse/AIRFLOW-5074
Other than increasing the number of slots or core parallelism those don't provide a good solution though.

PR #1097, opened 2020-06-25
Fix view regressions

jklukas:
Fixes several view regressions introduced in https://github.com/mozilla/bigquery-etl/pull/1092


jklukas:
I've published these for the time being, since the events view regression was breaking mozanalysis, as reported in `#fx-metrics`.

PR #1102, opened 2020-06-29
Generate documentation for mozfun

scholtzan:
This adds a script for generating Markdown files that can be rendered by mkdocs. I went with mkdocs for now, since it is relatively easy to set up but I am also open to alternatives. 

I haven't added a CircleCI step yet as I am not sure if mozfun specific documentation should be hosted as part of this repository, somewhere else or if we want to generate documentation also for other bigquery-etl parts (datasets? UDFs from other projects? ...) and mozfun is just part of the documentation.

As a next step, I'll look into validating examples.


jklukas:
I think definitely we should generate docs specific to this repo, probably on a gh-pages branch. And it would be worth making it somewhat generic so that we can add additional docs beyond mozfun. So I'd vote that mozfun docs show up under a subdirectory, and perhaps we have a top-level index page that links to the function docs, links to the repo itself, and mentions that rendering of additional documentation about tables, etc. may come in the future.

wlach:
> I think definitely we should generate docs specific to this repo, probably on a gh-pages branch. And it would be worth making it somewhat generic so that we can add additional docs beyond mozfun. So I'd vote that mozfun docs show up under a subdirectory, and perhaps we have a top-level index page that links to the function docs, links to the repo itself, and mentions that rendering of additional documentation about tables, etc. may come in the future.

This makes sense to me too. Longer term I'd like to think about moving (at least) the table definitions for derived datasets to the [schema dictionary](https://mozilla-schema-dictionary.netlify.app/) but I think this is a great stop-gap as it will take a while to get that working. 

mkdocs is a good initial choice for this repo. It's probably the simplest of all documentation systems with which to get up and running, at least for python-based projects like this one.

.circleci/config.yml
```diff
@@ -106,6 +106,20 @@ jobs:
...
             cd telemetry-airflow
             bash bin/test-parse
+  docs:
+    docker: *docker
+    steps:
+    - checkout
+    - *build
+    - run:
+        name: Install dependencies
+        command: pip install mkdocs markdown-include
```
scholtzan, line 116: We'll want to have a step here to validate doc examples. I'll do it in a separate PR since this is already getting bloated.

bigquery_etl/docs/generate_docs.py
```diff
@@ -0,0 +1,147 @@
...
+    help="Directories of projects documentation is generated for.",
+    nargs="+",
+    default=DEFAULT_PROJECTS,
+)
+parser.add_argument(
+    "--docs_dir",
+    "--docs-dir",
+    default=DOCS_DIR,
+    help="Directory containing static documentation.",
+)
```
scholtzan, line 31: My thinking here: We might have some static documentation in `docs/` and want to merge it with the generated documentation.

bigquery_etl/docs/generate_docs.py
```diff
@@ -0,0 +1,147 @@
...
+
+                        # parse the doc directory structure
+                        # used in mkdocs.yml
+                        path_parts = root.split(os.sep)
+                        config = dir_structure
+
+                        for part in path_parts:
+                            config = config.setdefault(part, {})
+
+    # convert directory structure to mkdocs compatible format
```
scholtzan, line 86: Generating the `mkdocs.yml` based on the directory structure is where it gets hairy :(
In some cases, each UDF might have its own documentation in addition to the dataset documentation while in some cases there might just be documentation on a dataset level:

```yaml
nav:
  - project_name:
     - dataset_name1: 
        - Overview: path/to/index.md
        - udf1: path/to/index.md
        - udf2: path/to/index.md
      - dataset_name2: path/to/index.md
```

scholtzan:
I updated the PR to make it possible to have static documentation for the repo and merge it with generated documentations. I also added a step to CircleCI for pushing the docs to the `gh-pages` branch.

bigquery_etl/docs/generate_docs.py
scholtzan: Turns out mkdocs auto-generates navigation when not specified. So this is not needed.

scholtzan:
Currently, the docs would look like this:

Screen Shot 2020-06-30 at 20 09 40

jklukas:
> Alternatively, it could make sense to merge all documentation of a dataset into one file, even if there are UDF-specific doc files. However, this might result in very long web pages if there is a lot of documentation in one dataset.

I think part of creating a good user interface here is to keep any individual dataset from getting out of hand in terms of number of functions. My gut says we should have one page per dataset with dataset-level documentation at the top, and then function-specific docs following.

PR #1108, opened 2020-06-30
Add script to backfill glam tables for a glean product

acmiyaguchi:
This adds a script that will backfill a table for a glean product. It doesn't not check if the data has already been processed -- this means that it is possible to aggregate the same date twice. However, if the dates are being properly tracked in a log, then this should do the right amount of work to bring a table up to a current state.

### DIFF
--- a/script/glam/backfill_glean
+++ b/script/glam/backfill_glean
@@ -1,0 +1,35 @@
+#!/bin/bash
+# example: START_DATE=2020-04-01 END_DATE=2020-07-01 backfill_glean
+
+set -ex
+
+function ds_range {
+    DS_START=$1 DS_END=$2 python3 - <<EOD
+from datetime import date, timedelta, datetime
+from os import environ
+def parse(ds):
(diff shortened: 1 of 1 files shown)


script/glam/backfill_glean
```diff
@@ -0,0 +1,41 @@
...
+#   PRODUCT=org_mozilla_fenix \
+#       backfill_glean
+
+set -ex
+
+function ds_range {
+    DS_START=$1 DS_END=$2 python3 - <<EOD
+from datetime import date, timedelta, datetime
+from os import environ
+def parse(ds):
```
Ben-Wu, line 14: Can also just do `datetime.fromisoformat`

script/glam/backfill_glean
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+# example:
```
Ben-Wu, line 2: Script name is missing? Or is that intentional?

acmiyaguchi: Actually the script name starts after the environment variables

PR #1114, opened 2020-07-01
Replace null keys with 'Unknown' in amo_stats_dau

jklukas:
See https://github.com/mozilla/addons-server/issues/14790


willdurand:
@jklukas Thanks. Did you run a backfill yet? (I am still seeing `null` values in -dev)

jklukas:
It would normally get published overnight, but there may have been a hiccup. I just manually published the views, so you should no longer see null values in dev or prod.

willdurand:
> It would normally get published overnight, but there may have been a hiccup. I just manually published the views, so you should no longer see null values in dev or prod.

I confirm it works now, thanks!

PR #1116, opened 2020-07-01
Fix mozfun udf reference

jklukas:


jklukas:
I temporarily suspended restrictions so that I could go ahead and merge this. Retroactive r? would be appreciated.

PR #1117, opened 2020-07-01
mozfun.histogram.extract cleanup

jklukas:
Follow-up to #1000 now that the function is published and I've had some chance to use it.


mozfun/hist/extract/udf.sql
```diff
@@ -98,6 +98,36 @@ SELECT
 FROM
   extracted;
 
+-- We test a histogram with a small values array along with a null histogram;
+-- when there is at least one null row, something about the behavior changes such
+-- that use of unsafe OFFSET indexing can raise errors, even when contained under
```
scholtzan, line 103: I couldn't reproduce this issue. But it's probably good to use `SAFE_OFFSET` here in any case

PR #1120, opened 2020-07-02
Reduce complexity of getting histogram sum in main_summary

### LINKED ISSUE #1119: main_summary_v4 dry run failing as too complex
Recent PRs are failing dry runs with:

>  [{'code': 400, 'errors': [{'message': 'Resources exceeded during query execution: Not enough resources for query planning - too many subqueries or query is too complex.', 'domain': 'global', 'reason': 'resourcesExceeded'}]

My guess is that it's due to https://github.com/mozilla/bigquery-etl/pull/1117 which increases the complexity of the json_extract_histogram udf and adds an additional layer of indirection, calling out to the mozfun version of the function.

My inclination is to provide a new udf specifically for extracting `sum` from a histogram, vastly reducing the necessary complexity for the function, but still meeting the use case here.

jklukas:
Fixes #1119

### DIFF
--- a/sql/telemetry_derived/main_summary_v4/part1.sql
+++ b/sql/telemetry_derived/main_summary_v4/part1.sql
@@ -112,20 +112,20 @@ SELECT
   payload.info.reason,
   payload.info.timezone_offset,
   -- Different types of crashes / hangs; format:off
-  udf.json_extract_histogram(udf.get_key(payload.keyed_histograms.subprocess_crashes_with_dump, 'pluginhang')).sum AS plugin_hangs,
-  udf.json_extract_histogram(udf.get_key(payload.keyed_histograms.subprocess_abnormal_abort, 'plugin')).sum AS aborts_plugin,
-  udf.json_extract_histogram(udf.get_key(payload.keyed_histograms.subprocess_abnormal_abort, 'content')).sum AS aborts_content,
-  udf.json_extract_histogram(udf.get_key(payload.keyed_histograms.subprocess_abnormal_abort, 'gmplugin')).sum AS aborts_gmplugin,
-  udf.json_extract_histogram(udf.get_key(payload.keyed_histograms.subprocess_crashes_with_dump, 'plugin')).sum AS crashes_detected_plugin,
-  udf.json_extract_histogram(udf.get_key(payload.keyed_histograms.subprocess_crashes_with_dump, 'content')).sum AS crashes_detected_content,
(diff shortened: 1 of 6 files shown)


sql/telemetry_derived/main_summary_v4/part1.sql
```diff
@@ -180,16 +180,14 @@ SELECT
   -- buckets for these particular histograms is not fine enough for the median
   -- to give a more accurate value than the mean.
   udf.histogram_to_mean(
-    udf.json_extract_histogram(payload.histograms.places_bookmarks_count)
+    mozfun.hist.extract(payload.histograms.places_bookmarks_count)
```
jklukas, line 183: I believe the biggest single contributor to "resources exceeded" is the number of persistent UDF references. This PR removes all references to `udf.json_extract_histogram` with references to the underlying mozfun.hist.extract, eliminating a level of indirection.

sql/telemetry_derived/main_summary_v4/part1.sql
```diff
@@ -112,20 +112,20 @@ SELECT
...
-  udf.json_extract_histogram(udf.get_key(payload.keyed_histograms.process_crash_submit_attempt, 'main-crash')).sum AS crash_submit_attempt_main,
-  udf.json_extract_histogram(udf.get_key(payload.keyed_histograms.process_crash_submit_attempt, 'content-crash')).sum AS crash_submit_attempt_content,
-  udf.json_extract_histogram(udf.get_key(payload.keyed_histograms.process_crash_submit_attempt, 'plugin-crash')).sum AS crash_submit_attempt_plugin,
-  udf.json_extract_histogram(udf.get_key(payload.keyed_histograms.process_crash_submit_success, 'main-crash')).sum AS crash_submit_success_main,
-  udf.json_extract_histogram(udf.get_key(payload.keyed_histograms.process_crash_submit_success, 'content-crash')).sum AS crash_submit_success_content,
-  udf.json_extract_histogram(udf.get_key(payload.keyed_histograms.process_crash_submit_success, 'plugin-crash')).sum AS crash_submit_success_plugin,
-  udf.json_extract_histogram(udf.get_key(payload.keyed_histograms.subprocess_kill_hard, 'ShutDownKill')).sum AS shutdown_kill,
+  udf.extract_histogram_sum(udf.get_key(payload.keyed_histograms.subprocess_crashes_with_dump, 'pluginhang')) AS plugin_hangs,
```
jklukas, line 115: Getting the sum of a histogram is by far the most common operation, so we provide a new implementation for just that operation, ignoring the rest of the histogram content.

jklukas:
Note that I've already deployed the UDF changes in this PR, since that's the only way to make them take effect for dry runs.

PR #1129, opened 2020-07-07
Inject function names and descriptions into docs

jklukas:


bigquery_etl/docs/generate_docs.py
```diff
@@ -86,14 +87,31 @@ def main():
...
                                 dataset_doc_file.write("\n\n")
-                                dataset_doc_file.write(load_with_examples(src))
+                                # Inject a level-2 header with the UDF name
+                                if not docfile_content.strip().startswith("#"):
+                                    dataset_doc_file.write(f"## {name}\n\n")
+                                # Inject the "description" from metadata.yaml
+                                if description:
+                                    dataset_doc_file.write(f"{description}\n\n")
+                                # Inject the contents of the README.md
+                                dataset_doc_file.write(docfile_content)
```
scholtzan, line 114: If the UDF-specific doc file does have a section starting with `#`, we'd prepend the description which would be followed by the section. So, I think we could end up with something like:

```
Some description - foo bar

## get_key (or some section name)

content of the README.md
```

jklukas: I'll remove this conditional. I think we should always expect for the function name to be injected.

PR #1137, opened 2020-07-07
Use mozfun.hist.extract in generated desktop GLAM queries

jklukas:
This will make GLAM queries tolerate the new compact histogram encodings


jklukas:
It's unclear to me what the process is for generating and checking in queries when we update the logic, so looking for some advice from @emtwo and/or @acmiyaguchi here. I'm not 100% sure yet that this change yields identical results, so some testing will be necessary.

jklukas:
It's unclear to me what the process is for generating and checking in queries when we update the logic, so looking for some advice from @emtwo and/or @acmiyaguchi here. I'm not 100% sure yet that this change yields identical results, so some testing will be necessary.

acmiyaguchi:
If I'm not mistaken, you can run https://github.com/mozilla/bigquery-etl/blob/master/script/glam/generate_and_run_desktop_sql to just generate the relevant SQL.

sql/telemetry_derived/clients_daily_histogram_aggregates_v1.sql.py
```diff
@@ -43,49 +43,23 @@
...
-                UNNEST(histograms) AS histogram,
-                UNNEST(get_keyval_pairs(JSON_EXTRACT(histogram, "$.values"))) AS keyval
-              WHERE histogram IS NOT NULL
-                AND JSON_EXTRACT(histogram, "$.values") IS NOT NULL
-              GROUP BY key));
+        f"""-- Query generated by: sql/telemetry_derived/clients_daily_histogram_aggregates_v1.sql.py --agg-type {agg_type}
+        CREATE TEMP FUNCTION udf_aggregate_json_sum(histograms ARRAY<STRING>) AS (
+          ARRAY(
+            SELECT AS STRUCT
+              FORMAT('%d', values_entry.key) AS key,
```
jklukas, line 52: We use FORMAT here to turn `key` from an INT64 into a string to match what's expecting for the existing query.

sql/telemetry_derived/clients_daily_histogram_aggregates_v1.sql.py
```diff
@@ -43,49 +43,23 @@
...
+              SUM(values_entry.value) AS value
+            FROM
+              UNNEST(histograms) AS histogram,
+              UNNEST(mozfun.hist.extract(histogram).values) AS values_entry
```
jklukas, line 56: The hist.extract function handles parsing `values` into an array of `STRUCT<key INT64, value INT64>` pairs, so we no longer need the intermediate JS functions for turning the values array into keyval strings and then re-extracting.

sql/telemetry_derived/clients_daily_keyed_boolean_aggregates_v1/query.sql
```diff
@@ -100,7 +104,7 @@ sampled_data AS (
     OR (channel = "release" AND os != "Windows")
     OR (channel = "release" AND os = "Windows" AND MOD(sample_id, @sample_size) = 0)
 ),
--- Using `min` for when `agg_type` is `count` returns null when all rows are null
+        -- Using `min` for when `agg_type` is `count` returns null when all rows are null
```
jklukas, line 107: I don't understand why this space is now getting injected where it didn't before. Could be dedent behavior changing in python versions?

jklukas:
Tests are failing due to new fields that are not present in the schemas checked into the test resources. I'm still thinking about the best way to handle that, but otherwise this is ready for review.

sql/telemetry_derived/clients_daily_keyed_boolean_aggregates_v1/query.sql
acmiyaguchi: This is strange, I don't see any place where this behavior necessarily changed.

sql/telemetry_derived/clients_daily_scalar_aggregates_v1.sql.py
```diff
@@ -50,8 +50,7 @@ def generate_sql(
 ):
     """Create a SQL query for the clients_daily_scalar_aggregates dataset."""
     query = textwrap.dedent(
-        f"""-- Query generated by:
-            -- templates/clients_daily_scalar_aggregates.sql.py --agg-type {agg_type}
+        f"""-- Query generated by: sql/telemetry_derived/clients_daily_scalar_aggregates.sql.py --agg-type {agg_type}
```
acmiyaguchi, line 53: Maybe set this on a newline and lstrip the query to fix the indentation issues?

sql/telemetry_derived/clients_daily_keyed_boolean_aggregates_v1/query.sql
acmiyaguchi: Another explanation could be that the query was not updated when the code was, since the query is generated at runtime in the Airflow DAG.

jklukas: > Another explanation could be that the query was not updated when the code was, since the query is generated at runtime in the Airflow DAG

I think that's likely. I suspect this that the additional space is comes from that fact that the script calls `reformat` for our SQL formatting library. I believe that lib attempts to preserve indentation of comments while it radically changes indentation of surrounding code.

We're injecting various variables into the text here and I'd imagine they are not all indented consistently. Anyways, this is a small issue that I'm going to ignore.

sql/telemetry_derived/clients_daily_histogram_aggregates_v1.sql.py
```diff
@@ -43,49 +43,25 @@
...
+            WHERE
+              histogram IS NOT NULL
+            GROUP BY
+              values_entry.key
+            ORDER BY
+              values_entry.key
```
jklukas, line 62: I've rebased on master and added an ORDER BY here that appears to be necessary to get the tests to run successfully.

jklukas:
Tests should now be passing. I ended up taking out the generated queries. Instead, I've manually applied just the UDF changes to the affected query. The outlay of probes has changed enough since these queries were last generated that it would have required a significant effort to update all the test cases to match the set of probes actually present in the freshly-generated queries.

Issue #1133, opened 2020-07-07
Run UDF tests for mozfun functions


jklukas:
cc @scholtzan 

It appears that the current CI configuration misses functions defined under `mozfun` when running UDF tests. We should make sure those test cases run and that CI fails on logic regressions.

PR #1138, opened 2020-07-07
Run pytest on mozfun UDFs

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/1133

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -17,3 +17,4 @@ testpaths =
     udf
     udf_js
     script
+    mozfun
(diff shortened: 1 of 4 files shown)


pytest.ini
```diff
@@ -17,3 +17,4 @@ testpaths =
     udf
     udf_js
     script
+    mozfun
```
jklukas, line 20: It would have taken me a long time to find this...

jklukas:
It looks like tests fail after merging in #1136 because the mozfun.map.get_key function reference was rewritten as if a temp function, but that temp udf wasn't injected into the query.

PR #1139, opened 2020-07-08
Script for migrating UDFs to mozfun

scholtzan:
I added a script to make migrating UDFs to mozfun easier. It basically traverses the UDF directory, asks whether a UDF should be migrated, to what dataset and how referenced UDFs should be renamed.

Screen Shot 2020-07-07 at 16 58 07

Or to migrate a specific UDF:

`./script/migrate_to_mozfun --udf udf/glean_timespan_seconds.sql`


mozfun/bitmask/bitmask_365/udf.sql
```diff
@@ -0,0 +1,6 @@
+/*
+A bitmask for 365 bits
+*/
+CREATE OR REPLACE FUNCTION bitmask.bitmask_365() AS (
+  CONCAT(b'\x1F', REPEAT(b'\xFF', 45))
+);
```
scholtzan, line 6: Are the `bitmask` UDFs used outside of queries?

scholtzan:
I migrated a few UDFs to `hist` and `glean`, but I guess it's easier to review if migrations happen in different PRs based on datasets etc.

mozfun/bitmask/bitmask_365/udf.sql
jklukas: I doubt it, but cc @fbertsch. I think we would need a tighter naming scheme for these before putting in mozfun.

mozfun/bits28/active_in_range/metadata.yaml
```diff
@@ -1,2 +1,6 @@
-friendly_name: Bits28 Days Active in Range
-description: Return a boolean indicating if any bits are set in the specified range of a bit pattern.
+description: 'Return a boolean indicating if any bits are set in the specified range
+  of a bit pattern. The start_offset must be zero or a negative number indicating
```
jklukas, line 2:
```suggestion
  of a bit pattern. The `start_offset` must be zero or a negative number indicating
```

We should consistently use md code formatting for parameter names.

mozfun/glean/timespan_nanos/metadata.yaml
```diff
@@ -0,0 +1,3 @@
+description: Returns the number of nanoseconds represented by a Glean timespan struct.  See
```
jklukas, line 1: It would be nice to preserve newlines in these descriptions so that it ends up rendering with the link on a separate line.

mozfun/hist/merge/metadata.yaml
```diff
@@ -0,0 +1,4 @@
+description: Merge an array of histograms into a single histogram. - The histogram
+  values will be summed per-bucket - The count will be summed - Other fields will
+  take the mode_last
```
jklukas, line 3: We'll probably need to manually change whitespace here to ensure the list renders well.

mozfun/mode/last/metadata.yaml
```diff
@@ -0,0 +1,5 @@
+description: 'Returns the most frequently occuring element in an array.  In the case
```
jklukas, line 1: `mode` seems to specific to me to be a reasonable dataset-level name. I'd like us to think more about what the logical grouping would be here for functions like `mode_last`.

script/migrate_to_mozfun
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env python3
```
jklukas, line 1: Since this is going to be a short-lived script, I don't think it should live directly under `script`. Perhaps it should go under `script/legacy` to indicate that it's a short-lived script that isn't intended to be maintained long-term.

mozfun/bitmask/bitmask_365/udf.sql
fbertsch: Agreed, I don't think anyone is using these at the moment. I can check in with the search team today though.

scholtzan:
I checked Redash for some commonly used UDFs and migrated some of them.

mozfun/glean/percentile/metadata.yaml
```diff
@@ -0,0 +1,2 @@
+description: 'Glean Percentile'
```
jklukas, line 1: There isn't sufficient documentation here. I think we should drop `glean/percentile` from this PR and it can be added later on a standalone basis. I'm also concerned that there's no obvious indication that this is a JS function.

mozfun/glean/timespan_nanos/metadata.yaml
```diff
@@ -0,0 +1,3 @@
+description: 'Returns the number of nanoseconds represented by a Glean timespan struct. 
+  See https://mozilla.github.io/glean/book/user/metrics/timespan.html'
```
jklukas, line 2:
```suggestion

  See https://mozilla.github.io/glean/book/user/metrics/timespan.html'
```

mozfun/glean/timespan_seconds/metadata.yaml
```diff
@@ -0,0 +1,4 @@
+description: 'Returns the number of seconds represented by a Glean timespan struct,
+  rounded down to full seconds. 
+  See https://mozilla.github.io/glean/book/user/metrics/timespan.html'
```
jklukas, line 3:
```suggestion

  See https://mozilla.github.io/glean/book/user/metrics/timespan.html'
```

mozfun/hist/mean/metadata.yaml
```diff
@@ -0,0 +1,4 @@
+description: Given histogram h, return floor(mean) of the measurements in the bucket.
+  That is, the histogram sum divided by the number of measurements taken. 
+  https://github.com/mozilla/telemetry-batch-view/blob/ea0733c/src/main/scala/com/mozilla/telemetry/utils/MainPing.scala#L292-L307
```
jklukas, line 3:
```suggestion
  https://github.com/mozilla/telemetry-batch-view/blob/ea0733c/src/main/scala/com/mozilla/telemetry/utils/MainPing.scala#L292-L307
  
```

mozfun/hist/merge/metadata.yaml
```diff
@@ -0,0 +1,5 @@
+description: 'Merge an array of histograms into a single histogram. 
+  - The histogram values will be summed per-bucket
```
jklukas, line 2:
```suggestion

  - The histogram values will be summed per-bucket 
```

mozfun/metadata/normalize/metadata.yaml
```diff
@@ -0,0 +1,3 @@
+description: Accepts a pipeline metadata struct as input and returns a modified struct
```
jklukas, line 1: What would you think about a `norm` namespace? So this would become `mozfun.norm.metadata` and then this would be a natural grouping for `norm.os`, `norm.os_version`, etc. as we need for other normalization functions.

mozfun/stats/mode_last/metadata.yaml
```diff
@@ -0,0 +1,5 @@
+description: 'Returns the most frequently occuring element in an array. In the case
```
jklukas, line 1:
```suggestion
description: 'Returns the most frequently occuring element in an array. 

  In the case
```

jklukas, line 1: I like `stats` as a dataset for this.

mozfun/glean/normalize_baseline_client_info/metadata.yaml
```diff
@@ -0,0 +1,3 @@
+description: Accepts a glean client_info struct as input and returns a modified struct
```
jklukas, line 1: If we implement a `norm` dataset as suggested below, I'd argue that this should be moved there as `norm.glean_baseline_client_info`.

mozfun/bits28/active_in_range/metadata.yaml
```diff
@@ -1,2 +1,7 @@
-friendly_name: Bits28 Days Active in Range
-description: Return a boolean indicating if any bits are set in the specified range of a bit pattern.
+description: 'Return a boolean indicating if any bits are set in the specified range
+  of a bit pattern. The `start_offset` must be zero or a negative number indicating
+  an offset from the rightmost bit in the pattern. n_bits is the number of bits to
+  consider, counting right from the bit at `start_offset`. 
+  
+  See detailed docs for the bits28 suite of functions: https://docs.telemetry.mozilla.org/cookbooks/clients_last_seen_bits.html#udf-reference'
+friendly_name: Active In Range
```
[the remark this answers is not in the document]
scholtzan: I ran the migration script on that UDF early on and made some changes after. Must have missed changing the friendly name back.

scholtzan:
Ah, indeed they do.

PR #1140, opened 2020-07-08
Add FXA support to shredder

relud:


bigquery_etl/shredder/amplitude.py
```diff
@@ -132,7 +144,7 @@ def _deletions(start):
                 "user_ids": ids[start:end],
                 "requester": "shredder",
                 "ignore_invalid_id": "True",
-                "delete_from_org": "True",
+                "delete_from_org": "False",
```
relud, line 147: provided credentials should be for the relevant project, so we don't want to `delete_from_org`

relud: breaking this bit out into a separate PR

bigquery_etl/shredder/amplitude.py
```diff
@@ -85,11 +88,20 @@ def main():
...
     )
 
+    conditions = ""
+    if args.submission_date is not None:
+        conditions += f"DATE(submission_timestamp) = DATE '{args.submission_date}'"
+    if args.where is not None:
+        if conditions:
+            conditions += " AND "
+        conditions += args.where
+    where_clause = f" WHERE {conditions}" if conditions else ""
```
jklukas, line 98: I'd find this block cleaner if `conditions` were a list that we appended to and then add ANDs at the end:

```suggestion
    where_clause = f" WHERE {' AND '.join(conditions)}" if conditions else ""
```

relud: would fix, but this change is being replaced by a derived table.

PR #1143, opened 2020-07-08
Reschedule DAGs

scholtzan:


bigquery_etl/query_scheduling/utils.py
```diff
@@ -55,3 +55,45 @@ def is_schedule_interval(interval):
     or a timedelta []d[]h[]m
     """
     return SCHEDULE_INTERVAL_RE.match(interval)
+
+
+def schedule_interval_delta(schedule_interval1, schedule_interval2):
+    """Return the time delta between two schedule intervals as timedelta string."""
```
scholtzan, line 61: This probably doesn't work for all possible cases, but I think for the ones we currently have in our DAGs it should work fine. I'll open an issue to investigate this further later.

dags/bqetl_activity_stream.py
```diff
@@ -36,6 +36,7 @@
         task_id="wait_for_copy_deduplicate_copy_deduplicate_all",
         external_dag_id="copy_deduplicate",
         external_task_id="copy_deduplicate_all",
+        execution_delta=datetime.timedelta(days=-1, seconds=82800),
```
scholtzan, line 39: This is equivalent to `timedelta(hours=-1)`

bigquery_etl/query_scheduling/task.py
```diff
@@ -280,6 +281,25 @@ def with_dependencies(self, client, dag_collection):
                     and d.task_id == upstream_task.task_name
                     for d in self.depends_on
                 ):
-                    dependencies.append(upstream_task)
+                    upstream_schedule_interval = dag_collection.dag_by_name(
```
scholtzan, line 284: For the dependencies that are automatically determined we have to calculate the timedelta between the scheduling intervals of the different DAGs. This is done here

dags/bqetl_activity_stream.py
jklukas: This is strange enough that I think it's worth fixing as part of this PR.

scholtzan: Fix it as in replace it with `timedelta(hours=-1)`?

Because in the python shell I also get:

```
>>> timedelta(hours=-1)
datetime.timedelta(-1, 82800)
```

jklukas: > Fix it as in replace it with `timedelta(hours=-1)`?

Yes, that's what I meant. But you make a good point that `datetime.timedelta(-1, 82800)` is the canonical form. I'm okay with moving forward, but I will file an issue for this, because I think it's going to be confusing for anyone looking at the text of the DAG.

scholtzan: `execution_delta` always confuses me. Looking at it more closely, it needs to be timedelta(hours=1) for the DAGs here.

PR #1146, opened 2020-07-09
Airflow task priority

scholtzan:
I have a suspicion of what is going on here:
Tasks that are waiting on tasks they depend on to finish get queued until they can run. Once slots free up tasks are selected from that queue based on their `priority_weight`, which is the sum of the `priority_weights` of all their downstream tasks (at least that is how I understand it [from the docs](https://airflow.apache.org/docs/stable/_api/airflow/operators/index.html#package-contents)).

I think what is happening in our case is that `main_summary` has a low `priority_weight` since it is a single task with no explicit downstream dependencies. A lot of tasks depend on `main_summary` but that information is lost since we are using `ExternalTaskSensor`s. So, `main_summary` is always stuck in the queue and dependent tasks cannot run resulting in a deadlock.

A quick fix is to manually set the `priority_weights` and re-scheduling tasks. I think even better would be to automatically determine the right `priority_weight` when generating the DAGs.

Another improvement that is mentioned in [the docs](https://airflow.apache.org/docs/stable/concepts.html#pools) is to assign sensors to a separate `pool`. This will prevent sensors blocking tasks from running.

There is also an interesting blog post with some Airflow gotchas I found here: https://gtoonstra.github.io/etl-with-airflow/gotchas.html


bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -81,6 +85,7 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
         {% if task_ref.execution_delta -%}
         execution_delta={{ task_ref.execution_delta | format_timedelta | format_repr }},
         {% endif -%}
+        pool='sensor_pool',
```
scholtzan, line 88: For this to not fail, we'd first need to create this pool. I think it is not *strictly necessary* to fix our issue but potentially a good improvement.

sql/telemetry_derived/main_summary_v4/metadata.yaml
```diff
@@ -9,7 +9,8 @@ labels:
 scheduling:
   dag_name: bqetl_main_summary
   start_date: '2019-10-25'
+  priority: 90
```
scholtzan, line 12: Manually increasing the priority for main_summary so that it gets scheduled first

jklukas: We might want to also increase priority for clients_daily and clients_last_seen (and maybe the core and baseline variants), since many other DAGs have sensors that wait for those.

PR #1150, opened 2020-07-09
Bug 1649074 Extract a table for FxA device-command events

jklukas:
Based on example query from :rfkelly in https://bugzilla.mozilla.org/show_bug.cgi?id=1649074#c1


sql/firefox_accounts_derived/fxa_log_device_command_events_v1/query.sql
```diff
@@ -0,0 +1,39 @@
...
+  WHERE
+    key_id = 'fxa_hmac_prod'
+)
+--
+SELECT
+  `timestamp`,
+  jsonPayload.type,
+  TO_HEX(
+    udf.hmac_sha256((SELECT * FROM hmac_key), CAST(jsonPayload.fields.uid AS BYTES))
+  ) AS user_id,
```
rfk, line 19: Could we please apply the same hmacing logic to the `index`, `target` and `sender` fields? Like `uid`, these fields are unique identifiers stored on the server that could be used to tie metrics back to a specific account.

sql/firefox_accounts_derived/fxa_log_device_command_events_v1/query.sql
```diff
@@ -0,0 +1,39 @@
...
+  jsonPayload.fields.index,
+  jsonPayload.fields.command,
+  jsonPayload.fields.target,
+  jsonPayload.fields.targetOS AS target_os,
+  jsonPayload.fields.targetType AS target_type,
+  jsonPayload.fields.sender,
+  jsonPayload.fields.sender AS sender_os,
```
rfk, line 26: nit: `fields.sender` here should be `fields.senderOS` IIUC.

sql/firefox_accounts_derived/fxa_log_device_command_events_v1/query.sql
```diff
@@ -0,0 +1,48 @@
...
+  jsonPayload.type,
+  TO_HEX(
+    udf.hmac_sha256((SELECT * FROM hmac_key), CAST(jsonPayload.fields.uid AS BYTES))
+  ) AS user_id,
+  TO_HEX(
+    udf.hmac_sha256(
+      (SELECT * FROM hmac_key),
+      CAST(FORMAT('%d', CAST(jsonPayload.fields.index AS INT64)) AS BYTES)
+    )
+  ) AS index,
```
jklukas, line 25: `index` required some additional logic because it's represented as a float in the BQ table. It appears these are all really integers, so we cast to int, format that int as a string, and then cast the unicode string to bytes to send through the hmac hash. As long as the assumption that these are logically integers is sound, then this should give a stable hash for different integers.

jklukas:
After this gets merged, I'm going to create a separate PR for moving all these fxa queries into a DAG generated by this repo, which will include scheduling this new query.

PR #1155, opened 2020-07-10
bqetl_asn_aggregates DAG

scholtzan:
Migrating `asn_aggregates` query to their own DAG

Generates https://github.com/mozilla/telemetry-airflow/pull/1086

### DIFF
--- a/bigquery_etl/query_scheduling/task.py
+++ b/bigquery_etl/query_scheduling/task.py
@@ -103,6 +103,7 @@ class Task:
     multipart: bool = attr.ib(False)
     sql_file_path: Optional[str] = None
     priority: Optional[int] = None
+    skip_dry_run: bool = attr.ib(False)
     referenced_tables: Optional[List[Tuple[str, str]]] = attr.ib(None)
 
     @owner.validator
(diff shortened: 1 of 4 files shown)


sql/telemetry_derived/asn_aggregates_v1/metadata.yaml
```diff
@@ -9,3 +9,14 @@ owners:
 labels:
   incremental: true
   schedule: daily
+scheduling:
+  dag_name: bqetl_asn_aggregates
+  parameters: ["n_clients:INT64:500"]
+  skip_dry_run: true   # query accesses payload_bytes_raw.telemetry
```
scholtzan, line 15: The dry run would fail here since it doesn't have permissions to access payload_bytes_raw

bigquery_etl/query_scheduling/task.py
```diff
@@ -103,6 +103,7 @@ class Task:
     multipart: bool = attr.ib(False)
     sql_file_path: Optional[str] = None
     priority: Optional[int] = None
+    skip_dry_run: bool = attr.ib(False)
```
jklukas, line 106: This is going to be immediately useful for migrating the fxa_events DAG 👍

PR #1156, opened 2020-07-10
Optionally provide referenced_tables in metadata.yaml

relud:
because queries that reference main_v4 are slow to dry run

### DIFF
--- a/bigquery_etl/query_scheduling/task.py
+++ b/bigquery_etl/query_scheduling/task.py
@@ -7,7 +7,7 @@ import os
 import re
 import logging
 from google.cloud import bigquery
-from typing import List, Optional
+from typing import List, Optional, Tuple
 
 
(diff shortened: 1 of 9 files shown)


bigquery_etl/query_scheduling/task.py
```diff
@@ -103,6 +103,7 @@ class Task:
     multipart: bool = attr.ib(False)
     sql_file_path: Optional[str] = None
     priority: Optional[int] = None
+    referenced_tables: Optional[List[Tuple[str, str]]] = attr.ib(None)
```
jklukas, line 106: Why is this a list of 2-tuples rather than a map?

relud: it's the return type of `self._get_referenced_tables(client)`, which is a list of table references as a tuple of `dataset_id` and `table_id`. neither `dataset_id` nor `table_id` would be a unique key for each element.

PR #1159, opened 2020-07-10
Reference default_args for email when generating dag tasks

relud:
Without this most of our dags only email the owners.


dags/bqetl_main_summary.py
```diff
@@ -46,7 +51,12 @@
         dataset_id="telemetry_derived",
         project_id="moz-fx-data-shared-prod",
         owner="jklukas@mozilla.com",
-        email=["jklukas@mozilla.com"],
+        email=[
+            "dthorn@mozilla.com",
+            "frank@mozilla.com",
+            "jklukas@mozilla.com",
+            "telemetry-alerts@mozilla.com",
```
scholtzan, line 58: Do we always want to email everyone listed under `email` for the DAG for each task? Just wondering, since that hasn't been the case in the past: https://github.com/mozilla/telemetry-airflow/blob/b872df04bad82e6e4c20c06d9404f95822920e49/dags/main_summary.py#L368

relud: no, but in those cases we should set `email` in `metadata.yaml` to reduce the list.

scholtzan: Ah, I see.

PR #1165, opened 2020-07-13
Add experiment analysis tables to shredder

scholtzan:
This adds experiment analysis tables to shredder.

Tables for experiment analyses are stored as part of `moz-fx-data-experiments`. Currently, there is only one dataset, `mozanalysis`, with tables that are named like: `<some experiment identifier>_<day|week|overall>_<n>`. The dataset also contains some views that have names similar to the tables and tables with aggregated data that start with `statistics_` but do not contain any user identifiers.

Is there a good way to test this?

### DIFF
--- a/bigquery_etl/shredder/config.py
+++ b/bigquery_etl/shredder/config.py
@@ -121,6 +121,7 @@ impression_id_target = partial(DeleteTarget, field=IMPRESSION_ID)
 cfr_id_target = partial(DeleteTarget, field=CFR_ID)
 fxa_user_id_target = partial(DeleteTarget, field=FXA_USER_ID)
 user_id_target = partial(DeleteTarget, field=USER_ID)
+experiment_analysis_target = partial(DeleteTarget, field=CLIENT_ID)
 
 DELETE_TARGETS = {
     client_id_target(
@@ -354,4 +355,27 @@ def find_glean_targets(pool, client, project=SHARED_PROD):
             )
             if any(field.name == CLIENT_ID for field in table.schema)
         },
+    }
+
+
+EXPERIMENT_ANALYSIS = "moz-fx-data-experiments"
+
+
+def find_experiment_analysis_targets(pool, client, project=EXPERIMENT_ANALYSIS):
+    """Return a dict like DELETE_TARGETS for experiment analysis tables."""
+    datasets = {dataset.dataset_id for dataset in client.list_datasets(project)}
+
+    tables = [
+        table
+        for tables in pool.map(
+            client.list_tables,
+            [bigquery.DatasetReference(project, dataset_id) for dataset_id in datasets],
+        )
+        for table in tables
+        if table.table_type != "VIEW" and not table.table_id.startswith("statistics_")
+    ]
+
+    return {
+        experiment_analysis_target(qualified_table_id(table)): DESKTOP_SRC
+        for table in tables
     }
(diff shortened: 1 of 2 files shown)


bigquery_etl/shredder/config.py
```diff
@@ -355,3 +356,26 @@ def find_glean_targets(pool, client, project=SHARED_PROD):
...
+def find_experiment_analysis_targets(pool, client, project=EXPERIMENT_ANALYSIS):
+    """Return a dict like DELETE_TARGETS for experiment analysis tables."""
+    datasets = {dataset.dataset_id for dataset in client.list_datasets(project)}
+
+    tables = [
+        table
+        for tables in pool.map(
+            client.list_tables,
+            [bigquery.DatasetReference(project, dataset_id) for dataset_id in datasets],
+        )
```
relud, line 373: There's no need to construct the dataset references here, they are available via `DatasetListItem.reference`. Additionally, setting `chunksize=1` ensures that each thread in the pool only claims one dataset at a time to work on, which is ideal for small workloads like this.
```suggestion
    datasets = {dataset.reference for dataset in client.list_datasets(project)}

    tables = [
        table
        for tables in pool.map(
            client.list_tables,
            datasets,
            chunksize=1,
        )
```

bigquery_etl/shredder/delete.py
```diff
@@ -396,9 +401,15 @@ def main():
             )
     with ThreadPool(args.parallelism) as pool:
         glean_targets = find_glean_targets(pool, client)
+    with ThreadPool(args.parallelism) as pool:
+        experiment_analysis_targets = find_experiment_analysis_targets(pool, client)
```
relud, line 405: there's no need to recreate this thread pool
```suggestion
    with ThreadPool(args.parallelism) as pool:
        glean_targets = find_glean_targets(pool, client)
        experiment_analysis_targets = find_experiment_analysis_targets(pool, client)
```

bigquery_etl/shredder/config.py
```diff
@@ -355,3 +356,26 @@ def find_glean_targets(pool, client, project=SHARED_PROD):
...
+        for table in tables
+        if table.table_type != "VIEW" and not table.table_id.startswith("statistics_")
+    ]
+
+    return {
+        experiment_analysis_target(qualified_table_id(table)): DESKTOP_SRC
```
relud, line 379: this should reuse `client_id_target`, instead of defining `experiment_analysis_target` with identical behavior
```suggestion
        client_id_target(table=qualified_table_id(table)): DESKTOP_SRC
```

bigquery_etl/shredder/config.py
```diff
@@ -121,6 +121,7 @@ def fields(self) -> Tuple[str]:
 cfr_id_target = partial(DeleteTarget, field=CFR_ID)
 fxa_user_id_target = partial(DeleteTarget, field=FXA_USER_ID)
 user_id_target = partial(DeleteTarget, field=USER_ID)
+experiment_analysis_target = partial(DeleteTarget, field=CLIENT_ID)
```
relud, line 124:
```suggestion
```

relud:
> Is there a good way to test this?

you can dry run the shredder delete script like this, so long as you have sufficient permissions to read and list the new tables:
```sh
script/shredder_delete --end-date=2020-06-29 --start-date=2020-05-04 --only mozanalysis --dry-run --read-only --partition-limit=2
```
That will also print all skipped tables, and if you also comment out [these two lines](https://github.com/mozilla/bigquery-etl/pull/1165/files#diff-14b9db663f895d76f61cc3a1c2175a8cR409-R410) it should not skip any tables.

Another concern here is making sure that airflow has permissions to modify these tables.

scholtzan:
Just checked, Airflow should have all necessary permissions to modify the tables.

PR #1168, opened 2020-07-14
Fix asn_aggregates and fxa_events

scholtzan:
Some fxa_events queries couldn't perform a dryrun since they are accessing tables from different projects.

Generates https://github.com/mozilla/telemetry-airflow/pull/1093


sql/firefox_accounts_derived/fxa_auth_bounce_events_v1/metadata.yaml
```diff
@@ -9,3 +9,4 @@ labels:
 scheduling:
   dag_name: bqetl_fxa_events
   arguments: ['--schema_update_option=ALLOW_FIELD_ADDITION']
+  referenced_tables: []
```
jklukas, line 12: I was about to ask:

> Why wasn't DAG generation failing if these failed to dry run?

And then I realized it's because DAG generation happens locally, and I _do_ have access to the fxa log tables, so I was able to run this while others would not be able to.

But, I also thought that dag generation ran as part of tests. Would it be possible to run `generate_airflow_dags` as part of CI and validate that what's committed matches what's generated? Or we maybe can't do that due to not having BQ creds as part of CI?

scholtzan: CI would need credentials for performing dry runs. The dry run functionality provided by the Cloud Function doesn't seem to return information about referenced tables, only if the dry run failed or succeeded.
To avoid running into these errors we could consider extending the Cloud Function or provide a separate one that dry runs and returns referenced tables and can be used locally and from the CI.

PR #1170, opened 2020-07-14
Allow field addition for specified dates in airflow

relud:
requires https://github.com/mozilla/telemetry-airflow/pull/1094


jklukas:
What's the use case this is solving?

jklukas:
It would be helpful to see an example use of this functionality included in this PR.

sql/search_derived/search_clients_daily_v8/metadata.yaml
```diff
@@ -6,3 +6,5 @@ labels:
   schedule: daily
 scheduling:
   dag_name: bqetl_search
+  allow_field_addition_dates:
+  - 2020-07-13
```
relud, line 10: > It would be helpful to see an example use of this functionality included in this PR.

done

relud:
> What's the use case this is solving?

In https://github.com/mozilla/bigquery-etl/pull/1167#issuecomment-658336949 I recommended manually running a query with `--schema_update_option=ALLOW_FIELD_ADDITION`, and I felt like it would be better if that were actually run by airflow. Also, if there were a table in this repo that data eng doesn't have permissions to edit, we could still add fields in this way.

I'm not 100% sure we should actually do this, or do it in this way. I felt that because it was effectively no extra effort, including implementation would be less ambiguous about my intent.

scholtzan:
Is there an advantage to restricting this to specific dates? Or could we just have `ALLOW_FIELD_ADDITION` enabled for all or specific queries in general?

sql/search_derived/search_clients_daily_v8/metadata.yaml
jklukas: Can you also run dag generation and commit the results?

relud:
> Is there an advantage to restricting this to specific dates? Or could we just have `ALLOW_FIELD_ADDITION` enabled for all or specific queries in general?

When it's not set we are protected against unexpected schema changes that can't be easily reverted, and that has previously been my motivation for not suggesting it be enabled by default.

sql/search_derived/search_clients_daily_v8/metadata.yaml
relud: done

dags/bqetl_search.py
```diff
@@ -70,6 +70,7 @@
         email=["bewu@mozilla.com", "frank@mozilla.com", "telemetry-alerts@mozilla.com"],
         date_partition_parameter="submission_date",
         depends_on_past=False,
+        allow_field_addition_dates=("2020-07-13",),
```
jklukas, line 73: There's probably not much value to keeping old dates in this list, and in fact it would be slightly safer to remove old dates so that we don't accidentally allow date addition in backfills. So how about we change this field to a single date `allow_field_addition_on_date` and expect it to be overwritten for each new schema update.

dags/bqetl_mobile_search.py
```diff
@@ -29,6 +29,7 @@
         email=["bewu@mozilla.com", "telemetry-alerts@mozilla.com"],
         date_partition_parameter="submission_date",
         depends_on_past=False,
+        allow_field_addition_on_date=2020 - 07 - 13,
```
jklukas, line 32: Looks like we need to pick a different formatter here.

relud: it was missing quotes, so black added spaces. fixed.

PR #1171, opened 2020-07-15
Bug 1646362 - Accessibility inspector derived dataset

wlach:


sql/telemetry_derived/devtools_accessiblility_panel_usage_v1/metadata.yaml
```diff
@@ -0,0 +1,11 @@
+friendly_name: Devtools Panel Usage
+description: >
+  A daily number of usages of different accessibility devtools, partitioned by day.
+owners:
+  - wlachance@mozilla.com
```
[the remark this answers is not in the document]
yzen: If that helps yeah

sql/telemetry_derived/devtools_accessiblility_panel_usage_v1/query.sql
```diff
@@ -0,0 +1,133 @@
+WITH accessibility_panel_client_days AS (
+  SELECT
+    submission_date,
+    normalized_channel,
+    payload.processes.parent.scalars.devtools_accessibility_opened_count AS opened_count,
```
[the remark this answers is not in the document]
yzen: I think I'm probably the only frequent user of the dashboard. It is still on the team's/product's radar since, historically, other devtools dashboards do not really include much of the telemetry from the accessibility panel.

sql/telemetry_derived/devtools_accessiblility_panel_usage_v1/query.sql
```diff
@@ -0,0 +1,133 @@
...
+  WHERE
+    sample_id = 1
+    AND payload.processes.parent.scalars.devtools_accessibility_opened_count > 0
+)
+SELECT
+  submission_date,
+  normalized_channel,
+  COUNT(*) AS clients,
+  SUM(opened_count) AS panel_opens,
+  SUM(context_menu_opened_count) / (COUNT(*)) AS context_menu_opened_per_client_day,
```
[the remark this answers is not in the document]
yzen: Yes I believe that's right, basically see what fraction of daily users use the context menu.

yzen:
@wlach , let me know if I can help to clarify anything else, and thanks for looking at this! Regarding the date, yes the start date is too "new" because the query was not completing. I think ideally it should go back all the way into about august 2018 when I landed the panel enabled by default on all channels.

sql/telemetry_derived/devtools_accessiblility_panel_usage_v1/query.sql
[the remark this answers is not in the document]
wlach: 👍 let's stay consistent with the devtools panel then

sql/telemetry_derived/devtools_accessiblility_panel_usage_v1/query.sql
```diff
@@ -0,0 +1,134 @@
...
+    sample_id = 1
+    AND DATE(submission_timestamp) = @submission_date
+    AND payload.processes.parent.scalars.devtools_accessibility_opened_count > 0
+)
+SELECT
+  submission_date,
+  normalized_channel,
+  COUNT(*) AS clients,
+  SUM(opened_count) AS panel_opens,
+  SUM(context_menu_opened_count) / (COUNT(*)) AS context_menu_opened_per_client_day,
```
jklukas, line 78: You may want to use `SAFE_DIVIDE` rather than `/` in case any of these counts returns 0.

sql/telemetry_derived/devtools_accessiblility_panel_usage_v1/query.sql
```diff
@@ -0,0 +1,134 @@
...
+  COALESCE(SUM(CAST(print_to_json_context_count > 0 AS INT64)), 0) AS print_to_json_context_users,
+  SUM(node_inspected_count) / (COUNT(*)) AS node_inspected_per_client_day,
+  COALESCE(SUM(CAST(node_inspected_count > 0 AS INT64)), 0) AS node_inspected_users,
+  COALESCE(SUM(CAST(picker_used_count > 0 AS INT64)), 0) AS picker_users,
+  COALESCE(SUM(CAST(picker_used_count > 0 AS INT64)), 0) / (
+    COUNT(*)
+  ) AS picker_users_per_client_day,
+  SUM(picker_used_count) / (COUNT(*)) AS picker_count_per_client_day,
+  SUM(browser_context_count) / (COUNT(*)) AS browser_context_per_client_day,
+  COALESCE(SUM(CAST(browser_context_count > 0 AS INT64)), 0) AS browser_context_users,
```
jklukas, line 89: Looking more carefully at the logic, I think the best way to get to correct counts is to make the initial query aggregate per client, and then these metrics would become:

```suggestion
  SAFE_DIVIDE(SUM(browser_context_count), (COUNT(*)) AS browser_context_per_client_day,
  COUNTIF(browser_context_count > 0) AS browser_context_users,
```

sql/telemetry_derived/devtools_accessiblility_panel_usage_v1/query.sql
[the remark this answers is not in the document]
jklukas: It looks like this is currently measuring the average number of times the context menu was opened per day for all clients that interacted with the accessibility inspector, and the name of the field seems to agree with that. If we want the fraction of clients who opened the context menu, then it should be a `COUNTIF(context_menu_opened > 0)` divided by the count of all users.

sql/telemetry_derived/devtools_accessiblility_panel_usage_v1/query.sql
```diff
@@ -0,0 +1,134 @@
...
+      "accessibility.simulation.tritanomaly"
+    ) AS simulation_tritanomaly_count,
+    udf.get_key(
+      payload.processes.parent.keyed_scalars.devtools_accessibility_simulation_activated,
+      "accessibility.simulation.contrastLoss"
+    ) AS simulation_contrast_loss_count
+  FROM
+    telemetry.main
+  WHERE
+    sample_id = 1
```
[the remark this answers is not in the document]
wlach: The original used `0`, I'm not sure what I was thinking. Let's change back to that.

wlach:
Updated based on @jklukas's most excellent feedback.

I re-ran these queries based on July's data and found no material difference between them, aside from some minor changes due to the fact that we're aggregating per-client now (which is actually more accurate than what it was doing previously). Compare & contrast:

https://sql.telemetry.mozilla.org/queries/73096
https://sql.telemetry.mozilla.org/queries/73089

The CI failures seem to be false positive.

sql/telemetry_derived/devtools_accessiblility_panel_usage_v1/query.sql
```diff
@@ -0,0 +1,172 @@
...
+  COUNTIF(simulation_tritanopia_count > 0) AS simulation_tritanopia_users,
+  SAFE_DIVIDE(SUM(simulation_tritanomaly_count), COUNT(*)) AS simulation_tritanomaly_per_client_day,
+  COUNTIF(simulation_tritanomaly_count > 0) AS simulation_tritanomaly_users,
+  SAFE_DIVIDE(
+    SUM(simulation_contrast_loss_count),
+    COUNT(*)
+  ) AS simulation_contrast_loss_per_client_day,
+  COUNTIF(simulation_contrast_loss_count > 0) AS simulation_contrast_loss_users,
+  SUM(
+    simulation_deuteranomaly_count + simulation_protanomaly_count + simulation_protanopia_count + simulation_deuteranopia_count + simulation_tritanopia_count + simulation_tritanomaly_count + simulation_contrast_loss_count
```
wlach, line 166: @yzen just a note, it looks like this always 0. is that expected? https://sql.telemetry.mozilla.org/queries/73089#183070

yzen: yeah, I think this query is incorrect, I did not get to fixing it just yet..

wlach: It would be nice to fix this before we merge, so we don't have to backfill twice.

I think the actual SUM part is fine, it's just we're not getting any values for the keyed scalars for... some reason? I notice that `simulation_<x>_count` is always 0.

Here's the definition in the probe dictionary if you haven't seen it before: https://probes.telemetry.mozilla.org/?search=devtools&optout=true&view=detail&probeId=scalar%2Fdevtools.accessibility.simulation_activated

sql/telemetry_derived/devtools_accessiblility_panel_usage_v1/query.sql
```diff
@@ -0,0 +1,172 @@
...
+  FROM
+    telemetry.main
+  WHERE
+    sample_id = 0
+    AND DATE(submission_timestamp) = @submission_date
+    AND payload.processes.parent.scalars.devtools_accessibility_opened_count > 0
+  GROUP BY
+    1,
+    2,
+    3
```
jklukas, line 112:
```suggestion
    submission_date,
    client_id,
    normalized_channel
```

We consider it good practice to name grouping columns explicitly to avoid unintentional changes when reordering columns.

sql/telemetry_derived/devtools_accessiblility_panel_usage_v1/query.sql
```diff
@@ -0,0 +1,172 @@
...
+  ) AS simulation_contrast_loss_per_client_day,
+  COUNTIF(simulation_contrast_loss_count > 0) AS simulation_contrast_loss_users,
+  SUM(
+    simulation_deuteranomaly_count + simulation_protanomaly_count + simulation_protanopia_count + simulation_deuteranopia_count + simulation_tritanopia_count + simulation_tritanomaly_count + simulation_contrast_loss_count
+  ) AS simulations
+FROM
+  accessibility_panel_client_days
+GROUP BY
+  1,
+  2
```
jklukas, line 172:
```suggestion
  submission_date,
  normalized_channel
```

jklukas:
dry runs were failing due to unrelated main_summary changes. I merged in master, which should allow the test to pass.

wlach:
Thanks @jklukas! Just waiting on @yzen to provide some feedback on https://github.com/mozilla/bigquery-etl/pull/1171#discussion_r459036765 and then I can do the rest of what you suggest.

sql/telemetry_derived/devtools_accessiblility_panel_usage_v1/query.sql
yzen: Turns out I was using incorrect keys for the scalar (I also needed to remove some keys that are no longer present and added a new one). They should be: `PROTANOPIA`, `DEUTERANOPIA`, `TRITANOPIA`, `ACHROMATOPSIA` and `CONTRAST_LOSS`. In the original query, the simulation related code looks like:
```
    udf.get_key(payload.processes.parent.keyed_scalars.devtools_accessibility_simulation_activated, "PROTANOPIA") * 100 AS simulation_protanopia_count,
    udf.get_key(payload.processes.parent.keyed_scalars.devtools_accessibility_simulation_activated, "DEUTERANOPIA") * 100 AS simulation_deuteranopia_count,
    udf.get_key(payload.processes.parent.keyed_scalars.devtools_accessibility_simulation_activated, "TRITANOPIA") * 100 AS simulation_tritanopia_count,
    udf.get_key(payload.processes.parent.keyed_scalars.devtools_accessibility_simulation_activated, "ACHROMATOPSIA") * 100 AS simulation_achromatopsia_count,
    udf.get_key(payload.processes.parent.keyed_scalars.devtools_accessibility_simulation_activated, "CONTRAST_LOSS") * 100 AS simulation_contrast_loss_count
```
And in the second query I changed simulation related code to:
```
    SUM(simulation_protanopia_count) / (COUNT(*) * 100) AS simulation_protanopia_per_client_day,
    COALESCE(SUM(CAST(simulation_protanopia_count > 0 AS INT64)), 0) * 100 AS simulation_protanopia_users,
    SUM(simulation_deuteranopia_count) / (COUNT(*) * 100) AS simulation_deuteranopia_per_client_day,
    COALESCE(SUM(CAST(simulation_deuteranopia_count > 0 AS INT64)), 0) * 100 AS simulation_deuteranopia_users,
    SUM(simulation_tritanopia_count) / (COUNT(*) * 100) AS simulation_tritanopia_per_client_day,
    COALESCE(SUM(CAST(simulation_tritanopia_count > 0 AS INT64)), 0) * 100 AS simulation_tritanopia_users,
    SUM(simulation_achromatopsia_count) / (COUNT(*) * 100) AS simulation_achromatopsia_per_client_day,
    COALESCE(SUM(CAST(simulation_achromatopsia_count > 0 AS INT64)), 0) * 100 AS simulation_achromatopsia_users,
    SUM(simulation_contrast_loss_count) / (COUNT(*) * 100) AS simulation_contrast_loss_per_client_day,
    COALESCE(SUM(CAST(simulation_contrast_loss_count > 0 AS INT64)), 0) * 100 AS simulation_contrast_loss_users,
    
    SUM(simulation_protanopia_count + simulation_deuteranopia_count + simulation_tritanopia_count + simulation_achromatopsia_count + simulation_contrast_loss_count) AS simulations
```

wlach: Is it just me or is the total number of simulations still equalling 0 in your query? https://sql.telemetry.mozilla.org/queries/73089/source

yzen: Yeah, this functionality depends on web render so if there would be very few users at the moment. There's also sampling. I tried without sampling and I got some numbers on non-release channels but the are very low (<10 total)

wlach: @yzen and I did some analysis on Slack. It looks like we get a fairly small set of results if we turn off sampling. Given that one month's worth of this query is only half a TB in size, we think it's worth turning off sampling so we can detect small changes in things like this.

wlach:
@jklukas am I correct in thinking that generating the dags is no longer necessary? It looks like we're going to want to backfill this back to August 2018, it sounds like we can do that via the airflow ui once this lands?

jklukas:
> @jklukas am I correct in thinking that generating the dags is no longer necessary? 

No, it will continue to be necessary to run `scripts/generate_airflow_dags.sh` locally and commit the results.

What's new is that it's no longer necessary to create a separate telemetry-airflow PR. Airflow pulls down the contents of bigquery-etl every ~5 minutes to get DAG updates.

Once a PR is merged that includes DAG changes, Airflow will pick up the DAG changes automatically.

> It looks like we're going to want to backfill this back to August 2018, it sounds like we can do that via the airflow ui once this lands?

Yes, it should be possible to run that via the Airflow backfill UI.

wlach:
@jklukas ok I think this is ready now

PR #1172, opened 2020-07-15
Bug 1653499 Base AMO usage stats on payload.info.addons

jklukas:
cc @willdurand I've created this table with one day of data (2020-07-14) for comparison.

### DIFF
--- a/sql/amo_prod/desktop_addons_by_client_v1/query.sql
+++ b/sql/amo_prod/desktop_addons_by_client_v1/query.sql
@@ -1,21 +1,37 @@
+-- See https://stackoverflow.com/a/44885334/1260237
+-- and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent
+CREATE TEMPORARY FUNCTION decode_uri_component(path STRING)
+RETURNS STRING
+LANGUAGE js
+AS
+  """
+if (path == null) {
+  return null;
+}
+try {
+  return decodeURIComponent(path);
+} catch (e) {
+  return path;
+}
+""";
+
 WITH per_client AS (
   SELECT
     DATE(submission_timestamp) AS submission_date,
     client_id,
     sample_id,
-    ARRAY_CONCAT(
-      -- active_addons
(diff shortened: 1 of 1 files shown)


sql/amo_prod/desktop_addons_by_client_v2/query.sql
```diff
@@ -0,0 +1,46 @@
...
+        ARRAY(
+          SELECT AS STRUCT
+            REPLACE(REGEXP_EXTRACT(addon, "(.+):"), '%40', '@') AS id,
+            REGEXP_EXTRACT(addon, ":(.+)") AS version
+          FROM
+            UNNEST(SPLIT(payload.info.addons)) AS addon
+        )
+      ),
+      -- themes
+      ARRAY_AGG(STRUCT(environment.addons.theme.id, environment.addons.theme.version))
```
willdurand, line 18: I think themes should also be present in `payload.info.addons`.

sql/amo_prod/desktop_addons_by_client_v2/query.sql
```diff
@@ -0,0 +1,41 @@
+WITH per_client AS (
+  SELECT
+    DATE(submission_timestamp) AS submission_date,
+    client_id,
+    sample_id,
+    ARRAY_CONCAT_AGG(
+      ARRAY(
+        SELECT AS STRUCT
+          REPLACE(REGEXP_EXTRACT(addon, "(.+):"), '%40', '@') AS id,
```
willdurand, line 9: I think we should also replace `%7D` with `}` and `%7B` with `{`.

jklukas: I've extracted this to a Javascript UDF where we can do proper URI decoding. It makes the query take a bit longer to run, but gets us significantly more correct results. That boosted us to 99% coverage of the v1 data.

willdurand:
Looks good to me, I think we can move forward: https://bugzilla.mozilla.org/show_bug.cgi?id=1653499 (we'll make sure that this Telemetry field doesn't get removed by accident).

PR #1174, opened 2020-07-15
Replace jq in generate_glean_sql

Ben-Wu:
I just noticed that the docker image doesn't have jq


Ben-Wu:
Ya I just felt it was overkill to add it to the dockerfile for this one use

PR #1179, opened 2020-07-16
Reduce complexity of main_summary by referencing mozfun udfs directly

relud:


relud:
> There is also a `mozfun.map.get_key` UDF (https://github.com/mozilla/bigquery-etl/blob/master/mozfun/map/get_key/udf.sql) which is the same as `udf.get_key`. It looks like `udf.get_key` doesn't wrap the mozfun UDF yet (probably because I missed it), but it might make sense to use `mozfun.map.get_key` here instead. Or we keep it in mind as something we need to look out for once `udf.get_key` gets updated.

added `mozfun.map.get_key`

Issue #1089, opened 2020-06-24
DAG generation: infer copy_dedupe dependencies


jklukas:
Currently, we must add explicit `depends_on` entries in `metadata.yaml` in order to generate `wait` sensors for stable tables in generated DAGs.

We already can determine via dry run that a given query depends on stable tables, and I think we should automatically inject sensors to wait on copy_deduplicate tasks based on stable table dependencies.

scholtzan:
Sounds like a good idea. As a first step before implementing this, I think all the copy_deduplicate tasks should be moved to a common `copy_deduplicate` DAG. Currenly, `copy_deduplicate_main_ping`, for example, is part of the `main_summary` DAG while other tasks are in the `copy_deduplicate` DAG.

jklukas:
I agree that there should be a single DAG for copy_dedupe and that `copy_deduplicate_main_ping` should be moved there.

PR #1182, opened 2020-07-17
Lower-friction scheduled queries: automatically determine copy_dedupe dependencies

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/1089


bigquery_etl/query_scheduling/task.py
```diff
@@ -74,6 +77,49 @@ def validate_execution_delta(self, attribute, value):
...
+        task_id="baseline_clients_last_seen",
+        schedule_interval="0 1 * * *"
+    ): ["*.baseline_clients_last_seen*"],
+    TaskRef(
+        dag_name="copy_deduplicate",
+        task_id="copy_deduplicate_all",
+        schedule_interval="0 1 * * *"
+    ): ["*_stable.*"]
+}
+
```
scholtzan, line 122: This maps tables that are referenced in a query to the corresponding telemetry-airflow copy_deduplicate task it depends on. Matching tables are specified as patterns.

We might want to have a config file for this in the future, but having this in code will do for now.

dags/bqetl_nondesktop.py
```diff
@@ -56,42 +56,38 @@
...
-        wait_for_copy_deduplicate_baseline_clients_last_seen
+        wait_for_telemetry_derived__core_clients_last_seen__v1
     )
 
     firefox_nondesktop_exact_mau28_by_client_count_dimensions.set_upstream(
         wait_for_telemetry_derived__core_clients_last_seen__v1
     )
-    firefox_nondesktop_exact_mau28_by_client_count_dimensions.set_upstream(
-        wait_for_copy_deduplicate_baseline_clients_last_seen
-    )
```
scholtzan, line 97: This dependency would get removed. I don't think it is actually required, but if it is then it can be added back using `depends_on`

bigquery_etl/query_scheduling/task.py
```diff
@@ -74,6 +77,49 @@ def validate_execution_delta(self, attribute, value):
...
+        schedule_interval="0 1 * * *",
+    ): ["telemetry_stable.main_v4"],
+    TaskRef(
+        dag_name="copy_deduplicate",
+        task_id="bq_main_events",
+        schedule_interval="0 1 * * *",
+    ): ["telemetry_derived.main_events_v1"],
+    TaskRef(
+        dag_name="copy_deduplicate",
+        task_id="events_events",
```
jklukas, line 108:
```suggestion
        task_id="event_events",
```

bigquery_etl/query_scheduling/task.py
```diff
@@ -74,6 +77,49 @@ def validate_execution_delta(self, attribute, value):
...
+        schedule_interval="0 1 * * *",
+    ): ["telemetry_derived.events_events_v1"],
```
jklukas, line 110:
```suggestion
    ): ["telemetry_derived.event_events_v1"],
```

dags/bqetl_internet_outages.py
```diff
@@ -32,6 +32,18 @@
...
+        external_dag_id="copy_deduplicate",
+        external_task_id="copy_deduplicate_main_ping",
+        execution_delta=datetime.timedelta(seconds=7200),
+        check_existence=True,
+        mode="reschedule",
+    )
+
+    internet_outages__global_outages__v1.set_upstream(
+        wait_for_copy_deduplicate_main_ping
+    )
```
jklukas, line 46: Ideally, we'd be able to avoid defining this extra sensor since we know that this query depends on clients_daily, and that clients_daily already depends on copy_dedupe_main_ping. Not necessary here, but would be nice to have.

dags/bqetl_nondesktop.py
jklukas: Indeed, looks like it hits only core pings. This is probably an oversight. I will file an issue about it. But for this PR, removing this dependency looks correct.

jklukas: https://github.com/mozilla/bigquery-etl/issues/1184

Issue #1130, opened 2020-07-07
Run ETL for all Fenix products in Glam


emtwo:
Right now only the release ETL is run.

Ben-Wu:
ETL now running for `org_mozilla_firefox`, `org_mozilla_firefox_beta`, `org_mozilla_fenix`, and `org_mozilla_fennec_aurora`

PRs:
https://github.com/mozilla/bigquery-etl/pull/1183
https://github.com/mozilla/telemetry-airflow/pull/1098

PR #1183, opened 2020-07-17
Generalize fenix glam generate and run code

Ben-Wu:
airflow pr: https://github.com/mozilla/telemetry-airflow/pull/1098
issue: https://github.com/mozilla/bigquery-etl/issues/1130

Generate queries aren't checked in in order to keep things clean since they'll be generated in the airflow task.  Let me know if they should be.

The default product values of `org_mozilla_fenix` were removed so that product needs to be set explicitly.  As far as I can tell this shouldn't affect anything other than the airflow tasks which I've updated.


bigquery_etl/glam/generate.py
```diff
@@ -78,9 +78,9 @@ def main():
     if not dataset_path.is_dir():
         raise NotADirectoryError(f"path to {dataset_path} not found")
 
-    build_date_udf_mapping = dict(
-        org_mozilla_fenix="`moz-fx-data-shared-prod`.udf.fenix_build_to_datetime"
-    )
+    build_date_udf_mapping = {
+        args.prefix: "`moz-fx-data-shared-prod`.udf.fenix_build_to_datetime"
```
acmiyaguchi, line 82: If the product is not fenix, this is the wrong udf to use. If this returns null by default though, it should be fine. This can be fixed easily if broken since it's not used in the core aggregation.

script/glam/export_csv
```diff
@@ -8,7 +8,12 @@ src_project=${SRC_PROJECT:-glam-fenix-dev}
 # TODO: glam-dev-bespoke-nonprod-dataops-mozgcp-net
 dst_project=${DST_PROJECT:-$src_project}
 dataset=${DATASET:-glam_etl_dev}
-product=${PRODUCT:-org_mozilla_fenix}
+product=${PRODUCT}
```
acmiyaguchi, line 11: A shorthand uses `?` substituton
```suggestion
product=${PRODUCT?PRODUCT must be defined}
```

script/glam/generate_glean_sql
```diff
@@ -70,10 +70,15 @@ function write_clients_daily_aggregates {
 function main() {
     cd "$(dirname "$0")/../.."
 
-    local project=${PROJECT:-moz-fx-data-shared-prod}
-    local product=${PRODUCT:-org_mozilla_fenix}
+    local project=${SRC_PROJECT:-moz-fx-data-shared-prod}
```
acmiyaguchi, line 73: What's the reason for the change in variable name?

bigquery_etl/glam/generate.py
Ben-Wu: Oh that's true.  I'll add the the specific ids

script/glam/export_csv
Ben-Wu: Wow this is useful.  I had no idea

script/glam/generate_glean_sql
Ben-Wu: This is so it doesn't collide with `PRODUCT` in `run_glam_sql` which is the destination project.  Keeping it separate lets us set source and destination separately in the airflow task

script/glam/export_csv
acmiyaguchi: I've written more bash than I've cared for. The Linux Documentation Project has a good reference on these parameter substitutions: https://www.tldp.org/LDP/abs/html/parameter-substitution.html

PR #1185, opened 2020-07-20
Add langpack data to `amo_dev.amo_stats_dau`

willdurand:
Langpacks are Mozilla-owned add-ons and their GUIDs (`addon_id`) are normalized.

### DIFF
--- a/sql/amo_dev/amo_stats_dau_v2/query.sql
+++ b/sql/amo_dev/amo_stats_dau_v2/query.sql
@@ -23,3 +23,4 @@ WHERE
     'notes@mozilla.com',
     '1b2383b324c8520974ee097e46301d5ca4e076de387c02886f1c6b1503671586@pokeinthe.io' -- Laboratory
   )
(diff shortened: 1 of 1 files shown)


willdurand:
cc @jklukas

sql/amo_dev/amo_stats_dau_v2/query.sql
```diff
@@ -23,3 +23,4 @@ WHERE
     'notes@mozilla.com',
     '1b2383b324c8520974ee097e46301d5ca4e076de387c02886f1c6b1503671586@pokeinthe.io' -- Laboratory
   )
+  OR addon_id LIKE 'langpack-%@firefox.mozilla.org'
```
jklukas, line 26: We'll need to wrap this in parentheses so that we get something like:

```
AND (
  addon_id IN (...)
  OR addon_id LIKE ...
)
```

jklukas:
This table is now backfilled with the new logic.

willdurand:
> This table is now backfilled with the new logic.

Thank you!

PR #1190, opened 2020-07-21
Bug 1654330 AMO install stats based on event pings

jklukas:


sql/amo_prod/amo_stats_installs_v2/init.sql
```diff
@@ -0,0 +1,50 @@
...
+    udf.get_key(event_map_values, 'addon_id') AS addon_id,
+    -- `step` is needed to only get one record for 1 download/install.
+    udf.get_key(event_map_values, 'step') AS extra_step,
+    -- We need to filter out installs not initiated from AMO (or DISCO).
+    udf.get_key(event_map_values, 'source') AS extra_source,
+    -- We retain `method` here to show how we could retrieve
+    -- attribution data instead (this isn't collected yet).
+    -- For AMO, we do NOT need this field. We'll have something
+    -- like `extra_attribution` or `extra_utm_params` probably.
+    udf.get_key(event_map_values, 'method') AS extra_method,
```
willdurand, line 22: In this case, `method` contains things like `link` or `amWebAPI`, etc. We could think of `method` here like `utm_source` in the future, so `extra_method` should be later "converted" to a column that contains:

```
{
  'link': 1234,
  'amWebAPI': 4567
}
```

(similar to `dau_by_app_version` in `amo_stats_dau` for example)

sql/amo_prod/amo_stats_installs_v2/init.sql
```diff
@@ -0,0 +1,50 @@
...
+--
+SELECT
+  submission_date,
+  addon_id,
+  count(*) AS total_downloads,
+  [STRUCT('unknown' AS key, count(*) AS value)] AS downloads_per_source,
+  [STRUCT('unknown' AS key, count(*) AS value)] AS downloads_per_medium,
+  [STRUCT('unknown' AS key, count(*) AS value)] AS downloads_per_content,
+  [STRUCT('unknown' AS key, count(*) AS value)] AS downloads_per_campaign,
+  ARRAY_AGG(DISTINCT extra_method IGNORE NULLS) AS extra_methods
```
willdurand, line 38: This shouldn't be exposed in the "final" table structure.

sql/amo_prod/amo_stats_installs_v2/init.sql
```diff
@@ -0,0 +1,50 @@
...
+FROM
+  installs
+WHERE
+  submission_date = '2020-06-30'
```
willdurand, line 42: Should we keep this?

sql/amo_prod/amo_stats_installs_v2/init.sql
jklukas: Do you want me to go ahead and create a `downloads_by_method` column? Or should we just remove this completely?

sql/amo_prod/amo_stats_installs_v2/init.sql
jklukas: I will clean this up before finalizing.

sql/amo_prod/amo_stats_installs_v2/init.sql
willdurand: I think we can remove it completely for now.

willdurand:
@jklukas something that could be useful to us is having this table in `amo_dev` too

jklukas:
Removed `method`, cleaned up, and created a `dev` version of the table.

willdurand:
> Removed `method`, cleaned up, and created a `dev` version of the table.

@jklukas thanks to the dev table, I was able to write most features for AMO (behind a feature flag). While we still don't have attribution data (we're not yet collecting it), I think it's fine to go ahead and get some new data if it's not too much effort (we don't have very recent data anymore).

jklukas:
> I think it's fine to go ahead and get some new data if it's not too much effort (we don't have very recent data anymore).

I backfilled the prod and dev tables up to 2020-08-03.

jklukas:
We now have v3 tables that support this.

PR #1193, opened 2020-07-22
Initial feature_usage table.

jmccrosky:
This is a little ugly.  I think it would be better to generate the query programatically with a specification of how to extract all the features of interest, but would like to get this moving so I can continue with my work.


sql/telemetry/feature_usage/query.sql
```diff
@@ -0,0 +1,114 @@
...
+  COALESCE(filled_password, 0) AS filled_password,
+  COALESCE(remember_password, 0) AS remember_password,
+  COALESCE(sync_configured, 0) AS sync_configured
+FROM
+  (
+    SELECT
+      client_id,
+      submission_date,
+      IF(event_method = 'show' AND event_object = 'protection_report', 1, 0) AS protection_report,
+      IF(event_category = 'pictureinpicture' AND event_method = 'create', 1, 0) AS pip,
```
jklukas, line 18: IIUC, the intention is that we want one row in the final table per client_id per day. If so, each one of these subqueries should `GROUP BY submission_date, client_id` and each of these fields would be an aggregation like:

```
LOGICAL_OR(event_category = 'pictureinpicture' AND event_method = 'create') AS pip,
```

For cost and query efficiency, each of these features should be encoded as a boolean rather than as an integer (1 bytes vs. 8 bytes).

sql/telemetry/feature_usage/query.sql
```diff
@@ -0,0 +1,114 @@
...
+    FROM
+      `moz-fx-data-shared-prod.telemetry.events`
+    WHERE
+      submission_date = @submission_date
+      AND normalized_channel = "release"
```
jklukas, line 23: Are we sure we want to restrict this to just release rather than allowing all channels as we do for KPIs?

sql/telemetry/feature_usage/query.sql
```diff
@@ -0,0 +1,114 @@
...
+  ) events_t
+FULL JOIN
```
jklukas, line 25: These joins may end up being untenable for the entire release population. Is there any chance we can use a 1% sample for this work, or at least start with a 1% sample?

sql/telemetry/feature_usage/query.sql
```diff
@@ -0,0 +1,114 @@
...
+  (
+    SELECT
+      client_id,
+      submission_date,
+      IF(scalar_parent_browser_engagement_max_concurrent_tab_pinned_count > 0, 1, 0) AS pinned_tab,
+    FROM
+      `moz-fx-data-shared-prod.telemetry.main_summary`
```
jklukas, line 32: We also need to group by client_id for main_summary, since there can be multiple main pings per client per day

jklukas: Also, we should prefer to hit `telemetry.main` directly rather than `main_summary` unless we have a specific reason to use main_summary. There are a few fields in main_summary extracted from additional_properties, for example, that would be expensive to extract from `main` directly.

sql/telemetry/feature_usage/query.sql
```diff
@@ -0,0 +1,114 @@
...
+FULL JOIN
+  (
+    SELECT
+      client_id,
+      submission_date,
+      IF(is_default_browser, 1, 0) AS default_browser,
+      IF(scalar_parent_browser_engagement_total_uri_count_sum > 10, 1, 0) AS many_uri,
+      IF(sync_count_desktop_sum > 0, 1, 0) AS sync_configured,
+    FROM
+      `moz-fx-data-shared-prod.telemetry.clients_daily`
```
jklukas, line 49: If we already have to calculate some aggregates over main_summary, then we should avoid joining with clients_daily. These fields can all be expressed as aggregates on top of main_summary.

sql/telemetry/feature_usage/query.sql
```diff
@@ -0,0 +1,114 @@
...
+  ) cd_t
+ON
+  events_t.client_id = cd_t.client_id
+  AND events_t.submission_date = cd_t.submission_date
+FULL JOIN
+  (
+    SELECT
+      client_id,
+      submission_date,
+      1 AS filled_password
```
jklukas, line 62:
```suggestion
      true AS filled_password
```

sql/telemetry/feature_usage/query.sql
```diff
@@ -0,0 +1,114 @@
...
+    FROM
+      (
+        (
+          SELECT
+            client_id,
+            submission_date
+          FROM
+            `moz-fx-data-shared-prod.telemetry.main_summary` ms_t
+          CROSS JOIN
+            ms_t.histogram_content_pwmgr_form_autofill_result res_a
```
jklukas, line 72: We can express these CROSS JOINs within the `main_summary` query. I will need to follow up with example code for that.

jklukas:
Here's a starting point for many of the refactorings I recommended. It joins just main_summary and events, uses booleans, etc.

```
SELECT
  events_t.client_id AS clients_id,
  events_t.submission_date AS submission_date,
  protection_report,
  pinned_tab,
  default_browser,
  many_uri,
  pip,
  filled_password,
  remember_password,
  sync_configured,
FROM
  (
    SELECT
      client_id,
      submission_date,
      logical_or(event_method = 'show' AND event_object = 'protection_report') AS protection_report,
      logical_or(event_category = 'pictureinpicture' AND event_method = 'create') AS pip,
    FROM
      `moz-fx-data-shared-prod.telemetry.events`
    WHERE
      AND normalized_channel = "release"
      group by submission_date, client_id
  ) events_t
FULL JOIN
  (
    SELECT
      client_id,
      submission_date,
      logical_or(scalar_parent_browser_engagement_max_concurrent_tab_pinned_count > 0) AS pinned_tab,
      SUM(
        (SELECT COUNTIF(key = 0) FROM UNNEST((histogram_parent_pwmgr_form_autofill_result))) +
        (SELECT COUNTIF(key = 0) FROM UNNEST((histogram_content_pwmgr_form_autofill_result)))        
        ) > 0 AS filled_password,
      SUM(
        (SELECT COUNTIF(key = 1) FROM UNNEST((histogram_parent_pwmgr_form_autofill_result))) +
        (SELECT COUNTIF(key = 1) FROM UNNEST((histogram_content_pwmgr_form_autofill_result)))        
        ) > 0 AS remember_password,
      udf.mode_last(ARRAY_AGG(is_default_browser ORDER BY `timestamp`)) AS default_browser,
      SUM(scalar_parent_browser_engagement_total_uri_count) > 10 AS many_uri,
      SUM(sync_count_desktop) > 0 AS sync_configured,

    FROM
      `moz-fx-data-shared-prod.telemetry.main_summary`
    WHERE
      AND normalized_channel = "release"
            group by submission_date, client_id

  ) ms_t
ON
  events_t.client_id = ms_t.client_id
  AND events_t.submission_date = ms_t.submission_date
```

It validates, but I have not vetted the content at all, so there may be issues with nulls, etc.

Also, this still does not fall back all the way to `telemetry.main`. We would need to basically just inject the same transformations used in the `main_summary` query to get this working with `main`.

sql/telemetry/feature_usage/query.sql
```diff
@@ -0,0 +1,114 @@
...
+  (
+    SELECT
+      client_id,
+      submission_date,
+      IF(scalar_parent_browser_engagement_max_concurrent_tab_pinned_count > 0, 1, 0) AS pinned_tab,
+    FROM
+      `moz-fx-data-shared-prod.telemetry.main_summary`
+    WHERE
+      submission_date = @submission_date
+      AND normalized_channel = "release"
```
relud, line 35:
```suggestion
      LOGICAL_OR(
        IF(scalar_parent_browser_engagement_max_concurrent_tab_pinned_count > 0, 1, 0)
      ) AS pinned_tab,
      LOGICAL_OR(
        0 IN (
          SELECT
            key
          FROM
            UNNEST(
              ARRAY_CONCAT(
                histogram_content_pwmgr_form_autofill_result,
                histogram_parent_pwmgr_form_autofill_result
              )
            )
        )
      ) AS filled_password,
    FROM
      `moz-fx-data-shared-prod.telemetry.main_summary`
    WHERE
      submission_date = @submission_date
      AND normalized_channel = "release"
    GROUP BY
      submission_date,
      client_id
```

sql/telemetry/feature_usage/query.sql
relud: example: https://github.com/mozilla/bigquery-etl/pull/1193/files#r459074119

sql/telemetry/feature_usage/query.sql
jmccrosky: Good point.  Will update.

sql/telemetry/feature_usage/query.sql
jmccrosky: I had been restricting my analysis to release, but this table may be useful for other applications, so I'll drop the restriction.

sql/telemetry/feature_usage/query.sql
jmccrosky: For my purposes, 1% is not possible.  Restricting to profiles that use particular features a particular number of days after first-seen-date makes numbers much smaller and the matching methods I use need fairly large sample sizes for statistical significance.

sql/telemetry/feature_usage/query.sql
jmccrosky: Will do.

sql/telemetry/feature_usage/query.sql
jmccrosky: Will do.

sql/telemetry/feature_usage/query.sql
jmccrosky: Cool!  Will update.

jmccrosky:
OK, I think I've addressed all the requested changes.  Please let me know.

I was thinking though: I originally started on this query as a quick hack to unblock my own work.  I didn't put a lot of thought into design.  I feel like here, there's some danger that this is going to become the standard feature usage table.  If it is, we should think about a few things:

 - Is the wide format the best choice?  Or should it be long, with a STRING feature field and a row per clientXdateXfeature?
 - Should it be reviewed by someone familiar with telemetry for each of the features?

My preference is to merge this and do a backfill so I can start on my work then, when data platforms has time to take this over, revisit all this, but if you feel we should do a little more thinking now, I understand.

jklukas:
> Is the wide format the best choice? Or should it be long, with a STRING feature field and a row per clientXdateXfeature?

From an efficiency perspective, the wide format is better both for storage size and performance.

> My preference is to merge this and do a backfill so I can start on my work then, when data platforms has time to take this over, revisit all this, but if you feel we should do a little more thinking now, I understand.

I am okay with moving forward with this to unblock your work, learning from it, and circling back to shore it up as it proves to be valuable. What do you have in mind for length of backfill?

sql/telemetry/feature_usage/query.sql
```diff
@@ -0,0 +1,83 @@
+SELECT
+  events_t.client_id AS clients_id,
+  events_t.submission_date AS submission_date,
+  COALESCE(protection_report, FALSE) AS protection_report,
+  COALESCE(pinned_tab, FALSE) AS pinned_tab,
+  COALESCE(default_browser, FALSE) AS default_browser,
+  COALESCE(many_uri, FALSE) AS many_uri,
```
jklukas, line 7: I would like to see each of these fields be named with a past-tense verb indicating an action that the user took, and for it to be as specific as possible. This becomes even more important as we turn this into a clients_last_seen-type table where we have to translate each of these names consistently into a `_bits` field name.

So this field would be `visited_10_uri` and the clients_last_seen field would be `visited_10_uri_bits`.

Other suggested names:

viewed_protection_report
used_pip
remembered_password

Outliers are these fields that don't indicate user actions, but rather represent configuration state, so these are more logical dimensions rather than usage indicators: 

default_browser
sync_configured

Are we considering these dimensions? Do we want to maintain 28 days of history for these, are would we just be keeping the most recently observed value? Can we make the distinction explicit?

sql/telemetry/feature_usage/query.sql
```diff
@@ -0,0 +1,83 @@
...
+  (
+    SELECT
+      client_id,
+      submission_date,
+      LOGICAL_OR(event_method = 'show' AND event_object = 'protection_report') AS protection_report,
+      LOGICAL_OR(event_category = 'pictureinpicture' AND event_method = 'create') AS pip,
+    FROM
+      `moz-fx-data-shared-prod.telemetry.events`
+    WHERE
+      submission_date = @submission_date
```
jklukas, line 22: What about sampling? Is that an option here?

sql/telemetry/feature_usage/query.sql
```diff
@@ -0,0 +1,83 @@
...
+                `moz-fx-data-shared-prod.udf.null_if_empty_list`(
+                  mozfun.hist.extract(
+                    payload.processes.content.histograms.pwmgr_form_autofill_result
+                  ).values
+                ),
+                `moz-fx-data-shared-prod.udf.null_if_empty_list`(
+                  mozfun.hist.extract(payload.histograms.pwmgr_form_autofill_result).values
+                )
+              )
+            )
```
jklukas, line 58:
```suggestion
            )
          WHERE
            value > 0
```

It's probably possible that an entry exists in the list of values with a 0 count, which I assume we wouldn't want to mark as active?

sql/telemetry/feature_usage/query.sql
```diff
@@ -0,0 +1,83 @@
...
+      `moz-fx-data-shared-prod.telemetry.main`
+    WHERE
+      DATE(submission_timestamp) = @submission_date
+    GROUP BY
+      submission_date,
+      client_id
+  ) main_t
+ON
+  events_t.client_id = main_t.client_id
+  AND events_t.submission_date = main_t.submission_date
```
relud, line 83:
```suggestion
USING
  (client_id, submission_date)
```

sql/telemetry/feature_usage/query.sql
```diff
@@ -0,0 +1,83 @@
+SELECT
+  events_t.client_id AS clients_id,
+  events_t.submission_date AS submission_date,
```
relud, line 3: if `JOIN` is via `USING` instead of `ON`, then this can be:
```suggestion
  clients_id,
  submission_date,
```

sql/telemetry/feature_usage/query.sql
```diff
@@ -0,0 +1,83 @@
...
+      LOGICAL_OR(
+        1 IN (
+          SELECT
+            key
+          FROM
+            UNNEST(
+              `moz-fx-data-shared-prod.udf.null_if_empty_list`(
+                mozfun.hist.extract(payload.histograms.pwmgr_prompt_remember_action).values
+              )
+            )
```
relud, line 70: null and empty lists behave the same within `UNNEST`
```suggestion
            UNNEST(
              mozfun.hist.extract(payload.histograms.pwmgr_prompt_remember_action).values
            )
```

sql/telemetry/feature_usage/query.sql
relud, line 58: this looks like it's trying avoid the case where `ARRAY_CONCAT` returns `NULL` if any parameter is `NULL`, but that means you don't want `null_if_empty_list`, you want the opposite: `IFNULL(..., [])` (i.e. `empty_list_if_null`):
```suggestion
            UNNEST(
              ARRAY_CONCAT(
                IFNULL(
                  mozfun.hist.extract(
                    payload.processes.content.histograms.pwmgr_form_autofill_result
                  ).values,
                  []
                ),
                IFNULL(
                  mozfun.hist.extract(payload.histograms.pwmgr_form_autofill_result).values,
                  []
                )
              )
            )
```

sql/telemetry/feature_usage/query.sql
jklukas: And it looks like we have a mispelling here: `clients_id`

sql/telemetry/feature_usage/query.sql
jmccrosky: I updated these.  You make a good point that there is a difference between features that are used discretely and those that persist until "turned off".  Although `has_pinned_tab` is sort of in the middle as it persists until the tab is closed.

Anyway, I'd like to treat them all the same.  I still want to look for patterns of co-occurrence of feature use across types and be able to determine if the feature was used/enabled at all during some period.

sql/telemetry/feature_usage/query.sql
jmccrosky: Not for my current application.

jmccrosky:
Thanks all.  I think I've addressed everything.  Please take a look.

jklukas:
I just pushed a change to move this to the telemetry_derived dataset (only views are supposed to go in `telemetry`) and to refactor the query to push subqueries to CTEs. I ran a test query, and was pleased to see it only scans about 50 GB per day, so will be much faster than I expected. It scans many fewer columns than clients_daily, etc.

But that also made me realize we have no dimensions at this point (country, channel, etc.). Do we care at this point? It'll be a little bit of a pain to figure out the logic, since we could have a client that appears only in main or only in events, so we'll have to aggregate dimension values across the two ping types.

PR #1195, opened 2020-07-22
ExternalTaskSensor pool

scholtzan:
There is now a `DATA_ENG_EXTERNALTASKSENSOR` pool in Airflow which can be used for `ExternalTaskSensors`. Currently, it has 10 slots, we might need to increase this number? Or first start testing by assigning the pool to only a couple of task sensors before going all in?


scholtzan:
I opened https://github.com/mozilla/telemetry-airflow/pull/1103 for adding the changes to telemetry-airflow.

And https://github.com/mozilla/telemetry-airflow/pull/1104 to undo them, if needed.

PR #1196, opened 2020-07-22
Bug 1648241 - Create mobile search clients last seen

Ben-Wu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1648241

**Very** derivative of existing search clients last seen so tagging future Frank for review.  Mobile version doesn't sample because total data is small


sql/search_derived/mobile_search_clients_last_seen_v1/query.sql
```diff
@@ -0,0 +1,135 @@
+WITH _derived_search_cols AS (
+  SELECT
+    udf.normalize_search_engine(engine) AS short_engine,
```
jklukas, line 3: Is it possible to delay normalizing the search engine to the view? If it's not normalized in the daily table, do we need to normalize it here?

sql/search_derived/mobile_search_clients_last_seen_v1/query.sql
```diff
@@ -0,0 +1,135 @@
...
+    * EXCEPT (short_engine)
+  FROM
+    _derived_search_cols
+),
+_grouped AS (
+  -- Here we get a single row per-client, containing
+  -- info from each engine, as well as overall info
+  SELECT
+    client_id,
+    ANY_VALUE(sample_id) AS sample_id,
```
jklukas, line 34: I'd prefer to put `sample_id` in the GROUP BY so that we can avoid using `ANY_VALUE`

```suggestion
    sample_id,
```

sql/search_derived/mobile_search_clients_last_seen_v1/query.sql
Ben-Wu: The difference is that the normalized engine is used in the computation here to get the monthly search totals.  We could however normalize in the view and then sum up the totals per normalized engine there.  I'll try that out

Ben-Wu: Also just noticed I didn't make the view

Ben-Wu:
Latest commit moves search engine normalization to view using a new udf.  udf ended up being a little more complex than I expected

udf/add_searches_by_index.sql
```diff
@@ -0,0 +1,80 @@
...
+        SUM(search.ad_click) AS ad_click,
+      FROM
+        UNNEST(searches) search
+      GROUP BY
+        index
+      ORDER BY
+        index
+    )
+    SELECT AS STRUCT
+      ARRAY_AGG(total_searches) total_searches,
```
jklukas, line 32: The ARRAY_AGG docs state:

> If ORDER BY is not specified, the order of the elements in the output array is non-deterministic, which means you might receive a different result each time you use this function

So I think we need to put the ORDER BY in the ARRAY_AGG calls rather than in the CTE above to guarantee ordering (even though your tests seem to indicate ordering is stable).

udf/normalize_monthly_searches.sql
```diff
@@ -0,0 +1,121 @@
...
+            engine.value.search_with_ads[OFFSET(index)],
+            engine.value.ad_click[OFFSET(index)],
+            index
+          )
+        )
+      ) AS value,
+    FROM
+      UNNEST(engine_searches) AS engine
+    CROSS JOIN
+      UNNEST(engine.value.total_searches) total_searches
```
jklukas, line 36:
```suggestion
      UNNEST(engine.value.total_searches) AS total_searches
```

PR #1198, opened 2020-07-23
[Bug 1654024] Script for monitoring partition sizes of stable tables

scholtzan:
I think having a Python script instead of a pure SQL query does make it easier to handle backfill cases. The script is scheduled in https://github.com/mozilla/telemetry-airflow/pull/1105

Before merging, the target table needs to be created for the query to run successfully.


sql/monitoring/telemetry_stable_sizes_v1/query.py
```diff
@@ -0,0 +1,80 @@
...
+    ).result()
+
+
+def main():
+    args = parser.parse_args()
+    client = bigquery.Client(args.project)
+
+    stable_tables = get_tables(client, args.project, args.dataset)
+
+    partition_sizes = [
```
jklukas, line 60: Can we parallelize this step? We're going to be spending all our time waiting on the dry run queries to return. And if we do parallelize, can we expand this job to take care of all stable datasets rather than doing just `telemetry_stable`?

sql/monitoring/stable_table_sizes_v1/query.py
```diff
@@ -0,0 +1,108 @@
...
+from argparse import ArgumentParser
+from fnmatch import fnmatchcase
+from google.cloud import bigquery
+from functools import partial
+from multiprocessing.pool import ThreadPool
+
+parser = ArgumentParser(description=__doc__)
+parser.add_argument("--date", required=True)  # expect string with format yyyy-mm-dd
+parser.add_argument("--project", default="moz-fx-data-shared-prod")
+parser.add_argument("--dataset", default="*_stable")  # pattern
```
scholtzan, line 14: Get sizes for tables in all stable datasets

sql/monitoring/stable_table_sizes_v1/query.py
```diff
@@ -0,0 +1,108 @@
...
+    args = parser.parse_args()
+    client = bigquery.Client(args.project)
+
+    stable_datasets = [
+        dataset.dataset_id
+        for dataset in list(client.list_datasets())
+        if fnmatchcase(dataset.dataset_id, args.dataset)
+    ]
+
+    with ThreadPool(8) as p:
```
scholtzan, line 85: Parallelize

jklukas: I think we can safely dial this up to 20.

jklukas:
These changes look great 👍

scholtzan:
I also created `moz-fx-data-shared-prod.monitoring.stable_table_sizes_v1`

PR #1201, opened 2020-07-24
Include sample_id in feature_usage

jklukas:
Used as clustering key and allows sampling


jklukas:
I am backfilling now with this structure.

PR #1202, opened 2020-07-24
Add sqrt to LTV calc

benmiroglio:
Performs a SQRT on the raw LTV values before normalization to handle large outliers.

PR #1203, opened 2020-07-24
Add ltv view without revenue fields

benmiroglio:
Adding a view for LTV that doesn't require permissions for the revenue tables. @fbertsch Please take a look when you get back. I'm not confident this is reading from the correct table.

### DIFF
--- a/sql/telemetry/client_ltv/view.sql
+++ b/sql/telemetry/client_ltv/view.sql
@@ -1,0 +1,20 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.client_ltv`
+AS
+SELECT
+  * EXCEPT (
+    avg_client_ad_click_value,
+    avg_client_search_with_ads_value,
+    avg_client_search_value,
+    avg_client_tagged_search_value,
+    ltv_ad_clicks_current,
+    ltv_search_with_ads_current,
+    ltv_search_current,
+    ltv_tagged_search_current,
+    ltv_ad_clicks_future,
+    ltv_search_with_ads_future,
+    ltv_search_future,
+    ltv_tagged_search_future
+  )
+FROM
+  `moz-it-eip-revenue-users.ltv_derived.client_ltv_v1`
--- a/bigquery_etl/dryrun.py
+++ b/bigquery_etl/dryrun.py
@@ -60,6 +60,7 @@ SKIP = {
     "sql/apple_app_store/metrics_by_web_referrer/query.sql",
     "sql/apple_app_store/metrics_total/query.sql",
     "sql/monitoring/telemetry_distinct_docids_v1/query.sql",
+    "sql/revenue_derived/client_ltv_normalized/query.sql",
     # Already exists (and lacks an "OR REPLACE" clause)
     "sql/org_mozilla_firefox_derived/clients_first_seen_v1/init.sql",
     "sql/org_mozilla_firefox_derived/clients_last_seen_v1/init.sql",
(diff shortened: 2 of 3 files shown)


sql/telemetry/client_ltv/view.sql
```diff
@@ -0,0 +1,20 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.client_ltv`
+AS
+SELECT
+  * EXCEPT (
```
fbertsch, line 5: I noticed that we're missing normalized tagged searches LTV. Want to add that to the other query so it's available here?

sql/telemetry/client_ltv/view.sql
```diff
@@ -0,0 +1,20 @@
+CREATE OR REPLACE VIEW
```
fbertsch, line 1: I think this should be a query rather than a view. We don't really have great capabilities for authorized views, and the other table is in a project we don't control which makes it more difficult.

sql/telemetry/client_ltv/view.sql
benmiroglio: How so? It looks to be there: https://github.com/mozilla/bigquery-etl/pull/1202/files/0906389fa0d5dbb728763892e74fe8a181bddad1#

sql/telemetry/client_ltv/view.sql
benmiroglio: Makes sense. Just to confirm I see that as meaning:

* move this over to the `telemetry_derived`dir
*  remove ```CREATE OR REPLACE VIEW AS `moz-fx-data-shared-prod.telemetry.client_ltv`
AS```
* rename this file as `query.sql`

Let me know if you something else in mind!

sql/telemetry/client_ltv/view.sql
fbertsch: Down below that: https://github.com/mozilla/bigquery-etl/blob/master/sql/revenue_derived/client_ltv_v1/query.sql#L246-L267

Looks like future tagged search is there, but not current.

sql/telemetry/client_ltv/view.sql
fbertsch: Nope, we should be fine keeping this in `revenue_derived`, we can keep that as a new dataset.

Other than that, what you said is correct. You can even try to [schedule the query yourself](https://github.com/mozilla/bigquery-etl#scheduling-queries-in-airflow)!

sql/telemetry/client_ltv/view.sql
benmiroglio: ah, I see. will fix!

sql/telemetry/client_ltv/view.sql
benmiroglio: `sql/revenue_derived/client_ltv_v1/query.sql` already exists (the locked down dataset). Would you prefer this go under a new directory, i.e. `sql/revenue_derived/client_ltv_reduced/query.sql`?

fbertsch: Yup, that would make sense. I would perhaps call it `client_ltv_normalized`, what do you think?

benmiroglio: that is much better name. SGTM.

benmiroglio:
FYI, I plan to create a PR to schedule this in airflow, but probably won't get to it for a couple of days!

fbertsch (approved):
This will take a bit of finagling with permissions to get it to run correctly.

PR #1204, opened 2020-07-24
Add some lower-friction scheduled query docs

scholtzan:
Some docs about lower-friction scheduled queries.


README.md
```diff
@@ -222,7 +222,54 @@ labels:
...
+    ```
+      - `task_id`: name of task query depends on
+      - `dag_name`: name of the DAG the external task is part of
+      - `execution_delta`: time difference between the `schedule_intervals` of the external DAG and the DAG the query is part of
+- To generate all Airflow DAGs run `./script/generate_airflow_dags`
+  - Generated DAGs are located in the `dags/` directory
+  - Dependencies between queries scheduled in bigquery-etl and dependencies to stable tables are detected automatically
+- Generated DAGs need to be manually copied to [telemetry-airflow](https://github.com/mozilla/telemetry-airflow) 
+
+More instructions for scheduling queries in Airflow can be found in this
```
scholtzan, line 272: Not sure about this cookbook. Either we will have to update it or will it be replaced by a new cookbook @wlach is working on?

README.md
```diff
@@ -222,7 +222,54 @@ labels:
...
+      start_date: '2020-04-05'  # YYYY-MM-DD
+      email: ['example@mozilla.com']
+      retries: 2    # number of retries if the query execution fails
+      retry_delay: 30m  
+  ```
+  - All DAG names need to have `bqetl_` as prefix.
+  - `schedule_interval` is either defined as a [CRON expression](https://en.wikipedia.org/wiki/Cron) or alternatively as one of the following [CRON presets](https://airflow.readthedocs.io/en/latest/dag-run.html): `once`, `hourly`, `daily`, `weekly`, `monthly`
+  - `start_date` defines the first date for which the query should be executed
+  - `email` lists email addresses alerts should be sent to in case of failures when running the query
+- To schedule a specific query, it requires a `metadata.yaml` file that includes a `scheduling` section, for example:
```
wlach, line 242:
```suggestion
- To schedule a specific query, add a `metadata.yaml` file that includes a `scheduling` section, for example:
```

README.md
wlach: We should probably merge some scheduling info into the new cookbook that @gkabbz and myself are working on, yeah. Thankfully scheduling is now easy enough that we can probably just include it as a last step in the instructions on how to add a query. For now, I'd just remove this link as it's misleading (I'll try to remember to add something back after we merge our work in progress).

README.md
```diff
@@ -222,8 +222,52 @@ labels:
 Scheduling Queries in Airflow
 ---
 
-Instructions for scheduling queries in Airflow can be found in this
-[cookbook](https://docs.telemetry.mozilla.org/cookbooks/bigquery-airflow.html).
+- bigquery-etl has tooling to automatically generate Airflow DAGs for scheduling queries.
+- Every query must be scheduled as part of a DAG that is specified in `dags.yaml`.
```
jklukas, line 226:
```suggestion
- To be scheduled, a query must be assigned to a DAG that is specified in `dags.yaml`.
```

Not every query is required to be scheduled.

README.md
```diff
@@ -222,8 +222,52 @@ labels:
...
+    default_args:
+      owner: example@mozilla.com
+      start_date: '2020-04-05'  # YYYY-MM-DD
+      email: ['example@mozilla.com']
+      retries: 2    # number of retries if the query execution fails
+      retry_delay: 30m  
+  ```
+  - All DAG names need to have `bqetl_` as prefix.
+  - `schedule_interval` is either defined as a [CRON expression](https://en.wikipedia.org/wiki/Cron) or alternatively as one of the following [CRON presets](https://airflow.readthedocs.io/en/latest/dag-run.html): `once`, `hourly`, `daily`, `weekly`, `monthly`
+  - `start_date` defines the first date for which the query should be executed
```
jklukas, line 240: It would be worth noting here that Airflow will not automatically backfill older dates if `start_date` is set in the past, as readers may assume this.

PR #1208, opened 2020-07-29
Rewrite duplicated bits28 udfs to mozfun shims

asiOvOtus:
The overall plan is to migrate all the udfs to mozfun. Since there are already a few functions duplicated in the both codebase, this PR remove the duplicated "bits28_" functions by wrapper around mozfun functions.

This is an ongoing work and will be followed by a few other PRs shortly.

Tested:

- Unit tests
- scripts/format_sql


asiOvOtus:
Thanks for pointing it out. I also thought about it in a while, but since this case can be caught by CI, I will make sure to modify the caller queries once it becomes a problem.

PR #1211, opened 2020-08-03
Rewrite duplicated map udfs to mozfun shims

asiOvOtus:
This is a trivial follow-up PR after #1208. It removes the duplicated `map/get_key` and `map/get_key_with_null` by wrapper around mozfun functions.


asiOvOtus:
Feedback wanted :D @acmiyaguchi @scholtzan

PR #1214, opened 2020-08-05
Add parsed_x_source_tags to metadata in user-facing views

jklukas:
As discussed in https://bugzilla.mozilla.org/show_bug.cgi?id=1657360#c1

### DIFF
--- a/mozfun/norm/metadata/udf.sql
+++ b/mozfun/norm/metadata/udf.sql
@@ -16,34 +16,61 @@ CREATE OR REPLACE FUNCTION norm.metadata(metadata ANY TYPE) AS (
               -- Even though it's not part of the spec, many clients append
               -- '+00:00' to their Date headers, so we strip that suffix.
               REPLACE(metadata.header.`date`, 'GMT+00:00', 'GMT')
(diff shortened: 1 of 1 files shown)


mozfun/norm/metadata/udf.sql
```diff
@@ -16,34 +16,61 @@ CREATE OR REPLACE FUNCTION norm.metadata(metadata ANY TYPE) AS (
...
               -- '+00:00' to their Date headers, so we strip that suffix.
               REPLACE(metadata.header.`date`, 'GMT+00:00', 'GMT')
-            ) AS parsed_date
+            ) AS parsed_date,
+            ARRAY(
+              SELECT
+                TRIM(t)
+              FROM
+                UNNEST(SPLIT(metadata.header.x_source_tags, ',')) t
+            ) AS parsed_source_tags
```
jklukas, line 25: I'm open to feedback on naming here. This could be `parsed_x_source_tags`.

fbertsch: `parsed_x_source_tags` might be better, since it makes it more obvious what it is derived from.

PR #1217, opened 2020-08-06
Bug 1655146 Add impression_stats_by_experiment

jklukas:
This table will be a bit larger than impression_stats_flat, about 1 TB/day
uncompressed. But it will be much more efficient for the class of queries
that are concerned with evaluating performance of a single experiment.

### DIFF
--- a/sql/activity_stream/impression_stats_by_experiment/view.sql
+++ b/sql/activity_stream/impression_stats_by_experiment/view.sql
@@ -1,0 +1,13 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.activity_stream.impression_stats_by_experiment`
+AS
+SELECT
+  tile_id,
+  tile_id_types.type AS tile_type,
+  stats.* EXCEPT (tile_id)
+FROM
+  `moz-fx-data-shared-prod.activity_stream_bi.impression_stats_by_experiment_v1` AS stats
+LEFT JOIN
+  `moz-fx-data-shared-prod.activity_stream.tile_id_types` AS tile_id_types
+USING
+  (tile_id)
--- a/sql/activity_stream_bi/impression_stats_by_experiment_v1/init.sql
+++ b/sql/activity_stream_bi/impression_stats_by_experiment_v1/init.sql
@@ -1,0 +1,30 @@
+CREATE OR REPLACE TABLE
+  activity_stream_bi.impression_stats_by_experiment_v1
+PARTITION BY
+  DATE(submission_timestamp)
+CLUSTER BY
+  experiment_id
+OPTIONS
+  (
+    require_partition_filter = TRUE, --
+    partition_expiration_days = 180
+  )
+AS
+SELECT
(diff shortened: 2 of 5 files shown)


sql/activity_stream/impression_stats_by_experiment/view.sql
```diff
@@ -0,0 +1,13 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.activity_stream.impression_stats_by_experiment`
```
jklukas, line 2: @kirill-demtchouk - This view contains the logical tile type info and would be the target for all your experiment analysis queries. Does this name look good?

kirill-demtchouk: Looks good! I see Mozilla uses Airflow too 👍 

One question on permissions for this new table: 

In addition to Redash access, can we also make sure that this table is accessible from our BI tool (Mode Analytics)? We use this service account for that tool: external-mode@moz-fx-data-bq-pocket.iam.gserviceaccount.com.

jklukas: That service account is granted read access to the `activity_stream_bi` dataset, so any new tables defined there will be accessible. This view won't be accessible, but the underlying `impression_stats_by_experiment_v1` table will be. Does that meet your needs? If having the view would be useful, we can duplicate it in `activity_stream_bi`.

kirill-demtchouk: That should be fine. The only difference between the view and the table is the join against `tile_id_types`.

sql/activity_stream_bi/impression_stats_by_experiment_v1/init.sql
```diff
@@ -0,0 +1,30 @@
...
+OPTIONS
+  (
+    require_partition_filter = TRUE, --
+    partition_expiration_days = 180
+  )
+AS
+SELECT
+  submission_timestamp,
+  experiment.key AS experiment_id,
+  experiment.value.branch AS experiment_branch,
```
jklukas, line 16: I know that your existing queries coalesce experiment info with `shield_id`. I assume the `shield_id` variants are no longer relevant. Is that correct?

Is there a particular amount of backfill that's relevant here? Would backfilling a month or two be sufficient, or do you have use cases where you need to look back on the full 6 months?

kirill-demtchouk: Yup, `shield_id` is in there from the period where we had a mix of data with `shield_id`/`experiments`.

One month of backfill should be sufficient!

jklukas:
I've backfilled the most recent 35 days of data into the table. Once this is merged, Airflow will pick it up and add new daily partitions alongside `impression_stats_flat`

kirill-demtchouk:
@jklukas I will spend some time early next week converting existing reporting to use the new table. Thanks for setting this up!

PR #1219, opened 2020-08-06
Don't include most recent month in revenue calc

fbertsch:


sql/revenue_derived/client_ltv_v1/query.sql
```diff
@@ -33,7 +33,7 @@ with_exploded_months AS (
     history.element.value.tagged_searches.list[OFFSET(off)].element AS tagged_sap,
   FROM
     with_exploded_history,
-    UNNEST(GENERATE_ARRAY(0, 11)) AS off
+    UNNEST(GENERATE_ARRAY(0, 10)) AS off
```
fbertsch, line 36: Once we add the extra month at the beginning of the array, this will change back to 11, and line 28 will change ot `12 - off` so we still don't include the most recent month.

fbertsch: We do an inner join with revenue data, so this will drop the month even after that.

benmiroglio: thanks for the added detail! LGTM

PR #1221, opened 2020-08-07
Add views for logical Fenix app ids in GLAM ETL

acmiyaguchi:
[spec](https://docs.google.com/document/d/1O_hDT8po5iEQcle62fN_eno4liVhZOhFXnKRBAoBaWk)


acmiyaguchi:
The airflow dag will probably remain in two parts, since it's easier to deal with the dependencies in this way. I've added an `all` option to `STAGE` though, which should make it easier to run this locally.

acmiyaguchi:
I've done some testing of the `STAGE=all` option, which requires the app_id in question to be manually added to the dictionary in `generate.py`. This seems to work as expected.

PR #1223, opened 2020-08-10
Add features

jmccrosky:
Add two features I need to work with.  Can we backfill with this?  It'll make my life much easier and save money as I don't keep pulling these features myself each time I run.

I realize I'm changing schema.  Would prefer to continue to call it v1 while developing, but if the schema change is a problem, let me know and I'll make this v2.


jklukas:
> I realize I'm changing schema. Would prefer to continue to call it v1 while developing, but if the schema change is a problem, let me know and I'll make this v2.

This is just adding fields, which is generally the happy path for schema evolution. We can make these changes without breaking any existing queries, so no need to change version.

jklukas:
I'll kick off a backfill today.

jklukas:
This is now backfilled from present back through 2019-09-01.

PR #1224, opened 2020-08-10
Bug 1654330 AMO download stats by UTM params

jklukas:
This new `v3` table uses events and replaces the `v1` table that was based on clients_daily.


sql/amo_dev/amo_stats_installs_v3/query.sql
```diff
@@ -0,0 +1,22 @@
+SELECT
+  *
+FROM
+  amo_prod.amo_stats_installs_v3
+WHERE
+  submission_date = @submission_date
+  AND (
+    hashed_addon_id IN (
+      '{db55bb9b-0d9f-407f-9b65-da9dd29c8d32}', -- :willdurand
```
jklukas, line 9: `hashed_addon_id` is a problem for the dev table. We either need the list of hashed IDs to insert here, or we need to be able to join back to unhashed IDs before doing this filtering.

jklukas:
> Why not keep the v1 query in the repository instead of removing it?

It's never actually been used in practice, but I suppose no harm in keeping it. I'll remove the scheduling section from the metadata so that while the query still exists, it no longer will be present in the DAG.

sql/amo_dev/amo_stats_installs_v3/query.sql
willdurand: We could hash those add-on IDs if needed.

sql/amo_prod/amo_stats_installs_v3/query.sql
```diff
@@ -0,0 +1,165 @@
...
+    hashed_addon_id,
+    submission_date,
+    utm_campaign,
+    utm_content,
+    utm_source,
+    utm_medium
+),
+-- The addon_id is hashed due to string size limitations in events;
+-- we create a "unhashing" lookup table here from the DAU table, which should
+-- include all the same addon_ids that we see in install_stats events.
```
willdurand, line 27: This is an interesting approach but are we 100% sure we'll have all the add-on IDs here?

jklukas: Logically, they should all be there, but there may be a small possibility of skew across the date boundary depending on when exactly different pings arrive.

I was hoping to simplify the system, but you're right to question this. We should err on the side of stability. I will refactor this table to be clustered on hashed_addon_id. I think I will still do a join for the dev table, though, so that we don't have to manually hash IDs every time we want to update the dev list to include new addons.

willdurand: Thanks. I really like the idea, though, because that would simplify a bunch of things for AMO. I feel like it's worth exploring if we can validate our theory, but it shouldn't be a blocker.

Issue #947, opened 2020-05-01
UDFs for canonical naming of mobile applications


jklukas:
Mobile usage data that feeds into the mobile MAU KPI and into GUD relies on some ad hoc filtering in various views and scheduled queries to determine which pings contribute.

It's currently difficult to follow the trail of queries, and you need to copy and paste logic out of those queries to recreate this filtering and naming for new use cases.

I initially wanted to provision a table to join with to get this cleaned info, but I don't see a clear way to do that since canonical app names can depend on source ping type, channel, an `app_name` field, etc.

The interface would be something like the following:

```
udf.extract_canonical_name_core(app_name, os, channel)
RETURNS STRUCT<
  canonical_name STRING,
  canonical_channel STRING,
  contributes_to_2019_kpi BOOLEAN
  >

udf.extract_canonical_name_glean(dataset_name, channel)
RETURNS STRUCT<
  canonical_name STRING,
  canonical_channel STRING,
  contributes_to_2019_kpi BOOLEAN
  >
```

cc @jmccrosky

jmccrosky:
I'm not totally clear on how this would be used.  Can you clarify?  @gkabbz was working on implementing the mobile activation metric that started this discussion, so may have some thoughts to share.

jklukas:
Currently, we have logic in the `core_clients_last_seen` and `nondesktop_clients_last_seen` views that decides how to clean and filter app names.  We would refactor those so that they call these functions instead. Then the exact_mau and GUD queries would make use of those canonical names.

I may need to go ahead and implement this in a PR to make the discussion more concrete.

jmccrosky:
I understand I think.  I guess generally I'm supportive of having a single clients_last_seen view that works across all our products and trust you to choose the best way to get there.  If you want me to have more opinions than just that though, let me know.

jklukas:
I suppose my real intention here is to have a canonical place I can point to when the question comes up "which pings exactly are included for the mobile MAU KPI?" These UDFs would be the concrete implementation of that logic and would also serve as the most concrete documentation of the KPI.

jmccrosky:
Sounds good to me :)

magorlick:
[...]

PR #1226, opened 2020-08-11
Add mozfun.norm.fenix_app_info UDF

jklukas:
Relevant to #947

We will use this function for how we present Firefox Preview and Fenix in GUD.


udf/extract_canonical_glean_app_info.sql
```diff
@@ -0,0 +1,71 @@
+/*
```
acmiyaguchi, line 1: Should this UDF live inside of mozfun instead?

udf/extract_canonical_glean_app_info.sql
```diff
@@ -0,0 +1,71 @@
...
+  WHEN
+    app_id LIKE 'org_mozilla_fenix%'
+  THEN
+    IF(
+      udf.fenix_build_to_datetime(app_build_id) < '2020-07-03',
+      STRUCT('Firefox Preview', 'beta', 'org.mozilla.fenix'),
+      STRUCT('Fenix', 'nightly', 'org.mozilla.fenix')
+    )
+  ELSE
+    ERROR("Bad data")
```
acmiyaguchi, line 46: Is there a way to test for this case?

udf/extract_canonical_glean_app_info.sql
```diff
@@ -0,0 +1,71 @@
+/*
+
```
acmiyaguchi, line 2: Perhaps link to the historical naming doc for more context on naming.

udf/extract_canonical_glean_app_info.sql
jklukas: I would indeed like to get it into `mozfun` eventually, but I see a few blockers:

- It depends on `udf.fenix_build_to_datetime`, which we'd also need to move to mozfun
- It currently only supports Fenix
- This UDF still feels a bit volatile; it's not clear that the two arguments we have in here now are going to be sufficient long-term to disambiguate app names and channels

So I'm thinking of this a bit as an incubation stage before committing to an interface and a long-term name in mozfun. And I'm definitely open to suggestions on naming.

udf/extract_canonical_glean_app_info.sql
jklukas: Are you referring to https://docs.google.com/document/d/1O_hDT8po5iEQcle62fN_eno4liVhZOhFXnKRBAoBaWk/edit?ts=5f31bb6d#heading=h.z5xwoh2241fq ?

udf/extract_canonical_glean_app_info.sql
jklukas: I meant to fill this out with a better message. But no, I'm not aware of a way at this point to write a test case that exercises this. If we raise an error, the test fails.

udf/extract_canonical_glean_app_info.sql
acmiyaguchi: No, I was thinking this spreadsheet: https://docs.google.com/spreadsheets/d/18PzkzZxdpFl23__-CIO735NumYDqu7jHpqllo0sBbPA/edit#gid=0

udf/extract_canonical_glean_app_info.sql
acmiyaguchi: It seems like it would be better to just commit with a single location for this UDF, especially if it's intended to be used across projects. FWIW, it would be useful to have `fenix_build_to_datetime` in mozfun too.

jklukas:
The more time I spend thinking about this, the more uncomfortable I am with releasing this UDF as an interface we recommend folks use. There's just too much uncertainty about what inputs we'll need for future Glean apps (or even for future evolution of Fenix).

I think perhaps a better solution is to push logic into views. In `scripts/generate_views`, we would inject logic such that the user-facing views for Fenix apps would replace `normalized_channel` and would add `normalized_app_name` and `normalized_app_id` fields. We would also propagate those fields to the baseline_clients_daily and baseline_clients_last_seen tables.

There are still some details to work out, but it gives us the flexibility to change what fields we use for calculating app names and channels without impacting users. It also means users don't ever need to know about a UDF; the values are built into the user-facing views.

acmiyaguchi:
There are a couple of problems to note with using a view over the stable tables:

* `org_mozilla_fenix` is split across a build_id boundary which is defined within the nested `client_info` field. A view would need to filter on this field requiring a full scan of each partition and would not be able to take advantage of clustering unless explicitly added to the pipeline family metadata.
* The schemas across the products are not guaranteed to the same, so it may not be as simple as unioning the tables. This works the Glam ETL because there the table is flattened in such a way that there a row for each (client, metric name, values).

jklukas:
> There are a couple of problems to note with using a view over the stable tables

Your points are valid if we try to make a single view that unifies all fenix apps. I was imagining that we still have the 5 separate app IDs, and we'd be updating the logic for the user-facing views under those 5 datasets. It's still asking a lot for users to have to union together these views.

The `baseline_clients_daily` and `baseline_clients_last_seen` tables are similar to the Glam case where we are generating the tables consistently, and can expect the schemas to match exactly for all glean app ids, thus a union can work. But it would indeed be nice to be able to cluster efficiently on channel.

mozfun/norm/fenix_app_info/udf.sql
```diff
@@ -0,0 +1,50 @@
+CREATE OR REPLACE FUNCTION norm.fenix_app_info(app_id STRING, app_build_id STRING)
+RETURNS STRUCT<app_name STRING, channel STRING, app_id STRING> AS (
```
jklukas, line 2: Changed the `canonical_app_id` field to just `app_id` for simplicity. Particularly because the function name no longer uses "canonical".

mozfun/norm/fenix_app_info/udf.sql
```diff
@@ -0,0 +1,50 @@
...
+    app_id LIKE 'org_mozilla_fenix_nightly%'
+  THEN
+    STRUCT('Firefox Preview', 'nightly', 'org.mozilla.fenix.nightly')
+  WHEN
+    app_id LIKE 'org_mozilla_fenix%'
+  THEN
+    IF(
+      -- See udf.fenix_build_to_datetime for info on build_id format;
+      -- the build_id constant here corresponds to 2020-07-03
+      app_build_id < '21850000',
```
jklukas, line 28: I took out the call to `udf.fenix_build_to_datetime` both for performance and to avoid needing a refactor of `udf.fenix_build_to_datetime` within the scope of this PR.

jklukas:
I've moved the function to `mozfun` and changed the name to be Fenix-specific.

Basically, I'm decreasing the overall scope here so that we can meet the immediate need of allowing data users to consistently determine channel of Fenix apps. We can defer the larger problem of how to configure relationships between glean app_ids consistently in ETL.

PR #1227, opened 2020-08-12
Use qualified table names and udf functions

wlach:
This makes backfilling inside a sandbox project easier

### DIFF
--- a/sql/telemetry_derived/devtools_accessiblility_panel_usage_v1/query.sql
+++ b/sql/telemetry_derived/devtools_accessiblility_panel_usage_v1/query.sql
@@ -17,79 +17,79 @@ WITH accessibility_panel_client_days AS (
       payload.processes.parent.scalars.devtools_accessibility_node_inspected_count
     ) AS node_inspected_count,
     SUM(
-      udf.get_key(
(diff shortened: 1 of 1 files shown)


sql/telemetry_derived/devtools_accessiblility_panel_usage_v1/query.sql
```diff
@@ -17,79 +17,79 @@ WITH accessibility_panel_client_days AS (
       payload.processes.parent.scalars.devtools_accessibility_node_inspected_count
     ) AS node_inspected_count,
     SUM(
-      udf.get_key(
+      `moz-fx-data-shared-prod`.udf.get_key(
```
jklukas, line 20: If we had test cases for this table, using fully qualified UDF paths could actually break tests. In this case, though, the `get_key` UDF exists in mozfun and we should use that instead.
```suggestion
      mozfun.map.get_key(
```

wlach:
Oops, I guess we're good now! Sorry for the noise

jklukas:
I think the integration test failed due to another branch running CI at the same time. I've seen that in the past and thought it had been fixed. But likely it will succeed if run again.

Issue #1169, opened 2020-07-14
Airflow DAG generation dry run uses local permissions


scholtzan:
As a follow-up to https://github.com/mozilla/bigquery-etl/pull/1168#discussion_r454481648

Currently, the dry run used for determining referenced tables during the Airflow DAG generation uses the local BigQuery permissions of the user generating the DAG. This might lead to situations where users have permissions to some private datasets or projects that are accessible only to them but not to other users. The CI and the DAG generation process should catch these cases.

PR #1228, opened 2020-08-12
Use GCP dry run function for generating Airflow DAGs

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/1169


scholtzan:
This depends on https://github.com/mozilla-services/cloudops-infra/pull/2406 getting deployed

tests/query_scheduling/test_dag_collection.py
```diff
@@ -3,13 +3,18 @@
...
 from bigquery_etl.query_scheduling.dag_collection import DagCollection
 from bigquery_etl.query_scheduling.dag import InvalidDag, DagParseException
 from bigquery_etl.query_scheduling.task import Task
 from bigquery_etl.metadata.parse_metadata import Metadata
+from bigquery_etl.dryrun import DryRun
 
 TEST_DIR = Path(__file__).parent.parent
+TEST_DRY_RUN_URL = (
+    "https://us-central1-bigquery-etl-integration-test.cloudfunctions.net/dryrun"
+)
```
scholtzan, line 17: I created a dry run function in the `bigquery-etl-integration-test` which is used for the tests here. It's probably also useful to have for testing if changes are made to the function itself.

bigquery_etl/dryrun.py
```diff
@@ -111,52 +107,100 @@
...
+    )
+
+    def __init__(self, sqlfile):
+        """Instantiate DryRun class."""
+        self.sqlfile = sqlfile
+        self.dry_run_result = None
+
+    def _execute(self):
+        """Dry run the provided SQL file."""
+        if self.dry_run_result:
```
jklukas, line 125: I can't remember if the base python version here is 3.7 or 3.8. If 3.8, dry_run_result seems like a prime candidate to be expressed as a [cached property](https://docs.python.org/dev/library/functools.html#functools.cached_property), eliminating the need for `_execute`

tests/query_scheduling/test_dag_collection.py
jklukas: How is that function managed? It seems potentially problematic that the function deployed in the test project could be outdated compared to the shared-prod version.

scholtzan: Currently, it's been created manually in `bigquery-etl-integration-test`. But maybe there is a way to deploy the Cloud Function in https://github.com/scholtzan/cloudops-infra/tree/master/projects/data-shared/tf/modules/cloudfunctions/src/bigquery_etl_dryrun to both projects?

jklukas: I think it's sufficient for now to put a comment in the code here about how this is currently deployed manually and could get out of date with what's deployed in shared-prod.

PR #1229, opened 2020-08-12
Fix GCS integration test issues

scholtzan:
Running integration tests in parallel resulted in some of them failing since one test run deleted files that were still accessed by another test run. This should fix the problem.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -76,9 +76,6 @@ def test_bucket():
 
     yield bucket
 
-    # cleanup test bucket
-    bucket.delete_blobs(bucket.list_blobs())
-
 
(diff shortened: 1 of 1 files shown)


conftest.py
```diff
@@ -76,9 +76,6 @@ def test_bucket():
 
     yield bucket
 
-    # cleanup test bucket
-    bucket.delete_blobs(bucket.list_blobs())
```
scholtzan, line 80: We don't need to delete all the files here. Instead, tests that access GCS write all their files to `temporary_gcs_folder` which will be removed at the end of the test (see below)

PR #1230, opened 2020-08-17
Verify that generated DAGs are up-to-date

scholtzan:
This adds a CI step to verify that generated DAGs which will get added to the repo are actually up to date. So, if users forget to re-generate DAGs, this will remind them (or reviewers) to run the `generate_airflow_dag` script.


bigquery_etl/query_scheduling/generate_airflow_dags.py
```diff
@@ -30,12 +30,6 @@
     help="File with DAGs configuration",
     default=DEFAULT_DAGS_FILE,
 )
-parser.add_argument(
-    "--project_id",
-    "--project-id",
-    default="moz-fx-data-shared-prod",
-    help="Dry run queries in this project to determine task dependencies.",
-)
```
scholtzan, line 38: We don't need access to BigQuery anymore since using the dry run Cloud function

bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -12,7 +12,7 @@ default_args = {{
 }}
 
 with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None -%}, schedule_interval={{ schedule_interval | format_timedelta | format_schedule_interval }}{%+ endif -%}) as dag:
-{% for task in tasks %}
+{% for task in tasks | sort(attribute='task_name') %}
```
scholtzan, line 15: Strangely, running the DAG generation in CircleCI puts the Airflow tasks in a completely different order. So, I think it makes sense to explicitly sort them here to avoid these issues.

.circleci/config.yml
```diff
@@ -106,6 +106,20 @@ jobs:
...
+    - *build
+    - run: 
+        name: Generate DAGs
+        command: |
+          mv dags/ existing_dags/
+          mkdir dags
+          PATH="venv/bin:$PATH" script/generate_airflow_dags
+    - run:
+        name: Verify that DAGs were correctly generated and are up-to-date
+        command: diff -aq existing_dags/ dags/
```
scholtzan, line 122: `diff` returns an exit code 1 if there are differences between files

jklukas: Under `verify-requirements`, we generate `requirements.txt` in place and use `git diff --exit-code` to verify that nothing has changed. Unless there's a clear benefit to using `diff -aq` here, I'd like to see us use a consistent methodology for verifying generated content.

PR #1233, opened 2020-08-18
[Bug 1655852] Experiments search aggregates dataset

scholtzan:
Addresses https://bugzilla.mozilla.org/show_bug.cgi?id=1655852

The revenue DS team wants to monitor aggregated search probes for experiments and experiment branches. This gives  them feedback about whether they’ve done something "castastrophic" relatively early on. The dataset create here will be used for adding some monitoring to the already existing experiments enrollments Grafana dashboard.

Everything is quite similar to the `experiment_enrollment_aggregates` dataset, which is already used in Grafana.

Scheduled in telemetry-airflow: https://github.com/mozilla/telemetry-airflow/pull/1128

cc @teonbrooks


sql/telemetry_derived/experiment_search_aggregates_v1/init.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE IF NOT EXISTS
+  `moz-fx-data-shared-prod.analysis.experiment_search_aggregates_v1`(
```
relud, line 2:
```suggestion
  `moz-fx-data-shared-prod.telemetry_derived.experiment_search_aggregates_v1`(
```

sql/telemetry_derived/experiment_search_aggregates_live_v1/view.sql.py
```diff
@@ -0,0 +1,149 @@
...
+              SUM(unnested_ad_clicks.value) AS ad_clicks_count,
+              SUM(unnested_search_with_ads.value) AS search_with_ads_count,
+              SUM(unnested_search_counts.count) AS search_count
+            FROM
+              all_experiments_searches_live
+            GROUP BY
+              experiment,
+              branch,
+              window_start,
+              window_end
```
relud, line 93: rather than repeat this logic, I think it would be better to store this aggregation as a view that is used both here and when generating `experiment_search_aggregates_v1`

scholtzan: There are some small differences between the logic here and in `experiment_search_aggregates_v1` that make this quite tricky:
 
This view queries all entries from the live table that haven't been written to `experiment_search_aggregates_v1` by doing `date(submission_timestamp) > '{submission_date}'`.

For the `experiment_search_aggregates_v1` we'd need to change this to `date(submission_timestamp) = '{submission_date}'` if we'd query from the live table. Or we leave this `WHERE` statement but would need to query from `telemetry_stable.main_v4`.

akkomar (approved):

sql/telemetry_derived/experiment_search_aggregates_live_v1/view.sql.py
relud: moved to #1239

PR #1234, opened 2020-08-18
Fix #1232 - Ignore glam_etl directory when publishing views

### LINKED ISSUE #1232: Error in publish_views: Unclosed identifier literal at [2:3] in moz-fx-data-shared-prod.glam_etl.org_mozilla_fenix_glam_release__view_clients_daily_histogram_aggregates_v1
/CC @acmiyaguchi 

It's not immediately obvious to me why this is missing since the source appears to have it: https://github.com/mozilla/bigquery-etl/blob/master/sql/glam_etl/org_mozilla_fenix_glam_release__view_clients_daily_histogram_aggregates_v1/view.sql, but the query as attempted to publish follows:

```
   1:CREATE OR REPLACE VIEW
   2:  `moz-fx-data-shared-prod.glam_etl.org_mozilla_fenix_glam_release__view_clients_daily_histogram_aggregates_v1
   3:AS
   4:SELECT
   5:  *
   6:FROM
   7:  `moz-fx-data-shared-prod`.glam_etl.org_mozilla_firefox__view_clients_daily_histogram_aggregates_v1
```

I suspect some issue with replacement logic for e.g. `--target-project` is the cause, though the invocation here failing uses `--target-project moz-fx-data-shared-prod`. I would further guess this has to do with a whole identifier being backticked in some cases but in other cases only the project id is.

acmiyaguchi:
This should fix #1232 

The `VIEWS_TO_SKIP` set is the following:

```
('activity_stream/tile_id_types/view.sql', 'firefox_accounts/fxa_amplitude_email_clicks/view.sql', 'pocket/pocket_reach_mau/view.sql', 'telemetry/buildhub2/view.sql', 'glam_etl/org_mozilla_firefox__view_clients_daily_histogram_aggregates_v1/view.sql', 'glam_etl/org_mozilla_fenix_glam_release__view_user_counts_v1/view.sql', 'glam_etl/org_mozilla_fenix_glam_release__view_clients_daily_scalar_aggregates_v1/view.sql', 'glam_etl/org_mozilla_fenix_glam_release__view_probe_counts_v1/view.sql', 'glam_etl/org_mozilla_fenix_glam_release__view_clients_daily_histogram_aggregates_v1/view.sql', 'glam_etl/org_mozilla_firefox__view_clients_daily_scalar_aggregates_v1/view.sql', 'sql/shredder_state/progress/view.sql')
```

### DIFF
--- a/script/publish_views
+++ b/script/publish_views
@@ -4,6 +4,7 @@
 
 from argparse import ArgumentParser
 from functools import partial
+from pathlib import Path
 import logging
 from multiprocessing.pool import ThreadPool
 import os
@@ -14,6 +15,8 @@ from google.api_core.exceptions import BadRequest
 from google.cloud import bigquery
 import sqlparse
 
+sql_root = Path(__file__).resolve().parent.parent / "sql"
+assert sql_root.exists(), f"{sql_root} not found"
 
 VIEWS_TO_SKIP = (
     # Access Denied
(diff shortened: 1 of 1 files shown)


script/publish_views
```diff
@@ -14,6 +15,8 @@ from google.api_core.exceptions import BadRequest
 from google.cloud import bigquery
 import sqlparse
 
+sql_root = Path(__file__).resolve().parent.parent / "sql"
+assert sql_root.exists(), f"{sql_root} not found"
```
jklukas, line 19: Having an assert when importing makes me slightly nervous, but I suppose it's true that this invariant should hold and that it's better to raise this message than have the logic fail more mysteriously.

acmiyaguchi: This was also necessary for testing, since there isn't a great way to prod into the behavior other than running the script itself and inserting print statements. I rather it fail here than mysteriously during automation.

PR #1237, opened 2020-08-18
fenix_app_info documentation fixups

jklukas:
Follow-up to #1226

### DIFF
--- a/mozfun/norm/fenix_app_info/README.md
+++ b/mozfun/norm/fenix_app_info/README.md
@@ -1,0 +1,1 @@
+@sql(../examples/fenix_app_info.sql)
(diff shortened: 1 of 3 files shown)


mozfun/norm/fenix_app_info/README.md
```diff
@@ -0,0 +1 @@
+@sql(../examples/fenix_app_info.sql)
```
acmiyaguchi, line 1: What is this?

jklukas: It injects the text of the example SQL file into the README. See rendered examples in https://mozilla.github.io/bigquery-etl/mozfun/bits28/

The examples are sent through dry run to verify that they work, and we also enforce formatting on them, so that we don't get stale or incorrect examples.

PR #1240, opened 2020-08-18
Update Glam ETL for Fenix

acmiyaguchi:
Followup to #1221 

This adds a new script that runs all of the newly generated ETL. It also fixes the views so they run properly e.g. `UNION ALL` instead of `UNION`. Finally, the channel gets completely ignored for the dataset. At this point, it might actually be worth taking the union of all of the products into a single logical one.


acmiyaguchi:
The `glam_fenix` DAG is disabled in the meanwhile, and I've generated CSV files for unblocking integration work.

PR #1244, opened 2020-08-20
More complete docs for fenix_app_info

### LINKED ISSUE #1243: fenix_app_info UDF docs could motivate sanitization more clearly
After reading https://mozilla.github.io/bigquery-etl/mozfun/norm/, I understand that fenix_app_info maps a couple of telemetry fields to a struct, but I'm not sure what fields the output struct contains or how the values in the output differ from the values in the input. I think it would be helpful to add some context for what values are possible in the output and why those are easier to reason about than the values in the pings.

jklukas:
Fixes #1243

### DIFF
--- a/mozfun/norm/examples/fenix_app_info.sql
+++ b/mozfun/norm/examples/fenix_app_info.sql
@@ -1,9 +1,34 @@
+-- Example of a query over all Fenix builds advertised as "Firefox Beta"
+CREATE TEMP FUNCTION extract_fields(app_id STRING, m ANY TYPE) AS (
+  (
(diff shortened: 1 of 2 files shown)


mozfun/norm/examples/fenix_app_info.sql
```diff
@@ -1,9 +1,34 @@
+-- Example of a query over all Fenix builds advertised as "Firefox Beta"
```
jklukas, line 1: This example should make it clear that there's still a lot of complexity involved in unioning tables together to get complete answers per channel.

Issue #1248, opened 2020-08-21
Fenix releases no longer follow format in fenix_build_to_datetime


jklukas:
cc @chutten 

It looks like around 2020-08-01, Fenix apps are reporting app_build as a 10-digit int string rather than 8, so the parsing of datetime is always giving times in early 2020. An example 10-digit value is '2015758259'. I'm seeing if I can get info from Fenix team about the change.

PR #1250, opened 2020-08-21
Tolerate new 10-digit Fenix app_build format

jklukas:
Fixes https://github.com/mozilla/bigquery-etl/issues/1248

~~Will consider this a draft until we here more from Fenix engineering in https://github.com/mozilla-mobile/fenix/issues/14031 about whether the hypothesis about epoch differences bears out.~~ This is now resolved and we have confirmation from the Fenix team that they're sticking with the modified epoch.

### DIFF
--- a/udf/fenix_build_to_datetime/udf.sql
+++ b/udf/fenix_build_to_datetime/udf.sql
@@ -1,36 +1,56 @@
-/*
-Convert the Fenix client_info.app_build-format string to a DATETIME.
-May return NULL on failure.
-
-The Fenix app_build format is documented here:
-https://github.com/mozilla-mobile/fenix/blob/c72834479eb3e13ee91f82b529e59aa08392a92d/automation/gradle/versionCode.gradle#L13
-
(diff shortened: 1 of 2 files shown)


udf/fenix_build_to_datetime/udf.sql
```diff
@@ -1,36 +1,70 @@
...
+    )
+  ELSE
+    NULL
+  END
 );
 
 -- Tests
 SELECT
-  assert_equals(DATETIME '2020-06-05 14:34:00', udf.fenix_build_to_datetime("21571434"))
+  assert_equals(DATETIME '2020-06-05 14:34:00', udf.fenix_build_to_datetime("21571434")),
```
acmiyaguchi, line 67: It would be useful to put tests for builds on the edges to document epoch.

```
0000000000
0000000001
000000000 # invalid build
000100000
000100001
```

acmiyaguchi (approved):
Looks good to me, suggestions for a few tests.

PR #1253, opened 2020-08-25
Allow baseline pings with 'foreground' reason in KPIs

jklukas:
After merging, I will need to backfill these tables back to early April.

@harterrt formally requested this change in https://jira.mozilla.com/browse/DS-1018


tdsmith:
+dependson #887

PR #1256, opened 2020-08-26
Add dataset for monitoring bigquery-etl query cost

scholtzan:
This adds a scripts which determines the total bytes processed for queries scheduled in bigquery-etl.

Users have previously asked about how much their newly added queries cost. While dry running the queries gives a rough idea, it is sometimes quite inaccurate (ignores sampling or some partitioning). I'm planning to add information about how much the previous run of a query cost to the new bigquery-etl CLI tool to make it easier for users to answer this question.

Additionally, this dataset might also be useful for monitoring cost (like we do with Redash scheduled queries) or for monitoring queries in general.

### DIFF
--- a/sql/monitoring/bigquery_etl_scheduled_queries_cost_v1/query.py
+++ b/sql/monitoring/bigquery_etl_scheduled_queries_cost_v1/query.py
@@ -24,6 +24,7 @@ def create_query(query_paths, date):
 
     return f"""
       SELECT
+        '{date}' AS submission_date,
         creation_time,
         destination_table.dataset_id AS dataset,
         destination_table.table_id AS table,
--- a/sql/monitoring/bigquery_etl_scheduled_queries_cost_v1/query.py
+++ b/sql/monitoring/bigquery_etl_scheduled_queries_cost_v1/query.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+
+"""Determine cost of previously scheduled bigquery-etl queries."""
+
+from argparse import ArgumentParser
+from fnmatch import fnmatchcase
+from google.cloud import bigquery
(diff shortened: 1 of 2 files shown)


sql/monitoring/bigquery_etl_scheduled_queries_cost_v1/query.py
```diff
@@ -0,0 +1,59 @@
...
+
+    tables = [path.parent.name + "*" for path in query_paths]
+
+    return f"""
+      SELECT
+        creation_time,
+        destination_table.dataset_id AS dataset,
+        destination_table.table_id AS table,
+        total_bytes_processed,
+        total_bytes_processed / 1024 / 1024 / 1024 / 1024 * 5 AS cost
```
akkomar, line 32: Nit:
```suggestion
        total_bytes_processed / 1024 / 1024 / 1024 / 1024 * 5 AS cost_usd
```

akkomar (approved):
Very nice!

sql/monitoring/bigquery_etl_scheduled_queries_cost_v1/query.py
```diff
@@ -0,0 +1,60 @@
...
+
+    return f"""
+      SELECT
+        '{date}' AS submission_date,
+        creation_time,
+        destination_table.dataset_id AS dataset,
+        destination_table.table_id AS table,
+        total_bytes_processed,
+        total_bytes_processed / 1024 / 1024 / 1024 / 1024 * 5 AS cost
+      FROM `region-us`.INFORMATION_SCHEMA.JOBS_BY_PROJECT
```
scholtzan, line 33: We'd need to make sure that Airflow has permissions to access this dataset.

PR #1258, opened 2020-08-26
More CLI query commands

scholtzan:
This adds a few more commands to the bqetl CLI:

```
bqetl format    -- to format SQL files
bqetl dryrun    -- to dry run specific SQL files or all SQL files in a directory
bqetl query backfill   -- to run a backfill; has some options to specify a start_date, end_date, dry_run etc
bqetl query info       -- shows information about all or specific queries, including cost and last_updated, if users have permissions to access BigQuery
```

I moved some of the logic from `script` to `bigquery_etl` which makes it easier to reuse and test. It might be good if @jklukas could take a look at these parts.


bigquery_etl/cli/__init__.py
```diff
@@ -1,22 +1,29 @@
...
 
     @click.group(commands=commands)
     @click.version_option(version=__version__)
     def group():
         """CLI tools for working with bigquery-etl."""
         pass
 
+    warnings.filterwarnings(
+        "ignore", "Your application has authenticated using end user credentials"
+    )
```
scholtzan, line 25: When authenticating via `gcloud auth default`, every time `google.cloud.bigquery.Client` makes a request the following warning shows up: 

```
UserWarning: Your application has authenticated using end user credentials from Google Cloud SDK. We recommend that most server applications use service accounts instead. If your application continues to use end user credentials from Cloud SDK, you might receive a "quota exceeded" or "API not enabled" error. For more information about service accounts, see https://cloud.google.com/docs/authentication/
  warnings.warn(_CLOUD_SDK_CREDENTIALS_WARNING)
```

This mutes this warning when using the CLI

bigquery_etl/format_sql/format.py
```diff
@@ -0,0 +1,252 @@
+"""Format SQL."""
```
scholtzan, line 1: This was moved here from `/scripts/format_sql`

bigquery_etl/run_query.py
```diff
@@ -0,0 +1,107 @@
+"""
```
scholtzan, line 1: This was moved here from `scripts/run_query`

jklukas: `scripts/run_query` should be changed to call this.

PR #1259, opened 2020-08-28
Fix entrypoint run_query call

scholtzan:
We should add some tests for the `entrypoint` script. I'll look into it.


scholtzan:
I'll merge now to make sure the DAGs will run successfully, but a retroactive review might still be good.

PR #1260, opened 2020-08-28
Add integration tests for entrypoint run query

scholtzan:
We were running into some issues with making changes and the `entrypoint` script not working properly anymore a couple of times in the past. This adds integration tests to (hopefully) prevent more issues in the future.

### DIFF
--- a/tests/test_entrypoint.py
+++ b/tests/test_entrypoint.py
@@ -1,0 +1,63 @@
+from pathlib import Path
+from google.cloud import bigquery
+import os
+import pytest
+import subprocess
+
+ENTRYPOINT_SCRIPT = Path(__file__).parent.parent / "script" / "entrypoint"
+
+
+class TestEntrypoint:
+    @pytest.mark.integration
+    def test_run_query(self, tmp_path, project_id):
+        query_file_path = tmp_path / "sql" / "test" / "query_v1"
+        os.makedirs(query_file_path)
+
+        query_file = query_file_path / "query.sql"
+        query_file.write_text("SELECT 1 AS a, 'abc' AS b;")
+
+        result = subprocess.check_call(
+            [ENTRYPOINT_SCRIPT, "query", "--project_id=" + project_id, str(query_file)]
+        )
+        assert result == 0
+
+    @pytest.mark.integration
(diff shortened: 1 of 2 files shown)


tests/test_entrypoint.py
```diff
@@ -0,0 +1,90 @@
...
+            assert b"No metadata.yaml found for {}" in result
+        except subprocess.CalledProcessError as e:
+            # running bq in CircleCI will fail since it's not installed
+            # but the error output can be checked for whether bq was called
+            assert b"No such file or directory: 'bq'" in e.output
+            assert b"No metadata.yaml found for {}" in e.output
+            assert (
+                b'subprocess.check_call(["bq"] + query_arguments, stdin=query_stream)'
+                in e.output
+            )
```
scholtzan, line 39: I tried installing the cloud SDK and bq in CircleCI but get the following error:

```
ERROR: gcloud failed to load: While reading from '/root/.config/gcloud/configurations/config_default' [line  5]: option 'project' in section 'core' already exists
    gcloud_main = _import_gcloud_main()
    import googlecloudsdk.gcloud_main
    from googlecloudsdk.api_lib.iamcredentials import util as iamcred_util
    from googlecloudsdk.api_lib.util import apis_internal
    from googlecloudsdk.core import transport
    from googlecloudsdk.core import log
    _log_manager = _LogManager()
    self.Reset(sys.stdout, sys.stderr)
    std_console_formatter = _ConsoleFormatter(stderr)
    use_color = not properties.VALUES.core.disable_color.GetBool(validate=False)
    named_configs.ActivePropertiesFile.Load(),
    ActivePropertiesFile._PROPERTIES = properties_file.PropertiesFile(
    self.__Load(properties_path)
    parsed_config.read(properties_path)
    self._read(fp, filename)
    raise DuplicateOptionError(sectname, optname,

This usually indicates corruption in your gcloud installation or problems with your Python interpreter.
```

So what I ended up doing here is checking if `bq` runs correctly, or if it is not installed, if the error output is as expected, which should also include the `subprocess.check_call` to `bq`.

I checked some of the recent issues that resulted in Airflow failures, these tests would have caught those.

PR #1261, opened 2020-08-31
Prepare search_clients_last_seen for 100% backfill

fbertsch:
- Fully qualify tables
- Remove `NULL` engine
- Remove sample_id limit


sql/search_derived/search_clients_last_seen_v1/init.sql
```diff
@@ -58,6 +58,6 @@ SELECT
   CAST(NULL AS BYTES) AS days_clicked_ads_bytes,
   CAST(NULL AS BYTES) AS days_created_profile_bytes
 FROM
-  search_clients_daily_v8
+  `moz-fx-data-shared-prod`.revenue_derived.search_clients_daily_v8
```
jklukas, line 61: This change is counter to https://github.com/mozilla/bigquery-etl#recommended-practices

> Should not specify a project or dataset in table names to simplify testing

I would prefer that you make this change locally for the backfill rather than change in the repo.

Also, is `revenue_derived` correct here? I would have expected `search_derived`, as that's the default dataset in this directory.

fbertsch: That seems fine, I can do this locally instead. And you're correct, it should be search_derived 🤦

PR #1270, opened 2020-08-31
Tolerate 10-digit Fenix app_build values when choosing channel

jklukas:
See https://github.com/mozilla/bigquery-etl/pull/1250 for proposed logic of how
to parse a date from the 10-digit format, but this change should be reliable
for the specific case of determining whether we are before or after the
2020-07-03 epoch where naming of apps and channels changed.

This will affect GLAM and GUD.

Reverts https://github.com/mozilla/bigquery-etl/pull/1249

I will plan to run a backfill for GUD on the same day I merge this, and then
communicate out the change to data users, since channel breakdowns in GUD
will change for Fenix and Firefox Preview.


jklukas:
Need to work out failing tests.

bigquery_etl/glam/templates/logical_app_id/org_mozilla_fenix_glam_beta__view_clients_daily_histogram_aggregates_v1.sql
```diff
@@ -7,7 +7,7 @@ WITH extracted AS (
   FROM
     `moz-fx-data-shared-prod`.glam_etl.org_mozilla_fenix__view_clients_daily_histogram_aggregates_v1
   WHERE
-    `moz-fx-data-shared-prod`.udf.fenix_build_to_datetime(app_build_id) < date "2020-07-03"
+    mozfun.norm.fenix_app_info('org_mozilla_fenix', app_build_id).channel = 'beta'
```
acmiyaguchi, line 10: This is an autogenerated file, it actually needs to be updated in https://github.com/mozilla/bigquery-etl/blob/f25a096536a10effd5f0f7b09051da9484a76234/bigquery_etl/glam/generate.py#L122-L127. The script is expected to have a callback of a single argument though.

acmiyaguchi: ^ This is inaccurate, since the mapping only affects the final published form of the dataset into csv. I remember creating these views by hand.

acmiyaguchi (approved):
The GLAM portion looks accurate to me, after looking at this with a closer eye.

PR #1272, opened 2020-08-31
UDF refactoring

scholtzan:
This migrates existing UDFs in `udf` and `udf_js` to use the same structure that is used in `mozfun`. This way publishing and parsing logic can be consolidated and simplified. This will also make it easier to implement support in the `bqetl` CLI for UDFs and re-use code there for mozfun UDFs and internal UDFs.

I also formatted all UDFs and added tests for the UDF parsing.


tests/data/udf/shift_28_bits_one_day/udf.sql
```diff
@@ -3,13 +3,11 @@
 Shift input bits one day left and drop any bits beyond 28 days.
```
jklukas, line 3: Why are some of these UDFs moved under `tests/data/udf` rather than `udf/`?

udf_js/crc32/metadata.yaml
```diff
@@ -0,0 +1,6 @@
+description: Calculate the CRC-32 hash of an input string.  The implementation here
+  could be optimized. In particular, it calculates a lookup table on every invocation
+  which could be cached and reused. In practice, though, this implementation appears
+  to be fast enough that further optimization is not yet warranted.  Based on https://stackoverflow.com/a/18639999/1260237
+  See https://en.wikipedia.org/wiki/Cyclic_redundancy_check
+friendly_name: Crc32
```
jklukas, line 6:
```suggestion
friendly_name: CRC-32
```

tests/data/udf/shift_28_bits_one_day/udf.sql
jklukas: I see now that these are duplicated in both spots, and that the `tests/data` versions are used for tests. I'm a bit worried about this duplication. I could imagine there being some confusion for someone browsing the repo looking for an implementation of one of these duplicated functions. Or if we need to make a modification to one of them, is it expected that we do so in both places?

Could we prefix all of these names with `example_` or something along those lines to make the separation more clear?

scholtzan: I prefixed them with `test_` which is in line with some of the other test data that we already have.

Issue #1271, opened 2020-08-31
Integration tests failing due to persistent table


relud:
integration tests are consistently failing with `409 Already Exists: Table *****************************:tmp.incremental_query_v1_20200315_temp` (e.g. https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/5471/workflows/89bd18b0-b4c5-4021-9368-c9647b4aec88/jobs/23358)

if the table is manually deleted then the next test will pass, but then it will be back in a failure-state.

We need to fix two aspects of this:

1. this resource needs to be cleaned up when the test ends
2. this resource needs to be named something unique for every run, so that concurrent CI tests don't conflict with each other.

PR #1273, opened 2020-08-31
Fix writing to temporary table when publishing JSON

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/1271

### DIFF
--- a/bigquery_etl/public_data/publish_json.py
+++ b/bigquery_etl/public_data/publish_json.py
@@ -91,29 +91,30 @@ class JsonPublisher:
         """Publish query results as JSON to GCP Storage bucket."""
         self.last_updated = datetime.datetime.utcnow()
 
-        if self.metadata.is_incremental_export():
-            if self.date is None:
-                logging.error(
-                    "Cannot publish JSON. submission_date missing in parameter."
-                )
-                sys.exit(1)
+        try:
+            if self.metadata.is_incremental_export():
+                if self.date is None:
+                    logging.error(
+                        "Cannot publish JSON. submission_date missing in parameter."
+                    )
+                    sys.exit(1)
 
-            # if it is an incremental query, then the query result needs to be
-            # written to a temporary table to get exported as JSON
-            self._write_results_to_temp_table()
-            self._publish_table_as_json(self.temp_table)
-        else:
-            # for non-incremental queries, the entire destination table is exported
-            result_table = f"{self.dataset}.{self.table}_{self.version}"
-            self._publish_table_as_json(result_table)
+                # if it is an incremental query, then the query result needs to be
+                # written to a temporary table to get exported as JSON
+                self._write_results_to_temp_table()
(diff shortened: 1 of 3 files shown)


bigquery_etl/public_data/publish_json.py
```diff
@@ -228,7 +230,10 @@ def _gcp_convert_ndjson_to_json(self, gcs_path):
...
     def _write_results_to_temp_table(self):
         """Write the query results to a temporary table and return the table name."""
-        table_date = self.date.replace("-", "")
+        table_date = self.date.replace("-", "") + "".join(
+            random.choice(string.ascii_lowercase) for i in range(12)
+        )
+
         self.temp_table = (
             f"{self.project_id}.tmp.{self.table}_{self.version}_{table_date}_temp"
         )
```
scholtzan, line 239: The temporary tables get deleted in https://github.com/mozilla/bigquery-etl/blob/1b6128ff5aee11c12ee440cda3567083b05d7aee/bigquery_etl/public_data/publish_json.py#L79

I think the problem here was that there were multiple integration test instances running in parallel, trying to create the same temporary table and failing when it already existed due to another test having created it.

tests/public_data/test_publish_public_data_json_script.py
```diff
@@ -46,12 +46,6 @@ def test_script_incremental_query(
 
         assert res.returncode == 0
 
-        with pytest.raises(NotFound):
-            temp_table = (
-                f"{project_id}.{temporary_dataset}.incremental_query_v1_20200315_temp"
-            )
-            bigquery_client.get_table(temp_table)
-
```
scholtzan, line 54: This was checking if the temporary table gets removed. Since the table name now has some random string in it, it's the actual table name is unknown and cannot be checked here :|

bigquery_etl/public_data/publish_json.py
relud: based on https://stackoverflow.com/a/1481512, that method is not guaranteed to be called at all, and `bq ls bigquery-etl-integration-test:tmp` shows that tables created in CI are not being cleaned up.

I think this needs to delete the temp table in a `finally` block somewhere, to ensure that it actually runs.

bigquery_etl/public_data/publish_json.py
```diff
@@ -114,6 +109,12 @@ def publish_json(self):
 
         self._publish_last_updated()
 
+        # delete temporary artifacts
+        if self.temp_table:
+            self.client.delete_table(self.temp_table)
+
+        self._clear_stage_directory()
+
```
scholtzan, line 117: I moved the deletion of the temporary tables and files to the end of the script here.

relud: this needs to occur in a `try ... finally` block to ensure that resources are deleted even when an error occurs.

scholtzan: Right, should be fixed now.

PR #1280, opened 2020-09-01
Add `clients_last_seen.days_any_interaction_bits`.

felixlawrence:
This column is a 28-day bit array containing whether the client has any active ticks (thus nonzero active hours) associated with each of the last 28 submission dates.

It's possible that three files in `tests/telemetry_derived/clients_last_seen_v1/test_single_day/` may need to be updated too; I didn't take the time to understand the test suite and was hoping that someone else could take a look at that!

The name `days_any_interaction_bits` came out of a discussion with @irrationalagent and @godelstheory. We think it's probably more intuitive than `days_nonzero_active_hours` or similar, even though its name gives less hint as to the quantities it's derived from?

As per our meeting last week, we're expecting not to backfill this quantity unless it proves to be super useful and there's a need for historical data.


jklukas:
> days_any_interaction_bits

I've been trying to keep to the form `days_<past tense verb>_<maybe an object here>_bits`. Would it be possible to recast this in that format? days_interacted_bits? days_seen_any_interaction_bits?

I will admit that `any_interaction` does feel pretty clear on its own, so I could be pursuaded to keep it as-is.

jklukas (approved):
I pushed a commit to update tests. This looks good to me otherwise. I'll merge once I have clean CI.

jklukas:
This field now exists and the view is updated. I told Airflow to rerun the query for today, so we should have the first non-null bits momentarily.

felixlawrence:
Awesome, thank you! Looking forward to using it once the data has trickled in!

PR #1281, opened 2020-09-02
Upgrade to pytest 6.0.1

relud:
Also upgrade and fix pytest plugins


bigquery_etl/public_data/__init__.py
```diff
@@ -0,0 +1 @@
+"""Public Data."""
```
relud, line 1: mypy gets angry if we don't have `__init__.py` files

bigquery_etl/util/bigquery_tables.py
```diff
@@ -36,7 +39,7 @@ def get_tables_matching_patterns(client: bigquery.Client, patterns):
                 datasets = [d for d in all_datasets[project] if fnmatchcase(d, dataset)]
             for dataset in datasets:
                 dataset = f"{project}.{dataset}"
-                tables = [(f"{dataset}.{table}", None)]
+                tables = [f"{dataset}.{table}"]
```
relud, line 42: r? @scholtzan 

I checked the usage of this function because `mypy` was complaining about inconsistent typing on `tables`, and it looks like this isn't supposed to be a `Tuple`?

bigquery_etl/util/bigquery_tables.py
```diff
@@ -45,6 +48,6 @@ def get_tables_matching_patterns(client: bigquery.Client, patterns):
                         for t in all_tables[dataset]
                         if fnmatchcase(t.table_id, table)
                     ]
-                    matching_tables += tables
+                matching_tables += tables
```
relud, line 51: r? @scholtzan 

It looks like this was skipping non-wildcard tables due to incorrect indent

pytest.ini
```diff
@@ -10,7 +10,7 @@ markers =
     integration
 norecursedirs =
     venv
-    script/legacy/athena-migration
+    script/legacy
```
relud, line 13: we don't want to update legacy scripts as linting rules change

script/copy_deduplicate
```diff
@@ -1,8 +1,7 @@
 #!/usr/bin/env python3
 
 """
-Copy a day's data from live ping tables to stable ping tables,
-deduplicating on document_id.
+Copy a day's data from live to stable ping tables, deduplicating on document_id.
```
relud, line 4: I had to reword a few docstrings now that pydocstyle is working again, because the sentence must be a single line.

script/copy_deduplicate
```diff
@@ -310,15 +309,15 @@ def list_live_tables(client, pool, project_id, only_tables, table_filter):
     ]
 
 
-def main():
+def _main():
```
relud, line 312: I didn't want to write docstrings for all the functions in scripts, now that pydocstyle working again, so instead I renamed the functions to be private.

This isn't great for `main` functions, but I figured it's fine for now because scripts can't be imported.

tests/format_sql/conftest.py
```diff
@@ -16,16 +16,15 @@ def pytest_collect_file(parent, path):
     """Collect format_sql tests."""
     if path.basename == "expect.sql":
         print(path)
-        return FormatTest(path, parent)
+        return FormatTest.from_parent(parent, fspath=path)
```
relud, line 19: Directly instantiating test classes was deprecated in favor of `.from_parent` in pytest 5.4, and removed in pytest 6

script/copy_deduplicate
scholtzan: Yes, that should be fine. Eventually, the logic from the `script/` scripts will be moved to `bigquery_etl`, so those can be renamed back to `main` when that happens.

PR #1288, opened 2020-09-08
Bug 1637926 Add an aet_clients_daily table for end-user analysis

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1637926

~~The first draft here defines two separate tables:~~
<details>
- `moz-fx-data-shared-prod.firefox_accounts_derived.aet_desktop_clients_daily_v0`
- `moz-fx-data-shared-prod.firefox_accounts_derived.aet_log_users_daily_v0`

These tables have been created with data from 2020-09-01 through 2020-09-07
for the sake of discussion. The `_v0` is meant to convey these are very much
WIP and are not expected to be productionized as-is.

These daily tables are going to rely on an early stage of ETL where we track
associations between ecosystem_user_ids, and associations of ecosystem_client_id
to ecosystem_user_id. We do not yet make any attempt here to join in canonical
IDs.

These tables also currently show exact aggregations. We will need to institute
bucketing of some sort, perhaps by having a first aggregation over all clients
to get quantiles for each metric, and then replacing the per-client aggregations
with the associated quantile. This could probably use DS help as we refine
the methodology.

This first draft is primarily intended to express a sense of what fields we
would present in an `ecosystem_clients_daily` view.</details>

This PR now represents a full implementation ready for review.


irrationalagent:
If we replace individual client measurements with their corresponding quantiles as defined on the distribution over all clients per day, then it could make it hard to interpret the per-client time series. For example, a client's actual, absolute usage might go up slightly in absolute terms over a period of days, but if usage of the population goes up over the same period at a slightly higher rate, then it can look like that client's usage actually went down. When the question concern's the client's usage relative to everyone else then that's probably OK, but if the question concerns within-client changes isolated from population-level trends then we could have problems.

If we need to bucket, then I think we should have some kind of scheme that maintains a fixed scale.

jklukas:
> If we need to bucket, then I think we should have some kind of scheme that maintains a fixed scale.

Understood. I think it would be reasonable to include some instructions in code comments here about how to run a query to produce quantiles for a new metric which we can then copy-and-paste into the query as the static binning for that new metric.

It's not clear to me how to decide on the number of bins and whether that should vary per metric.

jklukas:
This is now a complete implementation of `aet_clients_daily`. The tables as implemented here have been backfilled. Sample query showing full activity for a specific user:

```
SELECT
  *
FROM
  `moz-fx-data-shared-prod.account_ecosystem.aet_clients_daily`
WHERE
  submission_date >= '2020-09-01'
  AND user_id = '8d3de4dd9cf927f9e7c2e4cec7387d6ca264bc28b5a9c7c72f35b62842bea852'
ORDER BY
  submission_date DESC
LIMIT
  1000
```

sql/account_ecosystem/aet_clients_daily/metadata.yaml
```diff
@@ -0,0 +1,13 @@
+friendly_name: AET Clients Daily
+description: >
+  One row per user per service per day, showing metrics across services.
+
+  The `user_id` and `client_id` are directly related to the `ecosystem_user_id`
```
jklukas, line 5: Note the use of simple `user_id` and `client_id` as the field names for the final user-facing analysis table. I don't want to call them `ecosystem_user_id` and `ecosystem_client_id` because they are not directly those values. `canonical_id` seems jargony and the details of how we maintain continuity across password resets are not necessary for end users to think about.

I am very open to feedback on this naming decision, and also on the naming for `service` which ends up being "fxa - <some_oauth_client_id>" or "desktop". Calling desktop a "service" doesn't seem right. Maybe "source", but that seems _too_ generic.

sql/account_ecosystem/aet_clients_daily/view.sql
```diff
@@ -0,0 +1,57 @@
...
+    normalized_country_code AS country,
+  FROM
+    `moz-fx-data-shared-prod.account_ecosystem_derived.desktop_clients_daily_v1`
+),
+fxa_logging AS (
+  SELECT
+    submission_date,
+    canonical_id,
+    CAST(NULL AS string) AS ecosystem_client_id_hash,
+    -- We likely want to replace oauth_client_id with a human-readable service name.
```
jklukas, line 25: What's the source of truth for logical names of oauth_client_id?

sql/account_ecosystem_derived/desktop_clients_daily_v1/query.sql
```diff
@@ -0,0 +1,87 @@
+-- Function to coerce a raw value `v` to the nearest quantile to make fingerprinting
```
jklukas, line 1: This method could definitely use some DS scrutiny, but it's something we can evolve later and reprocess, so not critical to get perfect right now.

scholtzan (approved):

jklukas:
Merged, but additional commentary in here is still welcome and I can address in additional PRs.

sql/account_ecosystem_derived/desktop_clients_daily_v1/query.sql
irrationalagent: So if I'm understanding this right, we would define the quantile array for each measurement "by hand" (though this in principle could be automated), and then raw values will be replaced by the lower of the two quantile boundaries that the raw value falls between.

This maintains a fixed/absolute scale, which is good. I'm trying to think about what would happen if, for example, the observed distribution for a probe compresses on the low end so that most values start falling between lowest pairs, causing most values to be identical. We can then re-define the quantile array for that measurement to have more resolution in the low end, and since these numbers will still be on the same underlying scale this won't cause problems.

Also, I have to think about this more, but its good that we can pass in an arbitrary array because depending on the underlying distribution for the probe it may be that we do not always want evenly spaced quantiles. Consider a distribution where most of the probability mass is squished into the middle, but still has long, thin tails. If we use deciles in this case, for example, most raw values will be drawn from the middle of the distribution and thus will have similar (or the same) binned values. This will reduce the variance of the binned distribution, potentially to a point where we lose a lot of information. I believe there are other schemes in which the bins are spaced such that they are closer together nearer the center of probability mass. I believe there may be even a general-purpose algorithm that optimizes bin spacing like this given any distribution - I will research.

PR #1289, opened 2020-09-08
Android event types init

fbertsch:


udf/event_code_points_to_string/udf.sql
```diff
@@ -0,0 +1,43 @@
...
+          WHEN
+            i IS NULL
+          THEN
+            NULL
+            -- Skip commas
+          WHEN
+            i >= 43
+          THEN
+            i + 2
+            -- Skip double quote
```
fbertsch, line 18: We use double quotes to represent "any" for an event property. It is one of the lowest-valued representable (not e.g. `\u0001`) unicode values that is not a meta-character in regex.

udf/event_code_points_to_string/udf.sql
```diff
@@ -0,0 +1,43 @@
...
+      code_points IS NULL,
+      NULL,
+      ARRAY(
+        SELECT
+          CASE
+          WHEN
+            i IS NULL
+          THEN
+            NULL
+            -- Skip commas
```
fbertsch, line 13: Commas are present between every event, so we cannot use them as index chars.

udf/event_code_points_to_string/udf.sql
```diff
@@ -0,0 +1,43 @@
...
+            i + 1
+          ELSE
+            i
+          END
+        FROM
+          UNNEST(code_points) AS i
+      )
+    )
+  )
+);
```
scholtzan, line 31: I'm wondering if there is a good way to handle invalid code points, because certain code points like `udf.event_code_points_to_string([55296])` or ones that are too large, like `udf.event_code_points_to_string([2000000])` will throw an error. There is also `SAFE.CODE_POINTS_TO_STRING` which will return `NULL` instead of an error but it feels like these code points would require special handling.

sql/org_mozilla_firefox_derived/event_types_v1/init.sql
```diff
@@ -0,0 +1,132 @@
...
+    event_property,
+    event_property_index
+),
+per_event AS (
+  SELECT
+    category,
+    event,
+    first_timestamp,
+    primary_index AS numeric_index,
+    udf.event_code_points_to_string([primary_index]) AS index,
```
scholtzan, line 105: A question that came to my mind about this model in general: What if there are so many different events that not all of them can be represented as valid code points because they exceed the maximum valid value? I guess it might be possible to split up the code point and have multiple represent one event?

scholtzan (commented):
Looking good.

I left some questions/comments about handling invalid code points. Currently, running the query fails with `Invalid codepoint 55296`.

Also, most build errors should be easily fixed by running `black`.

sql/org_mozilla_firefox_derived/event_types_v1/init.sql
fbertsch: Right, there are limitations. One of them is that we are limited to ~1M event types, and ~1M values per event property. We have never really gotten close to that point, and in general I would direct people at that point to consolidate events and use event properties to differentiate between them (if they need to).

udf/event_code_points_to_string/udf.sql
fbertsch: The 55296 code point failure is surprising to me. In general I wanted this to hard fail on too-large (e.g. 2M) because then the entire job will fail and we will have to go in and manually fix things. That might mean dropping some event property or whatnot - eventually we will automate this to stop collecting new values once we are too large.

udf/event_code_points_to_string/udf.sql
scholtzan: Just wondering, do we actually have to have commas between events if they are self-contained as separate code points/characters anyway? Not having commas could simplify things, like writing regexes.

fbertsch: Indeed, we do, because different events have different numbers of event properties.

udf/event_code_points_to_string/udf.sql
```diff
@@ -0,0 +1,54 @@
...
+    CODE_POINTS_TO_STRING(NULL),
+    udf.event_code_points_to_string(CAST(NULL AS ARRAY<INT64>))
+  ),
+  assert_equals(CODE_POINTS_TO_STRING([NULL]), udf.event_code_points_to_string([NULL])),
+  assert_equals(CODE_POINTS_TO_STRING([]), udf.event_code_points_to_string([]));
+
+SELECT
+  assert_not_null(udf.event_code_points_to_string([n]))
+FROM
+  UNNEST(GENERATE_ARRAY(1, 1000000)) AS n
```
fbertsch, line 54: This test confirms the allowable 1M values for events and event properties.

udf/event_code_points_to_string/udf.sql
```diff
@@ -0,0 +1,54 @@
...
+  assert_equals("#", udf.event_code_points_to_string([34])),
+  assert_equals("-", udf.event_code_points_to_string([43])),
+  assert_equals(CODE_POINTS_TO_STRING([1, 2]), udf.event_code_points_to_string([1, 2])),
+  assert_equals(
+    CODE_POINTS_TO_STRING(NULL),
+    udf.event_code_points_to_string(CAST(NULL AS ARRAY<INT64>))
+  ),
+  assert_equals(CODE_POINTS_TO_STRING([NULL]), udf.event_code_points_to_string([NULL])),
+  assert_equals(CODE_POINTS_TO_STRING([]), udf.event_code_points_to_string([]));
+
```
scholtzan, line 50: Optionally, a test for the invalid code points >= 55294 could be added

scholtzan (approved):
Looks good to me

udf/event_code_points_to_string/udf.sql
fbertsch: I like the current one to guarantee 1M allowable values - if it ends up being slow we can remove it down the line.

PR #1290, opened 2020-09-08
CLI commands for UDFs

scholtzan:
Adds some CLI commands for working with mozfun and internal UDFs:

```
Commands:
  create    Create a new UDF
  info      Get UDF information
  publish   Publish UDFs to BigQuery.
  validate  Validate a UDF.
```

### DIFF
--- a/bigquery_etl/cli/udf.py
+++ b/bigquery_etl/cli/udf.py
@@ -2,6 +2,7 @@
 
 import click
 from pathlib import Path
+import pytest
 import re
 import string
 import sys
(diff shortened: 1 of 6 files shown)


bigquery_etl/cli/udf.py
```diff
@@ -0,0 +1,238 @@
...
+    ctx.ensure_object(dict)
+    ctx.obj["UDF_DIRS"] = ("udf", "udf_js")
+
+
+@click.group()
+@click.pass_context
+def mozfun(ctx):
+    """Create the CLI group for the mozfun command."""
+    ctx.ensure_object(dict)
+    ctx.obj["UDF_DIRS"] = ("mozfun",)
```
scholtzan, line 37: CLI commands are essentially the same for mozfun and internal UDFs, except for the default directories. Instead of duplicating every command, the default working directory is passed to the context based on whether the `mozfun` or `udf` command is invoked and then used in the command implementations.

acmiyaguchi (approved):
Looks good to me. :+1:

PR #1295, opened 2020-09-09
Query for updates to event_types_v1

fbertsch:
This query takes yesterday's events, yesterday's event_types,
and adds the new events, event_properties, and property values.

It writes it out to a new partition. This is not strictly
necessary but will aid debugging and redoes.

This depends on merging https://github.com/mozilla/bigquery-etl/pull/1289 and rebasing.

### DIFF
--- a/sql/org_mozilla_firefox_derived/event_types_v1/query.sql
+++ b/sql/org_mozilla_firefox_derived/event_types_v1/query.sql
@@ -1,0 +1,199 @@
+WITH
+  current_events AS (
+  SELECT
+    name AS event,
+    * EXCEPT (name)
+  FROM
+    org_mozilla_firefox.events
+  CROSS JOIN
+    UNNEST(events) AS event
+  WHERE
+    DATE(submission_timestamp) = @submission_date
+  ),
+  event_types AS (
+    SELECT *
+    FROM org_mozilla_firefox_derived.event_types_v1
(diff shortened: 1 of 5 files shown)


sql/org_mozilla_firefox_derived/event_types_v1/query.sql
```diff
@@ -0,0 +1,243 @@
+WITH current_events AS (
```
fbertsch, line 1: I understand this query is a _pain_ to look at. A bit of explanation here may be in order.

We're basically doing exactly what's in the `init` query (see the `per_event_property` and `per_event` CTEs). Ahead of that, we're stripping out known events, event_properties, and event_property_values; we then create the new values for each and `UNION` them to the existing ones (see e.g.`all_primary_event_types`). Once we have the three different sets of data (_all_ events, event_properties, and event_property_values) we do do exactly what we do in the init function.

tests/org_mozilla_firefox_derived/event_types_v1/test_new_day/org_mozilla_firefox_derived.event_types_v1.yaml
```diff
@@ -0,0 +1,57 @@
+---
```
fbertsch, line 1: This file is the same as the result from the `init` test.

PR #1305, opened 2020-09-11
Add DETERMINISTIC modifier to JS functions

jklukas:
This recently released feature allows query results using JS UDFs to be cached.

See https://issuetracker.google.com/issues/138310623


scholtzan (approved):
Neat!

PR #1307, opened 2020-09-11
bqetl CLI docs

scholtzan:
Some basic documentation for the bqetl CLI


README.md
```diff
@@ -5,6 +5,50 @@ BigQuery ETL
...
 Bigquery UDFs and SQL queries for building derived datasets.
 
+The `bqetl` CLI
+---
+
+The `bqetl` command-line tool aims to simplify working with the bigquery-etl repository by supporting
+common workflows, such as creating, validating and scheduling queries or adding new UDFs.
+
+`bqetl` can be installed by running `pip install mozilla-bigquery-etl` or by downloading the repository
+and running `pip install .`
```
jklukas, line 15:
```suggestion
`bqetl` can be installed by running `pip install mozilla-bigquery-etl` or by cloning the repository
and running `pip install --editable .`
```

The `-e` or `--editable` flag ensures that the installed executable continues to reference the source code rather than making a static copy at the time of install.

README.md
```diff
@@ -5,6 +5,50 @@ BigQuery ETL
...
+
+Options:
+  -p, --path DIRECTORY  Path to directory in which query should be created
+  -o, --owner TEXT      Owner of the query (email address)
+  -i, --init            Create an init.sql file to initialize the table
+  --help                Show this message and exit.
+```
+
+Running some commands, for example to create or query tables, require access to GCP which can
+be granted by running `gcloud auth login`.
```
jklukas, line 50: This should probably just say that it requires GCP access similar to the `bq` CLI and link to https://docs.telemetry.mozilla.org/cookbooks/bigquery/access.html#using-the-bq-command-line-tool  rather than mentioning any specific auth steps.

jklukas (approved):
r+wc

The "Scheduling Queries in Airflow" section should be updated to reference the relevant `bqetl` invocation to schedule a query and generate its DAG, which can be in this PR or left to a follow-up.

PR #1309, opened 2020-09-11
Draft query showing a proof-of-concept for clients_all_time

jklukas:
## clients_all_time

The concept of `clients_all_time` is to provide an interface very similar to `clients_last_seen`, but the underlying data model is a single row per client over all time. Instead of having a moving 28-day window, we rewrite the entire table every day, incorporating the new day's observations. At query time, the view logic unpacks this one-row-per-client representation into a per-client-per-day representation much like clients_last_seen.

Another way of looking at this is that it gives us clients_first_seen + clients_last_seen without needing a join. It also unlocks the possibility of doing LTV-type calculation that rely on history from 1 year ago.

### Prototype

The query given here considers just a single dimension, a single usage criterion, and is for a 1% sample of the desktop population, but it runs over the entire history contained in `main_v4` to demonstrate the concept.

It ran in approx. 45 minutes and scanned 100 GB. The resulting table is about 1 GB in size.

We can infer from this that the storage size on disk will be about 0.5 GB per BYTES column per sample_id.

### Trying it out

We provide a clients_last_seen-like view on top of it that can be queried like:

```
SELECT
  * 
FROM 
  `moz-fx-data-shared-prod.analysis.klukas_clients_all_time` 
LIMIT 1000
```

See a preview of results at https://sql.telemetry.mozilla.org/queries/74697/source

### DIFF
--- a/clients_all_time.sql
+++ b/clients_all_time.sql
@@ -1,0 +1,54 @@
+/*
+
+The underlying "clients_all_time_v0" table has one row per client for all time,
+but this view on top of it explodes the dates so that we get one row per client
+per day, similar to the clients_last_seen structure we're already familiar with.
+
+The differences between this view and clients_last_seen are:
+
+- The bit patterns are BYTES rather than INT64
+- The bit patterns include all history, so we could use them for LTV-type calculations
+  that depend on activity 1 year ago
+- This view includes first_seen_date, which previously required a separate table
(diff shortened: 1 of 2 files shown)


jklukas:
The actual performance of a view on top of this all_time table was very bad, so we are not pursuing this idea further in the near term.

PR #1314, opened 2020-09-15
Bug 1661250 - prepare Shredder script for adding Pioneer support

akkomar:
This adds a new `environment` parameter to Shredder delete script, accepting either `telemetry` or `pioneer` values.
The goal of this parameter is to have a way for choosing different source and target tables. By default it is set to `telemetry` (which keeps current behavior unchanged).

Next step is to define logic for discovering Pioneer study tables in `config.py` and invoke it in delete script (`NotImplementedError`).


akkomar:
Hey @relud this is how I'd like to support Pioneer in Shredder. I still need to figure out table configuration but I'd like to have your input on this approach.

relud (approved):
this approach looks good to me, very clean.

Issue #1164, opened 2020-07-13
mozfun: publish descriptions to BQ


jklukas:
Requested by @mcfunley

PR #1317, opened 2020-09-15
Publish descriptions for UDFs

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/1164

This turned out hairier than I would have hoped. Unlike tables, there is no API for adding descriptions to existing UDFs, so descriptions have to be added to the queries defining UDFs.

### DIFF
--- a/bigquery_etl/udf/publish_udfs.py
+++ b/bigquery_etl/udf/publish_udfs.py
@@ -1,6 +1,7 @@
 """Publish UDFs and resources to the public mozfun GCP project."""
 
 from argparse import ArgumentParser
+import json
 import os
 import re
 
@@ -17,8 +18,9 @@ DEFAULT_GCS_BUCKET = "moz-fx-data-prod-bigquery-etl"
 DEFAULT_GCS_PATH = ""
 
 OPTIONS_LIB_RE = re.compile(r'library = "gs://[^"]+/([^"]+)"')
+OPTIONS_RE = re.compile(r"OPTIONS(\n|\s)*\(")
 
-SKIP = {"udf/main_summary_scalars/udf.sql"}
+SKIP = ["udf/main_summary_scalars/udf.sql"]
 
 parser = ArgumentParser(description=__doc__)
 parser.add_argument(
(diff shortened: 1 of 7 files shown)


bigquery_etl/udf/publish_udfs.py
```diff
@@ -134,6 +136,14 @@ def publish_udf(
             fr'library = "gs://{gcs_bucket}/{gcs_path}\1"', definition
         )
 
+        # add UDF descriptions
+        if raw_udf.filepath not in SKIP:
+            escaped_description = json.dumps(str(raw_udf.description))
```
scholtzan, line 141: The descriptions need to be escaped since otherwise quotation marks and other characters will make the query invalid. Converting it to JSON seemed like the easiest thing to do here

bigquery_etl/udf/publish_udfs.py
```diff
@@ -134,6 +136,14 @@ def publish_udf(
...
+            query = OPTIONS_RE.sub(f"OPTIONS(description={escaped_description},", query)
```
scholtzan, line 142: JavaScript UDFs already have some `OPTIONS` for importing external scripts, descriptions need to be added to the existing `OPTIONS`

bigquery_etl/udf/publish_udfs.py
jklukas: BQ has triple-quoted strings. Can we wrap the description in `"""` instead of json-ifying? It seems like that would be slightly safer.

jklukas (approved):
r+wc

bigquery_etl/udf/publish_udfs.py
scholtzan: Json-ifying also handles escaping of some other characters, for example the `\x01` in `[...]SELECT MAX(n)     FROM UNNEST(GENERATE_ARRAY(0, 8 * BYTE_LENGTH(b)))
  AS n     WHERE BIT_COUNT(SUBSTR(b >> n, -1) & b''\x01'') > 0));  See also: bits_to_days_since_seen.sql[...]`. Just wrapping the description in `"""` will result in a `re.error: bad escape \x at position` error for these cases when used in `OPTIONS_RE.sub`.

jklukas: Sounds worth adding a comment to this line. 👍

PR #1327, opened 2020-09-18
Add tables for monitoring fission experiment

akkomar:


sql/telemetry_derived/fission_monitoring_crash_v1/query.sql
```diff
@@ -0,0 +1,101 @@
...
+    normalized_channel = 'nightly'
+    AND mozfun.map.get_key(environment.settings.user_prefs, 'fission.autostart') IS NOT NULL
+),
+main_ping_data AS (
+  SELECT
+    submission_timestamp,
+    mozfun.map.get_key(environment.settings.user_prefs, 'fission.autostart') AS experiment_branch,
+    environment.build.build_id,
+    0 AS main_crash,
+    0 AS content_crash,
```
godelstheory, line 62: Please add `oom_crashes` and `shutdown_crashes`.

akkomar: @godelstheory how do I get them? i don't see them in the crash ping.

godelstheory: For `oom_crashes` I have used:
```
 IF (payload.metadata.oom_allocation_size is not null,
      1,
      0) AS oom_crash**
```

For `shutdown_crashes`, maybe we should add three:
1. `shutdown_kill_crash`
```
payload.metadata.ipc_channel_error = 'ShutDownKill'
```
2. `shutdown_hang`
```
payload.metadata.moz_crash_reason LIKE 'MOZ_CRASH%'
```
3. `content_shutdown_crash`
```
IF (REGEXP_CONTAINS(payload.process_type, 'content')
   AND REGEXP_CONTAINS(payload.metadata.ipc_channel_error, 'ShutDownKill'),
       1,
       0) AS content_shutdown_crash,
```

akkomar: Adding oom_crashes and shutdown_hangs. As for the `shutdown_kill_crash` - defined this way (`payload.metadata.ipc_channel_error = 'ShutDownKill'` it will be a superset of `content_shutdown_crash` - is that expected?

godelstheory: Yeah, I would include both.

sql/telemetry_derived/fission_monitoring_main_v1/query.sql
```diff
@@ -0,0 +1,11 @@
+-- creates a pre-filtered main ping dataset for monitoring
+-- and ad-hoc analyses on fission experiment
+SELECT
+  *
```
scholtzan, line 4: Based on the `init.sql`, shouldn't this be something like this?
```suggestion
  DATE(submission_timestamp) AS submission_date,
  document_id,
  sample_id
```

sql/telemetry_derived/fission_monitoring_crash_v1/query.sql
```diff
@@ -0,0 +1,137 @@
...
+  SUM(content_crash) AS content_crashes,
+  SUM(startup_crash) AS startup_crashes,
+  SUM(content_shutdown_crash) AS content_shutdown_crashes,
+  SUM(oom_crashes) AS oom_crashes,
+  SUM(shutdown_hangs) AS shutdown_hangs,
+  SUM(gpu_crashes) AS gpu_crashes,
+  SUM(plugin_crashes) AS plugin_crashes,
+  SUM(gmplugin_crashes) AS gmplugin_crashes,
+  SUM(
+    LEAST(GREATEST(usage_seconds / 3600, 0), 24)
```
scholtzan, line 127: There are days with up to 25 hours ;) But I guess we can ignore those here

sql/telemetry_derived/fission_monitoring_main_v1/query.sql
akkomar: We want to have all the probes here, but also I don't want to specify all of them in the init.sql - therefore the `*` and `ALLOW_FIELD_ADDITION` [here](https://github.com/mozilla/telemetry-airflow/pull/1154/files#diff-da835b6a2ddbe3615ee058a2b838ccffR41)

sql/telemetry_derived/fission_monitoring_crash_v1/query.sql
akkomar: Good point! I think this is a common pattern used to remove outliers, but definitely worth keeping in mind (and maybe investigating implications?) /cc @godelstheory

sql/telemetry_derived/fission_monitoring_main_v1/query.sql
scholtzan: Ah, I see

scholtzan (approved):

PR #1323, opened 2020-09-17
Bug 1635906 Add bqetl support for scripts and script for AET lookup

jklukas:
There are some code changes here for DAG generation and for testing.

PR #1335, opened 2020-09-23
Bug 1635906 Add client ID lookup table for AET

jklukas:
Builds on top of the eco_uid lookup table added in #1323


scholtzan (approved):

sql/account_ecosystem_derived/ecosystem_client_id_lookup_v1/metadata.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - jklukas@mozilla.com
+labels:
+  application: aet
+  schedule: daily
+  incremental: true
+scheduling:
+  dag_name: bqetl_account_ecosystem
+  depends_on_past: True
+  # We access a restricted table for getting an HMAC key, so cannot dry run
+  # and must explicitly list referenced tables.
```
scholtzan, line 14: I guess this query will need to get added to `SKIP` in dryrun

Issue #860, opened 2020-03-30
Move all python logic inside bigquery_etl module


scholtzan:
The code should be refactored so that the python logic lives inside the `bigquery_etl` module and scripts become light entrypoints into that codebase. That would also make writing tests a lot more convenient.

See also https://github.com/mozilla/bigquery-etl/pull/839#discussion_r395075615

PR #1336, opened 2020-09-23
Move python scripts logic to bigquery_etl module

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/860

I just moved the scripts into bigquery_etl but left the logic unchanged.


jklukas (approved):
It doesn't look like telemetry-airflow contains any `python` invocations of `script/` files, so I think we should be good with just changing the shebang line as you've done here.

PR #1337, opened 2020-09-23
Update stripe schemas

relud:
and include them in load jobs when available

the schemas for each resource type were manually updated with all of the columns described in https://stripe.com/docs/api/, so they should be complete now, instead of only covering fields that were actually in use.


bigquery_etl/stripe/__init__.py
```diff
@@ -166,6 +190,10 @@ def import_(api_key, date, table, file, resource):
                 job.result()
             except Exception as e:
                 print(f"{job.job_id} failed: {e}", file=sys.stderr)
+                for error in job.errors or ():
+                    message = error.get("message")
+                    if message and message != getattr(e, "message", None):
+                        print(message, file=sys.stderr)
```
relud, line 196: the extra lines here include information like what field is missing from the schema, and the row at which the job failed.

acmiyaguchi (approved):
1337

r+

PR #1338, opened 2020-09-24
Don't assert on import when running bqetl command

relud:
this is breaking stripe imports


acmiyaguchi (approved):
This is because the file might be relative to a site-packages directory, presumably? Sorry about this.

relud:
> This is because the file might be relative to a site-packages directory, presumably?

yeah, when bqetl is installed (in docker), it runs out of site-packages.

> Sorry about this.

no worries, it's not actually in-use yet.

PR #1344, opened 2020-09-24
CLI support name patterns with wildcards

scholtzan:
This adds support for expressing query and UDF names with wildcards. Instead of specifying paths to queries or UDFs, the CLI now works with query and UDF names. This way, users do not necessarily need to know the underlying folder structure for using the CLI.

The CLI now supports things like:
`bqetl query info "*.clients*"` or `bqetl query info "telemetry_derived.*_v6"`

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -209,7 +209,7 @@ def create(name, path, owner, init):
 )
 def schedule(name, path, dag, depends_on_past, task_name):
     """CLI command for scheduling a query."""
-    query_files = queries_matching_name_pattern(name, path)
+    query_files = _queries_matching_name_pattern(name, path)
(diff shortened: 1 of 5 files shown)


bigquery_etl/cli/query.py
```diff
@@ -20,9 +20,40 @@
 
 
 QUERY_NAME_RE = re.compile(r"(?P<dataset>[a-zA-z0-9_]+)\.(?P<name>[a-zA-z0-9_]+)")
+QUERY_FILE_RE = re.compile(
```
jklukas, line 23: Using "sql file" rather than "query file" terminology would be more precise. I was interpreting "query file" as "query.sql" specifically.

jklukas (approved):

PR #1349, opened 2020-09-25
Bug 1667889 - Add stripe ETL to support Mozilla VPN dashboard

relud:
[Bug 1667889 - VPN subscription dashboard](https://bugzilla.mozilla.org/show_bug.cgi?id=1667889)
[Re:Dash Mozilla VPN Redux](https://sql.telemetry.mozilla.org/dashboard/mozilla-vpn-redux)

### DIFF
--- a/bigquery_etl/dryrun.py
+++ b/bigquery_etl/dryrun.py
@@ -63,6 +63,15 @@ SKIP = {
     "sql/moz-fx-data-shared-prod/apple_app_store/metrics_total/query.sql",
     "sql/moz-fx-data-shared-prod/monitoring/telemetry_distinct_docids_v1/query.sql",
     "sql/moz-fx-data-shared-prod/revenue_derived/client_ltv_normalized/query.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/plan_events/view.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/plans_v1/init.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/plans_v1/query.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/product_events/view.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/products_v1/init.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/products_v1/query.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/subscription_events/view.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/subscriptions_v1/init.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/subscriptions_v1/query.sql",
     # Already exists (and lacks an "OR REPLACE" clause)
     "sql/moz-fx-data-shared-prod/org_mozilla_firefox_derived/clients_first_seen_v1/init.sql",  # noqa E501
     "sql/moz-fx-data-shared-prod/org_mozilla_firefox_derived/clients_last_seen_v1/init.sql",  # noqa E501
--- a/sql/moz-fx-data-shared-prod/stripe_derived/plans_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/stripe_derived/plans_v1/metadata.yaml
@@ -1,0 +1,15 @@
+friendly_name: Stripe Plans
+description: Stripe plans as of UTC midnight. See bug 1667889.
+owners:
+  - dthorn@mozilla.com
+labels:
+  application: stripe
+  schedule: daily
+scheduling:
+  dag_name: bqetl_stripe
+  date_partition_parameter: date
+  depends_on:
+    - dag_name: stripe
+      task_id: stripe_import_events
+      schedule_interval: daily
+  referenced_tables: []
--- a/sql/moz-fx-data-shared-prod/stripe_derived/subscription_events/view.sql
+++ b/sql/moz-fx-data-shared-prod/stripe_derived/subscription_events/view.sql
@@ -1,0 +1,35 @@
+CREATE OR REPLACE VIEW
+  subscription_events
+AS
+WITH raw AS (
+  SELECT
+    created AS event_timestamp,
+    `data`.subscription.*,
+  FROM
+    stripe_external.events_v1
+  WHERE
+    `data`.subscription IS NOT NULL
+  UNION ALL
+  SELECT
+    created AS event_timestamp,
+    *,
+  FROM
+    stripe_external.subscriptions_v1
+)
+SELECT
+  -- limit fields in stripe_derived so as not to expose sensitive data
+  created,
+  cancel_at,
+  cancel_at_period_end,
+  canceled_at,
+  customer,
+  ended_at,
+  event_timestamp,
+  id,
+  metadata,
+  plan.id AS plan,
+  start_date,
+  status,
+  trial_end,
+FROM
+  raw
--- a/dags.yaml
+++ b/dags.yaml
@@ -208,3 +208,12 @@ bqetl_account_ecosystem:
     email: ['jklukas@mozilla.com']
     retries: 2
     retry_delay: 30m
+
+bqetl_stripe:
+  schedule_interval: daily
+  default_args:
+    owner: dthorn@mozilla.com
+    start_date: '2020-10-05'
+    email: ['telemetry-alerts@mozilla.com', 'dthorn@mozilla.com']
+    retries: 2
+    retry_delay: 5m
--- a/dags/bqetl_stripe.py
+++ b/dags/bqetl_stripe.py
@@ -1,0 +1,71 @@
+# Generated via https://github.com/mozilla/bigquery-etl/blob/master/bigquery_etl/query_scheduling/generate_airflow_dags.py
+
+from airflow import DAG
+from airflow.operators.sensors import ExternalTaskSensor
+import datetime
+from utils.gcp import bigquery_etl_query
+
+default_args = {
+    "owner": "dthorn@mozilla.com",
+    "start_date": datetime.datetime(2020, 10, 5, 0, 0),
+    "email": ["telemetry-alerts@mozilla.com", "dthorn@mozilla.com"],
+    "depends_on_past": False,
+    "retry_delay": datetime.timedelta(seconds=300),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
+}
+
+with DAG("bqetl_stripe", default_args=default_args, schedule_interval="@daily") as dag:
+
+    stripe_derived__plans__v1 = bigquery_etl_query(
+        task_id="stripe_derived__plans__v1",
+        destination_table="plans_v1",
+        dataset_id="stripe_derived",
+        project_id="moz-fx-data-shared-prod",
+        owner="dthorn@mozilla.com",
+        email=["dthorn@mozilla.com", "telemetry-alerts@mozilla.com"],
+        date_partition_parameter="date",
+        depends_on_past=False,
+        dag=dag,
+    )
+
+    stripe_derived__products__v1 = bigquery_etl_query(
+        task_id="stripe_derived__products__v1",
+        destination_table="products_v1",
+        dataset_id="stripe_derived",
+        project_id="moz-fx-data-shared-prod",
+        owner="dthorn@mozilla.com",
+        email=["dthorn@mozilla.com", "telemetry-alerts@mozilla.com"],
+        date_partition_parameter="date",
+        depends_on_past=False,
+        dag=dag,
+    )
+
+    stripe_derived__subscriptions__v1 = bigquery_etl_query(
+        task_id="stripe_derived__subscriptions__v1",
+        destination_table="subscriptions_v1",
+        dataset_id="stripe_derived",
+        project_id="moz-fx-data-shared-prod",
+        owner="dthorn@mozilla.com",
+        email=["dthorn@mozilla.com", "telemetry-alerts@mozilla.com"],
+        date_partition_parameter="date",
+        depends_on_past=False,
+        dag=dag,
+    )
+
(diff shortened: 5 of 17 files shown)


bigquery_etl/stripe/charge.schema.json
```diff
@@ -1136,10 +1136,6 @@
     "type": "RECORD",
     "description": "undocumented card",
     "fields": [
-      {
-        "name": "id",
-        "type": "STRING"
-      },
```
relud, line 1142: this schema needed to be sorted so that it matched `data.event` in `stripe_external.events_v1`

sql/stripe/revenue/view.sql
```diff
@@ -0,0 +1,16 @@
+CREATE OR REPLACE VIEW
+  revenue
```
relud, line 2: I still need to evaluate whether this is the right way to calculate revenue

sql/stripe/funnel/view.sql
```diff
@@ -0,0 +1,65 @@
+CREATE OR REPLACE VIEW
+  funnel
```
relud, line 2: r? @felixlawrence on this view

My goal here is to give a useful representation of the `subscribed` stage of the funnel, with room to add other stages later.

For context, the rest of the queries here (except the revenue view) are just to extract the most up to date values we have for each resource type, and hide any fields that shouldn't be accessible to everyone in re:dash.

felixlawrence: I think I don't understand what funnel we're trying to represent here.

The usual way I think of a funnel is as a series of ~tubes~ steps. Each of the steps has a starting event, and a range of possible outcomes. For each outcome, we can calculate the conversion rate towards that outcome.

When writing queries to analyse a funnel, I'd typically write a query for each step: start with a list of the starting events in a given time window, typically `LEFT JOIN` to the corresponding outcome event to see what the outcome was (limiting the conversion window to some fixed period of time, to avoid lagging metrics where history constantly gets rewritten), and optionally apply a `DISTINCT user_id` to the count of starting events (denominator) and the count of each outcome (numerator).

I can see how the `subscriptions_v1` table has already done some of the `LEFT JOIN`ing for us...

I can see how "start entering payment details" -> "successfully set up payment" is a one-step funnel, and "start entering payment details" -> "successfully set up payment" -> "renew for your second month" is a two-step funnel.

I can see a few funnels being relevant to the VPN product:
* Arrives on landing page -> clicks CTA on landing page -> signs up/in to a FxA -> ... -> Adds payment method -> first payment (or CC auth?) successful -> renews for 2nd month -> renews for 3rd month
* Renews for nth month -> renews for n+1th month
* Subscribes -> runs client app -> runs client app on 2nd device -> runs client app on 3rd device
* Has payment failure -> resolves payment failure

This view is, for each date, sorting subscriptions by their status and counting them, which doesn't match my understanding of funnel analyses - so it would be good to sync on Monday about what the underlying objectives here are, so I'd be better placed to review whether it's meeting them!

sql/stripe/funnel/view.sql
```diff
@@ -0,0 +1,96 @@
...
+      "cancelled"
+    WHEN
+      `date` = end_date
+    THEN
+      "failed renewal"
+    ELSE
+      "active"
+    END
+    AS status,
+    -- TODO decide whether to count customers or subscriptions, because in a single day a customer
```
felixlawrence, line 64: Yeah, I smell trouble here. We need to count the number of active (and trialling) subscriptions at UTC midnight each day, and we need to account for the day-to-day changes in those numbers (i.e. count new subscriptions, failed renewals, etc).

If one subscription can end in multiple ways (e.g. a payment fails then it's manually ended, perhaps on a later date) then we don't want to double-count its end (unless there's a balancing number of things counted in the "start" data).

Are we at risk of that here or does Stripe's data structure save us from it?

felixlawrence: W.r.t. "failed setup", I think that this kind of thing should be part of the funnel analysis; the visualisations we're currently aiming at should start at the point that the user has already signed up (or at least set up a trial).

sql/stripe/funnel/view.sql
```diff
@@ -0,0 +1,96 @@
...
+    CASE
+    WHEN
+      incomplete_expired
+    THEN
+      "failed setup"
+    WHEN
+      TIMESTAMP(`date`) < trial_end
+    THEN
+      "trialing"
+    -- TODO decide whether subscriptions that go through a trial are ever counted as "new"
```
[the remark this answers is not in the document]
relud: I think we should count trials with phases "new trial" (on the day the trial starts), "trialing", "trial converted" (on the day the paid subscription starts), and "trial cancelled" (on the day the trial ended without becoming paid)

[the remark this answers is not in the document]
relud: so essentially 2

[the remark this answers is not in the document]
felixlawrence: Ah - one wrinkle this might have highlighted! I think the count of active subscriptions for e.g. 2020-09-01 should include all subscriptions that were active at 2020-09-01 00:00:00 UTC, regardless of whether they are "new" (started immediately before this moment) or whether they were cancelled immediately after this moment. So a subscription can be "active" _and_ another state on a given date.

Basically, we're querying two quite-separate quantities:

1. "how many subscriptions were active/in a trial at each of these _moments_"
2. "how many subscriptions made transition _x_ during each of these _time periods_"

In (1) we're counting users in a state at midnight each day, and in (2) we're counting how many users change between sets of states during the time between midnights.

sql/stripe/funnel/view.sql
```diff
@@ -0,0 +1,96 @@
...
+  funnel
+AS
+WITH subscriptions AS (
+  SELECT
+    plan,
+    status = "incomplete_expired" AS incomplete_expired,
+    cancel_at_period_end
+    OR "cancelled_for_customer_at" IN (SELECT key FROM UNNEST(metadata)) AS cancelled_for_customer,
+    -- count subscriptions from the day after they start to the day after they end
+    -- TODO: decide whether to count users at the end of the day or the beginning of the day
```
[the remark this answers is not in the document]
relud: > convention (2) - so let's adopt that

:+1:

> "active" and "trialing" subscriptions

yeah, i'm going to break those up so active/trialing is separate from status

sql/stripe/funnel/view.sql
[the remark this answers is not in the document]
relud: going to split this into two queries, one for active subs by date, and a separate one for transitions

sql/stripe/subscription_transitions/view.sql
```diff
@@ -0,0 +1,100 @@
...
+WITH subscriptions AS (
+  SELECT
+    products_v1.name AS product,
+    (
+      cancel_at_period_end
+      OR "cancelled_for_customer_at" IN (SELECT key FROM UNNEST(subscriptions_v1.metadata))
+    ) AS cancelled_for_customer,
+    DATE(start_date) AS start_date,
+    DATE(trial_end) AS trial_end,
+    DATE(ended_at) AS end_date,
```
relud, line 13: so transitions like "new" occurs on the same date the subscription starts to be counted as `active`, and transitions like "cancelled" occur on the day after the last date subscription was counted as `active`

sql/stripe/subscription_volume/view.sql
```diff
@@ -0,0 +1,32 @@
...
+  stripe_derived.subscriptions_v1
+LEFT JOIN
+  stripe_derived.plans_v1
+ON
+  (subscriptions_v1.plan = plans_v1.id)
+LEFT JOIN
+  stripe_derived.products_v1
+ON
+  (plans_v1.product = products_v1.id)
+CROSS JOIN
```
felixlawrence, line 19: Oh wow, haven't seen this pattern before! It took me a couple min to wrap my head around. If this isn't already a common pattern, might this query be more readably expressed as the following?

(forgive the `moz-fx-data-shared-prod` everywhere - it was needed so that I could lint the query)

```WITH date_range AS (
    SELECT `date`
    FROM UNNEST(GENERATE_DATE_ARRAY(
        (SELECT DATE(MIN(start_date)) FROM moz-fx-data-shared-prod.stripe_derived.subscriptions_v1),
        CURRENT_DATE,
        INTERVAL 1 DAY
    )) AS `date`
)
SELECT
  `date`,
  products_v1.name AS product,
  COALESCE(`date` >= DATE(trial_end), TRUE) AS paying,
  COUNT(DISTINCT customer) AS subscription_count,
FROM
  date_range
LEFT JOIN
  moz-fx-data-shared-prod.stripe_derived.subscriptions_v1
ON
  (DATE(start_date) = `date`)
LEFT JOIN
  moz-fx-data-shared-prod.stripe_derived.plans_v1
ON
  (subscriptions_v1.plan = plans_v1.id)
LEFT JOIN
  moz-fx-data-shared-prod.stripe_derived.products_v1
ON
  (plans_v1.product = products_v1.id)
WHERE
  status NOT IN ("incomplete", "incomplete_expired")
GROUP BY
  `date`,
  product,
  paying
```

sql/stripe_derived/charge_events/view.sql
```diff
@@ -0,0 +1,31 @@
...
+  SELECT
+    created AS event_timestamp,
+    *,
+  FROM
+    stripe_external.charges_v1
+  -- TODO remove this filter in favor of deleting the unwanted data from the table
+  WHERE
+    DATE(created) < DATE '2020-08-10'
+)
+SELECT
```
felixlawrence, line 22: Are we intentionally keeping customer IDs and subscription IDs out of this? If not, shouldn't we include them here so we can e.g. join to subscriptions?

sql/stripe/subscription_transitions/view.sql
```diff
@@ -0,0 +1,100 @@
...
+  SELECT
+    `date`,
+    product,
+    transition,
+    -- make count negative for subscription ended counts
+    IF(
+      transition IN ("trial cancelled", "trial convert failed", "cancelled", "renew failed"),
+      -1,
+      1
+    ) * COUNT(DISTINCT customer) AS subscription_count,
```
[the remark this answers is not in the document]
relud: ooh, good point. yeah, let's drop distinct for that reason

sql/stripe/subscription_volume/view.sql
felixlawrence: Oh man I totally lost the plot writing that query, sorry

felixlawrence: I think I meant the following?
```
LEFT JOIN
  moz-fx-data-shared-prod.stripe_derived.subscriptions_v1
ON
  DATE(start_date) >= `date` AND DATE(ended_at) < `date`
```

felixlawrence: If I understood correctly, @relud pointed out over Zoom that the left join on `subscriptions_v1` is less efficient (it creates more rows that then need to be filtered out by the join condition) than his approach

sql/stripe_derived/charge_events/view.sql
relud: I'm going to remove charge_events for now, as it's not being used within this PR

sql/stripe/subscription_volume/view.sql
relud: as discussed in zoom, i'm going to keep the GENERATE_DATE_ARRAY, but isolate the application of it via CTE so that it's easier to read

sql/moz-fx-data-shared-prod/stripe_derived/subscription_events/view.sql
```diff
@@ -0,0 +1,35 @@
...
+    `data`.subscription IS NOT NULL
+  UNION ALL
+  SELECT
+    created AS event_timestamp,
+    *,
+  FROM
+    stripe_external.subscriptions_v1
+)
+SELECT
+  -- limit fields in stripe_derived so as not to expose sensistive data
```
acmiyaguchi, line 20:
```suggestion
  -- limit fields in stripe_derived so as not to expose sensitive data
```

sql/moz-fx-data-shared-prod/stripe_derived/plans_v1/metadata.yaml
```diff
@@ -0,0 +1,15 @@
+friendly_name: Stripe Plans
+description: Stripe plans as of UTC midnight
```
acmiyaguchi, line 2: Might be useful to include the bug number in the metadata section for reference.

acmiyaguchi (approved):
This looks good to me, assuming all of the SQL is valid.

bigquery_etl/dryrun.py
```diff
@@ -63,6 +63,15 @@
...
     "sql/moz-fx-data-shared-prod/revenue_derived/client_ltv_normalized/query.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/plan_events/view.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/plans_v1/init.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/plans_v1/query.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/product_events/view.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/products_v1/init.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/products_v1/query.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/subscription_events/view.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/subscriptions_v1/init.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/subscriptions_v1/query.sql",
```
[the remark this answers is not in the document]
relud: I can manually dry_run it to make sure it passes, but the dry_run script doesn't have sufficient permissions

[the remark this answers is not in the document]
relud: all of the dry_runs passed without issue

PR #1354, opened 2020-09-29
Fix #1329 - Use app_build_id as app_version in GLAM fenix nightly

### LINKED ISSUE #1329: Inconsistent versioning scheme in `org_mozilla_fenix` causes GLAM Fenix aggregates to be low
Current aggregates of the `org_mozilla_fenix_glam_nightly` are unexpectedly low. This is caused by pruning of clients in the `clients_scalar` and `clients_histogram` tables using the `latest_versions` table. The logic keeps the last 3 major revisions, based on the Firefox versioning scheme (release every 4-6 weeks with semantic versioning e.g. 80.0.1). `org_mozilla_fenix` and `org_mozilla_fennec_aurora` do not follow this convention, instead encoding the build date into the display version.

One example of this is follows:

```
Nightly 200820 06:08 
```

https://sql.telemetry.mozilla.org/queries/74943/source

One potential solution is to retain all versions and to filter out client ids based on their last seen build date. This assumes that build ids encode the build date in their string. However, this would work consistently across Fenix products.

acmiyaguchi:
#1329

---

I tested this query using the following steps in `glam_etl_dev`.

```bash
# generate the list of tables to delete
bq ls -n 1000 --format json glam-fenix-dev:glam_etl_dev | jq -r '.[].id' |  \
	grep aggregates | grep -v daily | grep nightly | grep org_mozilla_fenix_glam | \
	xargs -I {} echo "bq rm {}"

# run the echo'ed commands

bqetl glam glean backfill-incremental --dataset glam_etl_dev org_mozilla_fenix_glam_nightly 2020-08-01 2020-08-17
```

The resulting scalar aggregates table has a reasonable number of clients in the table during this period:

image

Since this change requires updating the clients_histogram_aggregates table (dropping the `latest_version` field in the `histogram_aggregates` struct), I'll be copying the table and running the a modified version of the following query for beta/release:

```sql
WITH
  unnested AS (
  SELECT
    * EXCEPT(histogram_aggregates)
  FROM
    glam_etl_dev.org_mozilla_fenix_glam_beta__clients_histogram_aggregates_v1,
    UNNEST(histogram_aggregates)),
  summed AS (
  SELECT
    * EXCEPT(latest_version,
      value),
    mozfun.map.sum(ARRAY_CONCAT_AGG(value)) AS value
  FROM
    unnested
  GROUP BY
    sample_id,
    client_id,
    ping_type,
    os,
    app_version,
    app_build_id,
    channel,
    metric,
    metric_type,
    key,
    agg_type )
SELECT
  * EXCEPT(metric,
    metric_type,
    key,
    agg_type,
    value),
  ARRAY_AGG( STRUCT< metric STRING, metric_type STRING, key STRING, agg_type STRING, aggregates ARRAY<STRUCT<key STRING, value INT64>> >(metric,
      metric_type,
      key,
      agg_type,
      value)) AS histogram_aggregates
FROM
  summed
GROUP BY
  sample_id,
  client_id,
  ping_type,
  os,
  app_version,
  app_build_id,
  channel
```


bigquery_etl/glam/templates/clients_histogram_aggregates_v1.init.sql
```diff
@@ -10,7 +10,6 @@ CREATE TABLE IF NOT EXISTS
     channel STRING,
     histogram_aggregates ARRAY<
       STRUCT<
-        latest_version INT64,
```
acmiyaguchi, line 13: This field causes issues when not using `latest_versions` to filter app versions. This doesn't apply on nightly because app versions are just build versions. 

I double checked the desktop aggregates, and it looks like the bucket count query doesn't reference latest versions at all, unlike the query in this module: https://github.com/mozilla/bigquery-etl/blob/master/sql/telemetry_derived/clients_histogram_bucket_counts_v1/query.sql

While it is being used in the histogram aggregates for filtering against the new data in the desktop queries, it is mistakenly being aggregated in the bucket counts here. I'm removing it since it doesn't need to be in the query.

acmiyaguchi:
Here's the associated diff: https://gist.github.com/acmiyaguchi/bd7fe4d0834e2abc3ed9f2ebc1e066e1

emtwo (approved):
Looks good to me!

I think maybe your reference to #1322 isn't correct though?

acmiyaguchi:
It should be #1329. I'm currently testing the query that will be used for migrating the clients histogram table. I'll comment with some of the outputs from my testing for documentation before merging,

PR #1357, opened 2020-09-30
Fix #1356 - Add extra validation before casting 8 digit build ids

### LINKED ISSUE #1356: udf.fenix_build_to_datetime breaks on input `11831860`: Input calculates to invalid datetime: 2019-01-01 0018:60:00
### summary

Running the following query leads to `Input calculates to invalid datetime: 2019-01-01 0018:60:00`:

```sql
select `moz-fx-data-shared-prod`.udf.fenix_build_to_datetime("11831860")
```

I expect the query to fail gracefully (or parse the build id).

### details
I ran the following query with a puzzling error:

```sql
select 
  `moz-fx-data-shared-prod`.udf.fenix_build_to_datetime(app_build_id) as build_date, 
  count(distinct client_id) as n
from `glam-fenix-dev`.glam_etl.org_mozilla_fenix_glam_beta__view_clients_daily_scalar_aggregates_v1
where app_version < 70
group by 1
order by 1
```

I narrowed this down to `11831860`.

```sql
%%bigquery
select distinct
  # `moz-fx-data-shared-prod`.udf.fenix_build_to_datetime(app_build_id) as build_date,
  app_build_id,
  count(*)
from `glam-fenix-dev`.glam_etl.org_mozilla_fenix_glam_beta__view_clients_daily_scalar_aggregates_v1
where app_version < 70 
  and length(app_build_id) = 8 
  and SUBSTR(app_build_id, 5, 2) = "18" 
  and SUBSTR(app_build_id, 7, 2) = "60"
group by 1
order by 1
```

Trying to parse this leads to the following error: `Input calculates to invalid datetime: 2019-01-01 0018:60:00`

The relevant code path is here:

https://github.com/mozilla/bigquery-etl/blob/3caa5eb38207985e6cd75a1a65b06480ab109ce6/udf/fenix_build_to_datetime/udf.sql#L7-L21

cc @jklukas do you have any ideas on what might be going on here?

acmiyaguchi:
This fixes the `fenix_build_to_datetime` for 8 digit builds with invalid minutes (and hours). See #1356 for details.

### DIFF
--- a/udf/fenix_build_to_datetime/udf.sql
+++ b/udf/fenix_build_to_datetime/udf.sql
@@ -6,18 +6,23 @@ CREATE OR REPLACE FUNCTION udf.fenix_build_to_datetime(app_build STRING) AS (
   THEN
     -- Ideally, we would use PARSE_DATETIME, but that doesn't support
     -- day of year (%j) or the custom single-character year used here.
-    DATETIME_ADD(
-      DATETIME(
-        2018 + SAFE_CAST(
-          SUBSTR(app_build, 1, 1) AS INT64
-        ), -- year
-        1, -- placeholder month
-        1, -- placeholder year
-        SAFE_CAST(SUBSTR(app_build, 5, 2) AS INT64),
-        SAFE_CAST(SUBSTR(app_build, 7, 2) AS INT64),
-        0  -- seconds is always zero
+    IF(
+      SUBSTR(app_build, 5, 2) < "24"
+      AND SUBSTR(app_build, 7, 2) < "60",
+      DATETIME_ADD(
+        DATETIME(
+          2018 + SAFE_CAST(
+            SUBSTR(app_build, 1, 1) AS INT64
+          ), -- year
+          1, -- placeholder month
+          1, -- placeholder year
+          SAFE_CAST(SUBSTR(app_build, 5, 2) AS INT64),
+          SAFE_CAST(SUBSTR(app_build, 7, 2) AS INT64),
+          0  -- seconds is always zero
+        ),
+        INTERVAL SAFE_CAST(SUBSTR(app_build, 2, 3) AS INT64) - 1 DAY
       ),
-      INTERVAL SAFE_CAST(SUBSTR(app_build, 2, 3) AS INT64) - 1 DAY
+      NULL
     )
   WHEN
     10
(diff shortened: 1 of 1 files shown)


udf/fenix_build_to_datetime/udf.sql
```diff
@@ -6,18 +6,23 @@ CREATE OR REPLACE FUNCTION udf.fenix_build_to_datetime(app_build STRING) AS (
...
-        2018 + SAFE_CAST(
-          SUBSTR(app_build, 1, 1) AS INT64
-        ), -- year
-        1, -- placeholder month
-        1, -- placeholder year
-        SAFE_CAST(SUBSTR(app_build, 5, 2) AS INT64),
-        SAFE_CAST(SUBSTR(app_build, 7, 2) AS INT64),
-        0  -- seconds is always zero
+    IF(
+      SUBSTR(app_build, 5, 2) < "24"
```
acmiyaguchi, line 10: This is actually not a problem in the dataset I'm currently querying, but might as well validate this too.

jklukas (changes requested):
Should we tolerate minute=60 by adding an hour rather than returning null?

I think that @chutten's original logic would handle this case. I refactored this in https://github.com/mozilla/bigquery-etl/pull/1250 to construct a DATETIME out of a list of ints with the intent of making it more readable, but the previous approach of using multiple DATETIME_ADD calls likely handles this case naturally.

I'm definitely open to discussion, but my preference would be to restore chutten's logic and confirm that it reasonably handles these cases.

acmiyaguchi:
No, I don't think so. If we extend the logic beyond the 60th minute, it doesn't hold up well. Should`11119999` should be treated as the 99th minute of the 99th hour of the day? It shouldn't be possible in the first place, since the SimpleDateFormat api is supposed to return a minute between 0-59.

The number of bad builds since 2020-06-01 has been ~35, so rejecting them is well within the tolerance of noise.

udf/fenix_build_to_datetime/udf.sql
jklukas: I'd prefer to move this IF into the CASE to avoid the extra level of nesting.

```
CASE
WHEN LENGTH(app_build) = 8 AND SUBSTR(app_build, 5, 2) < "24" AND SUBSTR(app_build, 7, 2) < "60"
THEN ...
WHEN LENGTH(app_build) = 10
THEN ...
ELSE null
END
```

jklukas (approved):
r+wc

PR #1359, opened 2020-09-30
Bug 1661250 - Add Pioneer support to Shredder

akkomar:
This adds Pioneer support to Shredder with automatic table discovery.


relud (approved):

whd (approved):
I'm currently unsure if/how we'll leverage the `_derived` datasets in pioneer, but we may want to traverse the study-specific analysis projects as well.

akkomar:
> I'm currently unsure if/how we'll leverage the `_derived` datasets in pioneer, but we may want to traverse the study-specific analysis projects as well.

@whd : are study analysis projects created automatically or by hand? I'm wondering if we can reliably discover them or maybe just pass them as a config parameter.

whd:
> @whd : are study analysis projects created automatically or by hand? I'm wondering if we can reliably discover them or maybe just pass them as a config parameter.

They're created by terraform configuration and are accessible as outputs of https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-pioneer/tf/prod/envs/prod/study-projects/outputs.tf#L22-L31, so it would not be difficult to pass them in as a configuration parameter. In theory they can be discovered from their naming convention under the gcp folder they are in but I'd opt for configuration instead (project naming conventions will possibly change).

akkomar:
> > @whd : are study analysis projects created automatically or by hand? I'm wondering if we can reliably discover them or maybe just pass them as a config parameter.
> 
> They're created by terraform configuration and are accessible as outputs of https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-pioneer/tf/prod/envs/prod/study-projects/outputs.tf#L22-L31, so it would not be difficult to pass them in as a configuration parameter. In theory they can be discovered from their naming convention under the gcp folder they are in but I'd opt for configuration instead (project naming conventions will possibly change).

Added an option to pass in study name and project id.

bigquery_etl/shredder/config.py
```diff
@@ -474,3 +475,85 @@ def find_experiment_analysis_targets(pool, client, project=EXPERIMENT_ANALYSIS):
...
+            pioneer_target(
+                table=qualified_table_id(table), project=PIONEER_PROD
+            ): sources[table.dataset_id]
+            for table in __get_tables_with_pioneer_id__(derived_datasets)
+        },
+        **{
+            # tables with pioneer_id located in study analysis projects
+            pioneer_target(
+                table=qualified_table_id(table), project=analysis_project
+            ): sources[study_name + "_stable"]
```
whd, line 550: What's the `_stable` here for? The analysis projects will only have an `analysis` dataset that needs to be inspected.

bigquery_etl/shredder/delete.py
```diff
@@ -44,6 +45,27 @@
...
+
+
+parser.add_argument(
+    "--pioneer-study-projects",
+    "--pioneer_study_projects",
+    default={},
+    metavar="STUDY_NAME_TO_PROJECT_ID_LIST",
+    help="Pioneer study-specific analysis projects to include in data deletion."
+    "This should be comma-delimited list of study-name:project-id pairs, e.g.:"
+    "study-one:project-one,study-two:project-two",
```
whd, line 66: Why do we need the study project name here? There are two datasets in each analysis project, `analysis` and `study-name`, but the latter is all views into the main pioneer project and won't need to be processed.

whd (commented):
I think this can be simplified. We don't actually have study name available from terraform output, except as a substring of the analysis projects themselves. If you need a unique identifier for each project the project id should be sufficient since there will only be one dataset that needs to be considered.

akkomar:
> I think this can be simplified. We don't actually have study name available from terraform output, except as a substring of the analysis projects themselves. If you need a unique identifier for each project the project id should be sufficient since there will only be one dataset that needs to be considered.

Oh, you're right, I missed that the study name isn't in the output. I'm using study output to find matching `deletion-request` table (hence the `sources[study_name + "_stable"]`). I wanted to avoid that but I can extract it from a project name instead.
Re. `analysis` dataset - that's great, will just look at this one.

bigquery_etl/shredder/config.py
akkomar: This is used to find a matching deletion request table. I added a comment explaining this.

whd (approved):

PR #1361, opened 2020-09-30
Remove `total_searches`

xluo-ds:
The way`total_searches` is defined as sum over several other search-related metrics is not proper (but ok in determining whether user searched as used in this case). Misusage of `total_searches` value can be misleading, so we prefer to remove it.


scholtzan (approved):
I fixed some of the tests that failed and fixed the SQL formatting.

There is also a CLI tool which makes interacting with the bigquery-etl repo easier, which can be installed via ` pip install mozilla-bigquery-etl`. This tool can be used to validate and re-format queries by running `bqetl format`

fbertsch (changes requested):
This is going to fail on the next run if we just drop this field as-is. It would require pausing ETL, moving the dataset to a new location, moving it back to the original location sans the field, and then starting the ETL back up.

Instead, why don't we drop this field from the view. Then the tests don't need to change and we don't need to alter the data.

xluo-ds:
> This is going to fail on the next run if we just drop this field as-is. It would require pausing ETL, moving the dataset to a new location, moving it back to the original location sans the field, and then starting the ETL back up.
> 
> Instead, why don't we drop this field from the view. Then the tests don't need to change and we don't need to alter the data.

Is the failure due to unmatching schema? If that's the case, I agree with Frank to just remove it from view. What do you think @scholtzan?

Also, I should mention earlier that this is the first time that I am working with bigquery-etl, so please bear with me. Thank you :)

scholtzan:
Yes, removing the field from the [search_clients_last_seen view](https://github.com/mozilla/bigquery-etl/blob/master/sql/search/search_clients_last_seen_v1/view.sql) is the best way to do this.

xluo-ds:
> Yes, removing the field from the [search_clients_last_seen view](https://github.com/mozilla/bigquery-etl/blob/master/sql/search/search_clients_last_seen_v1/view.sql) is the best way to do this.

Ok, in this case, I will close this PR and submit a new one under the view. Thank you all!

PR #1365, opened 2020-10-01
Use Glean telemetry for newer iOS Firefox data

jklukas:
See discussion in https://github.com/mozilla/gud/issues/117


sql/telemetry/nondesktop_clients_last_seen_v1/view.sql
```diff
@@ -99,9 +146,10 @@ SELECT
   country,
   locale,
   distribution_id,
-  metadata_app_version AS app_version
+  metadata_app_version AS app_version,
+  device AS device_model,
```
jklukas, line 150: Adding device_model to the view here, which will be needed to support https://github.com/mozilla/gud/issues/113

PR #1367, opened 2020-10-01
Dryrun stored procedures

fbertsch:


bigquery_etl/pytest_plugin/udf.py
```diff
@@ -96,5 +97,8 @@ def runtest(self):
             job_config = bigquery.QueryJobConfig(
                 use_legacy_sql=False, default_dataset=default_dataset
             )
-            job = bq.query(self.query, job_config=job_config)
+            job = bq.query(
+                self.query.replace(GENERIC_DATASET, default_dataset.dataset_id),
```
fbertsch, line 101: Google has confirmed to me that stored procedures will not support default datasets. To work around that we replace the dataset here.

bigquery_etl/udf/parse_udf.py
```diff
@@ -267,15 +265,22 @@ def udf_usage_definitions(text, raw_udfs=None):
     ]
 
 
-def persistent_udf_as_temp(raw_udf, raw_udfs=None):
-    """Transform persistent UDF into temporary UDF."""
-    sql = prepend_udf_usage_definitions(raw_udf, raw_udfs)
+def sub_local_routines(test, raw_udfs=None):
+    """
+    Transform persistent UDF into temporary UDFs.
```
scholtzan, line 270:
```suggestion
    Transform persistent UDFs into temporary UDFs.
```

scholtzan (approved):
Looks good!

PR #1371, opened 2020-10-02
Bug fixes for AET daily ETL

jklukas:
This fixes a bug about missing uids in the daily table as reported in
https://bugzilla.mozilla.org/show_bug.cgi?id=1637926#c5

It also adds some boolean fields to allow the daily table to specifically
address questions about whether a given client was 5-uri active per day.


tests/account_ecosystem_derived/ecosystem_user_id_lookup_v1/test_script/telemetry_stable.account_ecosystem_v4.yaml
```diff
@@ -0,0 +1,3 @@
+- submission_timestamp: 2020-04-01 22:09:03.243074 UTC
+  ecosystem_user_id: a0
+  previous_ecosystem_user_ids: null
```
scholtzan, line 3: Does this need to be something like?
```suggestion
  payload:
    ecosystem_user_id: a0
    previous_ecosystem_user_ids: null
```

scholtzan (approved):

PR #1374, opened 2020-10-02
[Bug 1664476] Stop anomdtct updates

scholtzan:
https://bugzilla.mozilla.org/show_bug.cgi?id=1664476

Depends on https://github.com/mozilla/telemetry-airflow/pull/1163


sql/telemetry_derived/deviations_v1/metadata.yaml
```diff
@@ -5,14 +5,7 @@ owners:
   - jmccrosky@mozilla.com
 labels:
   incremental: true
-  schedule: daily
   public_json: true
   public_bigquery: true
```
scholtzan, line 9: We'll keep the public table and the public json data in GCS around for now since they are linked in the public blog post, but updates for the data will stop.

jklukas (approved):

PR #1376, opened 2020-10-02
Update destination table for bigquery_etl_scheduled_query_usage

scholtzan:
I renamed the query but forgot to update the destination table name...


wlach (approved):
Maybe another argument to factor this code out into a common factory function. :)

PR #1377, opened 2020-10-02
Event counts

fbertsch:


mozfun/event_analysis/extract_event_counts/metadata.yaml
```diff
@@ -0,0 +1,6 @@
+description: >
+  Extract the events and their counts from an events string.
+
+  This function explicitly ignores event properties, and
+  retrieves just the counts of the top-level events.
+friendly_name: Event_analysis Extract Event Counts
```
jklukas, line 6:
```suggestion
friendly_name: Extract Event Counts
```

I don't think repeating the dataset name is necessary in the friendly name.

mozfun/event_analysis/extract_event_counts/udf.sql
```diff
@@ -0,0 +1,39 @@
+CREATE OR REPLACE FUNCTION event_analysis.extract_event_counts(events STRING)
+RETURNS ARRAY<STRUCT<index STRING, count INT64>> AS (
+  ARRAY(
+    SELECT AS STRUCT
+      index,
+      COUNT(*) AS count
+    FROM
+      UNNEST(regexp_extract_all(events, r'(.),')) AS index
```
jklukas, line 8: Is `index` the right name here? From my reading, it looks like each element of this array is a unicode char representing an event name, rather than a position in the array. Should this be called `event_char`?

jklukas: I might actually prefer that this fall into the general key/value struct paradigm, so would return `ARRAY<STRUCT<key STRING, value INT64>>`. That would make this compatible with `get_key`, etc.

jklukas (changes requested):

mozfun/event_analysis/extract_event_counts/metadata.yaml
fbertsch: I agree, we should change this default in the CLI.

mozfun/event_analysis/extract_event_counts/udf.sql
fbertsch: In general we've been using the word `index` to describe what the unicode char is that represents an event. I consider it akin to an index in an associative array.

I'm not sure that `udf.get_key` would be useful w.r.t. this udf. The use would be something like: `udf.get_key(event_analysis.extract_event_counts(events), "\U0025")` which should really just be an event count using the view structure.

jklukas (approved):

PR #1379, opened 2020-10-02
Update stripe schemas and fix subscription imports

relud:


bigquery_etl/stripe/event.schema.json
```diff
@@ -8974,6 +8974,10 @@
               }
             ]
           },
+          {
+            "name": "latest_attempt",
```
acmiyaguchi, line 8978: Where does this field come from? I don't see a reference to it in the import script or the event object definition (https://stripe.com/docs/api/events/object)

relud: the event object may contains the object that was modified, this part of the schema is for where it contains a setup_intent, which has this field: https://stripe.com/docs/api/setup_intents/object#setup_intent_object-latest_attempt

acmiyaguchi (approved):

PR #1380, opened 2020-10-02
Formalize product names in nondesktop_clients_last_seen

jklukas:
The `nondesktop_clients_last_seen_v1` view was developed mostly as an
internal implementation detail for downstream tables, but it has become
useful in its own right. This PR formalizes the view by providing an alias
without a version modifier and it adds a `product` field with application
names that are short but more meaningful than the `app_name` field.

See discussion in https://jira.mozilla.com/browse/DO-330 about confusion that
has resulted from the name "Fennec iOS" used in dashboards, etc. This is a
step toward reducing that kind of confusion.

This PR also adds `contributes_to_2019_kpi` and `contributes_to_2020_kpi` fields
as source of truth for how we count KPI metrics. That logic is currently
copied and pasted in several places, which could lead to errors.

This will need a fair amount of review from data users before moving forward.
It will also require backfilling several downstream tables and communicating
the change.

### DIFF
--- a/bigquery_etl/format_sql/format.py
+++ b/bigquery_etl/format_sql/format.py
@@ -87,7 +87,6 @@ SKIP = {
     "sql/moz-fx-data-shared-prod/telemetry_derived/smoot_usage_desktop_v2/query.sql",
     "sql/moz-fx-data-shared-prod/telemetry_derived/smoot_usage_fxa_v2/query.sql",
     "sql/moz-fx-data-shared-prod/telemetry_derived/smoot_usage_new_profiles_v2/query.sql",  # noqa E501
-    "sql/moz-fx-data-shared-prod/telemetry_derived/smoot_usage_nondesktop_v2/query.sql",
     "sql/moz-fx-data-shared-prod/telemetry_derived/surveygizmo_daily_attitudes/init.sql",  # noqa E501
     "sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/fenix_metrics.template.sql",  # noqa E501
     "sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/mobile_search_clients_daily.template.sql",  # noqa E501
--- a/sql/moz-fx-data-shared-prod/telemetry/fenix_clients_last_seen/view.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry/fenix_clients_last_seen/view.sql
@@ -1,0 +1,48 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.fenix_clients_last_seen`
+AS
+-- For context on naming and channels of Fenix apps, see:
+-- https://docs.google.com/document/d/1Ym4eZyS0WngEP6WdwJjmCoxtoQbJSvORxlQwZpuSV2I/edit#heading=h.69hvvg35j8un
+WITH fenix_union AS (
+  SELECT
+    *,
+    'org_mozilla_fenix' AS _dataset
+  FROM
+    `moz-fx-data-shared-prod.org_mozilla_fenix.baseline_clients_last_seen`
+  UNION ALL
+  SELECT
+    *,
+    'org_mozilla_fenix_nightly' AS _dataset
+  FROM
+    `moz-fx-data-shared-prod.org_mozilla_fenix_nightly.baseline_clients_last_seen`
+  UNION ALL
+  SELECT
+    *,
+    'org_mozilla_firefox' AS _dataset
+  FROM
+    `moz-fx-data-shared-prod.org_mozilla_firefox.baseline_clients_last_seen`
+  UNION ALL
+  SELECT
+    *,
+    'org_mozilla_firefox_beta' AS _dataset
+  FROM
+    `moz-fx-data-shared-prod.org_mozilla_firefox_beta.baseline_clients_last_seen`
+  UNION ALL
+  SELECT
+    *,
+    'org_mozilla_fennec_aurora' AS _dataset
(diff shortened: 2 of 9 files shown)


xluo-ds (commented):
Understood that these changes may lead to adjustment to KPI reporting. Leave for other reviewers to weigh in on approval :)

sql/telemetry/nondesktop_clients_last_seen_v1/view.sql
```diff
@@ -83,41 +36,138 @@ glean_union AS (
...
+-- Within this view, we present a `product` field that is intended to be
+-- short, readable, unambiguous, and reasonably close to the "official" names
+-- of the applications. The app_name field is a messy artifact of raw pings
+-- that we include only to help with debugging.
+-- For Android, we use the internal names "Fennec" and "Fenix" as the product name
+-- to disambiguate between the legacy and current implementations of Firefox for Android.
+--
+-- Expect the following mappings:
+--
+--   "Official" name               | product          | app_name
```
gkabbz, line 49: I really like this. Will this table live on dtmo or as a udf? Specifically asking with a focus on the official name. Would be a nice to have for it to be in a handy spot.

sql/telemetry_derived/firefox_nondesktop_day_2_7_activation_v1/query.sql
```diff
@@ -34,12 +12,13 @@ WITH base AS (
       AND BIT_COUNT(days_seen_bits << 1 & udf.bitmask_lowest_7()) > 0
     ) AS day_2_7_activated,
   FROM
-    `moz-fx-data-shared-prod.telemetry.nondesktop_clients_last_seen_v1`
+    `moz-fx-data-shared-prod.telemetry.nondesktop_clients_last_seen`
+  WHERE
+    contributes_to_2020_kpi
```
gkabbz, line 17: I like the addition of this where clause. Took a look a the [dashboard](https://datastudio.google.com/reporting/1L7dsFyqjT8XZHrYprYS-HCP5_k_gZGIb/page/0iERB) for the day 2-7 activation and it looks like it's currently only including products that contribute to the 2020 KPI only. I initially thought that was happening in the dashboard but it isn't. Looked at the query for creating the day_2_7 table and the clients_last_seen table and I'm not seeing where the exclusion is happening. The day 2-7 just miraculously knows what our KPI products are for 2020. Am I missing something?

sql/telemetry_derived/firefox_nondesktop_exact_mau28_v1/query.sql
```diff
@@ -9,51 +9,15 @@ SELECT
...
-    'Fenix',
-    'Firefox Preview',
-    'Fennec', -- Firefox for Android and Firefox for iOS
-    'Focus',
-    'Lockbox', -- Lockwise
-    'Zerda', -- Firefox Lite, previously called Rocket
-    'FirefoxForFireTV', -- Amazon Fire TV
-    'FirefoxConnect' -- Amazon Echo Show
-  )
-  -- There are also many strange nonsensical entries for os, so we filter here.
```
gkabbz, line 56: I think this context around why we filter for Android and iOS should carry through  to the new view

gkabbz (changes requested):

sql/telemetry_derived/firefox_nondesktop_day_2_7_activation_v1/query.sql
jklukas: Later on in this query there's an existing clause:

```
  AND product IN (
    'Fenix',
    'Firefox Preview',
    'Fennec Android',
    'Focus Android',
    'Fennec iOS',
    'Focus iOS',
    'Firefox Lite',
    'FirefoxConnect',
    'Lockwise Android'
  )
```

That's what's causing the filtering you see today. I'll need to remove that clause as part of this PR, since this `contributes_to_2020_kpi` line handles the same intention.

sql/telemetry/nondesktop_clients_last_seen_v1/view.sql
jklukas: I've refactored this logic to a new UDF `mozfun.norm.product_info` in this PR. That will produce the table in rendered documentation and be consistent with the implementation of the function.

sql/telemetry_derived/firefox_nondesktop_exact_mau28_v1/query.sql
jklukas: The logic for computing the `contributes_to_2020_kpi` field now includes checking of `os`, so a specific filter about OS is no longer needed in this view.

sql/mozfun/norm/product_info/metadata.yaml
```diff
@@ -0,0 +1,54 @@
+friendly_name: Normalize Product Info
+description: |
+  Returns a normalized `product` name and a `canonical_name` for a product
```
jklukas, line 3: @gkabbz Please take a look at this documentation for `mozfun.norm.product_info`. My intention is that this becomes the canonical reference for naming of mobile applications in telemetry. The function logic is generated from the markdown table.

jklukas:
If there are no objections to moving forward with the changes in this PR, then I'll prep documentation changes and plan to move this through to production. I'll send out some communication to data users as part of merging as well, so that folks can be aware of the change and what tables it affects.

gkabbz (approved):

jklukas:
Factored out the product_info UDF to https://github.com/mozilla/bigquery-etl/pull/1436 and that is now merged. Rebased this on top of master so that the product_info implementation is no longer part of this PR. I'll be testing the pieces and plan to deploy tomorrow.

PR #1382, opened 2020-10-05
Update ping selection criteria for fission monitoring tables

akkomar:
This adds filtering by experiment slug and filters out sessions where fission hasn't yet been enabled.

### DIFF
--- a/sql/telemetry_derived/fission_monitoring_crash_v1/query.sql
+++ b/sql/telemetry_derived/fission_monitoring_crash_v1/query.sql
@@ -1,15 +1,35 @@
-WITH crash_ping_data AS (
+WITH first_sessions AS (
+    -- these are the sessions where clients were enrolled in the experiment
+    -- since fission is enabled after restart, we want to filter them out from the main pings
+  SELECT
+    session_id
+  FROM
+    `moz-fx-data-shared-prod.telemetry.events`
+  WHERE
+    event_category = 'normandy'
+    AND normalized_channel = 'nightly'
+    AND event_method = 'enroll'
+    AND submission_date = @submission_date
+    AND event_string_value = 'bug-1660366-pref-ongoing-fission-nightly-experiment-nightly-83-100'
+),
+crash_ping_data_unfiltered AS (
   SELECT
     submission_timestamp,
     client_id,
-    -- TODO: update filtering when https://bugzilla.mozilla.org/show_bug.cgi?id=1667426 is finalized
+    payload.session_id,
     CASE
     WHEN
-      mozfun.map.get_key(environment.settings.user_prefs, 'fission.autostart') = 'true'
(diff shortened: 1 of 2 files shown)


sql/telemetry_derived/fission_monitoring_crash_v1/query.sql
```diff
@@ -41,21 +61,41 @@ WITH crash_ping_data AS (
...
     client_id,
-    -- TODO: update filtering when https://bugzilla.mozilla.org/show_bug.cgi?id=1667426 is finalized
+    payload.info.session_id,
     CASE
     WHEN
-      mozfun.map.get_key(environment.settings.user_prefs, 'fission.autostart') = 'true'
+      mozfun.map.get_key(
+        environment.experiments,
+        "bug-1660366-pref-ongoing-fission-nightly-experiment-nightly-83-100"
+      ).branch = 'fission-enabled'
```
godelstheory, line 91: Same here as for crashes:

Could a check be added to ETL to ensure that the `mozfun.map.get_key(environment.settings.user_prefs, 'fission.autostart') == 'true'`, when `branch == 'enabled'` and vice-versa?

sql/telemetry_derived/fission_monitoring_crash_v1/query.sql
```diff
@@ -1,15 +1,35 @@
...
     THEN
       'enabled'
     WHEN
-      mozfun.map.get_key(environment.settings.user_prefs, 'fission.autostart') = 'false'
```
godelstheory, line 12: Please add this preference be output for the `moz-fx-data-shared-prod.telemetry_derived.fission_monitoring_crash` table. Then a cross-check can be performed during analysis to make sure the preferences match.

godelstheory: This doesn't need to be done for the `fission_monitoring_main_v1` table, as the `environment.settings.user_prefs` payload is already available.

godelstheory (commented):
Looks good.

Added a comment about adding a column to the `fission_monitoring_crash` table that contains the `fission.autostart` pref. This will allow cross-checking during analysis to be sure the pref aligns with the branch.

sql/telemetry_derived/fission_monitoring_crash_v1/query.sql
akkomar: We should use the new pref instead, we can add it to the crash table when it [lands in the schema](https://bugzilla.mozilla.org/show_bug.cgi?id=1669208), then we can look into checks.
I wouldn't delay this change though given the tight timeline.

godelstheory: Sounds good.

godelstheory (approved):

PR #1384, opened 2020-10-05
Omit lists of stripe resources that have a separate table

relud:
fixes an issue where stripe event imports were failing because there's a payment intent event with too many charges attached

### DIFF
--- a/bigquery_etl/stripe/__init__.py
+++ b/bigquery_etl/stripe/__init__.py
@@ -14,6 +14,22 @@ from stripe.api_resources.abstract import ListableAPIResource
 import click
 import stripe
 
+# omit lists of resources that have a separate table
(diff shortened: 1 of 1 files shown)


bigquery_etl/stripe/__init__.py
```diff
@@ -14,6 +14,22 @@
 import click
 import stripe
 
+# omit lists of resources that have a separate table
```
acmiyaguchi, line 17: The events for these resources are captured in those tables?

acmiyaguchi (approved):

bigquery_etl/stripe/__init__.py
relud: yeah, more or less. the data is still captured, this is just reducing how much we repeat it.

PR #1396, opened 2020-10-06
Bug 1635918 Enable deletion of data from AET tables on user request

jklukas:


bigquery_etl/shredder/config.py
```diff
@@ -93,6 +95,14 @@ def fields(self) -> Tuple[str, ...]:
...
     field="_",
 )
+ECOSYSTEM_CLIENT_ID_HMAC_SRC = DeleteSource(
+    table="account_ecosystem_restricted.ecosystem_client_id_deletion_v1",
+    field=ECOSYSTEM_CLIENT_ID_HASH,
+)
+ECOSYSTEM_CLIENT_ID_SRC = DeleteSource(
+    table="account_ecosystem_restricted.ecosystem_client_id_deletion_v1",
+    field=ECOSYSTEM_CLIENT_ID,
+)
```
relud, line 105: these should also be added to `SOURCES`

relud (approved):
r+wc

PR #1397, opened 2020-10-06
Bug 1667891 - Export mozilla vpn waitlist from cloud SQL

relud:


sql/moz-fx-data-shared-prod/mozilla_vpn_derived/waitlist_v1/query.sql
```diff
@@ -0,0 +1,28 @@
+-- EXECUTE IMMEDIATE is necessary to use @date in the query for EXTERNAL_QUERY,
+-- because it requires a string literal
+EXECUTE IMMEDIATE CONCAT(
```
relud, line 3: it's ugly, but also it works? I'm open to suggestions.

jklukas:
```
SELECT FORMAT("DATE '%t'", DATE '2020-01-01')
```

yields `DATE '2020-01-01'`. 

So I would expect we could do:

```

    EXTERNAL_QUERY(
      "moz-fx-guardian-prod-bfc7.us.guardian-sql-prod",
      FORMAT("SELECT * FROM vpn_waitlist WHERE DATE(updated_at) = DATE '%t'", @date))
```

I can't do a full test because I don't have access to the db connection.

relud: EXTERNAL_QUERY requires a string literal, it does not allow the query to be generated at runtime:

> Unexpected TVF argument type found, expecting literal string at [17:7]

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/waitlist_v1/query.sql
```diff
@@ -0,0 +1,28 @@
...
+      _update
+    ).*
+  FROM
+    EXTERNAL_QUERY(
+      "moz-fx-guardian-prod-bfc7.us.guardian-sql-prod",
+      "SELECT * FROM vpn_waitlist WHERE DATE(updated_at) = DATE '""",
+  FORMAT_DATE("%F", @date),
+  """';"
+    ) AS _update
+  FULL JOIN
```
jklukas, line 24: Why express this with this join against the existing table rather atomically replacing just the partition for the target date?

It seems most flexible to do a daily extract of the previous day into BQ, and then worry about determining the current state per ID in a separate step, whether materialized or a view.

relud: I think you are trying to think of this as a stream of updates, when it isn't? I am trying to mirror a cloud sql table. Instead of reading the whole table from cloud sql every day, this query aims to reduce impact on the prod cloud sql db by reading only the rows that changed.

I don't understand what benefit there is to materializing the result of the `EXTERNAL_QUERY` as a separate step. The result is only useful for updating rows to match the cloud sql table, and it could result in people querying the wrong table.

jklukas:
It would help to have a `metadata.yaml` here that includes a robust description of what the guarantees are for this table and how to use it. Also, do we want to have this query scheduled via metadata.yaml?

I was imagining that we'd query for `DATE(updated_at) = @date`, so each partition would reflect rows that were updated on the given day. If the use case here is that we only ever care about approximating the _current_ state of the waitlist table in PG, then I can see how the approach here makes sense.

relud:
> It would help to have a `metadata.yaml` here that includes a robust description of what the guarantees are for this table and how to use it. Also, do we want to have this query scheduled via metadata.yaml?

I will add one, it will include scheduling.

> If the use case here is that we only ever care about approximating the `current` state of the waitlist table in PG, then I can see how the approach here makes sense.

current state in PG, updated daily, is the goal here.

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/waitlist_v1/query.sql
relud: I found out that it accepts a string literal *or query parameter*, so moving the whole external_database_query value to a string query parameter fixed this issue.

relud:
Added table description and a derived table for redash access, and modified the `EXTERNAL_QUERY` call to use a query parameter so `EXECUTE IMMEDIATE` is not longer necessary.

jklukas (approved):

Issue #1391, opened 2020-10-06
`bqetl udf create` should not default to including dataset in udf friendly name


fbertsch:
Currently it creates something like: `friendly_name: <dataset> <UDF name>`. We should drop the dataset.

Issue #1342, opened 2020-09-24
bqetl udf create with non-udf dataset results in untestable udf


fbertsch:
The default configuration for `bq udf create`, when used with a dataset that is not `udf`, is to create a new subdirectory within `/udf` with that dataset name. Unfortunately this causes any udfs in that directory to fail [here](https://github.com/mozilla/bigquery-etl/blob/master/bigquery_etl/udf/parse_udf.py#L135), because the udf name's dataset is presumed to be `udf`.

We need to treat these [like `mozfun` udfs](https://github.com/mozilla/bigquery-etl/blob/master/bigquery_etl/udf/parse_udf.py#L125).

scholtzan:
Fixed via https://github.com/mozilla/bigquery-etl/pull/1399

PR #1399, opened 2020-10-06
Routines

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/1391 and https://github.com/mozilla/bigquery-etl/issues/1342

This adds support for adding UDFs to generic datasets, and not only `udf` and `udf_js`. This way a lot of the logic already in place for mozfun can be reused.

Part of this PR is also renaming functions that process UDFs and stored procedures to `_routines_`.


bigquery_etl/cli/udf.py
```diff
@@ -122,7 +122,7 @@ def create(ctx, name, path):
     # create defaul metadata.yaml
     metadata_file = udf_path / "metadata.yaml"
     metadata = {
-        "friendly_name": string.capwords(dataset + " " + name.replace("_", " ")),
+        "friendly_name": string.capwords(name.replace("_", " ")),
```
scholtzan, line 125: https://github.com/mozilla/bigquery-etl/issues/1391

bigquery_etl/routine/parse_routine.py
```diff
@@ -144,186 +158,162 @@ def from_text(text, dataset, name, filepath=None, description="", is_defined=Tru
...
 
-def parse_udf_dirs(*udf_dirs):
-    """Read contents of udf_dirs into ParsedUdf instances."""
+def parse_routines(project_dir):
+    """Read routine contents of the project dir into ParsedRoutine instances."""
     # collect udfs to parse
-    raw_udfs = read_udf_dirs(*udf_dirs)
+    raw_routines = read_routine_dir(project_dir)
+    raw_routines.update(read_routine_dir(os.path.join(SQL_DIR, "mozfun")))
+    raw_routines.update(read_routine_dir(ASSERT_UDF_DIR))
```
scholtzan, line 217: These are all UDFs that are potentially referenced in tests

bigquery_etl/pytest_plugin/udf.py
```diff
@@ -6,32 +6,30 @@
...
     GENERIC_DATASET,
 )
 
-_parsed_udfs = None
+_parsed_routines = None
 
 
-def parsed_udfs():
+def parsed_routines():
     """Get cached parsed udfs."""
```
fbertsch, line 23:
```suggestion
    """Get cached parsed routines."""
```

bigquery_etl/pytest_plugin/udf.py
```diff
@@ -52,7 +50,7 @@ class UdfFile(pytest.File):
     def collect(self):
         """Collect."""
         self.add_marker("udf")
-        self.udf = parsed_udfs()[self.name]
+        self.udf = parsed_routines()[self.name]
```
fbertsch, line 53: I think that this general change in `s/udf/routine` should update this file as well, since we are relying on it for procedure testing.

bigquery_etl/routine/parse_routine.py
```diff
@@ -14,40 +14,49 @@
 import yaml
 
 from bigquery_etl.metadata.parse_metadata import METADATA_FILE
-from bigquery_etl.util.common import project_dirs
 
 
-UDF_DIRS = ("udf", "udf_js")
-MOZFUN_DIR = ("sql/mozfun",)
 UDF_CHAR = "[a-zA-z0-9_]"
 UDF_FILE = "udf.sql"
```
fbertsch, line 20: `s/(UDF_FILE, PROCEDURE_FILE)/ROUTINE_FILE` - they are used together in almost every instance now, no?

bigquery_etl/routine/parse_routine.py
```diff
@@ -14,40 +14,49 @@
...
+        for filename in files
+        if filename in (UDF_FILE, PROCEDURE_FILE)
+    ]
+
+
+def get_routines(project):
+    """Return all routines that could be referenced by the project."""
+    return (
+        get_routines_from_dir(project)
+        + get_routines_from_dir(os.path.join(SQL_DIR, "mozfun"))
```
fbertsch, line 52: We should really be using pathlib for these directories

bigquery_etl/routine/parse_routine.py
```diff
@@ -144,186 +158,162 @@ def from_text(text, dataset, name, filepath=None, description="", is_defined=Tru
...
 
-        return RawUdf(
+        return RawRoutine(
             internal_name,
             dataset,
             filepath,
             definitions,
             tests,
             # We convert the list to a set to deduplicate entries,
             # but then convert back to a list for stable order.
```
fbertsch, line 181:
```suggestion
```

bigquery_etl/routine/parse_routine.py
```diff
@@ -144,186 +158,162 @@ def from_text(text, dataset, name, filepath=None, description="", is_defined=Tru
...
-        for udf in MOZFUN_ROUTINES:
+        # get routines that could be referenced by the UDF
+        routines = get_routines(project)
+        dependencies = []
+        for udf in routines:
             if udf["name"] in "\n".join(definitions):
                 dependencies.append(udf["name"])
 
+        dependencies.extend(re.findall(TEMP_UDF_RE, "\n".join(definitions)))
+        dependencies = list(set(dependencies))
```
fbertsch, line 169: I don't think deduping is necessary anymore, since we're iterating through the list of routines (which should have no duplicates in itself).

bigquery_etl/routine/parse_routine.py
fbertsch: It's looking like we should allow a list of routine dirs: `read_routine_dirs(*project_dirs)`

bigquery_etl/routine/parse_routine.py
```diff
@@ -144,186 +158,162 @@ def from_text(text, dataset, name, filepath=None, description="", is_defined=Tru
...
+def routine_usages_in_text(text, project):
+    """Return a list of routine names used in the provided SQL text."""
+    sql = sqlparse.format(text, strip_comments=True)
+    routines = get_routines(project)
 
+    udf_usages = []
+
+    for routine in routines:
+        if routine["name"] in sql:
+            udf_usages.append(routine["name"])
```
fbertsch, line 253: This very much matches what we're doing up above, in the `RawRoutine.from_text` function. I'm wondering if we can instead use `RawRoutine.from_text(defined=False)` for these tests, like we do in the [hmac tests](https://github.com/mozilla/bigquery-etl/blob/master/tests/validation/hmac_sha256.py#L40).

bigquery_etl/routine/publish_routines.py
```diff
@@ -9,17 +9,16 @@
...
+from bigquery_etl.routine.parse_routine import (
+    read_routine_dir,
     accumulate_dependencies,
-    get_udf_dirs,
 )
 
-DEFAULT_PROJECT_ID = "moz-fx-data-shared-prod"
-DEFAULT_UDF_DIR = ["udf/", "udf_js/"]
-DEFAULT_DEPENDENCY_DIR = "udf_js/lib/"
+DEFAULT_DEPENDENCY_DIR = "lib/"
```
fbertsch, line 18: A more descriptive name for this directory might be in order, since we lost the parent (`udf_js`) to tell us what the library is for.

fbertsch (changes requested):
Thanks for working on cleaning up this codebase. I've put in a few ideas for next steps to make things clearer.

bigquery_etl/routine/parse_routine.py
fbertsch: (Note: I don't think we should block this PR from landing on this change, since the problem existed before it).

bigquery_etl/routine/parse_routine.py
fbertsch: (Note: I don't think we should block this PR from landing on this change, since the problem existed before it).

bigquery_etl/cli/udf.py
```diff
@@ -21,7 +21,7 @@
 UDF_NAME_RE = re.compile(r"^(?P<dataset>[a-zA-z0-9_]+)\.(?P<name>[a-zA-z0-9_]+)$")
 UDF_DATASET_RE = re.compile(r"^(?P<dataset>[a-zA-z0-9_]+)$")
 UDF_FILE_RE = re.compile(r"(^.*/|^)([a-zA-Z0-9_]+)/([a-zA-Z0-9_]+)/udf\.sql$")
-DEFAULT_DEPENDENCY_DIR = "udf_js/lib/"
+DEFAULT_UDF_DEPENDENCY_DIR = "lib/"
```
fbertsch, line 24: Just to be clear, I meant changing `lib/` to something like `udf_js_lib/` :)

scholtzan:
Just checked, the changes here will also fix https://github.com/mozilla/bigquery-etl/issues/1400

bigquery_etl/routine/parse_routine.py
scholtzan: It's still necessary for when routines in `mozfun` are parsed. `mozfun` routines are by default part of `get_routines` since they can be referenced by any routine in any project. So if `mozfun` is passed in as project routines should be parsed from then these UDFs will appear twice.

fbertsch (approved):
🚢

PR #1401, opened 2020-10-07
Create UDF for `extract_event_counts_with_properties`

fbertsch:
This is a follow-up to `extract_event_counts`. This is to support dashboards that need a breakdown of events w/ properties.

### DIFF
--- a/sql/mozfun/event_analysis/extract_event_counts_with_properties/udf.sql
+++ b/sql/mozfun/event_analysis/extract_event_counts_with_properties/udf.sql
@@ -77,7 +77,6 @@ SELECT
     with_event_props[OFFSET(0)]
   ),
   assert_struct_equals(
-  assert_struct_equals(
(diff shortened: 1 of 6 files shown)


sql/mozfun/event_analysis/extract_event_counts_with_properties/udf.sql
```diff
@@ -0,0 +1,140 @@
...
+  ),
+  assert_struct_equals(
+    STRUCT(
+      'a' AS event_index,
+      NULL AS property_index,
+      CAST(NULL AS STRING) AS property_value_index,
+      3 AS count
+    ),
+    with_event_props[OFFSET(0)]
+  ),
```
jklukas, line 87:
```suggestion
```

Repeat of previous test

jklukas (approved):
r+

PR #1407, opened 2020-10-08
Add stored procedure for enumerating geckoview versions in fenix nightly

acmiyaguchi:
See #1402 

This was tested in the `glam-fenix-dev` project using the following SQL:

<details>
<summary>Click to expand</summary>

```sql
DECLARE
  results ARRAY<STRUCT<build_hour DATETIME,
  geckoview_version STRING,
  n_builds INT64>>;
CREATE OR REPLACE PROCEDURE
  analysis.fenix_nightly_geckoview_versions(
    IN reference_date DATE,
    IN history_days INT64,
    OUT build_hour_version ARRAY<
      STRUCT<build_hour DATETIME, geckoview_version STRING, n_builds INT64>
    >
  )
BEGIN
  -- https://github.com/mozilla/bigquery-etl/blob/56defb1061445115277d90def06ad41185e519b5/sql/moz-fx-data-shared-prod/udf/fenix_build_to_datetime/metadata.yaml#L1-L23
  SET build_hour_version = (
    WITH extracted AS (
    -- We'll look at the metrics ping to estimate the major geckoview version
    -- the metrics section is aliased, so we must rename the table for this to
    -- work appropriately
      SELECT
        submission_timestamp,
        client_info.app_build,
        metrics.string.geckoview_version,
      FROM
        `moz-fx-data-shared-prod`.org_mozilla_fenix.metrics AS t1
      WHERE
        mozfun.norm.fenix_app_info('org_mozilla_fenix', client_info.app_build).channel = 'nightly'
      UNION ALL
      SELECT
        submission_timestamp,
        client_info.app_build,
        metrics.string.geckoview_version,
      FROM
        `moz-fx-data-shared-prod`.org_mozilla_fenix_nightly.metrics AS t1
      UNION ALL
      SELECT
        submission_timestamp,
        client_info.app_build,
        metrics.string.geckoview_version,
      FROM
        `moz-fx-data-shared-prod`.org_mozilla_fennec_aurora.metrics AS t1
    ),
    transformed AS (
      SELECT
        app_build,
        geckoview_version,
        datetime_TRUNC(
          `moz-fx-data-shared-prod`.udf.fenix_build_to_datetime(app_build),
          HOUR
        ) AS build_hour
      FROM
        extracted
      WHERE
        DATE(submission_timestamp) >= DATE_SUB(reference_date, INTERVAL history_days DAY)
    ),
    grouped_build_hours AS (-- count the number of geckoview versions for each build hour row over an interval
      SELECT
[...]

PR #1415, opened 2020-10-09
Allow for cross-project references in routines

fbertsch:
- Scrape all projects for routine defns when generating tests
- Create UDFs as non-temp for stored procedure tests
- Make assert functions default non-temp (to support above)
- Memoize raw routine scraping

Fixes https://github.com/mozilla/bigquery-etl/pull/1407.


bigquery_etl/routine/parse_routine.py
```diff
@@ -272,25 +289,33 @@ def routine_usage_definitions(text, project, raw_routines=None):
...
+    for name, routine in raw_routines.items():
+        if name in sql:
+            for defn in routine.definitions:
+                match = PERSISTENT_UDF_RE.match(defn)
+                dataset, name = match.group(1), match.group(2)
+                replace_name = f"{dataset}_{name}"
+                if stored_procedure_test:
+                    replace_name = f"{GENERIC_DATASET}.{replace_name}"
+                sql = re.sub(
+                    fr"`?(?:`?{routine.project}`?\.)?{dataset}.{name}`?",
```
acmiyaguchi, line 312: What is this doing, with all of the `?`? Conditionally setting the project if it exists?

tests/assert/array_equals/udf.sql
```diff
@@ -1,4 +1,4 @@
-CREATE TEMP FUNCTION assert_array_equals(expected ANY TYPE, actual ANY TYPE) AS (
+CREATE OR REPLACE FUNCTION assert.array_equals(expected ANY TYPE, actual ANY TYPE) AS (
```
acmiyaguchi, line 1: What's the reasoning behind putting this in it's own dataset instead of leaving it as a temporary function? Is this a functional requirement for stored procedure tests?

bigquery_etl/routine/parse_routine.py
```diff
@@ -16,20 +16,27 @@
...
+PERSISTENT_UDF_PREFIX = re.compile(PERSISTENT_UDF_PREFIX_RE_STR, re.IGNORECASE)
+PERSISTENT_UDF_RE = re.compile(
+    fr"{PERSISTENT_UDF_PREFIX_RE_STR}\s+({UDF_CHAR}*)\.({UDF_CHAR}+)`?", re.IGNORECASE
 )
-PERSISTENT_UDF_RE = re.compile(fr"((?:udf|assert){UDF_CHAR}*)\.({UDF_CHAR}+)")
 UDF_NAME_RE = re.compile(r"^([a-zA-Z0-9_]+\.)?[a-zA-Z][a-zA-Z0-9_]{0,255}$")
 GENERIC_DATASET = "_generic_dataset_"
 SQL_DIR = Path("sql/")
 ASSERT_UDF_DIR = "tests/assert"
+TEST_UDF_DIR = "tests/data/test_sql/moz-fx-data-test-project/udf"
```
scholtzan, line 37: Looks like this is unused here?

```suggestion
```

bigquery_etl/routine/parse_routine.py
```diff
@@ -16,20 +16,27 @@
 from bigquery_etl.metadata.parse_metadata import METADATA_FILE
 
 
+PROJECT_CHAR = "[a-zA-z0-9-]"
```
scholtzan, line 19: Looks like this is not used

```suggestion
```

scholtzan (approved):

bigquery_etl/routine/parse_routine.py
fbertsch: right - may or may not have backticks, may or may not have backticks around the project name, the project name may or may not exist.

tests/assert/array_equals/udf.sql
fbertsch: Good question. This came up because you cannot define a temporary udf and a permanent udf in the same query. So while we had to define the UDFs inside of the procedure as permanent, the assert udfs were all still temporary, causing it to fail. I figured it's easiest to rely on the existing infrastructure to transform the asserts to temporary only when they need to be.

acmiyaguchi: It's strange that you can't have temporary and permanent udfs in the same query. It seems reasonable to have these utilities in persistent locations, though.

acmiyaguchi (approved):

Issue #1403, opened 2020-10-07
Normalize fenix build ids to the build timestamp to ignore encoded architecture/device information in GLAM


acmiyaguchi:
The fenix build id includes extra information in the build id that covers things like the architecture (ARM/x86). This means there are ~4 builds per build time. Instead, we'd like to aggregate across a single build timestamp. To solve this, the Fenix builds should simply be the date timestamp truncated to the hour as a string.

PR #1418, opened 2020-10-09
Add udfs for creating and parsing custom build ids

acmiyaguchi:
This are utility functions for https://github.com/mozilla/bigquery-etl/issues/1403

The `format_build_hour` udf is intended to be used in the `logical_app_id` views to convert the `app_build_id` into a custom id that throws away architecture information. The `parse_build_hour` udf is meant to be used in the `generate.py` file as the `build_date_udf`.


sql/mozfun/glam/parse_build_hour/metadata.yaml
```diff
@@ -0,0 +1,3 @@
+description: |
+  Parses the custom build id used for Fenix builds in GLAM.
```
Ben-Wu, line 2:
```suggestion
  Parses the custom build id used for Fenix builds in GLAM to a datetime.
```

sql/mozfun/glam/parse_build_hour/udf.sql
```diff
@@ -0,0 +1,9 @@
+CREATE OR REPLACE FUNCTION glam.parse_build_hour(build_hour STRING)
```
Ben-Wu, line 1: Something like `build_hour_to_datetime` would be more descriptive

sql/mozfun/glam/format_build_hour/udf.sql
```diff
@@ -0,0 +1,12 @@
+CREATE OR REPLACE FUNCTION glam.format_build_hour(app_build_id STRING)
```
Ben-Wu, line 1:
```suggestion
CREATE OR REPLACE FUNCTION glam.fenix_build_to_build_hour(app_build_id STRING)
```

Ben-Wu (changes requested):
I think the function names are a little confusing.

Also should we change `udf.fenix_build_to_datetime` to call the mozfun version?

acmiyaguchi:
Thanks, these are certainly better names. We'll need to wait until the next deploy before we can reference the mozfun function in the udf module, but yes, we should.

sql/mozfun/glam/parse_build_hour/udf.sql
Ben-Wu, line 1:
```suggestion
CREATE OR REPLACE FUNCTION glam.build_hour_to_datetime(build_hour STRING)
```

Ben-Wu (approved):
Just the other renaming but otherwise good

Issue #1402, opened 2020-10-07
Use metrics.string.geckoview_version for fenix nightly aggregates in GLAM


acmiyaguchi:
This is an extension of #1329, where it's been determined that using the build date as the app_version is not sufficient for tracking changes over time. The app_version can be inferred from the metrics ping, however.

acmiyaguchi:
The basis of this will be a derived view that's referenced in the nightly job. Each of the builds will be truncated to the hour, with each hour associated with the most recent geckoview version.

![image](https://user-images.githubusercontent.com/3304040/95401576-6bc9d880-08c2-11eb-831e-ec3bf868903b.png)


https://sql.telemetry.mozilla.org/queries/75387/source#188214

PR #1419, opened 2020-10-09
Add table to determine Fenix nightly mapping of builds to geckoview versions

acmiyaguchi:
This query addresses #1402 by creating an incremental `org_mozilla_fenix_derived.geckoview_version_v1` table. 

Here's the init query run on STMO: https://sql.telemetry.mozilla.org/queries/75590/source#188662

It fails dryrun because the table hasn't been initialized. Once it is deployed, the dry-runs should start to work again.


tests/sql/moz-fx-data-shared-prod/org_mozilla_fenix_nightly_derived/geckoview_version_v1/metrics.schema.json
```diff
@@ -0,0 +1,39 @@
...
+            "type:": "STRING",
+            "name": "geckoview_version",
+            "mode": "NULLABLE"
+          }
+        ]
+      }
+    ]
+  }
+]
+
```
acmiyaguchi, line 39: I've been trying to set up tests with the schema to cast the relevant fields to string, but I get the following errors:

```python
>>> from google.cloud import bigquery

# works fine
>>> bq.schema_from_json("tests/sql/moz-fx-data-shared-prod/search_derived/mobile_search_aggregates_v1/test_aggregation/mobile_search_clients_daily_v1.schema.json")

>>> bq.schema_from_json("tests/sql/moz-fx-data-shared-prod/org_mozilla_fenix_nightly_derived/geckoview_version_v1/metrics.schema.json")

  File "<stdin>", line 1, in <module>
  File "/Users/amiyaguchi/Work/bigquery-etl/venv/lib/python3.8/site-packages/google/cloud/bigquery/client.py", line 3166, in schema_from_json
    return self._schema_from_json_file_object(file_obj)
  File "/Users/amiyaguchi/Work/bigquery-etl/venv/lib/python3.8/site-packages/google/cloud/bigquery/client.py", line 3147, in _schema_from_json_file_object
    return [SchemaField.from_api_repr(field) for field in json_data]
  File "/Users/amiyaguchi/Work/bigquery-etl/venv/lib/python3.8/site-packages/google/cloud/bigquery/client.py", line 3147, in <listcomp>
    return [SchemaField.from_api_repr(field) for field in json_data]
  File "/Users/amiyaguchi/Work/bigquery-etl/venv/lib/python3.8/site-packages/google/cloud/bigquery/schema.py", line 105, in from_api_repr
    fields=[cls.from_api_repr(f) for f in fields],
  File "/Users/amiyaguchi/Work/bigquery-etl/venv/lib/python3.8/site-packages/google/cloud/bigquery/schema.py", line 105, in <listcomp>
    fields=[cls.from_api_repr(f) for f in fields],
  File "/Users/amiyaguchi/Work/bigquery-etl/venv/lib/python3.8/site-packages/google/cloud/bigquery/schema.py", line 104, in from_api_repr
    field_type=api_repr["type"].upper(),
KeyError: 'type'
```

@relud would you happen to know where I've gone wrong with this schema? I tried this out in the BigQuery console, and it seems to create a new table fine.

tests/sql/moz-fx-data-shared-prod/org_mozilla_fenix_nightly_derived/geckoview_version_v1/metrics.schema.json
```diff
@@ -0,0 +1,39 @@
...
+    "name": "metrics",
+    "mode": "NULLABLE",
+    "fields": [
+      {
+        "type": "RECORD",
+        "name": "string",
+        "mode": "NULLABLE",
+        "fields": [
+          {
+            "type:": "STRING",
```
relud, line 30: there's a misplaced colon in the type key
```suggestion
        "type": "STRING",
        "name": "app_build",
        "mode": "NULLABLE"
      }
    ]
  },
  {
    "type": "RECORD",
    "name": "metrics",
    "mode": "NULLABLE",
    "fields": [
      {
        "type": "RECORD",
        "name": "string",
        "mode": "NULLABLE",
        "fields": [
          {
            "type": "STRING",
```

acmiyaguchi: I definitely would not have caught that on my own, thanks!

sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/geckoview_version_v1/query.sql
```diff
@@ -0,0 +1,161 @@
...
+  FROM
+    estimated_version
+  WHERE
+    build_hour
+    BETWEEN DATE_SUB(@submission_date, INTERVAL 30 DAY)
+    AND @submission_date
+  ORDER BY
+    build_hour
+),
+past_and_present AS (
```
Ben-Wu, line 137: I'm not understanding what this step does.  Why do we need to recalculate the past values instead of just appending the build hours for the current date?

sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/geckoview_version_v1/init.sql
```diff
@@ -0,0 +1,147 @@
...
+    datetime_TRUNC(mozfun.norm.fenix_build_to_datetime(app_build), HOUR) AS build_hour
+  FROM
+    extracted
+  WHERE
+    DATE(submission_timestamp)
+    BETWEEN DATE_SUB(@submission_date, INTERVAL 30 DAY)
+    AND @submission_date
+),
+grouped_build_hours AS (
+  -- Count the number of geckoview versions for each build hour row over the
```
Ben-Wu, line 49: Doesn't this count the number of pings per build_hour and geckoview_version? Not geckoview versions for each build hour.  If it is n_builds would be the wrong name.  But I think counting number of pings sounds reasonable.

sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/geckoview_version_v1/init.sql
```diff
@@ -0,0 +1,147 @@
...
+  HAVING
+    n_builds > 5
+  ORDER BY
+    build_hour DESC,
+    geckoview_version
+),
+top_build_hours AS (
+  -- Get the geckoview version for the build hour that has the most number
+  -- of rows.
+  SELECT
```
Ben-Wu, line 74: Subquery is a little clunky but I don't know of a better way.  

Maybe something like
```sql
SELECT
  *,
  ROW_NUMBER() OVER (PARTITION BY build_hour ORDER BY n_builds DESC) AS r,
FROM
  grouped_build_hours t
ORDER BY
  build_hour
```
and filtering by `r=1` in the next query avoids the subquery but I'm not convinced that's any better

Ben-Wu (commented):
I have some questions.  Mainly, why do we need to recursively compute the table instead of just adding the build hours of the latest date?

acmiyaguchi (commented):
It doesn't need to be incremental (and probably better that it isn't). I would like to be able to see the entire window over for each submission date so I can see the number of pings for a nightly build, days after it's been pushed out. I'd like to catch discontinuities between windows.

 I'll modify this so each date just appends the window to the table, and add a view that limits the output to one row per build hour.

tests/sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/geckoview_version_v1/test_aggregation/org_mozilla_fennec_aurora.metrics.yaml
```diff
@@ -0,0 +1,6 @@
+- client_info:
```
Ben-Wu, line 1: Can you add a test case for when there are more than 5 pings each for geckoversion negative, between 0 and 100, and >100 at the same timestamp?  My concern is that geckoview version is a string so doing MAX will put 99.0.0 > 100.0.0

Ben-Wu (changes requested):
Thanks for the clarification, I also somehow missed the very first comment in the query.  I think not having almost identical init and query now is much clearer.  It wouldn't hurt to still have an init.sql just to define the columns like a lot of queries do but I'm indifferent on that.

I'm still wondering about version >100.0.0 because it's a string unless I'm misunderstanding that step.

tests/sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/geckoview_version_v1/test_aggregation/org_mozilla_fennec_aurora.metrics.yaml
acmiyaguchi: Yep, I'm working through fixing the tests since they broken in that last change I made.

Ben-Wu (approved):
I'll assume major version is sufficient because that's what desktop uses.

PR #1420, opened 2020-10-09
Query for importing experimenter experiments

scholtzan:
Scheduled via https://github.com/mozilla/telemetry-airflow/pull/1167

### DIFF
--- a/sql/moz-fx-data-experiments/monitoring/experimenter_experiments_v1/query.py
+++ b/sql/moz-fx-data-experiments/monitoring/experimenter_experiments_v1/query.py
@@ -1,0 +1,211 @@
+#!/usr/bin/env python3
+
+"""Import experiments from Experimenter via the Experimenter API."""
+
+from argparse import ArgumentParser
+from google.cloud import bigquery
+import requests
+import attr
+import cattr
+import datetime
+import pytz
+from typing import List, Optional
+
+EXPERIMENTER_API_URL_V1 = (
+    "https://experimenter.services.mozilla.com/api/v1/experiments/"
+)
+
+# for nimbus experiments
+EXPERIMENTER_API_URL_V4 = (
+    "https://experimenter.services.mozilla.com/api/v4/experiments/"
+)
+
+parser = ArgumentParser(description=__doc__)
+parser.add_argument("--project", default="moz-fx-data-experiments")
+parser.add_argument("--destination_dataset", default="monitoring")
+parser.add_argument("--destination_table", default="experimenter_experiments_v1")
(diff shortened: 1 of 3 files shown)


sql/moz-fx-data-experiments/monitoring/experimenter_experiments_v1/query.py
```diff
@@ -0,0 +1,211 @@
...
+            normandy_slug=self.slug,
+            experimenter_slug=None,
+            type="v4",
+            status="Live" if self.active else "Complete",
+            start_date=self.startDate,
+            end_date=self.endDate,
+            proposed_enrollment=self.proposedEnrollment,
+            reference_branch=self.referenceBranch,
+            is_high_population=False,
+        )
```
scholtzan, line 140: `Experiment`, `ExperimentV4` and `ExperimentV1` is pretty much copied from jetstream. I removed branches and variants since they do not seem necessary.

sql/moz-fx-data-experiments/monitoring/experimenter_experiments_v1/query.py
```diff
@@ -0,0 +1,211 @@
...
+
+    converter = cattr.Converter()
+    converter.register_unstructure_hook(
+        datetime.datetime, lambda d: datetime.datetime.strftime(d, format="%Y-%m-%d")
+    )
+
+    client.load_table_from_json(
+        converter.unstructure(experiments), destination_table, job_config
+    ).result()
+    print(f"Loaded {len(experiments)} experiments")
```
scholtzan, line 207: We'll just dump all the experiments to the destination table

sql/moz-fx-data-experiments/monitoring/experimenter_experiments_v1/query.py
```diff
@@ -0,0 +1,211 @@
...
+            status="Live" if self.active else "Complete",
+            start_date=self.startDate,
+            end_date=self.endDate,
+            proposed_enrollment=self.proposedEnrollment,
+            reference_branch=self.referenceBranch,
+            is_high_population=False,
+        )
+
+
+def get_experiments():
```
tdsmith, line 143:
```suggestion
def get_experiments() -> List[Experiment]:
```

may as well annotate if we're gonna annotate :)

tdsmith (approved):
looks right! the branch ratios will be important if we want to add checks for branch balance, so we might want to keep a [branch, ratio] array on the loaded experiment.

Not sure we need to know the reference_branch, otoh.

scholtzan:
Good point. Added branches and ratios

Issue #1422, opened 2020-10-12
publish_views: google.api_core.exceptions.BadRequest: 400 Syntax error: Unclosed identifier literal


whd:
/CC @scholtzan 

Guessing this has something to do with replacement logic, because the view [looks fine](https://github.com/mozilla/bigquery-etl/blob/master/sql/moz-fx-data-shared-prod/mozilla_vpn/waitlist/view.sql#L2). I vaguely recall this happening before but for an unrelated reason (#1232).

```
   1:CREATE OR REPLACE VIEW
   2:  `moz-fx-data-shared-prod.mozilla_vpn.waitlist
   3:AS
   4:SELECT
   5:  *
   6:FROM
   7:  `moz-fx-data-shared-prod`.mozilla_vpn_derived.waitlist_v1
```

scholtzan:
The backticks need to enclose the entire view name, not just the project part. I pushed a fix in https://github.com/mozilla/bigquery-etl/pull/1423

relud:
@whd can you check that this fixed it?

PR #1425, opened 2020-10-12
Fix backticks in stripe derived views

relud:
additional fixes for #1422


fbertsch (approved):
This is fine, but should we fix the publishing script instead? We deal with backticks in the routine parsing. https://github.com/mozilla/bigquery-etl/blob/master/bigquery_etl/routine/parse_routine.py#L310

relud:
> This is fine, but should we fix the publishing script instead? We deal with backticks in the routine parsing. https://github.com/mozilla/bigquery-etl/blob/master/bigquery_etl/routine/parse_routine.py#L310

filed #1426 to do that, but for now I'll merge this as it's a much simpler fix

PR #1428, opened 2020-10-13
Fix field name error causing Shredder to miss FxA account deletions

jklukas:
While looking at Shredder logs, I noticed that all entries for FxA-related
derived tables show 0 deletions, like:

> 712215424909 bytes and deleted 0 rows from moz-fx-data-shared-prod.firefox_accounts_derived.fxa_users_daily_v1

It appears that `account.deleted` events populate a field named `uid` rather
than `user_id`. I was able to verify this by choosing a recent `uid` value
from a deletion event and counting events where that same value appears as
`user_id` in FxA logs. There were matching messages.


jklukas:
We will want to backfill this as far back as we have source logs (90 days).

relud (approved):

Issue #1421, opened 2020-10-09
Add support for multiple projects to publish tasks to the same DAG


scholtzan:
Different projects might have tasks that should be part of the same DAG. Currently, only tasks of the last project evaluated will overwrite the DAG.

PR #1429, opened 2020-10-13
Fix DAG generation for different projects

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/1421


jklukas (approved):

jklukas:
r+ but looks like there are some legitimate test failures

PR #1436, opened 2020-10-14
Add mozfun.norm.product_info UDF

jklukas:
Factoring this out from https://github.com/mozilla/bigquery-etl/pull/1380
since this UDF needs to be published before I can properly test the rest of the changes there.


scholtzan (approved):
Looks good to me

PR #1440, opened 2020-10-15
Bug 1654078 Limit geo to country level in regrets-reporter view

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1654078#c45


whd (approved):
LGTM assuming CI passes

PR #1444, opened 2020-10-16
Refactor stripe tables to reduce confusion

relud:
The current layout of stripe tables has an initial import of various resources, and then a table of events that update and add to those resources.

That had been causing some confusion, because people were expecting the `stripe_external.{resource}_v1` tables to be updated every day.

This PR depends on me moving the initial imports to `stripe_external.initial_{resource}_v1`, and then regularly updates `stripe_external.{resource}_v1` tables. Those are then copied or aggregated (well, not aggregated yet, but soon) into `stripe_derived`, with PII and individual payment information removed because that dataset is readable in Redash.


relud:
the dry-run is currently failing because the new table configuration conflicts with the current configuration. before merging i will rearrange the tables and re-run the dry-run to make sure it works

acmiyaguchi (approved):
Looks good to me

PR #1447, opened 2020-10-16
Add daily schedule for geckoview versions

acmiyaguchi:
This schedules the query in #1419 in support of #1402

### DIFF
--- a/dags/bqetl_org_mozilla_fenix_derived.py
+++ b/dags/bqetl_org_mozilla_fenix_derived.py
@@ -31,7 +31,6 @@ with DAG(
         email=["amiyaguchi@mozilla.com", "telemetry-alerts@mozilla.com"],
         date_partition_parameter="submission_date",
         depends_on_past=False,
-        parameters=["submission_date:DATE:{{ds}}"],
         dag=dag,
     )
 
--- a/dags/bqetl_org_mozilla_fenix_derived.py
+++ b/dags/bqetl_org_mozilla_fenix_derived.py
@@ -1,0 +1,50 @@
+# Generated via https://github.com/mozilla/bigquery-etl/blob/master/bigquery_etl/query_scheduling/generate_airflow_dags.py
+
+from airflow import DAG
+from airflow.operators.sensors import ExternalTaskSensor
+import datetime
+from utils.gcp import bigquery_etl_query
+
+default_args = {
+    "owner": "amiyaguchi@mozilla.com",
+    "start_date": datetime.datetime(2020, 10, 18, 0, 0),
+    "email": ["amiyaguchi@mozilla.com", "telemetry-alerts@mozilla.com"],
+    "depends_on_past": False,
+    "retry_delay": datetime.timedelta(seconds=1800),
+    "email_on_failure": True,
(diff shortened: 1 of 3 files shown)


dags/bqetl_org_mozilla_fenix_derived.py
```diff
@@ -0,0 +1,50 @@
...
+) as dag:
+
+    org_mozilla_fenix_derived__geckoview_version__v1 = bigquery_etl_query(
+        task_id="org_mozilla_fenix_derived__geckoview_version__v1",
+        destination_table="geckoview_version_v1",
+        dataset_id="org_mozilla_fenix_derived",
+        project_id="moz-fx-data-shared-prod",
+        owner="amiyaguchi@mozilla.com",
+        email=["amiyaguchi@mozilla.com", "telemetry-alerts@mozilla.com"],
+        date_partition_parameter="submission_date",
```
Ben-Wu, line 32: I don't think `date_partition_parameter` should be here since it isn't incremental.  Is `incremental: false` expected to take this out?

acmiyaguchi: `date_partition_parameter` is the suffix to the destination table e.g. `geckoview_version_v1$20201018`. The incremental keyword applies to `depends_on_past` in the airflow parameters.

Ben-Wu: Ok I thought this wasn't a partitioned table

sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/geckoview_version_v1/metadata.yaml
```diff
@@ -6,3 +6,9 @@ description: |
   reference date is 2020-10-01 will include hours up to midnight of that day.
 owners:
   - amiyaguchi@mozilla.com
+labels:
+  incremental: false
+  schedule: daily
+scheduling:
+  dag_name: bqetl_org_mozilla_fenix_derived
+  parameters: ["submission_date:DATE:{{ds}}"]
```
Ben-Wu, line 14: This parameter shouldn't be needed if `date_partition_parameter="submission_date"` since the operator wrapper function adds this already.  I don't know if it breaks if a parameter is defined multiple times.

Ben-Wu (approved):

dags/bqetl_org_mozilla_fenix_derived.py
Ben-Wu: I thought the reason you defined a submission_date parameter was because you didn't want `date_partition_parameter` defined

sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/geckoview_version_v1/metadata.yaml
acmiyaguchi: I see: https://github.com/mozilla/telemetry-airflow/blob/76e3b6bd2efaaa0c1b143bf531838174a2652c43/dags/utils/gcp.py#L493-L495

PR #1452, opened 2020-10-19
Add CREATE OR REPLACE VIEW clause to geckoview_versions and move it to org_mozilla_fenix

acmiyaguchi:
Not having a `CREATE OR REPLACE VIEW` DDL statement at the top of the view breaks validation when runnning `scripts/publish_view`.


Ben-Wu (commented):
Sorry I should have caught that.  Also, should this be in the `org_mozilla_fenix` dataset instead to match the pattern we use?

acmiyaguchi:
It looks like it should, I didn't realize derived was only for tables and not for views. I'll expand the scope a bit more.

Ben-Wu (approved):

PR #1455, opened 2020-10-19
Bug 1671517: Add event counts to clients daily

sunahsuh:
Per a data science request, adds an aggregation of `scalar_parent_telemetry_event_counts_sum` and `scalar_content_telemetry_event_counts_sum` to clients daily


sunahsuh:
Hmm looks like there were some changes to the permissions recently that's causing some of the tests to fail :/

jklukas:
> Hmm looks like there were some changes to the permissions recently that's causing some of the tests to fail :/

Yes, these failures look to be related to the table-level ACL changes @whd was planning to merge this week. I'll write up a data ops bug.

jklukas:
> I'll write up a data ops bug.

I thought about this more and realized it's likely not safe to give the dry run service these additional permissions, so this is all working as intended and I think we should instead exempt the affected queries from dry runs. I've requested your review on https://github.com/mozilla/bigquery-etl/pull/1462 which should allow tests here to succeed.

PR #1456, opened 2020-10-19
Fix #1402 - Use org_mozilla_fenix.geckoview_version view for app_version in GLAM

### LINKED ISSUE #1402: Use metrics.string.geckoview_version for fenix nightly aggregates in GLAM

acmiyaguchi:
Fixes #1402. Depends on #1458 due to the indiscriminate nature of the `replace_project_and_dataset` function in `run_glam_sql`. This means the reference to the geckoview_version view in shared-prod points to one in `glam-fenix-dev` which doesn't exist.

### DIFF
--- a/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__clients_histogram_aggregates_v1/query.sql
+++ b/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__clients_histogram_aggregates_v1/query.sql
@@ -72,7 +72,7 @@ filtered_accumulated AS (
       @submission_date,
       INTERVAL 365 day
     )
-    AND app_version >= (latest_version - 3)
+    AND app_version > (latest_version - 3)
 ),
 -- unnest the daily data
 extracted_daily AS (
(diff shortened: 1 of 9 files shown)


acmiyaguchi:
I tested this by dropping the `glam_etl_dev` dataset, running `./script/glam/test/test_glean_org_mozilla_fenix_glam_nightly`, and inspecting the output of the `org_mozilla_fenix_glam_nightly__extract*` tables. The outputs look reasonable, and the queries complete much faster than before due to proper versions.

sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__clients_histogram_aggregates_v1/query.sql
```diff
@@ -68,6 +72,7 @@ filtered_accumulated AS (
       @submission_date,
       INTERVAL 365 day
     )
+    AND app_version >= (latest_version - 3)
```
Ben-Wu, line 75: Doesn't this retain the last 4 versions and not 3 like the config indicates?

Ben-Wu (approved):

sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__clients_histogram_aggregates_v1/query.sql
acmiyaguchi: Yes, I believe you're right.

PR #1462, opened 2020-10-20
Exempt a few files from dry run due to new table-level ACLs

jklukas:
The dry run service can no longer perform queries with wildcard table specifications or access raw AET data. See mozilla-services/cloudops-infra#2599


sunahsuh (approved):
👍

PR #1464, opened 2020-10-20
Query for monitoring experiment analysis costs

scholtzan:
This is for adding some alerting based on the cost of queries executed by jetstream. I'll add alerts in Redash that will get triggered if the cost of a query exceeds a certain amount. Since Redash doesn't have access to `INFORMATION_SCHEMA.JOBS_BY_PROJECT` for `moz-fx-data-experiments`, we need this scheduled query which writes cost related data to a destination Redash has access to.


scholtzan:
@tdsmith I think this would still be good to have. Once we get this query running I can set up some alerting in Redash for when experiments exceed some cost threshold.

tdsmith (approved):

PR #1465, opened 2020-10-21
Add ETL for Mozilla VPN acquisition funnel

relud:
All of these tables feed into a redash query that combines them into a funnel: https://sql.telemetry.mozilla.org/queries/75572/source#188990

### DIFF
--- a/sql/moz-fx-data-shared-prod/mozilla_vpn_derived/login_flows_v1/init.sql
+++ b/sql/moz-fx-data-shared-prod/mozilla_vpn_derived/login_flows_v1/init.sql
@@ -1,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS
+  login_flows_v1(
+    flow_id STRING,
+    flow_started TIMESTAMP,
+    flow_completed TIMESTAMP,
+    fxa_uids ARRAY<STRING>,
+    viewed_email_first_page BOOL
+  )
--- a/sql/moz-fx-data-shared-prod/mozilla_vpn_derived/login_flows_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/mozilla_vpn_derived/login_flows_v1/metadata.yaml
@@ -1,0 +1,17 @@
+friendly_name: Mozilla VPN FxA Login Flows
+description: >
+  A list of users that attempted to login to FxA, to sign up for or login to Mozilla VPN.
+owners:
+  - dthorn@mozilla.com
+labels:
+  application: mozilla_vpn
+  schedule: daily
+scheduling:
+  dag_name: bqetl_mozilla_vpn
+  # destination is the whole table, not a single partition, so don't use date_partition_parameter
+  date_partition_parameter: null
+  parameters:
+  - "date:DATE:{{ds}}"
+  referenced_tables:
+  - ["moz-fx-data-shared-prod", "firefox_accounts_derived", "fxa_auth_events_v1"]
+  - ["moz-fx-data-shared-prod", "firefox_accounts_derived", "fxa_content_events_v1"]
--- a/sql/moz-fx-data-shared-prod/stripe_derived/customers_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/stripe_derived/customers_v1/query.sql
@@ -1,0 +1,7 @@
+SELECT
+  -- limit fields in stripe_derived so as not to expose sensitive data
+  created,
+  id,
+  metadata,
+FROM
+  stripe_external.customers_v1
--- a/bigquery_etl/dryrun.py
+++ b/bigquery_etl/dryrun.py
@@ -68,6 +68,7 @@ SKIP = {
     "sql/moz-fx-data-shared-prod/apple_app_store/metrics_total/query.sql",
     "sql/moz-fx-data-shared-prod/monitoring/telemetry_distinct_docids_v1/query.sql",
     "sql/moz-fx-data-shared-prod/revenue_derived/client_ltv_normalized/query.sql",
+    "sql/moz-fx-data-shared-prod/stripe_derived/customers_v1/query.sql",
     "sql/moz-fx-data-shared-prod/stripe_derived/plans_v1/query.sql",
     "sql/moz-fx-data-shared-prod/stripe_derived/products_v1/query.sql",
     "sql/moz-fx-data-shared-prod/stripe_derived/subscriptions_v1/query.sql",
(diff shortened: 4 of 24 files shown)


sql/moz-fx-data-shared-prod/mozilla_vpn_derived/acquisition_funnel_v1/query.sql
```diff
@@ -0,0 +1,107 @@
...
+),
+activated_vpn AS (
+  -- TODO: this is undercounted, and may not be fully measurable
+  SELECT
+    fxa_uid,
+    first_activated_vpn,
+  FROM
+    activated_vpn_v1
+)
+SELECT
```
felixlawrence, line 77: It would be useful for data exploration (and possibly more!) to store the table resulting from these joins, without the GROUP BY. For example, we could easily study the distribution over conversion times between arbitrary steps in the funnel.

Also, when doing so, would it make sense to replace the LEFT JOINs with OUTER JOINs, to deal with the fact that some users might be missing from some data sources but not others?

And if we really want to throw in the kitchen sink, could we pull in `referrer`, `utm_medium`, `utm_source`, `utm_campaign` by parsing the `attribution` column in `mozilla_vpn_external.users_v1` (or another data source if you've seen it)?

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/acquisition_funnel_v1/query.sql
```diff
@@ -0,0 +1,107 @@
...
+  GROUP BY
+    fxa_uid
+),
+activated_vpn AS (
+  -- TODO: this is undercounted, and may not be fully measurable
+  SELECT
+    fxa_uid,
+    first_activated_vpn,
+  FROM
+    activated_vpn_v1
```
felixlawrence, line 75: You didn't `GROUP BY fxa_uid` here - intentional because you know it's unique in `activated_vpn_1`, or an oversight?

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/acquisition_funnel_v1/query.sql
```diff
@@ -0,0 +1,107 @@
...
+    (fxa_uid)
+  WHERE
+    -- include entries in users_v1 that are not present in login_flows_v1
+    first_logged_in IS NULL
+    -- TODO: exclude login on desktop client, and second login attempt on mobile.
+    -- As an approximation, filter out flows that started after first login.
+    OR first_logged_in > flow_started
+  GROUP BY
+    fxa_uid,
+    -- when fxa_uid is missing, produce one row per flow_id
```
felixlawrence, line 28: This will undercount our conversion rates, since one person can be responsible for many `flow_id`s but IIUC at most one of those would be associated with any conversion? But such is life

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/acquisition_funnel_v1/query.sql
relud: because I know it's unique

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/acquisition_funnel_v1/query.sql
relud: > This will undercount our conversion rates, since one person can be responsible for many `flow_id`s but IIUC at most one of those would be associated with any conversion? But such is life

all of the flows that have a given fxa_uid are counted as a single conversion, yeah. I don't think we have anything longer lived than flow id that we could use to reduce that overcounting.

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/acquisition_funnel_v1/query.sql
relud: > It would be useful for data exploration (and possibly more!) to store the table resulting from these joins, without the GROUP BY. For example, we could easily study the distribution over conversion times between arbitrary steps in the funnel.

sure

> Also, when doing so, would it make sense to replace the LEFT JOINs with OUTER JOINs, to deal with the fact that some users might be missing from some data sources but not others?

yeah, that makes sense

> And if we really want to throw in the kitchen sink, could we pull in `referrer`, `utm_medium`, `utm_source`, `utm_campaign` by parsing the `attribution` column in `mozilla_vpn_external.users_v1` (or another data source if you've seen it)?

yeah. I'm planning to pull that in from the fxa events, but first I have to get that section working.

sql/moz-fx-data-shared-prod/stripe_derived/customers_v1/query.sql
```diff
@@ -0,0 +1,12 @@
...
+  created,
+  id,
+  ARRAY(
+    SELECT
+      IF(key = "userid", STRUCT("fxa_uid" AS key, TO_HEX(SHA256(value)) AS value), (key, value))
+    FROM
+      UNNEST(metadata)
+  ) AS metadata
+FROM
+  stripe_external.customers_v1
```
relud, line 12: @jklukas does this look right for hashing the fxa user id when it reaches a dataset accessible in redash?

relud: it's technically possible to hash it sooner, but making it hashed all the way back to where it first appears in `stripe_external.events_v1` would require hashing it in the python script before load, which feels like the wrong place to be applying this transformation.

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/acquisition_funnel_v1/query.sql
relud: > store the table resulting from these joins

I'm going to defer that to a separate PR

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/login_flows_v1/init.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE
```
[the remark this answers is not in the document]
relud: fixed, but also: `init.sql` isn't part of dry-runs

sql/moz-fx-data-shared-prod/stripe_derived/customers_v1/query.sql
relud: fixed

jklukas (approved):

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/login_flows_v1/metadata.yaml
```diff
@@ -0,0 +1,17 @@
+friendly_name: Mozilla VPN FxA Login Flows
+description: >
+  A list of users that attempted to login to Mozilla VPN.
```
felixlawrence, line 3: This makes it sound like they were existing VPN users, logging back in, whereas we want to study prospective VPN users, logging in to FxA as part of the VPN signup flow.

Is this accurate?
```suggestion
  A list of users that attempted to login to FxA, to sign up for Mozilla VPN.
```

Or should it be
```suggestion
  A list of users that attempted to login to FxA, to sign up for or log in to Mozilla VPN.
```

felixlawrence (approved):
I had a look through, and it seems to be sensible, i.e. aligned with what we need to measure. This is not a full technical review - let me know if you need me to provide one.

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/login_flows_v1/metadata.yaml
relud: this first suggestion lgtm

relud: oh, actually thinking it through the second one is more accurate

relud:
> I had a look through, and it seems to be sensible, i.e. aligned with what we need to measure. This is not a full technical review - let me know if you need me to provide one.

we should do one, but I don't think we should for that before merging this

PR #1468, opened 2020-10-21
Allow field addition for one day of clients_daily and friends

jklukas:
Allows Airflow to apply https://github.com/mozilla/bigquery-etl/pull/1455


sunahsuh (approved):

jklukas:
Note that this only worked for clients_daily. clients_last_seen and clients_first_seen both reference themselves, so the query itself fails if clients_daily contains new fields. Currently, it's necessary to do some manual manipulation of the queries or the schemas to propagate a new field to these tables.

PR #1473, opened 2020-10-21
Refactor shared UDFs used in GLAM queries into mozfun

acmiyaguchi:
These UDFs can be shared across desktop and glean queries. In particular, the method of the histogram aggregates being an estimator for the Dirichlet distribution and the bucketing methodology for scalars can be reused in the glean ETL.


sql/mozfun/glam/histogram_fill_buckets_dirichlet/metadata.yaml
```diff
@@ -0,0 +1,19 @@
...
+  Interpolate missing histogram buckets with empty buckets so it becomes a valid
+  estimator for the dirichlet distribution.
+
+  See: https://docs.google.com/document/d/1ipy1oFIKDvHr3R6Ku0goRjS11R1ZH1z2gygOGkSdqUg
+
+  Aggregate the histograms to the client level, to get a histogram {k1: p1,
+  k2:p2} where the p’s are proportions(and p1, p2 sum to 1). This is then the
+  clients estimated density. So every client is reduced to one row (i.e the
+  clients histograms are reduced to this single density). How can we create a
+  final estimator for the histogram density? The approach is similar to the
```
Ben-Wu, line 11: I don't think we need to mention case 2 here since it's not in this description

sql/mozfun/glam/histogram_fill_buckets_dirichlet/metadata.yaml
```diff
@@ -0,0 +1,19 @@
+description: |
+  Interpolate missing histogram buckets with empty buckets so it becomes a valid
+  estimator for the dirichlet distribution.
+
+  See: https://docs.google.com/document/d/1ipy1oFIKDvHr3R6Ku0goRjS11R1ZH1z2gygOGkSdqUg
+
+  Aggregate the histograms to the client level, to get a histogram {k1: p1,
+  k2:p2} where the p’s are proportions(and p1, p2 sum to 1). This is then the
+  clients estimated density. So every client is reduced to one row (i.e the
+  clients histograms are reduced to this single density). How can we create a
```
Ben-Wu, line 10:
```suggestion
  client's histograms are reduced to this single density). How can we create a
```

sql/mozfun/glam/histogram_fill_buckets_dirichlet/udf.sql
```diff
@@ -0,0 +1,70 @@
+-- udf_fill_buckets
+CREATE OR REPLACE FUNCTION glam.histogram_fill_buckets_dirichlet(
+  input_map ARRAY<STRUCT<key STRING, value FLOAT64>>,
+  buckets ARRAY<STRING>,
+  total_users INT64
```
Ben-Wu, line 5: I noticed there are two different `udf_fill_buckets` right now: one that divides by total users and one that doesn't.  Is it appropriate to have this dividing one be the persistent udf?

Ben-Wu (commented):
Are you planning on replacing all the temp function call with these when this lands?

sql/mozfun/glam/histogram_fill_buckets_dirichlet/udf.sql
acmiyaguchi: Yes, this is the variant used to aggregate histograms. I think it should replace `udf_fill_bucket` in scalar aggregates too, for consistency.

acmiyaguchi:
I'm planning on replacing the uses in the glam fenix etl first, but the uses in the desktop etl should also be replaced.

Ben-Wu (approved):

PR #1480, opened 2020-10-23
WIP Generated usage_all_time queries

jklukas:


bigquery_etl/usage/config/measures.yaml
```diff
@@ -0,0 +1,24 @@
...
+            sql: '>= 10'
+      - name: active_ticks_sum
+        sql: >-
+          SUM(COALESCE(
+            payload.processes.parent.scalars.browser_engagement_active_ticks,
+            payload.simple_measurements.active_ticks
+          ))
+        usages:
+          - name: had_active_ticks
+            sql: '> 0'
```
scholtzan, line 24: I'm wondering if it would make sense to define these measures in the `metadata.yaml` file instead. It might be a little easier to reason about how the queries work (get generated, scheduled, ...) if metadata is spread over fewer files? 
On the other hand, this would potentially increase the complexity of `metadata.yaml` files and make them harder to understand.

bigquery_etl/usage/__main__.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+with (HERE / "config" / "measures.yaml").open("r") as f:
+    data = yaml.safe_load(f)
+
+with (HERE / "templates" / "view_last_seen.sql").open("r") as f:
+    template = Template(f.read())
+
+for source in data["sources"]:
+    print(reformat(template.render(**source)))
```
scholtzan, line 20: I'm guessing this will be similar to what we do for the Airflow DAG generation in that we require users that add new measures to generate the query? And then have a similar process for the CI and running checks and dry runs for making sure the generated queries are valid?

bigquery_etl/usage/config/measures.yaml
jklukas: A limitation of our current approach of `metadata.yaml` files is that it treats each table/view as an independent entity. The only real concept we have of relations between tables is dependencies.

I think there's a need for a coherent approach to defining metadata that can influence the behavior of entire pipelines. For usage, that means being able to use one source of truth for what appears in clients_daily, clients_last_seen, clients_first_seen, and other downstream tables.

We have been doing this ad hoc with scripts, particularly for some of the emerging Glean ETL. So I definitely see opportunity for bringing more uniformity to how we approach that.

bigquery_etl/usage/__main__.py
jklukas: Yes, currently this sort of approach would rely on PR authors to run generation locally. I am definitely hesitant to add that complexity without a clear approach to CI, documentation, etc.

bigquery_etl/usage/config/measures.yaml
fbertsch: I've put together an approach for a straightforward query generation in https://github.com/mozilla/bigquery-etl/pull/1511. The idea there is to make it obvious how the templates and templating.yaml files interact, without any overhead of non-static arguments.

Eventually we would probably need to support something that is a bit more programmatic (ala baseline_clients_last_seen), but that seems to me very glean-specific and doable later.

PR #1486, opened 2020-10-26
Script for determining column storage sizes

scholtzan:
I used this for investigating https://bugzilla.mozilla.org/show_bug.cgi?id=1669288 which helped me finding the columns and histograms that contributed to the increase in partition sizes. I wonder if we would want to run this script on a daily (or maybe a less frequent period) basis?


sql/moz-fx-data-shared-prod/monitoring/column_size_v1/query.py
```diff
@@ -0,0 +1,121 @@
...
+from argparse import ArgumentParser
+from google.cloud import bigquery
+from functools import partial
+from multiprocessing.pool import ThreadPool
+
+parser = ArgumentParser(description=__doc__)
+parser.add_argument("--date", required=True)  # expect string with format yyyy-mm-dd
+parser.add_argument("--project", default="moz-fx-data-shared-prod")
+parser.add_argument(
+    "--tables", nargs="*", default=["telemetry_stable.main_v4"]
```
scholtzan, line 14: For now, this is a manually curated list of tables. I guess we could simply run it on all stable tables, but not sure if that's necessary. It would probably be good to add the `events` table since we also saw an increase there at some point.

jklukas: > It would probably be good to add the events table since we also saw an increase there at some point.

For the `events` table, I expect almost all of the concept will be in the events array, so probably won't be useful. It would probably be more actionable to maintain a count of `event_group` values, which is out of scope here.

jklukas (approved):

PR #1488, opened 2020-10-27
Add minimal set of tests for GLAM Fenix queries

### LINKED ISSUE #1459: Write minimal tests for Glean GLAM ETL scripts
There are no tests for glean GLAM ETL scripts, which makes it difficult to determine whether the scripts are working as intended. A set of tests would be useful.

acmiyaguchi:
This fixes #1459.


script/entrypoint
```diff
@@ -19,7 +19,8 @@ fi
 
 if [ "$#" = 0 ]; then
     # Default to 20 workers because most of the tests are io-bound
-    exec pytest --black --flake8 --mypy-ignore-missing-imports --pydocstyle --numprocesses 20
```
acmiyaguchi, line 22: This was causing test failures because of mypy or pytest-mypy, which is due to the free floating `data.py` files in each of the sql tests being treated as a duplicate module (https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/6996/workflows/ad741abd-36a0-4649-bf39-1e8e0d1616b1/jobs/35787).

tests/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__scalar_bucket_counts_v1/test_minimal/data.py
```diff
@@ -0,0 +1,164 @@
...
+                "metric_type": "counter",
+                "key": "",
+                "agg_type": "count",
+                "value": 8.0,
+            },
+        ],
+    },
+]
+
+# TODO: why are the range_min and range_max set at these values?
```
acmiyaguchi, line 50: The behavior of range_min and range_max is not intuitive, there might be something weird going on here.

tests/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__scalar_percentiles_v1/test_minimal/data.py
```diff
@@ -0,0 +1,84 @@
...
+                "metric_type": "counter",
+                "key": "",
+                "agg_type": "count",
+                "value": 8.0,
+            },
+        ],
+    },
+]
+
+# TODO: the total user count is very wrong
```
acmiyaguchi, line 50: This is the most egregious error, there should be at most 2 users.

acmiyaguchi (commented):
There are a couple of issues that should be sorted out.

tests/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__scalar_percentiles_v1/test_minimal/data.py
Ben-Wu: Looking at the query it looks like this is because of the cross join with the `*` combos.  In the final query you should be able to do `COUNT(DISTINCT(client_id)) AS total_users`

tests/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__scalar_probe_counts_v1/test_minimal/data.py
```diff
@@ -0,0 +1,161 @@
...
+        "metric": "places_manager_write_query_count",
+        "metric_type": "counter",
+        "os": "*",
+        "ping_type": "*",
+        "range_max": 5.0,
+        "range_min": 0.0,
+    }
+]
+
+# TODO: programmatic generation of aggregates.
```
Ben-Wu, line 31: Could you use a fake bucket count and range to make this simpler?

tests/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__clients_histogram_aggregates_v1/test_minimal/data.py
```diff
@@ -0,0 +1,97 @@
...
+CLIENTS_HISTOGRAM_AGGREGATES = [
+    {
+        "sample_id": 1,
+        "client_id": UUID,
+        "ping_type": "metrics",
+        "os": "Android",
+        "app_version": 84,
+        "app_build_id": APP_BUILD_ID,
+        "channel": "nightly",
+        "histogram_aggregates": [],
```
Ben-Wu, line 25: Although not exactly minimal, it would be more interesting if the test summed onto an existing aggregate

Ben-Wu (commented):
This is useful and it helped me understand the etl better.

I don't really understand what's happening with scalar bucket counts though.

script/entrypoint
acmiyaguchi: I think it has to do with pytest collecting all of the python files under the test directory as individual modules (although this is mostly conjecture). Since I have a `data.py` file in each test, these appear to be the same module several times. When run outside of pytest, there is no `__init__.py` file and therefore it's ignored. 

If I add a `__init__.py` file to each test going up the directory levels so each test appears like a unique module, then the pytest mypy plugin will work correctly. I want these scripts to be standalone, not part of a module though.

tests/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__histogram_percentiles_v1/test_minimal/data.py
```diff
@@ -0,0 +1,262 @@
...
+
+ROOT = Path(__file__).parent
+
+SUBMISSION_DATE = "2020-10-01"
+APP_BUILD_ID = "2020100100"
+
+HISTOGRAM_PROBE_COUNTS = [
+    {
+        "agg_type": "histogram",
+        "aggregates": [
```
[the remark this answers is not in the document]
acmiyaguchi: This is a fair assessment; I found it easier to copy and paste while putting these together, but it would be easier to maintain with a smaller set.

tests/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__scalar_percentiles_v1/test_minimal/data.py
acmiyaguchi: Thanks for catching this, this fixes the user counts.

tests/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__clients_histogram_aggregates_v1/test_minimal/data.py
acmiyaguchi: Added a separate test summing onto an existing aggregate.

tests/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__scalar_bucket_counts_v1/test_minimal/data.py
Ben-Wu: Can clients_scalar_aggregates have `*` values for ping_type and os?  Could this be caused by the cross join with the static combos?  Maybe not because app_build_id is defined.  If I understand correctly, min/max should be 2/3 right?

Ben-Wu (commented):
I think the range in scalar bucket counts should be figured out before merging.  But otherwise this is looks good to me.

tests/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__scalar_bucket_counts_v1/test_minimal/data.py
acmiyaguchi: Even with the cross join, I'm not sure how the range max (in log space) could be set to 5 when the max value is 8.

Ben-Wu: I have a feeling it might be related to the cross join and the input data having `"ping_type": "*", "os": "*",` but I don't really know.  If you can't figure out then I don't want to block merging this.  It would be nice to verify in case there's actually an issue in the etl though.

script/entrypoint
```diff
@@ -19,7 +19,8 @@ fi
...
+    mypy --ignore-missing-imports bigquery_etl tests
```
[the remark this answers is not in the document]
acmiyaguchi: This seems to work, I'll opt to use the glob ignore option.

Ben-Wu (approved):
I think this is acceptable with the range issue if we can't figure it out at all but maybe create an issue so we don't forget (although that might still happen).

tests/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__scalar_bucket_counts_v1/test_minimal/data.py
acmiyaguchi: It's tangentially related to the cross join. I'm going to follow up with this in a separate PR because this one is already very large for a minimal PR.

PR #1491, opened 2020-10-27
Use resource id, not event id, when updating stripe resource tables

relud:
resources are deduped by id, but this was incorrectly using event id, causing all update events to be considered a new resource


acmiyaguchi (approved):
Does the interpretation of https://github.com/mozilla/bigquery-etl/pull/1489 change based on this issue?

relud:
> Does the interpretation of #1489 change based on this issue?

nope, that's a completely separate issue.

PR #1492, opened 2020-10-27
Add derived tables for blog.m.o google analytics

Ben-Wu:
Port of [blogsSiteMetricsByLandingPageSummaryTable.py](https://github.com/mozilla/marketing-analytics/blob/master/Airflow/blogsSiteMetricsByLandingPageSummaryTable.py) and [blogsSiteMetricsSummaryTable.py](https://github.com/mozilla/marketing-analytics/blob/master/Airflow/blogsSiteMetricsSummaryTable.py).  The resulting tables are the same but I just unnested the queries and created some intermediate tables (goals and sessions) so the two final queries can share.

I validated that the tables match by doing a full join on the existing tables and the new tables:
```sql
SELECT
  *
FROM (
  SELECT
    *,
    'a' AS a,
  FROM
    `benwu-test-1.ga_derived.blogs_landing_page_summary_v1` )
FULL JOIN (
  SELECT
    *,
    'b' AS b,
  FROM
    `ga-mozilla-org-prod-001.66602784.blogs_landing_page_summary` )
USING
  (date, device_category, operating_system, browser, LANGUAGE, country, source, medium, campaign, content, blog, subblog, landing_page, cleaned_landing_page, sessions, downloads, socialShare, newsletterSubscription)
WHERE
  date BETWEEN '2020-10-20'
  AND '2020-10-26'
  AND (a is null or b is null)
ORDER BY
  date, device_category, operating_system, browser, LANGUAGE, country, source, medium, campaign, content, blog, subblog, landing_page, cleaned_landing_page, sessions, downloads, socialShare, newsletterSubscription
```


dags.yaml
```diff
@@ -240,3 +240,12 @@ bqetl_org_mozilla_fenix_derived:
     retry_delay: 30m
     start_date: '2020-10-18'
   schedule_interval: daily
+
+bqetl_google_analytics_derived:
+  schedule_interval: 0 16 * * *
+  default_args:
+    owner: bewu@mozilla.com
+    email: ['bewu@mozilla.com']
```
Ben-Wu, line 248: George, do you still want to get notifications for these queries?

sql/moz-fx-data-marketing-prod/ga_derived/blogs_daily_summary_v1/init.sql
```diff
@@ -0,0 +1,29 @@
+CREATE TABLE IF NOT EXISTS
+  `moz-fx-data-marketing-prod.ga_derived.blogs_daily_summary_v1`(
+    date DATE,
+    deviceCategory STRING,
```
Ben-Wu, line 4: The existing queries have a mix of camel case and underscores. I kept them as is for now so consumers don't need to change any logic but would it be easy to change those so we can standardize to underscore?

Ben-Wu (commented):
@gkabbz Some questions for you.  I'd appreciate an initial look over the queries and then I'll get a more thorough review from data platform

dags.yaml
gkabbz: Nope. Looking to make my escape.

sql/moz-fx-data-marketing-prod/ga_derived/blogs_daily_summary_v1/init.sql
gkabbz: I think this was because the GA drops have camel case and I hadn't really settled on a method. I'm fine with changing them. Once we migrate, all the end points that use the data will need to be updated so we may as well go in and make the changes now to be consistent with the way you guys do things.

sql/moz-fx-data-marketing-prod/ga_derived/blogs_daily_summary_v1/query.sql
```diff
@@ -0,0 +1,44 @@
...
+  SUM(socialShare) AS socialShare,
+  SUM(newsletterSubscription) AS newsletterSubscription
+FROM
+  `moz-fx-data-marketing-prod.ga_derived.blogs_sessions_v1` AS sessions_table
+LEFT JOIN
+  `moz-fx-data-marketing-prod.ga_derived.blogs_goals_v1`
+USING
+  (date, visitIdentifier)
+LEFT JOIN
+  `moz-fx-data-marketing-prod.static.standardized_country_list` AS standardized_country_list
```
gkabbz, line 26: Would be great if this standardized country list was put somewhere central to be shared across Moz. I'd made this list coz I was tired of trying to reconcile between GA and Telemetry. But I think I'm the only one using it / keeping it up to date. Would be great if we could make it something better maintained by Data Eng that applies against other data sources and is used centrally. *This is a nice to have outside this project*

gkabbz: FYI I just have a local csv that I update when the need arises and the I overwrite that table. Maybe that csv can live somewhere else.

gkabbz (approved):
Love the structure. Gonna be a lot easier to update changes to the logic going forward.

sql/moz-fx-data-marketing-prod/ga_derived/blogs_daily_summary_v1/query.sql
jklukas: We already have such a table. See `country_names` and `country_codes` in https://github.com/mozilla/bigquery-etl/tree/master/sql/moz-fx-data-shared-prod/static

The country_names table in particular includes listings for each of the country names that show up in the FxA metrics pipeline. If GA produces different logical country names, it likely makes sense to add them to `country_names` and use that int this query as well.

Ben-Wu: I made it into a csv here:
https://github.com/mozilla/bigquery-etl/blob/07603860fdb71bcabcd8b05fcc9e12267abaab1d/sql/moz-fx-data-marketing-prod/static/standardized_country_list/data.csv

We can just move this into shared-prod.  The logic for this table is a little different from the existing ones so I think a new table works best.

dags.yaml
```diff
@@ -240,3 +240,12 @@ bqetl_org_mozilla_fenix_derived:
     retry_delay: 30m
     start_date: '2020-10-18'
   schedule_interval: daily
+
+bqetl_google_analytics_derived:
+  schedule_interval: 0 16 * * *
```
jklukas, line 245: Can we comment here on why this runs at 16:00 UTC? I assume it has to do with the schedule for GA delivery to BQ.

sql/moz-fx-data-marketing-prod/ga_derived/blogs_daily_summary_v1/query.sql
```diff
@@ -0,0 +1,44 @@
...
+  SUM(social_share) AS social_share,
+  SUM(newsletter_subscription) AS newsletter_subscription,
+FROM
+  `moz-fx-data-marketing-prod.ga_derived.blogs_sessions_v1` AS sessions_table
+LEFT JOIN
+  `moz-fx-data-marketing-prod.ga_derived.blogs_goals_v1`
+USING
+  (date, visit_identifier)
+LEFT JOIN
+  `moz-fx-data-marketing-prod.static.standardized_country_list` AS standardized_country_list
```
jklukas, line 26:
```suggestion
  `moz-fx-data-shared-prod.static.standardized_country_list` AS standardized_country_list
```

sql/moz-fx-data-marketing-prod/ga_derived/blogs_daily_summary_v1/query.sql
```diff
@@ -0,0 +1,44 @@
...
+  campaign,
+  content,
+  blog,
+  subblog,
+  SUM(sessions) AS sessions,
+  SUM(downloads) AS downloads,
+  SUM(social_share) AS social_share,
+  SUM(newsletter_subscription) AS newsletter_subscription,
+FROM
+  `moz-fx-data-marketing-prod.ga_derived.blogs_sessions_v1` AS sessions_table
```
jklukas, line 20:
```suggestion
  blogs_sessions_v1 AS sessions_table
```

Per https://github.com/mozilla/bigquery-etl#queries our preferred style is to make table references as short as possible, so we should consistently leave off the project and dataset where they match the target project and dataset.

sql/moz-fx-data-shared-prod/static/standardized_country_names/description.txt
```diff
@@ -0,0 +1 @@
+Mappings from country codes or non-standard names to a standardized country name.
```
jklukas, line 1: We should make it clear that this is GA-specific, considering we have some odd entries in here like "worldwide" and "(not set)".  I would recommend changing the name of the table to `ga_standardized_country_names_v1` and mentioning Google Analytics specifically in this description.

jklukas (approved):
r+wc

Just a few details to sort out for the standardized country names. Otherwise, looks good.

dags.yaml
Ben-Wu: Yes the GA export runs at 15:00 UTC.  The existing derived table jobs run at 16:00 and there haven't been issues so I'll continue that here.  I'll add a comment.

PR #1493, opened 2020-10-28
Update glam fenix etl with updated scalar bucketing

### LINKED ISSUE #1134: Scalar Redesign for Fenix
Desktop glam has a new method for creating buckets for scalars that needs to be adopted for fenix/glean glam

acmiyaguchi:
This fixes #1134 to adopt an approach like https://github.com/mozilla/bigquery-etl/pull/1022.

This updates the scalar bucketing to mirror the desktop scalar bucketing. It's done slighly differently since the bucketing is in a single stage on the desktop etl. This also merges normalized histograms by the number of clients reporting that bucket (`histogram_fill_buckets_dirichlet`). 

I'm worried about using the updated binning methodology for scalars, because it's determined by the range of each probe, which may change from day to day. I did a quick query to validate the buckets -- I would expect there to be 100 buckets, but it seems like it changes between probes. I'd be more comfortable with parameterized bucketing like the functional bucketing used for glean histograms. 

@emtwo is the discussion around the scalar bucketing redesign documented anywhere? I'd like a bit more context to see if my understanding of the problem is generally correct. 

```sql
WITH
  probes AS (
  SELECT
    * EXCEPT(aggregates),
    ARRAY(
    SELECT
      key
    FROM
      UNNEST(aggregates)
    ORDER BY
      SAFE_CAST(key AS int64)) AS buckets
  FROM
    `glam-fenix-dev.glam_etl_dev.org_mozilla_fenix_glam_nightly__view_probe_counts_v1`
  WHERE
    ping_type = "*"
    AND os="*"
    AND app_build_id="*"
    AND channel="*"
    AND app_version=83
    AND metric_type IN ("counter",
      "quantity") )
SELECT
  metric,
  key,
  count(distinct bucket) as n_buckets
FROM
  probes,
  unnest(buckets) bucket
  group by 1,2
  order by 3 desc
```

```json
[
  {
    "metric": "gfx_display_primary_width",
    "key": "",
    "n_buckets": "106"
  },
  {
    "metric": "glean_upload_deleted_pings_after_quota_hit",
    "key": "",
    "n_buckets": "106"
  },
  {
    "metric": "gfx_adapter_primary_ram",
    "key": "",
    "n_buckets": "106"
  },
  {
    "metric": "gfx_display_count",
    "key": "",
    "n_buckets": "106"
  },
  {
    "metric": "gfx_display_primary_height",
    "key": "",
    "n_buckets": "106"
  },
  {
    "metric": "glean_upload_pending_pings",
    "key": "",
    "n_buckets": "105"
  },
  {
    "metric": "places_manager_write_query_count",
    "key": "",
    "n_buckets": "105"
  },
  {
    "metric": "glean_error_preinit_tasks_overflow",
    "key": "",
    "n_buckets": "105"
  },
  {
    "metric": "places_manager_read_query_count",
    "key": "",
    "n_buckets": "105"
  },
  {
    "metric": "glean_validation_metrics_ping_count",
    "key": "",
    "n_buckets": "105"
  },
  {
    "metric": "metrics_top_sites_count",
    "key": "",
    "n_buckets": "104"
  },
  {
    "metric": "logins_store_read_query_count",
    "key": "",
    "n_buckets": "103"
  },
  {
    "metric": "events_total_uri_count",
    "key": "",
    "n_buckets": "103"
  },
  {
    "metric": "metrics_tabs_open_count",
    "key": "",
    "n_buckets": "103"
  },
  {
    "metric": "logins_store_write_query_count",
    "key": "",
    "n_buckets": "102"
  },
  {
    "metric": "metrics_recently_used_pwa_count",
    "key": "",
    "n_buckets": "100"
  },
  {
    "metric": "logins_store_unlock_count",
    "key": "",
    "n_buckets": "96"
  }
]
```


Ben-Wu (commented):
Doesn't the desktop version have the same problem with the ranges changing from day to day? Parameterized bucketing sounds reasonable but wouldn't you need the range to determine the parameters?  Can you explain what you had in mind for that?

I'm having trouble wrapping my head around this.  I think the etl is just a bit intractable 😕

acmiyaguchi:
The desktop ETL does have this problem, but I'm going to stick with this for consistency. For the parameterized bucketing, you would need the ranges, but the bucket values wouldn't change between iterations (i.e. a new max value will add a new bucket instead of shifting existing values). I don't know how to choose good parameters, though.

I'm also not sure if the front-end code can support where the number of buckets of buckets changes between days.There's a little bit of discussion here: https://mozilla.slack.com/archives/GSTS19DTP/p1603986443142100

acmiyaguchi:
I ran the following commands for testing:

Copy daily tables from the "prod" dataset:

```bash
script/glam/list_tables glam_etl | \
    grep clients_daily | grep -v org_mozilla_firefox | grep -v view | \
xargs -I {} echo bq cp -f {} {} | sed 's/glam_etl/glam_etl_dev/2'
```

Remove the existing aggregates tables:

```bash
script/glam/list_tables glam_etl_dev | \
    grep -v clients_daily | grep aggregates_v1 | \
    xargs -I {} echo "bq rm -f {}"
```

Run the incremental backfill:

```bash
bqetl glam glean backfill-incremental org_mozilla_fenix_glam_nightly 2020-08-01 2020-11-01
```

Export the results:

```bash
bqetl glam glean export org_mozilla_fenix_glam_nightly 
```

bigquery_etl/glam/templates/probe_counts_v1.sql
```diff
@@ -1,8 +1,48 @@
...
+        WHEN
+          metric_type = 'custom_distribution_exponential'
+        THEN
+          mozfun.glam.histogram_generate_exponential_buckets(range_min, range_max, bucket_count)
+        WHEN
+          metric_type = 'custom_distribution_linear'
+        THEN
+          mozfun.glam.histogram_generate_linear_buckets(range_min, range_max, bucket_count)
+        ELSE
+          []
```
Ben-Wu, line 34: I know this isn't a new change but I'm curious if this is an explicitly allowed behaviour: passing an unknown metric type.  What's the behaviour if no buckets are returned?

Ben-Wu (approved):
Changes look fine to me.  It's just a matter of sorting out the changing buckets with the frontend.

script/glam/test/test_glean_org_mozilla_fenix_glam_nightly
```diff
@@ -28,19 +30,27 @@ if [[ $skip_generate == false ]]; then
...
     done
     PRODUCT=$logical_app_id STAGE=incremental $dir/generate_glean_sql
 fi
 
-for app_id in "${app_ids[@]}"; do
-    PRODUCT=$app_id STAGE=daily $dir/run_glam_sql
-done
+if [[ $generate_only != false ]]; then
+    exit
+    bqetl glam glean update-schemas
```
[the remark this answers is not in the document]
acmiyaguchi: This works like you'd expect -- I just put them in the wrong order.

bigquery_etl/glam/templates/probe_counts_v1.sql
acmiyaguchi: This would result in an empty list from this udf. If the buckets are unnested, it would cause the row to disappear from the resulting set. Otherwise, the histogram would probably just be empty.

```sql
select array_agg(cast(item as int64)) from unnest([]) item
-- {"_f0": []|}
```

PR #1494, opened 2020-10-28
Bug 1673903 - Add urlbar-searchmode to allowed SAP sources

wlach:


Ben-Wu (approved):
Does this require some backfill?

wlach:
> Does this require some backfill?

I am hoping we can avoid it, because it hasn't reached the release channel yet. I'll ask :mconnor on the bug.

Issue #1495, opened 2020-10-28
mozfun publishing failing for glam.percentile


scholtzan:
```
[2020-10-28 01:05:10,559] {logging_mixin.py:112} INFO - [2020-10-28 01:05:10,559] {pod_launcher.py:125} INFO - b'  File "/usr/local/lib/python3.8/sre_parse.py", line 1039, in parse_template\n'
[2020-10-28 01:05:10,559] {logging_mixin.py:112} INFO - [2020-10-28 01:05:10,559] {pod_launcher.py:125} INFO - b'    this = chr(ESCAPES[this][1])\n'
[2020-10-28 01:05:10,559] {logging_mixin.py:112} INFO - [2020-10-28 01:05:10,559] {pod_launcher.py:125} INFO - b"KeyError: '\\\\u'\n"
[2020-10-28 01:05:10,560] {logging_mixin.py:112} INFO - [2020-10-28 01:05:10,559] {pod_launcher.py:125} INFO - b'\n'
[2020-10-28 01:05:10,560] {logging_mixin.py:112} INFO - [2020-10-28 01:05:10,560] {pod_launcher.py:125} INFO - b'During handling of the above exception, another exception occurred:\n'
[2020-10-28 01:05:10,560] {logging_mixin.py:112} INFO - [2020-10-28 01:05:10,560] {pod_launcher.py:125} INFO - b'\n'
[2020-10-28 01:05:10,560] {logging_mixin.py:112} INFO - [2020-10-28 01:05:10,560] {pod_launcher.py:125} INFO - b'Traceback (most recent call last):\n'
[2020-10-28 01:05:10,560] {logging_mixin.py:112} INFO - [2020-10-28 01:05:10,560] {pod_launcher.py:125} INFO - b'  File "/usr/local/lib/python3.8/runpy.py", line 194, in _run_module_as_main\n'
[2020-10-28 01:05:10,560] {logging_mixin.py:112} INFO - [2020-10-28 01:05:10,560] {pod_launcher.py:125} INFO - b'    return _run_code(code, main_globals, None,\n'
[2020-10-28 01:05:10,560] {logging_mixin.py:112} INFO - [2020-10-28 01:05:10,560] {pod_launcher.py:125} INFO - b'  File "/usr/local/lib/python3.8/runpy.py", line 87, in _run_code\n'
[...]

PR #1496, opened 2020-10-28
Fix GLAM UDF descriptions

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/1495

I think we should add some checks in the CI for this.

### DIFF
--- a/sql/mozfun/glam/histogram_generate_functional_buckets/metadata.yaml
+++ b/sql/mozfun/glam/histogram_generate_functional_buckets/metadata.yaml
@@ -6,7 +6,7 @@ description: |
   A functional bucketing algorithm. The bucket index of a given sample is
   determined with the following function:
 
-  i = ⌊n log<sub>base</sub>(𝑥)⌋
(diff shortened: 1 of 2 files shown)


sql/mozfun/glam/histogram_generate_functional_buckets/metadata.yaml
```diff
@@ -6,7 +6,7 @@ description: |
   A functional bucketing algorithm. The bucket index of a given sample is
   determined with the following function:
 
-  i = ⌊n log<sub>base</sub>(𝑥)⌋
+  i = \floor{n log<sub>base</sub>(x)}
```
scholtzan, line 9: I believe this is the LaTeX way to write this, but am also open to alternatives

acmiyaguchi: Maybe something like:

```
$$ \lfloor{n log_{\text{base}}{(x)}}\rfloor $$
```

acmiyaguchi (approved):

PR #1497, opened 2020-10-28
Order mozfun.glam.histogram_fill_buckets by string sorting

acmiyaguchi:
https://github.com/mozilla/bigquery-etl/blob/1529510c6eafbcf2e5b5ed46aa123b99ee91e7c2/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__scalar_probe_counts_v1/query.sql#L329-L336

This udf will get used where the buckets are not necessarily integers.

### DIFF
--- a/sql/mozfun/glam/histogram_fill_buckets/udf.sql
+++ b/sql/mozfun/glam/histogram_fill_buckets/udf.sql
@@ -18,11 +18,7 @@ RETURNS ARRAY<STRUCT<key STRING, value FLOAT64>> AS (
         SAFE_CAST(key AS STRING) = e.key
     )
     SELECT
-      ARRAY_AGG(
-        STRUCT<key STRING, value FLOAT64>(SAFE_CAST(key AS STRING), value)
-        ORDER BY
-          CAST(key AS int64)
-      )
+      ARRAY_AGG(STRUCT<key STRING, value FLOAT64>(SAFE_CAST(key AS STRING), value) ORDER BY key)
     FROM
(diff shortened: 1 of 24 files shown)


sql/mozfun/glam/histogram_fill_buckets/udf.sql
```diff
@@ -18,11 +18,7 @@ RETURNS ARRAY<STRUCT<key STRING, value FLOAT64>> AS (
         SAFE_CAST(key AS STRING) = e.key
     )
     SELECT
-      ARRAY_AGG(
-        STRUCT<key STRING, value FLOAT64>(SAFE_CAST(key AS STRING), value)
-        ORDER BY
-          CAST(key AS int64)
-      )
+      ARRAY_AGG(STRUCT<key STRING, value FLOAT64>(SAFE_CAST(key AS STRING), value) ORDER BY key)
```
Ben-Wu, line 21: Is it ever possible that key isn't a string and the cast is needed?  I don't think it is because the parameter is a string but maybe there's a reason it's here.  Fine either way, I'm just curious.

Ben-Wu (approved):

sql/mozfun/glam/histogram_fill_buckets/udf.sql
acmiyaguchi: If the input_map key type were `ANY`, this would be needed, but so far I don't think this has been the case. I'll simplify this a bit.

PR #1499, opened 2020-10-29
Fix `bqetl mozfun publish`

acmiyaguchi:
This updates `bigquery_etl.cli.routine` so the `publish` command does not fail. It uses the default project consistently across the commands, and defaults to the project-id option if provided.

### DIFF
--- a/bigquery_etl/cli/routine.py
+++ b/bigquery_etl/cli/routine.py
@@ -68,8 +68,7 @@ project_id_option = click.option(
     "--project-id",
     "--project_id",
     help="GCP project ID",
-    default=None,
-    callback=is_valid_project,
+    callback=lambda *args: is_valid_project(*args) if args[-1] else args[-1],
 )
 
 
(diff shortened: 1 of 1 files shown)


bigquery_etl/cli/routine.py
```diff
@@ -312,31 +299,29 @@ def validate(ctx, name, sql_dir, project_id):
     help="The GCS path in the bucket where dependency files are uploaded to.",
 )
 @click.pass_context
-def publish(ctx, path, project, dependency_dir, gcs_bucket, gcs_path):
```
acmiyaguchi, line 315: This is the main bug.

bigquery_etl/cli/routine.py
```diff
@@ -312,31 +299,29 @@ def validate(ctx, name, sql_dir, project_id):
...
 
     if not is_authenticated(project_id):
         click.echo("User needs to be authenticated to publish routines.", err=True)
         sys.exit(1)
 
-    if path and is_valid_dir(path):
+    if path and is_valid_dir(ctx, None, path):
         click.echo(f"Publish routines to {project_id}")
-        publish_routines.publish(
-            project_dirs(project_id),
```
acmiyaguchi, line 332: `target` needs to a directory instead of a list of directories.

scholtzan (approved):
👍

PR #1507, opened 2020-11-02
Use union instead of join to make query more readable

relud:
reusing the pattern from https://github.com/mozilla/bigquery-etl/blob/415ee2f/sql/moz-fx-data-shared-prod/mozilla_vpn_derived/login_flows_v1/query.sql

### DIFF
--- a/sql/moz-fx-data-shared-prod/mozilla_vpn_derived/protected_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/mozilla_vpn_derived/protected_v1/query.sql
@@ -1,5 +1,5 @@
 -- WARNING: on mobile this is undercounted and may not be measurable
-WITH _current AS (
+WITH base AS (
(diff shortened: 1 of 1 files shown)


sql/moz-fx-data-shared-prod/mozilla_vpn_derived/protected_v1/query.sql
```diff
@@ -10,18 +10,16 @@ WITH _current AS (
...
+  FROM
+    protected_v1
 )
 SELECT
   fxa_uid,
-  IF(
-    _previous.first_protected IS NULL
-    OR _previous.first_protected > _current.first_protected,
-    _current,
-    previous
```
relud, line 20: this also fixes a bug where this line was supposed to be `_previous`

jklukas (approved):

PR #1509, opened 2020-11-03
Add support for --schema_update_option in run_multipart_query.py

scholtzan:
For `run_multipart_query` we need to add explicit support for `--schema_update_option`

### DIFF
--- a/bigquery_etl/run_multipart_query.py
+++ b/bigquery_etl/run_multipart_query.py
@@ -88,6 +88,11 @@ parser.add_argument(
     type=dirpath,
     help="Path to the query directory that contains part*.sql files",
 )
+parser.add_argument(
+    "--schema_update_option",
+    help="Optional options for updating the schema.",
+    default="",
+)
 
 temporary_dataset = None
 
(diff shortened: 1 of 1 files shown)


bigquery_etl/run_multipart_query.py
```diff
@@ -157,6 +162,17 @@ def main():
...
+
+            if args.schema_update_option == "ALLOW_FIELD_ADDITION":
+                schema_update_options = [
+                    bigquery.SchemaUpdateOption.ALLOW_FIELD_ADDITION
+                ]
+            elif args.schema_update_option == "ALLOW_FIELD_RELAXATION":
+                schema_update_options = [
+                    bigquery.SchemaUpdateOption.ALLOW_FIELD_RELAXATION
+                ]
+
```
scholtzan, line 175: It looks like there are only [2 schema update options](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.job.SchemaUpdateOption.html) that we need to consider here

sunahsuh (approved):
👍

PR #1511, opened 2020-11-03
WIP: Desktop events_daily & Generic query generation

fbertsch:


jklukas:
From https://github.com/mozilla/bigquery-etl/pull/1480#discussion_r516805647

> Eventually we would probably need to support something that is a bit more programmatic (ala baseline_clients_last_seen), but that seems to me very glean-specific and doable later.

We already have several streams of generated queries in this repository, each governed by independent scripts. We have glean usage (baseline_clients_last_seen and friends), we have GLAM ETL, we have copy_deduplicate, we have shredder, and probably more.

I do like that this PR provides a framework into which we could plug additional use cases, but from my initial glance through, I'm not convinced that it's flexible enough to justify adding the machinery.

The framework here does provide the ability to have a single set of templates produce multiple output tables based on different source input. But within that set of templates, there's still a lot of duplicated logic, making it easy to have logic drift between `init.sql` and `query.sql` for example. It also doesn't look possible here to generate the list of inputs as we would need for generic Glean-type pipelines.

If we want to introduce a query generation framework that's intended to be generic and fit a wide variety of use cases, we should invest the time to write a proposal and make sure we have a vision that will make the repo overall more understandable, and that ideally would be able to handle some of the existing generated query use cases.

I am somewhat skeptical that we'll be able to put together a single framework for query generation. Instead, we may want something of a plugin system where each use case is driven by python code, but we have those scripts organized in a clear directory structure, with any relevant configuration files living alongside the code of their plugin. We would need a clear contract for the plugins (is it expected that developers run them locally and commit the generated output? etc.) and a clear order in which they're run. This could give us the flexibility of our current system of isolated scripts, where the needs for each use case are different, but having them better organized and documented as a plugin system could make it substantially easier to reason about where generated queries are coming from.

fbertsch:
@jklukas thank you for the comment! I agree that a fully-formed proposal is best-suited for this use case. In the meantime I will move this code to its own directory in `bigquery_etl` while we work out how we want to support this in a more general case.

One disagreement I have is that we should support a plugin system. That is too heavy-handed for these use cases. My take is that we should support generated queries on:
- A static list of datasets, as this is doing
- A static list of datasets and pings
- Dynamic list of Glean applications
- Dynamic list of Glean applications and Glean pings

If we supported these 4 use cases, that (I believe) should cover GLAM and baseline_clients_*. I'm not sure if we should support shredder. I'll shoe-horn this into a proposal and we can hash out the details there.

fbertsch:
@jklukas it doesn't look like the current build failures are an actual problem with this code, and it's ready to be reviewed. However we'll have to fix that before being able to merge.

bigquery_etl/dryrun.py
```diff
@@ -223,12 +224,21 @@ def is_valid(self):
 
 def main():
     """Dry run all SQL files in the project directories."""
+    parser = ArgumentParser(description=main.__doc__)
```
fbertsch, line 227: This was to support local dryrunning of single queries

bigquery_etl/cli/query.py
```diff
@@ -59,12 +59,16 @@ def _queries_matching_name_pattern(pattern, sql_path, project_id):
...
 )
 
-project_id_option = click.option(
-    "--project-id",
-    "--project_id",
-    help="GCP project ID",
-    default=None,
-)
+
+def project_id_option(default=None):
```
fbertsch, line 63: This was initially used in the new `bqetl query generate`. I dropped that function, but this seems like a useful DRY addition.

bigquery_etl/format_sql/formatter.py
```diff
@@ -288,7 +288,7 @@ def inline_block_format(tokens, max_line_length=100):
                 last_token_was_opening_bracket = line.ends_with_opening_bracket
 
 
-def reformat(query, format_=inline_block_format):
+def reformat(query, format_=inline_block_format, trailing_newline=False):
```
fbertsch, line 291: This lets us generate formatted queries that don't need to be reformatted.

sql/moz-fx-data-shared-prod/telemetry/deanonymized_events/view.sql
```diff
@@ -0,0 +1,47 @@
+CREATE OR REPLACE VIEW
+  telemetry.deanonymized_events
```
fbertsch, line 2: Are we free to use the `tmp` dataset for these kinds of views? I don't actually want this to be available to end users.

jklukas: We have one or two cases where we put views in `_derived` because they weren't intended for end users, which is what I'd recommend here.

fbertsch:
@jklukas Much cleaner history now. It's not perfect but should be much easier to review.

sql/moz-fx-data-shared-prod/telemetry/event_types_update_v1/metadata.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - frank@mozilla.com
+labels:
+  application: telemetry
+  incremental: false
+  schedule: daily
+scheduling:
+  dag_name: bqetl_event_rollup
+  destination_table: null
+  query_file_path: 'sql/moz-fx-data-shared-prod/telemetry
+  /event_types/query.sql'
```
jklukas, line 14: This is probably incorrect with the linebreak.

fbertsch: It is odd, but is still parseable yaml. I think the yamllinter recommended it. I would prefer it on a newline with `>`, let me see if it complains with that approach.

jklukas (commented):
Finished initial round of comments. I may have more if I can get back to this today. General approach looks reasonable and leaves room for other use cases to be added into CI, etc. One of my main concerns is making sure there's enough documentation in the python files for folks to quickly grasp what's going on.

sql/moz-fx-data-shared-prod/telemetry/event_types_update_v1/metadata.yaml
jklukas: Based on the generated output, it looks like it collapses to a single space, which leads to an invalid path.

bigquery_etl/events_daily/generate_queries.py
```diff
@@ -0,0 +1,152 @@
...
+
+from argparse import ArgumentParser
+from bigquery_etl.format_sql.formatter import reformat
+from dataclasses import dataclass
+from jinja2 import Environment, FileSystemLoader
+from pathlib import Path
+from typing import List, Optional
+
+
+QUERY_FILES = {
```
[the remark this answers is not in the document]
fbertsch: I can see that being confusing -- it represents any valid files we might find in a query directory. Usually they are associated with queries. Perhaps `TEMPLATED_FILES`?

bigquery_etl/events_daily/generate_queries.py
```diff
@@ -0,0 +1,152 @@
...
+
+@dataclass
+class Template:
+    """A template, to be filled with args and saved as a file."""
+
+    name: str
+    env: Environment
+
+    def generate(self, write_path, args):
+        """Generate this template at the specified write_path with the specified args."""
```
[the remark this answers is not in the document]
fbertsch: Agreed, `render` makes more sense here.

sql/moz-fx-data-shared-prod/org_mozilla_firefox/event_types_update_v1/query.sql
```diff
@@ -0,0 +1,15 @@
...
+  AS
+  SELECT
+    * EXCEPT (submission_date)
+  FROM
+    org_mozilla_firefox_derived.event_types_v1
+  WHERE
+    submission_date = '""",
+  CAST(submission_date AS STRING),
+  "'"
+);
```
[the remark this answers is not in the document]
fbertsch: It would reduce complexity, but also add additional risk. This table is small, and this approach was kept so that we had a valid history we could walk through in case something went wrong. The query that updates this table is _very_ complicated so I wanted to ensure rollbacks and backfills were easy.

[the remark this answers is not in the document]
jklukas: Another way we could express this table is by having it partitioned by day, but each day includes only the new entries. So the incremental query would references `FROM event_types_v1 WHERE submission_date < @submission_date` and return only `new_event_types` rather than `all_event_types` to populate the day's partition. The view would then be a simple `SELECT *` without a filter on date.

That setup avoids needing to generate the view, but provides better characteristics for rewinding.

bigquery_etl/events_daily/query_templates/event_types_v1/init.sql
```diff
@@ -0,0 +1,172 @@
+{{ header }}
```
jklukas, line 1: The content of this `init.sql` is so similar to `query.sql` that I would prefer a single `query.sql` and we can pass in an additional parameter to the template `init: bool`. If `init` is true, we'd include the CREATE OR REPLACE statement, etc. This is the approach used in the glean_baseline query generation machinery.

fbertsch: I agree that this would be ideal, but these differ in pretty substantial ways: e.g. getting the `event_index` for `init` is just a `group by`, but we need to join with the previous results in `query.sql` to drop the ones we already have.

We could probably refactor things to be more amenable to this, but let's leave that as a separate issue. Alongside that refactoring we should also expand test scope.

sql/moz-fx-data-shared-prod/org_mozilla_firefox/event_types_update_v1/query.sql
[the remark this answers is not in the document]
jklukas: Then I'd advocate for turning the existing `event_types_v1` into a partitioned `event_types_history_v1` and then `event_types_v1` be an unpartitioned table that is rewritten every day. This would avoid having this surprising `EXECUTE IMMEDIATE CONCAT` query.sql that doesn't actually populate a table.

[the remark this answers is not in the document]
fbertsch: Hmm, then we'd end up with 3 queries: `event_types_history_v1`, `event_types_v1`, and `event_types` view. That adds complexity in a different way.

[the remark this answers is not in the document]
fbertsch: We may be able to make this a query, rather than an `EXECUTE IMMEDIATE`. I'm trying to recall the reason we used that formation.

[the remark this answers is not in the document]
jklukas: I assume a view definition can't contain a query parameter, so the EXECUTE IMMEDIATE is necessary to inject the query param into the view as a literal date.

> Hmm, then we'd end up with 3 queries: event_types_history_v1, event_types_v1, and event_types view. That adds complexity in a different way.

Yes, it involves two table + a view rather than one table + a view, but in the current formulation this PR contains a pseudo-query that appears under a directory path that looks like it should produce a table called `event_types_update_v1` but in fact no such table will ever exists. I would argue that this is corrupting the intent of the directory structure in order to plug in to the DAG generation machinery.

At the very least, this file should be changed to be called `script.sql` rather than `query.sql` to give some hint that it's not actually running a query that populates a table.

But again, I would prefer to introduce another table in the chain if that gets us to a state that's more obviously comprehensible for future developers.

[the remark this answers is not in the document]
fbertsch: Nope, that does not work: `Error in query string: Error processing job 'frank-sandbox:bqjob_r1c34be6c2896468c_00000175be402170_1': Query error: Query parameters cannot be used inside SQL view bodies at [11:21]`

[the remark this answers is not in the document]
fbertsch: Okay, given the current options I do think it's wise to create `events_daily_v1` as a materialized view on the most recent day -- I did hate the `EXECUTE IMMEDIATE`. Let me update the PR.

fbertsch:
@jklukas ready for final review - events_daily view is failing for telemetry, but that is expected since the underlying table doesn't exist yet.

sql/moz-fx-data-shared-prod/telemetry_derived/event_types_v1/query.sql
```diff
@@ -0,0 +1,259 @@
+-- Generated by /Users/frankbertsch/repos/bigquery-etl/bigquery_etl/events_daily
```
fbertsch, line 1: Hah, probably not exactly the path we were looking for

bigquery_etl/events_daily/generate_queries.py
```diff
@@ -0,0 +1,153 @@
...
+
+        if fpath.suffix == ".sql":
+            text = reformat(text, trailing_newline=True)
+
+        (write_path / self.name).write_text(text)
+
+
+@dataclass
+class TemplatedDir:
+    """A directory of templates, which will be generated per the templating.yaml."""
```
jklukas, line 57:
```suggestion
    """A directory of templates, which will be rendered per the templating.yaml."""
```

bigquery_etl/events_daily/generate_queries.py
```diff
@@ -0,0 +1,153 @@
...
+
+    name: str
+    path: Path
+    env: Optional[Environment] = None
+
+    def generate(self, write_path, dataset=None):
+        """Generate this TemplatedDir at write_path for the specified dataset."""
```
jklukas, line 64:
```suggestion
        """Render this TemplatedDir to write_path for the specified dataset."""
```

bigquery_etl/events_daily/query_templates/event_types/templating.yaml
```diff
@@ -0,0 +1,4 @@
+org_mozilla_firefox:
+  app_id: org_mozilla_firefox
+telemetry:
+  app_id: telemetry
```
jklukas, line 4: Can you add a comment here about why we set `app_id: telemetry`? This feels like a made-up value to satisfy some constraint, so needs explanation.

sql/moz-fx-data-shared-prod/org_mozilla_firefox/event_types/view.sql
```diff
@@ -0,0 +1,7 @@
+CREATE OR REPLACE VIEW
```
jklukas, line 1: I would love for _every_ generated file to include a comment that indicates it's been generated by a specific script, including all .sql variants and the metadata.yaml.

jklukas (approved):
r+wc

I commented on a few remaining documentation/commentary items that could have a significant impact to ease of use for those interacting with the repo.

PR #1514, opened 2020-11-03
Update options for script/run_multipart_query

relud:
Allow multiple schema update options, make `--destination_table` required, fix `TypeError: argument of type 'NoneType' is not iterable` when `--dataset_id` is missing, and use `get_temporary_table` from `bigquery_etl.util.temp_table`

### DIFF
--- a/bigquery_etl/run_multipart_query.py
+++ b/bigquery_etl/run_multipart_query.py
@@ -13,10 +13,11 @@ The query files must be in the same directory and all be prefixed with `part`.
 
 from argparse import ArgumentParser
 from multiprocessing.pool import ThreadPool
-from uuid import uuid4
 import os.path
 
 from google.cloud import bigquery
+
+from .util.temp_table import get_temporary_table
 
 
 def dirpath(string):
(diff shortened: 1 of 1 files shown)


bigquery_etl/run_multipart_query.py
```diff
@@ -90,23 +93,16 @@ def dirpath(string):
 )
 parser.add_argument(
     "--schema_update_option",
+    action="append",
+    choices=[
+        bigquery.SchemaUpdateOption.ALLOW_FIELD_ADDITION,
+        bigquery.SchemaUpdateOption.ALLOW_FIELD_RELAXATION,
+    ],
+    default=[],
+    dest="schema_update_options",
```
scholtzan, line 102: I'm wondering if we should rename `schema_update_option` to `schema_update_options` here so that it is consistent with the parameter naming in `bigquery.QueryJobConfig`. Doing that would require updating telemetry-airflow though.

scholtzan (approved):

bigquery_etl/run_multipart_query.py
relud: I think we should keep it as-is to match the `bq` cli tool.

Issue #1517, opened 2020-11-04
Bug 1672992 Fix multipart query behavior when not allowing schema updates


jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1672992

ETL is unstuck for today, but I expect main_summary to fail again tomorrow. There is a logic fix needed on top of the recent changes to schema update option for multipart queries. See https://github.com/mozilla/bigquery-etl/pull/1514

cc @relud @scholtzan

PR #1518, opened 2020-11-04
[Bug 1675293] Fix schema_update_option in main_summary

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/1517

### DIFF
--- a/bigquery_etl/run_multipart_query.py
+++ b/bigquery_etl/run_multipart_query.py
@@ -97,6 +97,11 @@ parser.add_argument(
     choices=[
         bigquery.SchemaUpdateOption.ALLOW_FIELD_ADDITION,
         bigquery.SchemaUpdateOption.ALLOW_FIELD_RELAXATION,
+        # Airflow passes an empty string when the field addition date doesn't
+        # match the run date.
+        # See https://github.com/mozilla/telemetry-airflow/blob/
+        # e49fa7e6b3f5ec562dd248d257770c2303cf0cba/dags/utils/gcp.py#L515
+        "",
     ],
     default=[],
     dest="schema_update_options",


bigquery_etl/run_multipart_query.py
```diff
@@ -97,6 +97,7 @@ def dirpath(string):
     choices=[
         bigquery.SchemaUpdateOption.ALLOW_FIELD_ADDITION,
         bigquery.SchemaUpdateOption.ALLOW_FIELD_RELAXATION,
+        "",
```
scholtzan, line 100: [telemetry-airflow passes an empty string here](https://github.com/mozilla/telemetry-airflow/blob/03ee091cdba5cb1fcb0a6aad3e62483d2bfbb3ee/dags/utils/gcp.py#L511)

jklukas: This would mean that we end up with `args.schema_update_options` being `[""]` which is passed to a `bigquery.QueryJobConfig` below. Do we know the behavior of that value in a job config?

scholtzan: Based on the tests I ran locally, it would just ignore schema update options. But maybe the better solution here is to rewrite the part in telemetry-airflow where the schema_update_option is added as a parameter and not passing an empty string there.

jklukas: Given that fixing this on the Airflow side looks untenable, I think I'd be fine with this solution, with commentary indicating why we need to accept `""`.

jklukas (approved):

bigquery_etl/run_multipart_query.py
scholtzan: Added a comment. I'll leave this open for now until @relud had a chance to take a look, maybe there is a more elegant solution here.

jklukas:
We should definitely rerun a day of main_summary after this is merged and built, to verify that this behaves as expected. I'd recommend we target 2020-11-01 which is a Sunday and thus lower volume.

bigquery_etl/run_multipart_query.py
relud: we could remove the empty string from the list in `main()`, given that it works with empty string in the list, this seems fine as-is

relud (approved):

scholtzan:
I successfully reran main_summary

PR #482, opened 2019-11-06
Clients daily main WIP

washort:
obviously has a bit left to do but I got a few fields converted.

PR #1519, opened 2020-11-04
Produce clients_daily from main_v4

### LINKED ISSUE #1516: Produce clients_daily directly from main_v4 rather than main_summary
We have deprecated `main_summary` and generally are not adding new fields there, but `clients_daily` still depends on it. One major reason I'm aware of is that `main_summary` contains more history than `main_v4`, allows `clients_daily` to stretch back to 2016.

But are there other reasons? It would be nice to simplify the ETL structure and remove `main_summary` as a failure mode for downstream desktop ETL. Could we consider "archiving" the current clients_daily query for large-scale backfill scenarios, and otherwise replace the query with one on top of `main_v4` directly?

cc @relud Who I expect has the most context here.

relud:
fixes #1516
supersedes #482


relud:
doing this in a single query causes "query too complex", so I created a query that is a subset of main summary, and called it "clients summary".

clients summary runs in a single query without using `script/run_multipart_query`, and also provides a path forward for updating clients_daily after deprecating main summary.

side-note: I don't love the name "clients summary", I just couldn't think of anything better. suggestions welcome.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_summary_v1/query.sql
```diff
@@ -0,0 +1,352 @@
...
+    environment.settings.attribution.source,
+    environment.settings.attribution.medium,
+    environment.settings.attribution.campaign,
+    environment.settings.attribution.content
+  ) AS attribution,
+  environment.settings.sandbox.effective_content_process_level AS sandbox_effective_content_process_level,
+  payload.info.timezone_offset,
+  udf.extract_histogram_sum(
+    mozfun.map.get_key(payload.keyed_histograms.subprocess_crashes_with_dump, 'pluginhang')
+  ) AS plugin_hangs,
```
jklukas, line 123: Have you tried collecting all these `extract_histogram_sum/get_key` calls into a single call on top of an array like count histograms? In this case, I think we would need to make an array of STRUCT<histogram, key>.

jklukas: I'm basically wondering if we could still have a path to avoiding clients_summary by fully buying into the strategy of packing arrays for all UDF calls.

relud: > I'm basically wondering if we could still have a path to avoiding clients_summary by fully buying into the strategy of packing arrays for all UDF calls.

I did attempt this, and it still exceeded query complexity limits

jklukas:
I'm concerned that we may overall be adding complexity here by introducing an additional table into the mix that's very main_summary-esque. Perhaps we'd be better off sticking with main_summary. I'm hoping to think some more about this today in the context of what the overall workflow for opting in new metrics to clients_daily can look like.

jklukas:
I think we should open a case with GCP again about this and see if there are any settings they might be able to tweak that could get us over the hump here.

jklukas:
I played with this a bit today, and I think we are _so close_ to being able to do this all in a single query. A few things I changed:

- Replace all `udf.foo` invocations with `mozfun` equivalents wherever possible, which avoids an extra wrapper udf invocation
- Remove the overactive subquery, instead putting a `HAVING pings_aggregated_by_this_row < 200000` clause on the final query
- Factor out `udf.extract_histogram_sum(mozfun.map.get_key` to an array:
```
    ARRAY(
      SELECT AS STRUCT
        udf.extract_histogram_sum(mozfun.map.get_key(histogram, key)) AS v
      FROM
        UNNEST(
          [
            STRUCT(
              payload.keyed_histograms.subprocess_crashes_with_dump AS histogram,
              'pluginhang' AS key
            ),
            STRUCT(payload.keyed_histograms.subprocess_abnormal_abort, 'plugin'),
            STRUCT(payload.keyed_histograms.subprocess_abnormal_abort, 'content'),
            STRUCT(payload.keyed_histograms.subprocess_abnormal_abort, 'gmplugin'),
            STRUCT(payload.keyed_histograms.subprocess_crashes_with_dump, 'plugin'),
            STRUCT(payload.keyed_histograms.subprocess_crashes_with_dump, 'content'),
            STRUCT(payload.keyed_histograms.subprocess_crashes_with_dump, 'gmplugin'),
            STRUCT(payload.keyed_histograms.process_crash_submit_attempt, 'main-crash'),
            STRUCT(payload.keyed_histograms.process_crash_submit_attempt, 'content-crash'),
            STRUCT(payload.keyed_histograms.process_crash_submit_attempt, 'plugin-crash'),
            STRUCT(payload.keyed_histograms.process_crash_submit_success, 'main-crash'),
            STRUCT(payload.keyed_histograms.process_crash_submit_success, 'content-crash'),
            STRUCT(payload.keyed_histograms.process_crash_submit_success, 'plugin-crash'),
            STRUCT(payload.keyed_histograms.subprocess_kill_hard, 'ShutDownKill')
          ]
        )
    ) AS hist_key_sums,

...

  hist_key_sums[OFFSET(0)] AS plugin_hangs,
  hist_key_sums[OFFSET(1)] AS aborts_plugin,
  hist_key_sums[OFFSET(2)] AS aborts_content,
  hist_key_sums[OFFSET(3)] AS aborts_gmplugin,
  hist_key_sums[OFFSET(4)] AS crashes_detected_plugin,
  hist_key_sums[OFFSET(5)] AS crashes_detected_content,
  hist_key_sums[OFFSET(6)] AS crashes_detected_gmplugin,
  hist_key_sums[OFFSET(7)] AS crash_submit_attempt_main,
  hist_key_sums[OFFSET(8)] AS crash_submit_attempt_content,
  hist_key_sums[OFFSET(9)] AS crash_submit_attempt_plugin,
  hist_key_sums[OFFSET(10)] AS crash_submit_success_main,
  hist_key_sums[OFFSET(11)] AS crash_submit_success_content,
  hist_key_sums[OFFSET(12)] AS crash_submit_success_plugin,
  hist_key_sums[OFFSET(13)] AS shutdown_kill,

```

With that, I was at a point where I only had to comment out a single field to get it over the line and be able to dry run. I'm optimistic that there are a few more optimizations we can make to UDFs to get this to a point where it's stable and we have headroom to add more fields.

The final query I got to is captured in https://gist.github.com/jklukas/64bc0622ab63bd47fb43e45f0769a79c

jklukas:
Also, we can hit `main_v4` directly rather than the `main` view, which should avoid several subqueries.

jklukas:
> Remove the overactive subquery, instead putting a HAVING pings_aggregated_by_this_row < 200000 clause on the final query

I'm realizing now that perhaps this won't have the desired effect. The HAVING clause is only applied after aggregations have happened. I assume the whole `overactive` filter exists to prevent aggregates from overflowing, and relying on the HAVING filter won't save us in that case.

jklukas:
I filed [GCP case 25952190](https://console.cloud.google.com/support/cases/detail/25952190?project=moz-fx-data-shared-prod) to ask for help in further improving this query.

jklukas:
I tried restoring the `overactive` subquery and join, which seems to have a substantial impact on complexity. But moving to `main_v4` rather than the view appears to overcome this. I published a [new gist](https://gist.github.com/jklukas/7608d32250e42184e1c8fe79070d24ee) that is able to dry-run, and I believe hasn't changed any of the logic. We'll definitely need more validating.

relud:
@jklukas I had this working, but after rebasing to include #1549 the complexity is too high *only during testing*, presumably because temporary UDFs have a higher complexity cost than persisent UDFs.

relud:
> after rebasing to include #1549 the complexity is too high _only during testing_

fixed this by using arrays to only have one call to `mozfun.map.sum`

jklukas (commented):
The logic here looks good. I'm holding off on approving until we have the test plan and results documented that verifies the output is unchanged on real main ping data.

relud:
> I'm holding off on approving until we have the test plan and results documented that verifies the output is unchanged on real main ping data.

my plan is to run this on a day of data, json serialize the results and ensure they are the same. will that suffice, or did you want something more thorough than that?

jklukas:
> my plan is to run this on a day of data, json serialize the results and ensure they are the same. will that suffice, or did you want something more thorough than that?

That sounds great.

relud:
I have validated that the updated query is equivalent to the original, via the scripts in [this gist](https://gist.github.com/relud/ae0cb924d25fe35ec7b083f1eb3b10e2)

notable results differences were that floats can't be compared directly, map sums didn't have deterministic order, and `ANY_VALUE` isn't deterministic

jklukas (approved):
Thanks so much for validating and providing the documentation!

The final thing I want to see here before merging is clear documentation about the change from main_summary to main_v4 which specifies the first date for which the new logic is valid (and thus the limit for backfills using this methodology) and a link to the final version of the main_summary-based query so that it's easy to find should we need to do a longer backfill. This can probably be a comment at the top of the clients_daily query file.

jklukas:
We should be sure to celebrate this once this is merged! It's removing a significant impediment to evolution of the clients_daily ETL stream.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
```diff
@@ -1,436 +1,872 @@
...
+  environment.addons.active_addons[].foreign_install
+  environment.addons.active_addons[].user_disabled
+  environment.addons.active_addons[].version
+  payload.simple_measurements.active_ticks
+  payload.simple_measurements.first_paint
+
+To backfill partitions for 2019-11-22 and earlier, use the last version of this query that read from main_summary_v4:
+https://github.com/mozilla/bigquery-etl/blob/813a485/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
+*/
+WITH base AS (
```
relud, line 13: > The final thing I want to see here before merging is clear documentation about the change from main_summary to main_v4 which specifies the first date for which the new logic is valid (and thus the limit for backfills using this methodology) and a link to the final version of the main_summary-based query so that it's easy to find should we need to do a longer backfill. This can probably be a comment at the top of the clients_daily query file.

how's this?

jklukas (approved):
The commentary looks great! 🚢

Issue #1520, opened 2020-11-04
Add dry run support for different projects


Ben-Wu:
There's already a todo in the dryrun script for adding support for different dryrun endpoints depending on project so I'm just copying that here.

Another option is to reuse the same endpoint and give the service account permissions to read the other projects and add a project argument to the cloud function.  I don't see either option as better than the other.

PR #1524, opened 2020-11-05
Add derived GA tables for www site

Ben-Wu:
cc @gkabbz 

Ported from https://github.com/mozilla/marketing-analytics/tree/master/Airflow (the siteMetrics and eventsByPage scripts)

Dryrun skip was removed for these tables so they can dryrun and referenced tables can be used for dag generation.  This means table references need to include project id.  I created issue https://github.com/mozilla/bigquery-etl/issues/1520 as a potential fix.

I verified that these match the existing tables.

airflow dag:
Screen Shot 2020-11-04 at 1 38 37 PM


sql/moz-fx-data-marketing-prod/ga_derived/www_site_events_metrics_v1/query.sql
```diff
@@ -0,0 +1,54 @@
...
+  `language`,
+  browser,
+  country,
+  source,
+  medium,
+  campaign,
+  ad_content,
+  page_name
+HAVING
+  event_category IS NOT NULL
```
jklukas, line 54: `HAVING` clauses generally have worse performance characteristics compared to `WHERE`. Is there a particular reason for this clause as opposed to adding this condition to `WHERE`?

jklukas (approved):
r+wc

sql/moz-fx-data-marketing-prod/ga_derived/www_site_events_metrics_v1/query.sql
Ben-Wu: This looks like an artifact of the original query that I rewrote.  I'll take it out

PR #1525, opened 2020-11-05
add Baidu to normalized engine udf

xluo-ds:
adding 'Baidu' as an engine to the normalized-engine udf
waiting on Greg to confirm the code is right.


jklukas:
This PR is no longer relevant. `normalize_search_engine` in this repo is now a stub.

PR #1526, opened 2020-11-05
Add total_user filter on glam exports

acmiyaguchi:
This filters probes where there are less than 10 users in the dimension.


robhudson (approved):
Thanks!

PR #1532, opened 2020-11-06
Remove udf_js.glean_percentile

acmiyaguchi:
As per :whd while running publish-udf.

```
Function not found: glam.percentile at [7:3]
```

This function is now deprecated with the `mozfun.glam.percentile` function, so there's no need to keep it around. I'm not sure why the tests were allowing this to pass, though.


Ben-Wu (approved):

Ben-Wu:
My guess is that the pytest plugin treats all udfs as the same project so the mozfun udf is recognized in the shared-prod test

PR #1536, opened 2020-11-09
Support python script query scheduling

scholtzan:
This adds support to schedule `query.py` files.

Next step will be to migrate the [monitoring](https://github.com/mozilla/telemetry-airflow/blob/master/dags/monitoring.py) and other telemetry-airflow DAGs to use this instead.

### DIFF
--- a/bigquery_etl/cli/dag.py
+++ b/bigquery_etl/cli/dag.py
@@ -196,7 +196,7 @@ def remove(name, dags_config, output_dir):
         sys.exit(1)
 
     for task in dag_tbr.tasks:
-        metadata = Metadata.of_sql_file(task.query_file)
+        metadata = Metadata.of_query_file(task.query_file)
         sql_path = Path(os.path.dirname(task.query_file))
         metadata.scheduling = {}
         metadata_file = sql_path / METADATA_FILE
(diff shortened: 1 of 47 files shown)


bigquery_etl/cli/dag.py
```diff
@@ -196,7 +196,7 @@ def remove(name, dags_config, output_dir):
         sys.exit(1)
 
     for task in dag_tbr.tasks:
-        metadata = Metadata.of_sql_file(task.query_file)
+        metadata = Metadata.of_query_file(task.query_file)
```
scholtzan, line 199: Renamed this since Python scripts can be scheduled which aren't SQL

bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -13,44 +13,57 @@ default_args = {{
...
-        {%+ endif -%}
-        {%+ if task.priority -%}
-        priority_weight={{ task.priority }},
-        {%+ endif -%}
-        {%+ if task.allow_field_addition_on_date -%}
-        allow_field_addition_on_date={{ task.allow_field_addition_on_date | format_repr }},
-        {%+ endif -%}
-        dag=dag,
-    )
+    {% if not task.is_python_script -%}
```
jklukas, line 16: I'd prefer to remove the `not` here and put the python script case first to make this easier to read.

jklukas (approved):
r+wc

This looks pretty clean 👍

PR #1540, opened 2020-11-10
Rename GA standardized country names table

Ben-Wu:
Spoke to George and this table shouldn't be GA specific so changing now before there are other uses.  There's also a `sql_file_path` fix here.


scholtzan (approved):

sql/moz-fx-data-shared-prod/static/standardized_country_names/description.txt
```diff
@@ -0,0 +1 @@
+One-to-many mapping from country codes or non-standard names to a standardized country name.
```
jklukas, line 1: I think there needs to be some documentation here about scope for this table. It includes several custom values like "Worldwide". If the scope here is not for Google Analytics, then what _is_ the scope? Where do these custom non-country values come from?

How can a user know how to choose between using this table and the country_names_v1 table?

Ben-Wu: The list was made by @gkabbz and there are values in it that aren't in google analytics.  From what I know it's intended to be used with any third-party data that has country values.  I don't see worldwide is even in GA so I'm assuming values like that have some use with another data source.

We also had the concern with ambiguous naming.  `third_party_standardized_country_names` may be a more descriptive name.

jklukas: I don't want to block merging this for now if it's going to fix some existing issues. I can live with `standardized_country_names` as the name, but I'd like to see a follow-up where @gkabbz adds some commentary in this description about what these additional non-country entries are and how they are expected to be used.

jklukas: And I think other PRs are failing CI right now because the ga_standardized_country_names table has already been removed.

Ben-Wu: Ok I added `ga_standardized_country_names` back again for now.  George said that this list was manually created over time working with a bunch of different data sources so I think third party is much more descriptive.

PR #1541, opened 2020-11-10
Schedule monitoring tasks

scholtzan:
Depends on https://github.com/mozilla/telemetry-airflow/pull/1181

### DIFF
--- a/sql/moz-fx-data-shared-prod/monitoring/telemetry_distinct_docids_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/monitoring/telemetry_distinct_docids_v1/metadata.yaml
@@ -1,0 +1,25 @@
+---
+friendly_name: Telemetry Distinct Document IDs
+description: >
+  Compares number of document IDs in telemetry decoded, live,
+  and stable tables.
+owners:
+  - bewu@mozilla.com
+labels:
+  schedule: daily
+  incremental: true
+scheduling:
+  dag_name: bqetl_monitoring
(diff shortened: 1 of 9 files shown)


sql/moz-fx-data-shared-prod/monitoring/telemetry_distinct_docids_v1/metadata.yaml
```diff
@@ -0,0 +1,25 @@
...
+       'telemetry_telemetry__*']
+    - ['moz-fx-data-shared-prod', 'telemetry_stable', '*']
+    - ['moz-fx-data-shared-prod', 'telemetry_live', '*']
+  depends_on:
+    - task_id: copy_deduplicate_main_ping
+      dag_name: copy_deduplicate
+      execution_delta: 1h
+    - task_id: copy_deduplicate_all
+      dag_name: copy_deduplicate
+      execution_delta: 1h
```
jklukas, line 25: Are both `referenced_tables` and `depends_in` necessary? Do we get any benefit from listing `payload_bytes_decoded` here considering it's not populated by Airflow?

In general for this PR, would it be possible to use `referenced_tables` everywhere instead of `depends_on`, or is there a reason the Airflow generation machinery does not support `referenced_tables` for python scripts?

In general, I find `referenced_tables` very nice because it prevents us having to manually track execution_delta. If it's possible to avoid `depends_on` here, let's do it.

jklukas (approved):
r+wc

PR #1560, opened 2020-11-17
Bug 1677609 Add core active fields to clients_last_seen

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1677609

This adds just the new underlying bit pattern fields that will need to be
backfilled, and these will be hidden from users initially.
After the backfill is complete, we will update the view to include these
fields along with the various fields derived from them.

We include days_visited_10_uri_bits which was not explicitly requested in
the context of this bug, but was proposed as part of the prototype feature_usage
table (https://github.com/mozilla/bigquery-etl/pull/1193); it may be useful
for future comparisons.


jklukas:
To make the relevant schema updates, I plan to run the init.sql query (without the CREATE TABLE part):

```
SELECT
  CAST(NULL AS INT64) AS days_seen_bits,
  CAST(NULL AS INT64) AS days_visited_1_uri_bits,
  CAST(NULL AS INT64) AS days_visited_5_uri_bits,
  CAST(NULL AS INT64) AS days_visited_10_uri_bits,
  CAST(NULL AS INT64) AS days_had_8_active_ticks_bits,
  CAST(NULL AS INT64) AS days_opened_dev_tools_bits,
  CAST(NULL AS INT64) AS days_interacted_bits,
  CAST(NULL AS INT64) AS days_created_profile_bits,
  ARRAY<STRUCT<experiment STRING, branch STRING, bits INT64>>[] AS days_seen_in_experiment,
  -- We make sure to delay * until the end so that as new columns are added
  -- to clients_daily, we can add those columns in the same order to the end
  -- of this schema, which may be necessary for the daily join query between
  -- the two tables to validate.
  *
FROM
  clients_daily_v6
WHERE
  -- Output empty table and read no input rows
  FALSE
```

In the query job, I will set options to allow field addition and to append the results to the table (0 rows will be added due to the WHERE clause here).

relud (approved):

irrationalagent (approved):

jklukas:
Schema is now up to date such that dry runs should succeed. I ran:

```
bq query --project_id moz-fx-data-shared-prod --dataset_id telemetry_derived --nouse_legacy_sql --append_table --destination_table clients_last_seen_v1 --schema_update_option ALLOW_FIELD_ADDITION < init.sql
```

where `init.sql` is the modified version I posted above.

jklukas:
The DAG is updated and I tested that I was able to clear yesterday's run and have it succeed.

PR #1561, opened 2020-11-17
Bug 1677609 Join clients_first_seen into clients_last_seen

jklukas:
Several folks on DS report that they have been getting great value from
clients_first_seen, as the first_seen_date there is a much more stable way
to define new profiles compared to using profile_created_date from pings.

Currently, using first_seen_date requires doing a join between these two tables.
This PR adds that join to the clients_last_seen query itself to make this
workflow more efficient. I'd like to get this merged before we proceed with
the backfill discussed in https://bugzilla.mozilla.org/show_bug.cgi?id=1677609

This change has a few operational implications. Most importantly, it makes
clients_last_seen dependent on clients_first_seen, so those queries can no
longer proceed in parallel. `clients_first_seen` takes on average 10 minutes
to run, so we'll be delaying all ETL downstream of `clients_last_seen` by
about 10 minutes, which seems acceptable. It also adds some mental complexity
to the model.

The extra join does not appear to significantly slow down the
`clients_last_seen` query itself; it scans about 15% more data and consumes
about 15% more slot time.
I expect the performance is dominated by the existing join between
clients_daily and the previous day of clients_last_seen.

Note that this PR intentionally does not update the user-facing view to include these fields. I want them to remain hidden until we perform the backfill. After that, I'll file a follow-up PR to add the fields and derived `days_since_first_seen` to the view.


sql/moz-fx-data-shared-prod/telemetry_derived/clients_last_seen_v1/metadata.yaml
```diff
@@ -3,6 +3,13 @@ friendly_name: Clients Last Seen
 description: >
   Captures history of activity of each client in 28 day
   windows for each submission date.
+
+  Generally, this is direct product of clients_daily and serves to make
```
relud, line 7:
```suggestion
  Generally, this is a direct product of clients_daily and serves to make
```

relud (approved):
r+wc

jklukas:
Note that I plan to merge this after the November break. We're letting a clients_last_seen backfill run over the break to a new table, and that backfill includes this logic. We'll copy the backfill over to the production table as part of updating the schema and merging this PR.

PR #1564, opened 2020-11-19
Bug 1678344 - Add attribution source and ua to onboarding_events_amplitude

ncloudioj:
This fixes bug [1678344](https://bugzilla.mozilla.org/show_bug.cgi?id=1678344).

Note that we don't need to do the backfilling here since attribution is a new collection.


fbertsch (approved):
LGTM!

PR #1565, opened 2020-11-19
Create script to copy ga_sessions tables between projects

Ben-Wu:
Copy between projects so there's some time to allow people to change queries

related https://github.com/mozilla/telemetry-airflow/pull/1189


Ben-Wu:
todo: point derived table queries to copied tables

jklukas:
I'd prefer to see the script here put underneath `scripts/` and then add a DAG in telemetry-airflow that calls the bigquery-etl image with that script as the entrypoint. It's less convenient than tying into the DAG generation machinery here, but it's difficult to infer what's going on in this PR. We have a `query.py` in the `sql` directory structure that doesn't represent a real table, which is surprising. I know this is temporary, but I want to avoid setting a precedent.

Ben-Wu:
Moved to scripts

There's an issue with the copied tables that's causing dry run to fail.  Older tables are missing some columns that the wildcard query tries to access so it fails.  For some reason though, the tables these are copied from are also missing those fields but the query works fine on that dataset.  Trying to figure that out now

Ben-Wu:
Dry run works now.  I manually updated the schema for older tables but I wasn't able to figure out why it was working with the original tables.

sql/moz-fx-data-marketing-prod/ga_derived/blogs_empty_check_v1/query.sql
```diff
@@ -1,6 +1,6 @@
 SELECT
   IF(COUNT(*) = 0, ERROR(CONCAT('No data for given date ', @submission_date)), 1)
 FROM
-  `ga-mozilla-org-prod-001.66602784.ga_sessions_*`
+  `moz-fx-data-marketing-prod.66602784.ga_sessions_*`
```
jklukas, line 4:
```suggestion
  66602784.ga_sessions_*
```

According to style guidelines for this repo, we should avoid fully-qualified table references except where referencing a different project from the project in which the query runs.

jklukas (approved):
r+wc

Thanks for taking the time for the refactor and separate telemetry-airflow DAG.

sql/moz-fx-data-marketing-prod/ga_derived/blogs_empty_check_v1/query.sql
Ben-Wu: This is needed for dryrun since the dryrun function uses shared-prod as the default project.  I made a issue https://github.com/mozilla/bigquery-etl/issues/1520 but I haven't had a chance to address it.

jklukas: I forgot about that issue. Understood.

PR #1566, opened 2020-11-19
Fix mozilla vpn waitlist update logic

relud:
in a join, the table alias is never NULL, only the contents, so using `COALESCE` was causing all rows not updated to have all-null columns.

### DIFF
--- a/sql/moz-fx-data-shared-prod/mozilla_vpn_external/waitlist_v1/init.sql
+++ b/sql/moz-fx-data-shared-prod/mozilla_vpn_external/waitlist_v1/init.sql
@@ -6,7 +6,4 @@ AS
 SELECT
   *
 FROM
-  EXTERNAL_QUERY(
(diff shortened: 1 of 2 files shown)


sql/moz-fx-data-shared-prod/mozilla_vpn_external/waitlist_v1/init.sql
```diff
@@ -8,5 +8,5 @@ SELECT
 FROM
   EXTERNAL_QUERY(
     "moz-fx-guardian-prod-bfc7.us.guardian-sql-prod",
-    "SELECT * FROM vpn_waitlist LIMIT 0;"
+    "SELECT * FROM vpn_waitlist"
```
relud, line 11: initialize the data, not just the schema

acmiyaguchi (approved):

PR #1568, opened 2020-11-19
Add mozfun.hist.string_to_json

scholtzan:
Will be used here: https://github.com/FirefoxGraphics/telemetry/pull/9


sql/mozfun/hist/string_to_json/metadata.yaml
```diff
@@ -0,0 +1,3 @@
+---
+friendly_name: Histogram String to JSON
+description: Convert a histogram string to a full histogram JSON blob.
```
jklukas, line 3:
```suggestion
description: >
  Convert a histogram string (in JSON or compact format) to a full histogram JSON blob.
```

jklukas (approved):

PR #1571, opened 2020-11-20
Add main_1pct view

jklukas:
Discussed with BQ engineering in [support case 25679061](https://console.cloud.google.com/support/cases/detail/25679061?project=moz-fx-data-shared-prod)

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/main_1pct_v1/init.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/main_1pct_v1/init.sql
@@ -1,0 +1,16 @@
+CREATE TABLE IF NOT EXISTS
+  `moz-fx-data-shared-prod.telemetry_derived.main_1pct_v1`
+PARTITION BY
+  DATE(submission_timestamp)
+CLUSTER BY
+  normalized_channel,
+  sample_id
+OPTIONS
+  (require_partition_filter = TRUE, partition_expiration_days = 180)
+AS
+SELECT
+  *
+FROM
+  telemetry_stable.main_v4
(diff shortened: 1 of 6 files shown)


jklukas:
My plan would be to backfill this with 6 months of data, make an announcement to data users, and point to a bug to collect feedback. If it's proving valuable, we can introduce events_1pct as well and further document.

jklukas:
I've created the table and started a backfill. This simple count takes 5 seconds to execute and scans only 4 MB:

```
SELECT
  DATE(submission_timestamp) AS dt,
  COUNT(*)
FROM
  `moz-fx-data-shared-prod.telemetry_derived.main_1pct_v1`
WHERE
  DATE(submission_timestamp) >= '2020-11-24'
  AND subsample_id = 0
GROUP BY
  1
ORDER BY
  1 DESC
```

sql/moz-fx-data-shared-prod/telemetry_derived/main_1pct_v1/init.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE IF NOT EXISTS
+  `moz-fx-data-shared-prod.telemetry_derived.main_1pct_v1`
+PARTITION BY
+  DATE(submission_timestamp)
+CLUSTER BY
+  normalized_channel,
+  sample_id,
+  subsample_id
+OPTIONS
+  (require_partition_filter = TRUE, partition_expiration_days = 180)
```
scholtzan, line 10: Should it be noted in the table description that this table has only the past 6 months of data?

scholtzan (approved):

PR #1573, opened 2020-11-27
Remove filter for search metrics in glam

### LINKED ISSUE #1483: Remove filter for desktop browser search metrics in GLAM
A user of GLAM searched for `browser_searchinit_engine_invalid_webextension` and it was not there. The metric was found in the probe search API so it appeared that GLAM would load it but the GLAM data response was empty.

The reason is that GLAM filters out metrics related to search to avoid exposing those numbers to any employee:
https://github.com/mozilla/bigquery-etl/blob/master/sql/moz-fx-data-shared-prod/telemetry/client_probe_counts_v1/view.sql#L29-L32

Frank shared that the search datasets are available to all employees now and this is no longer required. If GLAM ever does expose data to the general public we can filter out search metrics via the front-end.

Ben-Wu:
(fixes #1483)

For reference: https://bugzilla.mozilla.org/show_bug.cgi?id=1623180

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry/client_probe_counts_v1/view.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry/client_probe_counts_v1/view.sql
@@ -26,7 +26,3 @@ SELECT
   *
 FROM
   all_counts
-WHERE
-  metric != "search_counts"
-  AND metric NOT LIKE "%browser_search%"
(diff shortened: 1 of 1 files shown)


Ben-Wu:
cc @robhudson Do we still want to do this?

acmiyaguchi (approved):

robhudson (approved):

PR #1575, opened 2020-11-27
Add queries and views for fetch marketing spend data

Ben-Wu:
Marketing spend data in moz-fx-data-marketing-prod:fetch

### DIFF
--- a/sql/moz-fx-data-marketing-prod/fetch/spend_alignment_by_campaign_v1/metadata.yaml
+++ b/sql/moz-fx-data-marketing-prod/fetch/spend_alignment_by_campaign_v1/metadata.yaml
@@ -9,5 +9,3 @@ owners:
 scheduling:
   dag_name: bqetl_marketing_fetch
   date_partition_parameter: null
-  referenced_tables: []
-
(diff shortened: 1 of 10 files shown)


sql/moz-fx-data-marketing-prod/fetch/spend_alignment_by_campaign_v1/metadata.yaml
```diff
@@ -0,0 +1,13 @@
...
+friendly_name: Campaign Spend Alignment
+labels:
+  incremental: false
+  schedule: weekly
+owners:
+  - bewu@mozilla.com
+scheduling:
+  dag_name: bqetl_marketing_fetch
+  date_partition_parameter: null
+  referenced_tables: []
```
scholtzan, line 12: It would probably be best to list referenced tables here for correctness. They should still get ignored during dry runs. It's also possible to use wildcards in dataset and table names.

scholtzan (approved):

sql/moz-fx-data-marketing-prod/fetch/spend_alignment_by_campaign_v1/metadata.yaml
Ben-Wu: I'll just remove it since the dry run works fine.  I think I just had it there to speed up testing.

PR #1578, opened 2020-11-30
Bug 1677609 Add first_seen and core active to clients_last_seen view

jklukas:
Follow-up to https://github.com/mozilla/bigquery-etl/pull/1561 now that the backfill is complete.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1677609

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry/clients_last_seen_v1/view.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry/clients_last_seen_v1/view.sql
@@ -1,17 +1,23 @@
 CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.telemetry.clients_last_seen_v1`
 AS
+WITH days_since AS (
+  SELECT
+    DATE_DIFF(submission_date, first_seen_date, DAY) AS days_since_first_seen,
+    DATE_DIFF(submission_date, second_seen_date, DAY) AS days_since_second_seen,
+    mozfun.bits28.days_since_seen(days_seen_bits) AS days_since_seen,
+    mozfun.bits28.days_since_seen(days_visited_1_uri_bits) AS days_since_visited_1_uri,
+    mozfun.bits28.days_since_seen(days_visited_5_uri_bits) AS days_since_visited_5_uri,
+    mozfun.bits28.days_since_seen(days_visited_10_uri_bits) AS days_since_visited_10_uri,
+    mozfun.bits28.days_since_seen(days_had_8_active_ticks_bits) AS days_since_had_8_active_ticks,
+    mozfun.bits28.days_since_seen(days_opened_dev_tools_bits) AS days_since_opened_dev_tools,
+    mozfun.bits28.days_since_seen(days_created_profile_bits) AS days_since_created_profile,
+    mozfun.bits28.days_since_seen(days_interacted_bits) AS days_since_interacted,
+    *
+  FROM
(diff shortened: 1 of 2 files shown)


sql/moz-fx-data-shared-prod/telemetry/clients_last_seen_v1/view.sql
```diff
@@ -61,6 +66,46 @@ SELECT
...
+    BIT_COUNT(days_visited_1_uri_bits) >= 21
+  THEN
+    'core_user'
+  ELSE
+    'other'
+  END
+  AS activity_segments_v1,
+  days_since_created_profile = 6
+  AND BIT_COUNT(
+    days_seen_bits & udf.bits28_from_string('0000000000000000000001111111')
```
jklukas, line 99: We should probably convert these UDF calls to integer literals in hex notation to avoid the UDF overhead, and add commentary about the UDF equivalent.

sql/moz-fx-data-shared-prod/telemetry/clients_last_seen_v1/view.sql
```diff
@@ -61,6 +66,46 @@ SELECT
...
+    'regular_user'
+  WHEN
+    BIT_COUNT(days_visited_1_uri_bits) >= 21
+  THEN
+    'core_user'
+  ELSE
+    'other'
+  END
+  AS activity_segments_v1,
+  days_since_created_profile = 6
```
jklukas, line 97: @irrationalagent Now that first_seen_date is available here, do you want to recast these to use that rather than `days_since_created_profile`? Or should we stick with this for now and consider updating later?

irrationalagent: How much pain would it be to change later? IMO we should probably change, but I know people rely on this field and I'm not sure if there are use cases that would break by us changing without consulting first.

What's the pain in having both `days_since_created_profile` and `days_since_first_seen`? I believe having both is probably preferable even if it causes some confusion (though I think pretty much all DS already knows the difference)

jklukas: I wasn't clear in the scope of what I was proposing here. I definitely would want to keep both days_since_created_profile and days_since_first_seen. `days_since_created_profile` would continue to be based on profile_creation_date.

My intent was to suggest that we might want to define the new `new_profile_7_day_activated_v1`, etc. fields in terms of `days_since_first_seen` rather than `days_since_created_profile`.

irrationalagent: Oops, I see. We tend to enroll users in experiments using PCD, so I think we would probably want to keep it that way for now to avoid any mismatch there.

jklukas:
@irrationalagent - If this looks good to you, please leave an approving review and I'll get this merged and published.

sql/moz-fx-data-shared-prod/telemetry/clients_last_seen_v1/view.sql
irrationalagent: @jklukas After canvassing DS a bit, I retract my previous comment. Let's go ahead and use `first_seen_date` when calculating `new_profile_7_day_activated_v1`. Sorry for the noise

jklukas: Done.

irrationalagent (approved):

PR #1579, opened 2020-11-30
Categorize VPN user attribution

felixlawrence:
At present, there are two problems with VPN user attribution:
1) The attribution is stored as a JSON blob, rather than our standard array of structs (homebrew map)
2) The raw attribution data needs categorization in order to be useful for many tasks.

We should solve both these problems; I don't know the trick to solve problem (1), so I pretended `JSON_EXTRACT()` did the trick. This code solves problem (2), albeit with that magical step. An alternative approach that actually works, using `JSON_EXTRACT()`, is taken in [this query](https://sql.telemetry.mozilla.org/queries/76554/source#190893).

@relud: feel free to treat this PR as a spec, bin it, and solve problems (1) and (2) by separate means. Or if appropriate, swap out the call to `JSON_EXTRACT()` with whatever UDF black magic that we typically use?


sql/moz-fx-data-shared-prod/mozilla_vpn_derived/users_v1/query.sql
```diff
@@ -1,6 +1,133 @@
+WITH raw AS (
+  SELECT
+    fxa_uid,
+    created_at,
+    JSON_EXTRACT(attribution, '$') AS attribution,
```
relud, line 5: pending #1585 this can be decoded via:
```suggestion
    mozfun.json.extract_string_map(attribution) AS attribution,
```

relud (approved):
this lgtm, pending #1585 for a UDF to extract a map from JSON

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/users_v1/query.sql
relud: committed this, to see if CI passes

felixlawrence: Ah, I've got to get format-sql working first...

relud:
after this is merged I'll need to manually run it to update the table schema

PR #1585, opened 2020-12-01
Define UDF for extracting a string map from JSON

relud:
This is needed to decode VPN attribution in #1579

### DIFF
--- a/sql/mozfun/json/extract_string_map/metadata.yaml
+++ b/sql/mozfun/json/extract_string_map/metadata.yaml
@@ -1,0 +1,5 @@
+---
+friendly_name: JSON Extract String Map
+description: >-
+  Returns an array of key/value structs from a string representing a JSON map.
+  Non-string non-null values are encoded as json.
--- a/bigquery_etl/view/publish_views.py
+++ b/bigquery_etl/view/publish_views.py
@@ -58,6 +58,11 @@ def _process_file(client, args, filepath):
             # with a letter and may not end with a dash. For more information see also
             # https://github.com/mozilla/bigquery-etl/pull/1427#issuecomment-707376291
             project_id = target_view_orig.replace("`", "").rsplit(".", 2)[0]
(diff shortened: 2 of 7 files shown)


sql/mozfun/json/extract_string_map/metadata.yaml
```diff
@@ -0,0 +1,5 @@
+---
+friendly_name: JSON Extract String Map
+description: >-
+  Returns an array of key/value structs from a string representing a JSON map.
+  Non-string non-null values are encoded as json.
```
jklukas, line 5: This is the first JS UDF in mozfun other than glam-specific UDFs. Minimally, I'd like the description here to specify that this is a JS UDF and that there are performance implications. Perhaps we should somehow encode the fact that this is JS in the name of the function or namespace.

Alternately, if the VPN use case is simple, we may be able to make some assumptions about structure and achieve what we need using some string substitutions without leaving SQL.

acmiyaguchi (approved):
Aside from the points that Jeff brought up, this looks like a useful interface to have in the toolbox.

sql/mozfun/json/extract_string_map/metadata.yaml
relud: > Minimally, I'd like the description here to specify that this is a JS UDF and that there are performance implications.

```suggestion
  BigQuery Standard SQL JSON functions are insufficient to implement this function, so JS is being used and it may not perform well with large or numerous inputs.
  Non-string non-null values are encoded as json.
```

>  Perhaps we should somehow encode the fact that this is JS in the name of the function or namespace.

I think it would be appropriate and effective to rename this to `mozfun.json.js_extract_string_map`.

> Alternately, if the VPN use case is simple, we may be able to make some assumptions about structure and achieve what we need using some string substitutions without leaving SQL.

I'm pretty sure the VPN data isn't so kind as to disallow `"` and `,` within the recorded header values.

jklukas (approved):

PR #1591, opened 2020-12-02
Fix #1587 - fix inconsistent range_min and range_max in bucket counts

### LINKED ISSUE #1587: Fix inconsistent range_min and range_max in scalar bucket counts
https://github.com/mozilla/bigquery-etl/blob/4234c400402b5f838ed926c63b91dfcd8dbfb2a2/tests/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__scalar_bucket_counts_v1/test_minimal/data.py#L50

This test has unexpected results. This is due to a de-duplication process that is resulting in double counting (double counting so the value is 3 orders of magnitude larger...).

https://github.com/mozilla/bigquery-etl/blob/4234c400402b5f838ed926c63b91dfcd8dbfb2a2/bigquery_etl/glam/templates/scalar_bucket_counts_v1.sql#L17-L30

acmiyaguchi:
Followup to #1488. This fixes #1587 by removing a "deduplication" step that was actually just double counting values. This is likely a reason that the number of users was inflated during spot checks of histograms.


tests/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__scalar_probe_counts_v1/test_minimal/data.py
```diff
@@ -32,9 +32,9 @@
     {
         "agg_type": "histogram",
         "aggregates": [
-            {"key": "1.00", "value": 0.16666666666666666},
-            {"key": "2.00", "value": 0.6666666666666666},
-            {"key": "4.00", "value": 0.16666666666666666},
+            {"key": "1.00", "value": 0.166_666_666_666_666_66},
```
Ben-Wu, line 35: Nice.  I didn't even know you could do that

Ben-Wu (approved):

tests/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__scalar_probe_counts_v1/test_minimal/data.py
acmiyaguchi: It was strange seeing black do this, but it's been around for a while. I vaguely remember seeing this sometime last year.

PR #1592, opened 2020-12-03
[Bug 1646825] Compact histograms encoding support in queries

scholtzan:
This is to unblock https://bugzilla.mozilla.org/show_bug.cgi?id=1646825


sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_histogram_aggregates_parent_v1/query.sql
```diff
@@ -1,30 +1,18 @@
...
-          var z = new Array();
-          node = JSON.parse(y);
-          for (let [key, value] of Object.entries(node)) {
-            z.push(`${key}:${value}`);
-          }
-          return z;
-        ''';
-
+-- Query generated by: sql/telemetry_derived/clients_daily_histogram_aggregates_v1.sql.py
 CREATE TEMP FUNCTION udf_aggregate_json_sum(histograms ARRAY<STRING>) AS (
```
scholtzan, line 2: I'm wondering if this UDF is worth having a mozfun UDF for? Something like `mozfun.hist.aggregate_values_sum`?

jklukas: Sounds reasonable, although I would expect the output to have keys of type INT64 rather than STRING, matching how the `values` array is returned by `mozfun.hist.extract`. If we promote this to mozfun, I'd like to make sure the signature is consistent with the rest of `mozfun.hist`, which would require a bit of refactoring in these queries.

sql/moz-fx-data-shared-prod/telemetry_derived/main_summary_v4/part1.sql
```diff
@@ -355,7 +355,10 @@ SELECT
   ) AS events,
   -- bug 1339655
   SAFE_CAST(
-    JSON_EXTRACT_SCALAR(payload.histograms.ssl_handshake_result, '$.values.0') AS INT64
+    mozfun.map.get_key(
+      mozfun.hist.extract(payload.histograms.ssl_handshake_result).values,
```
jklukas, line 359: I'm a little nervous about adding several new calls to `mozfun.hist.extract` and `mozfun.map.get_key` here, which are going to increase query complexity. If this query already had a preliminary CTE, I would suggest doing a single `mozfun.hist.extract` per histogram there, and reusing the value here. This is probably fine as long as dry runs can succeed.

jklukas (approved):
r+wc

sql/moz-fx-data-shared-prod/telemetry_derived/main_summary_v4/part1.sql
```diff
@@ -45,6 +45,25 @@ CREATE TEMP FUNCTION extract_popup_notification_stats(h STRING) AS (
...
+    mozfun.hist.extract(
+      payload.histograms.plugins_notification_user_action
+    ) AS plugins_notification_user_action
+  FROM
+    `moz-fx-data-shared-prod.telemetry.main`
+  WHERE
+    DATE(submission_timestamp) = @submission_timestamp
+    AND normalized_app_name = 'Firefox'
+    AND document_id IS NOT NULL
+)
```
scholtzan, line 66: Would this address the complexity concern?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_histogram_aggregates_parent_v1/query.sql
scholtzan: I'll do this in a follow-up PR

sql/moz-fx-data-shared-prod/telemetry_derived/main_summary_v4/part1.sql
```diff
@@ -45,6 +45,25 @@ CREATE TEMP FUNCTION extract_popup_notification_stats(h STRING) AS (
...
+    document_id,
+    client_id,
+    sample_id,
+    mozfun.hist.extract(payload.histograms.ssl_handshake_result) AS ssl_handshake_result,
+    mozfun.hist.extract(
+      payload.histograms.plugins_notification_shown
+    ) AS plugins_notification_shown,
+    mozfun.hist.extract(
+      payload.histograms.plugins_notification_user_action
+    ) AS plugins_notification_user_action
```
jklukas, line 59:
```suggestion
    mozfun.hist.extract(payload.histograms.ssl_handshake_result) AS ssl_handshake_result,
    mozfun.hist.extract(
      payload.histograms.plugins_notification_shown
    ) AS plugins_notification_shown,
    mozfun.hist.extract(
      payload.histograms.plugins_notification_user_action
    ) AS plugins_notification_user_action,
    *
```

If we add `*` to this CTE, we can avoid the join. The main query would select from `extracted_histograms` rather than `main`, and thus have access to these extra derived fields.

In this case, the WHERE clause only needs to appear in one place, either here or in the main query but not both.

sql/moz-fx-data-shared-prod/telemetry_derived/main_summary_v4/part1.sql
jklukas: See comment above. We should avoid adding a join to this query, which would likely be a significant drag on performance.

Issue #1432, opened 2020-10-13
clients_scalar_probe_counts_v1 exceeds resources for shuffle


BenWu:
```
Resources exceeded during
query execution: Your project or organization exceeded the maximum disk and
memory limit available for shuffle operations. Consider provisioning more slots,
reducing query concurrency, or using more efficient logic in this job.
```

The `clients_scalar_probe_counts_v1` query intermittently fails due to resources exceeded.  It seems to work on the retry, possibly because there are fewer running queries at that time.  There may be some query optimizations we could make to avoid this.

Ben-Wu:
Changed task dependencies so client_scalar_probe_counts runs after scalar_percentiles https://github.com/mozilla/telemetry-airflow/pull/1169

This may have fixed the issue for now but a more reliable solution would be ideal

BenWu:
fixed by ^

PR #1594, opened 2020-12-03
Separate queries on clients_scalar_aggregates by app_version

Ben-Wu:
For #1432

This makes scalar_probe_counts and scalar_percentiles make multiple queries per app_version.  clients_scalar_aggregates_v1 is clustered by app_version so this is somewhat efficient.  app_versions are nicely distributed: proportions are about 0.39, 0.38, 0.19, 0.034, 0.0037.  Clustering for the larger proportions aren't very efficient and still need to scan most of the table though so this increases total query cost.

Airflow pr https://github.com/mozilla/telemetry-airflow/pull/1194

### DIFF
--- a/script/glam/run_scalar_agg_clustered_query.py
+++ b/script/glam/run_scalar_agg_clustered_query.py
@@ -1,5 +1,6 @@
 """Run a query on clients_scalar_aggregates on one app_version at a time."""
 import datetime
+import sys
 import time
 from pathlib import Path
 
@@ -43,6 +44,10 @@ def main(submission_date, dst_table, project, dataset):
 
     print(f"Found versions: {app_versions}")
 
+    if len(app_versions) == 0:
+        print("Source table empty", file=sys.stderr)
+        sys.exit(1)
+
     sql_path = SQL_BASE_DIR / dst_table / "query.sql"
 
     query_text = sql_path.read_text()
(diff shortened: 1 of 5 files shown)


script/glam/run_scalar_agg_clustered_query.py
```diff
@@ -0,0 +1,100 @@
...
+        query_config = bigquery.QueryJobConfig(
+            query_parameters=[
+                bigquery.ScalarQueryParameter(
+                    "submission_date", "DATE", str(submission_date)
+                ),
+                bigquery.ScalarQueryParameter("app_version", "INT64", app_version),
+            ],
+            destination=intermediate_table,
+            default_dataset=f"{project}.{dataset}",
+            write_disposition=bigquery.WriteDisposition.WRITE_TRUNCATE
```
acmiyaguchi, line 66: Maybe some parens around this inline if statement.

script/glam/run_scalar_agg_clustered_query.py
```diff
@@ -0,0 +1,100 @@
...
+        query_job = bq_client.query(query_text, job_config=query_config)
+
+        # Periodically print so airflow gke operator doesn't think task is dead
+        elapsed = 0
+        while not query_job.done():
+            time.sleep(10)
+            elapsed += 10
+            if elapsed % 200 == 10:
+                print("Waiting on query...")
+
```
acmiyaguchi, line 80: Might as well print out the elapsed time at the end.

script/glam/run_scalar_agg_clustered_query.py
```diff
@@ -0,0 +1,100 @@
...
+        results = query_job.result()
+        print(f"{results.total_rows} rows in {intermediate_table}")
```
acmiyaguchi, line 82: Could you include the job id for debugging for something like `bq show -j ${job_id}`?

acmiyaguchi (approved):
Looks good to me.

script/glam/run_scalar_agg_clustered_query.py
Ben-Wu: Yeah I'll add that.  One thing I noticed though is that each iteration takes more than the time.sleep number of seconds so it would only be an estimate

PR #1601, opened 2020-12-08
Bug 1681208 - Expose subset of `stripe_external.invoices_v1`

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1681208)

This exposes a subset of columns that does not expose sensitive user data and can therefore be viewable as `mozilla-confidential`.

### DIFF
--- a/sql/moz-fx-data-shared-prod/stripe_derived/invoices_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/stripe_derived/invoices_v1/query.sql
@@ -1,0 +1,22 @@
+SELECT
+  -- limit fields in stripe_derived so as not to expose sensitive data
+  id,
+  billing_reason,
+  customer,
+  status,
+  status_transitions,
+  ARRAY(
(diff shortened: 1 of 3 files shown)


sql/moz-fx-data-shared-prod/stripe_derived/invoices_v1/query.sql
```diff
@@ -0,0 +1,22 @@
...
+  id,
+  billing_reason,
+  customer,
+  status,
+  status_transitions,
+  ARRAY(
+    SELECT
+      STRUCT(
+        line.id,
+        line.amount,
```
relud, line 12: legal has told us that this should not be visible to the whole company in re:dash, so if we are exposing these values they need to be restricted to a separate data source with access limited to a specific working group.

acmiyaguchi:
It turns out the `amount` field is considered sensitive. Closing this out.

PR #1605, opened 2020-12-08
[Bug 1671986] Average ping size monitoring

scholtzan:
Python script for determining average ping sizes.

I experimented a little with BigQuery scripting. It is possible to express this using BigQuery scripting instead: https://github.com/mozilla/bigquery-etl/pull/1605/commits/99b3dbc33d3fdd250b39434def36a637d01ede46
The downside is, however, that it takes much longer to execute than the Python script and might be harder to comprehend.


scholtzan:
I created `monitoring.average_ping_sizes_v1` and backfilled two days for testing.

jklukas:
Apologies for entirely missing this review. I will take a look today.

jklukas (approved):

sql/moz-fx-data-shared-prod/monitoring/average_ping_sizes_v1/query.py
```diff
@@ -0,0 +1,122 @@
...
+        """
+
+        result = client.query(sql).result()
+        average_ping_size = [row.average_ping_size for row in result][0]
+
+        return {
+            "submission_date": date,
+            "dataset_id": dataset_id,
+            "table_id": table_id,
+            "average_byte_size": average_ping_size,
```
jklukas, line 57: We could maintain `byte_size` and `row_count` columns in this table as well. It may be useful to be able to define a dashboard that correlates these 3 values, so it would be convenient to have them all in one place.

PR #1618, opened 2020-12-14
Add links to metadata.yaml editor and source dictory (#1242)

### LINKED ISSUE #1242: Edit on Github link in docs doesn't work
From https://mozilla.github.io/bigquery-etl/mozfun/norm/, the "Edit on Github" link in the top right 404s.

Iinh:
Fixes #1242.

Per @jklukas's suggestion, I added 2 new links at the end of each function's docs: Source that takes the user to the function directory, and Edit to the metadata.yaml edit interface.

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -4,7 +4,7 @@ site_description: Mozilla BigQuery ETL
 site_author: Mozilla Data Platform Team
 
 # Repository
-repo_name: mozilla/bigquery-etl/
+repo_name: mozilla/bigquery-etl
 repo_url: https://github.com/mozilla/bigquery-etl/
 edit_uri: "" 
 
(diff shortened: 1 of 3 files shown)


docs/mkdocs.yml
```diff
@@ -3,8 +3,12 @@ site_name: BigQuery ETL
 site_description: Mozilla BigQuery ETL
 site_author: Mozilla Data Platform Team
 
+# Repository
+repo_name: mozilla/bigquery-etl/
```
scholtzan, line 7: Does this need to be set? Looking at the docs mkdocs seems to automatically extract this name from the `repo_url`: https://www.mkdocs.org/user-guide/configuration/#repo_name

Iinh: mkdocs-material's default repo name is just "Github", and I thought it would be nice if we change it to our actual repo name. I don't have a strong opinion about this though, happy to remove it if you don't find it necessary :) https://squidfunk.github.io/mkdocs-material/setup/adding-a-git-repository/#repository-name

scholtzan: Okay, cool. In that case we should remove the `/` at the end.
```suggestion
repo_name: mozilla/bigquery-etl
```

scholtzan (approved):

PR #1619, opened 2020-12-14
Bug 1673979 - Add Search dashboard queries

xluo-ds:
Create 3 tables for the search dashboard, these tables may not be optimized at this moment as we believe it might be short-lived considering that we are moving towards the new BI tool. 

mobile_search_aggregates_for_searchreport
desktop_search_aggregates_for_searchreport
desktop_search_aggregates_by_userstate


sunahsuh (approved):
Xuan and I paired on most of this PR

PR #1627, opened 2020-12-16
Schedule deviations query until EOY

scholtzan:
There is renewed interest in the deviations dataset. Data is only needed until the end of the year, so I added support to add an `end_date` to DAGs.

Depends on https://github.com/mozilla/telemetry-airflow/pull/1207

cc @Dexterp37

### DIFF
--- a/dags.yaml
+++ b/dags.yaml
@@ -315,3 +315,12 @@ bqetl_desktop_platform:
     retries: 2
     retry_delay: 30m
 
+bqetl_deviations:
+  schedule_interval: 0 4 * * *
+  default_args:
+      owner: ascholtz@mozilla.com
+      start_date: '2020-10-01'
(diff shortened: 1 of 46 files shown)


dags.yaml
```diff
@@ -315,3 +315,17 @@ bqetl_desktop_platform:
     retries: 2
     retry_delay: 30m
 
+bqetl_deviations:
+  schedule_interval: 0 4 * * *
+  default_args:
+      owner: ascholtz@mozilla.com
+      start_date: '2020-10-01'
```
scholtzan, line 322: The DAG got removed on 2020-10-02. The datasets are still there. Once this gets merged I'll run a backfill to update the dataset.

jklukas (approved):
LGTM

PR #1630, opened 2020-12-18
[Bug 1673979] - Add Search dashboard queries

xluo-ds:
Add a filter to remove outliers in search count for `desktop_search_aggregates_by_userstate` table


BenWu:
I think this should be a pull request from a new branch?  Most of the changes are already in master.

PR #1631, opened 2020-12-18
[Bug 1673979] Add filter to remove outliers in search count

xluo-ds:
Add a filter to remove outliers in search count for `desktop_search_aggregates_by_userstate` table


sunahsuh (approved):
LGTM!

PR #1632, opened 2020-12-18
Coalesce search metrics in desktop_search_aggregates_by_userstate_v1 …

sunahsuh:
…before filtering

Whoops – the previous query removed all clients where any of these metrics were null (including all clients before 6/30/2020 since it appears one of the columns is newer)


xluo-ds:
LGTM :)

xluo-ds (approved):

PR #1634, opened 2020-12-21
Bug 1683765 - Import VPN surveys from alchemer/surveygizmo

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1683765)

Taking the output table, extracting the schema using `bq show --format json`, and setting this as part of the LoadJob configuration will fail with the following error:

```
 Error while reading data, error message: JSON parsing error in row starting at position 0: Couldn't convert value to timestamp: Unrecognized timezone: EST Field: date_submitted; Value: 2020-12-18 22:57:59 EST 
```

Because of this, I'm omitting `date_submitted` and `date_completed` which are likely not useful for the types of surveys that are being run. This can be added to the schema at some other point.


sql/moz-fx-data-shared-prod/mozilla_vpn_external/survey_product_quality_v1/query.py
```diff
@@ -0,0 +1,4 @@
+from bigquery_etl.alchemer.survey import main
+
+if __name__ == "__main__":
+    main()
```
scholtzan, line 4: I'm assuming this will be added later?

scholtzan (approved):
lgtm

sql/moz-fx-data-shared-prod/mozilla_vpn_external/survey_product_quality_v1/query.py
acmiyaguchi: This is actually implemented, I'm just reusing the same click command in each one of these jobs.

scholtzan: oh I see, cool!

PR #1639, opened 2021-01-04
Bug 1684935 - Use non-ISO 8601 years for FxA Log retrieval

fbertsch:
ISO 8601 years are week-numbering years. Specifically, an ISO
year begins on the first Monday of Week 01; so if in the
Gregorian calendar the year starts on a Friday, the Friday,
Saturday, and Sunday will all fall in the previous year.

We run into this problem here. %g uses the ISO year, so for
2021-01-01 and 2021-01-02, that is year '20'. This has no match
in the underlying data (no 20-01-01 in FxA logs). Switching to
%y gives us year '21' for this data, and a match in the FxA logs.

https://en.wikipedia.org/wiki/ISO_8601#Week_dates


fbertsch:
Along with this change, we will have to backfill these tables. For the Amplitude exports we should be fine, since this will just add more events that we should have already sent.

jklukas (approved):
I was entirely unaware of this year numbering convention. This looks good and I've confirmed that BQ is able to prune partitions correctly with this change.

PR #1640, opened 2021-01-04
Pin pip-tools to 5.4 until dependabot updates

relud:
unblocks #1635 and #1636

this provides a path for explicitly updating to pip-tools 5.5 once dependabot/dependabot-core#2943 is merged


requirements.txt
```diff
@@ -703,9 +626,15 @@ yarl==1.6.0 \
...
     --hash=sha256:e32f0fb443afcfe7f01f95172b66f279938fbc6bdaebe294b0ff6747fb6db020 \
-    --hash=sha256:fcbe419805c9b20db9a51d33b942feddbf6e7fb468cb20686fd7089d4164c12a
+    --hash=sha256:fcbe419805c9b20db9a51d33b942feddbf6e7fb468cb20686fd7089d4164c12a \
     # via aiohttp
 
-# WARNING: The following packages were not pinned, but pip requires them to be
-# pinned when the requirements file includes hashes. Consider using the --allow-unsafe flag.
-# setuptools
+# The following packages are considered to be unsafe in a requirements file:
+pip==20.3.3 \
```
relud, line 633: `pip` is required by `pip-tools`, so we need to include hashes for it (via `--allow-unsafe`) to prevent `requirements.txt` from becoming uninstallable

acmiyaguchi (approved):

PR #1645, opened 2021-01-05
Streamline bqetl initialization and utilization

wlach:
This simplifies the multiple choice installation process into
a simplified version which uses a local virtualenv, which will
hopefully be easier to support and debug.


scholtzan (approved):
I agree, this could make usability a bit easier. 👍

PR #1656, opened 2021-01-08
Experiment enrollment monitoring queries

scholtzan:
Implements the ETL described in https://docs.google.com/document/d/1IU7c2iJ1Cx54D9aHWmzy0kTtpMSbSVjC4YTN8UoqrJo

Currently, only two of the Grafana queries are implemented (experiment enrollment cumulative population and daily active population). Once this PR gets approved I'll open a follow-up PR that implements the remaining queries.

Export script is scheduled in https://github.com/mozilla/telemetry-airflow/pull/1217

Dry run is currently failing because tables have not been created yet.


bigquery_etl/query_scheduling/task.py
```diff
@@ -155,6 +155,7 @@ class Task:
     allow_field_addition_on_date: Optional[str] = attr.ib(None)
     destination_table: Optional[str] = attr.ib(default=DEFAULT_DESTINATION_TABLE_STR)
     is_python_script: bool = attr.ib(False)
+    execution_delay: Optional[str] = attr.ib(None)
```
scholtzan, line 158: I added an `execution_delay` option here. For the hourly experiment enrollment aggregate task we want to make sure we account for potential BigQuery sink delays. The delay can be configured with this parameter in the `metadata.yaml` file.

dags/bqetl_experiments_hourly.py
```diff
@@ -0,0 +1,44 @@
...
+        date_partition_parameter="submission_timestamp",
+        depends_on_past=False,
+        parameters=["submission_timestamp:TIMESTAMP:{{ts}}"],
+        dag=dag,
+    )
+
+    experiment_enrollment_aggregates_hourly_execution_delay = TimeDeltaSensor(
+        task_id="experiment_enrollment_aggregates_hourly_execution_delay",
+        delta=datetime.timedelta(seconds=1800),
+    )
```
scholtzan, line 40: 30 minutes should be enough to account for the BigQuery sink delay.

script/experiments/export_experiment_monitoring_data.py
```diff
@@ -0,0 +1,250 @@
...
+                        TIMESTAMP_SECONDS(
+                            UNIX_SECONDS(time) - MOD(UNIX_SECONDS(time), 30 * 60) + 30 * 60
+                            ) AS time,
+                    FROM data
+                    ORDER BY time DESC
+                )
+            )
+            WHERE rn = 1
+            ORDER BY time DESC
+        """
```
scholtzan, line 120: @emtwo For the exported data, does it make sense to aggregate data into 30 minute intervals if the experiment has been running for 14 days or more? In the first two weeks of the experiment, the exported data will consist of 5 minute intervals.

sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_aggregates_hourly_v1/metadata.yaml
```diff
@@ -0,0 +1,17 @@
...
+owners:
+- ascholtz@mozilla.com
+scheduling:
+  dag_name: bqetl_experiments_hourly
+  task_name: experiment_enrollment_aggregates_hourly
+  parameters: ["submission_timestamp:TIMESTAMP:{{ts}}"]
+  referenced_tables: [
+    "moz-fx-data-shared-prod", "telemetry_derived", "events_live"
+  ]
+  date_partition_parameter: submission_timestamp
```
jklukas, line 16: I don't think this is going to work without changes to `bigquery_etl_query` in telemetry-ariflow. The relevant code is:

```
    if destination_table is not None and date_partition_parameter is not None:
        destination_table = destination_table + "${{ds_nodash}}"
        parameters += (date_partition_parameter + ":DATE:{{ds}}",)
```

So it assumes a DATE type, and it injects `ds_nodash`. A hourly partition destination table has to be specified with a suffix of `$YYYMMDDHH`.

Do we have any other potential use cases queued up for hourly queries that might want to use a pattern like this? If not, it may be best to express this as a custom `query.py` that handles both the partitioning logic and the execution_delta. Instead of introducing the execultion_delta parameter to tasks, the task could execute immediately, and adjust the timestamp back 30 minutes. The DAG would run on the half hour in that case.

jklukas:
I'm out most of the day today, so likely won't be able to give an in-depth review until tomorrow.

bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -1,7 +1,7 @@
 # Generated via https://github.com/mozilla/bigquery-etl/blob/master/bigquery_etl/query_scheduling/generate_airflow_dags.py
 
 from airflow import DAG
-from airflow.operators.sensors import ExternalTaskSensor
+from airflow.operators.sensors import ExternalTaskSensor, TimeDeltaSensor
```
fbertsch, line 4: Super nit - can this be `{%+ if task.execution_delay %}, TimeDeltaSensor{% endif %}` to prevent unnecessary changes to all of the dag files?

sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_aggregates_hourly_v1/metadata.yaml
scholtzan: We'll also need to set up an hourly query for monitoring experiment search aggregates. There is also some interest in monitoring engagement data, which would probably use the same mechanism, though nothing has been planned out for this use case yet.

scholtzan: I think I might have found a workaround by using the Airflow macro [`ds_format`](https://airflow.apache.org/docs/apache-airflow/stable/macros-ref.html#airflow.macros.ds_format) which can format the nodash timestamp into a timestamp that indicates the hourly table partition:

```
    experiment_enrollment_aggregates_hourly = bigquery_etl_query(
        ...
        destination_table='experiment_enrollment_aggregates_hourly_v1${{ macros.ds_format(ts_nodash, "%Y%m%d%H%M%S", "%Y%m%d%H") }}',
        date_partition_parameter=None,
        ...
    )
```

The [GKEPodOperator `argument` field is a templated field](https://airflow.apache.org/docs/apache-airflow/1.10.12/_api/airflow/contrib/operators/kubernetes_pod_operator/index.html#module-airflow.contrib.operators.kubernetes_pod_operator), so when `destination_table` gets passed as argument it should get evaluated correctly.

jklukas: Do you think it would be possible to also solve the TimeDeltaSensor problem using `ds_format`?

We could have the DAG set to run on `HH:30` and then pass in parameter `submission_timestamp:TIMESTAMP:{{ macros.ds_format(ts, "%Y-%m-%d-%H:00:00") }}` (ts truncated to the hour). Then there would be no need to introduce a delay.

scholtzan: Indeed! I updated the metadata.yaml for the query and removed TimeDeltaSensor

sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_cumulative_population_estimate_v1/query.sql
```diff
@@ -0,0 +1,149 @@
+WITH all_enrollments AS (
```
scholtzan, line 1: I did some performance testing here. Instead of querying the view when exporting the data to GCS, it is more than 5x faster if the data is persisted in a table that is queried by the export script. So I changed this into a query.

script/experiments/export_experiment_monitoring_data.py
emtwo: Very nice! I looked at the json files, they look great!
I am mostly thinking outloud here, but I am wondering how we would graph the intervals. E.g. I have a slight concern that 5 min intervals for 2 weeks graphed as-is will be slow on the front-end to graph. I see that Grafana customizes the intervals graphed based on how much data is there. That is possible to do on the front-end if we drop some data points depending on how many there are. 

It's also possible to do it on the ETL side, but I am worried then we cannot retrieve the data if we need it later.

I guess we can start with the 5min intervals for 14 days and 30min for more. If that is an issue, seems like something we can adjust later, right?

sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_aggregates_hourly_v1/metadata.yaml
```diff
@@ -0,0 +1,20 @@
...
+  from the past hour, partitioned by hour.
+labels:
+  incremental: true
+owners:
+- ascholtz@mozilla.com
+scheduling:
+  dag_name: bqetl_experiments_hourly
+  task_name: experiment_enrollment_aggregates_hourly
+  parameters: [
+    'submission_timestamp:TIMESTAMP:{{ macros.ds_format(ts_nodash, "%Y%m%d%H%M%S", "%Y-%m-%d %H:00:00") }}' # yamllint disable-line rule:line-length
```
jklukas, line 13:
```suggestion
    # This DAG runs 30 minutes after the hour to allow for some latency in data loading;
    # we truncate ts_nodash so that the query represents the previous full hour.
    'submission_timestamp:TIMESTAMP:{{ macros.ds_format(ts_nodash, "%Y%m%d%H%M%S", "%Y-%m-%d %H:00:00") }}' # yamllint disable-line rule:line-length
```

sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_aggregates_recents_v1/metadata.yaml
```diff
@@ -0,0 +1,16 @@
...
+labels:
+  incremental: false
+owners:
+- ascholtz@mozilla.com
+scheduling:
+  dag_name: bqetl_experiments_live
+  task_name: experiment_enrollment_aggregates_recents
+  parameters: ["submission_timestamp:TIMESTAMP:{{ts}}"]
+  referenced_tables: [
+    ["moz-fx-data-shared-prod", "telemetry_derived", "events_live"]
```
jklukas, line 14: `events_live` is a virtual view, so shouldn't appear in `referenced_tables`. We should replace these references with the underlying tables that the view hits.

jklukas (approved):
Thanks for working through the details of `ds_format` capabilities here. This is looking much cleaner now.

How important is it that the hourly table be correct? There will be rare occasions where the pipeline latency stretches past 30 minutes and the hourly result will be incomplete. We could have a daily query to re-populate the previous full day to effectively "heal" the table. Probably out of scope here, but something to keep in mind if we find this data becomes more critical.

script/experiments/export_experiment_monitoring_data.py
scholtzan: The underlying experiment_enrollment_aggregates tables aggregate the data in 5 minute intervals. If we'd want to aggregate the data into larger intervals to reduce front-end load, then we can adjust that in the export script easily.

scholtzan:
It's not too critical. The `experiment_enrollment_aggregates_v1` which is running daily is there to finalize numbers from the stable tables and make sure results are complete at the end of the day.

scholtzan:
The tables have been created and I did some test runs.

PR #1657, opened 2021-01-08
Speed up dry run

scholtzan:
Using `multiprocessing.Pool` instead of `multiprocessing.ThreadPool` does significantly improve the dry run speed. Locally, it takes `11:03.11` minutes with `multiprocessing.ThreadPool` and `1:29.26` minutes with `multiprocessing.Pool` for me. This also speeds up the DAG generation.

### DIFF
--- a/bigquery_etl/dryrun.py
+++ b/bigquery_etl/dryrun.py
@@ -11,7 +11,7 @@ accidentally running queries during tests and overwriting production data, we
 proxy the queries through the dry run service endpoint.
 """
 from functools import cached_property
-from multiprocessing.pool import ThreadPool
+from multiprocessing.pool import Pool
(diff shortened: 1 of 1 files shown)


bigquery_etl/dryrun.py
```diff
@@ -217,6 +217,11 @@ def is_valid(self):
         return True
 
 
+def sql_file_valid(sqlfile):
+    """Dry run SQL files."""
+    return DryRun(sqlfile).is_valid()
+
+
```
scholtzan, line 224: I had to move this out of `main` otherwise it raised `AttributeError: Can't pickle local object 'main.<locals>.sql_file_valid'`

relud (approved):

PR #1658, opened 2021-01-11
Fix suggestions for daily, weekly, ... scheduling options

wlach:
The comment (and help text) falsely implied you should prefix
them with an `@` symbol.

### DIFF
--- a/bigquery_etl/cli/dag.py
+++ b/bigquery_etl/cli/dag.py
@@ -93,7 +93,7 @@ def info(name, dags_config, with_tasks):
     help=(
         "Schedule interval of the new DAG. "
         "Schedule intervals can be either in CRON format or one of: "
-        "@once, @hourly, @daily, @weekly, @monthly, @yearly or a timedelta []d[]h[]m"
+        "once, hourly, daily, weekly, monthly, yearly or a timedelta []d[]h[]m"
     ),
     required=True,
 )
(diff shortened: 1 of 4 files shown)


scholtzan (approved):
There are a few more occurrences of this comment that should be fixed:

https://github.com/mozilla/bigquery-etl/blob/b07b98e8255cedce2b40f4dfee9e6811ec6ed17e/bigquery_etl/query_scheduling/task.py#L87

https://github.com/mozilla/bigquery-etl/blob/b07b98e8255cedce2b40f4dfee9e6811ec6ed17e/bigquery_etl/query_scheduling/utils.py#L54

https://github.com/mozilla/bigquery-etl/blob/b07b98e8255cedce2b40f4dfee9e6811ec6ed17e/bigquery_etl/query_scheduling/dag.py#L147

scholtzan (approved):
👍

wlach:
> There are a few more occurrences of this comment that should be fixed:

You're right, actually this PR didn't fix what I thought it should, it was editing the wrong instance of the string. I am kind of feeling like repeating the expected format in each comment is not needed (and kind of a violation of [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)), it's really just the command line help where we want to print that string. What do you think?

scholtzan:
Yeah, there is no need in repeating the same comment everywhere.

PR #1661, opened 2021-01-13
Bug 1683765 - Add views to survey tables in the external dataset

acmiyaguchi:
Yet another PR for [bug 1683765](https://bugzilla.mozilla.org/show_bug.cgi?id=1683765). This makes the tables visible to everyone with access to STMO. This should be the last PR in this series.


relud:
> This makes the tables visible to everyone with access to STMO.

They may be visible, but ``` `moz-fx-data-shared-prod`.mozilla_vpn_external``` isn't (and shouldn't be) accessible in re:dash, so it will throw access denied errors if someone tries to use it.

acmiyaguchi:
I see, is this why there is a derived table for each of the external tables in the mozilla vpn dataset?

relud:
> I see, is this why there is a derived table for each of the external tables in the mozilla vpn dataset?

yep

relud (changes requested):
waiting to r+ until after underlying tables are copied or moved to `mozilla_vpn_derived`

acmiyaguchi:
Does it make sense for these imports to live in `mozilla_vpn_derived`, instead of `external`?

relud:
I don't mind it, but @jklukas might have opionions

jklukas:
We do in practice have several tables in `*_derived` datasets that are populated via scripts, etc. I definitely want any table that references an external data source to be in an `external` dataset, but in this case if we have a script that is pulling data and loading into into a concrete BQ table, `_derived` is acceptable.

Are there any concerns about sensitivity of the data? If the results of the import are sufficiently sanitized that they can be mozilla-confidential, that makes it easier.

What else exists in the vpn derived and vpn external datasets? I just want to make we're at least being consistent within the realm of VPN data about what we're putting where.

acmiyaguchi:
We currently have tables being mirrored from a postgres database in `_external`, and queries that limit columns in `_derived`. As for the survey data themselves, the imports are removing information about IP and geography that is returned by default by the API, so it should be fine to make moco confidential (assuming that free text responses are acceptable for viewing by all Mozilla).

acmiyaguchi:
Ran the following command to move surveys into the correct location:

```bash
$ bq ls --format json  mozilla_vpn_external | \
  jq -r '.[] | .tableReference.tableId' | \
 grep survey | \
 xargs -I{} echo bq cp -f mozilla_vpn_external.{} mozilla_vpn_derived.{}

bq cp -f mozilla_vpn_external.survey_cancellation_of_service_v1 mozilla_vpn_derived.survey_cancellation_of_service_v1
bq cp -f mozilla_vpn_external.survey_intercept_q3_v1 mozilla_vpn_derived.survey_intercept_q3_v1
bq cp -f mozilla_vpn_external.survey_market_fit_v1 mozilla_vpn_derived.survey_market_fit_v1
bq cp -f mozilla_vpn_external.survey_product_quality_v1 mozilla_vpn_derived.survey_product_quality_v1
bq cp -f mozilla_vpn_external.survey_recommend_v1 mozilla_vpn_derived.survey_recommend_v1
```

relud (approved):

acmiyaguchi:
Ran the output of the following command to clean up the external dataset:

```bash
 bq ls --format json  mozilla_vpn_external | \
  jq -r '.[] | .tableReference.tableId' | \
 grep survey | \
 xargs -I{} echo bq rm -f mozilla_vpn_external.{}
```

PR #1663, opened 2021-01-13
Move shredder view to a shredder project directory

whd:
I noticed this when deploying `mozdata` views, where ops logic currently [computes the views that should be deployed to mozdata](https://github.com/mozilla-services/cloudops-infra/blob/mozdata_project/projects/data-pioneer/Jenkinsfile.bigquery#L209-L216) via diff. This method relies on knowing more about bqetl directory structure than ops infra probably should, but as `mozdata` becomes supported here I believe this'll be sorted out naturally. In any case, having this shredder view within the shared-prod directory doesn't actually affect the diff  and the view is skipped during publishing anyway, so this PR should mostly be a no-op.

This view is also referenced in https://github.com/mozilla/bigquery-etl/blob/master/bigquery_etl/view/publish_views.py#L27 but without a project prefix.


relud:
lol, I had independently remembered this view this morning, and decided that it should be moved to the `moz-fx-data-shared-prod:monitoring` dataset, which is done in #1666

PR #1667, opened 2021-01-13
Fix null handling for structs in clients daily

relud:
this has been broken all along, but it probably wasn't noticeable because everything impacted is in environment, which is sent every time and changes infrequently

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
@@ -180,11 +180,15 @@ clients_summary AS (
     environment.settings.update.channel AS update_channel,
     environment.settings.update.enabled AS update_enabled,
     environment.settings.update.auto_download AS update_auto_download,
-    STRUCT(
-      environment.settings.attribution.source,
-      environment.settings.attribution.medium,
-      environment.settings.attribution.campaign,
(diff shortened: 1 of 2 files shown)


jklukas (approved):

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
```diff
@@ -672,12 +709,12 @@ aggregates AS (
...
       ARRAY_AGG(previous_build_id ORDER BY submission_timestamp)
     ) AS previous_build_id,
-    UNIX_DATE(DATE(SAFE.TIMESTAMP(ANY_VALUE(subsession_start_date)))) - ANY_VALUE(
+    MAX(UNIX_DATE(DATE(SAFE.TIMESTAMP(subsession_start_date)))) - MAX(
       profile_creation_date
     ) AS profile_age_in_days,
     FORMAT_DATE(
       "%F 00:00:00",
-      SAFE.DATE_FROM_UNIX_DATE(ANY_VALUE(profile_creation_date))
+      SAFE.DATE_FROM_UNIX_DATE(MAX(profile_creation_date))
```
relud, line 717: This doesn't have to be `MAX`, but something deterministic is better than `ANY_VALUE` and I wanted to avoid `mozfun.stats.mode_last` because it's expensive in terms of query complexity that would probably better spent on new fields.

PR #1672, opened 2021-01-14
Fix daily_active_population query to match schema

scholtzan:


jklukas (approved):

dags/bqetl_experiments_hourly.py
```diff
@@ -25,7 +25,7 @@
 
     experiment_enrollment_aggregates_hourly = bigquery_etl_query(
         task_id="experiment_enrollment_aggregates_hourly",
-        destination_table='experiment_enrollment_aggregates_hourly_v1${{ macros.ds_format(ts_nodash, "%Y%m%dT%H%M%S", "%Y%m%d%H") }}',
+        destination_table='experiment_enrollment_aggregates_hourly_v1${{ macros.ds_format(execution_date + macros.timedelta(hours=-1), "%Y-%m-%d %H:%M:%S", "%Y%m%d%H") }}',
```
scholtzan, line 28: It turns out the destination table partition is one hour off. Good news is there is a `macros.timedelta` which helps in computing the right partition. Based on [a blog post I found](https://www.astronomer.io/guides/templating) `execution_date` gets represented as a `%Y-%m-%d %H:%M:%S` string, which can then get formatted into the right partition format. Hopefully. Not sure if there is a better way to test this rather than trail and error...

PR #1673, opened 2021-01-14
Add video_play_time_ms_sum to clients_daily

jklukas:
cc @SuYoungHong


jklukas:
@SuYoungHong asked:

> other time values seem processed to hours in clients daily, but it seems to be in the raw, MS units here. is that a deliberate decision?

We can consider that. My preference would be to leave it in milliseconds and provide a value normalized to hours at the view level. But if this is the only measure shown in millis, I'll stick to hours for consistency.

jklukas:
This is deferred to a broader effort to include feature usage data consistently in these tables.

PR #1674, opened 2021-01-14
Desktop Funnel

wlach:
Not ready for review!


sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_installs_v1/query.sql
```diff
@@ -0,0 +1,66 @@
...
+  THEN
+    [country, 'tier-1']
+  WHEN
+    country IN ('MX', 'BR')
+  THEN
+    [country, 'non-tier-1']
+  ELSE
+    ['non-tier-1']
+  END
+);
```
wlach, line 14: Should we create a UDF to do this?

sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_installs_v1/query.sql
```diff
@@ -0,0 +1,66 @@
...
+CROSS JOIN
+  UNNEST(ARRAY_CONCAT(['Overall'], bucket_country(normalized_country_code))) AS country_group
+WHERE
+  DATE(submission_timestamp) = @submission_date
+  AND succeeded = TRUE
+  AND (
+    silent = FALSE
+    OR silent IS NULL
+  ) -- ignore silent installs (most of the time silent is unset)
+  AND build_channel = 'release'
```
wlach, line 50: Should probably make this a dimension, it might occasionally be interesting to look at beta/nightly install data.

sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_installs_v1/query.sql
```diff
@@ -0,0 +1,66 @@
...
+  AND DATE_DIFF(  -- Only use builds from the last month
+    DATE(submission_timestamp),
+    SAFE.PARSE_DATE('%Y%m%d', SUBSTR(build_id, 0, 8)),
+    MONTH
+  ) <= 1
```
wlach, line 55: I suspect we could probably do something more sophisticated here to filter out older builds.

sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_installs_v1/query.sql
```diff
@@ -0,0 +1,66 @@
...
+  -- FIXME: change the below to "not Firefox", but make sure there *is*
+  -- attribution (i.e. is not null, not 0)
+  AND get_ua_attribution(attribution) IN ('chrome', 'ie', 'edge')
```
wlach, line 58: This should also probably be a dimension. We can do the filtering in the final view?

sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_new_profiles_v1/query.sql
```diff
@@ -0,0 +1,92 @@
...
+    client_id,
+    submission_date,
+    country AS country_code,
+    any_value(days_seen_bits) AS days_seen_bits,
+  FROM
+    telemetry.clients_last_seen
+  WHERE
+    submission_date
+    BETWEEN @submission_date
+    AND DATE_ADD(@submission_date, INTERVAL 8 DAY)
```
wlach, line 51: For this to work, this query should only be processed 8 days after the date. Hopefully this is possible with airflow?

wlach (commented):
This is a straw man WIP based on the proposal here: https://docs.google.com/document/d/1T0WmStR8K1u_djo06u1EI9Rgyo6blsLIgUilksKf2C4/edit#

Having gone into the implementation, I am now feeling like maybe we should provide *more* dimensions in each of the individual sub-datasets, while putting the conditions in the view.

wlach:
I tagged a very long list of people for review this, but I don't need everyone to review everything. Looking for feedback from:

* @sghoseWI @gkabbz on the GA query (`telemetry_derived/attributed_windows_desktop_funnel_sessions_v1/query.sql`)
* @SuYoungHong overall review of all queries for correctness
* @scholtzan @jklukas review airflow integration, address question on how to schedule `telemetry_derived/attributed_windows_desktop_funnel_new_profiles_v1/query.sql`, maybe provide thoughts on how we should build a view that "brings these three datasets together", thoughts on anything I should do to make this data easier to access in Looker.

sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_sessions_v1/query.sql
```diff
@@ -0,0 +1,60 @@
...
+        return "FR"
+    } else if (input === "Canada") {
+        return "CA"
+    } else if (input === "Brazil") {
+        return "BR"
+    } else if (input === "Mexico") {
+        return "MX"
+    }
+    return "OTHER";
+    """;
```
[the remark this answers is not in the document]
gkabbz: I usually do this as a join to a standardized country table and then use a standard country name that would be joinable across datasets. There's `moz-fx-data-derived-datasets.static.country_names_v1` which standardizes names for telemetry country codes. I also have this one `moz-fx-data-derived-datasets.analysis.standardized_country_list` which standardizes both telemetry and Google Analytics data into the same country name. Makes it easier to join the different datasets by `standardizedCountry` once both datasets have been joined to it. At some point, we should probably add my analysis table to the static table.

sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_sessions_v1/query.sql
```diff
@@ -0,0 +1,60 @@
...
+FROM
+  `moz-fx-data-marketing-prod.ga_derived`.www_site_metrics_summary_v1
+CROSS JOIN
+  UNNEST(
+    ARRAY_CONCAT(['Overall'], bucket_country(map_country_name_to_prefix(standardized_country_name)))
+  ) AS country_group
+WHERE
+  date = @submission_date
+  AND operating_system = 'Windows'
+  AND browser != 'Mozilla'
```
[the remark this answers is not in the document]
gkabbz: Better still we should probably fix the derived table `www_site_metrics_summary_v1` to exclude browser = 'Mozilla' from the non_fx_downloads aggregation.

sql/moz-fx-data-shared-prod/telemetry/attributed_windows_desktop_funnel_installs/view.sql
```diff
@@ -0,0 +1,7 @@
+CREATE OR REPLACE VIEW
```
jklukas, line 1: Do we expect to provide a single unified view on top of installs, new profiles, and sessions? If that's what we want end users to reference, then it might not make sense to publish views for each of these individually.

sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_installs_v1/query.sql
jklukas: I don't think we should create a user-facing mozfun UDF out of this, since it involves logic that's pretty specific to this dataset. We could factor it out to a UDF under shared-prod, though, in which case I think it should have a name that makes it clear that it's for this use case (perhaps `desktop_funnel_bucket_country`).

sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_installs_v1/query.sql
jklukas: That makes sense to me, as long as the `get_ua_attribution` is fairly low cardinality (on order 10 distinct values). And we may want to make it a clustering field if we include it.

sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_installs_v1/query.sql
fbertsch: tier-1 and non-tier-1 are common operations, but what's weird is this returns an array. I could see having a public function `get_country_tier` which returns a string, and then this specific function uses it to return `[country, get_country_tier(country)]`

sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_sessions_v1/query.sql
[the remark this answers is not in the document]
wlach: I'm going to chat with @SuYoungHong about this but I think we should just skip this, and just use the table directly. Filed #1692 to make this behaviour the default

sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_sessions_v1/query.sql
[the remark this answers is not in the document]
BenWu: There's `moz-fx-data-shared-prod.static.third_party_standardized_country_names` which I believe is the same as the analysis table.  I created while moving the ga queries over

sql/moz-fx-data-shared-prod/telemetry/attributed_windows_desktop_funnel_installs/view.sql
wlach: This makes sense to me, I'll take these out and actually create the unified view.

dags.yaml
```diff
@@ -331,3 +331,19 @@ bqetl_desktop_platform:
...
+    - wlachance@mozilla.com
+    - telemetry-alerts@mozilla.com
+    email_on_failure: true
+    email_on_retry: true
+    end_date: null
+    owner: wlachance@mozilla.com
+    retries: 2
+    retry_delay: 30m
+    start_date: '2020-07-01'
+  schedule_interval: daily
```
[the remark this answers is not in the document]
wlach: I think this should no longer depend on `main_summary` now that clients_daily / clients_last_seen derives directly from main. I'll regenerate the DAGs.

SuYoungHong (dismissed):
notes from our working session

sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_sessions_v1/query.sql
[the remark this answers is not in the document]
wlach: Pulled this out.

sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_sessions_v1/query.sql
```diff
@@ -0,0 +1,60 @@
...
+    [country, 'tier-1']
+  WHEN
+    country IN ('MX', 'BR')
+  THEN
+    [country, 'non-tier-1']
+  ELSE
+    ['non-tier-1']
+  END
+);
+
```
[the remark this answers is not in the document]
wlach: @fbertsch I'm with @SuYoungHong -- I'd prefer to avoid expressing "tier 1" vs "rest of world" in this dataset, and just express that somehow with Looker (or the query accessing this dataset, in the cases where we fall back to redash). This is the sort of thing we could express with lookml, right? Am I safe to just leave it out of here for now?

sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_sessions_v1/query.sql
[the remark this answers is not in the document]
wlach: Awesome, sounds like we should normalized countries to this table.

sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_new_profiles_v1/query.sql
```diff
@@ -0,0 +1,92 @@
...
+    [country, 'non-tier-1']
+  ELSE
+    ['non-tier-1']
+  END
+);
+
+WITH pop AS (
+  SELECT
+    DISTINCT client_id,
+    normalized_country_code AS country_code,
```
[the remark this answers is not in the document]
gkabbz: +1

gkabbz (commented):
I think Su has a better feel for this dataset and its usage than I do. I do agree that we should avoid the country grouping in the dataset itself and we can do it on the presentation. Gives us more flexibility should the teams want to reclass countries and tiers.

sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_installs_v1/query.sql
wlach: I think I'm going to skip this tier-level stuff for the initial dataset. It's not clear to me that it's needed in the Looker world.

sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_installs_v1/query.sql
wlach: I actually took this logic out altogether for now, as I'm not sure if it should be here or in the Looker/BI layer. There might be some cases where people would want to understand/see old versions of Firefox being installed.

sql/moz-fx-data-shared-prod/telemetry_derived/attributed_windows_desktop_funnel_new_profiles_v1/query.sql
wlach: I split these in two.

wlach:
Update: I cleaned this up a bunch per @SuYoungHong's feedback and others. This dataset is now much more "raw" and unopinionated, and it includes many dimensions and fields that we would otherwise filter out or make assumptions about in the [numbers that matter dashboard](https://protosaur.dev/numbers-that-matter).

In the original proposal, I was thinking we'd provide a "view" on the data that we could use in something like redash, but I don't know if that makes sense in a Looker world. I'm going to circle back on this next week and see where the Looker implementation is at. I'm pretty sure what's in here now is what we want, but there might be better/different ways to go now if our end goal is to display this data in Looker.

wlach:
Update: I generated a snapshot of what these tables would look like manually, and attempted to reconstitute the numbers that matters information:

https://sql.telemetry.mozilla.org/queries/78309/source#194720

Installs, new profiles, and activations are all *higher* than what's in the existing numbers that matter dashboard, so that needs to figure it out. Some of the new profile part may be due to the fact that we're only using the new profile ping to measure it (rather than looking for a match in clients_last_seen, which I think only uses main pings?). 

Sessions and downloads are all the same (as one would expect, it's not a complicated query).

jklukas:
> clients_last_seen, which I think only uses main pings

That is correct. It's only main pings.

sql/moz-fx-data-shared-prod/telemetry_derived/desktop_funnel_installs_v1/query.sql
```diff
@@ -0,0 +1,46 @@
...
+  build_id,
+  silent,
+  succeeded,
+  get_ua_attribution(attribution) AS ua_attribution,
+  COUNTIF(had_old_install = FALSE) AS successful_new_installs,
+  COUNTIF(had_old_install = TRUE) AS successful_paveovers
+FROM
+  firefox_installer.install
+LEFT JOIN
+  `moz-fx-data-derived-datasets`.static.country_names_v1 country_names
```
scholtzan, line 33: I found the reason for why the numbers are higher here. It turns out the country `code`s in `static.country_names` are not unique. Some countries have multiple entries. So `LEFT JOIN`ing here counts some installs multiple times for some countries.

```sql
select code, count(code) 
from `moz-fx-data-derived-datasets`.static.country_names_v1 country_names
group by 1
order by 2 desc

-- AX 2
-- BN 2
...
```

scholtzan:
This is ready for review. 

I fixed the inflated installs, new profiles, and activations numbers and verified that the numbers match the ones currently shown in the NtM dashboard.

sql/moz-fx-data-shared-prod/telemetry_derived/desktop_funnel_activation_v1/query.sql
```diff
@@ -0,0 +1,108 @@
...
+        code,
+        name
+      FROM
+        `moz-fx-data-derived-datasets`.static.country_names_v1 country_names
+    )
+  WHERE
+    rn = 1
+),
+pop AS (
+  SELECT DISTINCT
```
jklukas, line 20: Why are we using SELECT DISTINCT here? If the intention is to make sure that we only get one entry per client, this doesn't accomplish that. If a client were to send two new_profile pings with any difference in any of the fields listed here, they would both be in the result set.

I'd prefer to see this expressed with ROW_NUMBER() like in the distinct_countries CTE above.

sql/moz-fx-data-shared-prod/telemetry_derived/desktop_funnel_activation_v1/schema.yaml
```diff
@@ -0,0 +1,31 @@
...
+  type: STRING
+- mode: NULLABLE
+  name: build_id
+  type: STRING
+- mode: NULLABLE
+  name: os
+  type: STRING
+- mode: NULLABLE
+  name: os_version
+  type: NUMERIC
```
jklukas, line 19: I think it will be surprising to have this field stored as `NUMERIC` and that we should cast it to string.

jklukas (commented):
Leaving comments so far. I wasn't able to fully read through this before my EOD.

sql/moz-fx-data-shared-prod/telemetry_derived/desktop_funnel_activation_v1/query.sql
```diff
@@ -0,0 +1,108 @@
...
+eight_days_later AS (
+  SELECT
+    a.*,
+    b.days_seen_bits
+  FROM
+    pop a
+  LEFT JOIN
+    raw_info b
+  ON
+    (a.client_id = b.client_id AND DATE_DIFF(b.submission_date, a.date, DAY) = 6)
```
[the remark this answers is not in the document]
scholtzan: Related to comment https://github.com/mozilla/bigquery-etl/pull/1674#discussion_r561324111

sql/moz-fx-data-shared-prod/telemetry_derived/desktop_funnel_activation_v1/query.sql
```diff
@@ -0,0 +1,87 @@
...
+  os,
+  CAST(os_version AS STRING) AS os_version,
+  attribution_source,
+  distribution_id,
+  attribution_ua,
+  count(*) AS num_activated
+FROM
+  client_conditions
+LEFT JOIN
+  `moz-fx-data-derived-datasets`.static.country_codes_v1 country_codes
```
jklukas, line 72:
```suggestion
  mozdata.static.country_codes_v1 country_codes
```

The static tables are published to multiple projects, but `moz-fx-data-derived-datasets` is now deprecated and `mozdata` should be the canonical entrypoint.

jklukas:
I don't see any `view.sql` in the current version of this PR. Is there an intention to present a final user-facing view as discussed in https://github.com/mozilla/bigquery-etl/pull/1674#discussion_r562224712 ?

If so, I'd recommend doing final cleaning like the join to `country_codes` at the view level.

wlach:
> I don't see any `view.sql` in the current version of this PR. Is there an intention to present a final user-facing view as discussed in [#1674 (comment)](https://github.com/mozilla/bigquery-etl/pull/1674#discussion_r562224712) ?
> 
> If so, I'd recommend doing final cleaning like the join to `country_codes` at the view level.

So I may not have as much context anymore, but based on where things were a couple weeks ago I would advise against providing a stable interface to this data. As I see it, there are three reasons to finish this work:

1. (less important) Update the numbers that matter dashboard on a faster cadence (daily vs. every week) at a reasonable price.
2. (more important) Provide an interim tool that might be useful to product management and others for answering short-term questions while we wait for the Looker implementation to finish.
3. (most important) Develop internal competence and understanding of this data model.

Providing a stable, supported interface to this data is not needed or desired to achieve the above and might actually be counter-productive.

sql/moz-fx-data-shared-prod/telemetry_derived/desktop_funnel_activation_v1/metadata.yaml
```diff
@@ -0,0 +1,16 @@
...
+owners:
+  - ascholtz@mozilla.com
+scheduling:
+  dag_name: bqetl_desktop_funnel
+  parameters: [
+    # Set the submission_date to the date from 8 days ago
+    'submission_date:DATE:{{ macros.ds_format((execution_date + macros.timedelta(days=-8)).to_datetime_string(), "%Y-%m-%d %H:%M:%S", "%Y-%m-%d") }}'   # yamllint disable-line rule:line-length
+  ]
+  date_partition_parameter: null
+  destination_table: desktop_funnel_activation_v1${{ macros.ds_format((execution_date + macros.timedelta(days=-8)).to_datetime_string(), "%Y-%m-%d %H:%M:%S", "%Y%m%d") }}   # yamllint disable-line rule:line-length
```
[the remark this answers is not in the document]
scholtzan: I changed the logic to be backward-looking. This way macros for the `submission_date` parameter aren't needed.
Since we won't have user-facing views, I kept the parameter for the table partition.

scholtzan:
I did a bit more cleaning up. Should be ready for another look

dags.yaml
```diff
@@ -419,3 +419,19 @@ bqetl_adm_engagements_daily:
     email: ['akomar@mozilla.com']
     retries: 2
     retry_delay: 30m
+
+bqetl_desktop_funnel:
+  description: >
+    This DAG schedules desktop funnel queries used to power the
+    [Numbers that Matter dashboard](https://protosaur.dev/numbers-that-matter/)
+  schedule_interval: 0 3 * * *
```
jklukas, line 427: We should probably move this to 04:00 since it's a use-case-specific DAG that doesn't have downstream dependencies.

sql/moz-fx-data-shared-prod/telemetry_derived/desktop_funnel_activation_v1/query.sql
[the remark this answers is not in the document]
jklukas: `raw_info` is still a windowed query over 8 days and I don't think there's any need for that to exist.

Based on my mental model of the intent here, I think this query should be a join between (using terminology from the [forward-looking windows docs](https://docs.telemetry.mozilla.org/cookbooks/clients_last_seen_bits.html#retention-forward-looking-windows):
- Deduped new profile pings from day 0
- clients_last_seen from day 6

The clients_last_seen row from day 6 contains all the data we need for understanding days 0 through 6.

It appears, though, that perhaps there's an extra offset here. So we receive a new profile ping on day 0, then we want to measure activity on days 1 through 7 (is that correct?). In which case this becomes a join between:
- Deduped new profile pings from day 0
- clients_last_seen from day 7

I still think that we should populate partitions based on the Airflow run date (for least operational surprise) and that we should provision a view (this can be under telemetry_derived to have it not be user-facing) to handle changing the date. In this case we have:

- `date` exposed in the view = `DATE_SUB(submission_date, interval 7 day)` (and the view does not expose submission_date)
- new_profile `WHERE submission_date = DATE_SUB(@submission_date, interval 7 day)`
- clients_last_seen `WHERE submission_date = @submission_date`
- `coalesce(udf.bitcount_lowest_7(days_seen_bits), 0) >= 5 AS activated`

It is entirely possible that I am missing nuance here, but I want to make sure that complexity in this query is necessary to minimize possibility of subtle bugs.

jklukas (changes requested):

sql/moz-fx-data-shared-prod/telemetry_derived/desktop_funnel_activation_v1/query.sql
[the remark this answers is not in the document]
scholtzan: I used the existing data and queries to check if there is an offset and the join should be on clients_last_seen from day 6 to reproduce the number in the existing queries.

sql/moz-fx-data-shared-prod/telemetry_derived/desktop_funnel_activation_6_day_offset_v1/metadata.yaml
```diff
@@ -0,0 +1,12 @@
+description: >
+  Measures the number of activated new profiles in the Firefox desktop
+  funnel. The acitvation date is offset by 6 days to reduce complexity 
+  for scheduling in Airflow. For querying desktop funnel activation data, 
+  use telemetry_derived.desktop_funnel_activation_v1 instead.
+friendly_name: Firefox Desktop Funnel Activation 6 Day Offset
```
scholtzan, line 6: Naming is hard. I'm open to better suggestions for this query.

sql/moz-fx-data-shared-prod/telemetry_derived/desktop_funnel_activation_v1/view.sql
```diff
@@ -0,0 +1,16 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry_derived.desktop_funnel_activation_v1`
+AS
+SELECT
+  DATE_SUB(submission_date, INTERVAL 6 day) AS date,
```
scholtzan, line 5: View which will adjust the date offset

sql/moz-fx-data-shared-prod/telemetry_derived/desktop_funnel_activation_6_day_offset_v1/metadata.yaml
jklukas: This seems roughly in line with some existing terminology for the `smoot_usage_day_6` view, etc. Those tables for GUD rely heavily on looking at bit patterns at different offsets for forward-looking windows. I think it's good to codify the offset in the table name here for least surprise.

jklukas (approved):
Thanks for bearing with all the conversation here 👍

PR #1675, opened 2021-01-14
execution_date to to_datetime_string in experiment_enrollment_aggregates_hourly scheduling

scholtzan:
I set up a local Airflow instance to test this. For a `2021-01-13T00:00:00Z` run timestamp the partition date part got rendered into `2021011223`

Before it was failing with ` ERROR - strptime() argument 1 must be str, not Pendulum`

### DIFF
--- a/script/experiments/export_experiment_monitoring_data.py
+++ b/script/experiments/export_experiment_monitoring_data.py
@@ -33,7 +33,6 @@ parser.add_argument(
 )
 parser.add_argument(
     "--datasets",
-    required=True,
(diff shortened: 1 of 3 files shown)


script/experiments/export_experiment_monitoring_data.py
```diff
@@ -33,7 +33,6 @@
 )
 parser.add_argument(
     "--datasets",
-    required=True,
```
scholtzan, line 36: Not required since `default` is set

jklukas (approved):

PR #1670, opened 2021-01-14
Improve ergonomics of creating/scheduling queries with long names

wlach:
Before, you would get a confusing error about the autogenerated task
name being too long with `./bqetl query schedule`. In theory, you
should have been able to get around this with the `--task` parameter,
but in practice it would autogenerate a name when getting the list
of DAGs, causing a failure that could only be worked around by
manually editing the metadata.

This PR moves the --task-name parameter to query creation and fails
in a (hopefully obvious) way if the autogenerated one would be too long
(if it isn't specified).

PR #1676, opened 2021-01-14
Fix bqetl query schedule --task_name option

scholtzan:
Fixes problems described in https://github.com/mozilla/bigquery-etl/pull/1670


bigquery_etl/cli/query.py
```diff
@@ -240,7 +241,7 @@ def schedule(name, sql_dir, project_id, dag, depends_on_past, task_name):
 
     sql_dir = Path(sql_dir)
 
-    dags = get_dags(None, sql_dir.parent / "dags.yaml")
+    dags = DagCollection.from_file(sql_dir.parent / "dags.yaml")
```
scholtzan, line 244: `get_dags` parses all DAG and task definitions. Since we do not use any of the parsed task definitions here, we can use `DagCollection.from_file` which parses only the DAG definitions.

wlach (approved):

PR #1677, opened 2021-01-15
Move non-user facing monitoring tables to monitoring_derived

scholtzan:
I saw this mentioned in https://bugzilla.mozilla.org/show_bug.cgi?id=1671934#c1

This PR creates a `monitoring_derived` dataset which moves the non-user facing tables into this dataset. This way the monitoring dataset will also follow the general structure used for all/most other dataset.

One thing to note here: for the user-facing views the versions have been removed from the view name. This will break some dashboards that query the old views and tables that had the version number as part of their name. So these would need to be updated once merged. 
Alternatively, to prevent anything from breaking, the versions could be added back to the view names, however since nothing critical seems to depend on these views I think we should rather follow the general practice of not having versions in view names.

Dry run and DAG generation are currently failing as `monitoring_derived` doesn't exist yet.


whd:
> Dry run and DAG generation are currently failing as `monitoring_derived` doesn't exist yet.

A corresponding PR needs to be filed against cloudops-infra to add a `derived` dataset definition: https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/tf/prod/envs/prod/bigquery/namespaces.tfvars.json#L287

jklukas:
As @whd says, we'll need to create the derived dataset first. Note that removing views in this PR won't actually cause the views to be removed, so that kind of naturally gives us a transition period where we can fix up consumers to hit the new view names before cleaning up the old ones.

jklukas:
Mark me for review again once we have the cloudops-infra change merged and the dataset exists.

whd:
`monitoring_derived` has been provisioned.

scholtzan:
I created the destination tables in `monitoring_derived`. Once this gets merged, the data from the old tables in `monitoring` can be copied into the new tables. There are also some dashboards/Redash queries that query from the old versioned views and tables that are in `monitoring`.

jklukas (approved):

scholtzan:
I created all the views and copied the data from the tables in the `monitoring` dataset to the corresponding tables in `monitoring_derived`

PR #1680, opened 2021-01-15
Migrate remaining enrollments grafana queries

scholtzan:
Depends on https://github.com/mozilla/bigquery-etl/pull/1675

Migrates the remaining experiment enrollment queries that are currently implemented in Grafana. Except for "Unenrollment Reasons", I'll need to find out if that is going to be part of Experimenter.

Depends on https://github.com/mozilla/telemetry-airflow/pull/1219


sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_aggregates_live/view.sql
```diff
@@ -0,0 +1,24 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry_derived.experiment_enrollment_aggregates_live`
```
scholtzan, line 2: This view replaces the existing `moz-fx-data-shared-prod.telemetry_derived.experiment_enrollment_aggregates_live` view. The advantage of simply replacing it is that queries that query the view won't break (like the queries currently implemented in Grafana)

jklukas (approved):

Issue #1413, opened 2020-10-09
Explicit CI check to make sure project IDs aren't missing for referenced tables in view definitions


whd:
https://github.com/mozilla/bigquery-etl/issues/1318
https://github.com/mozilla/bigquery-etl/issues/1409
https://github.com/mozilla/bigquery-etl/issues/1422
https://github.com/mozilla/bigquery-etl/pull/1705

jklukas:
We may be able to solve this somewhat naturally as part of implementing a `mozdata` project. Once we "cut over" to mozdata for user queries, I think bigquery-etl should put all user-facing views underneath `sql/mozdata`, and then we could have CI attempt to publish the views in a test project rather than using the shared-prod dryrun service.

whd:
Given related issues like https://github.com/mozilla/bigquery-etl/issues/1422#issuecomment-706827788 it does seem like the easiest thing to do here would be to attempt to publish views on CI. Dry run is clearly insufficient as we seem to be having an increasing number of these failures lately.

relud:
this CI check can be accomplished by modifying the dry-run script to not provide a default project or dataset when dry-running view definitions

relud:
> this CI check can be accomplished by modifying the dry-run script to not provide a default project or dataset when dry-running view definitions

I was wrong, the bigquery API doesn't have a default project parameter. When project is missing it defaults to the project where the query was submitted.

This means a dry run can only detect if `datasetId` was not specified or does not exist.

A workaround here would be to dry run views in a project with no BQ datasets, and translate the error from `Not
  found: Dataset some-project:some_dataset` to something like `Dataset name "some_dataset" missing project in view definition`.

scholtzan:
https://github.com/mozilla/bigquery-etl/pull/1711 adds a step to the CI for validating view definitions. It checks that view names have project IDs and makes sure directory the structure matches the view content.

fbertsch:
Still an issue. This was uncaught: https://github.com/mozilla/bigquery-etl/pull/1797/files

jklukas:
[...]

PR #1681, opened 2021-01-18
Add option for publishing user-facing views only

scholtzan:
This tries to address https://github.com/mozilla-services/cloudops-infra/pull/2808/files#r556195486 and would allow to publish user-facing views only by adding the `--user-facing-only` flag:

https://github.com/mozilla-services/cloudops-infra/blob/4a331d886628ea4013dd7833569095d12361811f/projects/data-pioneer/Jenkinsfile.bigquery#L219 would change to `./script/publish_views --user-facing-only --target-project ${MozdataProjectId} /tmp/sql`.


jklukas (approved):
LGTM!

bigquery_etl/view/publish_views.py
```diff
@@ -135,8 +153,12 @@ def main(**kwargs):
     for target in args.target:
         if os.path.isdir(target):
             for root, dirs, files in os.walk(target):
-                if "view.sql" in files:
-                    sql_files.append(os.path.join(root, "view.sql"))
+                dataset_dir = os.path.dirname(root)
```
whd, line 156: One thing to note is that I'm not sure which places in this repo leverage the actual directory structure as source of truth (https://github.com/mozilla/bigquery-etl/pull/1663 being a case where the directory was not a source of truth). `publish_views` I think until this PR did not rely on directory structure and instead used the view content in e.g. the logic for deciding whether to treat a view specially in conjunction with `--target-project`. It may be preferable to leverage the same filtering mechanism for both operations, or to enforce (perhaps in CI) a constraint that makes sure directory structure matches the (authoritative) view content.

This said, the main point of a flag like this is so that downstream logic doesn't have to worry about things like directory structure as much, and I think this will do the right thing.

whd (approved):

bigquery_etl/view/publish_views.py
scholtzan: Having CI check would probably be the easiest way and could at the same time solve https://github.com/mozilla/bigquery-etl/issues/1413

I think in the future we'll want to specify in the `metadata.yaml` file whether a view is user-facing or not instead of relying on the dataset name. This way we publishing views won't depend on the directory structure and we don't have to maintain the list of suffixes.

whd: > I think in the future we'll want to specify in the metadata.yaml file whether a view is user-facing or not instead of relying on the dataset name. This way we publishing views won't depend on the directory structure and we don't have to maintain the list of suffixes.

At some point the plan is to externalize the [datasets metadata](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/tf/prod/envs/prod/bigquery/namespaces.tfvars.json) to bq-etl or MPS or similar. I haven't quite figured out what this looks like, but something to note here is that whether a view is user-facing is mostly a property of the dataset containing the view and not the view itself. A system where we have dataset metadata that propagates to individual views would likely be easiest to reason about.

See https://docs.google.com/document/d/1ug7bP7-WJ6w0WypjDoSDWEdwav7suyJ6ECjGVIcRVgE/edit# for some preliminary design around this.

PR #1690, opened 2021-01-20
Add main_nightly, similar to main_1pct

jklukas:
Should provide a significant performance improvement to queries like https://sql.telemetry.mozilla.org/queries/65380/source and sidestep redash issues with errors due to dry run timeouts.


wlach (approved):
💯

jklukas:
I'm starting a backfill now using `moz-fx-data-backfill-slots` which will likely take most of a day.

PR #1692, opened 2021-01-20
Site Metrics: Expand definition of what "Firefox" is

wlach:
From: https://github.com/mozilla/bigquery-etl/pull/1674#discussion_r560237430

Include "Mozilla" in browsers that are considered to come from us, also
create a small UDF for this so we can reuse this logic in other things.


sql/moz-fx-data-shared-prod/udf/ga_is_mozilla_browser/udf.sql
```diff
@@ -0,0 +1,23 @@
+/*
+
+Determine if a browser in a Google Analytics data is one produced by Mozilla
+
+*/
+CREATE OR REPLACE FUNCTION udf.ga_is_mozilla_browser(browser STRING)
+RETURNS BOOLEAN AS (
+  CASE
+  WHEN
+    browser LIKE 'Firefox'
```
BenWu, line 10: Is using `LIKE` instead of `=` a conscious decision?  I don't think there's any functional difference since there's no pattern matching but I think `LIKE` is slightly less performant.

sql/moz-fx-data-marketing-prod/ga_derived/www_site_metrics_summary_v1/query.sql
```diff
@@ -11,7 +11,7 @@ WITH sessions_table AS (
     trafficSource.campaign,
     trafficSource.adContent AS ad_content,
     COUNT(*) AS sessions,
-    SUM(IF(device.browser != 'Firefox', 1, 0)) AS non_fx_sessions,
+    COUNTIF(udf.ga_is_mozilla_browser(device.browser)) AS non_fx_sessions,
```
BenWu, line 14:
```suggestion
    COUNTIF(NOT udf.ga_is_mozilla_browser(device.browser)) AS non_fx_sessions,
```

BenWu (changes requested):
Can you also publish the udf to dry run?

sql/moz-fx-data-shared-prod/udf/ga_is_mozilla_browser/udf.sql
wlach: No, just an error on my part. I'll change it to `=` to keep consistency with the previous behaviour

sql/moz-fx-data-marketing-prod/ga_derived/www_site_metrics_summary_v1/query.sql
wlach: 🤦 thank you

wlach:
> Looks fine other than the `NOT`. Can you also publish the udf to dry run?

How should I do this? `./bqetl routine publish`?

BenWu:
`bqetl routine publish` should work and maybe there are options to only publish that one udf.  I'm not too familiar with that part of the cli though and I would just run the sql in the bq console.

I also noticed the query and udf are in different projects so you would need to fully qualify the function call

sql/moz-fx-data-marketing-prod/ga_derived/www_site_downloads_v1/query.sql
```diff
@@ -35,7 +35,7 @@ WITH download_event_count AS (
 SELECT
   *,
   COUNTIF(download_events > 0) AS downloads,
-  COUNTIF(download_events > 0 AND browser != 'Firefox') AS non_fx_downloads,
+  COUNTIF(download_events > 0 AND NOT udf.ga_is_mozilla_browser(browser)) AS non_fx_downloads,
```
BenWu, line 38: query runs in a different project and I think it makes sense to keep the udf in shared-prod with everything else
```suggestion
  COUNTIF(download_events > 0 AND NOT `moz-fx-data-shared-prod.udf.ga_is_mozilla_browser`(browser)) AS non_fx_downloads,
```

BenWu (changes requested):

sql/moz-fx-data-marketing-prod/ga_derived/www_site_downloads_v1/query.sql
BenWu: I realize this is a confusing side effect of how the projects and bigquery-etl is set up

scholtzan:
`./bqetl routine publish sql/moz-fx-data-shared-prod/udf/ga_is_mozilla_browser/` should work to publish this specific UDF

wlach:
@BenWu I think we should be good now, after some fighting the UDF is published for dry run. How big are the tables we're working with here? Would you expect this to be a pain to backfill?

BenWu (approved):
The tables are small so it should be quick and cheap to backfill.  Would you like me to do that?

wlach:
> The tables are small so it should be quick and cheap to backfill. Would you like me to do that?

If you don't mind, that would be great! Feel free to merge this PR when it's convenient.

Issue #1678, opened 2021-01-15
Productionize fields from ccd_feature_set


jklukas:
I've been chatting with @SuYoungHong @irrationalagent @godelstheory about productionizing an experimental table for feature usage:

- [draft query](https://console.cloud.google.com/bigquery?project=moz-fx-data-derived-datasets&folder=&organizationId=&sq=630180991450:407629d0964442b3a814efcaed9837ae)
- [spreadsheet documenting the fields](https://docs.google.com/spreadsheets/d/1zlGi4C6qqEEUkyh2VFAIbyGl6y2J_AnfaH6RYNXFbx8/edit#gid=0)

jklukas:
## Extracting event counts from histograms

I understand the fields here that are looking at the '$.sum' from a histogram, but it's less clear to me what the intention is behind lines like:

```
SUM(COALESCE(ARRAY_LENGTH(`moz-fx-data-shared-prod.udf.json_extract_histogram`(payload.processes.content.histograms.video_play_time_ms).values), 0)) AS video_play_time_ms_count
```

This ARRAY_LENGTH calculation is perhaps trying to get the number of buckets that are populated? And we infer that the number of populated buckets corresponds to the number of distinct videos played? If so, I think that methodology is problematic as it assumes that every entry in the `values` array has a value of 1. But many histograms include empty buckets on the extremes, or may have multiple entries in a given bucket.

```
  (SELECT SUM(value) FROM UNNEST(mozfun.hist.extract(payload.processes.content.histograms.video_play_time_ms).values)) AS video_play_time_ms_count
```

See https://sql.telemetry.mozilla.org/queries/77400/source

Am I interpreting intent correctly?

godelstheory:
Yes, your interpretation is correct, and the `ARRAY_LENGTH` query is incorrect consequently. Your query using `SUM(value)` gets at the desired result.

irrationalagent:
FWIW I revised the query based on Jeff's feedback, and also fixed a couple of things in the events section. If I find anything else ill continue to update the link below. I can't guarantee that its complete in the sense of having every measure we might care about (it almost certainly doesn't)

https://sql.telemetry.mozilla.org/queries/77408/source

jklukas:
[...]

PR #1584, opened 2020-12-01
Add event-based usage criteria to clients_last_seen

jklukas:
This is proposed as part of the Data Warehouse [daily per-client aggregations project](https://docs.google.com/document/d/1Lml-hWiqhvUazjn_-sDF6TUvwAEA3jMG3LJbyIC7fEc/edit#) to allow event-based feature usage definitions.

This is initially presented as a draft for feedback on the approach; scheduling has not yet been implemented.

PR #1696, opened 2021-01-20
Add event ping fields to clients_daily and clients_last_seen

jklukas:
Supports the new desktop ETL structure proposed in the [daily per-client aggregations doc](https://docs.google.com/document/d/1Lml-hWiqhvUazjn_-sDF6TUvwAEA3jMG3LJbyIC7fEc/edit#). Supersedes https://github.com/mozilla/bigquery-etl/pull/1584

I'm implementing just a few event-based fields in this PR to make it easier to review the overall methodology. I'll file a follow-up after this to integrate all the event-based fields described in https://github.com/mozilla/bigquery-etl/issues/1678#issuecomment-762980255

Also, this PR does not yet update the user-facing views to reference the new joined tables. I want to get more fields integrated, perform a backfill, and do some validation before making all this available to users.


dags/bqetl_main_summary.py
```diff
@@ -197,6 +273,37 @@
...
+        external_task_id="bq_main_events",
+        execution_delta=datetime.timedelta(seconds=3600),
+        check_existence=True,
+        mode="reschedule",
+        pool="DATA_ENG_EXTERNALTASKSENSOR",
+    )
+
+    telemetry_derived__clients_daily_event__v1.set_upstream(wait_for_bq_main_events)
+    wait_for_event_events = ExternalTaskSensor(
+        task_id="wait_for_event_events",
```
scholtzan, line 288: I wonder if we should set the priority for [`event_events`](https://github.com/mozilla/telemetry-airflow/blob/ceda595853271cb71be7b4d1cacda51ebc6fe7c4/dags/copy_deduplicate.py#L65) to something >85. If Airflow, for some reason, happens to schedule `telemetry_derived__clients_daily_event__v1` and `event_events` at the same time, `telemetry_derived__clients_daily_event__v1` will have a higher priority. Likely won't make a huge difference in this single case though hopefully.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_event_v1/init.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE IF NOT EXISTS
+  clients_daily_event_v1
```
scholtzan, line 2: Not sure if we want to maintain a init.sql file for this query? I'm guessing it will change as new events get added.

Currently, the init.sql is missing `n_created_pictureinpicture` and `n_viewed_protection_report`

sql/moz-fx-data-shared-prod/telemetry_derived/clients_last_seen_event_v1/query.sql
```diff
@@ -0,0 +1,51 @@
...
+  SELECT
+    days_logged_event_bits,
+    days_used_pictureinpicture_bits,
+    days_viewed_protection_report_bits,
+    * EXCEPT (
+      days_logged_event_bits,
+      days_used_pictureinpicture_bits,
+      days_viewed_protection_report_bits,
+      submission_date
+    )
```
scholtzan, line 22: Wouldn't this be the same as?

```sql
_previous AS (
  SELECT
    * EXCEPT (
      submission_date
    )
  FROM
...
```

scholtzan (approved):

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_event_v1/init.sql
relud: yeah, I don't think this is the right place for an `init.sql`. The table doesn't self-join, so `query.sql` should be sufficient on its own.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_last_seen_event_v1/query.sql
relud: as fields are added this will change the order of the fields so that they match between `_current` and `_previous`, which is necessary for the `IF(..., _current, _previous).*` clause.

relud (approved):
r+wc

sql/moz-fx-data-shared-prod/telemetry_derived/clients_last_seen_event_v1/query.sql
jklukas: I'll document this in a comment. It's indeed surprising to see.

dags/bqetl_main_summary.py
jklukas: See https://github.com/mozilla/telemetry-airflow/pull/1232

PR #1708, opened 2021-01-22
Parallelize the generate_views script for faster execution

jklukas:


bigquery_etl/view/generate_views.py
```diff
@@ -77,81 +80,89 @@ def main():
...
-        if not os.path.exists(target_file):
-            # We put this BQ API all inside the conditional to speed up execution
-            # in the case target files already exist.
-            table = client.get_table(target)
-            replacements = ["mozfun.norm.metadata(metadata)" " AS metadata"]
-            # Add special replacements for Glean pings.
-            schema_id = table.labels.get("schema_id", None)
-            if schema_id == "glean_ping_1":
+
+def create_view_if_not_exists(client, view, tables, exclude, sql_dir):
```
jklukas, line 95: We've changed the signature here, but the body of the function is unchanged other than missing one level of indent since the function now operates on a single view rather than iterating over an object. The iteration now happens in the `pool.starmap` call.

scholtzan (approved):

whd (approved):

Issue #1110, opened 2020-07-01
Make SHARED_PROD configurable or drop views publishing from stage


whd:
I didn't have time to read https://github.com/mozilla/bigquery-etl/pull/1105 before it was merged, but it would be nice to have the `SHARED_PROD` be configurable in some way (or perhaps be `SHARED_PROJECT` so it could be used for the nonprod case).

The logic used to publish views to stage is [here](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/Jenkinsfile.bigquery.stage#L122-L135).

Today's stage deploy went out with all views failing like so:

     ERROR: /tmp/sql/org_mozilla_vrbrowser/metrics/view.sql contains a CREATE VIEW statement that does not target {SHARED_PROD}! Quitting...

(Side note: I'm guessing the fstring is supposed to replace `{SHARED_PROD}` with an actual value but it doesn't appear to.)

Admittedly running `sed s/moz-fx-data-shared-prod/moz-fx-data-shar-nonprod-efed/g` in order to get (generated) views publishing to work in stage is already a hack, but perhaps now is the time to consider the following alternatives:

1. better support for non-shared-prod in views publishing
2. dropping views publishing from stage and any projects other than shared-prod and derived-datasets (the latter of which will eventually go away)
3. move udf/views publishing to ops logic

I'm fine with dropping views publishing from stage entirely, but I generally like to have the environments be as similar as possible where they can be (even though e.g. copy-deduplicate doesn't run in stage) and autogenerated views is one of those things I think is nice to have. There are also in theory a class of errors that would be caught by views publishing to stage failing that I can at least investigate when they fail (even though they currently don't block schemas promotion to production).

The other case where this could be useful is when the central/main project is not shared e.g. pioneer. For various practical and security reasons I re-implemented the user-facing views publishing [in terraform](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-pioneer/tf/modules/bigquery/main.tf#L263) so this recent change won't affect pioneer, but my original intention was to use the bigquery-etl script to generate and publish pioneer views (until I learned that the default view uses a UDF that won't be accessible from the pioneer environment). It's unlikely (but not highly unlikely) that we end up spinning off more pioneer-like sequestered environments, and in case we do having the central/main project be configurable would help make the views scripts more usable there.

[...]

PR #1709, opened 2021-01-22
Add glean_usage ETL generation to generate_all_views

jklukas:
The new `generate_all_views` script is intended to replace `generate_views`
as the entrypoint for Jenkins. Its usage is demonstrated in the
`generate_and_publish_views` script.

This supports the move to user queries in the `mozdata` project.

### DIFF
--- a/script/generate_and_publish_views
+++ b/script/generate_and_publish_views
@@ -28,6 +28,6 @@ done
 
 # We additionally make sure we have identical view definitions in moz-fx-data-derived-datasets and mozdata
 if [[ "$TARGET_PROJECT" == "moz-fx-data-shared-prod" ]]; then
-    ./script/publish_views --target-project moz-fx-data-derived-datasets "${SQL_DIR}"
-    ./script/publish_views --target-project mozdata "${SQL_DIR}"
+    ./script/publish_views --user-facing-only --target-project moz-fx-data-derived-datasets "${SQL_DIR}"
+    ./script/publish_views --user-facing-only --target-project mozdata "${SQL_DIR}"
 fi
--- a/script/generate_and_publish_views
+++ b/script/generate_and_publish_views
@@ -2,11 +2,32 @@
 
 set -e
 
-# Fill in any missing view definitions in the moz-fx-data-shared-prod/ folder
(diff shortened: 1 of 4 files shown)


script/generate_and_publish_views
```diff
@@ -2,11 +2,32 @@
...
+esac
+done
+
+./script/generate_all_views --target-project "${TARGET_PROJECT}" --sql-dir "${SQL_DIR}"
+./script/publish_views --target-project "${TARGET_PROJECT}" "${SQL_DIR}"
+
+# We additionally make sure we have identical view definitions in moz-fx-data-derived-datasets and mozdata
+if [[ "$TARGET_PROJECT" == "moz-fx-data-shared-prod" ]]; then
+    ./script/publish_views --target-project moz-fx-data-derived-datasets "${SQL_DIR}"
+    ./script/publish_views --target-project mozdata "${SQL_DIR}"
```
scholtzan, line 32: When publishing to mozdata here should the `--user-facing-only` flag be added to ensure only user-facing views are published?

```suggestion
    ./script/publish_views --user-facing-only --target-project mozdata "${SQL_DIR}"
```

scholtzan (approved):

script/generate_and_publish_views
whd: Any time we hard-code projects in bigquery-etl we lose the flexibility to support the [nonprod](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-pioneer/Jenkinsfile.bigquery#L37) environment.

https://github.com/mozilla/bigquery-etl/issues/1110

jklukas: Indeed. This `generate_and_publish_views` script is intended to demonstrate usage. My intention is that Jenkins would simply move its existing call to `generate_views` with `./script/generate_all_views --target-project ${project} --sql-dir /tmp/sql`.

This generate_and_publish_views logic is mimicing what Jenkins does. Only for shared-prod does it do the extra publishing to mozdata and derived-datasets. This block is skipped for nonprod where we don't want to republish.

whd (approved):
:+1:

PR #1710, opened 2021-01-22
Use is_mozilla_browser udf in www landing page metrics

BenWu:


sql/moz-fx-data-marketing-prod/ga_derived/www_site_landing_page_metrics_v1/query.sql
```diff
@@ -21,7 +21,9 @@ WITH sessions_table AS (
     ad_content,
     browser,
     SUM(entrances) AS sessions,
-    SUM(CASE WHEN browser != 'Firefox' THEN entrances ELSE 0 END) AS non_fx_sessions,
+    SUM(
```
wlach, line 24: I think `COUNTIF` would be a bit clearer, but lgtm

wlach (approved):

sql/moz-fx-data-marketing-prod/ga_derived/www_site_landing_page_metrics_v1/query.sql
BenWu: I don't think there's a way to make simpler because it's summing `entrances` instead of 1

wlach: Ah, got it.

Issue #1713, opened 2021-01-25
Re-evaluate fxa_users_services_first_seen


jklukas:
cc @irrationalagent @ksiegler1 

The [firefox_accounts_derived__fxa_users_services_first_seen__v1 task in Airflow is failing consistently](https://workflow.telemetry.mozilla.org/log?execution_date=2021-01-24T01%3A30%3A00%2B00%3A00&task_id=firefox_accounts_derived__fxa_users_services_first_seen__v1&dag_id=bqetl_fxa_events). The basic error is "The query could not be executed in the allotted memory" with the top consumer listed as `sort operations used for analytic OVER() clauses: 98%`.

This `first_seen` query is also among the most expensive queries we run every day. It scans the entire history of fxa user data we keep, and we have never invested in turning it into an incremental query.

Are the `fxa_users_services_*` tables actively used? I want to make sure they're providing value before investing in performance improvements.

jklukas:
I am going to deschedule these first_seen and last_seen queries for now in https://github.com/mozilla/bigquery-etl/pull/1714

irrationalagent:
Sorry for the late reply, it has been several months since I have used this table, so I think its fine to deschedule for now.

ksiegler1:
Same here, I haven't used this table in a while.

PR #1716, opened 2021-01-25
Remove views for fxa_users_services_*

jklukas:
See https://github.com/mozilla/bigquery-etl/issues/1713#issuecomment-767122674

I will delete these views manually after merging.


sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/metadata.yaml
```diff
@@ -7,7 +7,8 @@ labels:
   application: fxa
   incremental: true
   schedule: daily
-# # Descheduled for now. Last run was 2021-01-23.
+  referenced_tables: []  # exempts this query from dry runs
```
scholtzan, line 10: `referenced_tables` only exists as part of `scheduling`. To exempt the query from dry runs, it would need to be added to `SKIP` in the `dryrun` script

scholtzan (approved):

Issue #1539, opened 2020-11-10
Support schema files in query directories


jklukas:
The proposal here is motivated by the [Daily Per-Client Aggregations](https://docs.google.com/document/d/1Lml-hWiqhvUazjn_-sDF6TUvwAEA3jMG3LJbyIC7fEc/edit#) workstream which seeks to allow us to more confidently deploy new fields to clients_daily, etc.

This repository already contains some BQ schemas as `schema.yaml` files in tests and in generated GLAM queries. We could build in direct support for schema files as part of the bqetl execution machinery with the following benefits:

- Allows a natural place to insert field descriptions, which we currently lack for derived tables and views
- We could detect schema changes as part of dry runs, providing a foothold to solve the "field addition surprise" failure mode for clients_daily and clients_last_seen

We would need to provide the following new affordances:

- A new bqetl invocation to allow easily pulling down the schema of an existing table or view, writing out a `schema.yaml` file in the appropriate directory (as a way to opt a query into this new functionality)
- A new bqetl invocation apply `schema.yaml` to update the schema of a table
- In the dryrun cloud function, return the schema of the result as part of the output; see [JobStatistics2.schema](https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#jobstatistics2); this should be present in JS as `job[0].metadata.statistics.schema` or in python as `job._job_statistics()['schema']`
- In the bqetl dryrun machinery, if `schema.yaml` exists, check it against the returned schema from the dryrun service; fail the dry run if they do not match (we'll probably need some logic to ensure stable ordering)

If we had the above, we could produce `schema.yaml` files for clients_daily, clients_last_seen, and clients_first_seen. When we add a new field to `clients_daily`, this would cause dry run to fail with a mismatch showing the mismatch with `schema.yaml`; the developer would then be sure to add that field to `schema.yaml` as part of the PR, including a description for the field, and then run `bqetl query schema apply` to push that new field to shared-prod. This will now cause clients_last_seen and clients_first_seen to fail dry runs due to the new field in clients_daily, prompting the developer to apply the same schema evolution steps to clients_last_seen.

We would document the workflow for evolving clients_daily/clients_last_seen, and hopefully the initially dryrun failure for clients_daily would jog the developer's memory to look up the docs before proceeding with a sequence of ad hoc schema updates and new dry run failures.

[...]

PR #1718, opened 2021-01-26
[WIP] Add command: `bqetl query schema`

jklukas:
A potential alternative to https://github.com/mozilla/bigquery-etl/issues/1539

With this functionality, adding a new field to clients_daily would look like
the following:

- File a PR that updates the clients_daily_v6 query.sql
- Get review (some tests are still failing)
- When ready to merge, run:

```bash
bqetl query schema --update moz-fx-data-shared-prod.telemetry_derived.clients_daily_event_v1
bqetl query schema --update moz-fx-data-shared-prod.telemetry_derived.clients_last_seen_event_v1
bqetl query schema --update moz-fx-data-shared-prod.telemetry_derived.clients_daily_v1
bqetl query schema --update moz-fx-data-shared-prod.telemetry_derived.clients_last_seen_v1
bqetl query schema --update moz-fx-data-shared-prod.telemetry_derived.clients_first_seen_v1
bqetl query schema --update moz-fx-data-shared-prod.telemetry_derived.clients_daily_joined_v1
bqetl query schema --update moz-fx-data-shared-prod.telemetry_derived.clients_last_seen_joined_v1
```

- All tests should now pass
- Merge the PR
- Incremental queries should all now run successfully.

This is currently marked as a draft. I'm seeking feedback about the general approach here before deciding on whether to fully flesh it out and document the workflow for applying schema updates to the clients_daily ETL stream.

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -1,5 +1,6 @@
 """bigquery-etl CLI query command."""
 
+import json
 import re
 import string
 import sys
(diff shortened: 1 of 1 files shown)


bigquery_etl/cli/query.py
```diff
@@ -462,6 +463,66 @@ def backfill(ctx, name, sql_dir, project_id, start_date, end_date, exclude, dry_
...
+        config = bigquery.QueryJobConfig(
+            dry_run=True,
+            default_dataset=f"{project}.{dataset}",
+            query_parameters=[
+                bigquery.ScalarQueryParameter(
+                    "submission_date", "DATE", "2000-01-01"
+                ),
+            ]
+        )
+        new_schema = client.query(sql, config)._job_statistics()['schema']
```
jklukas, line 520: Note that this relies on a non-public `_job_statistics` method to get at the schema part of the response.

scholtzan:
I do like that this option builds on the existing bigquery-etl structure by using `init.sql` files. Handling of errors, for example, the `init.sql` changes are incompatible with the existing schema etc. will be necessary.

Long-term I do think that replacing the `init.sql` files by schema files would be advantageous (most of the reasons have been mentioned in https://github.com/mozilla/bigquery-etl/issues/1539: column descriptions, schema files could be autogenerated from queries, schema files could be generated for existing tables, ...). From what I gathered, the main reason supporting schema files might not be ideal is the complexity added to the workflow making it less user friendly. The workflow proposed here seems to be almost identical, but less powerful, compared to schema files. `init.sql` also might imply to users that it is used only for initializing the table and not for ongoing table schema updates.

jklukas:
> From what I gathered, the main reason supporting schema files might not be ideal is the complexity added to the workflow making it less user friendly. The workflow proposed here seems to be almost identical, but less powerful, compared to schema files. init.sql also might imply to users that it is used only for initializing the table and not for ongoing table schema updates.

These are good points. Part of what I found problematic about checking in schema files is that modifying them by hand could be problematic. So perhaps we need some combination of schemas checked in and tooling to maintain and apply them. We'd probably need to maintain logic for diffing the schemas intelligently, though, which would be significantly more to maintain than what we have here.

jklukas:
Superseded by #1749

PR #1726, opened 2021-01-29
Fission experiment monitoring - add os+version dimensions to crash table

akkomar:


godelstheory (approved):
Thanks for making this happen so fast. This is exactly what is needed to segment by OS.

PR #1734, opened 2021-02-01
Delay VPN dag

jklukas:
We get alert emails every day about the wait_for_firefox_accounts_derived
task, since this DAG starts at midnight and the upstream DAG doesn't start
until 1:30. This reschedules the VPN DAG to occur 15 minutes after the upstream
DAG to avoid these alerts.

@relud - I'm relying on you to determine whether there would be any ill effects of delaying these queries, or if the logic relies on the timestamp rather than just the date.

### DIFF
--- a/dags.yaml
+++ b/dags.yaml
@@ -82,6 +82,16 @@ bqetl_fxa_events:
     retries: 1
     retry_delay: 10m
 
+bqetl_mozilla_vpn:
+  # Depends on bqetl_fxa_events, so run a bit after that one.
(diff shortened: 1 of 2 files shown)


relud (approved):

relud:
> @relud - I'm relying on you to determine whether there would be any ill effects of delaying these queries, or if the logic relies on the timestamp rather than just the date.

shouldn't be any ill effects, other than potentially losing some history in the UI.

PR #1735, opened 2021-02-01
Generate summaries of derived tables (fixes #1589)

### LINKED ISSUE #1589: Generate summaries of derived tables generated in this repo
In #1102, @scholtzan added auto-generated documentation for the UDFs, which are now published at https://mozilla.github.io/bigquery-etl/ 

For example:

https://mozilla.github.io/bigquery-etl/mozfun/bits28/

It would be handy to *also* generate some basic documentation for the views and derived tables that we have in this repository, as a useful intermediate step before we make this information available in the Glean dictionary ([see "phase 2" of the proposal](https://docs.google.com/document/d/1OkTWA3rsSJ0m5g9GDnxXVUMkJP-xJMQk_bDgDq-Z9xM/edit#heading=h.aw6yt7u3a2r2)). I expect that the audience for this would mostly be data engineering, although people from data science may also occasionally find this useful.

It's not completely clear to me, off the cuff, what the best initial organization would be of this information. Views are currently quite bare-bone (with no metadata at all), since they're intended mainly as a convenience mechanism for accessing the underlying table:

https://github.com/mozilla/bigquery-etl/tree/master/sql/moz-fx-data-shared-prod/telemetry/main_summary_v4

Derived tables have a few more fields, but are also fairly bare bone:

https://github.com/mozilla/bigquery-etl/tree/master/sql/moz-fx-data-shared-prod/telemetry_derived/main_summary_v4

However, we could probably augment this information in many cases with metadata that is in docs.tmo, for example:

https://docs.telemetry.mozilla.org/datasets/batch_view/main_summary/reference.html

(this might be better as a follow-on to an initial implementation)

--

For an initial attempt, just surfacing the basic metadata (description, name, owner, scheduling information) of the derived datasets (ignoring views) along with a link to the file in the repository seems like a sensible place to start. After that's done, perhaps we can figure out a strategy for displaying views at the same time. To provide some kind of loose organization for the documentation, I would propose breaking it down by GCP project, which would include:

* glam-fenix-dev/glam_etl 
* moz-fx-data-marketing-prod 
* moz-fx-data-shared-prod 

You can see the structure by looking at the contents of the `sql` directory (https://github.com/mozilla/bigquery-etl/tree/master/sql) -- note that there are some directories in there which do not refer to derived datasets.

Iinh:
Fixes #1589.

Work in progress. It's still quite rough (no styling yet...). I just wanted to get some feedback if I'm on the right track and not missing anything major.

Preview: https://iinh.github.io/bigquery-etl/.

What I did so far:
- surfacing basic metadata: name, owner, description, schedule information. E.g. https://iinh.github.io/bigquery-etl/moz-fx-data-shared-prod/account_ecosystem_restricted/
-  `*_derived` directories were excluded.
- only `metadata.yaml` and `view.sql` files were parsed. 
- I'm still not sure on the requirements for `glam-fenix-dev/glam_etl`. Since this project doesn't have metadata or view files, what info should I surface?

Tagging a few people to review, if you're available: @scholtzan, @wlach, @acmiyaguchi, @klukas. Thank you 😃


scholtzan:
Looking good so far! A couple of things:
* I wonder if it would be possible to keep the project names as they are. Currently the dashes are getting removed and they get capitalized:
Screen Shot 2021-02-01 at 10 01 58
* Also `_external`, `_bi` and `_restricted` datasets can be ignored. For a full list of suffixes that can get ignored see https://github.com/mozilla/bigquery-etl/blob/40abe8618995d9e2088617c7c5d372aad5d9c3bf/bigquery_etl/view/publish_views.py#L29
* I think we don't need to add the view SQL to the docs. In most cases it is just `SELECT * FROM ...`. Having a link to the `view.sql` file is probably good enough here
* It might also be nice to link to the "non-user-facing" query. In most cases that should correspond to `sql/<project_id>/<dataset>_derived/<view_name>_<some_version>`. I'm not sure if there are a lot of edge cases here that might make this difficult
* Some of the metadata is currently in the directory of the "non-user-facing" queries, for example, there is not metadata file for the [`telemetry.ssl_ratios`](https://github.com/mozilla/bigquery-etl/tree/master/sql/moz-fx-data-shared-prod/telemetry/ssl_ratios) view but the metadata lives in [`telemetry_derived.ssl_ratios_v1`](https://github.com/mozilla/bigquery-etl/tree/master/sql/moz-fx-data-shared-prod/telemetry_derived/ssl_ratios_v1)
* For generating the docs it might be worth looking at [Jinja templates](https://jinja.palletsprojects.com/en/2.11.x/). We use the for generating [Airflow DAGs in this repository](https://github.com/mozilla/bigquery-etl/tree/master/bigquery_etl/query_scheduling/templates) and also in a couple of other places. Having these templates instead of concatenating together bits and pieces of documentation that get written to a Markdown file might reduce the code complexity here.

wlach:
@scholtzan's feedback seems like a good place to start with. With my usability hat on, I think we should probably also look at:

* Describing what each project represents (similarly to what we already do for mozfun)
* Make sure to update the index to reflect the fact that we're now indexing all derived datasets (https://iinh.github.io/bigquery-etl/)

Also, expanding on what @scholtzan said, I think a lot of the metadata we actually care about (in terms of descriptions of the datasets) is actually in the `_derived` table definition. I think we should figure out a way to merge the views + the table definition into *one* entry, and then generating a sensible documentation artifact for the whole. For example, for `search_aggregates` we might render the following markdown:

```md
# Search Aggregates

Daily search clients, aggregated across unique sets of dimensions and partitioned by day

* table name: `mozdata.search_aggregates`
* scheduling: daily
* owners: [bewu@mozilla.com](mailto:bewu@mozilla.com)

[View source](https://github.com/mozilla/bigquery-etl/blob/master/sql/moz-fx-data-shared-prod/search/search_aggregates/view.sql) | [Table source](https://github.com/mozilla/bigquery-etl/blob/master/sql/moz-fx-data-shared-prod/search_derived/search_aggregates_v8/query.sql) | [Metadata source](https://github.com/mozilla/bigquery-etl/blob/master/sql/moz-fx-data-shared-prod/search_derived/search_aggregates_v8/metadata.yaml)
```

I believe all of this should be either in the metadata.yaml or the files themselves.

acmiyaguchi (commented):
There's good feedback from Anna and Will, here is some of my take on things (alongside some lower level review of the code).

> For generating the docs it might be worth looking at Jinja templates. 

+1 on going this route for generating the markdown documentation, it will be easier to maintain than building up the representation as a list of strings.

> I think a lot of the metadata we actually care about (in terms of descriptions of the datasets) is actually in the _derived table definition. I think we should figure out a way to merge the views + the table definition into one entry, and then generating a sensible documentation artifact for the whole. 

This doesn't exist in the metadata files afaik. There's the convention that a versioned name of the table in `_derived` has an unversioned view in the non-derived dataset e.g. `telemetry.clients_daily` references `telemetry_derived.clients_daily_v6` . It's also worth keeping in mind that reality doesn't always reflect the convention. Take for example `telemetry.crash_summary` which points to `telemetry.crash_summary_v2` in the same dataset.

Parsing SQL can be tricky, but doable if extracted table names are normalized by dropping backticks and keeping relative `{dataset}.{table}` references.

jklukas:
First, I want to say that seeing the rendered docs in this draft is very exciting. I think this has the potentially to fill a very big hole in current discoverability and usability of derived datasets.

## Document only user-facing datasets

I want to give my take on the relationship between tables and views, related to what @scholtzan and @wlach and @acmiyaguchi have said.

Per the [table layout and naming documentation](https://docs.telemetry.mozilla.org/cookbooks/bigquery/querying.html#table-layout-and-naming), views really are the user-facing contract, and I really want to encourage that viewpoint in the docs we generate here. Ideally, we would only show the user-facing views and we wouldn't show underlying tables in `_derived` datasets at all.

> I think we should figure out a way to merge the views + the table definition into one entry, and then generating a sensible documentation artifact for the whole

I definitely understand the motivation here, but I worry an effort at merging these into one entry is going to have ugly edge cases. There's not necessarily a 1:1 relationship between views and tables. In some cases, tables are intermediate ETL stages that shouldn't have a user-facing view. A view could reference more than one table.

Certainly it's the case today that views tend to have no documentation, and tables have much richer metadata. I would argue that's just showing that we need to invest in adding great descriptions for the views we're presenting to users.

> Parsing SQL can be tricky, but doable if extracted table names are normalized by dropping backticks and keeping relative {dataset}.{table} references.

I want to avoid parsing SQL. Instead, we have the luxury here that we can dry-run views, which returns a list of tables that they reference. We use that method for resolving dependencies when we do DAG generation and we could do the same here.

I think it would be a great first step to have our generated docs show only user-facing views, but have the view entry include a list of links to table definition directories, one for each table that the view references (as determined by dry-run). These links would be very similar to the `View` links that are already for each view.

## Show mozdata rather than moz-fx-data-shared-prod

We publish all views defined under `moz-fx-data-shared-prod` to `mozfun` and we are encouraging users to reference `mozfun` going forward. So, I would prefer for the heading here to be `mozfun` rather than `moz-fx-data-shared-prod`, since `mozfun` is going to be _the_ user-facing project for accessing most data. Perhaps we can defer visualizing the other projects to a future PR.

## Missing generated content

For Glean applications, we have [some machinery that generates and runs queries](https://github.com/mozilla/bigquery-etl/tree/master/bigquery_etl/glean_usage) that don't appear under the `sql` folder that's checked in, so they don't appear in these generated docs. For example, there's a user-facing `org_mozilla_ios_firefox` dataset that's missing here.

This could be more motivation for maintaining a generated-sql branch (see https://github.com/mozilla/bigquery-etl/issues/1110#issuecomment-652411078), but shouldn't block this effort. This could be another future improvement.

wlach:
> > I think we should figure out a way to merge the views + the table definition into one entry, and then generating a sensible documentation artifact for the whole
> 
> I definitely understand the motivation here, but I worry an effort at merging these into one entry is going to have ugly edge cases. There's not necessarily a 1:1 relationship between views and tables. In some cases, tables are intermediate ETL stages that shouldn't have a user-facing view. A view could reference more than one table.

Makes sense, I'm more concerned about presenting a coherent and easy-to-understand picture to end-users-- I'm pretty open-minded in how we get there. If that's some combination of better documentation inside the views along with some metadata generated during dry-run, fine by me.

jklukas:
> Should I file issues for these 2 tasks so we can keep track of them?

I agree that those both should be out of scope for this PR, and it would be great if you could file issues to track them!

bigquery_etl/docs/generate_docs.py
```diff
@@ -5,13 +5,20 @@
...
 
 from bigquery_etl.util import standard_args
 import yaml
 
-DEFAULT_PROJECTS_DIRS = ["sql/mozfun/"]
+
+DEFAULT_PROJECTS_DIRS = [
+    "sql/mozfun/",
+    "sql/moz-fx-data-marketing-prod/",
+    "sql/moz-fx-data-shared-prod/",
```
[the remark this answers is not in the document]
Iinh: We're only generating docs for 2 repositories right now: mozfun and moz-fx-data-shared-prod, and we use this list to filter out the rest of the projects.

Iinh:
So another round of WIP that addresses the issues mentioned (I apologize if this PR seems all over the place!): using Jinja templates to generate docs for mozdata, fixing nav bar items, and updating main page. Please let me know if I've missed anything. Preview: https://iinh.github.io/bigquery-etl/.

bigquery_etl/docs/mozdata/generate_mozdata_docs.py
```diff
@@ -0,0 +1,65 @@
...
+def generate_mozdata_docs(out_dir, project_dir):
+    """Generate documentation for mozdata."""
+    project_doc_dir = Path(os.path.join(out_dir, "mozdata"))
+
+    data_sets = [
+        item
+        for item in os.listdir(project_dir)
+        if os.path.isdir(os.path.join(project_dir, item))
+        and all(name not in item for name in NON_USER_FACING_DATASET_SUFFIXES)
+    ]
```
acmiyaguchi, line 26: Using Path methods can be more concise:

```suggestion
    data_sets = [
        item
        for item in Path(project_dir).glob("**/*")
        if item.is_dir()
        and all(name not in item.name for name in NON_USER_FACING_DATASET_SUFFIXES)
    ]
```

bigquery_etl/docs/mozdata/generate_mozdata_docs.py
```diff
@@ -0,0 +1,65 @@
...
+
+    for table in data_sets:
+        output = []
+        source_urls = {}
+
+        with open(project_doc_dir / f"{table}.md", "w") as dataset_doc:
+            # Manually set title to avoid Mkdocs from auto-formatting file names
```
acmiyaguchi, line 33: Do you have a link to a specific issue?

bigquery_etl/docs/mozdata/generate_mozdata_docs.py
```diff
@@ -0,0 +1,65 @@
...
+    "_external",
+    "_bi",
+    "_restricted",
+)
+SOURCE_URL = "https://github.com/mozilla/bigquery-etl/blob/master"
+
+
+def generate_mozdata_docs(out_dir, project_dir):
+    """Generate documentation for mozdata."""
+    project_doc_dir = Path(os.path.join(out_dir, "mozdata"))
```
acmiyaguchi, line 19: If you're using Path, you can often avoid using os.

```suggestion
    project_doc_dir = Path(out_dir) / "mozdata"
```

bigquery_etl/docs/mozdata/generate_mozdata_docs.py
```diff
@@ -0,0 +1,65 @@
...
+
+    for table in data_sets:
+        output = []
+        source_urls = {}
+
+        with open(project_doc_dir / f"{table}.md", "w") as dataset_doc:
+            # Manually set title to avoid Mkdocs from auto-formatting file names
+            dataset_doc.write(f"---\ntitle: {table}\n---\n\n")
+
+            for root, dirs, files in os.walk(f"{project_dir}{table}"):
```
acmiyaguchi, line 36: It's better to explicitly use `os.path.join(project_dir, table)` or `Path(project_dir) / table` so you don't have to worry whether the directory being passed as a string ends with a separator or not.

bigquery_etl/docs/mozdata/generate_mozdata_docs.py
```diff
@@ -0,0 +1,65 @@
...
+                if dirs == []:
+                    dataset_name = root.split("/")[-1]
+                    source_urls["Table"] = f"{SOURCE_URL}/{root}"
```
jklukas, line 39:
```suggestion
                    source_urls["Table"] = f"{SOURCE_URL}/{root}"
```

This renders a link labeled "Table Source", which is confusing to me, since in most cases it's a view being defined rather than a table. This could perhaps be "Source Directory".

jklukas: I suppose "View Source" is also a bit problematic because it can be read as a verb, as in "click this link to view the source".

So perhaps the links can be `Source Directory | Metadata File | View Definition`

bigquery_etl/docs/generate_docs.py
```diff
@@ -74,9 +75,13 @@ def main():
     """Generate documentation for project."""
     args = parser.parse_args()
     out_dir = os.path.join(args.output_dir, "docs")
-    if os.path.exists(out_dir):
+    override_dir = os.path.join(args.output_dir, "overrides")
```
acmiyaguchi, line 78: What is the purpose of the overrides directory?

jklukas (commented):
The new homepage is great. This is making us look more and more professional by the moment.

While datasets are listed alphabetically, the ordering of views within a dataset page looks random. We likely want to sort these in alphabetical order as well.

I also commented on the names for the links. I'm happy to see, though, that the `static` dataset renders nicely here; it is the one case where we expect to have actual tables in `mozdata` rather than views, and presentation is pretty reasonable.

I would be happy to see this merged once the above two issues are addressed.

bigquery_etl/docs/mozdata/generate_mozdata_docs.py
```diff
@@ -0,0 +1,65 @@
...
+NON_USER_FACING_DATASET_SUFFIXES = (
+    "_derived",
+    "_external",
+    "_bi",
+    "_restricted",
+)
+SOURCE_URL = "https://github.com/mozilla/bigquery-etl/blob/master"
+
+
+def generate_mozdata_docs(out_dir, project_dir):
```
acmiyaguchi, line 17: Is there something that makes this specific to mozdata, instead of being a general purpose function?

Iinh: yeah how we generate docs for mozfun and mozdata is different (e.g. you need to load examples for mozfun, when it's not needed for mozdata). Also the main function was getting quite long, so I thought it'd be better to abstract this out.

bigquery_etl/docs/generate_docs.py
Iinh: It's so that we can override the default home page. I made some customization to the main page so an override was needed. I will add a comment here to explain it.

bigquery_etl/docs/mozdata/generate_mozdata_docs.py
Iinh: @scholtzan suggested fixing the mkdocs' default behavior of removing underscores and capitalizing file names in her comment above, and this is one way to fix it. I will add more comment for better clarification. 
image

bigquery_etl/docs/mozdata/generate_mozdata_docs.py
acmiyaguchi: It seems like this function would work for something like `moz-fx-data-marketing-prod`, if we wanted to generate docs there too.

wlach (commented):
Don't think I have anything to add about the presentation above what others have mentioned at this point. Looking forward to the followup PRs to fill this out. :)

bigquery_etl/docs/mozdata/generate_mozdata_docs.py
Iinh: I tried using Path here, but realized that it would not work since I want the output to be a list of strings, and if I use Path I would get a list of Path objects instead. To make use of the Path objects I'd have to then convert it to string, split by "/" to get just the data set names, which defeats the purpose of reducing complexity.

bigquery_etl/docs/mozdata/generate_mozdata_docs.py
Iinh: Good point! I changed the function name to `generate_derived_dataset_docs`. Hope that's better :)

bigquery_etl/docs/mozdata/generate_mozdata_docs.py
acmiyaguchi: You can use `item.name` to get the filename, `item.parent.name` to get the directory name, or `str(item)` to get the path name.

Iinh:
@jklukas Could I get some feedback on this rendered output https://iinh.github.io/bigquery-etl/. Views in datasets are now sorted alphabetically and I also changed the source link display. Thanks :)

bigquery_etl/docs/overrides/home.html
```diff
@@ -0,0 +1,74 @@
...
+            alt=""
+            width="800"
+            height="800"
+            draggable="false"
+          >
+        </div>
+                       <!-- Hero content -->
+        <div class="tx-hero__content">
+          <h2>BigQuery ETL is Mozilla's framework for creating derived datasets and user-defined functions in BigQuery.</h2>
+          <p>At present, this site contains documentation for the Google Cloud Platform projects <code>mozdata</code> -- the primary home for user analysis, and <code>mozfun</code> -- the user-defined functions (UDFs) available in BigQuery.</p>
```
scholtzan, line 48: Nit: em dashes

```suggestion
          <p>At present, this site contains documentation for the Google Cloud Platform projects <code>mozdata</code> — the primary home for user analysis, and <code>mozfun</code> — the user-defined functions (UDFs) available in BigQuery.</p>
```

bigquery_etl/docs/mozdata/generate_mozdata_docs.py
Iinh: data_sets = [
        item.parent.name
        for item in Path(project_dir).glob("**/*")
        if item.is_dir()
        and all(name not in str(item.parent.name) for name in NON_USER_FACING_DATASET_SUFFIXES)
    ]

Thanks for the tips! I'm now running my script with the above code but it's taking much longer to run the scripts (5 minutes vs 10 seconds). Do you know if it's normal to downgrade in performance with `Path.glob` compared to `os.listdir` when looping like this?

acmiyaguchi: No, there shouldn't be a significant difference. Is the result correct? `item.is_dir()` means that you probably want `item.name` as the result instead of `item.parent.name`. Another explaination I can think of is that `project_dir` is set to the root, and you're traversing all the files in a virtualenv (which os.walk would be doing too, hmm).

jklukas:
> Could I get some feedback on this rendered output https://iinh.github.io/bigquery-etl/. Views in datasets are now sorted alphabetically and I also changed the source link display. Thanks :)

Yes, current version looks great and has addressed my concerns. +1 from me. I'll defer formal approval to those who are deeply reviewing the implementation.

bigquery_etl/docs/mozdata/generate_mozdata_docs.py
Iinh: Yes the result is correct. `item.parent.name` because I only want the derived dataset names, not all the view names. Is it okay to use my approach instead since it's much quicker? I know it's not not as concise, would it help if I provide more comments?

acmiyaguchi: Yes, it's a nitpicky thing (mostly because I think the Path interface is generally nicer to use). Feel free to use what works; there's no definitive way of doing a thing. I am very suspicious of the difference in run time however.

bigquery_etl/docs/overrides/home.html
Iinh: Thank you. This is now fixed.

docs/mozdata/introduction.md
```diff
@@ -0,0 +1 @@
+`mozdata` is the main GCP project that includes datasets for user analysis as of February 2021. This project is filled with views that reference underlying tables in `moz-fx-data-shared-prod`. Users can use data tools like [STMO](https://sql.telemetry.mozilla.org/) to issue queries in this project.
```
jklukas, line 1:
```suggestion
`mozdata` is the main GCP project that includes datasets for user analysis as of February 2021. This project is filled with views that reference underlying tables in `moz-fx-data-shared-prod`. Users can use data tools like [STMO](https://sql.telemetry.mozilla.org/) to issue queries in this project.

For more context, see the [Projects, Datasets, and Tables in BigQuery](https://docs.telemetry.mozilla.org/cookbooks/bigquery/querying.html#projects-datasets-and-tables-in-bigquery) article on docs.telemetry.mozilla.org.
```

It would be good to cross-link to DTMO here. We should reference these docs from that DTMO page too once this PR is merged.

bigquery_etl/docs/derived_datasets/generate_derived_dataset_docs.py
```diff
@@ -0,0 +1,72 @@
...
+            # Manually set title to prevent Mkdocs from removing
+            # underscores and capitalizing file names
+            # https://github.com/mkdocs/mkdocs/issues/1915#issuecomment-561311801
+            dataset_doc.write(f"---\ntitle: {table}\n---\n\n")
+
+            for root, dirs, files in os.walk(Path(project_dir) / table):
+                # show views in an alphabetical order
+                dirs.sort()
+
+                if dirs == []:
```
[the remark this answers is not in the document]
Iinh: Thank you for these great tips!

scholtzan (approved):
This looks great 🎉

bigquery_etl/docs/derived_datasets/generate_derived_dataset_docs.py
```diff
@@ -0,0 +1,72 @@
...
+        source_urls = {}
+        with open(project_doc_dir / f"{table}.md", "w") as dataset_doc:
+            # Manually set title to prevent Mkdocs from removing
+            # underscores and capitalizing file names
+            # https://github.com/mkdocs/mkdocs/issues/1915#issuecomment-561311801
+            dataset_doc.write(f"---\ntitle: {table}\n---\n\n")
+
+            for root, dirs, files in os.walk(Path(project_dir) / table):
+                # show views in an alphabetical order
+                dirs.sort()
```
acmiyaguchi, line 42: Does this actually do anything? The logic seems like it just skips over the directories if they exist.

acmiyaguchi (approved):
Looks good to me, if this code reflects the current state of https://iinh.github.io/bigquery-etl/

bigquery_etl/docs/derived_datasets/generate_derived_dataset_docs.py
Iinh: You mean `dirs.sort()`? It sorts the directories that we're looping so when we extract info from root and files things will be parsed to the template in an alphabetical order

acmiyaguchi: Ah I see, it's mutating shared state. It wasn't obvious to me at first glance.

PR #1738, opened 2021-02-02
Fix #1729 - Add timespan to metrics for GLAM ETL queries for Glean

### LINKED ISSUE #1729: Add timespans to GLAM ETL for Glean
The ETL for GLAM is missing timespans, which should be added to the ETL. See https://github.com/mozilla/glam/issues/1141

https://github.com/mozilla/bigquery-etl/blob/40abe8618995d9e2088617c7c5d372aad5d9c3bf/bigquery_etl/glam/clients_daily_scalar_aggregates.py#L93

https://mozilla.github.io/glean/book/user/metrics/timespan.html

Timespans should be added and the ability to backfill a limited set of metrics should be investigated (to avoid expensive backfill spanning months).

acmiyaguchi:
Fixes #1729 by adding timespans to the set of metrics. It turns out that it's not as simple as adding it to the set of unlabeled metrics due to the structure:

```sql
select metrics.timespan.startup_timeline_framework_start
from mozdata.org_mozilla_fenix.startup_timeline
where date(submission_timestamp) = "2021-01-30"
```

```json
[
  {
    "startup_timeline_framework_start": {
      "time_unit": "nanosecond",
      "value": "370000000"
    }
  },
  {
    "startup_timeline_framework_start": {
      "time_unit": "nanosecond",
      "value": "740000000"
    }
  },
...
]
```

### DIFF
--- a/bigquery_etl/glam/clients_daily_scalar_aggregates.py
+++ b/bigquery_etl/glam/clients_daily_scalar_aggregates.py
@@ -90,7 +90,7 @@ def get_scalar_metrics(schema: Dict, scalar_type: str) -> Dict[str, List[str]]:
     """
     assert scalar_type in ("unlabeled", "labeled")
     metric_type_set = {
-        "unlabeled": ["boolean", "counter", "quantity"],
+        "unlabeled": ["boolean", "counter", "quantity", "timespan"],
         "labeled": ["labeled_counter"],
     }
     scalars: Dict[str, List[str]] = {
--- a/bigquery_etl/glam/clients_daily_scalar_aggregates.py
+++ b/bigquery_etl/glam/clients_daily_scalar_aggregates.py
@@ -64,17 +64,20 @@ def get_unlabeled_metrics_sql(probes: Dict[str, List[str]]) -> str:
         )
 
     for metric_type, _probes in probes.items():
+        # timespans are nested within an object that also carries the unit of
+        # of time associated with the value
+        suffix = ".value" if metric_type == "timespan" else ""
         for probe in _probes:
             for agg_func in ["max", "avg", "min", "sum"]:
                 probe_structs.append(
                     (
                         f"('{probe}', '{metric_type}', '', '{agg_func}', "
-                        f"{agg_func}(CAST(metrics.{metric_type}.{probe} AS INT64)))"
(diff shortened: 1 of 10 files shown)


acmiyaguchi:
I can confirm that timespans are being aggregated in the latest nightly aggregates.

```sql
SELECT
  * except(scalar_aggregates)
FROM
  `glam-fenix-dev.glam_etl_dev.org_mozilla_fenix__view_clients_daily_scalar_aggregates_v1`,
  unnest(scalar_aggregates) as agg
where agg.metric_type = "timespan"
LIMIT
  1000
```

However, I'm not seeing [`startup_timeline_framework_start`](https://dictionary.protosaur.dev/apps/fenix/tables/startup-timeline?search=metrics.timespan.startup_timeline_framework_start) in the set of aggregates for nightly. In the ping tables:"

```sql
SELECT metrics.timespan.startup_timeline_framework_start.*
FROM org_mozilla_firefox.startup_timeline t
WHERE metrics.timespan.startup_timeline_framework_start.value IS NOT NULL
  AND date(submission_timestamp) = "2021-02-01"
LIMIT 100
```

The following query is empty:

```sql
  * except(scalar_aggregates)
FROM
  `glam-fenix-dev.glam_etl_dev.org_mozilla_fenix__view_clients_daily_scalar_aggregates_v1`,
  unnest(scalar_aggregates) as agg
where agg.metric = "startup_timeline_framework_start"
LIMIT
  1000
```

Strangely enough, it looks like the startup-timeline ping is nowhere to be seen:

```
SELECT
  client_id, ping_type, submission_date
FROM
  `glam-fenix-dev.glam_etl_dev.org_mozilla_fenix__view_clients_daily_scalar_aggregates_v1`
where ping_type like "%startup%"
LIMIT
  1000
```

This is weird because the query is visible in the checked in files.

acmiyaguchi:
The reason that `startup_timeline_framework_start` is missing is because the `startup-timeline` ping does not have a client_id, which filters out all of the data points.

acmiyaguchi:
I wrote up https://github.com/mozilla/glam/issues/1141#issuecomment-772098108 to discuss pings that are missing client ids. It's out of scope for this PR, and an entirely different issue from what I originally expected.

As far as backfilling the timespan data, it might require a separate script. I think we can append backfilled data to the aggregates table with a separate backfill script that scans over history for a specific metrics.

BenWu (approved):

PR #1743, opened 2021-02-02
Add "truncate version" UDF

wlach:
This should be useful for truncating operating system versions, both in derived datasets and queries.

Based partly on discussions we had in #data-help a while back: https://mozilla.slack.com/archives/C4D5ZA91B/p1611337464017000


jklukas:
Note that we have [os version normalization logic in the pipeline](https://github.com/mozilla/gcp-ingestion/blob/4fb9201adb1cb6eb46ab5a2af43ef5f1532aae52/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/NormalizeAttributes.java#L76-L91) which retains just the numeric parts up to a patch version. I'd be comfortable with having an `mozfun.norm.os_version` that exactly matches that behavior. We should keep any normalization decisions here as close to that as is feasible.

But this function is making a decision to truncate to at most minor version. I think the `truncate_version` name is not specific enough. I would be more comfortable with `mozfun.norm.minor_version`, although that sounds like it's extracting _only_ the minor part of the version string. Perhaps `mozfun.norm.major_minor`? Or we provide only `mozfun.norm.os_version` and expect consumers to use that with some additional logic to drop patch version.

wlach:
> But this function is making a decision to truncate to at most minor version. I think the `truncate_version` name is not specific enough. I would be more comfortable with `mozfun.norm.minor_version`, although that sounds like it's extracting _only_ the minor part of the version string. Perhaps `mozfun.norm.major_minor`? Or we provide only `mozfun.norm.os_version` and expect consumers to use that with some additional logic to drop patch version.

I'd argue hard to include this function in some form, as it was not at all obvious to me how to do this until I asked. `major_minor` (or maybe `major_minor_version`?) sound fine to me.

sql/mozfun/norm/truncate_version/udf.sql
```diff
@@ -0,0 +1,10 @@
+CREATE OR REPLACE FUNCTION norm.truncate_version(os_version STRING)
+RETURNS STRING AS (
+  REGEXP_REPLACE(os_version, r"^([0-9]+.[0-9]+).*", "\\1")
+);
+
+-- Tests
+SELECT
+  assert.equals("16.1", norm.truncate_version("16.1.1")),
+  assert.equals("10", norm.truncate_version("10")),
+  assert.equals("5.1", norm.truncate_version("5.1.0-ubuntu-foobar"))
```
acmiyaguchi, line 10: Can you add one more test case that tests something that doesn't fit the version pattern at all, like `foo-bar`, or `10-a`? I expect it to return null.

acmiyaguchi (commented):
Maybe `norm.truncate_major_minor`?

jklukas:
I vote for `norm.major_minor_version`.

wlach:
How about keeping it as `truncate_version` but adding a parameter-- either "major" or "minor"?

sql/mozfun/norm/truncate_version/udf.sql
```diff
@@ -0,0 +1,23 @@
...
+  END
+);
+
+-- Tests
+SELECT
+  assert.equals("16.1", norm.truncate_version("16.1.1", "minor")),
+  assert.equals("16", norm.truncate_version("16.1.1", "major")),
+  assert.equals("10", norm.truncate_version("10", "minor")),
+  assert.equals("5.1", norm.truncate_version("5.1.5-ubuntu-foobar", "minor")),
+  assert.equals(NULL, norm.truncate_version("foo-bar", "minor"))
```
acmiyaguchi, line 23: Another test for completeness please, where the truncation_level is neither "minor" nor "major".

acmiyaguchi, line 23: I think the test framework is casting the null to an number by default:

```suggestion
  assert.equals(CAST(NULL AS STRING), norm.truncate_version("foo-bar", "minor"))
```

acmiyaguchi (commented):
This looks good to me, although I'd personally default to `split(os_version, ".")[offset(0)]` over `mozfun.norm.truncate_version(os_version, "major")`.

acmiyaguchi (approved):
r+ with passing tests.

wlach:
> This looks good to me, although I'd personally default to `split(os_version, ".")[offset(0)]` over `mozfun.norm.truncate_version(os_version, "major")`.

Sure, but I think it's good to have this option for people who are less fluent in SQL.

jklukas (approved):
This is reasonably self-documenting with the `truncate_version` name and the truncation level passed in as a string 👍

sql/mozfun/norm/truncate_version/udf.sql
```diff
@@ -0,0 +1,24 @@
...
+);
+
+-- Tests
+SELECT
+  assert.equals("16.1", norm.truncate_version("16.1.1", "minor")),
+  assert.equals("16", norm.truncate_version("16.1.1", "major")),
+  assert.equals("10", norm.truncate_version("10", "minor")),
+  assert.equals("5.1", norm.truncate_version("5.1.5-ubuntu-foobar", "minor")),
+  assert.equals("5.1.5-ubuntu-foobar", norm.truncate_version("5.1.5-ubuntu-foobar", "patch")),
+  assert.equals(CAST(NULL AS STRING), norm.truncate_version("foo-bar", "minor"))
```
wlach, line 24: @acmiyaguchi It appears that the tests are failing because this UDF actually returns `foo-bar` in this case. Is that acceptable behaviour or do you think it should return `NULL`? I can see pros/cons to each.

jklukas: My preference would be for it to return NULL for values that do not start with numbers.

sql/mozfun/norm/truncate_version/udf.sql
```diff
@@ -0,0 +1,24 @@
+CREATE OR REPLACE FUNCTION norm.truncate_version(os_version STRING, truncation_level STRING)
+RETURNS STRING AS (
+  CASE
+  WHEN
+    truncation_level = "minor"
+  THEN
+    REGEXP_REPLACE(os_version, r"^([0-9]+[.]?[0-9]+).*", "\\1")
```
acmiyaguchi, line 7: Behavior might be more inline with extract than replace:

```suggestion
    REGEXP_EXTRACT(os_version, r"^([0-9]+[.]?[0-9]+).*")
```

sql/mozfun/norm/truncate_version/udf.sql
wlach: That's what we settled on

PR #1748, opened 2021-02-03
Add app_name and app_id to Experimenter tables

tdsmith:
Facilitates https://github.com/mozilla/jetstream/issues/456.


scholtzan (approved):
I'll add the new fields to the table schema and will then merge

PR #1749, opened 2021-02-03
Commands for validating queries using schemas

scholtzan:
This is yet another potential alternative to https://github.com/mozilla/bigquery-etl/issues/1539 and https://github.com/mozilla/bigquery-etl/pull/1718

Adding new fields to `clients_daily_v6.sql` would look like:
1. Edit `clients_daily_v6.sql` and add the field
2. Run `bqetl query schema generate telemetry_derived.clients_daily_v6`. This generates or updates the `schema.json` file this is part of the query directory.
3. File a PR
4. Tests are still failing
  * Reviewer or user (if approved) can update the table schema by running `bqetl query schema deploy telemetry_derived.clients_daily_v6`
  * Rerun tests, will fail again until all dependencies have been updated
5. All tests should now pass
6. Merge the PR
7. Queries should run successfully

*Future improvements*

Ideally, when a query gets updated all the queries that reference it should get updated automatically too when running the `bqetl query schema generate` command. This turns out to be tricky. One way I could imagine this working is by deploying updated schemas to the `bigquery-etl-integration-test` (or some other) project, let all queries reference this project and then dry run the queries to get their schema updates. Repeat until all schemas have been updated.
There might be a more elegant way to do this.


scholtzan:
cc @jklukas

bigquery_etl/cli/query.py
```diff
@@ -534,3 +536,183 @@ def initialize(name, sql_dir, project_id, dry_run):
...
+                        err=True,
+                    )
+
+                table = client.get_table(f"{project_name}.{dataset_name}.{table_name}")
+                table_schema = io.StringIO("")
+                client.schema_to_json(table.schema, table_schema)
+
+                if defined_schema != json.dumps(
+                    json.loads(table_schema.getvalue()), sort_keys=True
+                ):
```
scholtzan, line 710: There is probably a better way to compare schemas here

jklukas: A complete solution would probably be to traverse the schema files, descending into structs where necessary, building a list of field names in dotted notation (`client_info.client_id`, etc.). We would be able to compare alphabetically sorted lists to determine if the new schema is missing any existing fields, and we'd be able to report what new fields are added. This technique would also conveniently ignore the complication of having column comments not exist at the query level.

Hmm... this would also need to take into account data type to make sure there's not an incompatible change.

jklukas:
In order to support column descriptions, we'd need to to have the `schema generate` command support pulling the new schema from the query, merging in any column descriptions already defined in the checked-in schema file.

> One way I could imagine this working is by deploying updated schemas to the bigquery-etl-integration-test (or some other) project, let all queries reference this project and then dry run the queries to get their schema updates

I definitely want some testing workflow like this to exist, but I expect it would be a significant effort, and one which might require us to introduce some level of templating into our queries to make the conversion of one project to another more reliable. It also would run into issues for any query that has to reference tables outside shared-prod.

bigquery_etl/cli/query.py
whd: :amiyaguchi has definitely done some work around this for the safety-checks for MSG, using an approach I think similar to what :klukas describes.

One thing to note here is that the bigquery API sometimes returns legacy SQL for schemas and it can make diffing a lot more complicated: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/bigquery_table#schema. This is perhaps not an issue for query schemas where legacy sql not used.

wlach:
Is there any chance we could use an abbreviated yaml format for the schemas? I understand that we ultimately do need a generated JSON file to interface with BigQuery, but it would be nice if they were a little easier to view/edit/understand from a user's perspective.

jklukas:
> Is there any chance we could use an abbreviated yaml format for the schemas? I understand that we ultimately do need a generated JSON file to interface with BigQuery, but it would be nice if they were a little easier to view/edit/understand from a user's perspective.

I very much agree. I think we should use yaml for checked-in schemas. Converting to JSON where necessary should be trivial.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/metadata.yaml
```diff
@@ -1,21 +1,34 @@
...
-description: |-
-  A daily aggregation over `main` pings from each Firefox
+bigquery:
+  cluster_by:
+  - normalized_channel
+  - sample_id
+  partition_by:
+    field: submission_date
+    require_partition_filter: true
+    type: day
```
scholtzan, line 8: In order to get the schema of a deployed table for validation via a dry run the partitioning fields need to be known. So I think having this information as part of the metadata might be useful. 

This information is also required if a new table is created based on a schema. Since the schema doesn't include this information we'll need to specify it here.

scholtzan:
I made some updates to the initial proposal PR:
* Schema traversal when comparing schemas
* Add partitioning and clustering information to the `metadata.yaml` file which is used when validating queries (this way the dry run service can be used to get table schemas) and deploying destination tables
* Merging schemas when running `bqetl query schema update ...`. This way descriptions manually added to a schema file won't be overwritten
* Use YAML for schema files

Currently, the CLI has the following commands:
* `bqetl query schema update <dataset>.<table>`
    * Updates the `schema.yaml` file based on the `query.sql` schema and deployed schema of the destination table. If no `schema.yaml` file exists, this will create it based on the query.sql` schema and deployed schema of the destination table. This also updates the `metadata.yaml` and pulls in information about partitioning fields and clustering fields from the destination table.
* `bqetl query schema deploy <dataset>.<table>`
    * Updates the schema for the destination table based on the `schema.yaml` file or creates a new destination table if it doesn't exist yet. Partitioning and clustering for the new table can be configured via the `metadata.yaml` file.
* `bqetl query schema validate <dataset>.<table>`
    * Compares `schema.yaml` with the schema of the destination table and the query schema and shows whether they are all the same. This works entirely by using the dry run service, so we can add a step to the CI to run this validation.

bigquery_etl/cli/query.py
```diff
@@ -534,3 +537,213 @@ def initialize(name, sql_dir, project_id, dry_run):
...
+@click.option(
+    "--project-id",
+    "--project_id",
+    help="GCP project ID",
+    default="moz-fx-data-shared-prod",
+    callback=is_valid_project,
+)
+@click.pass_context
+def deploy(ctx, name, sql_dir, project_id):
+    """CLI command for deploying destination table schemas."""
```
scholtzan, line 633: todo: validate schemas before deploying

bigquery_etl/cli/query.py
```diff
@@ -534,3 +537,213 @@ def initialize(name, sql_dir, project_id, dry_run):
...
+                metadata.set_bigquery_partitioning(
+                    field=table.time_partitioning.field,
+                    partition_type=table.time_partitioning.type_.lower(),
+                    required=table.time_partitioning.require_partition_filter,
+                )
+                click.echo(f"Partitioning metadata added to {metadata_file_path}")
+
+            if table.clustering_fields and metadata.bigquery.cluster_by is None:
+                metadata.set_bigquery_clustering(table.clustering_fields)
+                click.echo(f"Clustering metadata added to {metadata_file_path}")
```
jklukas, line 608: I like this approach. We add the partitioning/clustering to metadata if not already set, but otherwise expect the metadata.yaml to be the source of truth for this info.

bigquery_etl/schema/__init__.py
```diff
@@ -0,0 +1,152 @@
...
+            self._traverse(
+                "root", self.schema["fields"], other.schema["fields"], update=False
+            )
+        except Exception as e:
+            print(e)
+            return False
+
+        return True
+
+    def _traverse(self, prefix, columns, other_columns, update=False):
```
jklukas, line 82: This traversal approach is great. I feel confident we'll be able to evolve this to meet our needs.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/metadata.yaml
jklukas: I'm very much in favor of codifying partioning/clustering info here. I'd like us to mimic the BQ API representation in the metadata as much as possible, which I expect will make it more natural to be able to evolve this to take advantage of new features in the future. So I would expect this to look like:

```
bigquery:
  require_partition_filter: true
  clustering:
    fields:
    - normalized_channel
    - sample_id
  time_partitioning:
    field: submission_date
    require_partition_filter: true
    type: day
 ```

**Comment by jklukas** [comment]
> Tests etc. still need to be updated, but first I wanted to see what people are thinking about moving forward with this.

I'm in support of moving forward with the direction. This PR is adds the fundamentals we need to work with schema files. We'll probably need to gain a bit of experience with it by doing a round or two of updates to clients_* ETL streams using this machinery and be able to evolve it into a much nicer and safer workflow than what we have today.

File: bigquery_etl/schema/__init__.py [hunk 6]
```diff
@@ -0,0 +1,191 @@
...
+
+    def compatible(self, other: "Schema") -> bool:
+        """
+        Check if schema is compatible with another schema.
+
+        If there is a field missing in the schema that is part of the "other" schema,
+        the schemas are still compatible. However, if there are fields missing in the
+        "other" schema they are not compatible since, e.g. inserting data into the "other"
+        schema that follows this schema would fail.
+        """
```
scholtzan, line 94: This might need a different name. 
Currently, it is used to check if the query results are compatible with the destination table schema. For example, if there are some extra fields in the destination table, the query can still write results to it without any issues. However, it there are new fields in the query that haven't been added to the destination table schema yet, this method will return `False`

.circleci/config.yml
```diff
@@ -158,6 +158,16 @@ jobs:
...
+  validate-schemas:
+    docker: *docker
+    steps:
+      - checkout
+      - *build
+      - run:
+          name: Validate query schemas
+          command: |
+            pip install .
+            bqetl query schema validate "*"
```
scholtzan, line 170: This will validate schemas only for queries that have a `schema.yaml` defined.

scholtzan:
This is ready for review. 

I added and updated tests and made sure the `bigquery` metadata follows the BQ API representation

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/metadata.yaml
```diff
@@ -1,21 +1,35 @@
...
+bigquery:
+  clustering:
+    fields:
+    - normalized_channel
+    - sample_id
+  time_partitioning:
+    field: submission_date
+    require_partition_filter: true
+    type: day
+description: 'A daily aggregation over `main` pings from each Firefox
```
jklukas, line 10: Can we restore the description to the `|-` version without the extra newlines?

bigquery_etl/cli/query.py
```diff
@@ -532,3 +536,256 @@ def initialize(name, sql_dir, project_id, dry_run):
...
+    except FileNotFoundError:
+        pass
+
+    table_schema = Schema.for_table(
+        project_name, dataset_name, table_name, partitioned_by
+    )
+
+    if not query_schema.compatible(table_schema):
+        click.echo(
+            f"Schema for query in {query_file_path} "
```
jklukas, line 753: Might be good to prepend this with `ERROR: ` to make it stand out.

bigquery_etl/metadata/parse_metadata.py
```diff
@@ -146,7 +207,26 @@ def write(self, file):
...
+        class Literal(str):
+            pass
+
+        def literal_presenter(dumper, data):
+            return dumper.represent_scalar("tag:yaml.org,2002:str", data, style="|")
+
+        yaml.add_representer(Literal, literal_presenter)
+
+        if "description" in metadata_dict:
+            metadata_dict["description"] = Literal(metadata_dict["description"])
```
scholtzan, line 220: This ensures that descriptions are presented as literals (`|-`)

bigquery_etl/metadata/parse_metadata.py
```diff
@@ -146,7 +207,26 @@ def write(self, file):
...
+
+        converter = cattr.Converter()
+        file.write_text(
+            yaml.dump(
+                converter.unstructure(metadata_dict),
+                default_flow_style=False,
+                sort_keys=False,
```
scholtzan, line 227: This ensures that keys appear in the `metadata.yaml` file in the order they are defined in the `Metadata` class. Which means the friendly name and description are at the top of the file

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/metadata.yaml
scholtzan: This was easier to configure than I thought.

bigquery_etl/metadata/parse_metadata.py
```diff
@@ -146,7 +207,26 @@ def write(self, file):
...
 
-        file.write_text(yaml.dump(metadata_dict))
+        # Use literal yaml representation for descriptions
+        class Literal(str):
+            pass
+
+        def literal_presenter(dumper, data):
+            return dumper.represent_scalar("tag:yaml.org,2002:str", data, style="|")
+
+        yaml.add_representer(Literal, literal_presenter)
```
jklukas, line 217: Does this affect `yaml` library behavior globally? Should this be happening once in global scope rather than nested here?

bigquery_etl/metadata/parse_metadata.py
jklukas: This is great!

jklukas (approved):
I am very happy with this! I did some testing with clients_last_seen, and it was able to successfully detect incompatibilities when I deleted a field and when I changed a field type. The metadata.yaml produced by `update` also now looks great.

PR #1753, opened 2021-02-04
Expose scrubbed data from apple receipts to redash

relud:


sql/moz-fx-data-shared-prod/mozilla_vpn_derived/subscriptions_v1/query.sql
```diff
@@ -1,3 +1,147 @@
...
+  status INT64,
+> DETERMINISTIC
+LANGUAGE js
+AS
+  "return JSON.parse(receipt);";
+
+CREATE TEMP FUNCTION udf_scrub_apple_receipt(apple_receipt ANY TYPE) AS (
+  STRUCT(
+    apple_receipt.environment,
+    -- extract a list of start and end dates that doesn't overlap or repeat
```
relud, line 85: @gkabbz this should be the minimum info we need to integrate apple subscriptions with what we have from Stripe.

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/subscriptions_v1/query.sql
```diff
@@ -1,3 +1,147 @@
...
+    request_date STRING,
+    request_date_ms STRING,
+    request_date_pst STRING,
+    version_external_identifier INT64,
+  >,
+  status INT64,
+> DETERMINISTIC
+LANGUAGE js
+AS
+  "return JSON.parse(receipt);";
```
acmiyaguchi, line 80: I'd like to see this is a udf with a validation document test, like we do mozilla-pipeline-schemas (as long as bigquery supports multtline strings).

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/subscriptions_v1/query.sql
```diff
@@ -1,3 +1,147 @@
...
+          IF(
+            ANY_VALUE(date) OVER (
+              ORDER BY
+                UNIX_DATE(date) RANGE
+                BETWEEN 1 PRECEDING
+                AND 1 PRECEDING
+            ) IS NULL,
+            date,
+            NULL
+          ) AS start_date,
```
acmiyaguchi, line 99: This queries the start date by finding a value that doesn't have anything before it? Is this equivalent to `if(lag(date) over (order by date) is null, date, null)`?

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/subscriptions_v1/query.sql
```diff
@@ -1,3 +1,147 @@
...
+    request_date_pst STRING,
+    version_external_identifier INT64,
+  >,
+  status INT64,
+> DETERMINISTIC
+LANGUAGE js
+AS
+  "return JSON.parse(receipt);";
+
+CREATE TEMP FUNCTION udf_scrub_apple_receipt(apple_receipt ANY TYPE) AS (
```
acmiyaguchi, line 82: Can you add the return type here? Is it `array(struct(start_date timestamp, end_date timestamp))`?

acmiyaguchi: Could you also add a short comment at the head of the udf on what scrubbing the receipt means?

relud: > Can you add the return type here? Is it `array(struct(start_date timestamp, end_date timestamp))`?

`STRUCT<environment STRING, array<struct<start_date DATE, end_date DATE>> active_periods>`

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/subscriptions_v1/query.sql
relud: > This queries the start date by finding a value that doesn't have anything before it?

yeah. also note that this is using `RANGE`, not `ROWS`, so it's specifically checking if one day before `date` is present, not one row before.

> Is this equivalent to `if(lag(date) over (order by date) is null, date, null)`?

i'm not familiar with a `lag` aggregate function, but I could define one where that would be correct

acmiyaguchi: No need to redefine the function, I'm just trying to get a handle on how it worked. If LAG and LEAD do work in this context though, I think it would be a littler clearer in terms of intent.

relud: looking at the definitions, `LAG` and `LEAD` only do preceding row, not range

sql/mozfun/iap/scrub_apple_receipt/udf.sql
```diff
@@ -0,0 +1,117 @@
...
+              CAST(UNIX_MILLIS(TIMESTAMP "2020-01-01T20:00:00") AS STRING),
+              "false"
+            ),
+            (
+              CAST(UNIX_MILLIS(TIMESTAMP "2020-01-02") AS STRING),
+              CAST(UNIX_MILLIS(TIMESTAMP "2020-01-03") AS STRING),
+              "false"
+            ),
+            (
+              CAST(UNIX_MILLIS(TIMESTAMP "2020-01-05") AS STRING),
```
acmiyaguchi, line 104: Is this going to be the case in the production environment where the receipts are going to have a 1 day gap?

acmiyaguchi (approved):
Excellent, heavy +1 on the udfs. I have one question about the shape of the data in production, but otherwise looks good to me.

sql/mozfun/iap/scrub_apple_receipt/udf.sql
relud: if there's a one day gap where they didn't have a valid subscription, then yes

acmiyaguchi: Okay, I see now :+1:

PR #1755, opened 2021-02-04
Add descriptions for major desktop views

jklukas:
In support of https://github.com/mozilla/bigquery-etl/pull/1735

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry/clients_daily/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/telemetry/clients_daily/metadata.yaml
@@ -1,0 +1,13 @@
+---
+friendly_name: Clients Daily
+description: >-
+  A daily aggregate of main and event pings from each Firefox
+  desktop client.
+
+  Clustering fields: `normalized_channel`, `sample_id`
+
+  See also: `clients_last_seen`, `clients_first_seen`
+labels:
+  application: firefox
+owners:
+  - dthorn@mozilla.com
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/metadata.yaml
@@ -4,9 +4,10 @@ description: >-
   A daily aggregation over `main` pings from each Firefox
   desktop client, partitioned by day.
 
-  Currently, `telemetry.clients_daily` is a view directly on top of this table,
-  though by end of Q1 2021, it will transition to point at the downstream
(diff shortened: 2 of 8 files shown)


sql/moz-fx-data-shared-prod/telemetry/clients_daily/metadata.yaml
```diff
@@ -0,0 +1,13 @@
+---
+friendly_name: Clients Daily
+description: >-
+  A daily aggregate of main and event pings from each Firefox
+  desktop client.
+
+  Clustering fields: `normalized_channel`, `sample_id`
+
+  See also: `clients_last_seen`, `clients_first_seen`
```
wlach, line 9: This seems out of sync with the description in the derived dataset: 

https://github.com/mozilla/bigquery-etl/blob/master/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/metadata.yaml

It would be nice if we only had a single source of truth for these descriptions-- I wonder if that means that we should make it more clear that the queries in telemetry_derived are "behind the scenes" in their metadata and keep any notes in there implementation-specific.

wlach (changes requested):
Less a "request for changes" than a "what do you think about this". 😄

sql/moz-fx-data-shared-prod/telemetry/clients_daily/metadata.yaml
jklukas: Very good point. Will update the table descriptions to that effect.

jklukas: I've simplified the descriptions for the underlying tables, and they now reference the user-facing view for more details.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/metadata.yaml
```diff
@@ -1,7 +1,10 @@
 ---
 friendly_name: Clients Daily
-description: A daily aggregate of main pings from each Firefox
-  desktop client, partitioned by day
+description: >-
+  A daily aggregation over `main` pings from each Firefox
+  desktop client, partitioned by day.
+
+  See view: `telemetry.clients_daily`
```
wlach, line 7: Maybe we should be even more explicit about this, something like "This is the backing implementation of `clients_daily`. Normally you would access this via the view `telemetry.clients_daily`" ?

jklukas: Updated to address the spirit of your language here. I also added a note and link about upcoming change to bring in event ping data.

wlach (approved):
👍

PR #1763, opened 2021-02-05
Get referenced tables for views

Iinh:
Follow up to #1735.


bigquery_etl/docs/derived_datasets/generate_derived_dataset_docs.py
```diff
@@ -56,6 +57,11 @@ def generate_derived_dataset_docs(out_dir, project_dir):
                 if VIEW_FILE in files:
                     source_urls["View"] = f"{SOURCE_URL}/{root}/{VIEW_FILE}"
 
+                    view_file = os.path.join(root, VIEW_FILE)
+                    referenced = DryRun(view_file).get_referenced_tables()
+                    if referenced:
+                        print('Referenced tables: ', referenced)
+
```
Iinh, line 64: @acmiyaguchi So I’m working on getting referenced tables for the user facing views, but there must be something wrong with my (very naive) approach because this prints 0 referenced table for all view files. What am I missing?

Iinh: @wlach Could you help also? Thanks!

acmiyaguchi: Are there any print statements logged during the process? You may want to set logging or a trace in the dryrun code to debug:

https://github.com/mozilla/bigquery-etl/blob/1dbb06ce9233c59b9c52a93c36d95f882a56494f/bigquery_etl/dryrun.py#L216-L224

I'd recommend either:

```python
print(self.dry_run_result)

# alternatively, to get into the debugger
import pdb; pdb.set_trace()
```

Iinh: Yes there are print statements logged, but it only prints empty lists for all views:

Screen Shot 2021-02-05 at 12 17 59 PM

Thanks for the debugging tip, I'll keep investigating.

acmiyaguchi: It'd be more informative to look at the dry-run response directly, it may give insight into why the field is missing. I can imagine permission issues being one possible error.

Iinh: Yep it's permission issue. That's so strange I already set up GCP and have access to BigQUery

acmiyaguchi: I'm not sure, the dry-run endpoint may be publicly accessible because I don't see an [oidc token](https://cloud.google.com/scheduler/docs/http-target-auth) in the request header. Do you mind sharing a full response with the auth issue?

@scholtzan do you know how the cloud function is configured, and whether the behavior here is expected?

scholtzan: The cloud function should be publicly accessible, so no authentication should be required

Iinh: Full error log:

```
{

    'valid':False,
    'errors':[
        {
            'code':403,
            'errors':[
                {
                    'message':'Access Denied: Dataset moz-fx-data-shared-prod:messaging_system: User does not have bigquery.tables.create permission for dataset moz-fx-data-shared-prod:messaging_system.',
                    'domain':'global',
                    'reason':'accessDenied'
                }
            ],
            'response':{
                'headers':{
                    'alt-svc':'h3-29=":443"; ma=2592000,
                    h3-T051=":443"; ma=2592000,
                    h3-Q050=":443"; ma=2592000,
                    h3-Q046=":443"; ma=2592000,
                    h3-Q043=":443"; ma=2592000,
                    quic=":443"; ma=2592000; v="46,
                    43"', 'cache-control':'private',
                    'content-encoding':'gzip',
                    'content-type':'application/json; charset=UTF-8',
                    'date':'Fri,
                    05 Feb 2021 21:10:05 GMT',
                    'server':'ESF',
                    'transfer-encoding':'chunked',
                    'vary':'Origin,
                    X-Origin, Referer', 'x-content-type-options':'nosniff',
                    'x-frame-options':'SAMEORIGIN',
                    'x-xss-protection':'0'
                }
            },
            'message':'Access Denied: Dataset moz-fx-data-shared-prod:messaging_system: User does not have bigquery.tables.create permission for dataset moz-fx-data-shared-prod:messaging_system.'
        }
    ]

}
SyntaxError: JSON.parse: expected property name or '}' at line 1 column 2 of the JSON data
```

scholtzan: Oh, right. You need to check if the view is part of the `SKIP` list and skip them since the dry run will fail for them.

acmiyaguchi: That would explain a part of the problem, but it's unclear to me why the ones with valid permissions would be returning empty references.

Iinh: Maybe I'm not understanding this correctly but isn't the SKIP check already done here?

https://github.com/mozilla/bigquery-etl/blob/1dbb06ce9233c59b9c52a93c36d95f882a56494f/bigquery_etl/dryrun.py#L211-L215

scholtzan: Looks like there is a `return []` missing. Currently, it is just printing out `...Ignoring dryrun results for` and then moves  on to dry running the file anyway

Iinh: The referenced code is from the `get_referenced_tables()` in `dryrun.py`, which already includes `return []`. 
https://github.com/mozilla/bigquery-etl/blob/1dbb06ce9233c59b9c52a93c36d95f882a56494f/bigquery_etl/dryrun.py#L209-L224
Since I imported the function I assumed that the SKIP check is already included, and I think the views that needed to be skipped were skipped, for example:

Screen Shot 2021-02-05 at 12 17 59 PM

The error that I got is: `Access Denied: Dataset moz-fx-data-shared-prod:messaging_system: User does not have bigquery.tables.create permission for dataset moz-fx-data-shared-prod:messaging_system.` 

If it's not too long an answer, could you help explain what `bigquery.tables.create permission` is?

scholtzan: For creating new views, e.g. via `CREATE OR REPLACE VIEW` (which is what all our view definitions look like), the service account running the SQL needs `bigquery.tables.create` permissions. It looks like view definitions are a special case that is handled in the dry run script: https://github.com/mozilla/bigquery-etl/blob/350291bd1a9608d6659f6cbd4c9123f4a366434f/bigquery_etl/dryrun.py#L244
Dry running view definitions always returns an error due to the missing permissions, but if this error is returned dry run will pretend the dry run was successful.

scholtzan: In order for you to be able to get referenced tables, you could retrieve the view query and then dry run that one. There is some documentation here on how to access view metadata including the query: https://cloud.google.com/bigquery/docs/view-metadata

Iinh: That makes total sense to me now! Thank you @scholtzan!

Iinh:
@scholtzan @acmiyaguchi I've updated DryRun to return referenced tables for views that require a date filter. Could you review if this approach is reasonable, I'll add tests if this is ok. Thanks!

Preview: https://iinh.github.io/bigquery-etl/mozdata/account_ecosystem/

CleanShot 2021-02-10 at 00 11 03@2x

scholtzan:
Yeah, I think the overall approach looks reasonable

bigquery_etl/dryrun.py
```diff
@@ -245,18 +302,61 @@ def is_valid(self):
...
+            # (submission_date, submission_timestamp, etc.)
+            # We mark these requests as valid and add a date filter
+            # in get_referenced_table()
+            print(f"{self.sqlfile:59} OK")
+        elif (
+            error
+            and error.get("code", None) in [400, 403]
+            and "Syntax error: Expected end of input but got keyword WHERE"
+            in error.get("message", "")
+        ):
```
acmiyaguchi, line 320: I don't think we should be overriding the default dry-run behavior to return OK if there are legitimate errors. Maybe this should be conditional based on whether a flag is set during the constructor?

bigquery_etl/dryrun.py
```diff
@@ -245,18 +302,61 @@ def is_valid(self):
...
-            print(f"{self.sqlfile:59} ERROR\n", self.dry_run_result["errors"])
+            print(f"{self.sqlfile:59} INVALID ERROR\n", self.dry_run_result["errors"])
             return False
 
         return True
 
+    def get_error(self):
+        error = {}
+        if "errors" in self.dry_run_result and len(self.dry_run_result["errors"]) == 1:
+            error = self.dry_run_result["errors"][0]
```
acmiyaguchi, line 334: `error` is never returned?

bigquery_etl/dryrun.py
```diff
@@ -245,18 +302,61 @@ def is_valid(self):
             # we expect CREATE VIEW and CREATE TABLE to throw specific
             # exceptions.
             print(f"{self.sqlfile:59} OK")
+        elif (
+            error
+            and error.get("code", None) in [400, 403]
+            and "without a filter over column(s)" in error.get("message", "")
```
acmiyaguchi, line 308: Maybe refactor/reuse the results of `get_error` here?

acmiyaguchi (commented):
Overall, the approach looks fine. I'm worried about overriding errors in the dry-run for the sake of generating documentation. Try to keep the changes to the `DryRun` api backwards compatible, where we can still obtain the unmodified dry-run calls.

Tests would definitely be useful here, making sure that you cover each one of the branches in the `get_referenced_tables` method.

bigquery_etl/docs/derived_datasets/generate_derived_dataset_docs.py
```diff
@@ -58,6 +60,23 @@ def generate_derived_dataset_docs(out_dir, project_dir):
                 if VIEW_FILE in files:
                     source_urls["View Definition"] = f"{SOURCE_URL}/{root}/{VIEW_FILE}"
 
+                    view_file = os.path.join(root, VIEW_FILE)
+                    content = ""
+
+                    with open(view_file) as f:
+                        view_content = f.read()
+                        # get the view content by removing CREATE OR REPLACE VIEW
```
jklukas, line 68: Why do we need to strip the CREATE OR REPLACE VIEW part? Do we not get table references when dry running a DML statement?

jklukas:
I'm hoping that we can avoid the SQL parsing logic that's going on here, which may be difficult to maintain.

If there's a way that we can improve the dry run service to be able to return the info we need from CREATE VIEW statements and avoid the need to manipulate the content, I would prefer to do that.

bigquery_etl/docs/derived_datasets/generate_derived_dataset_docs.py
Iinh: I believe because the current dry run service only has READ permission so if a query has "CREATE OR REPLACE VIEW" it will return a permission error. We have been bypassing this error here:
https://github.com/mozilla/bigquery-etl/blob/5148a7f775a968e1d46205b67cca160daa762ae7/bigquery_etl/dryrun.py#L239-L251

> If there's a way that we can improve the dry run service to be able to return the info we need from CREATE VIEW statements and avoid the need to manipulate the content, I would prefer to do that.

It definitely sounds like a better approach. I'm not familiar enough with the cloud run service to know if this is possible, any advice @scholtzan @acmiyaguchi?

jklukas: Thank you for the specific link. I had forgotten about the read-only restriction, which indeed complicates matters. Maybe we can still figure out a workaround.

acmiyaguchi: The error message makes sense. I actually don't understand why we need the DML statement in the first place (I don't need them in terraform or the console to create views in tables). 

I mentioned an alternative in a previous comment, which is to use `select * from {view} where {partition filter}`. This assumes the view has already been deployed, but does not require modifying the SQL itself.

jklukas: > I actually don't understand why we need the DML statement in the first place (I don't need them in terraform or the console to create views in tables).

I agree that it would be better not to have DML in our `.sql` files, but it would be some significant effort to change the design at this point.

bigquery_etl/dryrun.py
```diff
@@ -221,42 +244,124 @@ def get_referenced_tables(self):
...
             # We want the dryrun service to only have read permissions, so
             # we expect CREATE VIEW and CREATE TABLE to throw specific
             # exceptions.
             print(f"{self.sqlfile:59} OK")
+        elif self.get_error() == Errors.DATE_FILTER_NEEDED and self.strip_dml:
+            # Some queries require a partition filter
+            # (submission_date, submission_timestamp, etc.)
+            # We mark these requests as valid and add a date filter
+            # in get_referenced_table()
+            print(f"{self.sqlfile:59} OK but DATE FILTER NEEDED")
```
acmiyaguchi, line 319: This is changing the behavior of the dry run because jobs that would have previously failed now pass. I don't think this should cause any issues though.

tests/test_dryrun.py
```diff
@@ -34,6 +36,18 @@ def test_sql_file_invalid(self, tmp_path):
...
 
+    def test_view_file_valid(self):
+        view_file = (
+            TEST_DIR
+            / "sql"
+            / "moz-fx-data-shared-prod"
+            / "messaging_system"
+            / "cfr_users_last_seen"
+            / "view.sql"
+        )
```
acmiyaguchi, line 47: We generally haven't used the `tests/sql` directory to store test assets like this. Could put these in a `test_fixture` dataset instead e.g. `moz-fx-data-shared-prod/test_fixtures/cfr_users_last_seen`, or point directly to the view in the top-level `sql` directory.

Even better would be to simplify the queries even further and write them directly into a temporary file here e.g.

```python
def test_view_file_valid(self, tmp_path):
    sql = """CREATE OR REPLACE
    -- add the minimal amount of query necessary to reproduce the error
    ...
    """
    path = (tmp_path / "view.sql").write_text(sql)
    assert DryRun(str(path)).is_valid()
```

And so forth.

tests/test_dryrun.py
```diff
@@ -54,3 +68,55 @@ def test_get_referenced_tables(self, tmp_path):
...
+    def test_get_referenced_tables_dml_stripped(self):
+
+        view_file = (
+            TEST_DIR / "sql" / "moz-fx-data-shared-prod" / "telemetry" / "view.sql"
+        )
+
+        dml_stripped = DryRun(sqlfile=str(view_file), strip_dml=True)
+        dml_unstripped = DryRun(str(view_file))
+
+        assert (
```
acmiyaguchi, line 81: Also directly test `get_sql` method here. The idea is that if `get_referenced_tables()` were broken, this test wouldn't be affected.

tests/test_dryrun.py
```diff
@@ -54,3 +68,55 @@ def test_get_referenced_tables(self, tmp_path):
...
+            TEST_DIR
+            / "sql"
+            / "moz-fx-data-shared-prod"
+            / "messaging_system"
+            / "onboarding_events_amplitude"
+            / "view.sql"
+        )
+
+        view_1 = DryRun(sqlfile=str(view_file_1), strip_dml=True)
+        view_2 = DryRun(sqlfile=str(view_file_2), strip_dml=True)
```
acmiyaguchi, line 112: I would test `get_error` directly here and ensure that you cover every enum value.

bigquery_etl/docs/derived_datasets/generate_derived_dataset_docs.py
```diff
@@ -57,6 +58,11 @@ def generate_derived_dataset_docs(out_dir, project_dir):
                             print(error)
                 if VIEW_FILE in files:
                     source_urls["View Definition"] = f"{SOURCE_URL}/{root}/{VIEW_FILE}"
+                    view_file = os.path.join(root, VIEW_FILE)
+
```
acmiyaguchi, line 62: Just to consider the option where parsing SQL is not required, you could do something like this:

```python
import tempfile
...
if VIEW_FILE in files:
    source_urls["View Definition"] = f"{SOURCE_URL}/{root}/{VIEW_FILE}"

    # find the dataset/table information
    path = Path(root)
    dataset = path.parent.parent.name
    view = path.parent.name

    # create a temp file to pass into dryrun
    temp_sqlfile = Path(tempfile.mkdtemp()) / "view.sql"
    temp_sqlfile.write_text(f"select * from {dataset}.{view}")
    referenced_tables = DryRun(sqlfile=tmp_sqlfile)
```

Iinh: I think we need to pass in the original view file so that it can be tested against the SKIP list. But I'll add this logic in the DryRun class with `strip_dml` flag to get the referenced tables

bigquery_etl/dryrun.py
Iinh: I believe it would only change the behavior of jobs that have `strip_dml` flag on, so it shouldn't affect any existing jobs

bigquery_etl/docs/derived_datasets/generate_derived_dataset_docs.py
Iinh: After implementing this method, I found that it returns a lot of unfound table errors, like: `'message': 'Not found: Dataset moz-fx-data-shared-prod:telemetry.clients_scalar_aggregates_v1 was not found in location US', 'domain': 'global', 'reason': 'notFound'`
After comparing the names of sql directories in bigquery-etl to BigQuery datasets, I found that there's not a direct 1:1 correlation between the dir names, so extracting view names from directories like this would not work

tests/test_dryrun.py
```diff
@@ -41,16 +60,96 @@ def test_get_referenced_tables_empty(self, tmp_path):
...
+
+        invalid_dml_stripped = """
+        SELECT
+        *
+        FROM
+          `moz-fx-data-shared-prod.telemetry_derived.clients_daily_v6`
+        WHERE something
+        WHERE submission_date > current_date()
+        """
+        print(
```
acmiyaguchi, line 132: Print statements are usually not needed in pytest if you assert the value or use the `--pdb` option.

tests/test_dryrun.py
```diff
@@ -41,16 +60,96 @@ def test_get_referenced_tables_empty(self, tmp_path):
...
+        view_file = tmp_path / "telmetry_derived" / "view.sql"
+        view_file.write_text(
+            """
+            CREATE OR REPLACE VIEW
+            `moz-fx-data-shared-prod.telemetry.clients_daily`
+            AS
+            SELECT
+            *
+            FROM
+            `moz-fx-data-shared-prod.telemetry_derived.clients_daily_v6`
```
acmiyaguchi, line 87: Can you add another test with multiple referenced tables?

```
select document_id from mozdata.org_mozilla_firefox.baseline where submission_timestamp > current_timestamp()
union all
select document_id from mozdata.org_mozilla_fenix.baseline where submission_timestamp > current_timestamp()
```

tests/test_dryrun.py
```diff
@@ -41,16 +60,96 @@ def test_get_referenced_tables_empty(self, tmp_path):
...
+        """
+        )
+        query_dryrun = DryRun(str(query_file)).get_referenced_tables()
+        view_dryrun = DryRun(str(view_file), strip_dml=True).get_referenced_tables()
 
-        assert len(response) == 1
-        assert response[0]["datasetId"] == "telemetry_derived"
-        assert response[0]["tableId"] == "clients_daily_v6"
+        assert len(query_dryrun) == 1
```
acmiyaguchi, line 93: Can you move the test setup directly before the asserts? This is mostly a matter of style, but I think it's useful to group tests as if you were to put each case into a function.

acmiyaguchi (commented):
Okay, thanks for exploring the indirect method of checking tables. I have a few more comments, but this looks good in general.

tests/test_dryrun.py
Iinh: Thank you! This is now fixed.

acmiyaguchi (approved):
Looks good to me, looking forward to seeing the generated docs.

Issue #1770, opened 2021-02-08
UDF for sortable version parsing


jklukas:
From discussion on Slack:

> For example, we'll release Firefox v100 early next year and truncate_version(app_version, 'minor') > 80 would evaluate false for pings with an app_version of 100.x. I think with semantic versioning (afaik we use for all products), casting major and major.minor to a numeric should work as expected.

jklukas:
cc @wlach @ecsmyth

wlach:
Since the `truncate_version` UDF *just* landed, I am tempted to break its API and make it return a float, since I think that should work in all the cases we care about. Any thoughts?

jklukas:
I think a reasonable way forward here might be to make a function that returns a struct. So `mozfun.norm.semantic_version("102.12.13-beta")` function that returns a struct:

```
major: 102 (int)
minor: 12 (int)
patch: 13 (int)
suffix: "-beta" (string)
major_minor: 102.12 (float)
```

jklukas:
I'm a little leery of using a float to represent major+minor version. Semantically, it should be an array of integers. A float certainly allows for more ergonomic sorting, but a float may be unstable if used as a grouping key.

jklukas:
> Since the truncate_version UDF just landed, I am tempted to break its API and make it return a float, since I think that should work in all the cases we care about. Any thoughts?

To be clear, I don't want to block this change. @ecsmyth points out some clear usability problems of the string version, and there's a path here to make it nicely sortable by changing the return type to float.

I would just want to see some tests that show we don't end up with "12.3" becoming 12.30000000000000004, etc.

ecsmyth:
@jklukas I like your struct-based approach. The name ties it to a specific versioning scheme and it is flexible enough for a few different use cases. I think that being unable to use the returned value to group or in a `SELECT DISTINCT` is ok, since I'd image you'd be operating on a field of the struct rather than the struct itself.

wlach:
> > Since the truncate_version UDF just landed, I am tempted to break its API and make it return a float, since I think that should work in all the cases we care about. Any thoughts?
> 
> To be clear, I don't want to block this change. @ecsmyth points out some clear usability problems of the string version, and there's a path here to make it nicely sortable by changing the return type to float.
> 
> I would just want to see some tests that show we don't end up with "12.3" becoming 12.30000000000000004, etc.

I took this approach in #1772. I believe using NUMERIC should protect us against that pathological case. 

[...]

PR #1772, opened 2021-02-08
Make `truncate_version` return NUMERIC

wlach:
This should work reliably for all the cases we care about and
allows doing things like sorting by versions reliably (since one
can't assume lexicographical ordering).

Using NUMERIC should protect us from weird floating point edge cases:

https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#decimal_types

See also #1770

### DIFF
--- a/sql/mozfun/norm/truncate_version/udf.sql
+++ b/sql/mozfun/norm/truncate_version/udf.sql
@@ -1,14 +1,14 @@
 CREATE OR REPLACE FUNCTION norm.truncate_version(os_version STRING, truncation_level STRING)
-RETURNS STRING AS (
+RETURNS NUMERIC AS (
   CASE
   WHEN
     truncation_level = "minor"
   THEN
-    REGEXP_EXTRACT(os_version, r"^([0-9]+[.]?[0-9]+).*")
+    CAST(REGEXP_EXTRACT(os_version, r"^([0-9]+[.]?[0-9]+).*") AS NUMERIC)
   WHEN
     truncation_level = "major"
   THEN
-    REGEXP_EXTRACT(os_version, r"^([0-9]+).*")
+    CAST(REGEXP_EXTRACT(os_version, r"^([0-9]+).*") AS NUMERIC)
   ELSE
     NULL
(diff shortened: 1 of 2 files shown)


sql/mozfun/norm/truncate_version/udf.sql
```diff
@@ -1,24 +1,25 @@
...
 -- Tests
 SELECT
-  assert.equals("16.1", norm.truncate_version("16.1.1", "minor")),
-  assert.equals("16", norm.truncate_version("16.1.1", "major")),
-  assert.equals("10", norm.truncate_version("10", "minor")),
-  assert.equals("5.1", norm.truncate_version("5.1.5-ubuntu-foobar", "minor")),
-  assert.equals(CAST(NULL AS STRING), norm.truncate_version("5.1.5-ubuntu-foobar", "patch")),
-  assert.equals(CAST(NULL AS STRING), norm.truncate_version("foo-bar", "minor"))
+  assert.equals(16.1, norm.truncate_version("16.1.1", "minor")),
+  assert.equals(16.03, norm.truncate_version("16.03.1", "minor")),
```
jklukas, line 20: This case actually worries me a bit. Do we see minor versions like "03" in the wild? I could imagine a minor version "2" without the zero padding that should sort before "03" logically, but the numeric value we return is `16.03`.

I think I am worrying too much, and this is probably fine.

jklukas (approved):
r+wc

I think the docs for this function should be updated to explicitly discuss the return type and the benefits of NUMERIC for sorting.

sql/mozfun/norm/truncate_version/udf.sql
wlach: Ubuntu numbers things like "16.04" (YY.MM), so it's not unheard of. But I don't think they mix/match versioning like you suggest for your pathological case: the next version after "16.04" was "16.10" IIRC.

wlach:
@jklukas quick sanity check on the doc updates?

jklukas (approved):

PR #1781, opened 2021-02-09
New cookbook on creating a derived dataset

wlach:
This PR attempts to add a cookbook to the generated documentation. It also refactors the front page a bit to incorporate the documentation in a sensible way, here's what it looks like:

image

Thanks to @scholtzan for reviewing my first draft of the cookbook, hopefully there isn't *too* much more work to be done.

Curious @Iinh if you have any thoughts on this, especially with respect to how this fits into other documentation work you've been doing recently.

I've elected to leave the README as-is for now, even though much of its content is redundant with what's in this tutorial. I figure we can deal with that in a followup.


docs/cookbooks/creating_a_derived_dataset.md
```diff
@@ -0,0 +1,208 @@
...
+
+Then create your pull request, either from the GitHub web interface or the command line, per your preference.
+
+Note this example assumes that `origin` points to your fork. Adjust the last push invocation appropriately if you have a different [remote](https://git-scm.com/docs/git-remote) set.
+
+Speaking of forks, note that if you're making this pull request from a fork, many jobs will currently fail due to lack of credentials. In fact, even if you're pushing to the origin, you'll get failures because the table is not yet created. That brings us to the next step, but before going further it's generally best to get someone to review your work: at this point we have more than enough for people to provide good feedback on.
+
+## Creating an initial table
+
+To bootstrap an initial table, the normal best practice is to create another SQL attempt, similar to the incremental query above, which creates the dataset. We'll write out another file called `init.sql` in `sql/moz-fx-data-shared-prod/org_mozilla_mozregression_derived/mozregression_aggregates_v1`:
```
jklukas, line 159:
```suggestion
To bootstrap an initial table, the normal best practice is to create another SQL file, similar to the incremental query above, which creates the table. We'll write out another file called `init.sql` in `sql/moz-fx-data-shared-prod/org_mozilla_mozregression_derived/mozregression_aggregates_v1`:
```

docs/cookbooks/creating_a_derived_dataset.md
```diff
@@ -0,0 +1,208 @@
...
+GROUP BY
+  date,
+  mozregression_version,
+  mozregression_variant,
+  app_used,
+  os,
+  os_version;
+```
+
+Note the `PARTITION BY DATE(date)` in the statement. This makes it so BigQuery will partition the table by date. This isn't too big a deal for mozregression (where even the size of unaggregated data is very small) but [can be a godsend for datasets where each day is hundreds of gigabytes or terrabytes big](https://docs.telemetry.mozilla.org/cookbooks/bigquery/optimization.html).
```
jklukas, line 189:
```suggestion
Note the `PARTITION BY DATE(date)` in the statement. This makes it so BigQuery will partition the table by date. This isn't too big a deal for mozregression (where even the size of unaggregated data is very small) but [can be a godsend for datasets where each day is hundreds of gigabytes or terabytes big](https://docs.telemetry.mozilla.org/cookbooks/bigquery/optimization.html).
```

docs/mkdocs.yml
```diff
@@ -24,7 +24,9 @@ plugins:
     - awesome-pages
 nav:
     - index.md
-    - mozdata:
+    - Cookbooks:
+        - Creating a derived dataset: cookbooks/creating_a_derived_dataset.md
+    - Datasets:
```
jklukas, line 29: We may still want to add other projects besides mozdata here. If we were to add marketing data, is there a way to nest that under here? I suspect it may be a better user experience to have each of these stay at the top-level, in which case I think it may be good to continue to have these headings be `mozfun` and `mozdata`, and we can rely on the index page to give the context of what those terms mean.

jklukas (changes requested):
Left a few suggestions for changes. Overall, adding this structure seems reasonable.

docs/cookbooks/creating_a_derived_dataset.md
```diff
@@ -0,0 +1,208 @@
...
+
+In this case, we want to create a simple table of aggregated metrics related to mozregression's use that we can use to power dashboards (to prioritize feature development internally inside Mozilla), as well as syndication as a [public dataset](https://blog.mozilla.org/data/2020/09/25/data-publishing-mozilla/).
+
+## Initial steps
+
+Set up bigquery-etl on your system per the instructions in the [README.md](https://github.com/mozilla/bigquery-etl/blob/master/README.md).
+
+## Create the Query
+
+First, you'll need to create a query files to work off of. For this step, you'll need to know what you want your derived dataset to be called. In this case, we'll name it `org_mozilla_mozregression_derived.mozregression_aggregates`.
```
scholtzan, line 19:
```suggestion
First, you'll need to create a query file to work off of. For this step, you'll need to know what you want your derived dataset to be called. In this case, we'll name it `org_mozilla_mozregression_derived.mozregression_aggregates`.
```

docs/cookbooks/creating_a_derived_dataset.md
```diff
@@ -0,0 +1,208 @@
...
+./bqetl query validate org_mozilla_mozregression_derived.mozregression_aggregates
+```
+
+If there are no problems, you should see no output.
+
+## Creating a DAG
+
+BigQuery-ETL has some facilities in it to automatically add your query to [telemetry-airflow](https://github.com/mozilla/telemetry-airflow) (our instance of Airflow).
+
+Before scheduling your query, you'll need to find an [Airflow DAG](https://airflow.apache.org/docs/apache-airflow/stable/concepts.html#dags) to run it off of. In some cases, one may already exist but in this one, we don't have one -- so we'll create it:
```
scholtzan, line 106: Is it worth mentioning here how to find existing DAGs? Either by looking into the `dags.yaml` or by running `bqetl dag info`

docs/cookbooks/creating_a_derived_dataset.md
```diff
@@ -0,0 +1,208 @@
...
+
+Note that we are scheduling the generation of the underlying _table_ which is `org_mozilla_mozregression_derived.mozregression_aggregates_v1` rather than the view.
+
+After doing this, you will also want to generate the actual airflow configuration which telemetry-airflow will pick up. Run:
+
+```bash
+./bqetl dag generate
+```
+
+This may take a while, as it currently does a dry run through every query defined in bigquery-etl.
```
scholtzan, line 130: `bqetl dag generate bqetl_mozregression` would be faster here ;)

scholtzan (approved):
This looks good!

docs/mkdocs.yml
Iinh: It's possible to nest marketing data under `Datasets` when we get to it. 
I agree with @jklukas that it might be better to navigate the page if we have `mozdata` and `mozfun` at the top. However I feel like it's cleaner if we nest all projects under "Datasets". Using "Datasets" and "UDFs" instead of `mozdata` and `mozfun` would be more intuitive for users who are new to the repo. Maybe we could give more context about the project names such as `mozdata` and `mozfun` in the index under each nav item?

Iinh (approved):
> Curious @Iinh if you have any thoughts on this, especially with respect to how this fits into other documentation work you've been doing recently.

I think it looks great. Seems like the only category left to add besides "Cookbook", "Datasets", and "UDFs" is a "Reference" (?) tab that includes the technical stuff from the README (bqetl CLI, best practices to schedule queries, etc.)

docs/mkdocs.yml
wlach: Yep, I think subdividing between GCP project within each category makes sense -- and in fact we already do. 😄 This is how it looks currently:

image

Remember also that eventually at least some of (maybe all?) of this information will live in the glean dictionary.

docs/cookbooks/creating_a_derived_dataset.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Initial steps
+
+Set up bigquery-etl on your system per the instructions in the [README.md](https://github.com/mozilla/bigquery-etl/blob/master/README.md).
+
+## Create the Query
+
+First, you'll need to create a query files to work off of. For this step, you'll need to know what you want your derived dataset to be called. In this case, we'll name it `org_mozilla_mozregression_derived.mozregression_aggregates`.
+
+The `org_mozilla_mozregression_derived` part represents a "dataset" in BigQuery (which is Google's unfortunately-chosen identifier for a "group of tables"). By convention, we use the `_derived` postfix to hold derived tables like this one.
```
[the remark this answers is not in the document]
wlach: Makes sense! I agree.

docs/cookbooks/creating_a_derived_dataset.md
wlach: Does that work? I'll adjust it if so. 😄

wlach: It does!

docs/cookbooks/creating_a_derived_dataset.md
wlach: Yup! Added some notes about this.

jklukas (approved):

Issue #1779, opened 2021-02-09
Adapt glean baseline ETL to reference probeinfo


jklukas:
Follow-up to https://github.com/mozilla/bigquery-etl/pull/1775 that will allow us to include baseline_clients_* tables in `generated-sql` and generated docs.

PR #1786, opened 2021-02-10
Use probeinfo rather than BQ calls for glean_usage sql generation

jklukas:
Addresses https://github.com/mozilla/bigquery-etl/issues/1779


bigquery_etl/glean_usage/common.py
```diff
@@ -35,40 +42,52 @@ def write_sql(output_dir, full_table_id, basename, sql):
...
+    resources, and this helper lets us skip those.
     """
+    with tempfile.TemporaryDirectory() as tdir:
+        tfile = Path(tdir) / "query.sql"
+        with tfile.open("w") as f:
+            f.write(
+                f"SELECT * FROM `{project_id}.{dataset_id}.baseline_v1`"
+                " WHERE DATE(submission_timestamp) = '2000-01-01'"
+            )
+            dryrun = DryRun(str(tfile))
```
acmiyaguchi, line 59: This can be dedented at least one level.

bigquery_etl/glean_usage/common.py
```diff
@@ -35,40 +42,52 @@ def write_sql(output_dir, full_table_id, basename, sql):
         f.write("\n")
 
 
-def list_baseline_tables(client, pool, project_id, only_tables, table_filter):
-    """Make parallel BQ API calls to grab all Glean baseline stable tables.
+def baseline_exists_in_dataset(project_id, dataset_id):
```
acmiyaguchi, line 45: You can reuse `referenced_table_exists`.

bigquery_etl/glean_usage/baseline_clients_daily.py
```diff
@@ -37,12 +36,14 @@
     help="Also write the query text underneath the given sql dir",
 )
 parser.add_argument(
-    "--views_only",
-    "--views-only",
+    "--output_only",
+    "--output-only",
+    "--views_only",  # Deprecated name
```
acmiyaguchi, line 41: `--output-only` and `--views-only` are aliases?

script/generate_sql
```diff
@@ -0,0 +1,41 @@
...
+    ;;
+esac
+done
+
+# Fill in any missing view definitions in the moz-fx-data-shared-prod/ folder
+python -m bigquery_etl.view.generate_stable_views --sql-dir "${SQL_DIR}" --parallelism=20
+
+# Fill in definitions for generated Glean ETL
+PLACEHOLDER_DATE="2000-01-01"
+./script/run_glean_baseline_clients_daily --project-id "${TARGET_PROJECT}" --output-dir "${SQL_DIR}/${TARGET_PROJECT}" --output-only --date="${PLACEHOLDER_DATE}"
```
acmiyaguchi, line 40: Line continuations might help with the long line length.

acmiyaguchi (approved):
Looks good to me, this switches from enumerating the tables to educated guessing about dataset existence.

Hopefully we don't run into dry run issues with a larger volume of calls.

bigquery_etl/glean_usage/baseline_clients_daily.py
jklukas: `views_only` was implemented to allow Jenkins to stably generate the view files for views publishing. But for the generated-sql context, we want to output the generated _queries_ as well. The views publishing use case isn't hurt by having the queries generated as well. I want to leave `views_only` in place since Jenkins currently hits that flag, but it's no longer semantically correct.

We can clean this up later here if we update the Jenkins job.

jklukas:
cc @whd - This adds baseline_clients_* queries to the generated-sql branch. It's intended to maintain compatibility with the existing Jenkins logic. I want to coordinate deploy with you so that we can make sure this doesn't affect view publishing.

PR #1773, opened 2021-02-08
Add Fenix experiments to experiment monitoring datasets

scholtzan:
Fixes https://github.com/mozilla/jetstream/issues/477

PR #1787, opened 2021-02-10
Fix experiment enrollment aggregates query

scholtzan:
I accidentally added this field in https://github.com/mozilla/bigquery-etl/pull/1773
If we just had schema validation ;)


sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_aggregates_base/view.sql
```diff
@@ -69,7 +69,7 @@ all_events AS (
     fenix
 )
 SELECT
-  `timestamp`,
+  date(`timestamp`) AS submission_date,
```
scholtzan, line 72: Instead of grouping on the timestamp, we should group on the date instead to aggregate data and make querying more efficient. I've done this previously with the search metrics dataset, now that we also have Fenix data the enrollment queries fail to run due to memory exceeded errors

jklukas (approved):
👍

PR #1789, opened 2021-02-10
Bug: Generated stable views were missing Glean normalization

jklukas:


bigquery_etl/view/generate_stable_views.py
```diff
@@ -220,7 +223,7 @@ def main():
                     continue
                 schemas.append(
                     SchemaFile(
-                        schema_id=schema["$schema"],
+                        schema_id=schema.get("$id", ""),
```
jklukas, line 226: This line was the real problem. We were fetching `$schema` rather than `$id` so no table ever matched based on `schema_id`.

fbertsch (approved):
Hmm, this is an important piece of code for our users, might be good to add some tests.

PR #1799, opened 2021-02-11
Add submission_date param to event_types_v1

fbertsch:


bigquery_etl/events_daily/query_templates/event_types_v1/metadata.yaml
```diff
@@ -11,5 +11,6 @@ labels:
 scheduling:
   dag_name: {{ dag_name }}
   date_partition_parameter: null
+  parameters: ["submission_date:DATE:{% raw %}{{ds}}{% endraw %}"]
```
jklukas, line 14: I have often wondered what it would look like to have two levels of Jinja templating. Good to know there are some hooks to make it possible.

fbertsch: Now we know! Also possible to do it with: {{ `{{ds}}` }}, but I think this syntax is more obvious.

jklukas (approved):

PR #1802, opened 2021-02-12
Use zetasql to find sql dependencies

relud:
Adds bqetl command to show sql file dependencies

I have follow-up PRs ready that use this for view validation and dag generation

Currently this is importing zetasql via jni, but that's actually a java jni wrapper around a c++ library, so in the future we may be able to drop the java dependency by generating an ffi library.


bigquery_etl/dependency.py
```diff
@@ -0,0 +1,86 @@
...
+
+    try:
+        Analyzer = jnius.autoclass("com.google.zetasql.Analyzer")
+        AnalyzerOptions = jnius.autoclass("com.google.zetasql.AnalyzerOptions")
+    except jnius.JavaException:
+        # replace jnius.JavaException because it's not available outside this function
+        raise ImportError(
+            "failed to import java class via jni, please download java dependencies "
+            "with: mvn dependency:copy-dependencies"
+        )
```
scholtzan, line 29: We should probably also update the [test docs](https://github.com/mozilla/bigquery-etl/blob/master/tests/README.md) and add `mvn dependency:copy-dependencies` there.

scholtzan (approved):
This looks good!

PR #1803, opened 2021-02-12
Onboarding events daily

fbertsch:


bigquery_etl/events_daily/query_templates/events_daily_v1/init.sql
```diff
@@ -21,8 +21,10 @@ SELECT
   CAST(NULL AS STRING) AS country,
   CAST(NULL AS STRING) AS subdivision1,
   -- normalized fields
+  {% if include_normalized_fields %}
```
fbertsch, line 24: The normalized fields for `messaging_system.onboarding` don't seem to fill-in correctly. The only one that does is channel. Instead I put this block as optional and include channel explicitly.

sql/moz-fx-data-shared-prod/messaging_system_derived/normalized_onboarding_events/view.sql
```diff
@@ -0,0 +1,28 @@
...
+  `moz-fx-data-shared-prod.messaging_system_derived.normalized_onboarding_events`
+AS
+SELECT
+  DATE(submission_timestamp) AS submission_date,
+  submission_timestamp AS timestamp,
+  version AS app_version,
+  message_id AS category,
+  ARRAY(
+    SELECT AS STRUCT key, value
+    FROM UNNEST([
```
fbertsch, line 11: The `event_properties` come in as a JSON field. I'd love to automatically extract all event properties from that JSON blob, but BQ JSON primitives aren't good for that kind of processing right now, and it's not worth it to expend the energy on some complicated regex expressions.

Instead, we'll pull them out individually; we'll just check-in with the UJET team about:
- which fields they need
- how we want to handle new fields

sql/moz-fx-data-shared-prod/messaging_system_derived/normalized_onboarding_events/view.sql
```diff
@@ -0,0 +1,28 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.messaging_system_derived.normalized_onboarding_events`
+AS
+SELECT
+  DATE(submission_timestamp) AS submission_date,
+  submission_timestamp AS timestamp,
+  version AS app_version,
+  message_id AS category,
```
fbertsch, line 8: This is... not exactly correct - message_id is not a category of events. But I do know that this is the primary dimension UJET needs to filter on, so it should be top-level (not an event property); and we can rename it in the Looker view without changing these underlying ETL queries.

jklukas:
I won't be able to give this a full review today. Please reassign if this is urgent.

fbertsch:
Not urgent - I have already materialized a few days of this dataset to begin Looker work. We can backfill this Tuesday/Wednesday next week.

sql/moz-fx-data-shared-prod/messaging_system_derived/normalized_onboarding_events/view.sql
jklukas:
```suggestion
  -- message_id is not a category of events, but it _is_ the primary dimension
  -- UJET needs to filter on, so it should be top-level (not an event property); 
  -- we can rename it in the Looker view.
  message_id AS category,
```

jklukas (approved):
r+wc

PR #1805, opened 2021-02-12
Fix dry run permission filtering

scholtzan:
This should fix the CI issues

### DIFF
--- a/bigquery_etl/dryrun.py
+++ b/bigquery_etl/dryrun.py
@@ -239,10 +239,12 @@ class DryRun:
         elif (
             error
             and error.get("code", None) in [400, 403]
-            and "does not have bigquery.tables.create permission for dataset"
-            in error.get("message", "")
-            or "Permission bigquery.tables.create denied on dataset"
-            in error.get("message", "")
+            and (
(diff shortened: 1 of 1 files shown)


bigquery_etl/dryrun.py
```diff
@@ -241,6 +241,8 @@ def is_valid(self):
             and error.get("code", None) in [400, 403]
             and "does not have bigquery.tables.create permission for dataset"
             in error.get("message", "")
+            or "Permission bigquery.tables.create denied on dataset"
+            in error.get("message", "")
```
scholtzan, line 245: It looks like this is the new error message when we are trying to run `init.sql` files

jklukas, line 245:
```suggestion
            and (
              "does not have bigquery.tables.create permission for dataset"
              in error.get("message", "")
              or "Permission bigquery.tables.create denied on dataset"
              in error.get("message", "")
            )
```

I don't trust my understanding of python conditional ordering enough to know whether this logic is correct. I'd prefer and extra level of parens here.

jklukas (approved):

PR #1810, opened 2021-02-16
Remove segment limit in mobile search in-content source

BenWu:
For the new top sites search attribution

cc @mmccorks


sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/mobile_search_clients_daily.template.sql
```diff
@@ -150,10 +150,13 @@ fenix_flattened_searches AS (
...
     THEN
+      -- Drop engine from search key to get source
+      -- Key should look like `engine.in-content.sap.code` with possible
+      -- additional period-separated segments (e.g. .ts for top sites)
       IF(
-        ARRAY_LENGTH(SPLIT(search.key, '.')) < 2,
+        STRPOS(search.key, '.') = 0,
         NULL,
-        ARRAY_TO_STRING(udf.array_slice(SPLIT(search.key, '.'), 1, 3), '.')
+        SUBSTR(search.key, STRPOS(search.key, '.') + 1, LENGTH(search.key))
```
sunahsuh, line 159: Doesn't seem like the length argument is necessary here

sunahsuh: Also, note if `.` is the last element we'll end up with an empty string (while if there's no `.` at all we'll get `NULL`) – it'd be better if we were consistent in the results of these edge cases

sunahsuh (approved):
This looks like it'll work fine, just a few comments to consider. FWIW, looking at the implementation for array_slice it does look like we could make the `end_index` argument arbitrarily large without causing any issues for shorter arrays, but that'd mean relying on some implementation details that I suppose may not be guaranteed in the future

sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/mobile_search_clients_daily.template.sql
BenWu: Changed to 
```sql
IF(
  STRPOS(search.key, '.') IN (0, LENGTH(search.key)),
  NULL,
  SUBSTR(search.key, STRPOS(search.key, '.') + 1)
)
```

PR #1814, opened 2021-02-16
Add option to generate stable views without dry run

relud:
generate-sql is currently the slowest CI job on this repo, usually taking 4-5min to run. This should speed that up by about 2-3 minutes.

~This is in draft because it needs data ops to set up permissions (or a cloud function) so that circleci can list stable tables in prod.~ edit: permissions no longer needed

### DIFF
--- a/bigquery_etl/view/generate_stable_views.py
+++ b/bigquery_etl/view/generate_stable_views.py
@@ -74,6 +74,13 @@ parser.add_argument(
     default="moz-fx-data-shared-prod",
     help="The project where the stable tables live.",
 )
+parser.add_argument(
+    "--no-dry-run",
+    action="store_false",
+    default=True,
+    dest="dry_run",
+    help="Don't use dry run to check whether stable tables actually exist.",
+)
 standard_args.add_log_level(parser)
 standard_args.add_parallelism(parser)
 
@@ -112,7 +119,9 @@ class SchemaFile:
         )
 
 
-def write_view_if_not_exists(target_project: str, sql_dir: Path, schema: SchemaFile):
+def write_view_if_not_exists(
+    target_project: str, sql_dir: Path, schema: SchemaFile, dry_run: bool
+):
     """If a view.sql does not already exist, write one to the target directory."""
     target_dir = (
         sql_dir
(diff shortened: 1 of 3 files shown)


bigquery_etl/view/generate_stable_views.py
```diff
@@ -166,16 +192,23 @@ def write_view_if_not_exists(target_project: str, sql_dir: Path, schema: SchemaF
...
-        with tfile.open("w") as f:
-            f.write(full_sql)
-        dryrun = DryRun(str(tfile))
-        if 404 in [e.get("code") for e in dryrun.errors()]:
-            print(
-                f"Not creating {schema.user_facing_view} since stable table does not exist"
-            )
-            return
+    stable_table_missing = False
+    if stable_tables is not None:
```
relud, line 196: using `if stable_tables:` would results in falling back to dry-run when no stable tables are detected, which occurs when the active credentials don't have permissions to list datasets and tables in `moz-fx-data-shared-prod`

script/generate_sql
```diff
@@ -32,8 +32,12 @@ case $key in
 esac
 done
 
-# Fill in any missing view definitions in the moz-fx-data-shared-prod/ folder
-python -m bigquery_etl.view.generate_stable_views --sql-dir "${SQL_DIR}" --parallelism=20
+# Fill in any missing view definitions in the target project folder
+time python -m bigquery_etl.view.generate_stable_views \
```
relud, line 36: this step goes from ~3 minutes to ~10 seconds with `--list-stable-tables`

relud:
@jasonthomas would it be preferrable to grant CI permission to list tables in prod, or to set up a cloud function for it?

jklukas:
I want to take a step back here and ask whether having some mechanism to list stable tables in BQ is the right approach here. We have previously discussed having a `generated-schemas-prod` branch of MPS where we would push schemas once `terraform apply` had run, ensuing they exist. I recently discussed this with @whd again, but then decided to work around it by using a combination of `generated-schemas` and the dryrun service to validate existence. Clearly, the amount of time taken to do the dry runs is prohibitive for CI purposes. We should think through the specifics of what a deploy workflow would look like for `generated-schemas-prod`.

@relud - Do you agree that having a `generated-schemas-prod` branch serve the need here?

relud:
> @relud - Do you agree that having a `generated-schemas-prod` branch serve the need here?

sounds perfect

jklukas:
I opened https://bugzilla.mozilla.org/show_bug.cgi?id=1693353 to track discussion of `generated-schemas-prod`.

relud:
> I opened https://bugzilla.mozilla.org/show_bug.cgi?id=1693353 to track discussion of `generated-schemas-prod`.

While that gets worked out, I have modified this to instead add a flag that skips the dry run part and set it for PRs. r? @jklukas

jklukas (approved):
I like it.

PR #1815, opened 2021-02-16
LOOK-1: Emulating glean counters using legacy iOS events

acmiyaguchi:
[jira ticket](https://jira.mozilla.com/browse/LOOK-1)

This outlines an approach to unifying the looker charts for org_mozilla_ios_firefox using historical ios events. First, we generate counts for each event by client by day. Then, we aggregate counts by the relevant client_info data before creating a dataset that is a subset of schema columns in the current glean tables.


acmiyaguchi:
@fbertsch f? This is not close to ready, but I want to verify that this approach looks sane and will work for the looker dashboards. There are a couple of missing metrics, the biggest of which are the search counts.

Barring that, there is still the work of backfilling the dataset and somehow unioning this with the glean tables in looker.

sql/moz-fx-data-shared-prod/org_mozilla_ios_derived/legacy_mobile_event_counts/view.sql
```diff
@@ -0,0 +1,65 @@
...
+    method,
+    object,
+    string_value
+)
+SELECT
+  *
+FROM
+  counts
+JOIN
+  meta_recent
```
fbertsch, line 63: I'm not sure this will work for historical client dimensions. Take, for example, `os_version`. We expect that over the period of 2019-2020 for `os_version` to change significantly as iOS updates and people buy new phones. We won't be able to see this change when using just the latest metadata.

sql/moz-fx-data-shared-prod/org_mozilla_ios_derived/legacy_mobile_event_counts_transformed/view.sql
```diff
@@ -0,0 +1,121 @@
...
+  FROM
+    extracted
+  JOIN
+    labeled
+  USING
+    (client_id)
+  GROUP BY
+    1
+)
+SELECT
```
fbertsch, line 100: This seems to be missing date

fbertsch (commented):
This is getting towards what would work with Looker. If you're going to have it be un-pivoted (string col for event name, int col for count), then we should ensure it is clustered on the event name column.

Note that the glean counters all have their own column, with one row per-client-day. It might be best to try and match that schema here, so we can union them all together in a view:
```
SELECT *, 'legacy' AS telemetry_system
FROM telemetry.legacy_ios_event_counts
UNION ALL
SELECT *, 'glean'
FROM org_mozilla_firefox_ios.event_counts
```

sql/moz-fx-data-shared-prod/org_mozilla_ios_derived/legacy_mobile_event_counts/view.sql
acmiyaguchi: This metadata is being done per client per day, so we should expect to see differences in the os_version over time from view. It was my mistake for not pushing the latest version of this view (which is now an incremental query).

acmiyaguchi:
I've made these queries incremental and have backfilled a few days of data to test this out.

```
bqetl query backfill org_mozilla_ios_firefox_derived.legacy_mobile_event_counts_v1 -s 2021-01-01 -e 2021-01-07 --dry_run
bqetl query backfill org_mozilla_ios_firefox_derived.legacy_metrics_v1 -s 2021-01-01 -e 2021-01-07 --dry_run
```

The metrics table can be updated to the most recent version of the glean schema, to be fully schema compatible.

```
bq show --format=json mozdata:org_mozilla_ios_firefox.metrics | jq '.schema.fields' > /tmp/org_mozilla_ios_firefox_metrics.schema.json
bq update moz-fx-data-shared-prod:org_mozilla_ios_firefox_derived.legacy_metrics_v1 /tmp/org_mozilla_ios_firefox_metrics.schema.json
```

It's unclear if the table can be appended once schema evolution has been applied.

acmiyaguchi:
Union requires having the columns in the same order. I will have to write a generated view in order to overcome this limitation, unfortunately.

acmiyaguchi:
This should be close to the final iteration of this query. The logic is summarized as follows:

1. Generate a row per client with aggregated metrics from core
2. Generate a row per client with aggregated metrics from event counts
3. Generate a table that is the full join of the above
4. Generate a view view of the above that unions glean ios metrics with the legacy table using the last deployed schema as the source of truth

We should then be able to drop this into the looker view as-is, probably conditionally filtering on the `telemetry_system` based on dates. We can also validate the metrics between the two systems.

I need to backfill the data and do a few queries in stmo to double check, but I don't anticipate large changes from this logic.

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox/unified_metrics/query.py
```diff
@@ -0,0 +1,143 @@
...
+                res += traverse(f"{prefix}.{name}", node["fields"])
+            else:
+                res += [f"{prefix}.{name} {dtype}"]
+        return res
+
+    res = traverse("root", schema)
+    return sorted(res)
+
+
+def generate_query(columns, table):
```
fbertsch, line 35: I'm very excited by this function, because it could be a jumping off point to unioning stable data from across all channels of an application (wherein those views don't necessarily have all the same columns, and they are not in the same order).

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox/unified_metrics/query.py
```diff
@@ -0,0 +1,143 @@
...
+    """Generate a SQL query given column names."""
+    # You would think that it's this simple (e.g. selecting columns), but it's
+    # not because we want to maintain the nested structure afterwards
+    #   formatted_columns = ",\n".join(columns)
```
fbertsch, line 39: Leftover?

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox/unified_metrics/query.py
```diff
@@ -0,0 +1,143 @@
...
+
+    acc = ""
+    prev = []
+    # sorting the columns is important
+    for col in sorted(columns):
```
fbertsch, line 44: Some description of what this algorithm is doing would be useful.

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox/unified_metrics/query.py
```diff
@@ -0,0 +1,143 @@
...
+            acc = acc.rstrip(",")
+            acc += f") as {c},"
+
+    return reformat(f"select {acc} from `{table}`")
+
+
+def main():
+    # get the most schema deploy (to the nearest 15 minutes)
+    deploys_url = (
+        "https://protosaur.dev/mps-deploys/data/mozilla_pipeline_schemas/deploys.json"
```
fbertsch, line 79: Hmm, view publishing will depend on this protosaur data being available?

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox_derived/legacy_metrics_v1/query.sql
```diff
@@ -0,0 +1,206 @@
...
+  `values` ARRAY<STRUCT<key STRING, value INT64>>,
+  labels ARRAY<STRING>
+) AS (
+  (
+    WITH summed AS (
+      SELECT
+        IF(a.key IN (SELECT * FROM UNNEST(labels)), a.key, "__unknown__") AS k,
+        SUM(a.value) AS v
+      FROM
+        UNNEST(`values`) AS a
```
fbertsch, line 11: It might make more sense for this to be `FROM UNNEST(labels)`, and then get `SUM(map.get_key(...))` (and we can avoid that subquery)

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox_derived/legacy_metrics_v1/query.sql
```diff
@@ -0,0 +1,206 @@
...
+      labeled_counter_bookmarks_delete AS bookmarks_delete,
+      labeled_counter_tabs_open AS tabs_open,
+      labeled_counter_tabs_close AS tabs_close,
+      labeled_counter_reading_list_add AS reading_list_add,
+      labeled_counter_reading_list_delete AS reading_list_delete
+    ) AS labeled_counter
+  ) AS metrics
+FROM
+  aggregated
+FULL JOIN
```
fbertsch, line 199: Aggregated was already created from extract_event, why do we need to full join here?

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox_derived/legacy_mobile_event_counts_v1/query.sql
```diff
@@ -0,0 +1,73 @@
...
+    method,
+    object,
+    string_value
+)
+SELECT
+  *
+FROM
+  counts
+JOIN
+  meta_recent
```
fbertsch, line 71: Why is this taking the most recent metadata, instead of per-day?

fbertsch (commented):
Just a few questions - we'll pick this back up on Monday

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox/unified_metrics/query.py
jklukas: I'm hoping to be able to provide this information via dry run service instead: https://github.com/mozilla-services/cloudops-infra/pull/2916

acmiyaguchi: I guess we can directly figure out the commit hash by looking at the dataset tag via a bq api call.

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox_derived/legacy_mobile_event_counts_v1/query.sql
acmiyaguchi: `meta_recent` is the per client per day metadata.

fbertsch: Got it, I misread that.

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox_derived/legacy_metrics_v1/query.sql
acmiyaguchi: Good point, I added this here because of the full join that is necessary on the derived core table. This full join is not necessary.

fbertsch: IIUC, this will only run once, so we do not need to spend time optimizing if it's correct.

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox/unified_metrics/query.py
acmiyaguchi: Kind of, I do want to highlight that the naive solution of querying all the columns doesn't work in this case though.

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox/unified_metrics/query.py
acmiyaguchi: This snippet gets the commit-hash without going through a dryrun: 

```bash
bq = bigquery.Client()
commit_hash = (
    bq.get_dataset("moz-fx-data-shared-prod.telemetry")
    # labels are a dictionary on the dataset object
    .labels.get("schemas_build_id")
    # the schemas build id is the concatenation of a timestamp and commit hash
    .split("_")[1]
)
```

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox_derived/legacy_metrics_v1/query.sql
acmiyaguchi: Is it possible to get the `__unknown__` key in this case though?

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox/unified_metrics/query.py
```diff
@@ -0,0 +1,216 @@
...
+        # check if we go deeper
+        if len(split) > 1 and len(split) > len(prev):
+            # the number of times to start nesting
+            if len(prev) == 0:
+                k = len(split) - 1
+            else:
+                k = len(split) - len(prev)
+            acc += "struct(" * k
+        # sometimes we have two structs that are the same depth e.g. metrics
+        if len(split) > 1 and len(split) == len(prev) and split[-2] != prev[-2]:
```
fbertsch, line 68: `a.b.c.d`, `a.e.f.g` => this section assumes they share the ancestor `e`. It should be more like: `while split[-i] != prev[-i]`, and then finish off the previous struct.

This is a real quagmire. Even with that adjustment, strings like `a.b.c.d.e` and `a.f.c.g.h` will fail, because the algo will see `c`, think it's done, but miss the parent `b`/`f` difference.

I see two options here:
1. Fix this algorithm to build up from the bottom (`a`/`a`, then `b`/`e` differ)
2. Clearly identify these flaws, state this is not for general purpose use yet, and do some sort of processing at the beginning to ensure the columns are part of an accepted set (i.e. glean-shaped data only, where we know max depth and comparisons).

acmiyaguchi: This is certainly an issue, it does need to consider the base case for two nested structs like metadata and metrics.

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox_derived/legacy_metrics_v1/query.sql
acmiyaguchi: Glean uses `__unknown__` as the default key, so I think this is appropriate. I think it should be a `group by k`, good catch.

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox/unified_metrics/query.py
fbertsch: Does this new method satisfy the test of `a.b.c.d` and `a.f.c.g`?

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox_derived/legacy_metrics_v1/query.sql
```diff
@@ -0,0 +1,206 @@
...
+),
+aggregated AS (
+  SELECT
+    client_id,
+    submission_date,
+    -- background can be calculated too
+    SUM(
+      IF(object = "app" AND method = "foreground", value, 0)
+    ) AS counter_glean_validation_foreground_count,
+    ANY_VALUE(
```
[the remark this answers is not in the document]
acmiyaguchi: The issue I was running into is that we can't do aggregations of aggregates. Even pushing the `SUM` into the one above, we would still have do some kind of `JOIN` or `GROUP BY`.

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox/unified_metrics/query.py
acmiyaguchi: No :(

I wish there were a way to just generate all the possible cases of depth 4 with their solutions to test the python code.

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox_derived/legacy_metrics_v1/query.sql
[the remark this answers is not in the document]
fbertsch: I may be missing something, but this worked for me:
```
aggregated AS (
  SELECT
    client_id,
    submission_date,
    labeled_counter(ARRAY_AGG(
      IF(object = "bookmark" AND method = "open", (method, value), NULL) IGNORE NULLS
    ), ["awesomebar-results", "bookmarks-panel"]) AS labeled_counter_bookmarks_open,
    SUM(IF(object = "reader-mode-open-button", value, 0)) AS counter_reader_mode_open
    ...other fields...
)
```

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox/unified_metrics/query.py
acmiyaguchi: This new method satisfies the test case that you brought up.

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox_derived/legacy_metrics_v1/query.sql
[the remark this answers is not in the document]
acmiyaguchi: This does in fact work, interesting. I definitely prefer it this way because it's more difficult to make small typos.

acmiyaguchi:
Updated commands for testing:

```
 bqetl query backfill org_mozilla_ios_firefox_derived.legacy_mobile_core_v1 -s 2021-01-01 -e 2021-01-07
bqetl query backfill org_mozilla_ios_firefox_derived.legacy_mobile_event_counts_v1 -s 2021-01-01 -e 2021-01-07
bqetl query backfill org_mozilla_ios_firefox_derived.legacy_metrics_v1 -s 2021-01-01 -e 2021-01-07
python sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox/unified_metrics/query.py
```

sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox/unified_metrics/query.py
fbertsch: Hah, then we would use that function for the code itself!

Why does this require the prev and cur be the same length? How about `a.b.c.d` and `a.b.e`? Sorry for all these edge cases :/

acmiyaguchi: I want to say this will be okay, since we'll pop out of that nested level, and then handle it assuming that the two columns are the same level. I'll add the test case though.

fbertsch (approved):
Okay, I think this is good to go now. Thanks for sticking with it.

acmiyaguchi:
I'm going to merge this in, but it's surely going to need a follow-up PR. I am thinking of cutting out the intermediate tables to make this easier to debug in the console, and I need to figure out why figures are bloated approximately 6-8x.

PR #1818, opened 2021-02-16
Restore venv from cache before build

relud:
because most of our CI steps are building a venv from scratch every time instead of using the cache

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,16 +6,19 @@ jobs:
       - image: python:3.8
     steps:
       - checkout
-      - restore_cache:
+      - &restore_venv_cache
+        restore_cache:
           keys:
             # when lock files change, use increasingly general
             # patterns to restore cache
             - &cache_key
-            # yamllint disable-line rule:line-length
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -6,16 +6,19 @@ jobs:
...
             # when lock files change, use increasingly general
             # patterns to restore cache
             - &cache_key
-            # yamllint disable-line rule:line-length
-             python-packages-v1-{{ .Branch }}-{{ checksum "requirements.in" }}-{{ checksum "requirements.txt" }}
-            - python-packages-v1-{{ .Branch }}-{{ checksum "requirements.in" }}-
-            - python-packages-v1-{{ .Branch }}-
-            - python-packages-v1-
+              # yamllint disable-line rule:line-length
+              python-3.8-packages-v1-{{ .Branch }}-{{ checksum "requirements.in" }}-{{ checksum "requirements.txt" }}
```
relud, line 16: this had been broken because the cache was poisoned by symlinks from when it was for python 3.7, and python packages are installed to a version-specific `site-packages` directory.

jklukas (approved):

PR #1811, opened 2021-02-16
Use generated SQL content as source for docs

Iinh:
Addresses #1778. Thank you @jklukas for the detailed instruction on the issue!

PR #1819, opened 2021-02-17
Strip temporary dir name from filepath for docs generation

Iinh:
After https://github.com/mozilla/bigquery-etl/pull/1811, docs now source files from `tmp/generated-sql` directory at build time. However, since the [SKIP](https://github.com/mozilla/bigquery-etl/blob/52ab0ea6bd903caca115eaf48390a0ef741d00c5/bigquery_etl/dryrun.py#L24) list only includes string literals such as `sql/moz-fx-data-shared-prod/monitoring_derived/schema_error_counts_v1/view.sql`, we need to strip `tmp/generated-sql/` from file names so that they can be checked against the SKIP list successfully.

### DIFF
--- a/bigquery_etl/docs/generate_docs.py
+++ b/bigquery_etl/docs/generate_docs.py
@@ -129,7 +129,9 @@ def main():
                         if UDF_FILE in files or PROCEDURE_FILE in files:
                             # UDF-level doc; append to dataset doc
                             dataset_name = os.path.basename(path)
-                            dataset_doc = out_dir / path.parent / f"{dataset_name}.md"
+                            dataset_doc = (
+                                Path(out_dir) / "mozfun" / f"{dataset_name}.md"
+                            )
                             docfile_content = load_with_examples(src)
                             with open(dataset_doc, "a") as dataset_doc_file:
                                 dataset_doc_file.write("\n\n")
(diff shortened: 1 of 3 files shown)


jklukas:
I'd prefer not to add logic within the package that makes a special case for the literal path `tmp/generated-sql`. Perhaps it would be possible to adapt this logic so that it strips all but the last 4 components of the path name before doing comparison, so that it can support any target directory. The project part of the path should be the first significant piece for determining whether a file should be skipped.

bigquery_etl/docs/generate_docs.py
```diff
@@ -150,7 +152,7 @@ def main():
                                 dataset_doc_file.write(f"{sourced}\n\n")
                         else:
                             # dataset-level doc; create a new doc file
-                            dest = out_dir / path / f"{name}.md"
+                            dest = Path(out_dir) / "mozfun" / f"{name}.md"
```
Iinh, line 155: Now that we're using `tmp/generated-sql` dir, we need to be more specific with the destination folder name

bigquery_etl/dryrun.py
```diff
@@ -240,10 +240,14 @@ def dry_run_result(self):
 
     def get_referenced_tables(self):
         """Return referenced tables by dry running the SQL file."""
-        if self.sqlfile not in SKIP and not self.is_valid():
+        # strip file path to 'sql/project/dataset/table/*.sql' format
+        # to be checked against the SKIP list
+        filepath = self.sqlfile[self.sqlfile.find("/sql/") + 1 :]
```
jklukas, line 245: Clever. I was assuming we'd need to split on `/`, but this is readable and looks like it should be effective.

jklukas (approved):

PR #1823, opened 2021-02-18
Add firefox_desktop_usage_2021 for KPI reporting (CDOU, new profiles)

jklukas:
Expect an additional layer after this to join with forecasts.

This includes `os` as a dimension per @irrationalagent's request.

PR #1812, opened 2021-02-16
Add app_name to mozfun.norm.product_info output

jklukas:
See the new documentation in `product_info/metadata.yaml` for full context.

I want to use this when creating derived tables for serving 2021 KPI info for mobile products in https://github.com/mozilla/bigquery-etl/pull/1824

PR #1824, opened 2021-02-18
Add mobile_usage_2021 for KPI reporting (CDOU)

jklukas:
Mobile counterpart to https://github.com/mozilla/bigquery-etl/pull/1823

Note that we don't yet have `first_seen_date` ETL for mobile, so can't include
`new_profiles` here (which we have for desktop).

This is building on the existing nondesktop exact MAU table, which filters on
`contributes_to_2020_kpi` which so far is identical to the 2021 list, so should
be fine for now.

We may need to rebuild and clean up this ETL pipeline as we improve ETL for
Glean apps in H1.


sql/moz-fx-data-shared-prod/telemetry/mobile_usage_2021/view.sql
```diff
@@ -0,0 +1,33 @@
...
+  SELECT
+    *,
+    IF(product LIKE '% iOS', 'iOS', 'Android') AS os,
+  FROM
+    `moz-fx-data-shared-prod.telemetry_derived.mobile_usage_v1`
+),
+with_product_info AS (
+  SELECT
+    *,
+    mozdata.analysis.klukas_product_info(product, os).*,
```
jklukas, line 14: I will update this once https://github.com/mozilla/bigquery-etl/pull/1812 lands.

sql/moz-fx-data-shared-prod/telemetry/mobile_usage_2021/view.sql
```diff
@@ -0,0 +1,33 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.mobile_usage_2021`
+AS
+WITH base AS (
+  SELECT
+    *,
+    IF(product LIKE '% iOS', 'iOS', 'Android') AS os,
```
acmiyaguchi, line 7: Do you have an example of a product name? Do they always have a space in between?

sql/moz-fx-data-shared-prod/telemetry/mobile_usage_2021/view.sql
```diff
@@ -0,0 +1,33 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.mobile_usage_2021`
```
acmiyaguchi, line 2: This table is named 2021, but it has data since 2017. Should be be called `mobile_usage_yearly` instead?

acmiyaguchi (approved):
The query looks fine to me, but I don't have much context on what problem this is solving. Do you have a related bug, issue, or ticket?

sql/moz-fx-data-shared-prod/telemetry/mobile_usage_2021/view.sql
irrationalagent: > The query looks fine to me, but I don't have much context on what problem this is solving. Do you have a related bug, issue, or ticket?

Jeff walked me through this the other day, its helping us build the official corporate KPI dashboards for this year in looker.

irrationalagent: The values that are in the table now always have a space and iOS apps always end in iOS so this should work.

irrationalagent (approved):

sql/moz-fx-data-shared-prod/telemetry/mobile_usage_2021/view.sql
jklukas: See https://mozilla.github.io/bigquery-etl/mozfun/norm/#product_info-udf for the table of product names. This is definitely a bit hacky, but is a stopgap until we get more fully-formed baseline ETL with historical first_seen_date, etc. I will add more commentary to the query.

sql/moz-fx-data-shared-prod/telemetry/mobile_usage_2021/view.sql
jklukas: This is a good thing to discuss. I likely should add some more prose to the description about this.

This table is primarily aimed at serving source of truth for KPI data in 2021. For 2019 and 2020, we used `firefox_nondesktop_exact_mau28` for that purpose, but the schema there has become a bit stale. I want to express that this table is source of truth for reporting in 2021, but that KPI reporting needs and measures may change in future years, necessitating a change in schema. So I want this user-facing view to be obviously 2021-specific.

There's also the Shredder problem here. The data that we backfill will not reflect users who have sent deletion requests, whereas the incremental update through the year will not reflect deletion requests. This view should give stable results; I don't want numbers to drop due to new backfills. Although for the mobile case, I expect we may need to do at least one backfill in coming months as the underlying Glean ETL matures.

PR #1828, opened 2021-02-22
Fix shredder handling for tables partitioned by _PARTITIONTIME

relud:


bigquery_etl/shredder/delete.py
```diff
@@ -231,7 +231,7 @@ def get_partition_expr(table):
     if table.range_partitioning:
         return table.range_partitioning.field
     if table.time_partitioning:
-        return f"CAST({table.time_partitioning.field} AS DATE)"
+        return f"CAST({table.time_partitioning.field or '_PARTITIONTIME'} AS DATE)"
```
relud, line 234: `table.time_partitioning.field` is `None` when the correct partitioning column is `_PARTITIONTIME`

acmiyaguchi (approved):

PR #1829, opened 2021-02-22
Bug 1693141: handle engine suffixes in search scalar keys

sunahsuh:


sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v8/query.sql
```diff
@@ -12,6 +12,38 @@ CREATE TEMP FUNCTION get_search_addon_version(active_addons ANY type) AS (
...
+
+-- Bug 1693141: add organic suffix to source or type if the engine suffix is "organic"
+CREATE TEMP FUNCTION organicize_source_or_type(engine STRING, original STRING) AS (
+CASE
+  WHEN
+    ENDS_WITH(engine, ':organic')
+  THEN
+    CASE
+      WHEN
+        # For some reason, the ad click source is "ad-click:" but type is "ad-click"
```
BenWu, line 36: I don't see any reason for it but we shouldn't change the source string now without a good reason.  You could add `:` to the type though since it's only used in this query and it would simplify this function.  I don't have a preference either way

I didn't even know you could comment with `#`.  I think it should be `--` for consistency though
```suggestion
        -- For some reason, the ad click source is "ad-click:" but type is "ad-click"
```

BenWu (commented):
I took a look at the  probe data and this looks right to me

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v8/query.sql
sunahsuh: Oh this shouldn't cause continuity issues – this just ensures both the source and the type in the case of the organic tag end up as `ad-click:organic` vs `ad-click::organic`

BenWu: Yeah I understand.  I just meant ideally we probably don't want the `:` but too late to change it now

sunahsuh:
@BenWu mind giving this another pass now that the tests pass?

BenWu (approved):
Thank you for rewriting the tests and I'm sorry you had to go through that but they were helpful for me to verify.  I just have the one question about the new column but everything looks good to me.

Are you going to backfill as well?  You should be able to just clear the dag

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v8/query.sql
```diff
@@ -268,7 +302,9 @@ windowed AS (
     SUM(IF(type = 'tagged-follow-on', count, 0)) OVER w1 AS tagged_follow_on,
     SUM(IF(type = 'sap', count, 0)) OVER w1 AS sap,
     SUM(IF(type = 'ad-click', count, 0)) OVER w1 AS ad_click,
+    SUM(IF(type = 'ad-click:organic', count, 0)) OVER w1 AS ad_click_organic,
```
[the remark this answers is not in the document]
sunahsuh: Yes, we weren't counting them before at all so these are new counts and we're breaking them out for continuity

sunahsuh:
Yes, I'll backfill to when this landed in nightly for completeness. We probably also want to modify and backfill search aggregates for the new columns as well, which will be a follow-up PR

PR #1830, opened 2021-02-23
Move generated-sql into place as sql/ in docs deploy

jklukas:
This is a quick workaround for the underlying issue with hard-coded paths for dry run skips discussed in https://github.com/mozilla/bigquery-etl/pull/1819

This will get docs publishing working again in the short term while we think about updating the logic to tolerate other paths.


jklukas:
Looks like #1819 will meet this need.

PR #1831, opened 2021-02-23
Skip dag validation if unmodified

relud:
using a modified version of the step we use to skip tests in gcp-ingestion: https://github.com/mozilla/gcp-ingestion/blob/329c0fa/.circleci/config.yml#L48-L67

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,8 +82,8 @@ jobs:
           name: Early return if this build is from a forked PR
           command: |
             if [ -n "$CIRCLE_PR_NUMBER" ]; then
-            echo "Cannot pass creds to forked PRs,
-               so marking this step successful"
+              echo "Cannot pass creds to forked PRs," \
+                "so marking this step successful"
               circleci step halt
             fi
       - *restore_venv_cache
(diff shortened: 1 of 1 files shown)


jklukas (changes requested):
The purpose of this job is different from the example we're following from gcp-ingestion.

From my experience, `validate-dags` regularly catches cases where a new query has been added with `scheduling` metadata, but the PR author hasn't run the `bqetl dag generate` machinery locally and checked in changes. The success of this task depends on the content of `sql/` and `dags/` being in sync.

relud:
> From my experience, `validate-dags` regularly catches cases where a new query has been added with `scheduling` metadata, but the PR author hasn't run the `bqetl dag generate` machinery locally and checked in changes. The success of this task depends on the content of `sql/` and `dags/` being in sync.

perhaps you are confusing this job with `verify-dags-up-to-date`?

jklukas:
> perhaps you are confusing this job with verify-dags-up-to-date?

Probably. Will look again.

jklukas (approved):
This makes perfect sense now. Thank you for the polite redirection.

PR #1835, opened 2021-02-23
Move `generated-sql/sql` content to `sql/` for docs deploy

### LINKED ISSUE #1833: Generated docs links contain `tmp/generated-sql` in paths
After https://github.com/mozilla/bigquery-etl/pull/1819 was merged, we now see all the generated views, which is great!

The links to GitHub now contain `tmp/generated-sql` in the path names. For example, the [vrbrowser.baseline docs](https://mozilla.github.io/bigquery-etl/mozdata/org_mozilla_vrbrowser/#baseline) contain a link to https://github.com/mozilla/bigquery-etl/blob/master//tmp/generated-sql/sql/moz-fx-data-shared-prod/org_mozilla_vrbrowser/baseline

The links contain `/blob/master//tmp/generated-sql/sql` which probably now needs to be `/blob/generated-sql/sql`, so they look at the generated-sql branch under top-level directly `sql`.

I believe we can sidestep the `tmp/` path problem by copying `/tmp/generated-sql` into place as in https://github.com/mozilla/bigquery-etl/pull/1830

cc @Iinh

Iinh:
Resolves #1833 

This is a continuation of the work that was started in #1830: we now move the source content for docs from `tmp/generated-sql/sql/` to `sql/`. I also updated the source URL to point at the `generated-sql` branch.

### DIFF
--- a/bigquery_etl/docs/generate_docs.py
+++ b/bigquery_etl/docs/generate_docs.py
@@ -18,8 +18,8 @@ METADATA_FILE = "metadata.yaml"
 DOCS_DIR = "docs/"
 INDEX_MD = "index.md"
 SQL_REF_RE = r"@sql\((.+)\)"
-SOURCE_URL = "https://github.com/mozilla/bigquery-etl/blob/master"
-EDIT_URL = "https://github.com/mozilla/bigquery-etl/edit/master/"
+SOURCE_URL = "https://github.com/mozilla/bigquery-etl/blob/generated-sql"
+EDIT_URL = "https://github.com/mozilla/bigquery-etl/edit/generated-sql"
 
 parser = ArgumentParser(description=__doc__)
 parser.add_argument(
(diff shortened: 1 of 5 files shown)


bigquery_etl/docs/generate_docs.py
```diff
@@ -18,8 +18,8 @@
 DOCS_DIR = "docs/"
 INDEX_MD = "index.md"
 SQL_REF_RE = r"@sql\((.+)\)"
-SOURCE_URL = "https://github.com/mozilla/bigquery-etl/blob/master"
-EDIT_URL = "https://github.com/mozilla/bigquery-etl/edit/master/"
+SOURCE_URL = "https://github.com/mozilla/bigquery-etl/blob/generated-sql"
+EDIT_URL = "https://github.com/mozilla/bigquery-etl/edit/generated-sql"
```
jklukas, line 22: Looking at this, I now realize it's more complicated than I thought 😭 

The "Edit" URL should definitely point at `master` rather than `generated-sql` since the content of `generated-sql` is machine-generated. But some of the views are don't exist in `master`. So what do we do?

I think for now, we should go ahead with this PR as-is, since it will clearly be a better state than what we currently have. But then we'll need to think about a longer-term solution.

Maybe we need to have metadata to track whether a view is automatically generated or not, and treat them differently. At least we can keep this in mind for implementing things in the Glean dictionary.

jklukas (approved):

bigquery_etl/docs/generate_docs.py
Iinh: This is indeed more complicated than I thought. Thank you for your patience in reviewing my many attempts at this.
> Maybe we need to have metadata to track whether a view is automatically generated or not, and treat them differently.
👍

Issue #1836, opened 2021-02-24
Dry run failing on org_mozilla_mozregression_derived/mozregression_aggregates_v1/init.sql


jklukas:
See https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/8210/workflows/fd478802-0847-4ddc-a4d3-c44abd8ea75a/jobs/46472

> sql/moz-fx-data-shared-prod/org_mozilla_mozregression_derived/mozregression_aggregates_v1/init.sql ERROR
 [{'code': 400, 'errors': [{'message': 'moz-fx-data-shared-prod:org_mozilla_mozregression_derived.mozregression_aggregates_v1 is not allowed for this operation because it is currently a VIEW.', 'domain': 'global', 'reason': 'invalid'}], 'response': {'headers': {'alt-svc': 'h3-29=":443"; ma=2592000,h3-T051=":443"; ma=2592000,h3-Q050=":443"; ma=2592000,h3-Q046=":443"; ma=2592000,h3-Q043=":443"; ma=2592000,quic=":443"; ma=2592000; v="46,43"', 'cache-control': 'private', 'content-encoding': 'gzip', 'content-type': 'application/json; charset=UTF-8', 'date': 'Wed, 24 Feb 2021 14:34:12 GMT', 'server': 'ESF', 'transfer-encoding': 'chunked', 'vary': 'Origin, X-Origin, Referer', 'x-content-type-options': 'nosniff', 'x-frame-options': 'SAMEORIGIN', 'x-xss-protection': '0'}}, 'message': 'moz-fx-data-shared-prod:org_mozilla_mozregression_derived.mozregression_aggregates_v1 is not allowed for this operation because it is currently a VIEW.'}]

cc @wlach @scholtzan

wlach:
@scholtzan I guess we need to change this file to reference `mozilla-public-data` instead of shared-prod?

https://github.com/mozilla/bigquery-etl/blob/master/sql/moz-fx-data-shared-prod/org_mozilla_mozregression_derived/mozregression_aggregates_v1/init.sql

See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1694447

wlach:
Strangely enough the Airflow DAG seems to be updating things fine: https://workflow.telemetry.mozilla.org/tree?dag_id=bqetl_internal_tooling even though the query also references the wrong project (?):

[...]

PR #1837, opened 2021-02-24
Set mozilla-public-data project in mozregression_aggregates_v1 init.sql

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/1836


wlach (changes requested):
Thanks! While you're here, could you also modify the query in:

https://github.com/mozilla/bigquery-etl/blob/master/docs/cookbooks/creating_a_derived_dataset.md

... and specify why you're using `mozilla-public-data` and not `moz-fx-data-share-prod`?

wlach (approved):
💯

PR #1841, opened 2021-02-24
Docs for common workflows: query scheduling

scholtzan:
Docs for common bigquery-etl workflow steps. Does this format seem useful? Are there any that should be added here?


docs/cookbooks/common_workflows.md
```diff
@@ -0,0 +1,97 @@
+# Common bigquery-etl workflows
+
+This is a quick guide of how to perform common workflows in bigquery-etl using the `bqetl` CLI.
+
+For any workflow, the bigquery-etl repositiory needs to be locally available, for example by cloning the repository, and the `bqetl` CLI needs to be installed by running `./bqetl bootstrap`.
+
+## Adding a new scheduled query
+
+1. Run `./bqetl query create <dataset>.<table>`
```
jklukas, line 9:
```suggestion
1. Run `./bqetl query create <dataset>.<table>_<version>`
```

docs/cookbooks/common_workflows.md
```diff
@@ -0,0 +1,97 @@
...
+    * Add BigQuery information such as table partitioning or clustering
+        * See [clients_daily_v6](https://github.com/mozilla/bigquery-etl/blob/master/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/metadata.yaml) for reference
+1. Run `./bqetl query validate <dataset>.<table>` to dry run and format the query
+1. To schedule the query, first select a DAG from the `./bqetl dag info` list or create a new DAG `./bqetl dag create <bqetl_new_dag>` 
+1. Run `./bqetl query schedule <dataset>.<table> --dag <bqetl_dag>` to schedule the query
+1. Run `./bqetl dag generate <bqetl_dag>` to update the DAG file
+1. Create a pull request
+    * CI fails since table doesn't exist yet
+1. PR gets reviewed and eventually approved
+1. Create destination table: `./bqetl query schema deploy`
```
jklukas, line 26: We should mention that DE will need to do this step. Perhaps it's worth assuming in these workflows that the audience does not have DE creds.

docs/cookbooks/common_workflows.md
```diff
@@ -0,0 +1,97 @@
...
+1. Merge pull-request
+1. Backfill data
+    * Option 1: via Airflow interface
+    * Option 2: `./bqetl query backfill <dataset>.<table>`
+
+## Update an existing query
+
+1. Open the `query.sql` file of the query to be updated and make changes
+1. Run `./bqetl query validate <dataset>.<table>` to dry run and format the query
+1. If the query is scheduled, run `./bqetl dag generate <bqetl_dag>` to update the DAG file
```
jklukas, line 36:
```suggestion
1. If the query scheduling metadata has changed, run `./bqetl dag generate <bqetl_dag>` to update the DAG file
```

docs/cookbooks/common_workflows.md
```diff
@@ -0,0 +1,97 @@
...
+1. Optionally, run `./bqetl query schema update <dataset>.<table>` to make local `schema.yaml` updates
```
jklukas, line 37:
```suggestion
1. If the query adds new columns, run `./bqetl query schema update <dataset>.<table>` to make local `schema.yaml` updates
```

jklukas (approved):
r+wc

The content here looks great. I do like that the format is pretty compact, which I think will make it less overwhelming for readers to scan through. This is a great initial set of workflows. We should talk about this in WG meeting and encourage folks to follow these steps next time they make changes, so that we can shake out any missing pieces.

docs/cookbooks/common_workflows.md
```diff
@@ -0,0 +1,97 @@
+# Common bigquery-etl workflows
+
+This is a quick guide of how to perform common workflows in bigquery-etl using the `bqetl` CLI.
+
+For any workflow, the bigquery-etl repositiory needs to be locally available, for example by cloning the repository, and the `bqetl` CLI needs to be installed by running `./bqetl bootstrap`.
+
+## Adding a new scheduled query
```
jklukas, line 7: We should cross-link between this section and wlach's longer-form article on adding a new scheduled query. These are good complementary views.

docs/cookbooks/common_workflows.md
```diff
@@ -0,0 +1,97 @@
...
+
+1. Run `./bqetl query create <dataset>.<table>`
+    * Specifiy the desired destination dataset and table name for `<dataset>.<table>`
```
jklukas, line 10:
```suggestion
    * Specify the desired destination dataset and table name for `<dataset>.<table>`
```

docs/cookbooks/common_workflows.md
wlach: As a note, the cookbook I made was based on an earlier version that @scholtzan shared with me. I do worry a bit about these going out of sync with each other -- I guess my question is, does it have value to have this short form version? The tutorial isn't *that* long.

docs/cookbooks/common_workflows.md
```diff
@@ -0,0 +1,97 @@
...
+    * Directories and files are generated automatically 
+1. Open `query.sql` file that has been created in `sql/moz-fx-data-shared-prod/<dataset>/<table>/` to write the query
+1. Run `./bqetl query schema update <dataset>.<table>` to generate the `schema.yaml` file
+    * Optionally add column descriptions to `schema.yaml`
```
wlach, line 14: The tutorial does not currently have this information, which was just added.

wlach (changes requested):
In general, I think this is good! This would have saved me a bunch of time in the last couple of months. The one thing I am not sure of is whether we need to duplicate the existing tutorial in short form. Scanning this over, I've already noticed at least one place where the tutorial is now out of date (the schema part) -- my preference would be to just have one place which we need to update when things change.

Open to hearing other arguments, but if it were up to me I would probably take that part out and have a table of contents like:

* Creating a derived dataset
* Other common workflows

docs/cookbooks/common_workflows.md
jklukas: I do think it's more likely that we'll keep this short form up to date. Updating the long-form involves wordsmithing and explaining context that requires more effort than updating a bullet point. Given that the long-form version is told in a specific context for creating a specific dataset, I think there's some challenge there in terms of what it means to keep it "up to date" since we'd be talking about actions that were not actually performed for the case in question. It may be reasonable to preface that long-form piece with a statement that it's a snapshot in time of the workflow as of January 2021.

scholtzan: Currently, adding a schema is not strictly required (we only have it for clients_daily) so the steps in the tutorial are still very much valid as described. I do think that in the long run we might want to migrate over to schemas, that's why I included them here. Once that becomes the standard it'd make sense to update the tutorial.

wlach: Ok, got it. Maybe best to add a comment here that this is not yet required, but recommended. Sounds like we can update the tutorial later.

wlach (approved):
Based on the discussion we just had in the platform infra meeting, I think there are still some unanswered questions here. That said, it seems like this PR is a strict improvement over what's in there now so 👍 from me.

PR #1845, opened 2021-02-25
Add Google Analytics to track docs usage

Iinh:
This should give us a better idea of how the users are using the docs.


wlach (approved):
Let's give this a try! My one concern is whether google analytics can handle analytics for a subdirectory of a domain, but let's see.

Issue #1795, opened 2021-02-10
bqetl CLI should have examples in help menu


fbertsch:
It's not traditional, but having a few examples in the CLI `--help` menu would be tremendously useful.

scholtzan:
Added in https://github.com/mozilla/bigquery-etl/pull/1846

Examples also get rendered in the new `bqetl` reference page: https://mozilla.github.io/bigquery-etl/bqetl/

PR #1846, opened 2021-02-25
Add some docs for bqetl commands

scholtzan:
Just want to see if having docs specifically on bqetl commands is useful? I think that especially having usage examples would be quite helpful.

I think this is also relatively easy to maintain. Whenever a command gets added or changed this would be the place to add docs and examples.

I'll add the rest of the docs if this is something we'd like to have.

Also addresses https://github.com/mozilla/bigquery-etl/issues/1795


docs/mkdocs.yml
```diff
@@ -30,3 +30,4 @@ nav:
         - ... | mozdata/**.md
     - UDFs:
         - ... | mozfun/**.md
+    - bqetl CLI: bqetl.md
```
wlach, line 33: I was thinking we should have a reference section for some of the stuff that's currently in the README. This would go well as a sub-topic under it.

```suggestion
    - Reference: 
      - bqetl CLI: bqetl.md
```

docs/bqetl.md
```diff
@@ -0,0 +1,126 @@
...
+  dag         Commands for managing DAGs.
+  dependency  Build and use query dependency graphs.
+  dryrun      Dry run SQL.
+  format      Format SQL.
+  glam        Tools for GLAM ETL.
+  mozfun      Commands for managing mozfun routines.
+  query       Commands for managing queries.
+  routine     Commands for managing routines.
+  stripe      Commands for Stripe ETL.
+  view        Commands for managing views.
```
wlach, line 35: I would probably skip showing the output, as it's just one more thing to keep up to date.

```suggestion
```

docs/bqetl.md
```diff
@@ -0,0 +1,126 @@
+# bqetl CLI
+
+The `bqetl` command-line tool aims to simplify working with the bigquery-etl repository by supporting common workflows, such as creating, validating and scheduling queries or adding new UDFs.
+
+## Installation
+
+Please follow the [Quick Start](https://github.com/mozilla/bigquery-etl#quick-start) for setting up bigquery-etl and the bqetl CLI.
```
wlach, line 7:
```suggestion
Follow the [Quick Start](https://github.com/mozilla/bigquery-etl#quick-start) to set up bigquery-etl and the bqetl CLI.
```

docs/bqetl.md
```diff
@@ -0,0 +1,126 @@
...
+
+Commands for managing Airflow DAGs.
+
+#### `create`
+
+When creating new DAGs, the DAG name must have a `bqetl_` prefix. Created DAGs are added to the `dags.yaml` file.
+
+```bash
+# Create new DAG
+$ ./bqetl dag create bqetl_core --schedule-interval="0 2 * * *" --owner=example@mozilla.com --description="Tables derived from the legacy telemetry `core` ping sent by various mobile applications." --start-date=2019-07-25
```
wlach, line 67: Is there any way we could pull these examples from the source?

wlach (commented):
Looks great, I have some initial thoughts.

bigquery_etl/cli/dag.py
```diff
@@ -83,7 +90,15 @@ def info(name, dags_config, with_tasks):
...
 @dag.command(
-    help="Create a new DAG with name bqetl_<dag_name>, for example: bqetl_search"
+    help="""Create a new DAG with name bqetl_<dag_name>, for example: bqetl_search
+
+    Examples:
+
+    ./bqetl dag create bqetl_core --schedule-interval="0 2 * * *"
+     --owner=example@mozilla.com
+     --description="Tables derived from `core` pings sent by mobile applications."
+     --start-date=2019-07-25
```
jklukas, line 100:
```suggestion
    ./bqetl dag create bqetl_core --schedule-interval="0 2 * * *" \\
      --owner=example@mozilla.com \\
      --description="Tables derived from `core` pings sent by mobile applications." \\
      --start-date=2019-07-25
```

I think most shells would require `\` for this to work.

docs/bqetl.md
jklukas: This is a tough one. I agree there's danger of this getting out of date, but it's also a nice concise summary of what to expect, which could provide motivation for users to go ahead and invest a few minutes in getting the CLI installed to play around with it more.

jklukas (commented):
I think the examples within click are great; they're likely to stay up to date because they exist alongside the source code that defines the commands, and they're available as users are playing around with the CLI.

I'm not confident that the longer-form documentation is going to be maintained. Would it be possible to generate this kind of content?

bigquery_etl/cli/alchemer.py
```diff
@@ -6,7 +6,22 @@
 from bigquery_etl.alchemer.survey import get_survey_data, insert_to_bq
 
 
-@click.group(help="Commands for importing alchemer data.")
+@click.group(
+    help="""Commands for importing alchemer data.
+
+Examples:
+
+\b
```
scholtzan, line 14: The `\b` marker indicates that the formatting is to be preserved for the following block

bigquery_etl/docs/bqetl/templates/commands.md
```diff
@@ -0,0 +1,40 @@
...
+{{ "#" * n }} `{{ cmd.name }}`
+
+{{ cmd.description }}
+
+{% if cmd.commands | length > 0 -%}
+{% for command in cmd.commands -%}
+    {{ command_docs(command, n + 1) }}
+{% endfor -%}
+{% else -%}
+**Usage**
```
scholtzan, line 11: Not defining this as a section to prevent mkdocs from adding an entry to the TOC.

bigquery_etl/docs/bqetl/templates/commands.md
```diff
@@ -0,0 +1,40 @@
+{% macro command_docs(cmd, n) -%}
+{{ "#" * n }} `{{ cmd.name }}`
+
+{{ cmd.description }}
+
+{% if cmd.commands | length > 0 -%}
+{% for command in cmd.commands -%}
+    {{ command_docs(command, n + 1) }}
+{% endfor -%}
```
scholtzan, line 9: Recursively traverse nested commands using Jinja macros

docs/bqetl.md
scholtzan: I think this is useful. Unfortunately, I couldn't find a way to get this information programmatically.

scholtzan:
I found a way to generate the `bqetl` command docs from the commands help texts.

Here is a screenshot of what it looks like rendered:

Screen Shot 2021-03-25 at 11 49 41

bigquery_etl/cli/alchemer.py
jklukas: At what level is the formatting getting stripped? Can we use markdown here? It would be good to avoid custom formatting concepts if possible, so any context you can provide about how you arrived at the `\b` solution would be appreciated.

bigquery_etl/cli/query.py
```diff
@@ -204,7 +217,20 @@ def create(name, sql_dir, project_id, owner, init):
 
 
 @query.command(
-    help="Schedule an existing query",
+    help="""Schedule an existing query
+
+    Examples:
```
jklukas, line 222: It looks like we have a mix of some of these multiline strings having content indented all the way left, and sometimes indented at the same level as the `help` kwarg. My preference would be to consistently indent like is done here. https://docs.python.org/3/library/textwrap.html may be useful if the indentation is causing problems.

jklukas (commented):
I'm excited to see a solution here that includes generated help text! My only concern here is about better understanding the options for marking up the help text.

bigquery_etl/cli/alchemer.py
scholtzan: This is just to ensure that the formatting in the CLI looks right. Without these markers click puts a paragraph on a single line, which looks like this:

```
./bqetl query backfill --help
Usage: bqetl query backfill [OPTIONS] NAME

  Run a backfill for a query. Additional parameters will get passed to bq.

  Examples:

  # Backfill for specific date range # second comment line ./bqetl query
  backfill telemetry_derived.ssl_ratios_v1 \   --start_date=2021-03-01 \
  --end_date=2021-03-31

  # Dryrun backfill for specific date range and exclude date ./bqetl query
  backfill telemetry_derived.ssl_ratios_v1 \   --start_date=2021-03-01 \
  --end_date=2021-03-31 \   --exclude=2021-03-03 \   --dry_run
```

Click automatically wraps long lines, but the parameters and multiline comments do not look right.

bigquery_etl/docs/bqetl/generate_bqetl_docs.py
```diff
@@ -0,0 +1,96 @@
...
+
+def extract_examples_from_help_text(help_text):
+    """Return the examples from the command help text."""
+    examples = None
+    help_text = help_text.split("\n\n")
+
+    if len(help_text) > 1:
+        examples = "\n\n".join(help_text[1:])
+        examples = examples.replace("Examples:\n\n", "")
+        examples = examples.replace("\b", "")
```
scholtzan, line 44: Removing `\b` here. It shows up in the rendered Markdown as newlines

bigquery_etl/cli/alchemer.py
jklukas: I hadn't understood that this was a Click feature that's documented: https://click.palletsprojects.com/en/7.x/documentation/#preventing-rewrapping

I thought we had custom logic here to interpret `\b`. I feel better about this now.

bigquery_etl/docs/bqetl/generate_bqetl_docs.py
jklukas: Can you add this concern as a code comment and link to https://click.palletsprojects.com/en/7.x/documentation/#preventing-rewrapping

bigquery_etl/cli/alchemer.py
scholtzan: Right, I should have linked to the docs 👍

docs/bqetl.md
wlach: Could you run `./bqetl` by itself in CI (as you do in the example), capture its output, then paste it in here?

docs/bqetl.md
```diff
@@ -0,0 +1,45 @@
+# bqetl CLI
```
wlach, line 1: Can you reference this in the README, e.g. via a link? 

https://github.com/mozilla/bigquery-etl#the-bqetl-cli

wlach (approved):
This looks very useful! Thanks for doing it. The only thing I would request before landing is an update to the README. I also have one other suggestion which you can take or leave.

docs/bqetl.md
scholtzan: I think this could be possible. My concern would be that the documentation process it might get harder to understand (some things happen in CI, some as Python scripts, some parts are static).

I'll open a follow-up issue to investigate this further at some point in the future.

PR #1847, opened 2021-02-25
Fix dryrun "'str' object has no attribute 'glob'" error

relud:


bigquery_etl/dryrun.py
```diff
@@ -416,8 +416,8 @@ def main():
         if isdir(path):
             sql_files.union(
                 sql_file
-                for glob in file_names
-                for sql_file in glob.glob(f"{args.path}/**/{glob}", recursive=True)
+                for _glob in file_names
```
acmiyaguchi, line 419: `pattern` would be a more descriptive name that doesn't clash.

.circleci/config.yml
```diff
@@ -64,7 +64,9 @@ jobs:
       - run:
           name: Dry run queries
           command: |
-            if [ "$CIRCLE_BRANCH" = master ]; then
+            if [ "$CIRCLE_BRANCH" = master ] ||
+                git log --format=%B --no-merges -n 1 |
+                grep -qF '[run-tests]'; then
```
acmiyaguchi, line 69: What's going on here? It's parsing the tag to run all the tests regardless of branch?

acmiyaguchi (approved):

PR #1848, opened 2021-02-25
Allow field addition for backfill

sunahsuh:
Forgot about this piece re: the backfill


BenWu (approved):
I assume 2020-11-21 is the first day in the dag to clear

sunahsuh:
Yes, I queried the table to make sure, and `2020-11-21` is the first day we saw the new tags

PR #1849, opened 2021-02-25
Don't run validate-dags for irrelevant changes to ci config

relud:


jklukas (changes requested):
You're actively working on changes to `.circleci` so are feeling this pain, but I don't think `.circleci` changes are a common enough workflow to warrant the additional complexity here. The bash conditional logic is getting difficult to read, and adding Ruby code, even for this limited context, feels like it could lead to difficulty debugging what's up should the behavior here change unexpectedly.

PR #1851, opened 2021-02-26
Bug 1695073 - search_aggregates: handle organic search

akkomar:


BenWu (commented):
Just need to add the columns to the unit tests.  You might also want to add `allow_field_addition_on_date` for 2020-11-21 to the search_aggregates task in the dag so you don't need to manually add the field when you backfill

BenWu (approved):

PR #1852, opened 2021-02-26
Add maven to dependabot

relud:


.github/dependabot.yml
```diff
@@ -8,5 +8,9 @@ updates:
     reviewers:
       - relud
       - jklukas
-    labels:
-      - dependencies
```
relud, line 12: this was redundant because it's the default value

acmiyaguchi (approved):

PR #1856, opened 2021-02-26
[Bug 1695333] Materialized views for experiment monitoring

scholtzan:
I would like some feedback if this approach (more details in https://bugzilla.mozilla.org/show_bug.cgi?id=1695333) makes sense.

<hr>

Things I learned:
* UDFs are not supported in materialized VIEWs
* UNION ALL is not supported
* Materialized VIEWs requires aggregations
* Materialized views cannot be replaced (`CREATE OR REPLACE` doesn't work here)
* No `STRUCT` support
* Partitioning cannot be applied to materialized views
* `CURRENT_DATE` not supported
* wrong unnesting behaviour


sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/experiment_events_live_v1/view.sql
```diff
@@ -0,0 +1,64 @@
+CREATE MATERIALIZED VIEW org_mozilla_fenix_derived.experiment_events_live_v1
```
scholtzan, line 1: I'm not sure what the right destination datasets for these materialized are. We will have materialized views which are almost identical for the following datasets:
* `org_mozilla_fenix_live`
* `org_mozilla_fenix_nightly_live`
* `org_mozilla_firefox_beta_live`
* `org_mozilla_fennec_aurora_live`
* `org_mozilla_firefox_live`

and I'm not sure whether all of them should live in `telemetry_derived` or in the Glean dataset.

The reason for having separate materialized views is that previously [a `UNION ALL`](https://github.com/mozilla/bigquery-etl/blob/9b1d4091857cec6d091c60e671afa26dbd1694be/sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_aggregates_base/view.sql#L16) was done to get live events from all Fenix-related datasets. However, materialized views do not support `UNION ALL`.

sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_aggregates_live/view.sql
```diff
@@ -35,14 +37,9 @@ SELECT
...
 FROM
-  `moz-fx-data-shared-prod.telemetry.experiment_enrollment_aggregates_hourly`
+  `moz-fx-data-shared-prod.org_mozilla_fenix_derived.experiment_events_live_v1`
 WHERE
-  window_start > (
-    SELECT
-      MAX(window_end)
-    FROM
-      `moz-fx-data-shared-prod.telemetry_derived.experiment_enrollment_aggregates_v1`
-  )
```
scholtzan, line 45: Keeping this query will cause the materialized view to re-run the underlying query, which is expensive. Using `CURRENT_DATE` prevents it from doing that.

sql/moz-fx-data-shared-prod/telemetry_derived/experiment_events_live_v1/view.sql
```diff
@@ -0,0 +1,63 @@
...
+  COUNTIF(event_method = 'update') AS update_count,
+  COUNTIF(event_method = 'enrollFailed') AS enroll_failed_count,
+  COUNTIF(event_method = 'unenrollFailed') AS unenroll_failed_count,
+  COUNTIF(event_method = 'updateFailed') AS update_failed_count,
+  COUNTIF(event_method = 'disqualification') AS disqualification_count,
+  COUNTIF(event_method = 'expose' OR event_method = 'exposure') AS exposure_count
+FROM
+  desktop_live
+WHERE
+  timestamp > TIMESTAMP('2021-03-01')
```
scholtzan, line 56: This prevents the materialized view to query data from the beginning of time (which would be expensive)

scholtzan: Materialized views do not support `CURRENT_DATE()` at the moment. Also [partition expiration can't be set on materialized views.](https://cloud.google.com/bigquery/docs/materialized-views#partition_expiration)

bigquery_etl/view/publish_views.py
```diff
@@ -23,6 +23,11 @@
     # Dataset glam-fenix-dev:glam_etl was not found
     # TODO: this should be removed if views are to be automatically deployed
     *[str(path) for path in Path("sql/glam-fenix-dev").glob("glam_etl/**/view.sql")],
+    # Materialized views
+    "telemetry_derived/experiment_events_live_v1/view.sql",
+    "org_mozilla_firefox_derived/experiment_events_live_v1/view.sql",
+    "org_mozilla_firefox_beta_derived/experiment_events_live_v1/view.sql",
+    "org_mozilla_fenix_derived/experiment_events_live_v1/view.sql",
```
jklukas, line 30: Maybe we want to call these `script.sql` or `init.sql` instead of `view.sql`? I could imagine that in the future we would reconfigure `view.sql` files to contain just the query content and not the full `CREATE VIEW` statement, but we'd likely still need to have materialized views be full DDL definitions.

`init.sql` captures the idea that this is something that's run once.

sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/experiment_events_live_v1/metadata.yaml
```diff
@@ -0,0 +1,8 @@
+friendly_name: Experiment Events Live
+description: |-
+  Materialized view of experimentation related events
+  coming from Fenix clients.
```
jklukas, line 4:
```suggestion
  coming from Fenix nightly clients.
```

sql/moz-fx-data-shared-prod/org_mozilla_firefox_derived/experiment_events_live_v1/metadata.yaml
```diff
@@ -0,0 +1,8 @@
+friendly_name: Experiment Events Live
+description: |-
+  Materialized view of experimentation related events
+  coming from Fenix clients.
```
jklukas, line 4:
```suggestion
  coming from Fenix release clients.
```

jklukas (approved):

scholtzan:
Depends on https://github.com/mozilla/telemetry-airflow/pull/1265 and https://github.com/mozilla/docker-etl/pull/19

scholtzan:
Dry run is currently failing since the views haven't been deployed, yet. I did deploy the materialized views in my sandbox and everything seems to be running fine.

Once https://github.com/mozilla/telemetry-airflow/pull/1265 and https://github.com/mozilla/docker-etl/pull/19 get approved, the changes here can also be deployed.

sql/moz-fx-data-shared-prod/org_mozilla_firefox_derived/experiment_events_live_v1/init.sql
```diff
@@ -0,0 +1,66 @@
...
+  SELECT
+    submission_timestamp AS `timestamp`,
+    event.category AS `type`,
+    -- using UNNEST twice on the same field doesn't work in materialized views
+    REGEXP_EXTRACT(TO_JSON_STRING(event.extra), '.*{"key":"branch","value":"([^"]*)"}.*') AS branch,
+    REGEXP_EXTRACT(
+      TO_JSON_STRING(event.extra),
+      '.*{"key":"experiment","value":"([^"]*)"}.*'
+    ) AS experiment,
+    event.name AS event_method
```
scholtzan, line 26: The materialized views showed some unexpected behaviour when `UNNEST`ing and `CROSS JOINING` the same field twice in this query. The `branch` value ended up being set to the `experiment`. When running the query as is (without creating the materialized view) in the console `experiment` and `branch` values showed up correctly. 

I was able to re-produce this issue with a simpler query: 
https://console.cloud.google.com/bigquery?sq=1061347137128:a129bb55dadc4a1f88fcb48a26276803

I'll open a bug for this in the BigQuery issue tracker. For now this workaround does also return the correct results.

PR #1858, opened 2021-03-01
Bug 1695654: Add a method to fetch dataset labels from dryrun service results

jklukas:
Depends on https://github.com/mozilla-services/cloudops-infra/pull/2916 ; marking as draft until that's merged.

### DIFF
--- a/bigquery_etl/dryrun.py
+++ b/bigquery_etl/dryrun.py
@@ -328,6 +328,24 @@ class DryRun:
 
         return {}
 
+    def get_dataset_labels(self):
+        """Return the labels on the default dataset by dry running the SQL file."""
+        if self.sqlfile not in SKIP and not self.is_valid():
+            raise Exception(f"Error when dry running SQL file {self.sqlfile}")
+
+        if self.sqlfile in SKIP:
+            print(f"\t...Ignoring dryrun results for {self.sqlfile}")
+            return {}
+
+        if (
+            self.dry_run_result
+            and self.dry_run_result["valid"]
+            and "datasetLabels" in self.dry_run_result
+        ):
+            return self.dry_run_result["dataset_labels"]
+
+        return {}
+
     def is_valid(self):
         """Dry run the provided SQL file and check if valid."""
         if self.dry_run_result is None:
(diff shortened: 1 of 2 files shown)


bigquery_etl/dryrun.py
```diff
@@ -328,6 +328,24 @@ def get_schema(self):
...
+        if self.sqlfile in SKIP:
+            print(f"\t...Ignoring dryrun results for {self.sqlfile}")
+            return {}
+
+        if (
+            self.dry_run_result
+            and self.dry_run_result["valid"]
+            and "datasetLabels" in self.dry_run_result
+        ):
+            return self.dry_run_result["dataset_labels"]
```
scholtzan, line 345:
```suggestion
            return self.dry_run_result["datasetLabels"]
```

scholtzan (approved):
I think the dryrun cloud function changes got merged

jklukas:
Some tests are failing, and I think the problem is that the tests are using a path that contains a placeholder dataset name, so the getDataset call in the dry run fails, and now if fails the whole dry run. Example error message:

> [{'code': 404, 'errors': [{'message': 'Not found: Dataset moz-fx-data-shared-prod:test_get_referenced_tables0', 'domain': 'global', 'reason': 'notFound'}], 'response': {'headers': {'alt-svc': 'h3-29=":443"; ma=2592000,h3-T051=":443"; ma=2592000,h3-Q050=":443"; ma=2592000,h3-Q046=":443"; ma=2592000,h3-Q043=":443"; ma=2592000,quic=":443"; ma=2592000; v="46,43"', 'cache-control': 'private', 'content-encoding': 'gzip', 'content-type': 'application/json; charset=UTF-8', 'date': 'Mon, 01 Mar 2021 20:44:19 GMT', 'server': 'ESF', 'transfer-encoding': 'chunked', 'vary': 'Origin, X-Origin, Referer', 'x-content-type-options': 'nosniff', 'x-frame-options': 'SAMEORIGIN', 'x-xss-protection': '0'}}, 'message': 'Not found: Dataset moz-fx-data-shared-prod:test_get_referenced_tables0'}]

So in the dry run service, I think we need to change the logic to still be able to return a valid result even if the getDataset part fails. Or we need to refactor tests to accommodate this.

scholtzan:
Updating the dry run service seems like a good option. I can push an update for this.

jklukas:
It also looks pretty straight-forward to update tests here to include a realistic directory structure, so I'm making an effort at that.

scholtzan:
I also pushed a fix here for the dry run function: https://github.com/mozilla-services/cloudops-infra/pull/2918

PR #1859, opened 2021-03-01
Byte udfs

fbertsch:


jklukas:
Quick initial feedback is that it's potentially problematic to name the dataset here the same as built-in type `BYTES`. And I like that `bits28` feels like kind of an unambiguous brand; you can talk about the "`bits28` functions" clearly. I don't have any concrete suggestions here yet.

sql/mozfun/bytes/extract_bits/udf.sql
```diff
@@ -0,0 +1,30 @@
+CREATE OR REPLACE FUNCTION bytes.extract_bits(b BYTES, BEGIN
+  INT64,
+  length INT64
+)
+RETURNS BYTES AS (
```
fbertsch, line 5: Dear lord, just saw this formatting. We need a way to disable the sql formatter in some cases 😱

fbertsch:
> Quick initial feedback is that it's potentially problematic to name the dataset here the same as built-in type `BYTES`. And I like that `bits28` feels like kind of an unambiguous brand; you can talk about the "`bits28` functions" clearly. I don't have any concrete suggestions here yet.

On the one hand I agree, but on the other, `bits28` is a data type that we've created. If it was available natively in BQ, we would use that type instead. See also: `json`, `hist`, and `map`.

sql/mozfun/bytes/extract_bits/udf.sql
```diff
@@ -0,0 +1,30 @@
...
+),
+8
+),
+(mozfun.bytes.bit_pos_to_byte_pos(length) * 8 - length)
+),
+mozfun.bytes.bit_pos_to_byte_pos(BEGIN
+),
+mozfun.bytes.bit_pos_to_byte_pos(length - 1)
+)
+);
```
relud, line 21: `BEGIN` is a keyword, so it needs to be escaped as ``` `begin` ```. Also, `mozfun.bytes.right_left_shift` appears to have been renamed to `mozfun.bytes.zero_right`
```suggestion
CREATE OR REPLACE FUNCTION bytes.extract_bits(b BYTES, `begin` INT64, length INT64)
RETURNS BYTES AS (
  SUBSTR(
    mozfun.bytes.zero_right(
      b << MOD(IF(`begin` > 0, `begin` - 1, LENGTH(b) * 8 + `begin`), 8),
      (mozfun.bytes.bit_pos_to_byte_pos(length) * 8 - length)
    ),
    mozfun.bytes.bit_pos_to_byte_pos(`begin`),
    mozfun.bytes.bit_pos_to_byte_pos(length - 1)
  )
);
```

relud (approved):
r+wc

jklukas:
I have been calling the `bits28`-type fields "bit patterns", which I'm not 100% confident in. I would love a name here that expresses that these functions are for interpreting bytes as a pattern representing dates of activity, which I'm assuming is the unifying abstraction here. `bytespat`, `bytepat`, `bytedates`, `bytenav`?

relud:
> I have been calling the `bits28`-type fields "bit patterns", which I'm not 100% confident in. I would love a name here that expresses that these functions are for interpreting bytes as a pattern representing dates of activity, which I'm assuming is the unifying abstraction here. `bytespat`, `bytepat`, `bytedates`, `bytenav`?

I'm used to hearing this described as bit packing, so perhaps `bitpack`?

i don't like the `bytedates`, because this is more general than just dates.

sql/mozfun/bytes/extract_bits/udf.sql
```diff
@@ -0,0 +1,20 @@
...
+
+-- Tests
+SELECT
+  extract_bits(b'\x01\xFE', 8, 8) = b'\xFF',
+  extract_bits(b'\xFF', 5, 4) = b'\xF0',
+  extract_bits(REPEAT(b'\xFF', 100), 50, 8) = b'\xFF',
+  extract_bits(b'\x0F\xF0', -12, 8) = b'\xFF',
+  extract_bits(b'\x0F\x77', 0, 8) = b'\x0F',
+  extract_bits(b'\x0F\xF0', -10, 8) = b'\xFC',
```
relud, line 20:
```suggestion
  bytes.extract_bits(b'\x01\xFE', 8, 8) = b'\xFF',
  bytes.extract_bits(b'\xFF', 5, 4) = b'\xF0',
  bytes.extract_bits(REPEAT(b'\xFF', 100), 50, 8) = b'\xFF',
  bytes.extract_bits(b'\x0F\xF0', -12, 8) = b'\xFF',
  bytes.extract_bits(b'\x0F\x77', 0, 8) = b'\x0F',
  bytes.extract_bits(b'\x0F\xF0', -10, 8) = b'\xFC',
```

fbertsch:
> I'm used to hearing this described as bit packing, so perhaps `bitpack`?
> 
> i don't like the `bytedates`, because this is more general than just dates.

I agree with that, these are more about manipulating bits in a byte array. Maybe `bytearray`?

fbertsch:
I don't know if I like `bitpack`, I'm not sure if it indicates what those functions work on.

jklukas:
If these functions are for manipulating "packed BYTES" fields, then perhaps `pbytes`? We have so far mostly used "bits" for INT64 fields, so it seems beneficial to use "bytes" rather than "bit" or "bits" to make it clear these are for byte-type fields.

But I'm also open to defaulting to `bytes`, as I intially started this discussion but haven't been able to come up with an alternative that I really want to champion.

jklukas:
> how about we keep this as bytes, and then if we end up creating more specific use-cased functions (similar to bits28), we can move those to a new dataset. Basically keep this clean with generic byte functions.

You have my blessing.

PR #1860, opened 2021-03-01
Fix use of parse_routine.sub_local_routines in validate-docs

relud:


bigquery_etl/docs/validate_docs.py
```diff
@@ -67,12 +38,21 @@ def validate(project_dirs):
...
                         # store sql in temporary file for dry_run
-                        tmp_dir = Path(tempfile.mkdtemp()) / Path(root)
+                        # DryRun will use tmp_example_file.parent.parent.name as the
+                        # default dataset, which must exist in moz-fx-data-shared-prod
+                        tmp_dir = (
+                            Path(tempfile.mkdtemp())
+                            / "moz-fx-data-shared-prod"
+                            / "tmp"
+                            / "example"
+                        )
```
relud, line 55: this fixes an issue where udf examples would have a default dataset in the wrong project. For example `mozfun.norm.fenix_app_info`, but `moz-fx-data-shared-prod:norm` does not exist because the default dataset should be `mozfun:norm`. DryRun doesn't support default datasets in a different project, so instead this overrides the default dataset to be `tmp`

bigquery_etl/docs/validate_docs.py
```diff
@@ -27,35 +27,6 @@
...
-        for udf, _ in parsed_routines.items():
-            # temporary UDFs cannot contain dots, rename UDFS
-            dry_run_sql = dry_run_sql.replace(udf, udf.replace(".", "_"))
-
-        # remove explicit project references
-        dry_run_sql = dry_run_sql.replace(os.path.basename(project_dir) + ".", "")
-
-    return dry_run_sql
-
-
```
relud, line 58: `sub_local_routines` already handles recursive references and transforming `dataset.func` and `project.dataset.func` to `datataset_func`, so I replaced this function with a call directly to `sub_local_routines`

scholtzan (approved):

PR #1861, opened 2021-03-01
Fix default dataset not found in dry runs

relud:
Over the weekend bigquery started returning errors if the default dataset does not exist for a dry run.


acmiyaguchi (approved):
LGTM, with passing dry runs.

scholtzan:
This might be related to https://github.com/mozilla-services/cloudops-infra/pull/2918. Which should fix the problem that errors are returned

relud:
fixed by mozilla-services/cloudops-infra#2918

PR #1862, opened 2021-03-02
[WIP] create adm daily engagement table

teonbrooks:
This is a mockup of the table I need to generate for a set of experiments that have been released and a rollout plan that has been set into motion. I'm tagging @harterrt as he is currently the DS contact for this project. I am happy to assist however I can but I need to push this off to someone else, either Harter or someone of the DE team.

Things that need to be completed:
- [ ] check logic to make sure it's all correct
- [ ] figure out how far back we would like to backfill (Phase 2 started August 24)
- [ ] perhaps optimize for less data reading (might already be optimized)

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/adm_engagements_daily_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/adm_engagements_daily_v1/metadata.yaml
@@ -1,0 +1,17 @@
+---
+friendly_name: adMarketplace Daily Engagements
+description: >
+  Click Volume, partitioned by day with the following facets:
+    - submission_date
+    - phase
+    - country
+    - placement
+    - partner
+    - version
+    - normalized_channel
+owners:
+  - tbrooks@mozilla.com
+labels:
+  application: firefox
+  schedule: daily
+  incremental: true
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/adm_engagements_daily_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/adm_engagements_daily_v1/metadata.yaml
@@ -10,7 +10,7 @@ description: >
     - version
     - normalized_channel
 owners:
-  - tbrooks@mozilla.com
+  - rharter@mozilla.com
 labels:
   application: firefox
   schedule: daily
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/adm_engagements_daily_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/adm_engagements_daily_v1/query.sql
@@ -1,0 +1,198 @@
+WITH
+  topsites_temp AS (
+  SELECT
+    submission_date,
+    -- might be a problem if the experiment targeting and the geo we detect are the different
+    country,
+    -- for phase 2, the events were structured differently. info contained in map
+    CASE
+      WHEN mozdata.udf.get_key(experiments, 'bug-1655185-pref-topsites-redirect-launch-us-v5-release-80-80') IS NOT NULL THEN mozdata.udf.get_key(event_map_values, 'source')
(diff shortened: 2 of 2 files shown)


sql/moz-fx-data-shared-prod/telemetry_derived/adm_engagements_daily_v1/metadata.yaml
```diff
@@ -0,0 +1,17 @@
...
+  Click Volume, partitioned by day with the following facets:
+    - submission_date
+    - phase
+    - country
+    - placement
+    - partner
+    - version
+    - normalized_channel
+owners:
+  - tbrooks@mozilla.com
```
harterrt, line 13: @teonbrooks, is this OK? I want to end your responsibility to this query, but I also don't want to plagiarize your work.

```suggestion
  - rharter@mozilla.com
```

sql/moz-fx-data-shared-prod/telemetry_derived/adm_engagements_daily_v1/metadata.yaml
```diff
@@ -0,0 +1,17 @@
...
+labels:
+  application: firefox
+  schedule: daily
+  incremental: true
```
harterrt, line 17: AFAICT, this query is not incremental (though we should make it so).

sql/moz-fx-data-shared-prod/telemetry_derived/adm_engagements_daily_v1/query.sql
```diff
@@ -0,0 +1,198 @@
...
+    event_method
+  END
+    AS interaction,
+    SPLIT(app_version, '.')[
+  OFFSET
+    (0)] AS version,
+    normalized_channel,
+    CASE
+    -- phase 2
+      WHEN mozdata.udf.get_key(experiments, 'bug-1655185-pref-topsites-redirect-launch-us-v5-release-80-80') IS NOT NULL THEN 'phase2'
```
harterrt, line 62: I'd like to avoid the repetition of these experiment slugs. I think we can simplify by creating the `phase` column in a CTE, then filtering to only rows with a non-null `phase`. AFAICT, this would prevent all of the other enumerations of these experiment slugs.

Here's a quick attempt at a UDF that might reduce some of the repetition in creating the `phase` column itself.

```sql
CREATE OR REPLACE FUNCTION map.contains_any_of(map ANY TYPE, k ANY TYPE) AS (
  SELECT
    LOGICAL_OR(key_value IN k)
  FROM
    UNNEST(map) AS key_value
);
```

harterrt (changes requested):
Thanks for this PR, Teon!

I think this query will need some clean-up before check in. I'm happy to handle the remaining work, but I'll probably fork this PR and request @teonbrooks review in a separate PR. Mostly this will ensure I understand this query and will be able to maintain it.

@teonbrooks  - does that work for you?

sql/moz-fx-data-shared-prod/telemetry_derived/adm_engagements_daily_v1/metadata.yaml
teonbrooks: yep yep, totally agree, just didn't know how to make that happen :)

sql/moz-fx-data-shared-prod/telemetry_derived/adm_engagements_daily_v1/query.sql
teonbrooks: oh cool! 😎 
can you use the function on the temp table for the case statements, there's the event schema change that puts a wrinkle in it. would be cool to have it phase dependent instead of the repeated slugs as you mentioned

teonbrooks:
> I think this query will need some clean-up before check in. I'm happy to handle the remaining work, but I'll probably fork this PR and request @teonbrooks review in a separate PR. Mostly this will ensure I understand this query and will be able to maintain it.

sure, that works for me. let's do that

PR #1863, opened 2021-03-02
Use dataset labels to speed up stable view generation

jklukas:
Builds on new dry run affordance from https://github.com/mozilla/bigquery-etl/pull/1858

We also remove the `--no-dry-run` option now since only the single dry run is now needed, and stable view generation completes in less than 2 seconds.

### DIFF
--- a/bigquery_etl/view/generate_stable_views.py
+++ b/bigquery_etl/view/generate_stable_views.py
@@ -109,9 +109,7 @@ class SchemaFile:
         )
 
 
-def write_view_if_not_exists(
-    target_project: str, sql_dir: Path, schema: SchemaFile, dry_run: bool
-):
+def write_view_if_not_exists(target_project: str, sql_dir: Path, schema: SchemaFile):
     """If a view.sql does not already exist, write one to the target directory."""
     target_dir = (
         sql_dir
@@ -265,7 +263,6 @@ def main():
                 write_view_if_not_exists,
                 args.target_project,
                 Path(args.sql_dir),
-                dry_run=args.dry_run,
             ),
             schemas,
             chunksize=1,
--- a/bigquery_etl/view/generate_stable_views.py
+++ b/bigquery_etl/view/generate_stable_views.py
@@ -12,7 +12,6 @@ checking them into the sql/ tree of the default branch of the repository.
 import json
 import logging
 import tarfile
-import tempfile
 import urllib.request
 from argparse import ArgumentParser
 from dataclasses import dataclass
@@ -21,6 +20,7 @@ from io import BytesIO
 from itertools import groupby
 from multiprocessing.pool import ThreadPool
 from pathlib import Path
+
 from typing import List
 
 from bigquery_etl.dryrun import DryRun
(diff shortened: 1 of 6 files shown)


scholtzan (approved):

jklukas:
Specific question for @relud: Can you check that removal of the `--no-dry-run` looks good? I'm expecting there are no external uses that depend on this option, so as long as we've removed all uses within the repo, we'll be good to go.

bigquery_etl/view/generate_stable_views.py
```diff
@@ -241,6 +217,25 @@ def get_stable_table_schemas() -> List[SchemaFile]:
...
+    with the most recent production schemas deploy.
+    """
+    with tempfile.TemporaryDirectory() as tdir:
+        d = Path(tdir) / target_project / "telemetry_derived" / "myquery"
+        d.mkdir(parents=True)
+        tfile = Path(d) / "query.sql"
+        with tfile.open("w") as f:
+            f.write("SELECT 1")
+        dryrun = DryRun(str(tfile))
+        build_id = dryrun.get_dataset_labels()["schemas_build_id"]
```
relud, line 234: `DryRun` does not support checking the project directory, and the service does not allow the default dataset to include a project, so accepting `target_project` here is misleading.

Also, if `content` is passed to `DryRun`, then `sqlfile` doesn't need to exist.

```suggestion
def schemas_uri():
    """Return URI for the schemas tarball deployed to moz-fx-data-shared-prod.

    We construct a fake query and send it to the dry run service in order
    to read dataset labels, which contains the commit hash associated
    with the most recent production schemas deploy.
    """
    dryrun = DryRun(
        sqlfile=str(Path("telemetry_derived") / "table" / "query.sql"),
        content="SELECT 1",
    )
    build_id = dryrun.get_dataset_labels()["schemas_build_id"]
```

bigquery_etl/view/generate_stable_views.py
```diff
@@ -241,6 +217,25 @@ def get_stable_table_schemas() -> List[SchemaFile]:
...
+    We construct a fake query and send it to the dry run service in order
+    to read dataset labels, which contains the commit hash associated
+    with the most recent production schemas deploy.
+    """
+    with tempfile.TemporaryDirectory() as tdir:
+        d = Path(tdir) / target_project / "telemetry_derived" / "myquery"
+        d.mkdir(parents=True)
+        tfile = Path(d) / "query.sql"
+        with tfile.open("w") as f:
+            f.write("SELECT 1")
```
relud, line 232: This section is removed by my other suggestion, but `Path` has `write_text` and `write_bytes` methods, so this could done with:
```suggestion
        tfile.write_text("SELECT 1")
```

relud (approved):
removal of `--no-dry-run` looks good

bigquery_etl/view/generate_stable_views.py
jklukas: > Also, if content is passed to DryRun, then sqlfile doesn't need to exist.

This makes me so happy!

Issue #1871, opened 2021-03-03
CDOU column in firefox_desktop_usage_2021 appears to be wrong


irrationalagent:
See [this query](https://sql.telemetry.mozilla.org/queries/78447/source)

This should be a cumulative sum of DAU for the year up until `submission_date`. 

I believe the problem might be here: 

https://github.com/mozilla/bigquery-etl/blob/404abcae10bf0deee1c4c2ff086fc457cef420bd/sql/moz-fx-data-shared-prod/telemetry_derived/firefox_desktop_usage_v1/query.sql#L5-L17

I think we may need to add a `ORDER BY submission_date` in the window frame.

jklukas:
I think you are entirely right. I will fix this up right now.

PR #1872, opened 2021-03-03
Bug: Incorrect window frame for CDOU calculations

jklukas:
Fixes https://github.com/mozilla/bigquery-etl/issues/1871


irrationalagent (approved):

jklukas:
I've gone ahead and applied this for desktop, and the results look correct when I rerun https://sql.telemetry.mozilla.org/queries/78447/source

jklukas:
Of course immediately after merging this, I realized it's still wrong. We're only ordering by submission_date, so the frame clause can lose some segments for the current day. I will file another follow-up PR and validate that I get the same number as I do from a query on the underlying table directly.

PR #1873, opened 2021-03-03
Bug: Fill in transient slices in CDOU calculations

jklukas:
Additional bug fix on https://github.com/mozilla/bigquery-etl/pull/1872

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/firefox_desktop_usage_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/firefox_desktop_usage_v1/query.sql
@@ -1,9 +1,108 @@
+WITH base AS (
+  SELECT
+    *,
+    EXTRACT(YEAR FROM submission_date) AS submission_year,
+  FROM
+    telemetry_derived.firefox_desktop_exact_mau28_by_dimensions_v2
+  WHERE
+    -- We completely recreate this table every night since the source table
+    -- is small and this query windows over a large time range.
+    submission_date >= '2017-01-01'
+),
+--
+-- We need a continuous list of dates to aid us later in transforming into
+-- a dense representation where each slice appears every day.
+continuous_dates AS (
+  SELECT
+    submission_date,
+    EXTRACT(YEAR FROM submission_date) AS submission_year
+  FROM
+    UNNEST(
+      GENERATE_DATE_ARRAY(
+        (SELECT MIN(submission_date) FROM base),
+        (SELECT MAX(submission_date) FROM base)
+      )
+    ) AS submission_date
+),
+--
+-- We aggregate all the counts for a given slice into an array per year,
+-- so that we can operate on entire years of data consistently.
+nested_counts_by_slice_year_sparse AS (
+  SELECT
+    submission_year,
+    id_bucket,
+    activity_segment,
+    os,
+    channel,
+    `source`,
+    medium,
+    campaign,
+    content,
+    country,
+    distribution_id,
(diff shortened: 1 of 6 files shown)


sql/moz-fx-data-shared-prod/telemetry_derived/firefox_desktop_usage_v1/query.sql
```diff
@@ -15,7 +15,17 @@ SELECT
       country,
       distribution_id
     ORDER BY
-      submission_date
+      submission_date,
```
jklukas, line 18: I need to put submission_date at the end.

jklukas:
I need to do some additional validation of this logic before proceeding.

jklukas:
There's something really subtle going on. The results from the window function are slightly lower than they should be and I have not been able to figure out why. My hunch at the moment is that there's some null handling we aren't accounting for, but haven't been able to support that with any specific evidence.

jklukas:
New theory here, that I'll pick back up on tomorrow:

If a given slice appears on 2021-01-01, but not on 2021-01-02, we don't get a 2021-01-02 row for that slice, and thus that slice's cdou doesn't show up on that day.

jklukas:
> If a given slice appears on 2021-01-01, but not on 2021-01-02, we don't get a 2021-01-02 row for that slice, and thus that slice's cdou doesn't show up on that day.

This turned out to be correct, and the answer was to fill in missing days to ensure that each slice is present consistently throughout a given year.

The complexity of the query escalated quickly due to this small wrinkle, and I've tried to be very explicit in documenting the methodology. 

@irrationalagent - Please do read through this carefully and point out any bits that aren't clear about what they're doing.

If you rerun https://sql.telemetry.mozilla.org/queries/78447/source you should find that the results are now correct. In particular, the number I get there for CDOU on 2021-01-02 exactly matches with a simple sum of dau on top of the same base table:

```
SELECT
       sum(dau) AS cdou
FROM  
  `moz-fx-data-shared-prod`.telemetry_derived.firefox_desktop_exact_mau28_by_dimensions_v2
WHERE submission_date between '2021-01-01' and '2021-01-02'
```

irrationalagent:
[I confirmed](https://sql.telemetry.mozilla.org/queries/78495/source) that the cdou column matches the running sum of total DAU per day from `firefox_desktop_exact_mau28_by_dimensions_v2`

I *think* I understand what is going on here, but a couple things:

1. To check my understanding: the value of cdou for a slice that has non-0 DAU on day n but has 0 DAU on day n+1 will be the same on both days? And the value of DAU for that slice on day n+1 will be 0 (or will it be null?)
2. It shouldn't be too much of a problem here if, for example, we suddenly start seeing a whole bunch of new unique values for (e.g.) the `source` column? This approach will add extra rows to each day, even if we only ever saw 1 DAU for that value (on just one day). I think I'm ok with this as a user, but just want to make sure a) that I'm understanding this and b) that this won't cause us something to blow up.

jklukas:
> the value of cdou for a slice that has non-0 DAU on day n but has 0 DAU on day n+1 will be the same on both days? 

Correct.

> And the value of DAU for that slice on day n+1 will be 0 (or will it be null?)

It will be NULL, but we can coalesce that to 0 in the user-facing view if that's a usability enhancement.

> if, for example, we suddenly start seeing a whole bunch of new unique values for (e.g.) the source column? This approach will add extra rows to each day, even if we only ever saw 1 DAU for that value (on just one day). I think I'm ok with this as a user, but just want to make sure a) that I'm understanding this and b) that this won't cause us something to blow up.

a) You are understanding this correctly.
b) It's a valid concern, but the pathological case is unlikely and we can adapt the logic if it does become a problem

irrationalagent (approved):

PR #1875, opened 2021-03-04
Record table references in metadata.yaml

relud:
recording this in the `generated-sql` branch would allow [mozilla/lookml-generator](https://github.com/mozilla/lookml-generator) to avoid depending on zetasql and jni.

~this adds a `references.yaml` alongside `metadata.yaml`, but maybe it would be better to add a `generated-metadata.yaml` file.~

~it is *possible* to update `metadata.yaml` via `ruamel.yaml` while preserving comments, and even enforce that this be committed and up-to-date, but it would somewhat change the formatting of the files. for example, `sql/moz-fx-data-shared-prod/mozilla_vpn_derived/subscriptions_v1/metadata.yaml` gets these changes:~
```diff
 scheduling:
   dag_name: bqetl_mozilla_vpn
   # destination is the whole table, not a single partition,
   # so don't use date_partition_parameter
-  date_partition_parameter: null
-  referenced_tables:
-    [["moz-fx-data-shared-prod", "mozilla_vpn_external", "subscriptions_v1"]]
+  date_partition_parameter:
+  referenced_tables: [["moz-fx-data-shared-prod", "mozilla_vpn_external", "subscriptions_v1"]]
```

edit: this appends a `references` section to `metadata.yaml` without checking it's content, and as a result is *not* idempotent.


jklukas:
I do not want to introduce a `references.yaml` file. This information seems clearly in the spirit of metadata that we put in `metadata.yaml` and I think it's the right place for it.

@scholtzan has added some logic for outputting a more human-friendly and stable yaml format in https://github.com/mozilla/bigquery-etl/blob/master/bigquery_etl/metadata/parse_metadata.py which may help, but doesn't address the problem of preserving comments, which in some cases are pretty critical.

> maybe it would be better to add a generated-metadata.yaml file

If the fields we're generating are specific to generation and never should appear in the `metadata.yaml`, can we directly append the generated fields to the `metadata.yaml` text?

relud:
> I do not want to introduce a `references.yaml` file. This information seems clearly in the spirit of metadata that we put in `metadata.yaml` and I think it's the right place for it.
> 
> @scholtzan has added some logic for outputting a more human-friendly and stable yaml format in https://github.com/mozilla/bigquery-etl/blob/master/bigquery_etl/metadata/parse_metadata.py which may help, but doesn't address the problem of preserving comments, which in some cases are pretty critical.

maybe it would be fine to format with `ruamel.yaml` only in the generated-sql branch, because then it doesn't matter if we have small equivalent changes like `key: null` vs `key:` and whatnot.

> > maybe it would be better to add a generated-metadata.yaml file
> 
> If the fields we're generating are specific to generation and never should appear in the `metadata.yaml`, can we directly append the generated fields to the `metadata.yaml` text so that the `metadata.yaml` in `generated-sql` keeps its formatting but includes the generated fields?

This could work, but I think we would need to have yamllint enforce strict indent. Otherwise it could end up with inconsistent indent, because it's not trivial to detect the indent configuration. For example it could result in this:
```yaml
owners:
  - dthorn@mozilla.com
references:
  view.sql:
  - project.dataset.some_table
```
instead of this:
```yaml
owners:
- dthorn@mozilla.com
references:
  view.sql:
  - project.dataset.some_table
```

relud:
> can we directly append the generated fields to the `metadata.yaml`

I've updated the PR to use this method

bigquery_etl/dependency.py
```diff
@@ -101,30 +130,31 @@ def dependency():
...
+    """Record table references in yaml."""
+    for parent, group in groupby(_get_references(paths), lambda e: e[0].parent):
+        references = {
+            path.name: table_references
+            for path, table_references in group
+            if table_references
+        }
+        if not references:
+            continue
+        with open(parent / "metadata.yaml", "a") as f:
```
jklukas, line 159: We should first open `metadata.yaml` with `yaml.load` and raise an error if it already contains a top-level `references` key. That behavior should also be noted in the command description.

bigquery_etl/dependency.py
```diff
@@ -101,30 +130,31 @@ def dependency():
...
+    help="Record table references in yaml. Requires Java.",
+)
+@click.argument(
+    "paths",
+    nargs=-1,
+    type=click.Path(file_okay=True),
+)
+def record(paths: Tuple[str, ...]):
+    """Record table references in yaml."""
+    for parent, group in groupby(_get_references(paths), lambda e: e[0].parent):
```
jklukas, line 151: How quickly does this run? Does it need to be parallelized?

jklukas (changes requested):
This looks good with the addition of some additional error checking for existing `references` content. I don't think that consistent indentation is a prerequisite here as long as we clearly label the generated part of the file.

bigquery_etl/dependency.py
relud: it takes 5 seconds in CI, so i don't think it needs to be parallelized

script/generate_sql
```diff
@@ -52,3 +52,6 @@ PLACEHOLDER_DATE="2000-01-01"
     --output-dir "${SQL_DIR}/${TARGET_PROJECT}" \
     --output-only \
     --date="${PLACEHOLDER_DATE}"
+
+# Record dependencies in yaml files
+time ./script/bqetl dependency record "${SQL_DIR}"
```
jklukas, line 57:
```suggestion
./script/bqetl dependency record "${SQL_DIR}"
```

jklukas (approved):

PR #1877, opened 2021-03-08
Correct extract_bits

fbertsch:
- Make tests actually run
- Fix extract of first byte
- Fix extract for long byte strings
- Fix extract for begin=0

### DIFF
--- a/sql/mozfun/bytes/extract_bits/udf.sql
+++ b/sql/mozfun/bytes/extract_bits/udf.sql
@@ -2,22 +2,25 @@ CREATE OR REPLACE FUNCTION bytes.extract_bits(b BYTES, `begin` INT64, length INT
 RETURNS BYTES AS (
   SUBSTR(
     mozfun.bytes.zero_right(
-      b << MOD(IF(`begin` > 0, `begin` - 1, LENGTH(b) * 8 + `begin`), 8),
-      (mozfun.bytes.bit_pos_to_byte_pos(length) * 8 - length)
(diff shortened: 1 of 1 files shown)


sql/mozfun/bytes/extract_bits/udf.sql
```diff
@@ -2,22 +2,25 @@ CREATE OR REPLACE FUNCTION bytes.extract_bits(b BYTES, `begin` INT64, length INT
 RETURNS BYTES AS (
   SUBSTR(
     mozfun.bytes.zero_right(
-      b << MOD(IF(`begin` > 0, `begin` - 1, LENGTH(b) * 8 + `begin`), 8),
-      (mozfun.bytes.bit_pos_to_byte_pos(length) * 8 - length)
+      b << IF(`begin` >= 0, GREATEST(`begin` - 1, 0), LENGTH(b) * 8 + `begin`),
+      8 * LENGTH(b) - length
```
relud, line 6: If we're going to safely handle `begin = 0`, then we should probably also handle `length > 8 * LENGTH(b)`
```suggestion
      GREATEST(8 * LENGTH(b) - length, 0)
```

relud (approved):
r+wc

sql/mozfun/bytes/extract_bits/udf.sql
fbertsch: I'll add a test for this as well.

PR #1878, opened 2021-03-08
Enforce more yaml lint rules

relud:


sql/moz-fx-data-shared-prod/account_ecosystem_derived/ecosystem_user_id_lookup_v1/metadata.yaml
```diff
@@ -17,11 +17,9 @@ scheduling:
...
   depends_on_past: true
   referenced_tables:
-    - [
-        "moz-fx-data-shared-prod",
-        "firefox_accounts_stable",
-        "account_ecosystem_v1",
-      ]
+    - - "moz-fx-data-shared-prod"
+      - "firefox_accounts_stable"
+      - "account_ecosystem_v1"
```
jklukas, line 22: Seeing this syntax here rather than `[]` is surprising to me.

docs/mkdocs.yml
```diff
@@ -19,18 +19,18 @@ theme:
...
 plugins:
     - search
     - awesome-pages
 nav:
     - index.md
     - Cookbooks:
-        - Common workflows: cookbooks/common_workflows.md
-        - Creating a derived dataset: cookbooks/creating_a_derived_dataset.md
+          - Common workflows: cookbooks/common_workflows.md
+          - Creating a derived dataset: cookbooks/creating_a_derived_dataset.md
```
jklukas, line 32: Why this much indentation here? I'd prefer to see this whole file use 2 spaces, which would be more consistent with other files in this repo.

jklukas (changes requested):
I'm concerned with the level of restriction we're placing here. Is it getting to the point where it's going to be onerous to appease the linter when humans are updating these yaml files? Do we need to have an auto-formatter in place that can apply these rules without manual edits?

docs/mkdocs.yml
relud: When filing this PR I went with whatever involved the smallest change. For the linting rules that meant consistent indent, rather than strict indent. For this file that meant 4-space indent. I'm happy to change this file to 2-space indent.

sql/moz-fx-data-shared-prod/account_ecosystem_derived/ecosystem_user_id_lookup_v1/metadata.yaml
relud: It was unexpected that yamllint didn't agree with the original indent here. I chose somewhat arbitrarily based on personal preference between this and the other option that passed yamllint indent checks, which is:
```diff
-    - - "moz-fx-data-shared-prod"
-      - "firefox_accounts_stable"
-      - "account_ecosystem_v1"
+    - [
+      "moz-fx-data-shared-prod",
+      "firefox_accounts_stable",
+      "account_ecosystem_v1"
+    ]
```

Two advantages of my choice are that it's shorter, and using `ruamel.yaml` in round-trip mode doesn't force it onto a single line.

Advantages of the alternative include preserving the existing `[]` notation, and that the separation of top-level list items is slightly more clear visually.

relud:
> I'm concerned with the level of restriction we're placing here. Is it getting to the point where it's going to be onerous to appease the linter when humans are updating these yaml files? Do we need to have an auto-formatter in place that can apply these rules without manual edits?

I don't think this will be too onerous. We have a lot of yaml files in this repo, and only a small portion didn't already pass. There were 4 rules that this PR had to fix in yaml files to pass lint:

- consistent indent
  - This is what motivated this PR in the first place, mostly because inconsistent indent makes yaml harder for me to read
  - This is the only one that would have been correctly fixed by adding additional auto-formatting, and when I tried to fix it by using `ruamel.yaml` in round-trip mode, it introduced line-length issues.
- no duplicate keys
   - I consider this an error in need of manual fixing, and it only counts as lint because it's technically allowed by the spec
- no trailing whitespace
   - already enforced and auto-applied on file changes via pre-commit hook
- trailing newline in files
   - no human-generated yaml had this issue, it was entirely due to jinja templating

jklukas (approved):
Thanks for the changes here and the analysis of effect on workflows. That addresses my major concerns, but I'd like to make sure we have a second opinion from @scholtzan before merging.

scholtzan (approved):
This looks good to me.

PR #1879, opened 2021-03-08
Materialized views search metrics

scholtzan:

PR #1882, opened 2021-03-10
Remove submission_date field from experiment_search_aggregates_v1

scholtzan:
Removes `submission_date` field that was accidentally introduced in https://github.com/mozilla/bigquery-etl/pull/1879 causing the daily runs to fail.

### DIFF
--- a/bigquery_etl/schema/__init__.py
+++ b/bigquery_etl/schema/__init__.py
@@ -49,7 +49,7 @@ class Schema:
         query = f"SELECT * FROM {project}.{dataset}.{table}"
 
         if partitioned_by:
-            query += f" WHERE {partitioned_by} = DATE('2020-01-01')"
+            query += f" WHERE DATE({partitioned_by}) = DATE('2020-01-01')"
 
         # write query to temporary file so it can get dry run
         tmp = NamedTemporaryFile()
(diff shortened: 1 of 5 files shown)


bigquery_etl/schema/__init__.py
```diff
@@ -49,7 +49,7 @@ def for_table(cls, project, dataset, table, partitioned_by=None):
         query = f"SELECT * FROM {project}.{dataset}.{table}"
 
         if partitioned_by:
-            query += f" WHERE {partitioned_by} = DATE('2020-01-01')"
+            query += f" WHERE DATE({partitioned_by}) = DATE('2020-01-01')"
```
scholtzan, line 52: I was playing around with schema validation here. Ran into this bug that causes the query to fail if `partitioned_by` is a `TIMESTAMP`

jklukas: If `partitioned_by` is already a DATE, this may fail. Can you verify that this is compatible with both dates and timestamps?

jklukas (approved):
r+wc

bigquery_etl/schema/__init__.py
scholtzan: For `clients_daily_v6` the `partitioned_by` is a `DATE` and it's running fine.

Generally, converting a `DATE` to a `DATE` shouldn't cause any problems I can think of:

```sql
WITH ts_data AS (
  SELECT TIMESTAMP('2021-03-01 11:12:22') as ts
)
SELECT ts from ts_data
WHERE DATE(DATE(DATE(ts))) > DATE('2021-01-01')
```

PR #1883, opened 2021-03-10
Bug 1695336 Pocket MAU import

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1695336

Includes some bug fixes for `bqetl query schema`, as I was trying to follow https://mozilla.github.io/bigquery-etl/cookbooks/common_workflows/#adding-a-new-scheduled-query in putting this together and ran into a few issues.

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -704,9 +704,11 @@ def deploy(ctx, name, sql_dir, project_id):
 
                 if metadata.bigquery and metadata.bigquery.time_partitioning:
                     new_table.time_partitioning = bigquery.TimePartitioning(
-                        metadata.bigquery.time_partitioning.type.bigquery_type(),
-                        metadata.bigquery.time_partitioning.field,
-                        metadata.bigquery.time_partitioning.require_partition_filter,
+                        metadata.bigquery.time_partitioning.type.bigquery_type,
+                        field=metadata.bigquery.time_partitioning.field,
+                        require_partition_filter=(
+                            metadata.bigquery.time_partitioning.require_partition_filter
+                        ),
                     )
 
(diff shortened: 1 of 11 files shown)


bigquery_etl/cli/query.py
```diff
@@ -704,9 +704,9 @@ def deploy(ctx, name, sql_dir, project_id):
 
                 if metadata.bigquery and metadata.bigquery.time_partitioning:
                     new_table.time_partitioning = bigquery.TimePartitioning(
-                        metadata.bigquery.time_partitioning.type.bigquery_type(),
```
jklukas, line 707: This was raising an error. `bigquery_type` is defined with a `@property` annotation.

bigquery_etl/cli/query.py
```diff
@@ -704,9 +704,9 @@ def deploy(ctx, name, sql_dir, project_id):
...
-                        metadata.bigquery.time_partitioning.field,
-                        metadata.bigquery.time_partitioning.require_partition_filter,
```
jklukas, line 709: The third positional param was getting interpreted as expiration millis, so this was throwing an error.

scholtzan (approved):
Looking good!

dags.yaml
```diff
@@ -402,3 +402,23 @@ bqetl_release_criteria:
...
+
+bqetl_pocket:
+  default_args:
+    depends_on_past: false
+    email:
+      - jklukas@mozilla.com
+      - telemetry-alerts@mozilla.com
+    email_on_failure: true
+    email_on_retry: true
+    end_date: null
```
scholtzan, line 414: I suppose this is something automatically added by the CLI. Might be nice to omit though in a future version.

PR #1885, opened 2021-03-12
Search metrics fix inflation

scholtzan:
The previous implementation of search metrics monitoring was inflating ` ad_clicks_count`s and `search_with_ads_count`s. This attempts to fix this problem. Since materialized views are quite limited the only way I found was to use a combination of dumping fields as JSON, extract parts using regex and creating new data structures by concatenating strings.


sql/moz-fx-data-shared-prod/telemetry_derived/experiment_search_events_live_v1/init.sql
```diff
@@ -4,57 +4,38 @@ IF
...
+    ) AS ad_clicks,
+    TO_JSON_STRING(
+      REGEXP_EXTRACT_ALL(
+        TO_JSON_STRING(payload.processes.parent.keyed_scalars.browser_search_with_ads),
+        '"value":([^,}]+)'
+      )
+    ) AS search_with_ads,
+    TO_JSON_STRING(
+      REGEXP_EXTRACT_ALL(TO_JSON_STRING(payload.keyed_histograms.search_counts), '"value":([^,}]+)')
+    ) AS search_count
```
scholtzan, line 27: This is not ideal. Querying from this materialized view ends up being quite expensive since no aggregation is being done here.

fbertsch: Can this instead `SUM` the array, and then do a `SUM` over those summed arrays? Might be something in materialized views that prevent us from doing so.

scholtzan: `SUM`ming the arrays could be done if materialized views would support sub-queries. Which is unfortunately not the case atm

sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/experiment_search_events_live_v1/init.sql
```diff
@@ -8,24 +8,91 @@ IF
...
+  --     "[[7, 0, 0], [4, 0, 0]]"
+  -- (4) Convert the nested array string to an array and concatenate
+  --     nested arrays of ping to a single array and dump as JSON string,
+  --     for example: "[[7, 0, 0], [4, 0, 0], [0, 1, 0], [0, 0, 3]]"
+  --
+  -- (5) Convert the nested array string into an array, unnest and sum
+  --     individual counts based on their array index:
+  --        ad_clicks_count = 11
+  --        search_with_ads_count = 1
+  --        search_count = 3
```
scholtzan, line 33: This is a hairy workaround. I'm hoping this step by step way of explaining what is happening here does help. I'm also open to suggestions for better solutions. At least this does produce the correct results.

sql/moz-fx-data-shared-prod/telemetry_derived/experiment_search_aggregates_v1/query.sql
```diff
@@ -3,95 +3,137 @@ WITH desktop AS (
...
+      )
+    ) AS search_with_ads_count,
+    SUM(
+      (
+        SELECT
+          SUM(`moz-fx-data-shared-prod`.udf.extract_histogram_sum(value.value))
+        FROM
+          UNNEST(payload.keyed_histograms.search_counts) AS value
+      )
+    ) AS search_count
```
scholtzan, line 29: All historical data in this derived table is basically incorrect. So this should be backfilled. I think backfilling a couple of months back should be sufficient, I'm not aware of anything, except of Grafana depending on this data.

sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/experiment_search_events_live_v1/init.sql
fbertsch: What if we just had separate materialized views for each metric type? That shouldn't lead to inflated counts, and would be more readable. Downside is some repeated code.

scholtzan: That would also be an option, though we'd end up having 12 materialized views then ([`org_mozilla_fenix`, `telemetry`, `org_mozilla_firefox_beta`, `org_mozilla_firefox_beta`] * [`ad_clicks_count`, `search_with_ads_count`, `search_count`])

fbertsch: Could we deploy those within a `sql.py` file?

jklukas: I really appreciate the thought that's gone into the documentation for this workaround. In particular, marking lines with `(3)`, etc. to refer to the logical steps makes parsing it much easier.

That said, it still would be pretty difficult to modify or debug this code if we find the results don't look right.

I think we have sufficient motivation at this point for generating these queries, which could make maintaining 12 views feasible. As long as we make sure all the generated files contain a comment refer referring to the script that created them, it should be reasonably comprehensible.

scholtzan: I ended up implementing a script for generating these queries. One limitation I didn't consider and only noticed when I test deployed the views is: https://cloud.google.com/bigquery/docs/materialized-views-faq#join_and_union_all

When we have multiple materialized views that query from the same base table (as we do for every metric) and then `JOIN` or `UNION ALL` these views, the actual base tables get queried from. Which is expensive. I added a comment here for more context: https://github.com/mozilla/bigquery-etl/pull/1904#discussion_r597216365

This issue doesn't occur in the implementation here, which now seems to be preferable?

jklukas: This is fascinating.

> This issue doesn't occur in the implementation here, which now seems to be preferable?

I agree with this conclusion. It sounds like we need to have only one materialized view per table, and string manipulation is the only viable option.

sql/moz-fx-data-shared-prod/telemetry_derived/experiment_search_aggregates_v1/query.sql
```diff
@@ -3,95 +3,137 @@ WITH desktop AS (
...
-        FROM
-          UNNEST(payload.keyed_histograms.search_counts),
-          UNNEST([REPLACE(key, 'in-content.', 'in-content:')]) AS _key,
-          UNNEST([LENGTH(REGEXP_EXTRACT(_key, '.+[.].'))]) AS pos
-      )
-    ) AS unnested_search_counts
+  GROUP BY
+    1,
+    2,
+    3
```
jklukas, line 39:
```suggestion
  GROUP BY
    submission_timestamp,
    experiment,
    branch
```

sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/experiment_search_events_live_v1/init.sql
fbertsch: Can the endpoint issue separate queries? I'd imagine this is going into Grafana, in which case these would end up in separate plots?

sql/moz-fx-data-shared-prod/org_mozilla_firefox_beta_derived/experiment_search_events_live_v1/init.sql
```diff
@@ -8,24 +8,91 @@ IF
...
+              -- (2)
+              TO_JSON_STRING(
+                REGEXP_EXTRACT_ALL(
+                  -- (1)
+                  TO_JSON_STRING(metrics.labeled_counter.browser_search_ad_clicks),
+                  '"value":([^,}]+)'
+                )
+              ),
+              r'"([^"]+)"',
+              '[\\1, 0, 0]'
```
jklukas, line 54: Are we able to also use a raw string here to avoid double backslash?

jklukas (approved):
r+

One final request is to write out grouping columns by name.

jklukas:
And again, I want to emphasize that I'm impressed with the documentation here that makes a tangled mess of string manipulation actually quite readable.

sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/experiment_search_events_live_v1/init.sql
scholtzan: Besides the Grafana plots, the data is and will also be used in additional monitoring queries to show warnings to users. Currently, we only have enrollment related queries, for example, to show a warning if the number of enrollments is too low or the number of unenrollments is too high. There'll be more such queries in the future (a proposal is in the works for Q2) which also might look at search metrics and possibly have queries that end up querying from different materialized views with the same base table again.

fbertsch: I see - I'll let you speak to whether it's doable to split those. Given it's not, this code is clearly the path forward!

sql/moz-fx-data-shared-prod/org_mozilla_firefox_derived/experiment_search_events_live_v1/init.sql
```diff
@@ -8,24 +8,91 @@ IF
...
-      unnested_search_counts.value AS search_count,
+      -- (4)
+      TO_JSON_STRING(
+        ARRAY_CONCAT(
+          JSON_EXTRACT_ARRAY(
+            -- (3)
+            REGEXP_REPLACE(
+              -- (2)
+              TO_JSON_STRING(
+                REGEXP_EXTRACT_ALL(
```
fbertsch, line 47: In the before-times of when we couldn't use UDFs in views, they would be created as a UDF and tested, and then copy+pasted into place. That approach would be good here, so that we can see the edge cases of this code. We could then point point to that file in the docs and put the long (and very helpful) docstring there.

scholtzan: Good idea!

scholtzan:
Materialized views have been deployed

PR #1886, opened 2021-03-12
Fix date partition param of pocket mau query

jklukas:
Fixup to #1883

### DIFF
--- a/dags/bqetl_pocket.py
+++ b/dags/bqetl_pocket.py
@@ -49,8 +49,9 @@ with DAG(
         project_id="moz-fx-data-shared-prod",
         owner="jklukas@mozilla.com",
         email=["jklukas@mozilla.com", "telemetry-alerts@mozilla.com"],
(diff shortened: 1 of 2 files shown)


dags/bqetl_pocket.py
```diff
@@ -49,8 +49,9 @@
         project_id="moz-fx-data-shared-prod",
         owner="jklukas@mozilla.com",
         email=["jklukas@mozilla.com", "telemetry-alerts@mozilla.com"],
-        date_partition_parameter="submission_date",
+        date_partition_parameter=None,
         depends_on_past=False,
+        parameters=["submission_date:DATE:{{ds}}"],
```
jklukas, line 54: This is our pattern for a `query.sql` on a non-partitioned table. The job is currently failing because it's using the `date_partition_parameter` to try to decorate the destination table with a date partition.

fbertsch (approved):

PR #1888, opened 2021-03-13
Add fixes for emulating ios glean pings

acmiyaguchi:
Followup to #1815, fixes a few issues in the queries. Simple counters look good now, but I noticed that values coming from the core ping are difficult to add metadata to. In particular, `COALESCE(t1.metadata, t2.metadata).*` fails because coalesce doesn't work with structs. Approximate half of the total clients seen across the two tables appear in both tables, which leaves some porttion of them missing metadata.

I made a v2 version which instead makes both the event and core tables subset of the metrics table, which is then merged together updating the schema. I tested this in looker which returns very reasonable results on counters. The queries that look over labeled counters either take a long time to finish or don't finish at all (e.g. search). This is the case prior to the view, anyhow.

```bash
bqetl query backfill org_mozilla_ios_firefox_derived.legacy_mobile_core_v2 -s 2021-01-01 -e 2021-01-01
bqetl query backfill org_mozilla_ios_firefox_derived.legacy_mobile_event_counts_v2 -s 2021-01-01 -e 2021-01-01
python sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox/unified_metrics/query.py
```

A backfill is in order, but I'd be nice to verify these results with a use-case of some kind. One thing to consider is the legacy id in the deletion-request, but I'm not sure how important that is with the counters. It's approximately 10gb/query/day (so 20gb/day for backfill). Your thoughts @fbertsch on backfill?


fbertsch (approved):
We should at least backfill through the beginning of 2020, to start with. It might be good to reach out to Ashley for some use case, my guess is search is the primary one. In that case we could also confer with @mmccorks who was going to start working on an iOS search dash in Looker.

PR #1891, opened 2021-03-15
Create adMarketplace Daily Engagements table

akkomar:


akkomar:
@harterrt this is a rewrite of #1862. I extracted experiment slugs corresponding to each phase to variable to make it a little clearer.

sql/moz-fx-data-shared-prod/telemetry/adm_engagements_daily/view.sql
```diff
@@ -0,0 +1,7 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.adm_engagements_daily`
+AS
+SELECT
+  *
+FROM
+  `moz-fx-data-shared-prod.telemetry_derived.adm_engagements_daily_v1`
```
harterrt, line 7: Looks like a check is failing because this table is not defined yet? @teonbrooks - where are these tables being saved currently?

harterrt (approved):
Thanks, @akkomar. One nit inline to address the failing CI check. Otherwise, LGTM!

Pulling the experiment slugs into `SET`s made this much more readable! Appreciate it.

When this is merged can you please start a backfill?

Thanks again!

sql/moz-fx-data-shared-prod/telemetry/adm_engagements_daily/view.sql
teonbrooks: currently there are no tables being saved or written. it's currently only forming a temporary table in redash

teonbrooks:
i'm in the git history anymore 😢 mind amending the git history to have me as a co-author? it will also help if anyone has questions down the line, they could see all the authors in the git history

sql/moz-fx-data-shared-prod/telemetry/adm_engagements_daily/view.sql
akkomar: This check is failing because this table is defined in this PR and not yet created.
I'm not sure though if you need it exposed as a view in `mozdata`?

akkomar:
> i'm in the git history anymore 😢 mind amending the git history to have me as a co-author? it will also help if anyone has questions down the line, they could see all the authors in the git history

Yeah, great idea - I added you as an owner and commit co-author.

sql/moz-fx-data-shared-prod/telemetry/adm_engagements_daily/view.sql
harterrt: We probably want this saved as a table in `mozdata`. IIUC, creating this as a view in re:dash was a convenient work around.

teonbrooks: to clarify, i was mentioned the current state in redash. for the view here, i think the issue is that the reference table for the view is being computed in this PR. are we able to create the table and the view in the same PR or do we need to land the table first, and then put the view in afterwards?

akkomar: We'll have table in shared-prod and view in mozdata. This is in line with our convention and will allow for easy access from Redash.

harterrt: SGTM!

teonbrooks (approved):

teonbrooks:
closes #1862

PR #1892, opened 2021-03-15
Remove fxa_amplitude_email_clicks

jklukas:
Per @cvalaas, the SFMC email system is now retired, so the underlying table
here is no longer receiving updates.

Depends on https://github.com/mozilla/telemetry-airflow/pull/1268


.circleci/config.yml
```diff
@@ -75,7 +75,7 @@ jobs:
               PATHS=sql
             else
               echo "Check dry run for modified queries"
-              PATHS="$(git diff origin/master... --name-only -- sql)"
+              PATHS="$(git diff origin/master... --name-only --diff-filter=d -- sql)"
```
jklukas, line 78: Needed to add this `diff-filter` because the CI logic here was detecting the deleted file, and trying to dry run it, though it no longer exists. `--diff-filter=d` means exclude deleted files.

fbertsch (approved):
👍

PR #1894, opened 2021-03-15
Bug 1697602 Remove derived tables for AET

jklukas:
Part of overall work for https://bugzilla.mozilla.org/show_bug.cgi?id=1697602

We will need to remove the underlying tables once this is merged.


bigquery_etl/shredder/config.py
```diff
@@ -95,14 +92,6 @@ def fields(self) -> Tuple[str, ...]:
...
     field="_",
 )
-ECOSYSTEM_CLIENT_ID_HMAC_SRC = DeleteSource(
-    table="account_ecosystem_restricted.ecosystem_client_id_deletion_v1",
-    field=ECOSYSTEM_CLIENT_ID_HASH,
-)
-ECOSYSTEM_CLIENT_ID_SRC = DeleteSource(
-    table="account_ecosystem_restricted.ecosystem_client_id_deletion_v1",
-    field=ECOSYSTEM_CLIENT_ID,
-)
```
jklukas, line 105: We are now dropping any remaining AET messages in the pipeline, so we will not be accruing any new entries in the stable tables, and there should be no further need for processing `ecosystem_client_id` as part of deletion requests.

bigquery_etl/shredder/config.py
```diff
@@ -369,9 +341,6 @@ def fields(self) -> Tuple[str, ...]:
         client_id_target(table="eng_workflow_stable.build_v1"),
         # other
         DeleteTarget(table="telemetry_stable.pioneer_study_v4", field=PIONEER_ID),
-        DeleteTarget(
-            table="telemetry_stable.pre_account_v4", field=DESKTOP_ECOSYSTEM_CLIENT_ID
-        ),
```
jklukas, line 374: I believe `pre_account` was added here in error. It does indeed have a field called `ecosystem_client_id`, but it is always set to "unknown". The `pre-account` ping was associated with a previous incarnation of "ecosystem telemetry" rather than AET.

relud (approved):

PR #1895, opened 2021-03-15
Add attributed to fx_usage

fbertsch:
@irrationalagent is this the correct defn?


jklukas (approved):
LGTM, and it's consistent with GUD: https://github.com/mozilla/bigquery-etl/blob/master/sql/moz-fx-data-shared-prod/telemetry_derived/smoot_usage_desktop_v2/query.sql#L5

PR #1896, opened 2021-03-15
Allow schema updates on backfill command

acmiyaguchi:
I ran into a situation where updating the schema during a backfill was useful to have around. This should add the functionality in a backwards compatible way.


scholtzan (approved):
👍

PR #1897, opened 2021-03-16
Add tables for storing and querying forecasts

fbertsch:
Here's the tables I mentioned earlier, Leif. We can update these in some way if we decide to generate other forecasts in simpleprophet.


fbertsch:
The referenced explore is here: https://github.com/mozilla/looker-spoke-default/pull/5

irrationalagent:
Looks good. What should we do about mobile? Maybe it would be best to have a similar but separate table. It would have a similar structure, the differences being:

1. And additional `app_name` column
2. No `new_profiles` and associated columns
3. Instead of a constant 1.05 target multiplier there would be a `target_lift` column that holds the target for each `app_name`. 

I wrote a table to BQ yesterday that holds the official mobile forecasts, `mozdata.analysis.mobile_forecasts_official`. The query I used is [here](
https://console.cloud.google.com/bigquery?sq=630180991450:622632e0fa804270bc844f9f271087ce). 

Alternately, we could combine desktop and mobile into one table, but that would require us to go change the lookml that was just merged. Maybe its best to keep them separate.

irrationalagent (approved):

sql/moz-fx-data-shared-prod/telemetry/forecasts_cache/view.sql
```diff
@@ -0,0 +1,51 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.forecasts_cache`
+AS
+SELECT
+  *
+FROM
+  `moz-fx-data-shared-prod.telemetry_derived.forecasts_cache_v1`
+UNION ALL
+SELECT
+  '' AS key,
```
jklukas, line 10: Is this intentional?

sql/moz-fx-data-shared-prod/telemetry/forecasts_cache/view.sql
```diff
@@ -0,0 +1,51 @@
...
+  CAST(ds AS TIMESTAMP) AS submission_date,
```
jklukas, line 11: Why timestamp rather than DATE here?

irrationalagent: I believe this is because looker requires it.

sql/moz-fx-data-shared-prod/telemetry_derived/forecasts_cache_v1/init.sql
```diff
@@ -0,0 +1,22 @@
+CREATE OR REPLACE TABLE
+  telemetry_derived.forecasts_cache_v1
```
jklukas, line 2: Can we name this something more specific like `looker_kpi_forecast_cache`?

sql/moz-fx-data-shared-prod/telemetry/forecasts_cache/view.sql
jklukas: If so, we should add a comment here to explain the mismatch.

fbertsch: Right, Looker `date` support is a bit of an oddity right now. Once I'm able to stamp down just what the behavior is for different filter comparisons I'll open an issue with them to support `date` more robustly.

sql/moz-fx-data-shared-prod/telemetry_derived/forecasts_cache_v1/init.sql
fbertsch: I think that's appropriate. I'll update the table

fbertsch:
> Alternately, we could combine desktop and mobile into one table, but that would require us to go change the lookml that was just merged. Maybe its best to keep them separate.

Let's go this route. It will let us iterate on the mobile dash without affecting the desktop one, and vice-versa.

sql/moz-fx-data-shared-prod/looker_forecasts_cache/view.sql
```diff
@@ -0,0 +1,57 @@
...
+      date_trunc(ds, YEAR)
+    ORDER BY
+      ds ASC
+  ) * 1.05 AS cum_new_profiles_target,
+  np_forecast.yhat_lower AS new_profiles_forecast_lower,
+  np_forecast.yhat_upper AS new_profiles_forecast_upper,
+FROM
+  `mozdata.analysis.loines_desktop_dau_forecast_2021-01-19` AS dau_forecast
+JOIN
+  `mozdata.analysis.loines_desktop_new_profiles_forecast_2021-01-19` AS np_forecast
```
jklukas, line 55:
```suggestion
  `moz-fx-data-shared-prod:static.desktop_dau_forecast_20210119` AS dau_forecast
JOIN
  `moz-fx-data-shared-prod:static.desktop_new_profiles_forecast_20210119` AS np_forecast
```

jklukas: I went ahead and copied these into prod `static` datasets (and removed potentially problematic dashes from the names) so that we aren't displaying contents from `analysis` dataset which could be clobbered by anyone.

sql/moz-fx-data-shared-prod/looker_forecasts_cache/view.sql
```diff
@@ -0,0 +1,57 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.looker_forecasts_cache`
```
jklukas, line 2: Is this view only intended to be used by Looker? If so, I'd say put it in `telemetry_derived` so that it's not user-facing for other contexts.

jklukas (approved):
r+wc

PR #1904, opened 2021-03-18
Generate experiment search metrics materialized views

scholtzan:
Supersedes https://github.com/mozilla/bigquery-etl/pull/1885

Edit: this doesn't work quite as expected due to some materialized view limitations (see comment below). So https://github.com/mozilla/bigquery-etl/pull/1885 is actually a more preferable/cheaper implementation.


sql/moz-fx-data-shared-prod/telemetry_derived/experiment_search_aggregates_live_v1/view.sql
```diff
@@ -8,10 +9,36 @@ WITH all_searches AS (
...
+  UNION ALL
+  SELECT
+    branch,
+    experiment,
+    window_start,
+    window_end,
+    0 AS ad_clicks_count,
     search_with_ads_count,
-    search_count
+    0 AS search_count,
```
scholtzan, line 26: Turns out when combining all materialized views we are running into the following limitation: https://cloud.google.com/bigquery/docs/materialized-views-faq#join_and_union_all

When a base table scan appears multiple times in a query plan (like it's the case here, for each metric the same `metrics` base table gets scanned) the actual base tables get scanned and queried from, so querying `experiment_search_aggregates_live` becomes very expensive.

scholtzan:
Use https://github.com/mozilla/bigquery-etl/pull/1885 instead

PR #1905, opened 2021-03-18
Add DAG for iOS Firefox ETL

acmiyaguchi:
Followup to #1815, to schedule the unified_metrics view. The directory is renamed to _v1, but it still publishes to the non-version prefix.


fbertsch (approved):

dags.yaml
```diff
@@ -415,7 +414,23 @@ bqetl_adm_engagements_daily:
...
   default_args:
     owner: akomar@mozilla.com
-    start_date: '2020-01-01'
-    email: ['akomar@mozilla.com']
+    start_date: "2020-01-01"
+    email: ["akomar@mozilla.com"]
     retries: 2
     retry_delay: 30m
+
+bqetl_org_mozilla_ios_firefox:
```
jklukas, line 422: Does this DAG only interact with the release channel of iOS, or does it reference the other namespaces as well? If it's across all channels, I think we should call the DAG `bqetl_firefox_ios` (using the `app_name`) rather than using the bq dataset name.

fbertsch: @jklukas That makes sense to me as well. We should write that up as a general guideline for Glean ETL development. I believe that should live in the README?

jklukas: Yes, it would be reasonable to document this in the README for the time being. My recommendation would definitely be to use `app_name` whenever creating a derived table that unifies several channels of a glean app. When it comes to actual locations in BQ, I think best practice right now would be to use appId-specific BQ datasets for appId-specific ETL, but to use at the point where we're unioning channels, we should put tables in `telemetry` and `telemetry_derived`, with the `app_name` embedded in the table's name. In the future, we could consider deploying `<app_name>_derived` datasets for this sort of unioned data, but that's operationally difficult right now.

fbertsch: I'm in agreement up until putting things in `telemetry` and `telemetry_derived`. I'm of the opinion that those unioned tables should belong in the release channel bq_dataset_family. A few reasons:

1. It is much easier for automated infra to find co-located tables (lookml-generator, Glean Dictionary)
2. It would be confusing for end-users to find Glean data in a `telemetry` dataset (and indeed, might not be possible)
3. Long-term we've discussed deprecating the name `telemetry`, and indeed we're trying to not use it anywhere in Looker. I'm hesitant to support adding more datasets there.

What are your thoughts?

jklukas: > I'm in agreement up until putting things in telemetry and telemetry_derived. I'm of the opinion that those unioned tables should belong in the release channel bq_dataset_family

That's definitely a reasonable alternative strategy. My counterarguments would be:

- `app_name` is fundamentally a more user-friendly and obvious identifier compared to the appId-based dataset names which are really a technical artifact of Glean ingestion. I'm not in love with pointing users at `org_mozilla_ios_firefox.unified_metrics` compared to `telemetry.firefox_ios_unified_metrics`
- Putting these tables in the release channel dataset obscures the fact that they're referencing data outside that dataset. Views under `telemetry` are more obviously representing the whole application, since that's the convention used by legacy telemetry.

I'm inclined to say let's go with the idea of putting unions under release channel bq_dataset_family, but unified tables should reference the `app_name`. Under that model, we would have `org_mozilla_ios_firefox.firefox_ios_metrics`. It's a bit verbose, but this way the table name fully defines the scope of the content, and users familiar with the convention would be able to disambiguate between this and channel-specific tables/views like `baseline_clients_daily`.

acmiyaguchi: I'll rename the dag to `bqetl_firefox_ios`, which seems to fit into existing names under the `dags.yaml`.

I'm disinclined to put this in the telemetry or telemetry_derived because the data from telemetry will be static while continuing to be updated via the Firefox iOS glean release channel. It's supposed to be a stand-in for the `org_mozilla_ios_firefox.metrics` when filtered appropriately on `telemetry_system`.

Is the `firefox_ios_metrics` evident in the context of the glean dataset (not that unified is evident in any meaningful way other than being the union between legacy and non-legacy data)?

PR #1906, opened 2021-03-18
Bug 1698578 - Parse fenix tagged ad click keys

BenWu:
dry run for search_agg is failing because new column isn't in the prod table.  I'll add the columns manually before merging instead of through airflow.

The changes:

- This adds a field `ad_click_organic` that counts only ad clicks with key `engine.in-content.organic`

- Tagged ad clicks are still counted in the same column but the source is now `in-content.sap|sap-follow-on.code`

- Old clients that still only send `engine` as the ad click key are counted with the ad click column but with source `ad-click`

@mmccorks does this logic sound correct?  Should we be using `ad-click` as the source for tagged ad clicks for continuity?


akkomar (approved):
LGTM, seems to be consistent with desktop.

PR #1907, opened 2021-03-22
Airflow DAGs: Don't send alert email on rescheduled sensor retry

akkomar:
This updates Airflow DAG template and disables retry emails for rescheduled sensors.

Similar change was pushed to https://github.com/mozilla/telemetry-airflow/pull/1253 last month in order to limit the number of alerting emails a bit.

### DIFF
--- a/bigquery_etl/query_scheduling/templates/airflow_dag.j2
+++ b/bigquery_etl/query_scheduling/templates/airflow_dag.j2
@@ -103,6 +103,7 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
         check_existence=True,
         mode='reschedule',
         pool='DATA_ENG_EXTERNALTASKSENSOR',
+        email_on_retry=False,
     )
     {{ wait_for_seen.append((dependency.dag_name, dependency.task_id)) or "" }}
     {% endif -%}
@@ -124,6 +125,7 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
         check_existence=True,
         mode='reschedule',
         pool='DATA_ENG_EXTERNALTASKSENSOR',
+        email_on_retry=False,
         dag=dag,
     )
     {{ wait_for_seen.append((task_ref.dag_name, task_ref.task_id)) or "" }}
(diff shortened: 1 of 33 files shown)


akkomar:
Hey @scholtzan wdyt? Are these alerts useful to you?

scholtzan (commented):
We had some cases in the past where these emails came in handy. If for some reason an external upstream dependency in another DAG for a task didn't successfully complete then task owners will get these retry emails as their tasks didn't actually finish. If these emails do not get sent then we might end up having tasks and DAGs running without knowing.

There are some things we could do, because most of these emails aren't very useful:
* A [`dagrun_timeout`](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/index.html#airflow.models.DAG) can be set for every DAG to a reasonably high duration. For example, if a DAG is running for more than 24 hours we could let it timeout/fail and then would get an email notification for that. This, however, seems to interfere with [the Airflow backfill](https://stackoverflow.com/questions/56601861/does-dagrun-timeout-interfere-when-backfilling-clearing-old-dags) potentially.
* Set `execution_timeout`s on the task level. Again, this could be something reasonably high, like 24 hours. `execution_timeout` could be added when generating the Airflow DAGs.

If an execution timeout is set on some level, then setting `email_on_retry=False` should be a safe thing to do.

This is actually something I talked to @jklukas about the other day. Maybe he has additional thoughts on this?

jklukas:
I think the simplest thing to do here would be to adjust DAG schedules so that we generally expect all sensors to succeed on the first try. At that point, emails about rescheduling tasks would become more meaningful.

I do want us to have some signal of failures early enough in the day that we can intervene and hopefully fix the ETL before midday Pacific when the greatest concentration of data users are going to be impacted.

I don't want to remove the signal of emails on rescheduled sensors until we implement some other layer of alerting to notify us of a bad state.

I don't know enough about the behavior of dagrun_timeout and execution_timeout to be able to recommend them, but it would be interesting to test them out and evaluate whether they're useful for us.

The "Human intervention" alternative is worth taking seriously as well. Sunah proposed a weekly rotation for Airflow failure triage, which was accepted but never implemented: https://docs.google.com/document/d/132xecjR9TowcCaAlfeKuaGgC9LKDvty18mz0zb1cO6I/edit

This is worth a discussion at data platform infra wg.

scholtzan:
I added an item to the agenda for the next wg meeting

PR #1915, opened 2021-03-25
Cluster desktop usage on attribution and country

jklukas:


jklukas:
Dry run will probably fail here due to materializing a new field. Once this is approved, I'll make the schema change and ensure tests are passing before merging.

sql/moz-fx-data-shared-prod/telemetry/firefox_desktop_usage_2021/view.sql
```diff
@@ -10,7 +10,6 @@ SELECT
     COALESCE(new_profiles, 0) AS new_profiles,
     COALESCE(cumulative_new_profiles, 0) AS cumulative_new_profiles
   ),
-  (source IS NOT NULL OR campaign IS NOT NULL) AS attributed,
```
jklukas, line 13: We materialize this now at the table level so we can cluster on it.

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_desktop_usage_v1/metadata.yaml
```diff
@@ -25,3 +25,12 @@ scheduling:
...
   date_partition_parameter: null
+bigquery:
+  time_partitioning:
+    field: submission_date
+    type: day
+    require_partition_filter: false
+  clustering:
+    fields:
+    - attributed
+    - country
```
fbertsch, line 36: We should confirm that BQ is able to pushdown this predicate through the join, since we're actually filtering on `country_name`

fbertsch (approved):
r+ with confirmation

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_desktop_usage_v1/metadata.yaml
fbertsch: If not, we can materialize `country_name` as well.

jklukas: Good callout, and I highly doubt it will be able to take advantage of this. Are we able to push the responsibility of mapping between ISO codes and country names to the Looker layer? It's not clear to me whether the following is applicable to our case:

https://docs.looker.com/reference/field-params/map_layer_name

fbertsch: I don't think so - `map_layer_name` is what Looker uses to translate the string to a specific location on a map. I'll check in with Looker support to make sure there's nothing else we can use.

fbertsch: Looker folks confirmed there's no hidden functionality for this. We could put in a huge CASE statement, but I'd much rather materialize this in the data than in the query itself :)

jklukas: Will materialize.

jklukas: I've added country_name below. We are clustering on both country and country_name, but that should be essentially free since they're a 1:1 mapping.

PR #1919, opened 2021-03-26
DAG scheduling and docs cleanup

jklukas:
This adjusts some DAG schedules in an attempt to minimize the number of
BQSensor rescheduling emails we receive under normal circumstances.

Of note, the `copy_deduplicate` DAG is often taking a little longer than an
hour to complete, meaning that DAGs starting at 02:00 are likely to hit
reschedules. We do not address that problem here in hopes that performance
improvements to copy_deduplicate can bring performance back under 1 hour.

We also make some documentation fixups, including consistently using `|`
rather than `>` or `>-` as our multi-line string indicator. This preserves
whitespace that is relevant for markdown processing.

This should partially alleviate concerns from https://github.com/mozilla/bigquery-etl/pull/1907

### DIFF
--- a/dags.yaml
+++ b/dags.yaml
@@ -27,7 +27,7 @@ bqetl_ssl_ratios:
 bqetl_amo_stats:
   schedule_interval: 0 3 * * *
   # yamllint disable rule:line-length
-  description: >-
+  description: |
     Add-on download and install statistics to power the
     [addons.mozilla.org](https://addons.mozilla.org) (AMO) stats pages.
 
@@ -42,7 +42,7 @@ bqetl_amo_stats:
 
 bqetl_vrbrowser:
   schedule_interval: 0 2 * * *
-  description: >-
+  description: |
     Custom ETL based on Glean pings from Mozilla VR Browser.
   default_args:
     owner: jklukas@mozilla.com
@@ -88,7 +88,7 @@ bqetl_mobile_search:
 
 bqetl_fxa_events:
   schedule_interval: 30 1 * * *
-  description: >-
+  description: |
     Copies data from a Firefox Accounts (FxA) project. Those source tables
     are populated via Cloud Logging (Stackdriver). We hash various fields
     as part of the import.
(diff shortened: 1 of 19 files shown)


dags.yaml
```diff
@@ -232,8 +241,12 @@ bqetl_asn_aggregates:
 # DAG for exporting query data marked as public to GCS
 # queries should not be explicitly assigned to this DAG (done automatically)
 bqetl_public_data_json:
-  schedule_interval: 0 4 * * *
-  description: The DAG exports query data marked as public to GCS
+  schedule_interval: 0 5 * * *
```
jklukas, line 244: In particular, mozreggression aggregates run at 04:00. If it's undesirable to delay public data publishing to GCS, then we can move mozregression up by 30 minutes.

cc @wlach

dags.yaml
```diff
@@ -272,7 +285,13 @@ bqetl_fenix_event_rollup:
...
 
 bqetl_stripe:
-  schedule_interval: daily
+  schedule_interval: 30 0 * * *
+  description: |
+    Daily derived tables on top of data imported from Stripe.
+
+    Depends on the `stripe` DAG which starts at midnight UTC;
+    we allow 30 minutes for that DAG to complete before this one
+    is scheduled to start.
```
jklukas, line 294: cc @relud on this Stripe scheduling change.

dags.yaml
```diff
@@ -281,6 +300,7 @@ bqetl_stripe:
     retry_delay: 5m
 
 bqetl_org_mozilla_fenix_derived:
+  schedule_interval: 0 2 * * *
```
jklukas, line 303: cc @acmiyaguchi on this scheduling change. This DAG has no hope of running until 02:00 since it relies on copy_deduplicate which runs at 01:00 and takes an hour.

dags.yaml
wlach: This information is definitely not time sensitive, so no worries with this one.

scholtzan (approved):
The scheduling changes look good to me

jklukas:
I'll hold off on merging this until Monday to avoid any surprises over the weekend.

dags.yaml
relud: the timing lgtm, but due to #1920 may break this dag

relud: #1921 provides a workaround for this dag until #1920 can be fixed.

akkomar (approved):

Issue #1914, opened 2021-03-24
Speed up doc generation


scholtzan:
Generating docs currently takes about 40 minutes. The time can be reduced, for example, by generating docs for derived datasets in parallel.

PR #1925, opened 2021-03-29
Use zetasql for doc generation

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/1914

I tested this on the `sql/` folder locally and it took about 8 seconds to generate the docs. Using dry run it takes significantly longer.


wlach (approved):
Hey @scholtzan -- this looks great and is indeed a ton faster! I tested it locally and it seems to produce identical output. I don't know if you wanted feedback from @acmiyaguchi about this specifically, but I'm comfortable merging this based on what I'm seeing.

PR #1926, opened 2021-03-29
Bug 1701712 Add events_1pct

jklukas:
This is a concrete step to allow querying events data over longer
time ranges. See https://bugzilla.mozilla.org/show_bug.cgi?id=1701712

It follows the existing pattern of `main_1pct` very closely.

Currently, Redash will reject longitudinal queries over `events` as being
too expensive, but this cannot take into account the benefits of the clustering
applied on the underlying tables. This 1% table sidesteps the issue for cases
where a 1% sample is sufficient.

We can consider expanding the time range beyond six months if we see clear use
cases for it.


wlach (approved):
Thanks Jeff! I bet this will be useful to many people. When you have a sec, can you update the documentation at https://docs.telemetry.mozilla.org/datasets/batch_view/events/reference.html to briefly mention this table?

(we should consolidate all of this documentation inside this repository, but I think I'm still not quite ready to pull the trigger on that-- working on dataset support for the glean dictionary this week)

PR #1930, opened 2021-03-30
Fix referenced_tables error in generate_docs

scholtzan:


wlach (approved):
👍 looks like the right fix! I'm kind of surprised that flake8 can't pick up things like this.

PR #1934, opened 2021-03-31
[DS-1424] Implementation of mobile clients first seen

acmiyaguchi:
[jira link](https://jira.mozilla.com/browse/DS-1424), also see the [mobile clients first seen proposal](https://docs.google.com/document/d/13u4FgT0IsMH1pphYZZiUCzAskfa3r3a98ymnBuNdDWM/edit#) and related [proposal for cross-ping mobile etl](https://docs.google.com/document/d/1aayc2G0qFJP7EOSvELgUsYklpyixeZChVMYtHlbCpw8/edit#heading=h.9zc3l8c1meuo).

This adds ETL in the same style as `baseline_clients_daily` and `baseline_clients_last_seen`. The responsibility of this ETL is to get the first submission timestamp for each client. If there is a way to link the glean client ids with a legacy client id, this is where the join will occur.


bigquery_etl/glean_usage/templates/baseline_clients_first_seen.view.sql
```diff
@@ -0,0 +1,9 @@
+{{ header }}
+
+CREATE OR REPLACE VIEW
+  `{{ first_seen_view }}`
```
jklukas, line 4: I don't think we should create this user-facing view right now. Once we get the first seen tables created and backfilled, we'll need to update the daily and last_seen queries to pull in `first_seen_date` and materialize a boolean `is_first_seen` field for clustering. The `baseline_clients_first_seen` view should point at `baseline_clients_daily_v1` and filter for `is_first_seen` being set.

bigquery_etl/glean_usage/templates/baseline_clients_first_seen_v1.sql
```diff
@@ -0,0 +1,97 @@
...
+  client_info.client_id
+FROM
+  `{{ baseline_table }}`
+WHERE
+  FALSE
+
+{% else %}
+
+WITH
+{% if fennec_id %}
```
jklukas, line 26: I had imagined crafting the Fenix and Firefox for iOS queries by hand in the normal `sql/` tree, but encoding all of this in one place as you're doing is more understandable and handles the dataset-per-channel problem nicely. So I like this.

bigquery_etl/glean_usage/templates/baseline_clients_first_seen.view.sql
acmiyaguchi: The script logic at the moment does best when there is a user facing view; it's a little tricky to decouple since the view is dry-run to get the table reference. I don't see too much harm in exposing this if it's going to be a subset of the final set of columns.

bigquery_etl/glean_usage/templates/baseline_clients_first_seen_v1.init.sql
```diff
@@ -0,0 +1,39 @@
...
+{% from "macros.sql" import core_clients_first_seen %}
+
+CREATE TABLE IF NOT EXISTS
+  `{{ first_seen_table }}`
+PARTITION BY
+  first_seen_date
+CLUSTER BY
+  sample_id
+OPTIONS
+  (require_partition_filter = FALSE)
```
acmiyaguchi, line 11: Setting this to `TRUE` breaks testing, which relies on `select * from {{ table }}`. It might not be that important to have a partition filter on these tables anyhow.

acmiyaguchi:
To generate schemas for tests, some form of the following command is useful:

```bash
bq show --format json moz-fx-data-shared-prod:org_mozilla_fenix_stable.baseline_v1 | jq '.schema.fields'  | pbcopy
```

To run the relevant tests:

```
./script/update_example_baseline_clients_first_seen
pytest tests -k baseline_clients_first_seen 
```

I've been suffixing the pytest commands with `; say done`, which helped with the ~2 minute test cycles. 

There are a total of 4 tests. There's one set for `org_mozilla_fenix` which is the nightly version of the Firefox for Android product, which contains a migration ping. There's another set for `org_mozilla_ios_firefox`, which is a typical Glean application that doesn't contain a migration ping (but could in the future). There's a test for init, which involves reading from all history. This is going to be a somewhat expensive up-front query -- for `org_mozilla_firefox`, it'll be ~1.3 TB. There's also a test for the query, which should cover most of the obvious cases.

The `telemetry_derived.core_clients_first_seen` table will need to be scheduled when adding this to airflow, which will mean rerunning the query since it hasn't been updated in several months.

bigquery_etl/glean_usage/baseline_clients_first_seen.py
```diff
@@ -0,0 +1,183 @@
...
+    project_id, baseline_table, date, dry_run, output_dir=None, output_only=False
+):
+    """Process a single table, potentially also writing out the generated queries."""
+    tables = table_names_from_baseline(baseline_table)
+
+    table_id = tables["first_seen_table"]
+    view_id = tables["first_seen_view"]
+    render_kwargs = dict(
+        header="-- Generated via bigquery_etl.glean_usage\n",
+        fennec_id=any(
```
jklukas, line 118: What's the status of Firefox for iOS in terms of having a ping that contains the legacy identifier? Do we know what that will look like and whether we'll be able to adapt this approach to handle iOS?

bigquery_etl/glean_usage/templates/baseline_clients_first_seen.metadata.yaml
```diff
@@ -0,0 +1,12 @@
+---
+# Generated via bigquery_etl.glean_usage
+
+friendly_name: Baseline Clients First Seen
+description: |-
+  Captures the earliest server timestamp that we observe a particular client.
+
+  Clustering fields: `normalized_channel`, `sample_id`
```
jklukas, line 8:
```suggestion
  Clustering fields: `sample_id`
```

bigquery_etl/glean_usage/templates/baseline_clients_first_seen.metadata.yaml
```diff
@@ -0,0 +1,12 @@
+---
+# Generated via bigquery_etl.glean_usage
+
+friendly_name: Baseline Clients First Seen
+description: |-
+  Captures the earliest server timestamp that we observe a particular client.
```
jklukas, line 6:
```suggestion
  Captures the earliest server date that we observe a particular client.
```

bigquery_etl/glean_usage/templates/baseline_clients_first_seen_v1.init.sql
```diff
@@ -0,0 +1,39 @@
...
+AS
+WITH
+  baseline AS (
+    SELECT
+      client_info.client_id,
+      DATE(MIN(submission_timestamp)) as first_seen_date,
+      MIN(sample_id) as sample_id
```
jklukas, line 18: I prefer to GROUP BY `sample_id, client_id` so that we can reference `sample_id` directly without an arbitrary aggregation function.

bigquery_etl/glean_usage/templates/baseline_clients_first_seen_v1.init.sql
```diff
@@ -0,0 +1,39 @@
...
+      DATE(submission_timestamp) > "2010-01-01"
+    GROUP BY
+      client_id
+  )
+{% if fennec_id %}
+  ,
+  {{ core_clients_first_seen(migration_table) }}
+  SELECT
+    client_id,
+    COALESCE(core.first_seen_date, baseline.first_seen_date) as first_seen_date,
```
jklukas, line 32: It may be worth maintaining two additional dates here: `first_seen_baseline_date` and `first_seen_legacy_date` (NULL for apps without a legacy equivalent) would record the first observation by ping type. This would primarily be useful for debugging/validation.

jklukas (commented):
Leaving a round of comments. Will circle back after looking at associated PRs.

bigquery_etl/glean_usage/baseline_clients_first_seen.py
acmiyaguchi: I believe there hasn't been movement on this for quite some time (https://jira.mozilla.com/browse/FXIOS-1415), but this approach should be fine in the future with some minor modifications.

bigquery_etl/glean_usage/templates/baseline_clients_first_seen_v1.init.sql
acmiyaguchi: This deviates from the specification quite a bit. A `submission_date` field may work just as well, where the `submission_date` is the `first_seen_baseline_date` and the `first_seen_legacy_date` is effectively the `first_seen_date`. Debugging for legacy data would be all rows where the `submission_date <> first_seen_date`.

jklukas: > A submission_date field may work just as well, where the submission_date is the first_seen_baseline_date and the first_seen_legacy_date is effectively the first_seen_date.

That sounds fine. I don't want to lose momentum here by trying to reach perfection in naming. It may be better to avoid this issue altogether for now and keep just `first_seen_date` for simplicity. I don't feel strongly about it and will support any of the formulations that have been discussed here.

jklukas (approved):
I trust your judgment on whether to add additional date fields here. Overall structure looks great, and thank you for going the extra mile here with tests, which I did non provide for the initial implementation of glean_usage.

bigquery_etl/glean_usage/templates/baseline_clients_first_seen_v1.init.sql
acmiyaguchi: One side-effect of adding another column is that it would increase the amount of data scanned every day, since there would be another date field (mostly the same in fenix, exactly the same for non-migration based tables).

PR #1935, opened 2021-04-01
Move several dags from 02:00 to 02:15

jklukas:
Follow-up to https://github.com/mozilla/bigquery-etl/pull/1919

As of today, most of the alert emails we receive for rescheduling are for
DAGs that start at 02:00.

We can see from the
[copy_deduplicate landing times](https://workflow.telemetry.mozilla.org/landing_times?base_date=2021-03-30+01%3A00%3A00%2B00%3A00&num_runs=50&root=&dag_id=copy_deduplicate&_csrf_token=IjYwMWZjYWRiMGIyODhjYjhmYTUzYTU3NDJiOTg5NWIyMWJiYmRhMzci.YGYUHQ.XJAJsqcSenJ1aQN6Ia1ZzAmWTbw)
that the DAG rarely has taken longer than 75 minutes, so this change should
prevent most of the remaining alerts from being triggered.


dags.yaml
```diff
@@ -188,7 +191,7 @@ bqetl_devtools:
     retry_delay: 30m
 
 bqetl_main_summary:
-  schedule_interval: 0 2 * * *
+  schedule_interval: 15 2 * * *
```
scholtzan, line 194: This will require some changes here: https://github.com/mozilla/telemetry-airflow/blob/fd54979f705b7d402b910407b79e099162b36f91/dags/attitudes_daily.py#L60

scholtzan: And here https://github.com/mozilla/telemetry-airflow/blob/fd54979f705b7d402b910407b79e099162b36f91/dags/firefox_public_data_report.py#L92

dags.yaml
```diff
@@ -57,7 +60,7 @@ bqetl_vrbrowser:
     retry_delay: 5m
 
 bqetl_core:
-  schedule_interval: 0 2 * * *
+  schedule_interval: 15 2 * * *
```
scholtzan, line 63: Requires changes in https://github.com/mozilla/telemetry-airflow/blob/fd54979f705b7d402b910407b79e099162b36f91/dags/incline_dash.py#L37

dags.yaml
scholtzan: And a couple of changes in the `jetstream` DAG: https://github.com/mozilla/telemetry-airflow/blob/fd54979f705b7d402b910407b79e099162b36f91/dags/jetstream.py#L58

scholtzan: And in the `kpi_forecasts` DAG https://github.com/mozilla/telemetry-airflow/blob/fd54979f705b7d402b910407b79e099162b36f91/dags/kpi_forecasts.py#L64

scholtzan (commented):
Some of these changes will affect DAGs in telemetry-airflow that depend on some tasks generated by bigquery-etl that will change their schedule interval here. These telemetry-airflow tasks need to be updated and then merged together with the changes here.

jklukas:
I'm hoping this is no longer necessary, as it looks like we've gotten copy_deduplicate into shape to be able to complete before 02:00 via https://github.com/mozilla/telemetry-airflow/pull/1279 and https://github.com/mozilla/telemetry-airflow/pull/1280

PR #1937, opened 2021-04-06
Pull out main logic and generate example queries for glean usage

acmiyaguchi:
This does some incremental changes on the glean usage queries:

* Moves the main logic and argument parsing into common
* Generates example queries for each query

I generate queries for fenix and ios, but it may only be useful to keep the query for one of them if clients daily and clients last seen do not change across different products. I may remove the ios queries if they are redundant. 

I'm still a little unclear on what the best way to refactor `run_query` into a common utility, so I'll leave it for later. I want to add some tests (at least for clients_daily) before I merge in new changes to pull in the clients first seen stuff.


script/update_example_glean_usage
```diff
@@ -0,0 +1,41 @@
...
+
+scripts=(
+    "run_glean_baseline_clients_daily"
+    "run_glean_baseline_clients_first_seen"
+    "run_glean_baseline_clients_last_seen"
+)
+
+# kill all children in the case when we kill the parent script
+# https://stackoverflow.com/a/360275
+trap "jobs -p | xargs -r kill" EXIT
```
jklukas, line 17: Is this slow enough that running these concurrently is necessary?

jklukas:
I'm a little concerned about it being possible for tests to get out of sync with the actual `glean_usage` code since they rely on the queries being generated and checked in. Perhaps it's fine as long as we are able to add commentary to tests to let developers know they should run the generation machinery.

jklukas (approved):

script/update_example_glean_usage
acmiyaguchi: It took about 10 seconds at some point before I parallelized it even further, so it might be a bit overkill. It's also a handy snippet, so it might be worth keeping around .

acmiyaguchi:
> I'm a little concerned about it being possible for tests to get out of sync with the actual glean_usage code since they rely on the queries being generated and checked in. 

I'll add some CI that uses the `update_example_glean_usage` script to ensure that the files don't change.

PR #1941, opened 2021-04-07
Add baseline clients daily test

acmiyaguchi:
This is preparing for a followup to #1934 for modifications to the clients daily tables. There are a couple of changes here:

* tables references are stripped of the project name because the testing framework does not replace the project name correctly (this may just be a bug, but it doesn't change the high level query since the default project is specified by the script).
* rewrites `require_partition_filter` in tests because it uses a select * to get query results
* adds simple tests for clients daily -- this doesn't include any of the boolean logic


sql/moz-fx-data-shared-prod/org_mozilla_fenix/baseline_clients_daily/view.sql
```diff
@@ -1,8 +1,8 @@
 -- Generated via bigquery_etl.glean_usage
 CREATE OR REPLACE VIEW
-  `moz-fx-data-shared-prod.org_mozilla_fenix.baseline_clients_daily`
+  `org_mozilla_fenix.baseline_clients_daily`
```
jklukas, line 3: This violates our policy of having views include only fully-qualified references (which used to be required by BQ). I'm not sure if there's any good way around this while still enabling the tests here.

bigquery_etl/glean_usage/baseline_clients_daily.py
```diff
@@ -27,7 +27,7 @@ def run_query(
     project_id, baseline_table, date, dry_run, output_dir=None, output_only=False
 ):
     """Process a single table, potentially also writing out the generated queries."""
-    tables = table_names_from_baseline(baseline_table)
+    tables = table_names_from_baseline(baseline_table, include_project_id=False)
```
jklukas, line 30: Maybe we should have `table_names_from_baseline` include additional entries in the dictionary. Something like "daily_table_qualified" that would be the version with project ID. Then the view template could reference the fully qualified name while the query template could reference the "daily_table" unqualified name, etc.

sql/moz-fx-data-shared-prod/org_mozilla_fenix/baseline_clients_daily/view.sql
acmiyaguchi: I see, it makes sense in the context of publishing views to `mozdata` and the like too.

bigquery_etl/glean_usage/baseline_clients_daily.py
acmiyaguchi: I might go with adding `project_id` as a variable that can be pulled in.

sql/moz-fx-data-shared-prod/org_mozilla_fenix/baseline_clients_daily/view.sql
acmiyaguchi: I'm not sure if you meant the references in the view or the view itself e.g. `moz-fx-data-shared-prod.org_mozilla_fenix_derived.baseline_clients_daily_v1` vs `moz-fx-data-shared-prod.org_mozilla_fenix.baseline_clients_daily`. I'm assuming the former and that the latter can be implicit based on the default project of the bq client.

jklukas: In practice, we have used a fully-qualified name for the view we are creating as well, and we may have an implicit dependency on this in our view publishing script or in Jenkins logic (which replaces shared-prod with shared-nonprod for publishing to the nonprod project. We should make sure this is fully qualified here for consistency.

jklukas (approved):
r+wc

Looks great. All I ask is that we make sure the CREATE VIEW statement targets a fully-qualified view name.

jklukas:
Whoops. Didn't realize auto-merge had been enabled. I will file a quick follow-up.

PR #1943, opened 2021-04-08
Use fully-qualified targets in glean_usage view creation

jklukas:
Follow-up to https://github.com/mozilla/bigquery-etl/pull/1941


acmiyaguchi (approved):
Apologies on the auto-merge, I should have turned it off with questions still in the air.

PR #1946, opened 2021-04-08
[DS-1424] Join baseline clients daily with first seen table

acmiyaguchi:
Yet another followup for [DS-1424](https://jira.mozilla.com/browse/DS-1424).

This joins the clients daily table with the first seen table. It also updates the view for baseline clients first seen so it refers to the clients daily table. I've made notes on the difference in semantics between `clients_first_seen` and `is_new_profile` which is caused by former using migration pings.


jklukas (changes requested):
I want to make sure we understand how we want to treat migrated clients before we proceed with this.

bigquery_etl/glean_usage/templates/baseline_clients_first_seen.metadata.yaml
```diff
@@ -3,7 +3,11 @@
 
 friendly_name: Baseline Clients First Seen
 description: |-
-  Captures the earliest server date that we observe a particular client.
+  Captures the earliest server date that we observe a particular client in the
+  baseline table. A client may have a `first_seen_date` that is earlier than the
+  epoch of the baseline table; this is the result of a migration from a legacy
+  data collection system. The `first_seen_date` and the `is_new_profile` fields
+  therefore have slightly different semantics.
```
jklukas, line 10: We should check with @gkabbz about terminology here. My understanding is that there really is no concept of a "profile" for mobile apps, although we tend to talk about "new profiles" for symmetry with desktop. is_new_profile is probably the most clear choice, even though one could argue it's technically incorrect.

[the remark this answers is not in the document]
gkabbz: Correct. I think we should define a new profile as of the date they first showed up in our systems regardless of the products. e.g. If user x was a fennec user and their first_seen_date on fennec was Mar 1, 2020 but they migrated to fenix with a first seen date of Sep 10, 2020. Then their first_seen_date (even with the fenix client_id) should be Mar 1, 2020 so we can maintain their tenure between products.

gkabbz: I don't have strong feelings about referring about it as new profiles. If it helps with symmetry with desktop I vote for it. Helps reduce the learning curve when switching between products. In desktop a profile = client_id which would hold here too right?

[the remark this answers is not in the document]
acmiyaguchi: If this is the case, rows that appear before the baseline ping epoch will not appear in this first seen table. I assume that this is fine though.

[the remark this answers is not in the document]
jklukas: > rows that appear before the baseline ping epoch will not appear in this first seen table

That seems fine to me, and worth documenting as part of the view's description.

acmiyaguchi: I'm also using the terminology from the [cross ping mobile ETL](https://docs.google.com/document/d/1aayc2G0qFJP7EOSvELgUsYklpyixeZChVMYtHlbCpw8/edit#heading=h.9zc3l8c1meuo) which uses `is_new_profile`.

jklukas: > In desktop a profile = client_id which would hold here too right

Correct.

Let's keep `is_new_profile` as the name.

bigquery_etl/glean_usage/templates/baseline_clients_daily_v1.sql
```diff
@@ -130,11 +130,25 @@ windowed AS (
...
+    _n = 1
 )
 --
-SELECT
-  * EXCEPT (_n)
-FROM
-  windowed
-WHERE
-  _n = 1
+select
```
jklukas, line 151:
```suggestion
SELECT
```

bigquery_etl/glean_usage/templates/baseline_clients_daily_v1.sql
```diff
@@ -130,11 +130,25 @@ windowed AS (
...
+  *
+from
```
jklukas, line 153:
```suggestion
FROM
```

bigquery_etl/glean_usage/templates/baseline_clients_daily_v1.sql
jklukas: If you can run script/format_sql on the output of this, it should catch any remaining style inconsistencies.

jklukas (approved):

bigquery_etl/glean_usage/templates/baseline_clients_daily_v1.sql
acmiyaguchi: I've explicitly ignored these files in the format_sql script, because the the double braces of the jinja templating tends to break.

acmiyaguchi:
Migration of the tables has been done via #1947, dry-runs are now successful.

PR #1952, opened 2021-04-12
Ignore materialized view dry runs

scholtzan:
The new BigQuery release introduced some change which disallows using `TO_JSON_STRING` with `STRUCT` input in materialized view: "Materialized views may not use the TO_JSON_STRING function with a STRUCT input". Unfortunately, our current strategy depends on using `TO_JSON_STRING` with `STRUCT` inputs. Materialized view that have been deployed in the past are still working, for now.

To reduce CI failure noise, this will skip dry runs for materialized views


jklukas (approved):
r+

If you have not already, please do file a GCP support case about this. I think it would warrant a P3 investigation since we're already relying on this functionality and we're currently unable to make changes.

scholtzan:
Will do

Issue #1927, opened 2021-03-30
public json export has frequent DELETE failures


jklukas:
Several of the public json export tasks have been needing to retry in recent days. It appears they're trying to delete objects that don't exist or have already been cleaned up.

See an example in https://workflow.telemetry.mozilla.org/log?execution_date=2021-03-29T05%3A00%3A00%2B00%3A00&task_id=export_public_data_json_mozregression_aggregates__v1&dag_id=bqetl_public_data_json

That task only succeeded on the third attempt.

cc @scholtzan

jklukas:
Here's another example: https://workflow.telemetry.mozilla.org/log?execution_date=2021-04-10T05%3A00%3A00%2B00%3A00&task_id=export_public_data_json_telemetry_derived__ssl_ratios__v1&dag_id=bqetl_public_data_json

From the logs, it looks like the problem is that we "rename_blob" 000000000000.json.tmp.gz to 000000000000.json.gz, etc. and then we immediately list the contents of the staging directory in order to delete them. From GCS's docs, I would expect the list operation after the delete would be consistent, but it appears that we end up with `000000000000.json.tmp.gz` in the listing and we attempt to delete it again after doing the rename.

Perhaps the simplest way around this would be to change "rename_blob" to "copy_blob", relying on the final `_clear_stage_directory` step to do all cleanup rather than issuing deletes as we move blobs into place.

PR #1955, opened 2021-04-13
Public data export temporary stage directory

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/1927

### DIFF
--- a/bigquery_etl/public_data/publish_json.py
+++ b/bigquery_etl/public_data/publish_json.py
@@ -55,7 +55,12 @@ class JsonPublisher:
         self.storage_client = storage_client
         self.temp_table = None
         self.date = None
-        self.stage_gcs_path = self.gcs_path + "stage/json/"
+        self.stage_gcs_path = (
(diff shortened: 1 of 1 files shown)


bigquery_etl/public_data/publish_json.py
```diff
@@ -55,7 +55,9 @@ def __init__(
         self.storage_client = storage_client
         self.temp_table = None
         self.date = None
-        self.stage_gcs_path = self.gcs_path + "stage/json/"
+        self.stage_gcs_path = self.gcs_path + "stage/json/" + "".join(
+            random.choice(string.ascii_lowercase) for i in range(12)
+        )
```
scholtzan, line 60: With multiple export tasks running in parallel and writing to the same stage directory, once one job is finished it will also delete staged files from all other jobs that are still running. Once they are complete and try to delete their staged files (which are already deleted) an error occurs

jklukas (approved):

PR #1956, opened 2021-04-13
Expose an `interval` field on iap subscriptions

relud:
also expose subscription id and device id for vpn tables, and product id for stripe subscriptions

`interval` maps to `Term` in [Data requirements for Finance](https://docs.google.com/document/d/1na1FlF8frDKF0_8kbc2hUwZXP9ljegYVAVVd1GPIut8/edit), uses naming consistent with Stripe, and correctly returns `"month"` for all the subs we have in prod.

the id fields are all for Looker, where it wants them to ensure it doesn't double count when one_to_many tables are joined in.

### DIFF
--- a/sql/mozfun/iap/derive_apple_subscription_interval/udf.sql
+++ b/sql/mozfun/iap/derive_apple_subscription_interval/udf.sql
@@ -1,0 +1,117 @@
+CREATE OR REPLACE FUNCTION iap.derive_apple_subscription_interval(
+  purchase_datetime_pst DATETIME,
+  expires_datetime_pst DATETIME
+) AS (
+  CASE
+  WHEN
+    DATETIME_ADD(purchase_datetime_pst, INTERVAL 1 YEAR) = expires_datetime_pst
+  THEN
+    "year"
+  WHEN
+    DATETIME_ADD(purchase_datetime_pst, INTERVAL 1 QUARTER) = expires_datetime_pst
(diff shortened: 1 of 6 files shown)


sql/mozfun/iap/derive_apple_subscription_interval/udf.sql
```diff
@@ -0,0 +1,117 @@
...
+  WHEN
+    DATETIME_ADD(purchase_datetime_pst, INTERVAL 1 WEEK) = expires_datetime_pst
+  THEN
+    "week"
+  WHEN
+    DATETIME_ADD(purchase_datetime_pst, INTERVAL 1 DAY) = expires_datetime_pst
+  THEN
+    "day"
+  ELSE
+    CONCAT(CAST(DATETIME_DIFF(expires_datetime_pst, purchase_datetime_pst, DAY) AS STRING), " day")
```
acmiyaguchi, line 27: Why is there a special case for having a number of days?

acmiyaguchi (approved):
This looks reasonable to me.

sql/mozfun/iap/derive_apple_subscription_interval/udf.sql
relud: because it's a fallback, and I didn't want it to throw and error or lose information if we add something weird in the future that this doesn't recognize

PR #1959, opened 2021-04-14
Fix typo in stripe invoice schema

relud:


bigquery_etl/stripe/event.schema.json
```diff
@@ -5068,7 +5068,7 @@
                 "type": "TIMESTAMP"
               },
               {
-                "name": "marked_uncollectable_at",
+                "name": "marked_uncollectible_at",
```
jklukas, line 5071: Does this field already exist? I expect we need to maintain the old name unless you're planning to rewrite the table to drop the old column.

relud: i rewrote the tables to drop the incorrectly named column, there was no data in the field until today

jklukas (approved):

PR #1960, opened 2021-04-15
Add vpn tables to support data studio dashboard

relud:
Requested by @gkabbz in [VPN Queries That Need To Be Productionized for Dashboard](https://docs.google.com/document/d/1VMy-6kU1AAzT_XLqi8bsu2I2fnqrrVqhObPUCBdjfGg/edit)

The dashboard linked in the request has already been updated to use these tables.

`subscriptions_with_attribution` was updated to include IAP subscriptions and named `all_subscriptions` because `subscriptions` is taken. In the future I plan to remove both `subscriptions` and `all_subscriptions` and instead use `subscription_platform.subscriptions`.

`site_metrics_summary` was originally two underlying tables unioned in a view, but the logic was very similar so I combined them into a single table.

Some fields have been renamed:
- stripe table ids are consistently suffixed with `_id`
- `rationalized_*` fields were renamed to `normalized_*` because I moved their logic to a UDF in mozfun.norm
- `website_channel_groups` was renamed to `website_channel_group` because it's a singular string value


dags.yaml
```diff
@@ -117,6 +117,21 @@ bqetl_mozilla_vpn:
...
     retry_delay: 30m
 
+bqetl_mozilla_vpn_site_metrics:
+  schedule_interval: 0 23 * * *
+  description: |
+    Daily extracts from the Google Analytics tables for Mozilla VPN as well as
+    derived tables based on that data.
+
+    Depends on Google Analytics exports, which have highly variable timing, so
+    is scheduled to run very late in the UTC day.
```
relud, line 127: over the 155 days since the GA exports were enabled, these are the UTC hours when the export for the full day became available:

| export_hour | days |
|-------------|------|
|    23:00:00 |    1 |
|    21:00:00 |   32 |
|    18:00:00 |   13 |
|    17:00:00 |   36 |
|    16:00:00 |   40 |
|    15:00:00 |   28 |
|    14:00:00 |    5 |

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/all_subscriptions_v1/query.sql
```diff
@@ -0,0 +1,208 @@
...
+    (customer_id)
+  LEFT JOIN
+    standardized_country
+  USING
+    (country)
+  LEFT JOIN
+    attribution
+  USING
+    (fxa_uid)
+  UNION ALL
```
acmiyaguchi, line 143: These two blocks could be split into their own cte statements e.g. `select * from fxa_subscriptions union all select * from apple_iap_subscriptions`.

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/funnel_ga_to_subscriptions_v1/query.sql
```diff
@@ -0,0 +1,163 @@
...
+      source
+    ).*,
+    sessions,
+    subscribe_intent_goal,
+    country,
+  FROM
+    site_metrics_summary_v1
+  WHERE
+    -- Populate from mozilla.org starting 2021-03-11, and from vpn.mozilla.org before that
+    IF(`date` < DATE "2021-03-11", site = "vpn.mozilla.org", site = "mozilla.org")
```
acmiyaguchi, line 19: TIL you can use if/case statements in the where clause

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/funnel_ga_to_subscriptions_v1/query.sql
```diff
@@ -0,0 +1,163 @@
...
+    normalized_content,
+    IF(
+      COUNT(DISTINCT COALESCE(website_channel_group, "NULL")) = 1,
+      ANY_VALUE(website_channel_group),
+      ERROR("website_channel_group must have one value per group")
+    ) AS website_channel_group,
+    SUM(sessions) AS sessions,
+    SUM(
+      IF(
+        country IN (
```
acmiyaguchi, line 41: Maybe stick this in a temp udf at the top so it only needs to be changed once when DE and FR are added.

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/site_metrics_summary_v1/query.sql
```diff
@@ -0,0 +1,189 @@
...
+      site = "vpn.mozilla.org",
+      STRUCT(
+        hits.eventInfo.eventLabel LIKE "%Subscribe%" AS subscribe_intent,
+        hits.eventInfo.eventLabel LIKE "%Join%" AS join_waitlist_intent,
+        hits.eventInfo.eventLabel LIKE "%Submit waitlist form%" AS join_waitlist_success,
+        hits.eventInfo.eventLabel LIKE "%Sign in%" AS sign_in_intent,
+        hits.eventInfo.eventLabel LIKE "%Download%" AS download_intent,
+        hits.eventInfo.eventLabel LIKE "%Download installer%" AS download_installer_intent
+      ),
+      STRUCT(
```
acmiyaguchi, line 43: These two structs don't match up? It looks like the latter one is missing `subscribe_intent`. Will it just get added as null in the final result because of the `.*`?

acmiyaguchi (approved):
Looks good to me. I'd recommend dumping the schemas for the tables if possible (for future review) alongside the tables, even if there isn't CI to check changes in schemas.

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/site_metrics_summary_v1/query.sql
relud: `subscribe_intent` was accidentally named `join_waitlist_intent` in the second struct. It didn't break because bigquery doesn't use names from the second struct, only ordering and types. will fix.

PR #1965, opened 2021-04-20
Remove allow_field_addition_on_date from clients_daily

scholtzan:
`allow_field_addition_on_date` in clients_daily changes the order of fields in the destination table schema based on how they appear in the query. This causes the schema of `clients_first_seen` and `clients_last_seen` to be different from `clients_daily` after adding new fields.

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/metadata.yaml
@@ -15,7 +15,6 @@ labels:
   application: firefox
   schedule: daily
 scheduling:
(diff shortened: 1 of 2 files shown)


sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/metadata.yaml
```diff
@@ -15,7 +15,6 @@ labels:
   application: firefox
   schedule: daily
 scheduling:
-  allow_field_addition_on_date: '2021-04-19'
```
scholtzan, line 18: We should remove this, just in case we ever run a backfill. Then the schema will change again.

jklukas (approved):

PR #1969, opened 2021-04-20
Update owners

jklukas:
Removes some folks who no longer are active on the assigned projects.

Also see https://github.com/mozilla/telemetry-airflow/pull/1288


dags.yaml
```diff
@@ -361,10 +361,10 @@ bqetl_search_dashboard:
     depends_on_past: false
     email:
       - telemetry-alerts@mozilla.com
-      - ssuh@mozilla.com
+      - bewu@mozilla.com
```
jklukas, line 364: @BenWu Do you have context on this search dashboard DAG such that you'd be an appropriate owner? I can assign myself if we don't know who else it would make sense for.

BenWu: I have some context.  I can take this

BenWu (approved):

PR #1970, opened 2021-04-20
[Bug 1706394] Internet outages dataset query from telemetry_stable.main_v4

scholtzan:
https://bugzilla.mozilla.org/show_bug.cgi?id=1706394

Directly querying from `telemetry_stable.main_v4` fixes the `Resources exceeded during query execution` error.

cc @Dexterp37

### DIFF
--- a/sql/moz-fx-data-shared-prod/internet_outages/global_outages_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/internet_outages/global_outages_v1/query.sql
@@ -229,7 +229,7 @@ histogram_data_sample AS (
       payload.processes.content.histograms.http_page_tls_handshake
     ).values AS tls_handshake,
   FROM
-    telemetry.main
+    telemetry_stable.main_v4
   WHERE
     DATE(submission_timestamp) = @submission_date
     -- Restrict to Firefox.


jklukas (commented):
The one field that will change with this is [subsession_length, which has normalization applied](https://github.com/mozilla/bigquery-etl/blob/8109d147e0cf68645ae4d35c5639a592de027a7b/sql/moz-fx-data-shared-prod/udf/normalize_main_payload/udf.sql#L14-L22).

We occasionally see very high values for subsession_length, which are replaced with NULL in the view. So there will be some pings that pass the `> 0` check on the `main_v4` table that would have failed on top of the `telemetry.main` view. Because we aren't summing or averaging the raw values, these outliers probably aren't a big deal, but we may want to check that. Or at least be prepared to examine whether we see any difference in behavior for this dataset.

scholtzan:
Isn't it just `session_length` that has normalization applied?

jklukas (approved):

jklukas:
> Isn't it just session_length that has normalization applied? subsession_length should be unaffected by the normalization

Indeed it is. Apologies for the misdirection.

Dexterp37:
Thanks for fixing this folks.

PR #1974, opened 2021-04-22
Bug 1673976 - Add glean ios search probes to mobile search tables

BenWu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1673976

@mmccorks I made temporary tables with a sample of ios data (both core and glean) from 2021-03-01 to 2021-04-10 at `mozdata.tmp.ios_search_aggregates_210422` and  `mozdata.tmp.ios_search_clients_daily_210422`.  Can you do some validation of this to make sure it looks reasonable?  I did a trivial check here https://sql.telemetry.mozilla.org/queries/79590/source#197718

Changes:
  - add ios glean to search tables (sap and in-content, no ad probes)
  - moves template code from sql/ to bigquery-etl
  -  mobile_search_clients_daily now grouped by channel since ios has same app name for all channels, shouldn't make a difference for everything else


akkomar (approved):
LGTM

BenWu:
Added a clause to hide organic counts from ios.  @akkomar can you review the last commit?

akkomar (approved):

PR #1975, opened 2021-04-26
Bug 1707571 - Ignore isverified in fxa_auth_events

jklukas:
This is hopefully a temporary workaround while we understand the incompatible schema change in the source table.

Very similar to the fix in https://github.com/mozilla/bigquery-etl/pull/1683

### DIFF
--- a/dags.yaml
+++ b/dags.yaml
@@ -95,6 +95,11 @@ bqetl_fxa_events:
 
     The DAG also provides daily aggregations on top of the raw log data,
     which eventually power high-level reporting about FxA usage.
+
+    Tasks here have occasionally failed due to incompatible schema changes
+    in the tables populated by Cloud Logging.
+    See https://github.com/mozilla/bigquery-etl/issues/1684 for an example
+    mitigation.
   default_args:
     owner: jklukas@mozilla.com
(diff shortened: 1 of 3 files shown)


wlach (approved):
Makes sense, thanks for looking into this!

jbuck:
Ack, sorry, I broke the ETL job here.

I was getting errors from Cloud Logging such as:

```
Field jsonpayload.fields.emailverified: Cannot convert bool value to floating point:1
Field jsonpayload.fields.isprimary: Cannot convert bool value to floating point:1
Field jsonpayload.fields.isverified: Cannot convert bool value to floating point:0
```

To fix these errors, I changed the schema of these three fields to `BOOLEAN` on 2011-04-25

This fixed the problem from Cloud Logging's perspective, but of course broke any downstream consumer

jklukas:
It looks like we're unstuck now with the additional fix in #1976, but these fields will be null. I'll need to think a bit more about whether we need a longer-term fix to make these fields available.

jklukas:
I looks like these fields are never populated for `jsonPayload.type = 'amplitudeEvent'` so I don't think we need to worry about backfilling or resolving this schema issue.

PR #1976, opened 2021-04-26
Bug 1707571 Ignore additional fields in fxa_auth_events

jklukas:
Fixup to #1975


jklukas:
Note that this particularly query is exempt from dryrun due to the source table having locked-down permissions.

jklukas:
This may turn out to be insufficient based on https://github.com/mozilla/bigquery-etl/pull/1975#issuecomment-826918800

I will do some more validation in the BQ console (since I do have special privs to be able to access this table).

jklukas:
I've verified that this query will dry run now in the BQ console using my own credentials.

wlach (approved):

PR #1978, opened 2021-04-26
Bug 1707640 - Calculate sample_id in baseline_clients_first_seen

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1707640#c8

### DIFF
--- a/bigquery_etl/glean_usage/templates/baseline_clients_first_seen_v1.init.sql
+++ b/bigquery_etl/glean_usage/templates/baseline_clients_first_seen_v1.init.sql
@@ -15,7 +15,6 @@ WITH
   baseline AS (
     SELECT
       client_info.client_id,
-      sample_id,
       -- Some Glean data from 2019 contains incorrect sample_id, so we
       -- recalculate here; see bug 1707640
       udf.safe_sample_id(client_info.client_id) AS sample_id,
--- a/bigquery_etl/glean_usage/templates/baseline_clients_first_seen_v1.init.sql
+++ b/bigquery_etl/glean_usage/templates/baseline_clients_first_seen_v1.init.sql
@@ -16,6 +16,9 @@ WITH
     SELECT
       client_info.client_id,
       sample_id,
+      -- Some Glean data from 2019 contains incorrect sample_id, so we
+      -- recalculate here; see bug 1707640
+      udf.safe_sample_id(client_info.client_id) AS sample_id,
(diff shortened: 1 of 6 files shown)


acmiyaguchi (approved):
This should be fine wrt to the incremental query because we end up taking the sample_id along the a row joined by the client id. It might be a good idea to recalculate sample ids there too as a precaution, but it's not strictly necessary since other tables that reference sample ids in the baseline table would be affected too.

bigquery_etl/glean_usage/templates/baseline_clients_first_seen_v1.query.sql
```diff
@@ -7,8 +7,10 @@ WITH
 _baseline AS (
   -- extract the client_id into the top level for the `USING` clause
   SELECT DISTINCT
-    sample_id,
-    client_info.client_id
+    client_info.client_id,
+    -- Some Glean data from 2019 contains incorrect sample_id, so we
+    -- recalculate here; see bug 1707640
+    udf.safe_sample_id(client_info.client_id) AS sample_id,
```
acmiyaguchi, line 13: Oh nevermind my main review comment, looks like it's being recalculated here.

jklukas: I think I slipped in the query.sql changes while you were reviewing, so this very well may not have been there when you first read through.

bigquery_etl/glean_usage/templates/baseline_clients_first_seen_v1.init.sql
```diff
@@ -16,6 +16,9 @@ WITH
     SELECT
       client_info.client_id,
       sample_id,
+      -- Some Glean data from 2019 contains incorrect sample_id, so we
+      -- recalculate here; see bug 1707640
+      udf.safe_sample_id(client_info.client_id) AS sample_id,
```
acmiyaguchi, line 21: The tests will start to fail because they use a dummy value for the sample id.

jklukas: I'm going to make the `expect.yaml` match the actual output for now. It may be good to come back to this and fix up the tests to use full UUID values.

acmiyaguchi: That, or a single test row with a legitimate uuid value.

PR #1986, opened 2021-04-29
Bug 1706166 - Add allow list for stripe fields

relud:
enforce the schema on disallowed fields, so that unknown fields added to the schema are still detected

fixes [Bug 1706166](https://bugzilla.mozilla.org/show_bug.cgi?id=1706166)


bigquery_etl/stripe/allowed_fields.py
```diff
@@ -0,0 +1,188 @@
...
+        self.root = bigquery.SchemaField.from_api_repr(
+            {"name": "root", "type": "RECORD", "fields": ujson.loads(path.read_text())}
+        )
+        self.filtered = list(self._filter_schema(self.root.fields, self.allowed))
+
+    def _filter_schema(
+        self, fields: List[bigquery.SchemaField], allowed: dict
+    ) -> Iterable[bigquery.SchemaField]:
+        for field in fields:
+            if field.name in allowed:
```
acmiyaguchi, line 55: Inverting conditions can reduce nesting e.g.

```python
if field.name not in allowed:
    continue
...
```

bigquery_etl/stripe/allowed_fields.py
```diff
@@ -0,0 +1,188 @@
...
+        is_list_item: bool = False,
+    ) -> Any:
+        # override obj for metadata and paged lists
+        if (
+            field.name in ("metadata", "custom_fields")
+            and isinstance(obj, dict)
+            and not is_list_item
+        ):
+            if "userid" in obj:
+                # hash fxa uid before it reaches BigQuery
```
acmiyaguchi, line 91: Why?

acmiyaguchi (approved):
Excellent tests. 

What's the plan for migrating over to these new schemas? Drop the old tables and reimport from scratch using the new version of this script?

bigquery_etl/stripe/allowed_fields.py
relud: @jklukas told me we don't want unhashed fxa uid anywhere in `moz-fx-data-shared-prod`

relud:
> What's the plan for migrating over to these new schemas? Drop the old tables and reimport from scratch using the new version of this script?

I'm going to export the tables as ndjson, load into temp tables with the new schema and `--ignore_unknown_values`, then select the temp tables into place. this preserves 7-day rollback capability, and should be reasonably fast because we only have ~4GB of events.

bigquery_etl/stripe/allowed_fields.py
jklukas: If the relevant tables here are Workgroup Confidential, then we could consider leaving this unhashed. But I'd indeed like to keep raw fxa userid out of shared-prod unless we have a clear business case.

relud: we don't, this is working fine

PR #1987, opened 2021-04-29
Bug 1708264 - Add initial support for dataset_metadata.yaml

jklukas:
This adds the machinery to create `dataset_metadata.yaml` as part of
`bqetl query create` and to validate that these files are valid, but these
are not yet actually used for anything.

The next step will be to fill out `dataset_metadata.yaml` for all
existing derived and user-facing datasets.

### DIFF
--- a/bigquery_etl/metadata/parse_metadata.py
+++ b/bigquery_etl/metadata/parse_metadata.py
@@ -13,6 +13,10 @@ from google.cloud import bigquery
 from bigquery_etl.query_scheduling.utils import is_email
 
 METADATA_FILE = "metadata.yaml"
+DATASET_METADATA_FILE = "dataset_metadata.yaml"
+DEFAULT_WORKGROUP_ACCESS = [
+    dict(role="roles/bigquery.dataViewer", members=["workgroup:mozilla-confidential"])
+]
 
 
 class Literal(str):
@@ -80,7 +84,7 @@ class BigQueryMetadata:
 @attr.s(auto_attribs=True)
 class Metadata:
     """
-    Representation of a Metadata configuration.
+    Representation of a table or view Metadata configuration.
 
     Uses attrs to simplify the class definition and provide validation.
     Docs: https://www.attrs.org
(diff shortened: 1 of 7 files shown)


sql/moz-fx-data-shared-prod/activity_stream/dataset_metadata.yaml
```diff
@@ -0,0 +1,11 @@
+friendly_name: Activity Stream
```
jklukas, line 1: We fill out the three activity stream datasets as an example here. The `_bi` dataset is a "custom" one with an additional ACL entry. The `_derived` directory didn't previously exist, but we add it here for completeness since the pipeline currently automatically creates a `_derived` dataset per document namespace.

jklukas:
Marking @scholtzan for review here as the author of most of the existing code for manipulating table-level `metadata.yaml` files and this code is largely trying to fit into that existing mold.

Marking @whd for review to validate the format and whether the `activity-stream` examples are matching your expectations.

bigquery_etl/metadata/parse_metadata.py
```diff
@@ -279,3 +283,70 @@ def set_bigquery_clustering(self, clustering_fields):
...
+    @classmethod
+    def from_file(cls, metadata_file):
+        """Parse dataset metadata from the provided file and create a new
+        DatasetMetadata instance."""
+        friendly_name = None
+        description = None
+        owners = []
+        labels = {}
+        scheduling = {}
+        bigquery = None
```
scholtzan, line 345:
```suggestion
```

whd (approved):
r+ with the activity_stream_bi change.

> The next step will be to fill out dataset_metadata.yaml for all existing derived and user-facing datasets.

The format looks good, and should be straightforward to translate back into `namespaces.tfvars.json` in the short term. There are some keys not used in the example but I'll look out for those in the PR following this one.

An issue I can think of if we're planning to translate all "user-facing" datasets to this format is dealing with datasets that are overridden on a per-environment or per-project basis. [env_namesaces.auto.tfvars.json](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-mozdata/tf/prod/envs/prod/bigquery/env_namespaces.auto.tfvars.json) (merged with and preferred over `namespaces.tfvars.json` for any particular project/environment deployment) is the current mechanism for dealing with cases such as:

1. `tmp` and `analysis`, granted access differently between mozdata and shared
2. `syndicate` datasets, where e.g. `glam_etl` is syndicated from [stage in stage](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-mozdata/tf/nonprod/envs/stage/bigquery/env_namespaces.auto.tfvars.json#L33-L47) and [prod in prod](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-mozdata/tf/prod/envs/prod/bigquery/env_namespaces.auto.tfvars.json#L17-L32).

If the plan in the first pass is to only deal with datasets that are relevant to this repository I don't think we need to figure this out.

jklukas:
> If the plan in the first pass is to only deal with datasets that are relevant to this repository I don't think we need to figure this out.

Indeed, I would love to avoid that complexity at the moment if it's reasonable on the ops logic side to maintain the concept of allowing `env_namesaces.auto.tfvars.json` to override content coming from `generated-sql`.

PR #1993, opened 2021-05-03
Make restricted access datasets view_restricted

whd:
This removes the default `projectReaders` access. We set this correctly on `subscription_platform` but not these two datasets.


whd:
I also noticed we're pretty inconsistent with ending descriptions with `.` or not. For the purposes of moving forward with implementation as long as they are the same between here and ops infra it doesn't matter.

jklukas (approved):

PR #2001, opened 2021-05-04
Bug 1708486 - Create aggregate table for contextual services

BenWu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1708486

### DIFF
--- a/sql/moz-fx-data-shared-prod/contextual_services_derived/event_aggregates_v1/init.sql
+++ b/sql/moz-fx-data-shared-prod/contextual_services_derived/event_aggregates_v1/init.sql
@@ -3,7 +3,8 @@ CREATE OR REPLACE TABLE
 PARTITION BY
   submission_date
 CLUSTER BY
-  source, event_type
+  source,
+  event_type
 OPTIONS
   (require_partition_filter = TRUE)
 AS
--- a/sql/moz-fx-data-shared-prod/contextual_services_derived/event_aggregates_v1/init.sql
+++ b/sql/moz-fx-data-shared-prod/contextual_services_derived/event_aggregates_v1/init.sql
@@ -1,0 +1,20 @@
+CREATE OR REPLACE TABLE
+  contextual_services_derived.event_aggregates_v1
+PARTITION BY
+  submission_date
+CLUSTER BY
+  source, event_type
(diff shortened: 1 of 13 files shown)


sql/moz-fx-data-shared-prod/contextual_services_derived/event_aggregates_v1/init.sql
```diff
@@ -0,0 +1,21 @@
+CREATE OR REPLACE TABLE
+  contextual_services_derived.event_aggregates_v1
+PARTITION BY
+  submission_date
+CLUSTER BY
+  source,
+  event_type
```
jklukas, line 7: Note: Now that we support clustering definitions under the `bigquery` key in `metadata.yaml` and `schema.yaml` files, that approach is preferred over defining an `init.sql`. See https://mozilla.github.io/bigquery-etl/cookbooks/common_workflows/#adding-a-new-scheduled-query

jklukas (approved):
r+wc

Switching to schema.yaml is optional here.

The query here is very clean. I would have done aggregations on each table individually, and then unioned at the end, but your solution here is more succinct and likely more maintainable.

sql/moz-fx-data-shared-prod/contextual_services_derived/event_aggregates_v1/init.sql
BenWu: Ok thanks I remembered there was something like that but I couldn't find the documentation so I thought I got it mixed up with something else.

BenWu:
@jklukas `schema.yaml` doesn't seem to have any use in this case because dryrun doesn't have permissions to validate the schema.  I also don't have permissions to create the table.  Can you create the table after merging using the `init.sql`?  Is it worth keeping a `schema.yaml` alongside the `init.sql` here?

jklukas:
> schema.yaml doesn't seem to have any use in this case because dryrun doesn't have permissions to validate the schema

Good point. Sorry for the misdirection, then.

>  Can you create the table after merging using the init.sql? 

I went ahead and ran the `init.sql`.

> Is it worth keeping a schema.yaml alongside the init.sql here?

No, let's just rely on the `init.sql`.

PR #2006, opened 2021-05-06
Update clients_last_seen_joined_v1 schema

scholtzan:
I already deployed the updated schema for `clients_last_seen_joined_v1` so that Airflow will finish successfully. The `clients_last_seen_joined_v1` had the fields added in https://github.com/mozilla/bigquery-etl/pull/2004 missing


jklukas (approved):

jklukas:
I added a commit here to include this table in the schema update workflow docs as well.

PR #2011, opened 2021-05-06
Automatically update downstream dependencies on query schema update

scholtzan:
This PR ensures that schemas of downstream dependencies get updated when updating a schema via `./bqetl query schema update ... --update-downstream`. If a schema cannot be updated automatically, which is the case for clients_last_seen and clients_first_seen, then a warning will be printed out.


README.md
```diff
@@ -353,7 +353,6 @@ Scheduling Queries in Airflow
     - `arguments` - a list of arguments passed when running the query, for example: `["--append_table"]`
     - `referenced_tables` - manually curated list of tables the query depends on; used to speed up the DAG generation process or to specify tables that the dry run doesn't have permissions to access, e. g. `[['telemetry_stable', 'main_v4']]`
     - `multipart` indicates whether a query is split over multiple files `part1.sql`, `part2.sql`, ...
-    - `allow_field_addition_on_date`: date for which new fields are allowed to be added to the existing destination table query results are written to
```
scholtzan, line 356: I propose that we drop support for `allow_field_addition_on_date` and rely on `schema.yaml` for adding fields from now on to prevent what happened in https://bugzilla.mozilla.org/show_bug.cgi?id=1706683

bigquery_etl/cli/query.py
```diff
@@ -759,7 +747,20 @@ def schema():
...
+    "--update_downstream",
+    help="Update downstream dependencies. GCP authentication required.",
+    default=False,
+    is_flag=True,
+)
+@click.option(
+    "--tmp-dataset",
+    "--tmp_dataset",
+    help="GCP datasets for creating updated tables temporarily.",
+    default="analysis",
```
scholtzan, line 761: For updating downstream dependencies, temporary tables with updated schemas will be created in GCP. I tried creating them in the `tmp` dataset, however the dryrun account doesn't have permissions to query this dataset. Not sure if there was a specific reason?

bigquery_etl/cli/query.py
```diff
@@ -768,73 +769,166 @@ def update(name, sql_dir, project_id):
...
-                "<dataset>.<table> to create the destination table."
-            )
+                tmp_schema_file = NamedTemporaryFile()
+                schema = Schema.from_schema_file(query_file.parent / SCHEMA_FILE)
+                schema.to_json_file(Path(tmp_schema_file.name))
+                bigquery_schema = client.schema_from_json(tmp_schema_file.name)
+                tmp_table = bigquery.Table(tmp_identifier, schema=bigquery_schema)
+                client.create_table(tmp_table)
+
+                tmp_tables[identifier] = tmp_identifier
```
scholtzan, line 807: Create a temporary table with the updated schema and use this temporary table in all downstream dependencies

bigquery_etl/cli/query.py
```diff
@@ -768,73 +769,166 @@ def update(name, sql_dir, project_id):
...
+    try:
+        query_schema = Schema.from_query_file(query_file_path, sql_content)
+    except Exception:
+        click.echo(
+            click.style(
+                f"Cannot automatically update {query_file_path}. "
+                f"Please update {query_file_path / SCHEMA_FILE} manually.",
+                fg="red",
+            ),
+            err=True,
```
scholtzan, line 858: For tables that reference themselves, like `clients_last_seen` or `clients_first_seen` the dry run fails here. I don't think there is a straightforward way to automatically update the schema for those.

sql/moz-fx-data-shared-prod/internet_outages/global_outages_v1/schema.yaml
```diff
@@ -0,0 +1,49 @@
+fields:
```
scholtzan, line 1: For testing this I ended up creating the schemas for all downstream dependencies of clients_daily

bigquery_etl/cli/query.py
jklukas: I can't think of a reason. It's likely an oversight. cc @whd

bigquery_etl/cli/query.py
jklukas: The only way forward I can think of would be adding support for "merging" the clients_daily schema.yaml content into the downstream tables. We'd probably need some additional metadata to indicate this is a reasonable thing to do, though, since that path would be fairly specific to the clients_last_seen problem.

bigquery_etl/cli/query.py
```diff
@@ -768,73 +769,166 @@ def update(name, sql_dir, project_id):
...
+    except NotFound:
+        click.echo(
+            f"Destination table {project_name}.{dataset_name}.{table_name} "
+            "does not exist in BigQuery. Run bqetl query schema deploy "
+            "<dataset>.<table> to create the destination table."
+        )
+    except FileNotFoundError:
+        click.echo(
+            f"No metadata file for {project_name}.{dataset_name}.{table_name}."
+            "Skip schema update."
```
jklukas, line 900:
```suggestion
            " Skip schema update."
```

sql/moz-fx-data-shared-prod/internet_outages/global_outages_v1/metadata.yaml
```diff
@@ -55,22 +54,31 @@ description: |-
...
-  public_json: false
-  public_bigquery: false
   review_bugs:
-    - 1640204
-  incremental_export: false
+  - '1640204'
 scheduling:
   dag_name: bqetl_internet_outages
-  # provide this value so that DAG generation does not have to dry run the
-  # query to get it, and that would be slow because main_v4 is referenced
```
jklukas, line 72: Let's keep this comment.

jklukas (approved):

PR #2017, opened 2021-05-07
Enable schema validation

scholtzan:
This enables the `validate-schemas` step in CI.

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -1037,7 +1037,7 @@ def _validate_schema(query_file):
     """
     if str(query_file) in SKIP or query_file.name == "script.sql":
         click.echo(f"{query_file} dry runs are skipped. Cannot validate schemas.")
-        return
+        return (True, query_file)
 
     query_file_path = Path(query_file)
     query_schema = Schema.from_query_file(query_file_path)
@@ -1130,3 +1130,5 @@ def validate_schema(name, sql_dir, project_id):
 
     if not all_valid:
         sys.exit(1)
+    else:
+        click.echo("\nAll schemas are valid.")
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -1030,6 +1030,11 @@ def deploy(ctx, name, sql_dir, project_id, force):
 
 
 def _validate_schema(query_file):
+    """
+    Check whether schema is valid.
+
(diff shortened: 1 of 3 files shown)


bigquery_etl/cli/query.py
```diff
@@ -1106,5 +1117,18 @@ def validate_schema(name, sql_dir, project_id):
...
+            if all_valid:
+                click.echo("\nSchemas for the following queries are invalid:")
+
+            all_valid = False
+            click.echo(r[1])
+
+    if not all_valid:
         sys.exit(1)
+    else:
+        click.echo("\nAll schemas are valid.")
```
scholtzan, line 1134: Since the schema validation step potentially logs out a lot of information (e.g. fields where descriptions differ etc.) it could become hard to find the schemas that don't pass validation. So it might be handy to list schemas that failed once validation is complete.

bigquery_etl/schema/__init__.py
```diff
@@ -46,13 +48,17 @@ def from_json(cls, json_schema):
...
-        query = f"SELECT * FROM {project}.{dataset}.{table}"
+        query = f"SELECT * FROM `{project}.{dataset}.{table}`"
 
         if partitioned_by:
             query += f" WHERE DATE({partitioned_by}) = DATE('2020-01-01')"
 
         # write query to temporary file so it can get dry run
-        tmp = NamedTemporaryFile()
+        # ensure that directory structure corresponds to original structure,
+        # otherwise dry run might fail when extracting the dataset from the path
```
scholtzan, line 58: I'm referring to https://github.com/mozilla/bigquery-etl/blob/8a55e8f429a76a309eb0568c2ca25adb918cf64b/bigquery_etl/dryrun.py#L262 here. By default when creating a temporary file in the Circle CI step, it will write the file to `tmp/some-random-name` which causes a failure in `Dryrun`.

jklukas: I've written shim code like this before, but I think you can pass content directly in the `DryRun` constructor rather than writing to a temp file, in which case we don't need to worry about directory structure. But I may be missing a detail here.

jklukas (approved):
r+ with some potential cleanups

bigquery_etl/schema/__init__.py
scholtzan: Indeed, you are right. No need for temporary files.

PR #2024, opened 2021-05-12
Fix typo in mozilla_vpn_derived/all_subscriptions_v1/query.sql

relud:


acmiyaguchi (approved):
For some reason, the 1 offset didn't seem wrong to me. r+

PR #2027, opened 2021-05-12
[Bug 1708169] Add automated Glean datasets, across all channels, to per-app datasets

scholtzan:
This PR adds a new command to the CLI: `glean_usage`

`./bqetl glean_usage generate` will generate all appId-specific and app_name-specific datasets.

Still a WIP, I'll add a `--app_name` option if this is the path we want to go. Also, the scripts for generating and running Glean queries need to be replaced in telemetry-airflow to instead use `./bqetl glean_usage generate` and `./bqetl query backfill`


bigquery_etl/glean_usage/common.py
```diff
@@ -113,88 +141,120 @@ def _extract_dataset_from_glob(pattern):
...
+    """Represents a generated Glean table."""
+
+    def __init__(self):
+        """Init Glean table."""
+        self.target_table_id = ""
+        self.prefix = ""
+        self.custom_render_kwargs = {}
+        self.no_init = True
+
+    def generate_per_app_id(
```
scholtzan, line 170: Most of the old `run_query` logic was pretty redundant. So I moved the logic into a `GleanTable` class (maybe there is a better name?) that the specific baseline table implementations can extend with some table specific configuration or logic.

bigquery_etl/glean_usage/common.py
```diff
@@ -113,88 +141,120 @@ def _extract_dataset_from_glob(pattern):
...
+        if output_only:
+            return
+
+        # dry run generated query
+        DryRun(
+            os.path.join(
+                output_dir, *list(baseline_table.split(".")[-2:]), query_filename
+            ),
+            query_sql,
+        ).is_valid()
```
scholtzan, line 231: This doesn't actually do anything if the dry run fails, except for printing an error (like the old implementation). Not sure if we'd want to handle failed dry run cases?

fbertsch:
Letting Jeff handle the review here.

jklukas (commented):
This refactor looks good, so is worth moving ahead with as long as we can develop a good strategy for initializing tables, as discussed in one of the comments.

bigquery_etl/glean_usage/common.py
```diff
@@ -113,88 +141,113 @@ def _extract_dataset_from_glob(pattern):
...
+        if not self.no_init:
+            try:
+                init_sql = render(init_filename, **render_kwargs)
+            except TemplateNotFound:
+                init_sql = render(query_filename, init=True, **render_kwargs)
+
+        if not (referenced_table_exists(view_sql)):
+            logging.info(
+                "Skipping view for table which doesn't exist:"
+                f" {self.target_table_id}"
```
jklukas, line 202: This should include a qualifier that indicates which app_id/dataset is involved.

bigquery_etl/cli/glean_usage.py
```diff
@@ -0,0 +1,106 @@
...
+    "--project-id",
+    "--project_id",
+    help="GCP project ID",
+    default="moz-fx-data-shared-prod",
+    callback=is_valid_project,
+)
+@click.option(
+    "--output-dir",
+    "--output_dir",
+    help="Output directory generated SQL is written to",
```
jklukas, line 42: To make this consistent with use elsewhere in the CLI, we should include `project_id` in the generated paths under output_dir. Currently, we have to pass in the project-level directory to this option, which is surprising. It seems like this refactor is the right place to fix this problem.

bigquery_etl/cli/glean_usage.py
```diff
@@ -0,0 +1,106 @@
...
+                partial(
+                    table.generate_per_app_id,
+                    project_id,
+                    output_dir=output_dir,
+                ),
+                baseline_tables,
+            )
+
+        # per app specific datasets
+        app_info = get_app_info()
```
jklukas, line 94: This should only be called once, outside the loop here.

bigquery_etl/glean_usage/common.py
```diff
@@ -113,88 +141,113 @@ def _extract_dataset_from_glob(pattern):
...
             )
+            return
+
+        view = f"{project_id}.{target_dataset}.{target_view_name}"
+
+        if output_dir:
+            write_sql(output_dir, view, "view.sql", sql)
+            write_dataset_metadata(output_dir, view)
```
jklukas, line 253: We need to write out the dataset metadata before we do the referenced_table_exists check. `referenced_table_exists` fails due to the per-app dataset not yet existing.

jklukas: We also may need to create the directory in write_dataset_metadata.

jklukas (commented):
I checked this out and tried running it. Leaving a round of comments about some errors I found.

bigquery_etl/glean_usage/templates/cross_channel.view.sql
```diff
@@ -0,0 +1,11 @@
+-- Generated via ./bqetl glean_usage generate
+CREATE OR REPLACE VIEW
+    `{{ project_id }}.{{ target_view }}`
+AS
+{% for (dataset, channel) in datasets -%}
+{% if not loop.first -%}
+UNION ALL
+{% endif -%}
+SELECT * REPLACE("{{ channel }}" AS normalized_channel)
+FROM {{ dataset }}.{{ table }}
```
jklukas, line 10:
```suggestion
FROM `{{ project_id }}.{{ dataset }}.{{ table }}`
```

Following our convention of always using fully-qualified names in views.

bigquery_etl/cli/query.py
```diff
@@ -581,15 +585,22 @@ def backfill(
...
+            depends_on_past = False
 
         if depends_on_past and exclude != []:
             click.echo(
                 f"Warning: depends_on_past = True for {query_file_path} but the"
                 f"following dates will be excluded from the backfill: {exclude}"
             )
 
+        if init:
+            ctx.invoke(initialize, name=query_file, dry_run=dry_run)
```
jklukas, line 602: Maybe instead of having an `init` flag, we can check for existence of the table via dry run. If the table does not exist, then we run `initialize`.

For the Airflow task, we need to have some a single invocation that runs successfully for both the initialization and the incremental update case. That could be `bqetl query backfill --init` but in that case we need to make sure that the init is a no-op if the table already exists.

bigquery_etl/glean_usage/common.py
```diff
@@ -45,6 +45,35 @@ def write_sql(output_dir, full_table_id, basename, sql):
...
 
 
+def write_dataset_metadata(output_dir, full_table_id):
+    """
+    Add dataset_metadata.yaml to public facing datasets.
+
+    Does not overwrite existing dataset_metadata.yaml files.
+    """
+    d = Path(os.path.join(output_dir, *list(full_table_id.split(".")[-2:])))
+    d.mkdir(parents=True, exist_ok=True)
```
jklukas, line 55:
```suggestion
    d.parent.mkdir(parents=True, exist_ok=True)
```

Currently, we're writing out empty directories like `moz-fx-data-shared-prod/fenix/baseline_clients_daily`

bigquery_etl/glean_usage/templates/cross_channel.view.sql
```diff
@@ -0,0 +1,11 @@
+-- Generated via ./bqetl glean_usage generate
+CREATE OR REPLACE VIEW
+    `{{ project_id }}.{{ target_view }}`
+AS
+{% for (dataset, channel) in datasets -%}
+{% if not loop.first -%}
+UNION ALL
+{% endif -%}
+SELECT * REPLACE("{{ channel }}" AS normalized_channel)
```
jklukas, line 9: I forgot that for Fenix, this channel thing is complicated enough to require a UDF.

So we should have logic here for `if app_name = "fenix"` and this line would be:

```
 SELECT * REPLACE(mozfun.norm.fenix_app_info("{{ dataset }}", app_build_id).channel AS normalized_channel)
```

bigquery_etl/cli/query.py
scholtzan: Instead of doing a dry run, it's probably easier to instantiate a BigQuery client and check if the table exists. When running the backfill BigQuery access is required anyway

bigquery_etl/util/__init__.py
```diff
@@ -1 +1,18 @@
...
+    """Extract table, dataset and project ID from query file."""
+    query_path = Path(path)
+
+    if query_path.is_file():
+        # path might be to query.sql, schema.yaml, ...
+        query_path = query_path.parent
+
+    table = query_path.name
+    dataset = query_path.parent.name
+    project = query_path.parent.parent.name
```
scholtzan, line 16: We do this all over the place. Might be nice to have a utility method that extracts these parts from file paths.

bigquery_etl/glean_usage/templates/cross_channel.view.sql
```diff
@@ -0,0 +1,11 @@
+-- Generated via ./bqetl glean_usage generate
+CREATE OR REPLACE VIEW
+    `{{ project_id }}.{{ target_view }}`
+AS
+{% for (dataset, channel) in datasets -%}
+{% if not loop.first -%}
+UNION ALL
+{% endif -%}
+SELECT * REPLACE(mozfun.norm.fenix_app_info("{{ dataset }}", app_build_id).channel AS normalized_channel)
```
jklukas, line 9:
```suggestion
{% if app_name == "fenix" -%}
SELECT * REPLACE(mozfun.norm.fenix_app_info("{{ dataset }}", app_build_id).channel AS normalized_channel)
{% else -%}
SELECT * REPLACE("{{ channel }}" AS normalized_channel)
{% endif -%}

```

bigquery_etl/cli/query.py
jklukas: It looks like we now rely on the existence check and ignore the `init` argument. Should we remove `init`?

scholtzan: Ah, yes. Removed

script/generate_sql
```diff
@@ -41,22 +41,7 @@ python -m bigquery_etl.view.generate_stable_views \
...
-    --project-id "${TARGET_PROJECT}" \
-    --output-dir "${SQL_DIR}/${TARGET_PROJECT}" \
-    --output-only \
-    --date="${PLACEHOLDER_DATE}"
-./script/run_glean_baseline_clients_first_seen \
-    --project-id "${TARGET_PROJECT}" \
-    --output-dir "${SQL_DIR}/${TARGET_PROJECT}" \
-    --output-only \
-    --date="${PLACEHOLDER_DATE}"
+./bqetl glean_usage generate --project-id "${TARGET_PROJECT}"
```
jklukas, line 44:
```suggestion
./bqetl glean_usage generate --project-id "${TARGET_PROJECT}" --output-dir "${SQL_DIR}"
```

bigquery_etl/glean_usage/common.py
```diff
@@ -113,88 +142,108 @@ def _extract_dataset_from_glob(pattern):
...
+
+        datasets = [
+            (a["bq_dataset_family"], a["app_channel"])
+            for a in app_info
+            if "app_channel" in a
+        ]
+
+        # Some apps only have a single channel, in which case the per-app_id dataset
+        # is already sufficient.
+        if len(datasets) == 0:
```
jklukas, line 227: This condition meets the spirit of the Per-Glean App Datasets proposal, but it leaves us in a spot where some apps will not have a namespace corresponding to the configured `app_name`. Looker automation may require this.

Currently, this logics creates per-app datasets only for `fenix` and `firefox_ios` (since those are the only multi-channel Glean apps). That's probably fine for now, and we can address this question as a future follow-up.

cc @fbertsch

jklukas (approved):
I've validated that the output of this new logic produces the same generated output as the old, except for the expected additions of `fenix` and `firefox_ios` datasets.

PR #2033, opened 2021-05-18
Bug 1698934 - support Rally datasets in Shredder

akkomar:


whd (commented):
I merged https://github.com/mozilla-services/cloudops-infra/pull/3198 preliminary to looking at this PR, but after reviewing this and taking a closer look at the data I think we wll need to fix `rally-zero-one` yet again. I believe :acmiyaguchi implemented a conversion to transform `pioneer_id` to `rally_id` for namespaces that start with `rally-`. This ended up being the wrong thing to do more or less since the top-level `rally_id` field we added was also the wrong thing to do, and the correct column for glean-based studies is actually `metrics.uuid.rally_id`.

This puts us in a place where `rally_zero_one` is unlike any other data, because it has `rally_id` but not `metrics.uuid.rally_id`. Currently, per https://bugzilla.mozilla.org/show_bug.cgi?id=1698934#c17, I believe the `deletion_requests` views expose the id column at the top-level only (e.g. `pioneer_id` or `rally_id`). Currently, `rally_zero_one` is exposing `pioneer_id` as `rally_id` (literally `pioneer_id AS rally_id` in the view), when it should be exposing `rally_id` as `rally_id` (whereas glean-based apps will be exposing `metrics.uuid.rally_id` as `rally_id` in the deletion view).

The current state is rather confusing and I believe that even when I correct the `rally_zero_one` view, shredder will still fail because it will expect e.g. `rally_zero_one.measurements_v1` to include `metrics.uuid.rally_id`.

I wanted to avoid treating `zero_one` as a special case in shredder but AFAICT there isn't a good way of handling this without shredder understanding `metrics.uuid.rally_id` vs `rally_id` for the one study. In theory we can update the `deletion_requests` views to include the proper id field to delete data based on, which would look like e.g.:
```
study-type uuid_field   upstream_field
pioneer    deletion_id  pioneer_id
rally      deletion_id  metrics.uuid.rally_id
zero_one   deletion_id  rally_id
```
In the above view there is only `deletion_id` with a separate field to read upstream field from.

I think our two main options are to do something like the above, or once I have fixed the zero-one view to use top-level `rally_id`, simply add another special case for `rally-zero-one`.

akkomar:
@whd:
Reading https://github.com/mozilla/bigquery-etl/pull/2033#pullrequestreview-683445560 I realized that per https://bugzilla.mozilla.org/show_bug.cgi?id=1698934#c17, Rally deletion views used top-level `rally_id` field instead of the nested one. I have pushed a fixup commit here to correctly handle that.

Second commit above adds special handling for `rally_zero_one` to Shredder. It now expects ping tables from this study to use `rally_id` as client identifier. As for the `deletion_requests` view, it expects a `rally_id` field there (similar to proper Rally studies).
Reading the comment above, infra PR and looking at `deletion_requests` schemas I'm not sure this solves the problem (I think a special case of deletion request view for `rally_zero_one` is still needed).

whd:
> Reading the comment above, infra PR and looking at deletion_requests schemas I'm not sure this solves the problem (I think a special case of deletion request view for rally_zero_one is still needed).

Yes I will need to fix zero one regardless. However, I think having special cases in both repos as you've just implemented is the simplest solution. I'll approve this once I have fixed the deletion view for zero-one.

whd (approved):
I think this combined with https://github.com/mozilla-services/cloudops-infra/pull/3201 will do the trick, but I'll wait for confirmation before merging + testing

PR #2036, opened 2021-05-18
Update transitive dependency to fix deploy CI

relud:
this should fix https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/9313/workflows/87c97852-936f-4950-9261-962edab01e87/jobs/56935


scholtzan (approved):
Great! No need for https://github.com/mozilla/bigquery-etl/pull/2035 then

PR #2038, opened 2021-05-18
Optional project_id parameter for is_authenticated

scholtzan:
Should fix failures in
https://github.com/mozilla/telemetry-airflow/blob/8be7a8493f77001e075b094d5eab8bc7060cf943/dags/copy_deduplicate.py#L165

### DIFF
--- a/bigquery_etl/cli/utils.py
+++ b/bigquery_etl/cli/utils.py
@@ -25,10 +25,14 @@ def is_valid_file(ctx, param, value):
     return value
 
 
-def is_authenticated(project_id="moz-fx-data-shared-prod"):
+def is_authenticated(project_id=None):
     """Check if the user is authenticated to GCP and can access the project."""
     client = bigquery.Client()
-    return client.project == project_id
+
+    if project_id:
+        return client.project == project_id
+
+    return True
 
 
 def is_valid_project(ctx, param, value):


bigquery_etl/cli/utils.py
```diff
@@ -25,10 +25,14 @@ def is_valid_file(ctx, param, value):
     return value
 
 
-def is_authenticated(project_id="moz-fx-data-shared-prod"):
```
scholtzan, line 28: The Airflow tasks seem to run in the `derived` project. So this check fails

bigquery_etl/cli/utils.py
```diff
@@ -25,10 +25,14 @@ def is_valid_file(ctx, param, value):
...
+def is_authenticated(project_id=None):
     """Check if the user is authenticated to GCP and can access the project."""
     client = bigquery.Client()
```
scholtzan, line 30: This will raise an exception [if the default credentials cannot be acquired](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.client.Client.html#google.cloud.bigquery.client.Client)

bigquery_etl/cli/utils.py
jklukas: It looks like gke_command by default uses `project_id=GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id).project_id` which gives derived-datasets because the GKE cluster still lives there. We should be launching these in shared-prod, so we can add an explict `project_id` param to these tasks in the DAG.

It looks like we updated `bigquery_etl_query` to use shared-prod by default, but not this function.

jklukas (approved):

PR #2041, opened 2021-05-18
[Bug 1710088] Generate events_unnested views for app datasets

scholtzan:
https://bugzilla.mozilla.org/show_bug.cgi?id=1710088

Creates `events_unnested` views for each app dataset as part of `./bqetl glean_usage generate`.

### DIFF
--- a/bigquery_etl/cli/glean_usage.py
+++ b/bigquery_etl/cli/glean_usage.py
@@ -12,6 +12,7 @@ from ..glean_usage import (
     baseline_clients_daily,
     baseline_clients_first_seen,
     baseline_clients_last_seen,
+    events_unnested,
 )
 from ..glean_usage.common import list_baseline_tables, get_app_info
 
@@ -109,3 +110,14 @@ def generate(project_id, output_dir, parallelism, exclude, only, app_name):
                 partial(table.generate_per_app, project_id, output_dir=output_dir),
                 app_info,
             )
+
+    # generate per-app events_unnested views
+    with ThreadPool(parallelism) as pool:
+        pool.map(
+            partial(
+                events_unnested.EventsUnnestedTable().generate_per_app,
+                project_id,
+                output_dir=output_dir,
+            ),
+            app_info,
+        )
--- a/bigquery_etl/cli/glean_usage.py
+++ b/bigquery_etl/cli/glean_usage.py
@@ -21,6 +21,7 @@ GLEAN_TABLES = [
     baseline_clients_daily.BaselineClientsDailyTable(),
     baseline_clients_first_seen.BaselineClientsFirstSeenTable(),
     baseline_clients_last_seen.BaselineClientsLastSeenTable(),
+    events_unnested.EventsUnnestedTable(),
 ]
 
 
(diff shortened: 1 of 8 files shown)


bigquery_etl/glean_usage/templates/cross_channel_events_unnested.view.sql
```diff
@@ -0,0 +1,16 @@
+-- Generated via ./bqetl glean_usage generate
+CREATE OR REPLACE VIEW
+    `{{ project_id }}.{{ target_view }}`
+AS
+{% for (dataset, channel) in datasets -%}
+{% if not loop.first -%}
+UNION ALL
```
scholtzan, line 7: `UNION ALL` on Fenix `event` views doesn't work since the ordering of the columns is different.

scholtzan: I left a comment in https://bugzilla.mozilla.org/show_bug.cgi?id=1710088 to discuss our options

bigquery_etl/glean_usage/templates/cross_channel_events_unnested.view.sql
```diff
@@ -0,0 +1,16 @@
...
+{% endif -%}
+{% if app_name == "fenix" -%}
+SELECT e.* EXCEPT (events) REPLACE(mozfun.norm.fenix_app_info("{{ dataset }}", client_info.app_build).channel AS normalized_channel), event.*
+{% else -%}
+SELECT e.* EXCEPT (events) REPLACE("{{ channel }}" AS normalized_channel), event.*
+{% endif -%}
+FROM `{{ project_id }}.{{ dataset }}.events` e
+LEFT JOIN UNNEST(e.events) AS event
```
jklukas, line 15: Why is this a LEFT JOIN? I don't think it makes sense to have rows with empty events, so I would expect a CROSS JOIN.

In practice, it probably doesn't matter because I don't think Glean is supposed to send empty event pings.

bigquery_etl/glean_usage/templates/cross_channel_events_unnested.view.sql
```diff
@@ -0,0 +1,16 @@
+-- Generated via ./bqetl glean_usage generate
+CREATE OR REPLACE VIEW
+    `{{ project_id }}.{{ target_view }}`
+AS
+{% for (dataset, channel) in datasets -%}
+{% if not loop.first -%}
+UNION ALL
+{% endif -%}
+{% if app_name == "fenix" -%}
+SELECT e.* EXCEPT (events) REPLACE(mozfun.norm.fenix_app_info("{{ dataset }}", client_info.app_build).channel AS normalized_channel), event.*
```
jklukas, line 10: As discussed in https://bugzilla.mozilla.org/show_bug.cgi?id=1710088#c2 we'd like these event fields to be prefixed with `event_`, so they'll have to be listed explicitly. We can probably factor that part out of the conditional so it doesn't have to be repeated.

bigquery_etl/cli/glean_usage.py
```diff
@@ -109,3 +110,14 @@ def generate(project_id, output_dir, parallelism, exclude, only, app_name):
...
+    # generate per-app events_unnested views
+    with ThreadPool(parallelism) as pool:
+        pool.map(
+            partial(
+                events_unnested.EventsUnnestedTable().generate_per_app,
+                project_id,
+                output_dir=output_dir,
+            ),
+            app_info,
+        )
```
jklukas, line 123: Can we include `EventsUnnestedTable` in `GLEAN_TABLES` and avoid this special case? It seems fine to have the `generate_per_app_id` be a no-op and let it run. Are there other reasons to have this be a special case?

bigquery_etl/glean_usage/templates/cross_channel_events_unnested.view.sql
fbertsch: +1. FWIW I've checked across lots of events tables and never seen an empty events array.

PR #2042, opened 2021-05-18
Pass project_id to run() when running backfill

scholtzan:
Currently, the `--project_id` parameter is ignored when running a backfill. This should fix it


scholtzan:
I'm going ahead and merge this to ensure Airflow is finishing successfully.

PR #2043, opened 2021-05-19
Add --no-partition option to backfill

scholtzan:
`bqetl query backfill` always writes results to a partition. This doesn't work for clients_first_seen tables though. I added a `--no-partition` option to write to the destination table instead.


scholtzan:
Merging now so the Airflow tasks can finish

jklukas (commented):
Retroactive r+

PR #2045, opened 2021-05-19
DO-500 Rename search_clients_daily

jklukas:
See https://jira.mozilla.com/projects/DO/issues/DO-500

`search_clients_daily` becomes `search_clients_engines_sources_daily`
which more accurately represents the data model. We also rename the mobile
equivalent and the related `last_seen` views.

The old views stick around for now. We will need to announce deprecation of the
old views and a timeline for turning them off.

The underlying tables are unchanged. This effort is only about changing the
user interface.

### DIFF
--- a/sql/moz-fx-data-shared-prod/search/mobile_search_clients_engines_sources_daily/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/search/mobile_search_clients_engines_sources_daily/metadata.yaml
@@ -1,14 +1,8 @@
 ---
-friendly_name: Mobile Search Clients Daily
+friendly_name: Mobile Search Clients Engines Sources Daily
 description: |
   A daily aggregate of baseline and metrics pings that have
   searches from each (client_id, engine, source) combination,
   partitioned by day.
-
-  Exposed to users as `search.mobile_search_clients_engines_sources_daily`.
 owners:
   - bewu@mozilla.com
-labels:
-  schedule: daily
-scheduling:
-  dag_name: bqetl_mobile_search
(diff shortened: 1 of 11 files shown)


sql/moz-fx-data-shared-prod/search/mobile_search_clients_engines_sources_daily/metadata.yaml
```diff
@@ -0,0 +1,8 @@
+---
+friendly_name: Mobile Search Clients Engines Sources Daily
+description: |
+  A daily aggregate of baseline and metrics pings that have
+  searches from each (client_id, engine, source) combination,
```
jklukas, line 5: `type` (or `search_type`) is also used as a key in the `GROUP BY`, but if I'm reading this logic correctly, each `engine` will only be associated with a `type`. If that's not correct, then this name may need to include `_type` as well.

BenWu: Search type can be determined by source and the field is a bit of an implementation detail so I don't think it's necessary to include `_type`

BenWu (approved):

sql/moz-fx-data-shared-prod/search/mobile_search_clients_engines_sources_daily/metadata.yaml
xluo-ds: LGTM. 

Just note that for `mobile` there is also `app_name`, which seems can be determined by (`source`, `client_id`, 
`engine`), I feel like `app_name` is more common dimension than `source`; however, your proposed name might be better in the spirit of consistent naming.

xluo-ds (approved):

jklukas:
> Just note that for mobile there is also app_name, which seems can be determined by (source, client_id,
engine)

Indeed. A given `client_id` should only ever appear associated with one app_name and one channel, so these extra dimensions should not affect the contract as to what each row represents.

PR #2050, opened 2021-05-20
Remove search_engines_clients_sources_last_seen views

jklukas:
Turns out that the last_seen table has a different grain compared to the daily table, so this is partially backing out https://github.com/mozilla/bigquery-etl/pull/2045

It also clarifies that the last_seen tables here have a year window rather than a 28-day window.

I will need to manually delete the already-deployed `search_engines_clients_sources_last_seen` views once this is merged.


BenWu (approved):
Thanks @fbertsch for correcting this

fbertsch (approved):
LGTM

jklukas:
I've removed these views from both shared-prod and mozdata.

PR #2054, opened 2021-05-20
Operational Monitoring.

emtwo:
This PR is in progress. Here are the items that need to be done for it to be ready for review:
- [x] init & query file generation
- [x] output to bq destination table
- [x] allow running from bqetl cli
- [x] dimensions/probes/slugs/branches should be pulled in based on external files

### DIFF
--- a/bigquery_etl/operational_monitoring/__init__.py
+++ b/bigquery_etl/operational_monitoring/__init__.py
@@ -1,0 +1,195 @@
+"""Generate and run the Operational Monitoring Queries."""
+import os
+import re
+import time
+import json
+import click
+import logging
+
+from datetime import datetime, date
+from pathlib import Path
+from multiprocessing.pool import ThreadPool
+
+from google.cloud import storage, bigquery
+
+from bigquery_etl.util.common import render, write_sql
+
+
+QUERY_FILENAME = "query.sql"
+INIT_FILENAME = "init.sql"
+BUCKET_NAME = "operational_monitoring"
+PROJECTS_FOLDER = "projects/"
+OUTPUT_DIR = "sql/moz-fx-data-shared-prod/"
+PROD_PROJECT = "moz-fx-data-shared-prod"
+
+
+def _download_json_file(project, bucket, filename):
+    blob = bucket.get_blob(filename)
+    return json.loads(blob.download_as_string()), blob.updated
+
+
+def _get_name_and_sql(query, reference_content, item_type):
+    return [
+        {"name": item, "sql": reference_content[item]["sql"]}
+        for item in query[item_type]
+    ]
+
+
+def _write_sql(project, dataset, slug, kwargs, init):
+    basename = INIT_FILENAME if init else QUERY_FILENAME
+    write_sql(
+        OUTPUT_DIR,
+        f"{project}.{dataset}.{slug}",
+        basename,
+        render(
+            basename,
+            template_folder="operational_monitoring",
+            **kwargs,
+            init=init,
+        ),
+    )
+
+
+def _bq_normalize_name(name):
+    return re.sub(r"[^a-zA-Z0-9_]", "_", name)
+
+
+def _query_up_to_date(dataset, slug, basename, project_last_modified):
+    """Check whether the project file was updated after the query.sql file was written."""
+    query_path = Path(os.path.join(OUTPUT_DIR, dataset, slug, basename))
+    if not os.path.exists(query_path):
+        return False
+
+    query_last_modified = datetime.utcfromtimestamp(query_path.stat().st_mtime)
+    if query_last_modified < project_last_modified.replace(tzinfo=None):
+        return False
+
+    return True
+
+
+def _generate_sql(project, dataset):
+    render_kwargs = {
+        "header": "-- Generated via bigquery_etl.operational_monitoring\n",
+        "gcp_project": project,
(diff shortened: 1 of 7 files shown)


emtwo:
Associated airflow PR: https://github.com/mozilla/telemetry-airflow/pull/1322

emtwo:
Right now the dry-run fails because I am referencing a `mozfun` function that I added in this PR. I can add it in a separate PR if that's better

bigquery_etl/operational_monitoring/templates/query.sql
```diff
@@ -0,0 +1,94 @@
...
+    `{{source}}`
+  WHERE
+    DATE(submission_timestamp) >= DATE_SUB(@submission_date, INTERVAL 30 DAY)
+  GROUP BY
+    1,
+    2,
+    3,
+    4,
+    5,
+    6
```
scholtzan, line 45: It looks like `dimensions` could be an arbitrary set of dimensions. Could this break the `GROUP BY` here (e.g. if `dimensions` happen to contain 100 different dimensions)?

bigquery_etl/operational_monitoring/templates/query.sql
```diff
@@ -0,0 +1,94 @@
...
+    {{ "," if not loop.last else "" }}
+  {% endfor %}
+)
+GROUP BY
+  1,
+  2,
+  3,
+  4,
+  5,
+  6
```
scholtzan, line 94: Same concern as above query. It might also be better to use the column names here instead to avoid confusion

sql/moz-fx-data-shared-prod/operational_monitoring/bug-1660366-pref-ongoing-fission-nightly-experiment-nightly-83-100/query.sql
```diff
@@ -0,0 +1,89 @@
+-- Generated via bigquery_etl.operational_monitoring
```
scholtzan, line 1: I guess generated files should not be checked in into the repository. Looks like the DAG in https://github.com/mozilla/telemetry-airflow/pull/1322/files would generate the queries before running them

sql/moz-fx-data-shared-prod/operational_monitoring/bug-1660366-pref-ongoing-fission-nightly-experiment-nightly-83-100/init.sql
```diff
@@ -0,0 +1,30 @@
+CREATE TABLE IF NOT EXISTS
```
scholtzan, line 1: This should probably also be removed before merging

bigquery_etl/operational_monitoring/__init__.py
```diff
@@ -0,0 +1,177 @@
...
+
+            _write_sql(project, dataset, slug, render_kwargs, init=True)
+            _write_sql(project, dataset, slug, render_kwargs, init=False)
+
+
+def _run_sql(project, dataset, submission_date):
+    bq_client = bigquery.Client(project=PROD_PROJECT)
+
+    for slug in os.listdir(os.path.join(OUTPUT_DIR, dataset)):
+        destination_table = f"{project}.{dataset}.{slug}"
```
scholtzan, line 117: For the `slug` part, generally the naming convention for tables is not to use dashes but to use underscores instead. This is how we do it in jetstream: https://github.com/mozilla/jetstream/blob/29efddb0825678b55504ccc07bfe86c4363bdad9/jetstream/__init__.py#L22

bigquery_etl/operational_monitoring/__init__.py
```diff
@@ -0,0 +1,177 @@
...
+            )
+
+            _write_sql(project, dataset, slug, render_kwargs, init=True)
+            _write_sql(project, dataset, slug, render_kwargs, init=False)
+
+
+def _run_sql(project, dataset, submission_date):
+    bq_client = bigquery.Client(project=PROD_PROJECT)
+
+    for slug in os.listdir(os.path.join(OUTPUT_DIR, dataset)):
```
scholtzan, line 116: This should probably also run in parallel.

bigquery_etl/operational_monitoring/__init__.py
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    client = storage.Client(project)
+    bucket = client.get_bucket(BUCKET_NAME)
+
+    probes, _ = _download_json_file(project, bucket, "probes.json")
+    dimensions, _ = _download_json_file(project, bucket, "dimensions.json")
+
+    # Iterating over all defined operational monitoring projects
+    for blob in bucket.list_blobs(prefix=PROJECTS_FOLDER):
```
scholtzan, line 79: I haven't checked how long the generation takes, but if it takes significant time we might also run this in parallel

bigquery_etl/operational_monitoring/templates/query.sql
emtwo: oh good catch, thank you

emtwo:
@scholtzan I addressed your comments. I didn't parallelize the query generation yet though (but parallelized the run). It seems to only take a couple seconds or less and right now we're only starting with 1 project anyways so we can add it later.

scholtzan (approved):

Issue #2056, opened 2021-05-21
core_clients_last_seen does not have updated first seen date


acmiyaguchi:
See https://sql.telemetry.mozilla.org/queries/80137/source

The last time the first seen date in the clients last seen table was updated was 2021-04-04. This is almost certainly incorrect.

jklukas:
This task is running daily, but perhaps it's having no effect: https://workflow.telemetry.mozilla.org/log?task_id=telemetry_derived__core_clients_first_seen__v1&dag_id=copy_deduplicate&execution_date=2021-05-20T01%3A00%3A00%2B00%3A00

acmiyaguchi:
It looks like the source of this issue is that the core clients first seen hasn't been updated since 2021-04-04: https://sql.telemetry.mozilla.org/queries/80138/source

acmiyaguchi:
It looks like the airflow logs show the core_clients_first_seen job failing silently:

```
RUNNING\r                                                                                      
\rWaiting on bqjob_r56fde488d9e4cec1_000001798c79156e_1 ... (1s) Current status: RUNNING\r                                                                                      
\rWaiting on bqjob_r56fde488d9e4cec1_000001798c79156e_1 ... (2s) Current status: RUNNING\r                                                                                      
\rWaiting on bqjob_r56fde488d9e4cec1_000001798c79156e_1 ... (3s) Current status: RUNNING\r                                                                                      
\rWaiting on bqjob_r56fde488d9e4cec1_000001798c79156e_1 ... (4s) Current status: RUNNING\r                                                                                      
\rWaiting on bqjob_r56fde488d9e4cec1_000001798c79156e_1 ... (4s) Current status: DONE   \n'
[2021-05-21 01:10:22,857] {pod_launcher.py:156} INFO - b'INFO: No metadata.yaml found for {} sql/moz-fx-data-shared-prod/telemetry_derived/core_clients_first_seen_v1/query.sql\n'
```

jklukas:
I don't think it's a silent failure. It's running to completion very quickly and producing no output. I'm able to find the job, and the result set is empty.

acmiyaguchi:
The use of `submission_date` for the date partition may be problematic; I'll put in a PR for that field. Otherwise, I'll write some unit tests for the query and see if I can debug the issue.

jklukas:
More proof that the query is running, is that we can see the empty partitions in information schema: https://sql.telemetry.mozilla.org/queries/80140/source

[...]

PR #2057, opened 2021-05-21
Fix empty result sets for incremental core clients first seen table

acmiyaguchi:
Should fix issues seen in #2056 after backfill. The condition in the main SQL script was causing the result set to always be empty: `cfs.first_seen_date < @submission_date AND cfs.client_id is NULL`.  In particular, the first condition of filtering out first_seen_date was also filtering out null values.

I also update the query to reflect the style of the baseline clients first seen queries.


sql/moz-fx-data-shared-prod/telemetry_derived/core_clients_first_seen_v1/query.sql
```diff
@@ -1,20 +1,27 @@
...
 FROM
-  clients_today
+  _current
 LEFT JOIN
-  telemetry_derived.core_clients_first_seen_v1 AS cfs
+  _previous
 USING
   (client_id)
 WHERE
-  cfs.first_seen_date < @submission_date
```
acmiyaguchi, line 19: This is the problematic statement; it's required because of the partition filter, but also happens to filter out null clients in the join (e.g. new clients to the table). I add this filter in a cte so it's not confounded here.

sql/moz-fx-data-shared-prod/telemetry_derived/core_clients_first_seen_v1/query.sql
```diff
@@ -1,20 +1,27 @@
...
   WHERE
     DATE(submission_timestamp) = @submission_date
+),
+_previous AS (
+  SELECT
+    *
+  FROM
+    telemetry_derived.core_clients_first_seen_v1
+  WHERE
+    first_seen_date > "2010-01-01"
```
jklukas, line 15:
```suggestion
    -- In the case we need to backfill older partitions of this table, we don't want newer partitions
    -- to alter results of the query.
    first_seen_date < @submission_date
```

sql/moz-fx-data-shared-prod/telemetry_derived/core_clients_first_seen_v1/query.sql
jklukas: Makes sense. Thanks for diving in and figuring this out.

jklukas (approved):
r+wc

acmiyaguchi:
Leaving this to be merged on Monday

PR #2062, opened 2021-05-24
Copy generated-sql content into docker container

jklukas:
This also adds a scheduled build at midnight UTC to ensure that the generated
content is updated shortly before nightly ETL starts.

This should allow us to relieve some awkwardness like in various contexts
where we need to run `script/generate_sql` before invoking an endpoint in the
container. In particular, it will allow us to revert
https://github.com/mozilla/telemetry-airflow/pull/1312
and may allow us to remove [generation when publishing views in Jenkins](https://github.com/mozilla-services/cloudops-infra/blob/de1f9e61e2a2236904449d8c482aa238cd7bffa2/projects/data-shared/Jenkinsfile.bigquery.prod#L139).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -324,6 +324,13 @@ jobs:
       - checkout
       - setup_remote_docker:
           docker_layer_caching: true
+      - attach_workspace:
+          at: /tmp/generated-sql
+      - run:
+          name: Move generated-sql into place
+          command:
+            rm -rf sql/
+            cp -r /tmp/generated-sql/sql sql
       - run:
           name: Determine docker image name
           command:
@@ -342,7 +349,7 @@ jobs:
 workflows:
   version: 2
   build:
-    jobs:
+    jobs: &build_jobs
       - build:
           context: data-eng-circleci-tests
       - verify-format-sql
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -324,6 +324,13 @@ jobs:
       - checkout
       - setup_remote_docker:
           docker_layer_caching: true
+      - attach_workspace:
+          at: /tmp/generated-sql
+      - run:
+          name: Move generated-sql into place
+          command:
+            rm -rf sql/
+            cp -r /tmp/generated-sql/sql sql
```
jklukas, line 333: This `deploy` step may turn out to be a good integration point for private-bigquery-etl code as well. The [proposal for private ETL code](https://docs.google.com/document/d/1GuuxseYBbUYiq-k4ut8LOEeVBtwwDB3cEBf35NmbTFE/edit#) currently discussed building a separated docker container, but @whd pointed out that consolidating artifacts is preferable from an ops management perspective.

relud (approved):

.circleci/config.yml
whd: I'm generally supportive of considerable simplification.

whd (approved):
I do have some reservations about this. This makes the docker container build non-idempotent, and leaves generated sql versioning somewhat lacking. But I generally support the end state this achieves, even though I would prefer it to eventually be via a mechanism that didn't involve querying production state.

PR #2063, opened 2021-05-24
Add options to stripe import for non-event resources

relud:


bigquery_etl/stripe/__init__.py
```diff
@@ -136,29 +149,36 @@ def import_(
...
+                ignore_unknown_values=False,
+                schema=filtered_schema.filtered,
+                source_format=bigquery.SourceFormat.NEWLINE_DELIMITED_JSON,
+                time_partitioning=bigquery.TimePartitioning(field="created"),
+                write_disposition=bigquery.WriteDisposition.WRITE_TRUNCATE,
+            )
+            if "$" in table:
+                job_config.schema_update_options = [
+                    bigquery.SchemaUpdateOption.ALLOW_FIELD_ADDITION
+                ]
```
relud, line 177: `schema_update_options` can only be set with `write_disposition=bigquery.WriteDisposition.WRITE_TRUNCATE` if the destination is a partition. This is because when it's a whole table `WRITE_TRUNCATE` causes the schema to be replaced instead of updated.

bigquery_etl/stripe/__init__.py
```diff
@@ -74,7 +80,7 @@ def _get_rows(
             yield FilteredSchema.expand(row)
 
 
-@click.group(help="Commands for Stripe ETL.")
+@click.group("stripe", help="Commands for Stripe ETL.")
 def stripe_():
```
relud, line 84: this results in the following change to the `--help` page for stripe commands:
```diff
-Usage: script/bqetl stripe- import [OPTIONS]
+Usage: script/bqetl stripe import [OPTIONS]
```

acmiyaguchi (approved):

PR #2065, opened 2021-05-24
The case of the missing sql/ directory

jklukas:
Whitespace. Pass it on.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -328,7 +328,7 @@ jobs:
           at: /tmp/generated-sql
       - run:
           name: Move generated-sql into place
-          command:
+          command: |
             rm -rf sql/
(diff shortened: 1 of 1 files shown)


jklukas:
This is a regression introduced in #2062 that caused views deploys to fail.

whd (approved):
:crying_cat_face:

.circleci/config.yml
```diff
@@ -328,7 +328,7 @@ jobs:
           at: /tmp/generated-sql
       - run:
           name: Move generated-sql into place
-          command:
+          command: |
```
jklukas, line 331: This was being interpreted as a single `rm -rf` invocation removing such illustrious directories as `cp` and `-r`.

relud (approved):

PR #2099, opened 2021-06-07
Bug 1714995 - add regrets reporter to shredder

relud:
confirmed via dry run this works as expected and doesn't break other tables.

[Bug 1714995](https://bugzilla.mozilla.org/show_bug.cgi?id=1714995)

### DIFF
--- a/bigquery_etl/shredder/config.py
+++ b/bigquery_etl/shredder/config.py
@@ -23,6 +23,7 @@ class DeleteSource:
     table: str
     field: str
     project: str = SHARED_PROD
+    conditions: Tuple[str, ...] = ()
 
     @property
     def table_id(self):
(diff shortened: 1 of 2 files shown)


bigquery_etl/shredder/config.py
```diff
@@ -98,6 +99,11 @@ def fields(self) -> Tuple[str, ...]:
 FXA_SRC = DeleteSource(
     table="firefox_accounts_derived.fxa_delete_events_v1", field=USER_ID
 )
+REGRETS_SRC = DeleteSource(
+    table="regrets_reporter_stable.regrets_reporter_update_v1",
+    field="data_deletion_request.extension_installation_uuid",
+    conditions=("data_deletion_request IS NOT NULL",),
```
acmiyaguchi, line 105: Accidentally making the tuple into a string would be caught by the type annotations?

acmiyaguchi (approved):

bigquery_etl/shredder/config.py
relud: it should, yeah.

PR #2105, opened 2021-06-09
Add operational monitoring views and dataset

emtwo:


emtwo:
@scholtzan hopefully I understood your explanation correctly =D

sql/moz-fx-data-shared-prod/operational_monitoring_derived/dataset_metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: Operational Monitoring Derived
+description: |-
+  Derived data based operational monitoring project requirements
```
scholtzan, line 3: It might be useful to add a mention here that the queries are generated via an Airflow task and link to it. Same for the operational_monitoring dataset

scholtzan (approved):
Yes, this should work 👍

PR #2108, opened 2021-06-09
Include interval count in VPN pricing plan

relud:


sql/moz-fx-data-shared-prod/mozilla_vpn_derived/all_subscriptions_v1/query.sql
```diff
@@ -264,7 +266,15 @@ vpn_subscriptions AS (
...
-  CONCAT(plan_interval, "-", plan_currency, "-", (plan_amount / 100)) AS pricing_plan,
+  CONCAT(
+    plan_interval_count,
+    "-",
+    plan_interval,
+    "-",
+    plan_currency,
+    "-",
+    (plan_amount / 100)
+  ) AS pricing_plan,
```
relud, line 277: This is needed because the 6-month plan available in euros was displaying incorrectly as `month-eur-41.97` and `6-month-eur-41.94` is less misleading.

This change doesn't make `pricing_plan` isn't any more or less unique, so dashboard counts won't be impacted.

This does not require any backfilling (no downstream tables that only update date partitions use this field)

sql/mozfun/iap/derive_apple_subscription_interval/udf.sql
```diff
@@ -1,119 +1,96 @@
...
+      (("month", 3), DATETIME "2021-03-15 00:00:00", DATETIME "2021-06-15 00:00:00"),
+      (("month", 1), DATETIME "2021-03-15 00:00:00", DATETIME "2021-04-15 00:00:00"),
+      (("week", 2), DATETIME "2021-03-15 00:00:00", DATETIME "2021-03-29 00:00:00"),
+      (("week", 1), DATETIME "2021-03-15 00:00:00", DATETIME "2021-03-22 00:00:00"),
+      (("day", 2), DATETIME "2021-03-15 00:00:00", DATETIME "2021-03-17 00:00:00"),
+      (("day", 1), DATETIME "2021-03-15 00:00:00", DATETIME "2021-03-16 00:00:00"),
+      -- last day of month
+      (("month", 1), DATETIME "2021-02-28 00:00:00", DATETIME "2021-03-31 00:00:00"),
+      (("month", 1), DATETIME "2021-01-31 00:00:00", DATETIME "2021-02-28 00:00:00"),
+      (("month", 1), DATETIME "2021-02-28 00:00:00", DATETIME "2021-03-28 00:00:00"),
```
acmiyaguchi, line 56: If the latter date were "2021-03-27", would this be ("month", 0)?

acmiyaguchi (approved):
What is an interval, and when can it be more than 1? A 6 month subscription might be an interval? It would increase for each new payment?

Otherwise the changes look good to me.

sql/mozfun/iap/derive_apple_subscription_interval/udf.sql
relud: that would come out to `("day", 27)`

relud:
> What is an interval, and when can it be more than 1? A 6 month subscription might be an interval? It would increase for each new payment?

interval is a unit of time, and interval_count is the quantity of time in those units. together they indicate how often a subscription needs to be paid. A 6 month subscription would be `("month", 6)` and you'd be paying for 6 months at a time. It does not change with each payment, but it could in theory be used to determine how many times a subscription has been renewed.

PR #2110, opened 2021-06-10
Bug 1709871 - Integrate private bigquery_etl into Docker image

akkomar:


.circleci/config.yml
```diff
@@ -343,11 +343,22 @@ jobs:
...
           command: |
             rm -rf sql/
             cp -r /tmp/generated-sql/sql sql
+      - run:
+          name: Move private sql into place
+          command: |
+            ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
+            git clone --single-branch --branch main \
+                          git@github.com:mozilla/private-bigquery-etl.git \
+            ~/private-bigquery-etl
```
jklukas, line 360:
```suggestion
            git clone --single-branch --branch main \
                          git@github.com:mozilla/private-bigquery-etl.git \
                          ~/private-bigquery-etl
```

.circleci/config.yml
```diff
@@ -343,11 +343,22 @@ jobs:
...
+            cp -r ~/private-bigquery-etl/sql/. sql/
```
jklukas, line 361:
```suggestion
            cp -r ~/private-bigquery-etl/sql/* sql/
```

I'm unfamiliar with the syntax of having a trailing dot in a path. I would expect `*` but I'm open to learn new things here.

jklukas (approved):
r+wc

.circleci/config.yml
jklukas, line 361:
```suggestion
            rsync --archive --ignore-existing ~/private-bigquery-etl/sql/ sql/
```

We need rsync here in order to merge the private content into the existing `sql/` directory. The `--archive` option makes it recursively traverse the two directory trees, and the `--ignore-existing` makes it prefer public content if there are two files with identical paths.

jklukas (changes requested):
We need to make sure we're merging the two `sql/` directories together, which I believe requires rsync.

jklukas (approved):

PR #2118, opened 2021-06-11
[Bug 1713623] Add field descriptions to stable views

scholtzan:
https://bugzilla.mozilla.org/show_bug.cgi?id=1713623

### DIFF
--- a/bigquery_etl/view/generate_stable_views.py
+++ b/bigquery_etl/view/generate_stable_views.py
@@ -65,6 +65,10 @@ description: |-
   Clustering fields: `normalized_channel`, `sample_id`
 """
 
+VIEW_CREATE_REGEX = re.compile(
+    r"CREATE OR REPLACE VIEW\n\s*[^\s]+\s*\nAS", re.IGNORECASE
+)
+
 parser = ArgumentParser(description=__doc__)
 parser.add_argument(
     "--sql-dir", default="sql/", help="The path where generated SQL files are stored."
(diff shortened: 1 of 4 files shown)


bigquery_etl/view/generate_stable_views.py
```diff
@@ -222,6 +225,19 @@ def write_view_if_not_exists(target_project: str, sql_dir: Path, schema: SchemaF
         with metadata_file.open("w") as f:
             f.write(metadata_content)
 
+    # get view schema with descriptions
+    try:
+        regex = re.compile(r"CREATE OR REPLACE VIEW\n\s*[^\s]+\s*\nAS", re.IGNORECASE)
```
jklukas, line 230: Can we make this a named constant to avoid recompiling the regex on every view?

bigquery_etl/view/generate_stable_views.py
```diff
@@ -222,6 +225,19 @@ def write_view_if_not_exists(target_project: str, sql_dir: Path, schema: SchemaF
...
+        content = regex.sub("", target_file.read_text())
+        content += " WHERE DATE(submission_timestamp) = '2020-01-01'"
+        view_schema = Schema.from_query_file(target_file, content=content)
+
+        stable_table_schema = Schema.from_json({"fields": schema.schema})
+        view_schema.merge(stable_table_schema, add_missing_fields=False)
+        view_schema.to_yaml_file(target_dir / "schema.yaml")
+    except Exception as e:
```
jklukas, line 238: Is it possible to be more specific about what exceptions we're handling here? There are three instance of `except Exception` in this PR where we print the message and then move on, which could lead to confusing behavior where it's unclear to the user whether the exception is expected or not. And particularly for automated running, that could lead to deployment being labeled as success even if there's some issue that causes all the views to fail.

jklukas (approved):
r+wc

bigquery_etl/view/generate_stable_views.py
scholtzan: Added some more specific exceptions where it makes sense. 
In some cases, e.g. when adding descriptions to the schema, I don't think we ever want deployment to be blocked in case of errors (e.g. dry run times out, schemas are incompatible, permissions missing). Worst case descriptions might not be available, but they don't seem very critical.

PR #2119, opened 2021-06-11
Add VPN funnel query from redash

relud:
originally defined in: https://sql.telemetry.mozilla.org/queries/76251/source#190143

Moving it here so it can be re-used in Looker


sql/moz-fx-data-shared-prod/mozilla_vpn_derived/funnel_fxa_login_to_protected_v1/metadata.yaml
```diff
@@ -0,0 +1,13 @@
+friendly_name: Funnel FxA Login to Protected
+description: >
+  Funnel from FxA Login to Protected with 2-day time limit
+owners:
+  - dthorn@mozilla.com
+labels:
+  application: mozilla_vpn
+  schedule: daily
```
scholtzan, line 8: For some queries that replace the entire destination table we've added `incremental: false` to labels

scholtzan (approved):

PR #2120, opened 2021-06-11
Fix pricing_plan for apple subs in Mozilla VPN

relud:
also:
- expose user_id for joining devices table in Looker
- use stricter definition of vpn subscriptions in retention
- update schemas


sql/moz-fx-data-shared-prod/mozilla_vpn_derived/all_subscriptions_v1/query.sql
```diff
@@ -236,13 +247,14 @@ apple_iap_subscriptions AS (
...
-    IF(`interval` = "month" AND interval_count = 1, 499, NULL) AS plan_amount,
+    NULL AS plan_amount,
     CAST(NULL AS STRING) AS billing_scheme,
-    IF(`interval` = "month" AND interval_count = 1, "USD", NULL) AS plan_currency,
+    CAST(NULL AS STRING) AS plan_currency,
     `interval` AS plan_interval,
     interval_count AS plan_interval_count,
     CAST(NULL AS STRING) AS product_id,
     "Mozilla VPN" AS product_name,
+    CONCAT(interval_count, "-", `interval`, "-", "apple") AS pricing_plan,
```
relud, line 257: apple IAP subscriptions are available in different currencies and prices, and we don't have a way to connect that information with user accounts here, so we'll have to settle for values like `1-month-apple`, instead of `1-month-usd-4.99`.

scholtzan (approved):

PR #2127, opened 2021-06-15
Bug 1716074 - Derive search clients daily from clients daily

BenWu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1716074

Some basic validation shows this works and search count totals per engine/source match up.  Need to to some more testing


sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v8/query.sql
```diff
@@ -46,77 +46,9 @@ CREATE TEMP FUNCTION organicize_source_or_type(engine STRING, original STRING) A
   END
 );
 
-WITH overactive AS (
```
BenWu, line 49: Overactive filter is identical to the one in `clients_daily` so it's not needed anymore

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v8/query.sql
```diff
@@ -46,77 +46,9 @@ CREATE TEMP FUNCTION organicize_source_or_type(engine STRING, original STRING) A
...
-  FROM
-    telemetry.main_summary
-  WHERE
-    submission_date = @submission_date
-  GROUP BY
-    client_id
-  HAVING
-    COUNT(*) > 200000
-),
-client_map_sums AS (
```
BenWu, line 62: All these map sums are done in clients_daily identically

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
```diff
@@ -980,5 +1010,7 @@ SELECT
   map_sum_aggregates[OFFSET(29)].map AS scalar_parent_urlbar_picked_topsite_sum,
   map_sum_aggregates[OFFSET(30)].map AS scalar_parent_urlbar_picked_unknown_sum,
   map_sum_aggregates[OFFSET(31)].map AS scalar_parent_urlbar_picked_visiturl_sum,
+  map_sum_aggregates[OFFSET(32)].map AS search_with_ads,
+  map_sum_aggregates[OFFSET(33)].map AS ad_clicks,
```
BenWu, line 1014: I chose these names without `scalar_parent_...` to match the existing `search_with_ads_count_all` and `ad_clicks_count_all` columns

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v8/query.sql
```diff
@@ -313,35 +209,18 @@ windowed AS (
...
         client_id,
         submission_date,
         engine,
         source,
         type
-      ORDER BY
-        `timestamp` ASC
-      ROWS BETWEEN
-        UNBOUNDED PRECEDING
-        AND UNBOUNDED FOLLOWING
```
BenWu, line 328: I'm not sure what the purpose of this window frame.  I think this is the default behaviour unless I'm missing something

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v8/query.sql
```diff
@@ -188,38 +120,18 @@ augmented AS (
...
-    client_map_sums
-  USING
-    (client_id)
-  WINDOW
-    w1 AS (
-      PARTITION BY
-        client_id,
-        submission_date
-    )
+    mozdata.tmp.clients_daily_with_search
```
BenWu, line 127: TODO: fix

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v8/query.sql
jklukas: If I'm reading correctly here, `w1` is now only used for `SUM` aggregations where order is not important. As long as that is the case, then yes defining `w1` with simply a PARTITION BY will give equivalent results.

But, if were still doing ARRAY_AGG here, order would be significant, and we'd need to specify both ORDER BY and window frame explicitly.

From https://cloud.google.com/bigquery/docs/reference/standard-sql/analytic-function-concepts#def_window_spec

> If neither the ORDER BY clause nor window frame clause are present, the window frame includes all rows in that partition.

> For aggregate analytic functions, if the ORDER BY clause is present but the window frame clause is not, the following window frame clause is used by default:

> RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v8/query.sql
```diff
@@ -293,10 +192,7 @@ windowed AS (
     scalar_parent_urlbar_searchmode_topsites_urlbar_sum,
     scalar_parent_urlbar_searchmode_touchbar_sum,
     scalar_parent_urlbar_searchmode_typed_sum,
-    SAFE_SUBTRACT(
-      UNIX_DATE(DATE(SAFE.TIMESTAMP(subsession_start_date))),
-      profile_creation_date
-    ) AS profile_age_in_days,
+    profile_age_in_days,
     SUM(IF(type = 'organic', count, 0)) OVER w1 AS organic,
     SUM(IF(type = 'tagged-sap', count, 0)) OVER w1 AS tagged_sap,
```
jklukas, line 197: Do these need to be sums? I assume that the search count aggregation in clients_daily ends up giving us an array where each engine/source combination will be unique. So maybe we don't need to do any windowing or aggregation here at all?

jklukas (commented):
A few questions about the data shape here before final review.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v8/query.sql
jklukas: Based on my reading of `udf.search_counts_map_sum`, we'll indeed be guaranteed that the `flattened` CTE here will not have more than one row per `client_id, engine, source`, so we shouldn't need this aggregation step any more.

jklukas: Ben points out that we do engine normalization, etc. in this query, so the values here are not exactly what's in the array from clients_daily, and we're not necessarily guaranteed that two different values haven't been normalized to the same thing. So we do need aggregation.

BenWu:
Rebased on top of the clients daily only changes and dry run passes.  I'll revisit this after the break

BenWu:
This should be fine to move forward now.  DS is in the loop with these changes but I don't see this as something that needs to be shared more widely as it's an implementation detail and the impact of the change is mostly inconsequential.

jklukas (approved):
LGTM!

PR #2133, opened 2021-06-16
Add ios support to experiment aggregates queries

scholtzan:
Missed to add ios support to the aggregates queries. This data needs to be backfilled to 2021-06-07

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_aggregates_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_aggregates_v1/query.sql
@@ -52,6 +52,19 @@ fenix_all_events AS (
   FROM
     `moz-fx-data-shared-prod.org_mozilla_firefox.events`
 ),
+ios_all_events AS (
+  SELECT
+    submission_timestamp,
+    events
+  FROM
+    `moz-fx-data-shared-prod.org_mozilla_ios_firefox.events`
+  UNION ALL
+  SELECT
+    submission_timestamp,
+    events
+  FROM
+    `moz-fx-data-shared-prod.org_mozilla_ios_firefoxbeta.events`
+),
 fenix AS (
   SELECT
     submission_timestamp AS `timestamp`,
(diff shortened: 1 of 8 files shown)


sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_aggregates_v1/query.sql
```diff
@@ -52,6 +52,19 @@ fenix_all_events AS (
...
+    submission_timestamp,
+    events
+  FROM
+    `moz-fx-data-shared-prod.org_mozilla_ios_firefox.events`
+  UNION ALL
+  SELECT
+    submission_timestamp,
+    events
+  FROM
+    `moz-fx-data-shared-prod.org_mozilla_ios_firefoxbeta.events`
```
jklukas, line 66: Should we also be including `org_mozilla_ios_fennec`?

sql/moz-fx-data-shared-prod/telemetry_derived/experiment_search_aggregates_v1/query.sql
```diff
@@ -135,6 +135,43 @@ fenix AS (
...
+    ) AS search_count,
+  FROM
+    `moz-fx-data-shared-prod.org_mozilla_ios_firefox_stable.metrics_v1`
+  LEFT JOIN
+    UNNEST(ping_info.experiments) AS experiment
+  GROUP BY
+    submission_timestamp,
+    experiment,
+    branch
+  UNION ALL
```
jklukas, line 156: In `experiment_enrollment_aggregates`, we union at the ping level and then aggregate in a subsequent step, which reduces code duplication for the aggregation. Is there a reason that we can't apply the same treatment here?

jklukas (approved):
r+wc

sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_aggregates_v1/query.sql
scholtzan: Yeah, I guess we should

sql/moz-fx-data-shared-prod/telemetry_derived/experiment_search_aggregates_v1/query.sql
scholtzan: The structure of `metrics` is different, for example, for the different fenix pings. So we can't directly union them but instead have to select the specific metrics we care about which means we have to do some (duplicated) `UNNEST`ing here.

PR #2134, opened 2021-06-16
Airflow task SLA and execution_timeout

scholtzan:
Based on proposal https://docs.google.com/document/d/1V8FwVH_oZonh-SLKRbghkzglofRlec7S51QwIIsLZbQ/edit#

I think it makes sense to set an `sla` for tasks but also ensure they time out after a while, otherwise we might end up with a lot of tasks retrying taking up all the slots if not detected during triage.

### DIFF
--- a/bigquery_etl/query_scheduling/task.py
+++ b/bigquery_etl/query_scheduling/task.py
@@ -160,6 +160,8 @@ class Task:
     retry_delay: Optional[str] = attr.ib(None)
     retries: Optional[int] = attr.ib(None)
     email_on_retry: Optional[bool] = attr.ib(None)
+    sla: Optional[str] = attr.ib(None)  # todo: set to 3h
+    execution_timeout: Optional[str] = attr.ib(None)  # todo: set to 24h
 
     @owner.validator
     def validate_owner(self, attribute, value):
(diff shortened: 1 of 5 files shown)


bigquery_etl/query_scheduling/templates/public_data_json_airflow_dag.j2
```diff
@@ -45,6 +45,9 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
         + ["--parameter={{ task.date_partition_parameter }}:DATE:{% raw %}{{ds}}{% endraw -%}"]
         {%- endif -%},
         image=docker_image,
+        sla=datetime.timedelta(minutes=5),
+        execution_timeout=datetime.timedelta(minutes=10),
+        email_on_retry=False,
```
scholtzan, line 50: I'm only changing the `bqetl_public_data_json` DAG here for testing. If everything works out we can update the remaining DAGs

bigquery_etl/query_scheduling/task.py
```diff
@@ -160,6 +160,8 @@ class Task:
     retry_delay: Optional[str] = attr.ib(None)
     retries: Optional[int] = attr.ib(None)
     email_on_retry: Optional[bool] = attr.ib(None)
+    sla: Optional[str] = attr.ib(None)  # todo: set to 3h
+    execution_timeout: Optional[str] = attr.ib(None)  # todo: set to 24h
```
scholtzan, line 164: If the testing is successful and we decide to go for setting `sla` and `execution_timeout` for the other tasks, then we should set them to the default values in the comments

bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -85,6 +91,12 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
             {%+ if task.email_on_retry != None -%}
             email_on_retry={{ task.email_on_retry }},
             {%+ endif -%}
+            {%+ if task.sla != None -%}
```
jklukas, line 94:
```suggestion
            {%+ if task.sla is not none -%}
```

Jinja has a `none` test: https://jinja.palletsprojects.com/en/2.11.x/templates/#none

My intuition says we should be using that rather than `!= None`, but I don't fully understand the implications in the Jinja context.

bigquery_etl/query_scheduling/templates/public_data_json_airflow_dag.j2
jklukas: +1 to this plan.

jklukas (approved):
Looks good. I'm very interested to see how this behaves.

PR #2139, opened 2021-06-21
Shredder fixes for Rally on Glean

akkomar:
This fixes two issues:
* automatic discovery of derived and analysis tables with nested `rally_id`
* incorrect handling of `rally_zero_one` study (Shredder query failed when it tried to access `metrics.uuid.rally_id` in `rally_zero_one_stable` table)


whd (approved):
Does the rally-zero-one special case work with the tables in the analysis datasets?

akkomar:
> Does the rally-zero-one special case work with the tables in the analysis datasets?

Yes, it does now.

PR #2141, opened 2021-06-21
Fix vpn retention table to correctly handle periods other than 1 month

relud:
previous definition of retention required an invoice to be paid in each month to count as renewed. This definition was insufficient because invoices aren't available for IAP subs, "month" is a timezone sensitive concept and IAP subs use `America/Los_Angeles`, and for non-monthly subscriptions (e.g. 6 month and 1 year subs) we don't have an invoice for each month.

New definition is based on localized start/end datetime and grace period in days, and is defined in a UDF for ease of testing.

### DIFF
--- a/sql/moz-fx-data-shared-prod/mozilla_vpn_derived/all_subscriptions_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/mozilla_vpn_derived/all_subscriptions_v1/query.sql
@@ -234,7 +234,7 @@ apple_iap_subscriptions AS (
     CAST(NULL AS TIMESTAMP) AS canceled_at,
     CAST(NULL AS TIMESTAMP) AS cancel_at,
     CAST(NULL AS BOOL) AS cancel_at_period_end,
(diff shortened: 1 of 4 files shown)


sql/moz-fx-data-shared-prod/mozilla_vpn_derived/all_subscriptions_v1/query.sql
```diff
@@ -234,7 +234,7 @@ apple_iap_subscriptions AS (
     CAST(NULL AS TIMESTAMP) AS canceled_at,
     CAST(NULL AS TIMESTAMP) AS cancel_at,
     CAST(NULL AS BOOL) AS cancel_at_period_end,
-    IF(end_time < CURRENT_DATE, end_time, NULL) AS ended_at,
+    IF(end_time < TIMESTAMP(CURRENT_DATE), end_time, NULL) AS ended_at,
```
relud, line 237: this fixes a bug introduced in #2142 where you can't use `<` between `TIMESTAMP` and `DATE`

scholtzan (approved):

PR #2142, opened 2021-06-21
Expose full start and end timestamp for IAP subscriptions

relud:
and include a timezone field, because the definition of a month is timezone sensitive

### DIFF
--- a/sql/mozfun/iap/scrub_apple_receipt/udf.sql
+++ b/sql/mozfun/iap/scrub_apple_receipt/udf.sql
@@ -2,114 +2,37 @@ CREATE OR REPLACE FUNCTION iap.scrub_apple_receipt(apple_receipt ANY TYPE)
 RETURNS STRUCT<
   environment STRING,
   active_periods ARRAY<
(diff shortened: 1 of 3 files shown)


sql/mozfun/iap/scrub_apple_receipt/udf.sql
```diff
@@ -2,114 +2,37 @@ CREATE OR REPLACE FUNCTION iap.scrub_apple_receipt(apple_receipt ANY TYPE)
...
       FROM
-        stage_4
+        UNNEST(apple_receipt.latest_receipt_info)
       WHERE
-        start_date IS NOT NULL
+        is_trial_period = "false"
+      ORDER BY
+        expires_date_ms DESC
+      LIMIT
+        1
```
relud, line 35: the apple receipt [excludes old transactions](https://github.com/mozilla-services/guardian-website/blob/c9a3d9def4c2643ebd177f1d1afa9963bdf87801/server/lib/apple.ts#L124), so this must assume that the subscription is continuous from `original_purchase_date_ms` through `expires_date_ms`

scholtzan (approved):

PR #2143, opened 2021-06-21
Github Action for pushing forks to upstream

scholtzan:
This wraps the [git-push-fork-to-upstream-branch](https://github.com/jklukas/git-push-fork-to-upstream-branch) script into a Github Action. The CircleCI config has been updated to have a failing tasks for forks that will show instructions of how to trigger the Github Actions so that integration tests will get executed for forks.

I successfully tested this approach in this repository: https://github.com/scholtzan/gh-actions-test

The main advantage here is that we don't have to share secrets with forks, however it might be a little unintuitive.

Another option that came to mind that we could try:
* All environment variables (our secrets) get moved into a restricted context
* The setup is essentially the same as described here: https://github.com/mozilla/glean.js/pull/422#issuecomment-863411793
* When sharing secrets also SSH keys are being shared, so we'd need to remove deploy keys from the repository settings and move them to environment variables managed by the restricted context
    * The CircleCI config would need to be rewritten to use the SSH keys from the environment variable

This seems a little less safe since it would require sharing secrets with forks (although if set up correctly it should not pose a security problem). Also not sure how much effort rewriting the CircleCI config would be.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -372,16 +372,23 @@ jobs:
       - checkout
       - run:
           name: Manually trigger integration tests for fork
+          # yamllint disable rule:line-length
           command: |
+            apt update
+            apt install jq -y
+
+            CIRCLE_PR_BRANCH=`curl -s https://api.github.com/repos/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/pulls/${CIRCLE_PR_NUMBER} | jq -r '.head.label'`
+
             echo "Integration tests for this fork need to be triggered manually"
             echo "Users with write access to the repository can trigger" \
               " integration tests by visiting: "
             echo "https://github.com/mozilla/bigquery-etl/actions/workflows" \
               "/push-to-upstream.yml".
-            echo "Trigger via 'Run workflow' and provide the " \
-              "<username>:<branch>" of this PR as parameters."
+            echo "Trigger via 'Run workflow' and provide " \
+              '$CIRCLE_PR_BRANCH' as parameter."
 
             exit 1
+          # yamllint enable rule:line-length
 
 workflows:
   version: 2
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -366,11 +366,31 @@ jobs:
...
+            echo "Trigger via 'Run workflow' and provide the " \
+              "<username>:<branch>" of this PR as parameters."
+
+            exit 1
 
 workflows:
   version: 2
   build:
     jobs: &build_jobs
+      - fail-on-fork:
```
jklukas, line 390:
```suggestion
      - manual-trigger-required-for-fork:
```

The name of this job and the status are the only things that show up on the PR itself, so I think we should name this task as explicitly as possible to indicate that a follow-up action is required.

.circleci/config.yml
```diff
@@ -366,11 +366,31 @@ jobs:
...
+      - run:
+          name: Manually trigger integration tests for fork
+          command: |
+            echo "Integration tests for this fork need to be triggered manually"
+            echo "Users with write access to the repository can trigger" \
+              " integration tests by visiting: "
+            echo "https://github.com/mozilla/bigquery-etl/actions/workflows" \
+              "/push-to-upstream.yml".
+            echo "Trigger via 'Run workflow' and provide the " \
+              "<username>:<branch>" of this PR as parameters."
```
jklukas, line 382: Do we have sufficient info in env vars here that we can inject the actual values into this message?

jklukas (approved):
r+wc

Don't hesitate to go ahead and merge this so that we can do some more concrete testing of this workflow, and then we can PR improvements as needed.

.circleci/config.yml
scholtzan: Ah yes that is possible! Updated the task and included the branch label in the output

scholtzan:
this repository doesn't seem to support the `webfactory/ssh-agent@v0.5.1` plugin:

```
webfactory/ssh-agent@v0.5.1 is not allowed to be used in mozilla/bigquery-etl. Actions in this workflow must be: within a repository that belongs to your Enterprise account, created by GitHub or match the following: !/mozilla/**, !mozilla/**, ./**, aws-actions/*, docker/*, pypa/gh-action-pypi-publish@release/v1.4.2.
```

PR #2155, opened 2021-06-23
Use ExternalTaskCompletedSensor for bqetl_public_data_json DAG

scholtzan:
To test the new [ExternalTaskCompletedSensor](https://github.com/mozilla/telemetry-airflow/pull/1330). Only used in the `bqetl_public_data_json` DAG for now. If it seems like it's working we can update the remaining DAGs after the wellness week


jklukas (approved):

scholtzan:
Just did some testing and everything worked like expected

PR #2151, opened 2021-06-23
Bug 1716074 - Add search related probes to clients_daily

BenWu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1716074

This is the clients_daily changes from https://github.com/mozilla/bigquery-etl/pull/2127 so clients daily can be backfilled separately first

PR #2156, opened 2021-06-24
Disambiguate ad_clicks in addons_daily query

jklukas:
The addons_daily_v1 query failed last night with an error about `ad_clicks`
being an ambiguous field name. See
[Airflow logs](https://workflow.telemetry.mozilla.org/log?execution_date=2021-06-23T04%3A00%3A00%2B00%3A00&task_id=telemetry_derived__addons_daily__v1&dag_id=bqetl_addons).

This is due to https://github.com/mozilla/bigquery-etl/pull/2151 which added
a field called ad_clicks to clients_daily and downstream tables. That new field
is present in a join.

This PR disambiguates the reference by specifying which table to pull it from.


BenWu (approved):
Looks good.  This query is in the dryrun skip list but it doesn't look like it should be.  This is possibly from back when search tables were restricted so we can take it out of the list

BenWu (approved):
👍  on the dryrun changes

PR #2157, opened 2021-06-24
Fix Github workflow

scholtzan:
Fixes the `/home/runner/work/_temp/a76e427e-af4c-4f69-8779-c68e2869d4ad.sh: line 52: /home/runner/.ssh/known_hosts: No such file or directory` error

### DIFF
--- a/.github/workflows/push-to-upstream.yml
+++ b/.github/workflows/push-to-upstream.yml
@@ -18,6 +18,7 @@ jobs:
       run: |
         eval "$(ssh-agent -s)"
         ssh-add - <<< "${{ secrets.GH_ACTION_SSH_KEY }}"
+        mkdir -p ~/.ssh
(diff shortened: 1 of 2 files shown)


.github/workflows/push-to-upstream.yml
```diff
@@ -18,7 +18,6 @@ jobs:
       run: |
         eval "$(ssh-agent -s)"
         ssh-add - <<< "${{ secrets.GH_ACTION_SSH_KEY }}"
-        ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
```
jklukas, line 21: We can probably make this work by adding `mkdir -p ~/.ssh` before running this. I don't know whether this line is necessary or not.

jklukas (approved):
r+wc

PR #2159, opened 2021-06-24
Update docs and add pull request template

scholtzan:
Docs for how to use the workflow + PR template. I'm not sure if there should be more items on the PR checklist.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -373,9 +373,11 @@ jobs:
         run:
           name: Early return if this build is running on upstream
           command: |
-            if [ "$CIRCLE_PR_NUMBER" ]; then
-              echo "Build on upstream"
+            if [ -n "$CIRCLE_PR_NUMBER" ]; then
+              echo "Build on fork"
+            else
               circleci step halt
+              echo "Build on upstream"
             fi
       - checkout
       - run:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -400,10 +402,7 @@ workflows:
   version: 2
   build:
     jobs: &build_jobs
-      - manual-trigger-required-for-fork:
-          filters:
-            branches:
-              only: /^pull\/.*$/
```
scholtzan, line 406: This needs to be re-executed when pushing to upstream otherwise it'll stay in failed state

.circleci/config.yml
```diff
@@ -373,9 +373,11 @@ jobs:
         run:
           name: Early return if this build is running on upstream
           command: |
-            if [ "$CIRCLE_PR_NUMBER" ]; then
-              echo "Build on upstream"
+            if [ -n "$CIRCLE_PR_NUMBER" ]; then
+              echo "Build on fork"
+            else
               circleci step halt
+              echo "Build on upstream"
```
jklukas, line 380: I think `circleci step halt` will immediately exit, so we'll need to put this message before the halt.

And it looks like modern invocation is `circleci-agent` vs. `circleci` but I don't understand the difference. https://circleci.com/docs/2.0/configuration-reference/#ending-a-job-from-within-a-step

jklukas (approved):

PR #2173, opened 2021-07-07
Bug 1719188  - Fix overflows by filtering out large histogram values

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1719188)

The airflow dag for glam-fenix has been failing consistently since 2021-06-30. There a few clients that have histograms with particularly large values, where `log2(value) ~ 62.8 bits`. With only a few of these values, the histogram aggregates are bound to overflow.

```sql
select
    client_id,
    agg.metric,
    agg.metric_type,
    val.key,
    val.value
from glam_etl.org_mozilla_firefox__clients_daily_histogram_aggregates_metrics_v1,
unnest(histogram_aggregates) agg,
unnest(agg.value) val
where _PARTITIONDATE = "2021-06-29"
order by val.value desc 
limit 1000
```

image

This filters out values greater than 2^40, which should suffice for individual pings. We should be able to aggregate 2^24 (16m) pings before running into overflows.

It's possible that this doesn't fix the problem if there is already a client that is already about to overflow.

Script was tested by running `./script/glam/test/test_glean_org_mozilla_fenix_glam_nightly` to completion.

Checklist for reviewer:


sql/glam-fenix-dev/glam_etl/org_mozilla_fenix__clients_daily_scalar_aggregates_startup_timeline_v1/query.sql
```diff
@@ -63,6 +63,111 @@ unlabeled_metrics AS (
         'sum',
         sum(CAST(metrics.counter.startup_timeline_clock_ticks_per_second AS INT64))
       ),
+      (
```
emtwo, line 66: I just wanted to confirm I remember this correctly. All these generated `query.sql` files that are being updated in this commit are just added for reference, right? But in the actual airflow job they are produced on the fly before running?

emtwo (approved):
lgtm!

sql/glam-fenix-dev/glam_etl/org_mozilla_fenix__clients_daily_scalar_aggregates_startup_timeline_v1/query.sql
acmiyaguchi: Yep, these checked in files are mostly for being able to diff changes. The actual jobs will generate the sql from scratch to get the new probe names.

PR #2174, opened 2021-07-08
Push private generated SQL to private repository

akkomar:
Fixes https://github.com/mozilla/private-bigquery-etl/issues/10
I have manually created the [private branch](https://github.com/mozilla/private-bigquery-etl/tree/private-generated-sql).

Checklist for reviewer:


.circleci/config.yml
```diff
@@ -440,6 +492,14 @@ workflows:
               only: main
             tags:
               only: /.*/
+      - private-generate-sql
```
jklukas, line 495:
```suggestion
      # The following "private" jobs are basically clones of the public jobs for generate-sql,
      # deploy, and push-generated-sql, except that they pull in some additional content from
      # an internal Mozilla repository for cases where ETL code cannot be public.
      # Although the CI logic is consolidated in this public repository, note that we are both pulling
      # from the internal repository and pushing generated results back to a branch on that internal
      # repository, which may be initially surprising.
      - private-generate-sql
```

.circleci/config.yml
```diff
@@ -337,28 +337,80 @@ jobs:
...
           command: |
             ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
             git clone --single-branch --branch main \
                           git@github.com:mozilla/private-bigquery-etl.git \
                           ~/private-bigquery-etl
             rsync --archive --ignore-existing ~/private-bigquery-etl/sql/ sql/
+      - run:
+          name: Generate SQL content
+          command: |
+            mkdir /tmp/private-generated-sql
```
jklukas, line 367: It would be nice to be able to factor out the logic for the generate-sql and push-generated-sql jobs so that there's less duplication with the private jobs, but my gut is telling me the duplication here is likely more maintainable than the code would be refactored, since we'd have to figure out how to templatize the paths involved.

jklukas (approved):
r+wc

A few small changes needed, and I suggested some code comments to better document the approach here.

.circleci/config.yml
akkomar: Yes, I agree, we're getting some duplication here and in some parts of private repo's config. In the future we can think of using inline orbs to fix this.

PR #2177, opened 2021-07-09
Fix private SQL generation

akkomar:
This fixes CircleCI workflow that is failing[1] because both `generate-sql` and `private-generate-sql` jobs are persisting `sql` directory into the workspace. In order to keep these jobs concurrent, `private-...` branch of the workflow will use `/tmp/generated-sql/private-sql/`.

[1] https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/10084/workflows/a835c08d-7d06-44cc-82a1-31f05795f701/jobs/64214/parallel-runs/0/steps/0-103

Checklist for reviewer:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -364,22 +364,22 @@ jobs:
       - run:
           name: Generate SQL content
           command: |
-            mkdir /tmp/private-generated-sql
-            cp -r sql/ /tmp/private-generated-sql/sql
+            mkdir /tmp/generated-sql
+            cp -r sql/ /tmp/generated-sql/private-sql
             ./bqetl bootstrap
             # Don't depend on dry run for PRs
             PATH="venv/bin:$PATH" ./script/generate_sql \
-              --sql-dir /tmp/private-generated-sql/sql/ \
+              --sql-dir /tmp/generated-sql/private-sql/ \
               --target-project moz-fx-data-shared-prod
       - persist_to_workspace:
-          root: /tmp/private-generated-sql
+          root: /tmp/generated-sql
           paths:
-            - sql
+            - private-sql
   push-private-generated-sql:
     docker: *docker
     steps:
       - attach_workspace:
-          at: /tmp/private-generated-sql
+          at: /tmp/generated-sql
       - add_ssh_keys:
           fingerprints: "cf:d6:25:9a:ee:26:66:39:c8:cc:48:f6:bb:3e:34:68"
       - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -364,22 +364,22 @@ jobs:
       - run:
           name: Generate SQL content
           command: |
-            mkdir /tmp/private-generated-sql
-            cp -r sql/ /tmp/private-generated-sql/sql
+            mkdir /tmp/generated-sql
+            cp -r sql/ /tmp/generated-sql/private-sql
```
jklukas, line 368: I'd prefer to keep the public and private paths more parallel, so that we aren't introducing another layer of terminology here.

How about this:

- We use `/tmp/workspace` as the root
- Public sql ends up under `/tmp/workspace/generated-sql/sql`
- Private sql ends up under `/tmp/workspace/private-generated-sql/sql`

That way the paths we're using here mirror the branches where the content will end up.

akkomar: Sounds good!

.circleci/config.yml
```diff
@@ -274,22 +274,22 @@ jobs:
...
             PATH="venv/bin:$PATH" ./script/generate_sql \
-              --sql-dir /tmp/generated-sql/sql/ \
+              --sql-dir /tmp/workspace/generated-sql/sql/ \
               --target-project moz-fx-data-shared-prod
       - persist_to_workspace:
-          root: /tmp/generated-sql
+          root: /tmp/workspace
           paths:
-            - sql
+            - generated-sql/sql
```
jklukas, line 287:
```suggestion
            - generated-sql
```

I think it's slightly clearer if we specify the base directory rather than the subdir here.

.circleci/config.yml
```diff
@@ -364,22 +364,22 @@ jobs:
...
             PATH="venv/bin:$PATH" ./script/generate_sql \
-              --sql-dir /tmp/private-generated-sql/sql/ \
+              --sql-dir /tmp/workspace/private-generated-sql/sql/ \
               --target-project moz-fx-data-shared-prod
       - persist_to_workspace:
-          root: /tmp/private-generated-sql
+          root: /tmp/workspace
           paths:
-            - sql
+            - private-generated-sql/sql
```
jklukas, line 377:
```suggestion
            - private-generated-sql
```

jklukas (approved):
Looks great!

One final suggestion which you can choose to ignore if you disagree.

Issue #2158, opened 2021-06-24
Separate CI task for running SQL/UDF tests


scholtzan:
Currently, SQL and UDF tests are executed as part of the `build` task. SQL and UDF tests will fail for forks, which means that the entire `build` task will show up as failed. Instead, we should create a separate CI tests that is skipped for forks.

PR #2178, opened 2021-07-09
Add separate CI step for SQL and routine tests

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/2158

The `build` CI tasks is currently failing for PRs from forks since some of the tests require GCP credentials. This moves the SQL and routine integration tests into a separate CI step that will be skipped for forks, but executed once the changes have been pushed to an upstream branch.

Checklist for reviewer:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,16 @@ jobs:
           command: PATH="venv/bin:$PATH" yamllint -c .yamllint.yaml .
       - run:
           name: PyTest with linters
-          command: PATH="venv/bin:$PATH" script/entrypoint
+          # yamllint disable rule:line-length
+          command: |
+            if [ -n "$CIRCLE_PR_NUMBER" ]; then
+              echo "Cannot pass creds to forked PRs," \
+                "so skipping routine and SQL tests"
+              PATH="venv/bin:$PATH" script/entrypoint -m "not (routine or sql or integration or java)"
+            else
+              PATH="venv/bin:$PATH" script/entrypoint
+            fi
+          # yamllint enable rule:line-length
       - save_cache:
           paths:
             - venv/
@@ -135,10 +144,10 @@ jobs:
           # GOOGLE_APPLICATION_CREDENTIALS
           # and use a file in that location for credentials if present;
           # See https://cloud.google.com/docs/authentication/production
-          # yamllint disable rule:line-length
           command: |
-            PATH="venv/bin:$PATH" script/entrypoint -m 'routine or sql or integration or java' --numprocesses 20
-          # yamllint enable rule:line-length
+            export GOOGLE_APPLICATION_CREDENTIALS="/tmp/gcp.json"
+            echo "$GCLOUD_SERVICE_KEY" > "$GOOGLE_APPLICATION_CREDENTIALS"
+            PATH="venv/bin:$PATH" script/entrypoint -m 'integration or java'
       - save_cache:
           paths:
             - ~/.m2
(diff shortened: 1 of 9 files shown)


script/entrypoint
```diff
@@ -19,7 +19,8 @@ fi
 
 if [ "$#" = 0 ]; then
     # Default to 20 workers because most of the tests are io-bound
-    exec pytest --black --flake8 --mypy-ignore-missing-imports --pydocstyle --numprocesses 20
+    # Ignore running routine and SQL tests since they require GCP access
+    exec pytest --black --flake8 --mypy-ignore-missing-imports --pydocstyle --numprocesses 20 -m "not sql" -m "not routine"
```
scholtzan, line 23: We specified custom markers for [routine](https://github.com/mozilla/bigquery-etl/blob/374325adc7abf87ea179eacd2fde5ecc6c1f6779/bigquery_etl/pytest_plugin/routine.py#L38) and [SQL tests](https://github.com/mozilla/bigquery-etl/blob/374325adc7abf87ea179eacd2fde5ecc6c1f6779/bigquery_etl/pytest_plugin/sql.py#L31). I _think_ it should be fine ignoring them here. The only problem I see is that these tests will also get ignored when running tests locally.

script/entrypoint
```diff
@@ -19,7 +19,8 @@ fi
 
 if [ "$#" = 0 ]; then
     # Default to 20 workers because most of the tests are io-bound
-    exec pytest --black --flake8 --mypy-ignore-missing-imports --pydocstyle --numprocesses 20
+    # Ignore running routine and SQL tests since they require GCP access
+    exec pytest --black --flake8 --mypy-ignore-missing-imports --pydocstyle --numprocesses 20 -m "not (routine or sql or integration or java)"
```
scholtzan, line 23: We specified custom markers for [routine](https://github.com/mozilla/bigquery-etl/blob/374325adc7abf87ea179eacd2fde5ecc6c1f6779/bigquery_etl/pytest_plugin/routine.py#L38) and [SQL tests](https://github.com/mozilla/bigquery-etl/blob/374325adc7abf87ea179eacd2fde5ecc6c1f6779/bigquery_etl/pytest_plugin/sql.py#L31). I think it should be fine ignoring them here. The only problem I see is that these tests will also get ignored when running tests locally.

jklukas:
This looks reasonable, but one test case still failed in the `integration` job due to permissions on a js resource in gcs. Do you understand what's going on there?

script/entrypoint
jklukas: This does seem like it will require at least an update to the testing docs: https://github.com/mozilla/bigquery-etl/blob/main/tests/README.md#how-to-run-tests

What's the invocation we would need to run a udf test now?

scholtzan:
This is very strange. :/ I'll see if I can figure out why this is happening

.circleci/config.yml
```diff
@@ -34,7 +34,16 @@ jobs:
...
+          # yamllint disable rule:line-length
+          command: |
+            if [ -n "$CIRCLE_PR_NUMBER" ]; then
+              echo "Cannot pass creds to forked PRs," \
+                "so skipping routine and SQL tests"
+              PATH="venv/bin:$PATH" script/entrypoint -m "not (routine or sql or integration or java)"
+            else
+              PATH="venv/bin:$PATH" script/entrypoint
+            fi
+          # yamllint enable rule:line-length
```
scholtzan, line 46: I figured out the problem: we run integration tests in the `bigquery-etl-integration-test` project which does not have permissions to access the `moz-fx-data-circleci-tests-bigquery-etl` storage bucket. The easiest workaround for now is to check for forks in the `build` step and only run SQL/routine tests for non-forks. This way `entrypoint` and how we run tests doesn't need to change.

scholtzan:
I checked that tests are succeeding as expected on the fork and then after triggering the Github action. So this should be ready for another look

jklukas (approved):
Looks good. And I verified that marking extra cases as integration doesn't interfere with the documented local workflows for testing i.e. udfs.

PR #2186, opened 2021-07-13
Bug 1719698 - Fxa users first seen

fbertsch:
Checklist for reviewer:

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/init.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/init.sql
@@ -1,0 +1,17 @@
+CREATE TABLE
+  `firefox_accounts_derived.fxa_users_first_seen_v1`
+PARTITION BY
+  first_seen_date
+CLUSTER BY
+  user_id
+AS
+SELECT
+  user_id,
+  DATE(MIN(submission_timestamp)) AS first_seen_date,
+  ARRAY_AGG(DISTINCT service) AS services_used,
+FROM
+  firefox_accounts.fxa_all_events
+WHERE
+  submission_timestamp > '2010-01-01'
+GROUP BY
+  user_id
--- a/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/init.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/init.sql
@@ -8,7 +8,7 @@ AS
 SELECT
   user_id,
   DATE(MIN(submission_timestamp)) AS first_seen_date,
-  ARRAY_AGG(DISTINCT service) AS services_used,
+  ARRAY_AGG(DISTINCT service ORDER BY service) AS services_used,
 FROM
   firefox_accounts.fxa_all_events
 WHERE
--- a/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/init.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/init.sql
@@ -8,7 +8,7 @@ AS
 SELECT
   user_id,
   DATE(MIN(submission_timestamp)) AS first_seen_date,
-  ARRAY_AGG(DISTINCT service ORDER BY service) AS services_used,
+  ARRAY_AGG(DISTINCT service IGNORE NULLS ORDER BY service) AS services_used,
 FROM
   firefox_accounts.fxa_all_events
 WHERE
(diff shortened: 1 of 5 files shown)


sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/init.sql
```diff
@@ -0,0 +1,17 @@
...
+  `firefox_accounts_derived.fxa_users_first_seen_v1`
+PARTITION BY
+  first_seen_date
+CLUSTER BY
+  user_id
+AS
+SELECT
+  user_id,
+  DATE(MIN(submission_timestamp)) AS first_seen_date,
+  ARRAY_AGG(DISTINCT service) AS services_used,
```
jklukas, line 11:
```suggestion
  ARRAY_AGG(DISTINCT service ORDER BY service) AS services_used,
```

It's probably good to make the ordering here deterministic.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/query.sql
```diff
@@ -0,0 +1,36 @@
...
+  COALESCE(_previous.first_seen_date, @submission_date) AS first_seen_date,
+  ARRAY(
+    SELECT DISTINCT
+      service
+    FROM
+      UNNEST(
+        ARRAY_CONCAT(COALESCE(_previous.services_used, []), COALESCE(_current.services_used, []))
+      ) AS service
+    WHERE
+      service IS NOT NULL
```
jklukas, line 29:
```suggestion
      service IS NOT NULL
    ORDER BY service
```

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/query.sql
```diff
@@ -0,0 +1,36 @@
...
+  WHERE
+    DATE(submission_timestamp) = @submission_date
+  GROUP BY
+    user_id
+),
+_previous AS (
+  SELECT
+    *
+  FROM
+    firefox_accounts_derived.fxa_users_first_seen_v1
```
jklukas, line 16:
```suggestion
    firefox_accounts_derived.fxa_users_first_seen_v1
  WHERE
    DATE(submission_timestamp) < @submission_date
```

This is important in case we need to do a backfill of a few days. We don't want more recent (potentially incorrect) data to influence the calculation. See https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v1/query.sql#L11-L21

jklukas (changes requested):
Left suggestions for two additional logical details here.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/query.sql
fbertsch: Hmm, this logic could break the one-row-per-client invariant. That would happen if the user is skipped on their first_seen date X, gets a row on day X+1, and we backfill just day X: they would be present for days X and X+1. My current logic would just cause the client to be present for day X+1, while what we want is the client to only be present for day X.

fbertsch: Would you oppose leaving this logic out, and changing line 20 to instead be `LEAST(_previous.first_seen_date, @submission_date)`?

jklukas: > and we backfill just day X: they would be present for days X and X+1. My current logic would just cause the client to be present for day X+1, while what we want is the client to only be present for day X.

We should never do this. Since this query depends on past, any backfill operation would have to be for day X and all following days.

fbertsch: Good call - forgot this is overwriting the entire table and not just the returned partitions.

fbertsch: @jklukas RFAL

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/query.sql
```diff
@@ -0,0 +1,40 @@
...
+  WHERE
+    first_seen_date < @submission_date
```
jklukas, line 18:
```suggestion
    -- If reprocessing scenarios, we must always backfill from the first affected date
    -- all the way to the present; to enforce that, we explicitly drop any data after
    -- the target @submission_date
    first_seen_date < @submission_date
```

This is surprising enough that I think it warrants commentary.

jklukas (approved):
r+wc

irrationalagent:
I was able to verify that the main values of `service` and `event_type` that I was expecting are present in the underlying `fxa_all_events` table, so that should lead to the correct service array being present for each user_id.

PR #2187, opened 2021-07-13
[DS-1643] iProspect data import V2

scholtzan:
See https://mozilla-hub.atlassian.net/browse/DS-1643?focusedCommentId=442361

Su and I went over the iProspect data requirements. This PR implements the following changes:
* Add a `iprospect.adspend_raw_v1` dataset where each partition represents the CSv data uploaded for that specific date. This table will be used for debugging and deriving data for `iprospect.adspend_v1`
* Add a `iprospect.adspend_meta_v1` dataset. The iProspect data we receive is mutable. The CSV files contain data for the past 30 days. Some historic data can change in more recent CSV reports due to reporting lags from vendors. We want to keep track of some specific changes in this dataset.
* Update `iprospect.adspend_v1` dataset to always update the most recent 30 days instead of just the most recent day.

Checklist for reviewer:


sql/moz-fx-data-marketing-prod/iprospect/adspend_meta_v1/query.sql
```diff
@@ -0,0 +1,58 @@
+SELECT
```
scholtzan, line 1: I'm open to renaming these fields if there are better names.

sql/moz-fx-data-marketing-prod/iprospect/adspend_meta_v1/query.sql
```diff
@@ -0,0 +1,58 @@
...
+  COUNT(DISTINCT today.date) = 0 AS has_missing_data,
+  COUNTIF(yesterday.fetch_ad_name IS NULL AND yesterday.date < @submission_date) AS new_row_count,
+  COUNTIF(
+    today.client_cost != yesterday.client_cost
+    AND yesterday.client_cost IS NOT NULL
+    AND today.client_cost > 0
+  ) AS client_cost_nonzero_changed_count,
+  COUNTIF(
+    today.imps_vendor != yesterday.imps_vendor
+    AND yesterday.imps_vendor IS NOT NULL
```
acmiyaguchi, line 13: What is imps?

sql/moz-fx-data-marketing-prod/iprospect/adspend_v1/metadata.yaml
```diff
@@ -0,0 +1,18 @@
+friendly_name: Adspend
+description: >
+  Import detailed adspend iProspect CSV data from
+  moz-fx-data-marketing-prod-iprospect storage bucket.
+labels:
+  incremental: true
+owners:
+  - ascholtz@mozilla.com
+scheduling:
```
acmiyaguchi, line 9: A depends on past condition, maybe? It looks like data from newer dates should be prioritized over older values. On the other hand, a missed day doesn't block the next day since there is a 30 window from the import.

sql/moz-fx-data-marketing-prod/iprospect/adspend_v1/metadata.yaml
```diff
@@ -0,0 +1,18 @@
+friendly_name: Adspend
+description: >
+  Import detailed adspend iProspect CSV data from
+  moz-fx-data-marketing-prod-iprospect storage bucket.
```
acmiyaguchi, line 4: Repeat the comment from the script, at the very least:

> Import iProspect data from moz-fx-data-marketing-prod.iprospect.adspend_raw_v1.

acmiyaguchi (approved):
I looked at the iprospect dataset in BigQuery, I assume that's from running these scripts. It looks good to me.

sql/moz-fx-data-marketing-prod/iprospect/adspend_meta_v1/query.sql
scholtzan: vendor impressions. I figured it might be best to just keep the original naming for these to avoid any unexpected naming mismatches.

sql/moz-fx-data-marketing-prod/iprospect/adspend_v1/metadata.yaml
scholtzan: Yeah, I don't think it is strictly necessary in this case. Not having `depends_on_past` would have the advantage of running potential backfills in Airflow without having to reduce the number of parallel task runs.

PR #2195, opened 2021-07-15
[Bug 1720218] Desktop Feature Usage Table

scholtzan:
This is a WIP for https://bugzilla.mozilla.org/show_bug.cgi?id=1720218

The desktop feature usage table is implemented as a SQL script to automatically `COALESCE` metric values based on eligible browser version defined in the metadata. There is also a step-by-step description of what it would take to add new metrics to the query.

The main question I have ([see as inline comment](https://github.com/mozilla/bigquery-etl/pull/2195#discussion_r670821744)) is how the metadata should be stored. I listed a few options that I could think of.

Depends on https://github.com/mozilla/bigquery-etl/pull/2194

Checklist for reviewer:


bigquery_etl/format_sql/tokenizer.py
```diff
@@ -438,6 +444,7 @@ class FieldAccessOperator(Operator):
     AngleBracketKeyword,
     SpaceBeforeBracketKeyword,
     ReservedKeyword,
+    ConcatenationOperator,
```
scholtzan, line 447: https://github.com/mozilla/bigquery-etl/pull/2194

sql/moz-fx-data-shared-prod/telemetry_derived/feature_usage_v2/script.sql
```diff
@@ -0,0 +1,760 @@
...
+      SELECT
+        [
+          STRUCT(
+            "pdf_viewer_time_to_view_ms_content" AS metric,
+            "38" AS min_version,
+            "INT64" AS `type`
+          ),
+          ("pdf_viewer_time_to_view_ms_content_count", "38", "INT64"),
+          ("pdf_viewer_doc_size_kb", "38", "INT64"),
+          ("pdf_viewer_doc_size_kb_content", "38", "INT64"),
```
scholtzan, line 30: How do we want to store the metadata (eligible browser versions, whether metrics are vetted)? I can think of a couple of options:

1. As common table expression (as it's implemented here). This would have the advantage of having this metadata live right beside the actual query. So if fields are to be added, fewer places need to be updated. The disadvantage is that information about which metrics have been vetted is not really accessible. We could add another `vetted` field here, but this wouldn't be accessible by other tools like Looker.
     * We could add `vetted` metadata to the field description. It would be nice if BigQuery had the option to add tags to columns, there are policy tags, but those aren't really intended for this use case.
2. Store this metadata in a separate table (e.g. `telemetry_derived.feature_usage_meta_v1`). This would make the metadata more accessible and the script here would pretty much work the same way (just need to remove the `version_metadata` temp table). The disadvantage I see is when adding new fields, we'd need to manually update that metadata table. I'm also not sure how Looker would be able to use the vetted information.
3. Track metadata in Google Sheets and connect sheet to BigQuery. Similar to option 2, we can connect a Google sheet to a BigQuery table that can be queried. This might make adding new metadata a little more convenient but still requires changes in multiple places.

sql/moz-fx-data-shared-prod/telemetry_derived/feature_usage_v2/script.sql
```diff
@@ -0,0 +1,760 @@
...
+        feature_usages f
+      LIMIT
+        1
+    ),
+    UNNEST(columns) column
+  LEFT JOIN
+    version_metadata
+  ON
+    metric = column
+);
```
scholtzan, line 760: This is where the magic happens to `COALESCE` metric values based on eligible browser version defined in the metadata.

irrationalagent:
Thanks so much for doing this Anna, I'll take a closer look tomorrow morning

jklukas (commented):
This is workable, but it may be a substantial problem for maintainability if adding new fields requires making code changes in sync in multiple different places.

I think this would be better modeled as a yaml configuration file that defines all the metric names alongside the SQL snippets used to define them, the minimum version, etc.. And then we'd have a python script that generates the final schema.yaml and the query.sql using a jinja template. It would implemented as a module under `bigquery_etl` that we'd hook into the `generate-sql` task.

What I'm proposing here is similar to the POC in https://github.com/mozilla/bigquery-etl/pull/1480 for generated usage_all_time queries.

Defining the metrics in a yaml file is the only way forward I see to really get all the per-metric configuration consolidated together in a single place. I guess it really depends on DS feedback about what looks like a reasonable interface to work with.

sql/moz-fx-data-shared-prod/telemetry_derived/feature_usage_v2/script.sql
irrationalagent: The immediate use case I had in mind for the `vetted` column was to indicate which metrics should and should not be visible in looker. 

In looker, there is the `hidden` field for measures that controls which can be seen, however I just did some looking around and it doesn't appear that this field can be set contingently (e.g. via a liquid variable). This means that for now (unless we build the look ml files with a script or something) a person will have to manually reference this field to know whether a particular measure should be exposed. Which means that option 1 will work for now.

If we BQ need a table with the metadata, maybe we can stick with this struct as a source of truth but write a small script that scrapes from it and updates the BQ table. That way the metadata only has to be changed in one place (this struct) but can still be accessed via BQ (naive suggestion, don't know how workable this really is).

sql/moz-fx-data-shared-prod/telemetry_derived/feature_usage_v2/script.sql
```diff
@@ -0,0 +1,760 @@
...
+-- for ineligible versions.
+-- For adding a new metric:
+-- (1) Add metadata to `version_metadata` for the new metric. Metadata is specified as a list of
+--   STRUCTs where `metric` specifies the column name for the new feature metric in the result table,
+--   `min_version` specifies the minimum eligible browser version and `type` the data type of the
+--   computed metric value.
+-- (2) Update `feature_usages` and specify the feature metric
+-- (3) Update the schema.yaml file and add the new field to the schema
+-- (4) Run `./bqetl query schema deploy telemetry_derived.feature_usage_v2 --force` to update the destination
+--   table schema.
```
irrationalagent, line 15: So I think I could personally manage this workflow, but my guess is that for broader usability people would find updating a yaml file more familiar. opinions from @SuYoungHong and @godelstheory welcome on this point.

I think the question is, if only data folks are going to be making updates here, then maybe this kind of workflow is OK (I'm talking purely based on the ergonomics). If the table is getting so much use that PMs want to be able to add their own metrics, then this workflow is likely going to be too involved. But, we don't have any evidence for that yet. 

That said, maybe there are technical benefits to taking the extra time to yaml-fy that I don't understand. For example, if it sets a precedent or creates res-usable code that can be used for similar tables in the future, then I would be all for that.

irrationalagent: Also, if I'm reading this correctly, if I go through this workflow and add a metric, then it will just be present for all dates after the first run of the ETL, correct? Its not going to automatically backfill previous dates, or will it?

scholtzan: I agree, having YAML files to configure the metrics DS is interested in probably would it make easier for DS to make updates. I'll see if there is a good generic path forward to support YAML files in bigquery-etl since we had some related use cases already. 

As for backfilling, previous dates are not going to automatically backfill in general. 
There are 2 recommended ways for performing a backfill:
* via `bqetl` command line tool: run `./bqetl query backfill telemetry_derived.feature_usage_v2 --start_date=<start> --end_date=<end>` locally
* or via the [Airflow Admin Backfill UI](https://workflow.telemetry.mozilla.org/admin/backfill/) (sometimes it's a bit flaky)

scholtzan:
Closed in favour of https://github.com/mozilla/bigquery-etl/pull/2202

PR #2197, opened 2021-07-16
Bug 1720938 - Allow ignore on unknown values during stripe import

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1720938)

The schemas are already populated, so this would be the easiest way to ignore new fields and have the job running over the weekend. This needs to be followed up to determine whether to add the new fields.

Checklist for reviewer:


acmiyaguchi:
Merging this in today, it will be a lot easier to not have to deal with failures from the weekend.

relud (approved):

PR #2202, opened 2021-07-20
[Bug 1720218] Desktop Feature Usage Table from YAML config

scholtzan:
Supersedes https://github.com/mozilla/bigquery-etl/pull/2195

Based on the suggestions in https://github.com/mozilla/bigquery-etl/pull/2195#pullrequestreview-708531454. This allows to configure feature usage table metrics in a YAML file and implements the machinery to generate the query by running `./bqetl feature_usage generate`. The `schema.yaml` file gets generated automatically, so when adding a new metric only the `templating.yaml` file needs to be updated.

@irrationalagent Does this look like it will be more convenient for adding and maintaining feature usage metrics?

Checklist for reviewer:


bigquery_etl/feature_usage/templating.yaml
```diff
@@ -0,0 +1,1085 @@
...
+        sql: client_id
+      - name: subsample_id
+        sql: MOD(ABS(FARM_FINGERPRINT(client_id)), 100)
+      - name: submission_date
+        sql: DATE(submission_timestamp)
+      - name: is_headless
+        sql: LOGICAL_OR(COALESCE(environment.system.gfx.headless, FALSE))
+        vetted: true
+        min_version: 72
+        type: BOOL
```
scholtzan, line 238: This is the file for configuring the metrics that will be used for generating the query. This config defines `sources` which are tables to query from and each source has a set of `measures` which are the feature usage metrics we are interested in. For each measure the `sql` expression needs to be defined, and optionally a `min_version` a `type` and whether the metric is `vetted`.

jklukas (approved):
This looks like a complete solution and it achieves the goal of keeping all configuration for a new metric together. From my perspective, this looks like a much more maintainable approach than the BQ scripting-based solution.

If @irrationalagent thinks this is a reasonable interface from the DS perspective, then I think we're good to move forward with it.

irrationalagent:
🎉 I will take a closer look later today!

bigquery_etl/feature_usage/templating.yaml
```diff
@@ -0,0 +1,1123 @@
...
+# the SQL query for `telemetry_derived.feature_usage_v2`.
+# Run `./bqetl query schema deploy telemetry_derived.feature_usage_v2` to
+# update the schema of the BigQuery destination table. This requires GCP
+# credentials.
+#
+# New metrics will not be automatically backfilled for previous dates. For
+# performing a backfill, after changes have been approved and merged, run:
+#   ./bqetl query backfill telemetry_derived.feature_usage_v2 \
+#       --start_date=<YYYY-MM-DD> \  # replace with date range
+#       --end_date=<YYYY-MM-DD>
```
scholtzan, line 38: I added some docs here that might be helpful for folks wanting to add new metrics.

irrationalagent (approved):
Amazing work, the instructions for adding columns are super clear and should be something every data scientist should be able to handle. 

I already have a couple more columns / features in mind from new telemetry that I'd like to add, so once this is merged I'll probably be the first to try out the new workflow.

I could see there *might* be some e.g. daily rollups that could lead to some derived tables coming off this, but first I want to see how much of that we can handle on the looker side (since this is just a 1 pct sample). Also different metrics will need to be rolled up differently etc.

Thanks again

PR #2223, opened 2021-07-26
Use the most recent bucket ranges in glam for categorical histograms (fixes #2220)

### LINKED ISSUE #2220: Use the most recent bucket ranges in glam for categorical histograms if multiple exist
An issue came up where glam upserts of new data were not working because the rows were not unique when they were expected to be. It turned out the issue was related specifically to categorical histograms increasing their bucket counts when there was an underlying assumption that these counts would not need to change.

The steps I did to discover this were:

The probe counts query here: https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/telemetry_derived/clients_histogram_probe_counts_v1/query.sql groups by first_bucket, last_bucket and num_buckets  for histograms... the assumption is that these would be the same for any given probe so they can be dropped after grouping and we assume there's only one row.

However with this query: https://gist.github.com/emtwo/35e7dd53d1d999ead572a474f687b227 I found that for a certain set of dimensions, there are two rows, each one with different values for the buckets. I think this must mean that someone recently updated the bucket range on the `probe external_protocol_handler_dialog_context_scheme`.

Separately, I confirmed with this: https://gist.github.com/emtwo/0bc0f0b0d95ca996941ee787f38e6607 that there is only the one metric, `external_protocol_handler_dialog_context_scheme`  that has updated buckets.

After discussion with chutten, I confirmed we should only ever expect categorical histograms to have additional categories/buckets to be added but no other histograms should have changing buckets. With that, we can choose to always look at the largest values for `last_bucket` and `num_buckets` when performing aggregations.

Specifically, given this query: https://gist.github.com/emtwo/35e7dd53d1d999ead572a474f687b227 which is a modification of the `clients_histogram_probe_counts` query showing a problem probe, we can address the issue by aggregating using `max(last_bucket)` and `max(num_buckets)` and test that the user counts and bucket values are within expected ranges.

Iinh:
fixes #2220.

Checklist for reviewer:


robhudson:
This looks good to me. Just squash the merges together before merging.

robhudson (approved):

Issue #2104, opened 2021-06-09
Add `bqetl view create` function


fbertsch:
It would be nice to parallel the creation of queries with a view creation that is easy to use.

PR #2224, opened 2021-07-26
CLI view support

scholtzan:
This adds the following CLI commands for views:

```
$ ./bqetl view --help                                                              
Usage: bqetl view [OPTIONS] COMMAND [ARGS]...

  Commands for managing views.

Options:
  --help  Show this message and exit.

Commands:
  create    Create a new view with name <dataset>.<view_name>, for...
  publish   Publish views.
  validate  Validate a view.
```

Fixes https://github.com/mozilla/bigquery-etl/issues/2104

Checklist for reviewer:


bigquery_etl/cli/query.py
```diff
@@ -31,80 +39,9 @@
...
-    """Generate a project-id option, with optional default."""
-    return click.option(
-        "--project-id",
-        "--project_id",
-        help="GCP project ID",
-        default=default,
-        callback=is_valid_project,
-    )
-
-
```
scholtzan, line 107: Move to `cli.common` since these CLI parameters are also used for other commands.

bigquery_etl/view/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+NON_USER_FACING_DATASET_SUFFIXES = (
+    "_derived",
+    "_external",
+    "_bi",
+    "_restricted",
+)
+
+
+@attr.s(auto_attribs=True)
+class View:
```
scholtzan, line 57: This class combines some of the code related to managing views and makes it easier to reuse some of the logic (e.g. for validation). I'm thinking of doing something similar for queries since currently the logic is spread across several scripts with lots of redundant code that is hard to test.

bigquery_etl/view/publish_views.py
```diff
@@ -1,184 +0,0 @@
...
-            # Only views for moz-fx-data-shared-prod will get published to other
-            # projects, for example to moz-fx-data-derived-datasets
-            if project_id != "moz-fx-data-shared-prod":
-                print(f"Skipping {filepath} because --target-project is set")
-                return True
-            target_view = target_view_orig.replace(project_id, args.target_project, 1)
-            # We only change the first occurrence, which is in the target view name.
-            sql = sql.replace(project_id, args.target_project, 1)
-        job_config = bigquery.QueryJobConfig(use_legacy_sql=False, dry_run=args.dry_run)
-        query_job = client.query(sql, job_config)
```
scholtzan, line 77: This validation code is almost identical to the logic in the old `validate_views.py` file. Moving both into a `View` class makes it easier to reuse the validation logic.

bigquery_etl/cli/glean_usage.py
```diff
@@ -4,7 +4,7 @@
 from pathlib import Path
 import click
 
-from ..cli.utils import (
+from ..cli.common import (
```
relud, line 7: why is this being renamed?

sql/moz-fx-data-test-project/test/simple_view/view.sql
```diff
@@ -0,0 +1 @@
+SELECT 1
```
relud, line 1: this file needs to be formatted with `script/format_sql`

script/publish_views
```diff
@@ -4,4 +4,8 @@
 
 cd "$(dirname "$0")/.."
 
-exec python3 -m bigquery_etl.view.publish_views "$@"
+./bqetl bootstrap
+./bqetl view publish "$@"
+
+wait
+trap - EXIT
```
relud, line 11: what is this waiting for, and why is it resetting the EXIT trap when it's never set?

script/publish_views
```diff
@@ -4,4 +4,8 @@
 
 cd "$(dirname "$0")/.."
 
-exec python3 -m bigquery_etl.view.publish_views "$@"
+./bqetl bootstrap
```
relud, line 7: why does this need to bootstrap every time it's run?

bigquery_etl/cli/glean_usage.py
scholtzan: probably not necessary. I'll undo the rename.

script/publish_views
scholtzan: Right, it doesn't seem necessary here. I was copying this over from another script: https://github.com/mozilla/bigquery-etl/blob/main/script/update_example_glean_usage

script/publish_views
scholtzan: Running `./bqetl bootstrap` will install requirements and setup `venv`. [Jenkins](https://github.com/mozilla-services/cloudops-infra/blob/623a8cc546d0084d837dbf4889d1239d562d4922/projects/data-pioneer/Jenkinsfile.bigquery#L243) will run this script and without running bootstrap `./bqetl view publish` won't work in that case.

relud (approved):

PR #2226, opened 2021-07-27
Add percentile jackknife function.

emtwo:
This PR adds a function for computing CI for percentiles of a histogram using jackknifing. I used a few existing pieces of code for reference:

1. Glam percentile computations for histograms (https://github.com/mozilla/bigquery-etl/blob/main/sql/mozfun/glam/percentile/udf.sql)
2. Existing jackknife mean ci function: https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/udf_js/jackknife_mean_ci/udf.sql

A couple of notes:
1. I wanted to directly call the glam percentile function from this jackknife function but bigquery didn't seem to allow a call from one js udf to another (or at least I couldn't get it to work myself), so I copied it over as part of this function.
2. I'm excluding the sql formatting on this file because it was actually breaking the unit tests by changing variables called `values` to upper case `VALUES`. I think it was treating it as a keyword perhaps.

For some context, the plan is to use this function in operational monitoring to provide confidence intervals for percentiles compared for different branches. Here is the opmon prototype where percentiles are used without confidence intervals: https://mozilladev.cloud.looker.com/dashboards-next/14?OS=Mac&Cores=4

Checklist for reviewer:


scholtzan:
cc @scholtzan 
So I get notified to address changes while Marina is out.

bigquery_etl/format_sql/format.py
```diff
@@ -141,6 +141,7 @@
     "sql/moz-fx-data-shared-prod/udf_js/jackknife_mean_ci.sql",
     "sql/moz-fx-data-shared-prod/udf_js/jackknife_ratio_ci.sql",
     "sql/moz-fx-data-shared-prod/udf_js/jackknife_sum_ci.sql",
+    "sql/moz-fx-data-shared-prod/udf_js/jackknife_percentile_ci/udf.sql",
```
godelstheory, line 144: Is the `udf.sql` at the end of the path necessary? The row above doesn't have this included in the path, though the SQL is defined in a `udf.sql` files.

sql/moz-fx-data-shared-prod/udf_js/jackknife_percentile_ci/udf.sql
```diff
@@ -0,0 +1,247 @@
...
+
+  function array_avg(arr) {
+      return array_sum(arr) / arr.length;
+  }
+
+  function getMeanErrorsPercentile(percentile, histogram, fullDataPercentile) {
+    var jk_percentiles = [];
+
+    histogram.values.forEach((bucket, i) => {
+      var histCopy = JSON.parse(JSON.stringify(histogram.values));
```
godelstheory, line 81: Nit: I am not familiar with copying a histogram in JS. Is this converting to a string then parsing back optimal?

godelstheory:
@scholtzan I have a couple small comments in the review above.

bigquery_etl/format_sql/format.py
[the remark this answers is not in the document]
scholtzan: The final `udf.js` is necessary here. We changed the directory structure for UDFs at some point, so the files above refer to old files that have been moved. I removed the references here to avoid any future confusion.
Also, skipping formatting for the new UDF here doesn't seem to be necessary.

sql/moz-fx-data-shared-prod/udf_js/jackknife_percentile_ci/udf.sql
scholtzan: This seems to be the most convenient way to create a deep-copy of the `histogram.values`. If we notice any impact on performance we could look into more efficient (which will also be a little more complex) alternatives.

scholtzan:
@godelstheory I've addressed you comments.

godelstheory (approved):

scholtzan:
I incorporated the UDF into the Looker dashboard: https://mozilladev.cloud.looker.com/dashboards-next/fission::fission_with_dimensions?OS=Windows&Cores=4&Percentile=50

PR #2227, opened 2021-07-27
[Bug 1722546] Workaround for copy_deduplicate copy_tables()

scholtzan:
This is just a temporary workaround to get the ETL running. I was able to copy the intermediate tmp tables into a single table by splitting them up over two `copy_table()` jobs.

This is probably not a good long-term solution since we copy data twice. I'd suggest we open a GCP support case and see if there is a better fix.

Checklist for reviewer:


relud (approved):

bigquery_etl/copy_deduplicate.py
```diff
@@ -254,8 +254,26 @@ def _copy_join_parts(client, stable_table, query_jobs):
...
+            copy_job_2.result()
+
+            copy_job = client.copy_table(
+                [tmp_table_1, tmp_table_2], stable_table, job_config=job_config
+            )
             copy_job.result()
+
             logging.info(f"Copied {len(query_jobs)} results to populate {stable_table}")
             for job in query_jobs:
                 client.delete_table(job.destination)
```
relud, line 279: need to clean up the intermediate tables:
```suggestion
            for job in query_jobs + [copy_job_1, copy_job_2]:
                client.delete_table(job.destination)
```

bigquery_etl/copy_deduplicate.py
```diff
@@ -254,10 +254,28 @@ def _copy_join_parts(client, stable_table, query_jobs):
...
+
+            tmp_table_2 = temp_dataset.temp_table()
+            copy_job_2 = client.copy_table(
+                sources[len(sources) // 2 :], tmp_table_2, job_config=job_config
+            )
+            copy_job_2.result()
+
+            copy_job = client.copy_table(
+                [tmp_table_1, tmp_table_2], stable_table, job_config=job_config
+            )
```
relud, line 274: for `sources` each table is formatted as `f"{sql_table_id(job.destination)}${partition_id}"`, does that need to be done here for `[tmp_table_1, tmp_table_2]` ?

relud: to be clear, if it works and doesn't erase all of main_v4, I don't think it's necessary. Just making sure this has been checked before merging.

relud: at least in my sandbox project I needed this, filed #2228 to fix

PR #2228, opened 2021-07-27
Add partition id to intermediate copy job sources

relud:
Checklist for reviewer:

### DIFF
--- a/bigquery_etl/copy_deduplicate.py
+++ b/bigquery_etl/copy_deduplicate.py
@@ -269,8 +269,13 @@ def _copy_join_parts(client, stable_table, query_jobs, temp_dataset):
             )
             copy_job_2.result()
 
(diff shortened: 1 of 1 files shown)


scholtzan (approved):
Did you test that this works? Just wondering, since `tmp_table_1` and `tmp_table_2` aren't going to be partitioned.

scholtzan:
Okay just tested it myself. Seems to work 👍

relud:
> Did you test that this works? Just wondering, since `tmp_table_1` and `tmp_table_2` aren't going to be partitioned.

> Okay just tested it myself. Seems to work 👍

i believe it works because copy jobs require source and destination tables to have the same partitioning, so `tmp_table_1` and `tmp_table_2` are implicitly partitioned.

PR #2244, opened 2021-08-04
CLI view support

scholtzan:
This is basically identical to https://github.com/mozilla/bigquery-etl/pull/2224 which I had to revert due to `./bqetl bootstrap` not working in Jenkins (see https://github.com/mozilla/bigquery-etl/pull/2238). I didn't realize it at the time, but there is also a `script/bqetl` which can be used instead and is already used for the `generate_all_views` and [`generate_sql`](https://github.com/mozilla/bigquery-etl/blob/main/script/generate_sql) scripts. This avoids running the `bootstrap` step, so Jenkins shouldn't run into any issues here.

Reverts mozilla/bigquery-etl#2238


script/publish_views
```diff
@@ -4,4 +4,4 @@
 
 cd "$(dirname "$0")/.."
 
-exec python3 -m bigquery_etl.view.publish_views "$@"
+./script/bqetl view publish "$@"
```
scholtzan, line 7: This avoids the issue in https://github.com/mozilla/bigquery-etl/pull/2238

.circleci/config.yml
```diff
@@ -244,7 +244,9 @@ jobs:
       - *java_deps
       - run:
           name: Validate views
-          command: PATH="venv/bin:$PATH" script/validate_views
+          command: |
+            ./bqetl bootstrap
+            ./bqetl view validate
```
relud, line 249:
```suggestion
            script/bqetl view validate
```

.circleci/config.yml
```diff
@@ -244,7 +244,7 @@ jobs:
       - *java_deps
       - run:
           name: Validate views
-          command: PATH="venv/bin:$PATH" script/validate_views
+          command: script/bqetl view validate
```
relud, line 247: oop, I got this wrong, should be:
```suggestion
          command: PATH="venv/bin:$PATH" script/bqetl view validate
```

PR #2245, opened 2021-08-04
Fix skipping publishing and validation of views

scholtzan:
We've been running into the following error

```
Error publishing /tmp/sql/moz-fx-data-shared-prod/telemetry/clients_probe_processes/view.sql. Invalid view definition.
```

The view should be skipped by wasn't because the current logic assumed it would be stored in `sql/moz-fx-data-shared-prod/` rather than `/tmp/sql/moz-fx-data-shared-prod/`

Checklist for reviewer:

### DIFF
--- a/bigquery_etl/cli/view.py
+++ b/bigquery_etl/cli/view.py
@@ -6,7 +6,7 @@ import re
 import string
 import sys
 
-from multiprocessing.pool import Pool
+from multiprocessing.pool import Pool, ThreadPool
 
 from ..view import View
 from .dryrun import dryrun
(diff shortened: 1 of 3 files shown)


bigquery_etl/cli/view.py
```diff
@@ -197,9 +197,10 @@ def publish(
     views = [View.from_file(f) for f in view_files]
     views = [v for v in views if not user_facing_only or v.is_user_facing]
 
-    with Pool(parallelism) as p:
+    with ThreadPool(parallelism) as p:
```
scholtzan, line 200: It appears that `ThreadPool` is actually a little faster than `Pool`. This is also what we used previously

relud (approved):

PR #2250, opened 2021-08-06
Fix skipping view validations when publishing

scholtzan:
Checklist for reviewer:

cc @whd

### DIFF
--- a/bigquery_etl/view/__init__.py
+++ b/bigquery_etl/view/__init__.py
@@ -194,7 +194,10 @@ class View:
 
         # avoid checking references since Jenkins might throw an exception:
         # https://github.com/mozilla/bigquery-etl/issues/2246
-        if self._valid_view_naming():
+        if (
+            any(str(self.path).endswith(p) for p in SKIP_VALIDATION)
+            or self._valid_view_naming()
+        ):
             client = bigquery.Client()
             sql = self.content
             target_view = self.view_identifier


bigquery_etl/view/__init__.py
```diff
@@ -194,7 +194,10 @@ def publish(self, target_project=None, dry_run=False):
 
         # avoid checking references since Jenkins might throw an exception:
         # https://github.com/mozilla/bigquery-etl/issues/2246
-        if self._valid_view_naming():
+        if (
+            any(str(self.path).endswith(p) for p in SKIP_VALIDATION)
```
scholtzan, line 198: We'll still need to skip views here, like `sql/moz-fx-data-shared-prod/telemetry/client_probe_counts_v1/view.sql`

bigquery_etl/view/__init__.py
```diff
@@ -194,7 +194,10 @@ def publish(self, target_project=None, dry_run=False):
...
+            or self._valid_view_naming()
```
relud, line 199: I think we need the opposite filter here:
```suggestion
            not any(str(self.path).endswith(p) for p in SKIP_VALIDATION)
            and self._valid_view_naming()
```

relud (approved):
r+wc

bigquery_etl/view/__init__.py
scholtzan: There are some views, like `telemetry.client_probe_counts_v1`, that we want to publish although their validation would fail (view name != file path). So I it should be correct as it is.

PR #2251, opened 2021-08-06
Revert "Fix skipping view validations when publishing"

relud:
Reverts mozilla/bigquery-etl#2250

### DIFF
--- a/bigquery_etl/view/__init__.py
+++ b/bigquery_etl/view/__init__.py
@@ -194,10 +194,7 @@ class View:
 
         # avoid checking references since Jenkins might throw an exception:
         # https://github.com/mozilla/bigquery-etl/issues/2246
-        if (
-            any(str(self.path).endswith(p) for p in SKIP_VALIDATION)
-            or self._valid_view_naming()
-        ):
+        if self._valid_view_naming():
             client = bigquery.Client()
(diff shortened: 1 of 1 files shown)


bigquery_etl/view/__init__.py
```diff
@@ -194,10 +194,7 @@ def publish(self, target_project=None, dry_run=False):
 
         # avoid checking references since Jenkins might throw an exception:
         # https://github.com/mozilla/bigquery-etl/issues/2246
-        if (
-            any(str(self.path).endswith(p) for p in SKIP_VALIDATION)
```
relud, line 198: this condition is already filtered at line 191 above

scholtzan: Line 191 is for `SKIP_PUBLISHING`, but there are some views, like `telemetry.client_probe_counts_v1`, that we want to publish although their validation would fail (view name != file path).

Running `./bqetl view publish` with this would fail with Error publishing sql/moz-fx-data-shared-prod/telemetry/client_probe_counts_v1/view.sql. Invalid view definition.

relud: ah, I get it now. I totally thought this logic was inverted from what it needed. thanks for the clarification!

PR #2252, opened 2021-08-06
Bug 1711797 - Add access point search probes to clients daily

BenWu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1711797
https://mozilla-hub.atlassian.net/browse/DS-1603

See https://github.com/mozilla/bigquery-etl/pull/2253 for downstream search clients daily changes

This adds the in-content search, search with ads, and ad click probes that are per access point, e.g. `browser_search_withads_urlbar`, `browser_search_withads_abouthome`.  Each probe is in its own column and the data in the same format as in the main ping (array of key-value pairs).  Keys look like `engine:tagged-follow-on`.  I tried to combine each search type (in-content, search with ads, ad click) into a single column by adding the source to the key (e.g. `engine:tagged-follow-on.urlbar`) which would make the schema and querying easier but it was running into the bigquery complexity limit.

Checklist for reviewer:


akkomar (approved):
LGTM, do these fields need to be manually added to the schema?

BenWu:
We can update the schema using the CLI https://mozilla.github.io/bigquery-etl/cookbooks/common_workflows/#add-a-new-field-to-clients_daily
And then retry the CI

PR #2253, opened 2021-08-06
Bug 1711797 Add access point search probes to search clients daily

BenWu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1711797
https://mozilla-hub.atlassian.net/browse/DS-1603

See https://github.com/mozilla/bigquery-etl/pull/2252 for upstream clients daily changes

The query tries to use the access point probe data first and uses the old probes if those are missing.

There are changes to the resulting sources which I'm not sure are acceptable or if they break anything:
- for tagged sap, the string in the source is now `tagged` instead of `sap`, e.g. `in-content:tagged...` or `in-content:tagged-follow-on...` instead of `in-content:sap...` or `in-content:sap-follow-on...`
- in-content, with ads, and ad click sources now end with `.access_point`, e.g. `in-content:tagged-follow-on:firefox-b-1-d.searchbar` instead of `in-content:sap-follow-on:firefox-b-1-d`
- with ads and ad click previously only looked like `search-with-ads:` or `ad-click:`, they now have tagged information and access point, e.g. `ad-click:organic.reload`


akkomar:
> There are changes to the resulting sources which I'm not sure are acceptable or if they break anything:
> 
>     * for tagged sap, the string in the source is now `tagged` instead of `sap`, e.g. `in-content:tagged...` or `in-content:tagged-follow-on...` instead of `in-content:sap...` or `in-content:sap-follow-on...`
> 
>     * in-content, with ads, and ad click sources now end with `.access_point`, e.g. `in-content:tagged-follow-on:firefox-b-1-d.searchbar` instead of `in-content:sap-follow-on:firefox-b-1-d`
> 
>     * with ads and ad click previously only looked like `search-with-ads:` or `ad-click:`, they now have tagged information and access point, e.g. `ad-click:organic.reload`

@xluo-ds would these break any queries or dashboards?

BenWu:
We had some discussion here https://mozilla-hub.atlassian.net/browse/DS-1603 and it sounds like it should be fine but I'll wait for confirmation

akkomar (approved):

PR #2258, opened 2021-08-11
Replace normalize search engine with stub implementation

BenWu:
Other repo changes: https://github.com/mozilla/private-bigquery-etl/pull/19

Moved to private-bigquery-etl and replaced the udf here with a stub so that tests depending on it can run.  The idea here is that in the deployed version we overwrite the udf here with the one in the private repo by taking out the `--ignore-existing` from rsync.

@jklukas are there any issues with this and is there a reason we wanted to explicitly set `--ignore-existing` in the first place?  Spoke with akomar about this and we don't see an issue


BenWu:
I tried fully qualifying yesterday and it wasn't working with an unrecognized name error.  Turns out I put the backticks around the argument as well which is wrong so fully qualifying does in fact work.  That requires access to shared-prod when running the test but that's probably safe to assume.

akkomar (approved):

PR #2267, opened 2021-08-16
Minor improvements to bqetl script

wlach:
* Check for a valid python installation
* Actually exit if `venv` directory already exists
* Don't require virtualenv package (use 'venv' builtin)
* Ensure that we don't attempt to recompile numpy or pandas when
  installing on more recent versions of Mac


wlach:
The netlify deploy errors will be fixed by #2268

relud (approved):

PR #2302, opened 2021-08-26
Fix non-private image builds

### LINKED ISSUE #2287: Main branch failing `deploy` task due to mvn error
It looks like we have not had a successful run of the `deploy` task for this repo in about a week, failing with a maven error while building the docker image. Strangely, the `deploy-to-private-gcr` task (which also has to build the image) works.

I notice that the docker build is using cache up until the failing `RUN mvn dependency:copy-dependencies` invocation, so perhaps there's something bad in the docker layer cache?

The error is:

```
Step 12/23 : RUN mvn dependency:copy-dependencies
 ---> Running in 2efdc878229b
ls: cannot access '/usr/bin/mvn': Operation not permitted
Error: Could not find or load main class org.codehaus.plexus.classworlds.launcher.Launcher
Caused by: java.lang.ClassNotFoundException: org.codehaus.plexus.classworlds.launcher.Launcher
The command '/bin/sh -c mvn dependency:copy-dependencies' returned a non-zero code: 1
```

cc @relud

relud:
also use docker orb and build docker images on branches, but only deploy on main and tags

Fixes #2287

Checklist for reviewer:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,7 @@ version: 2.1
 
 orbs:
   gcp-gcr: circleci/gcp-gcr@0.13.0
+  docker: circleci/docker@1.5
 
 jobs:
   build:
@@ -317,9 +318,7 @@ jobs:
               && git push \
               || echo "Skipping push since it looks like there were no changes"
   deploy:
-    docker:
-      # bash required for step: Determine docker image name
-      - image: relud/docker-bash
+    executor: docker/docker
     steps:
       - checkout
       - setup_remote_docker:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -331,20 +330,22 @@ jobs:
...
-          name: Deploy to Dockerhub
-          command: |
-           echo "${DOCKER_PASS:?}" | \
-            docker login -u "${DOCKER_USER:?}" --password-stdin
-            docker push "$IMAGE"
+      - when:
+          condition: &deploy-condition
+            or:
+              - equal: [main, << pipeline.git.branch >>]
+              - << pipeline.git.tag >>
```
jklukas, line 337: If this is git tag, do we still want to publish an image with tag `latest`? Can we add some commentary here about the intention of tagging?

.circleci/config.yml
```diff
@@ -419,13 +420,17 @@ jobs:
...
             cp -r /tmp/workspace/private-generated-sql/sql sql
-      - gcp-gcr/gcr-auth
-      - gcp-gcr/build-image:
-          image: bigquery-etl
-          tag: ${CIRCLE_TAG:-latest}
-      - gcp-gcr/push-image:
+      - when:
+          condition: *deploy-condition
+          steps:
+            - gcp-gcr/gcr-auth
```
jklukas, line 426: Can this gcr-auth step be delayed until after `build-image` so that it can be nested under the same `when` condition?

jklukas (approved):
r+wc

It would be good to include more commentary about intentions here, as this config is complex enough to be difficult to read.

.circleci/config.yml
relud: the tag is defined on line 344 below: `tag: ${CIRCLE_TAG:-latest}`, so the image tag matches the git tag, and latest is used for branch main. This condition is equivalent to the one in workflows that it replaced:
```
          filters:
            branches:
              only: main
            tags:
              only: /.*/
```

.circleci/config.yml
relud: this follows the pattern set in the [gcp-gcr orb job](https://github.com/CircleCI-Public/gcp-gcr-orb/blob/master/src/jobs/build-and-push-image.yml#L123-L147) and [docker orb job](https://github.com/CircleCI-Public/docker-orb/blob/master/src/jobs/publish.yml#L165-L200) of doing auth->build->push.

At least one reason for doing this could be that docker has a `--cache-from` flag now that allows one to use the previous image layers as a build cache: https://github.com/CircleCI-Public/docker-orb/blob/master/src/commands/build.yml#L128

PR #2307, opened 2021-08-30
Generate events daily queries dynamically

wlach:
Instead of relying on someone to run the script to generate the events_daily
queries, bundle this functionality into bqetl cli and run it by default
via the `generate_sql` script.


wlach:
I created a quick script to diff the generated versions of the stuff I deleted vs. what was checked into the tree, and found no difference: https://gist.github.com/wlach/fe9d76bc11927e21bdb3f303970e8629

wlach:
I'm not sure why `verify-dags-up-to-date` is failing, but I'm basically at my end of day. I'll try retrying the job...

wlach:
> I'm not sure why `verify-dags-up-to-date` is failing, but I'm basically at my end of day. I'll try retrying the job...

Seems to be reproducible, hmm. I can investigate more tomorrow but if anyone knows offhand what's going on lmk.

scholtzan:
I believe the reason for why `verify-dags-up-to-date` is failing is that the `bqetl_fenix_event_rollup` DAG doesn't have any tasks associated to it. `verify-dags-up-to-date` gets executed before the queries have been generated.
The fix should be simple: we need to run the query generation before validating the DAGs.

sql/moz-fx-data-shared-prod/org_mozilla_firefox_derived/events_daily_v1/metadata.yaml
```diff
@@ -1,21 +0,0 @@
...
-description: >
-  Packed event representation with one-row per-client
-owners:
-  - frank@mozilla.com
-labels:
-  application: org_mozilla_firefox
-  schedule: daily
-  incremental: true
-scheduling:
-  dag_name: bqetl_fenix_event_rollup
```
scholtzan, line 14: All the queries that referenced `bqetl_fenix_event_rollup` got removed and will be generated after the DAG generation step. This will result in the DAG having to tasks.

scholtzan:
This comment came up today and is pointing to the same problem: https://github.com/mozilla/bigquery-etl/pull/2308#issuecomment-909371508 I think it's currently not possible to remove the generated queries and just have them generated as part of sql-generate.

wlach:
> This comment came up today and is pointing to the same problem: [#2308 (comment)](https://github.com/mozilla/bigquery-etl/pull/2308#issuecomment-909371508) I think it's currently not possible to remove the generated queries and just have them generated as part of sql-generate.

Ok, let's put the generated queries back in while we figure out a more general solution to this problem.

scholtzan:
Yeah, I'm planning to put together a proposal some time soon

wlach:
`private-generate-sql` failed in this PR, looks like this is #2300 again. Aside from that I think this is looking ok now.

scholtzan (approved):

PR #2308, opened 2021-08-30
Generate experiment monitoring views

scholtzan:
I figured it would make sense to automatically generate experimentation related datasets instead of maintaining them manually. Especially, since we'll have to add Focus and Klar monitoring support asap. The generation of these queries is similar to feature_usage.

Checklist for reviewer:


jklukas (commented):
This generally looks good. I would like to see us remove the generated queries from `main` as part of this PR and add in a call to `bqetl experiment_monitoring` in `script/generate_sql` so that these are built and published to generated-sql. Does that sound feasible?

scholtzan:
It should be feasible, however one difficulty with removing the queries is that it will affect the Airflow DAG generation. Since the `metadata.yaml` files get removed and are only generated after the DAG generation runs, these queries will not be added as tasks to the Airflow DAG.
I guess a possible workaround is to always run generate-sql and then generate the Airflow DAGs, but that will also slow down the whole DAG generation process.

jklukas:
> It should be feasible, however one difficulty with removing the queries is that it will affect the Airflow DAG generation. Since the `metadata.yaml` files get removed and are only generated after the DAG generation runs, these queries will not be added as tasks to the Airflow DAG.

After writing, I felt almost certain there would be some reason blocking the way to take the generated queries out of `main`. The `glean_usage` machinery doesn't generate anything on `main`, but I realize it doesn't run into this issue because the actual query running is handled with bespoke python code that handles running that queries rather than DAG generation.

> I guess a possible workaround is to always run generate-sql and then generate the Airflow DAGs, but that will also slow down the whole DAG generation process.

I would agree that it's not currently reasonable to require the minutes-long generate-sql process in each run of DAG generation, so fine to move forward here with generated queries checked in.

jklukas (approved):

PR #2321, opened 2021-09-07
Fix for Bugzilla: 1729084 - Add UDF to remove outliers

ANich:
Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1729084

PR:
- Adds `filter_probe_counts_with_outlier_values` UDF and uses it to remove  metrics/probes in the relevant urlbar keyed scalar metrics if they pass a threshold value. 
This is similar to the approach in [`mobile_search_clients_daily_v1`](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/query.sql#L632-L633) and we use the same threshold here (10,000).
- Removed double call of `one_index` so position counts are 1-based instead of 2-based 
- Adds tests

Checklist for reviewer:

~- [ ] If the PR comes from a fork, trigger integration CI tests by running the [Push to upstream workflow](https://github.com/mozilla/bigquery-etl/actions/workflows/push-to-upstream.yml) and provide the `<username>:<branch>` of the fork as parameter. The parameter will also show up
in the logs of the `manual-trigger-required-for-fork` CI task together with more detailed instructions.~
~- [ ] If adding a new field to a query, ensure that the schema and dependent downstream schemas have been updated~

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/urlbar_clients_daily_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/urlbar_clients_daily_v1/query.sql
@@ -2,12 +2,33 @@ CREATE TEMP FUNCTION one_index(x ANY TYPE) AS (
   CAST(IF(SAFE_CAST(x AS INT64) < 0, SAFE_CAST(x AS INT64), SAFE_CAST(x AS INT64) + 1) AS STRING)
 );
 
-CREATE TEMP FUNCTION one_index_struct(x STRUCT<k STRING, v INT64>) AS (
-  STRUCT(one_index(x.k) AS key, x.v AS value)
+CREATE TEMP FUNCTION one_index_struct(record STRUCT<k STRING, v INT64>) AS (
+  STRUCT(one_index(record.k) AS key, record.v AS value)
 );
 
-CREATE TEMP FUNCTION one_index_array(x ARRAY<STRUCT<k STRING, v INT64>>) AS (
-  ARRAY(SELECT one_index_struct(e) FROM UNNEST(x) AS e)
+CREATE TEMP FUNCTION one_index_array(map ARRAY<STRUCT<k STRING, v INT64>>) AS (
+  ARRAY(SELECT one_index_struct(record) FROM UNNEST(map) AS record)
+);
+
+CREATE TEMP FUNCTION filter_probe_counts_with_outlier_values(
+  map ARRAY<STRUCT<k STRING, v INT64>>
+) AS (
+  -- Similar to search probes/metrics, count values over 10000 are considered too high
+  ARRAY(
+    SELECT
+      STRUCT(record.k AS key, record.v AS value)
+    FROM
+      UNNEST(map) AS record
+    WHERE
+      record.v <= 10000
+  )
+);
+
+CREATE TEMP FUNCTION transform_scalar_metric_sum_columns(map ARRAY<STRUCT<k STRING, v INT64>>) AS (
+    -- Transforms the sums of probe counts columns from clients_daily by:
+        -- Applying 1-indexing to reflex index -> position (of urlbar choice)
+        -- Replacing outlier sum values above a certain threshold with 0
+  one_index_array(filter_probe_counts_with_outlier_values(map))
 );
 
 WITH combined_urlbar_picked AS (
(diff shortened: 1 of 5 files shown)


sql/moz-fx-data-shared-prod/telemetry_derived/urlbar_clients_daily_v1/query.sql
```diff
@@ -28,64 +32,99 @@ WITH combined_urlbar_picked AS (
     [
       STRUCT(
         "autofill" AS type,
-        one_index_array(scalar_parent_urlbar_picked_autofill_sum) AS position
+        one_index_array(
+          replace_outlier_values_with_zero(scalar_parent_urlbar_picked_autofill_sum)
```
akkomar, line 36: How about calling `replace_outlier_values_with_zero` in the `one_index_array` function?
This would make these calls in the query shorter. `one_index_array` would need to be renamed to make it clearer what it does (e.g. `one_index_array_normalized`).

sql/moz-fx-data-shared-prod/telemetry_derived/urlbar_clients_daily_v1/query.sql
```diff
@@ -2,12 +2,16 @@ CREATE TEMP FUNCTION one_index(x ANY TYPE) AS (
...
 );
 
-CREATE TEMP FUNCTION one_index_array(x ARRAY<STRUCT<k STRING, v INT64>>) AS (
-  ARRAY(SELECT one_index_struct(e) FROM UNNEST(x) AS e)
+CREATE TEMP FUNCTION one_index_array(map ARRAY<STRUCT<k STRING, v INT64>>) AS (
+  ARRAY(SELECT one_index_struct(record) FROM UNNEST(map) AS record)
+);
+
+CREATE TEMP FUNCTION replace_outlier_values_with_zero(map ARRAY<STRUCT<k STRING, v INT64>>) AS (
+    ARRAY(SELECT STRUCT(record.k AS key, IF(record.v > 1000000, 0, record.v)) FROM UNNEST(map) AS record)
```
akkomar, line 14: I'd add a comment about this number and why we do this filtering.

akkomar (changes requested):
Looks great, thanks for taking this on!

My comments:
* SQL formatting check is failing - please follow quickstart instructions from the Readme to bootstrap cli tool and reformat (https://github.com/mozilla/bigquery-etl#formatting-sql)
* function call can be wrapped (see comment inline)
* commit message: would be good if it was formatted like `Bug XXXXXX - SUBJECT_TEXT` - this way it could be automatically linked to Bugzilla bug. Also: https://chris.beams.io/posts/git-commit/
* as for the right threshold:
  * maybe we can see what's the distribution of this values per client and for example cut off p995?
  * Or run this for a bunch of values and see either how many clients are cut off or how this affects Teon's query?
  * @teonbrooks what do you think?

sql/moz-fx-data-shared-prod/telemetry_derived/urlbar_clients_daily_v1/query.sql
teonbrooks: @rebecca-burwei, do you have any insights to what we should set as a threshold? I was told that the search counts have a similar procedure that's at 20k iirc

ANich: Here are the limits for [Desktop](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v8/query.sql#L263) and [Mobile](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/query.sql#L632-L633) search. Looking backwards seems like it was just more than expected for the sum of probe counts from a single client in a single day. 

Looking at a window around the anomaly day by client it was definitely one client with 1B+ probe count sums that day. The second highest was on the order of ~100,000 and a few with ~10,000.

teonbrooks: ok, cool. i think let's go with 10,000 similar to the above and make a comment about it.

ANich:
Updated with the recommendations thanks @teonbrooks @akkomar 👍 just getting a handle on the testing then I'll open for review

sql/moz-fx-data-shared-prod/telemetry_derived/urlbar_clients_daily_v1/query.sql
teonbrooks: just thought about this: should we be mapping this to zero or to NULL? they have very different implications. other patterns mentioned above map it to NULL

akkomar: @teonbrooks what implications are you thinking of?

I do agree however that `NULL` makes sense as opposed to `0`, which signifies "no interactions".

teonbrooks: mostly when it comes to computing picked per user, if a user actually didn't choose anything, that is worth noting as 0 whereas if we have the behavior of >10k, we would want to remove the user from consideration in the average

ANich: Agreed, just added a test that revealed that. We could filter the probes that fall outside the threshold altogether?

sql/moz-fx-data-shared-prod/telemetry_derived/urlbar_clients_daily_v1/query.sql
```diff
@@ -2,12 +2,30 @@ CREATE TEMP FUNCTION one_index(x ANY TYPE) AS (
...
+
+CREATE TEMP FUNCTION filter_probe_counts_with_outlier_values(map ARRAY<STRUCT<k STRING, v INT64>>) AS (
+  -- Similar to search probes/metrics, count values over 10000 are considered too high
+  ARRAY(
+    SELECT
+      STRUCT(record.k AS key, record.v as value)
+    FROM
+      UNNEST(map) AS record
+    WHERE record.v <= 10000
+  )
```
ANich, line 21: RE: https://github.com/mozilla/bigquery-etl/pull/2321#discussion_r705411601 

Filter rather than set to 0 @teonbrooks @akkomar

sql/moz-fx-data-shared-prod/telemetry_derived/urlbar_clients_daily_v1/query.sql
```diff
@@ -100,7 +125,7 @@ count_picked AS (
     COALESCE(SUM(position.value), 0) AS count_picked_total,
     mozfun.map.sum(ARRAY_AGG(STRUCT(type AS key, position.value AS value))) AS count_picked_by_type,
     mozfun.map.sum(
-      ARRAY_AGG(STRUCT(one_index(position.key) AS key, position.value AS value))
+      ARRAY_AGG(STRUCT(position.key AS key, position.value AS value))
```
ANich, line 128: I noticed this while writing tests as well, it seems like this call was causing all the index -> position conversions to be done twice, since it's already done above. So 0 -> 2 instead of 0 -> 1. Let me know if I'm misunderstanding though @teonbrooks

sql/moz-fx-data-shared-prod/telemetry_derived/urlbar_clients_daily_v1/query.sql
```diff
@@ -2,12 +2,33 @@ CREATE TEMP FUNCTION one_index(x ANY TYPE) AS (
   CAST(IF(SAFE_CAST(x AS INT64) < 0, SAFE_CAST(x AS INT64), SAFE_CAST(x AS INT64) + 1) AS STRING)
```
teonbrooks, line 2: i wonder if this should be remain as an INT? are there any advantages to recasting as a string other than that's the original type?

sql/moz-fx-data-shared-prod/telemetry_derived/urlbar_clients_daily_v1/query.sql
rebecca-burwei: nice catch. removing it looks right to me

sql/moz-fx-data-shared-prod/telemetry_derived/urlbar_clients_daily_v1/query.sql
rebecca-burwei: If it's a string, then count_picked_by_type and count_picked_by_position have the same nested types. I don't feel strongly about this either way.

rebecca-burwei (approved):
looks good to me

sql/moz-fx-data-shared-prod/telemetry_derived/urlbar_clients_daily_v1/query.sql
teonbrooks: good point, let's leave it as string to reduce complexity. we can always cast to int later if needed

akkomar (approved):

sql/moz-fx-data-shared-prod/telemetry_derived/urlbar_clients_daily_v1/query.sql
teonbrooks: nice catch. agreed!

teonbrooks (approved):

teonbrooks:
thanks @ANich. one last thing, has this table been overwritten to replace the first one since the first one incorrectly added +2 to the position instead of +1?

PR #2323, opened 2021-09-09
Support analysis for rollouts in opmon (by treating a boolean pref as a branch)

emtwo:
This PR adds the ability to specify in an opmon project config file that a certain boolean pref will be used instead of regular experiment branches for doing the comparative analysis.

TODO:
- [x] Use the pref in the sql templates that comes from the config

Checklist for reviewer:

### DIFF
--- a/bigquery_etl/operational_monitoring/__init__.py
+++ b/bigquery_etl/operational_monitoring/__init__.py
@@ -72,16 +72,14 @@ def _query_up_to_date(dataset, slug, basename, project_last_modified):
 
 
 def _write_sql_for_data_type(
-    query, project, dataset, om_project, render_kwargs, probes, data_type
+    query, project, dataset, slug, render_kwargs, probes, data_type
 ):
-    normalized_slug = _bq_normalize_name(om_project["slug"])
+    normalized_slug = _bq_normalize_name(slug)
     render_kwargs.update(
         {
             "source": query["source"],
             "probes": _get_name_and_sql(query, probes, "probes", data_type),
-            "slug": om_project["slug"],
-            "channel": om_project["channel"],
-            "pref": om_project.get("boolean_pref"),
+            "slug": slug,
         }
     )
     _write_sql(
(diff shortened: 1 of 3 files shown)


bigquery_etl/operational_monitoring/templates/scalar_query.sql
```diff
@@ -1,46 +1,70 @@
...
+    {% if branches %}
+    {% for branch in branches %}
+      "{{ branch }}"
+      {{ "," if not loop.last else "" }}
     {% endfor %}
-    branch
+    {% else %}
+    "enabled", "disabled"
+    {% endif %}
+)
```
wlach, line 70:
```suggestion
)

```

bigquery_etl/operational_monitoring/__init__.py
```diff
@@ -134,7 +136,7 @@ def _generate_sql(project, dataset):
             project, bucket, blob.name
         )
         normalized_slug = _bq_normalize_name(om_project["slug"])
-        render_kwargs.update({"branches": om_project["branches"]})
+        render_kwargs.update({"branches": om_project.get("branches", [])})
```
wlach, line 139: Should we have some kind of assertion here that either `branches` or `pref` is defined, but (I assume) not both?

I think it's better for things to break in the case of unexpected input than to have unpredictable behaviour.

bigquery_etl/operational_monitoring/templates/histogram_query.sql
```diff
@@ -82,10 +91,14 @@ merged_histograms AS (
   CROSS JOIN
     UNNEST(metrics)
   WHERE branch IN (
+    {% if branches %}
```
wlach, line 94: Likewise, a quick comment on the conditional logic here might help someone looking at this code later

bigquery_etl/operational_monitoring/__init__.py
```diff
@@ -72,14 +72,16 @@ def _query_up_to_date(dataset, slug, basename, project_last_modified):
...
-    normalized_slug = _bq_normalize_name(slug)
+    normalized_slug = _bq_normalize_name(om_project["slug"])
     render_kwargs.update(
         {
             "source": query["source"],
             "probes": _get_name_and_sql(query, probes, "probes", data_type),
-            "slug": slug,
+            "slug": om_project["slug"],
+            "channel": om_project["channel"],
+            "pref": om_project.get("boolean_pref"),
```
wlach, line 84: I could be missing something, but I feel like it might make more logical sense to update the `om_projexct`-specific `render_kwargs` in the function calling this, rather than in this method.

wlach (changes requested):
(filling in for @scholtzan, who's away)

In general this makes sense to me although I have some minor suggestions. Is this behaviour documented somewhere?

bigquery_etl/operational_monitoring/__init__.py
emtwo: good idea! I added an assertion. This is still pretty much in prototype phase where no one is actively using it but it will be handy in the future if people are trying to update the json files.

bigquery_etl/operational_monitoring/__init__.py
emtwo: Good point! `slug` is here because the raw slug needs to be set right before writing the query sql file but the others don't need to be there.

emtwo:
> In general this makes sense to me although I have some minor suggestions. Is this behaviour documented somewhere?

Thanks for the suggestions! I've made some updates. The behaviour is not actually documented somewhere but it's based on Corey's previous fission dashboard changes here: https://github.com/mozilla/fission-monitoring-nightly/commit/eacdfbb4599109c9b1336b3a7beb4b54a24d32a0 to support rollouts.

wlach (approved):
👍

PR #2333, opened 2021-09-15
Bug 1730190 - filter out overactive clients in events_daily queries

akkomar:
This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1730190.
Cutoff threshold of 3M filters out two clients that had unusually high number of submissions.

Checklist for reviewer:


bigquery_etl/events_daily/query_templates/events_daily_v1/query.sql
```diff
@@ -17,13 +17,22 @@ WITH sample AS (
       UNNEST(e.events) AS event
     {% else %}
     SELECT
+      {% if app_id == "telemetry" %}
+      *,
+      COUNT(*) OVER (PARTITION BY submission_date, client_id) AS client_event_count
```
wlach, line 22: I feel like it would be good to add this same filtering/counting for glean apps, where we might encounter the same problem. Can you add this?

bigquery_etl/events_daily/query_templates/events_daily_v1/query.sql
```diff
@@ -32,6 +41,10 @@ WITH sample AS (
       OR (@submission_date IS NULL AND submission_date >= '{{ start_date }}')
     )
     AND client_id IS NOT NULL
+    {% if app_id == "telemetry" %}
+    -- filter out overactive client crashing the telemetry query
```
wlach, line 45:
```suggestion
    -- filter out overactive clients: they distort the data and can cause the job to fail: https://bugzilla.mozilla.org/show_bug.cgi?id=1730190
```

wlach (changes requested):
Thanks @akkomar! I have one requested change but overall LGTM

Also, feel free to add me as a co-owner for these queries. I'm honestly not super with familiar with them yet, but I will be continuing to work in this space.

bigquery_etl/events_daily/query_templates/events_daily_v1/query.sql
akkomar: I'd opt for not adding this to glean query at this point. Distribution of event counts per client in this table does not have such long tail like in telemetry. I don't really know what cutoff level would make sense in the long term.

akkomar:
> Thanks @akkomar! I have one requested change but overall LGTM
> 
> Also, feel free to add me as a co-owner for these queries. I'm honestly not super with familiar with them yet, but I will be continuing to work in this space.

Great! Added you as an owner.

See my comment inline about glean.

bigquery_etl/events_daily/query_templates/events_daily_v1/query.sql
wlach: I feel like it makes sense to have some kind of sanity check here, just to guard against something weird happening unexpectedly in the future. We can deliberately set the threshold high for now (3,000,000 seems like a fine number to me). That said, I don't feel strongly enough about this to block merging this as-is.

wlach (approved):

PR #2334, opened 2021-09-15
Update jackknife_percentile_ci() for use with scalar float buckets

emtwo:
For some background/context:

The motivation of this PR is that the `jackknife_percentile_ci()` function can get really slow when it's applied to very long lists of numbers since it produces more subsamples with larger amounts of data and computes percentiles for each subsample.

A list of scalars with a wide variety of numbers was taking a long time (3min+) to compute these percentiles. By bucketing the scalars, a histogram with fewer numbers takes less time to compute percentiles on.

I used the same [scalar bucketing strategy used in glam](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/telemetry_derived/clients_scalar_probe_counts_v1/query.sql#L69-L86) which allows for buckets that are more appropriate per metric. However, the buckets are more precise so they are floats. So I updated the `jackknife_percentile_ci()` function to accept floats.

This PR also moves the flattening of scalars (i.e. `unnest(metrics)`) out of the scalar view and into the scalar query since it needs to be unnested to compute the buckets. This required a schema change so I'll be migrating into the new table when this PR merges.

### DIFF
--- a/bigquery_etl/operational_monitoring/templates/histogram_view.sql
+++ b/bigquery_etl/operational_monitoring/templates/histogram_view.sql
@@ -43,7 +43,7 @@ WITH normalized AS (
         branch,
         probe)
 
--- Cast histograms to have INT64 keys
+-- Cast histograms to have FLOAT64 keys
 -- so we can use the histogram jackknife percentile function.
 SELECT
     client_id,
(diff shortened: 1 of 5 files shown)


wlach (approved):
👍 Chatted about this over Zoom. Looks good! This makes sense to me, since we still have approximately the same amount of data to sample from (we're just changing how it's stored).

PR #2348, opened 2021-09-17
Replace `format_sql` with `bqetl format`

wlach:
* Make bqetl work equivalently to format_sql (which had some extra
options to format standard in)
* Remove `format_sql` and update everything that uses it to use
  `bqetl format` instead.


wlach:
image

relud (approved):

PR #2354, opened 2021-09-20
Support multiple data sources with a different query file for each.

emtwo:
The goal of this PR is to support multiple sources (e.g. crash table and main table) outputting data to the same data type table (e.g. histogram table). So some histograms can come from the crash table and some can come from the main table but they both end up in the same output table for the purpose of operational monitoring.

Checklist for reviewer:


scholtzan (approved):
LGTM! 👍

PR #2360, opened 2021-09-21
clients_daily_joined and clients_last_seen with first_seen_date

scholtzan:
Needed for https://github.com/mozilla/lookml-generator/pull/240 and based on some of the discussion in https://docs.google.com/document/d/10m3WryXZZtHMyI7GTeSALQo54-5XFVZm8bC3MI7coy8/edit#

Checklist for reviewer:


sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_joined_v1/query.sql
```diff
@@ -6,6 +19,10 @@ LEFT JOIN
   telemetry_derived.clients_daily_event_v1 AS cde
 USING
   (submission_date, sample_id, client_id)
+LEFT JOIN
+  first_seen
+USING
+  (submission_date, sample_id, client_id)
```
scholtzan, line 25: This is based on some of the discussion in https://docs.google.com/document/d/10m3WryXZZtHMyI7GTeSALQo54-5XFVZm8bC3MI7coy8/edit#

I'm wondering if it makes sense to add the first_seen date in `clients_daily_joined`? As far as I understand, eventually the `clients_daily` view will query this table (?), so it might be convenient to have the first seen dates be part of that.

Alternatively, we could also have a separate query (`clients_daily_with_first_seen`?) that executes the `first_seen` query.

jklukas: > eventually the clients_daily view will query this table

Yes, there are no blockers to making that change. We can swap this in whenever we feel like it's valuable.

Have you evaluated whether it would be more efficient to join with `clients_first_seen_v1` here rather than `clients_last_seen`? What you have here already is probably better and more extensible, but worth checking.

Perhaps my biggest concern here is just the extra dependency in the execution graph, since this table now can't be prepared until the `last_seen` query finishes. Something to watch out for in terms of DAG execution time.

jklukas (approved):
r+wc

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_joined_v1/query.sql
scholtzan: Yes, I checked if `clients_first_seen_v1` would be more efficient. It is only slightly faster time-wise when running the query (it only takes a couple of seconds to run anyways) and processes only a couple fewer GB. I think being more flexible here is probably worth this slightly worse efficiency.

PR #2351, opened 2021-09-20
Bug 1731093 - Generate events_daily for other Fenix variants

wlach:
In particular, Firefox for Android beta and nightly channels.

PR #2368, opened 2021-09-22
Derived dataset metadata for Fenix

wlach:
For testing (and eventually deploying) #2351 --
https://bugzilla.mozilla.org/show_bug.cgi?id=1731093


jklukas (approved):
r+

PR #2377, opened 2021-09-23
Use clients_daily_joined_v1 for telemetry.clients_daily

scholtzan:
All the data has been backfilled (https://bugzilla.mozilla.org/show_bug.cgi?id=1732155) so there shouldn't be anything blocking us from using `telemetry_derived.clients_daily_joined_v1` as basis for `telemetry.clients_daily`.

Checklist for reviewer:

### DIFF
--- a/dags/bqetl_devtools.py
+++ b/dags/bqetl_devtools.py
@@ -84,10 +84,10 @@ with DAG(
         wait_for_copy_deduplicate_main_ping
     )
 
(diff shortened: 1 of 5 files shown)


jklukas (approved):
I'm pretty sure that none of the queries here do a SELECT * from telemetry.clients_daily (in which case we would need to update schemas). All the `SELECT *` queries reference `clients_daily_v6` directly.

PR #2381, opened 2021-09-27
Add perms for contextual-services views

jklukas:
Access to these views is currently broken since we removed dataset-level ACLs.

`quicksuggest-impression` is already handled with a specific authorized view.

Checklist for reviewer:


whd:
This is probably more technically correct than what we've been doing for e.g. `telemetry` but the other option is to make the views dataset accessible to `workgroup:mozilla-confidential` since it should only contain views and can rely on downstream authorizations.

jklukas:
> the other option is to make the views dataset accessible to workgroup:mozilla-confidential since it should only contain views and can rely on downstream authorizations.

I would be happy to make this dataset accessible to `workgroup:contextual-services`, but we cannot make it mozilla-confidential because it contains an authorized view for `quicksuggest_impression` that isn't relying on downstream authorizations.

whd (approved):

whd:
This has been applied.

PR #2384, opened 2021-09-27
Mozfun udf for additional percentiles

alekhyamoz:
This is 1st step to resolve the [issue](https://github.com/mozilla/glam/issues/1576). 

The PR will create a udf in mozfun that will allow us to calculate the 99.9 and 99 percentiles (scale of 1)


sql/mozfun/glam/map_from_array_offsets_v2/udf.sql
```diff
@@ -0,0 +1,45 @@
...
+  -- required ordered
+  assert.array_equals(
+    ARRAY<STRUCT<key STRING, value FLOAT64>>[("0", 2.0), ("2", 4.0)],
+    glam.map_from_array_offsets([2.0, 0.0], [2.0, 3.0, 4.0])
+  ),
+  -- intended use-case, for approx_quantiles
+  assert.array_equals(
+    ARRAY<STRUCT<key STRING, value FLOAT64>>[("25", 1.0), ("50", 2.0), ("75", 3.0)],
+    glam.map_from_array_offsets(
+      [25.0, 50.0, 75.0],
```
emtwo, line 36: This should be able to handle 99.9% right? We should add that to the test.

sql/mozfun/glam/map_from_array_offsets_v2/udf.sql
```diff
@@ -0,0 +1,45 @@
+-- udf_get_values
+CREATE OR REPLACE FUNCTION glam.map_from_array_offsets_v2(
```
emtwo, line 2: We could call this something more specific to reflect that it help with higher precision. E.g. `map_from_array_offsets_precise`

sql/mozfun/glam/map_from_array_offsets_v2/udf.sql
```diff
@@ -0,0 +1,45 @@
...
+      )
+    FROM
+      UNNEST(required) AS key
+  )
+);
+
+SELECT
+  assert.array_equals(
+    ARRAY<STRUCT<key STRING, value FLOAT64>>[("0", 2.0), ("2", 4.0)],
+    glam.map_from_array_offsets([0.0, 2.0], [2.0, 3.0, 4.0])
```
emtwo, line 25: These tests should be calling `glam.map_from_array_offsets_v1` or whatever the new function name is. Is this the issue you had mentioned about permissions? I thought you meant it was being called from somewhere else. In the tests it should work. If it's not working, you may need to authenticate.

PR #2391, opened 2021-09-29
added the udf for scale 1 precision

alekhyamoz:
This is 1st step to resolve the [issue](https://github.com/mozilla/glam/issues/1576)

The udf will allow us to get values for 99 and 99.9 keys

### DIFF
--- a/sql/mozfun/glam/map_from_array_offsets_precise/udf.sql
+++ b/sql/mozfun/glam/map_from_array_offsets_precise/udf.sql
@@ -1,0 +1,45 @@
+-- udf_get_values
+CREATE OR REPLACE FUNCTION glam.map_from_array_offsets_precise(
+  required ARRAY<FLOAT64>,
+  `values` ARRAY<FLOAT64>
+)
+RETURNS ARRAY<STRUCT<key STRING, value FLOAT64>> AS (
+  (
+    SELECT
+      ARRAY_AGG(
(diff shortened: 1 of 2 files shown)


sql/mozfun/glam/map_from_array_offsets_precise/udf.sql
```diff
@@ -0,0 +1,45 @@
...
+  -- required ordered
+  assert.array_equals(
+    ARRAY<STRUCT<key STRING, value FLOAT64>>[("0", 2.0), ("0.2", 4.0)],
+    glam.map_from_array_offsets_precise([2.0, 0.0], [2.0, 3.0, 4.0])
+  ),
+  -- intended use-case, for approx_quantiles
+  assert.array_equals(
+    ARRAY<STRUCT<key STRING, value FLOAT64>>[("25", 1.0), ("50", 2.0), ("75", 3.0), ("99.9", 4.0)],
+    glam.map_from_array_offsets_precise(
+      [250.0, 500.0, 750.0, 999.0],
```
emtwo, line 36: Can we change this to `["25.0", "50.0", "75.0", "99.9"]` and then above on line 12 it can be `key * 10` and line 11 doesn't need to divide by 10.

I think this will be just more clear that the input to the function is the required percentiles.

emtwo (commented):
Just one more small change but otherwise this looks good!

emtwo (approved):

PR #2392, opened 2021-09-29
Add gcloud config set step for running tests

wlach:
(via @ANich) It feels slightly wrong to have to set this value twice (once via environment variable and once via gcloud).

Checklist for reviewer:

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -24,7 +24,8 @@ mvn dependency:copy-dependencies
 
 # run integration tests with 4 workers in parallel
 gcloud auth application-default login # or set GOOGLE_APPLICATION_CREDENTIALS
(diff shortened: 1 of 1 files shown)


jklukas:
> It feels slightly wrong to have to set this value twice (once via environment variable and once via gcloud).

I don't think setting the env var is necessary anymore if we're setting the project via gcloud.

scholtzan:
Yeah, we currently need the env variable for running the tests: https://github.com/mozilla/bigquery-etl/blob/9bca48821a8a0d40b1700cc14ecd8068d132ed06/conftest.py#L50 It's possible we can update this and determine the project some other way.

scholtzan (approved):

wlach:
I filed an issue to fix this at some point: https://github.com/mozilla/bigquery-etl/issues/2393

PR #2404, opened 2021-10-01
Optimize enrollment_other_events_overall

scholtzan:
Querying `telemetry.experiment_enrollment_other_events_overall` outside of shared-prod is not possible anymore due to the query exceeding complexity limitations. Rewriting the query fixed the problem.

Checklist for reviewer:

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_other_events_overall_v1/view.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_other_events_overall_v1/view.sql
@@ -20,11 +20,12 @@ SELECT
   window_start AS `time`,
   experiment,
   branch,
-  event.event AS event,
-  SUM(event.count) AS value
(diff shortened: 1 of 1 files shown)


sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_other_events_overall_v1/view.sql
```diff
@@ -6,55 +6,25 @@ WITH pivot AS (
...
   experiment,
   branch,
-  event,
-  SUM(`count`) AS value
+  event.event AS event,
+  SUM(event.count) AS value
 FROM
-  pivot
+  pivot,
+  UNNEST(events) AS event
```
jklukas, line 27: I'd prefer an explicit CROSS JOIN here, and calling this `e` rather than `event` so that we can have `e.event` rather than `event.event`

jklukas (approved):
r+wc

PR #2407, opened 2021-10-04
Bug 1733986 Fix date logic for adm_weekly_aggregates

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1733986

The intention of this query is to each day populate a partition that
windows over the previous 7 days. The GROUP BY should not include
`submission_date` and we need to inject the `@submission_date`
parameter into the query instead.

Checklist for reviewer:


wlach (approved):
A bit confusing but I think it makes sense!

PR #2416, opened 2021-10-07
Add normalized_channel to generated metrics datasets

scholtzan:
Having `normalized_channel` will probably come in handy

Checklist for reviewer:

### DIFF
--- a/bigquery_etl/glean_usage/templates/clients_last_seen_joined.metadata.yaml
+++ b/bigquery_etl/glean_usage/templates/clients_last_seen_joined.metadata.yaml
@@ -15,4 +15,5 @@ bigquery:
     require_partition_filter: true
   clustering:
     fields:
+      - normalized_channel
       - sample_id
--- a/bigquery_etl/glean_usage/templates/metrics_clients_daily.query.sql
+++ b/bigquery_etl/glean_usage/templates/metrics_clients_daily.query.sql
@@ -6,6 +6,11 @@
     DATE(submission_timestamp) AS submission_date,
     client_info.client_id AS client_id,
     sample_id,
+    {% if app_name == "fenix" -%}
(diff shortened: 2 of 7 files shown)


bigquery_etl/glean_usage/templates/clients_last_seen_joined.metadata.yaml
```diff
@@ -16,3 +16,4 @@ bigquery:
   clustering:
     fields:
       - sample_id
+      - normalized_channel
```
jklukas, line 19: Stable tables are clustered on `normalized_channel, sample_id` rather than the other way around. I think we should be consistent with that.

bigquery_etl/glean_usage/templates/metrics_clients_daily.query.sql
```diff
@@ -6,6 +6,7 @@
     DATE(submission_timestamp) AS submission_date,
     client_info.client_id AS client_id,
     sample_id,
+    normalized_channel,
```
jklukas, line 9: I expect `normalized_channel` here is often null. I think we want to inject `channel` here from the probeinfo data instead.

bigquery_etl/glean_usage/templates/metrics_clients_daily.query.sql
```diff
@@ -6,6 +6,11 @@
     DATE(submission_timestamp) AS submission_date,
     client_info.client_id AS client_id,
     sample_id,
+    {% if app_name == "fenix" -%}
+    mozfun.norm.fenix_app_info("{{ dataset }}", client_info.app_build).channel AS normalized_channel,
```
scholtzan, line 10: For fenix it's probably better to get the channel from the dataset

jklukas (approved):

Issue #2313, opened 2021-08-31
Write a plan for moving all generated content out of `main`


jklukas:
Per https://github.com/mozilla/bigquery-etl/pull/2307 we currently have a few different approaches to query generation coexisting in this codebase, which causes some confusion.

The ideal state is to have no generated code check into `main`, relying on CI to populate generated content in the `generated-sql` branch. But there are several realities today preventing that.

Chiefly, we run DAG generation separately from sql generation, which means that generated queries can't be included in the generated DAGs. We could consider moving DAG generation after SQL generation, but this may lead to overly long build times.

Similarly, our testing framework for exercising queries assumes that queries are checked in to `main`.

A proposal is needed here to figure out the full scope of changes needed to support a good workflow that pushes all SQL generation to CI.

scholtzan:
I started putting a proposal together: https://docs.google.com/document/d/1k9Gv_zJyWAVDhuR8vkEizDxfko35jyU6Shb4EPsBfzs/edit#

scholtzan:
Implementing the proposal will be broken out in the following steps (separate PRs):
- [ ] Move query generators out of `bigquery_etl/` into `sql_generators/`
- [ ] Update the CI to run the SQL + DAG generation
- [ ] Make changes to show diff when opening a PR
- [ ] Update CLI
- [ ] Remove generated query contents from `main` (until this point the generated queries will simply be overwritten when running the SQL generation, this helps to prevent breakages)
- [ ] Change Airflow DAG sync script to pick up generated DAGs from the `generated-sql` branch
- [ ] Remove generated DAGs from `main`
- [ ] Add support for testing generators

PR #2452, opened 2021-10-21
Speed up generate_sql

scholtzan:
This is one piece for working towards https://github.com/mozilla/bigquery-etl/issues/2313 for which I started drafting a proposal.

How do we feel about reducing the time for running `generate_sql` by 3x 😉 
The task now takes about 2.30m vs 8.30m

### DIFF
--- a/bigquery_etl/cli/glean_usage.py
+++ b/bigquery_etl/cli/glean_usage.py
@@ -128,13 +128,14 @@ def generate(project_id, output_dir, parallelism, exclude, only, app_name):
         for table in GLEAN_TABLES
     ]
 
-    # Parameters to generate per-app_id datasets consist of the function to be called
+    # Parameters to generate per-app datasets consist of the function to be called
     # and app_info
     generate_per_app = [
         (
             partial(table.generate_per_app, project_id, output_dir=output_dir),
-            app_info[0],
+            info,
         )
+        for info in app_info
         for table in GLEAN_TABLES
     ]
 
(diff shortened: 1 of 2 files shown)


bigquery_etl/view/generate_stable_views.py
```diff
@@ -345,15 +345,14 @@ def main():
...
-    with ThreadPool(args.parallelism) as pool:
+    with Pool(args.parallelism) as pool:
         pool.map(
             partial(
                 write_view_if_not_exists,
                 args.target_project,
                 Path(args.sql_dir),
             ),
             schemas,
-            chunksize=1,
```
scholtzan, line 356: Was there a reason for setting `chunksize=1` here? It does slow the whole process down by a lot.

bigquery_etl/cli/glean_usage.py
```diff
@@ -104,19 +104,37 @@ def generate(project_id, output_dir, parallelism, exclude, only, app_name):
...
+    generate_per_app = [
+        (
+            partial(table.generate_per_app, project_id, output_dir=output_dir),
+            app_info[0],
+        )
+        for table in GLEAN_TABLES
+    ]
+
+    with Pool(parallelism) as pool:
+        pool.starmap(run_generate, generate_per_app_id + generate_per_app)
```
scholtzan, line 135: It's more efficient running all the tasks in parallel instead of iterating over the Glean tables and running the generation for each Glean table in parallel and waiting for it to finish every time.

PR #2471, opened 2021-11-02
DVPN-69 - Add Wave IV countries

relud:
https://mozilla-hub.atlassian.net/browse/DVPN-69


relud:
this won't require backfill, but the impacted tables are wholly updated on each run.

ANich (approved):

PR #2476, opened 2021-11-04
add glam_etl to non user facing view

alekhyamoz:
Removing the multitude of tables in mozdata.glam_etl which are interim/raw tables that we would not want analysts to find/use.

### DIFF
--- a/bigquery_etl/view/__init__.py
+++ b/bigquery_etl/view/__init__.py
@@ -50,6 +50,7 @@ NON_USER_FACING_DATASET_SUFFIXES = (
     "_external",
     "_bi",
     "_restricted",
+    "glam_etl",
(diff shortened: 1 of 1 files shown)


scholtzan (approved):

whd:
I think this PR is unnecessary since these views are managed by syndication, see https://bugzilla.mozilla.org/show_bug.cgi?id=1643511 and https://docs.google.com/document/d/13EKQIB6P2bca8gAKtZ43DjNY1Iyqvl9oYoNi4vB9yto/edit#heading=h.i7zexsmcxlle for some background. https://mozilla-hub.atlassian.net/browse/DSRE-293 should capture removing the configuration for that.

PR #2478, opened 2021-11-04
Increase event_property threshold

scholtzan:
I'm not sure if increasing this threshold has any implications?

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/event_types_history_v1/init.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/event_types_history_v1/init.sql
@@ -108,7 +108,7 @@ per_event_property AS (
   USING
     (category, event, event_property)
   WHERE
-    event_property_value_index <= 1000
+    event_property_value_index <= 5000
   GROUP BY
     category,
     event,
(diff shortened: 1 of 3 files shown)


sql/moz-fx-data-shared-prod/telemetry_derived/event_types_history_v1/init.sql
```diff
@@ -108,7 +108,7 @@ per_event_property AS (
   USING
     (category, event, event_property)
   WHERE
-    event_property_value_index <= 1000
+    event_property_value_index <= 5000
```
scholtzan, line 111: Also open to setting a different threshold

wlach (changes requested):
I'm not aware of any issues with increasing the threshold. However, I believe the right approach is to modify the templating here and regenerate the queries:

https://github.com/mozilla/bigquery-etl/blob/821f9d45dc79c869bb3285c55ec1c7dcf0459fec/bigquery_etl/events_daily/query_templates/event_types_history_v1/templating.yaml#L12-L17

scholtzan:
Right, right

wlach (approved):
r+ with the yaml changes to non-Firefox reverted

PR #2480, opened 2021-11-04
Added JDK install to python-deps stage for pyjnius build on M1

ANich:
On Apple Silicon docker build isn't locating java when building pyjnius

Checklist for reviewer:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,11 @@ orbs:
   gcp-gcr: circleci/gcp-gcr@0.13.0
   docker: circleci/docker@1.5
 
+executors:
+  custom-ubuntu-executor:
+    machine:
+      image: ubuntu-2004:202111-02
+
 jobs:
   build:
     docker: &docker
@@ -361,11 +366,9 @@ jobs:
               || echo "Skipping push since it looks like there were no changes"
 
   deploy:
-    executor: docker/docker
+    executor: custom-ubuntu-executor
     steps:
       - checkout
-      - setup_remote_docker:
-          docker_layer_caching: true
       - *attach_generated_sql
       - *copy_generated_sql
       - when:
@@ -450,7 +453,7 @@ jobs:
               || echo "Skipping push since it looks like there were no changes"
           # yamllint enable rule:line-length
   deploy-to-private-gcr:
-    executor: gcp-gcr/default
+    executor: custom-ubuntu-executor
     steps:
       - checkout
       - *attach_generated_sql
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ orbs:
   docker: circleci/docker@1.5
 
 executors:
-  custom-ubuntu-executor:
+  ubuntu-machine-executor:
     machine:
       image: ubuntu-2004:202111-02
 
@@ -366,7 +366,7 @@ jobs:
               || echo "Skipping push since it looks like there were no changes"
 
   deploy:
-    executor: custom-ubuntu-executor
+    executor: ubuntu-machine-executor
     steps:
       - checkout
       - *attach_generated_sql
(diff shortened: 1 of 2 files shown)


kik-kik:
@jklukas did you have the same issue? Should we merge this or close the PR?
@haroldwoo any thoughts?

jklukas:
cc @relud 

I don't think I ever tried to do a docker build with this repo on my M1. I'm testing now.

I'm wondering if we should instead use:

```
FROM --platform=linux/amd64 python:${PYTHON_VERSION}-slim-buster AS base
```

which I believe would consistently make the image build for x86.

relud:
> I'm wondering if we should instead use:
> 
> ```
> FROM --platform=linux/amd64 python:${PYTHON_VERSION}-slim-buster AS base
> ```
> 
> which I believe would consistently make the image build for x86.

that sounds good to me

kik-kik:
> cc @relud
> 
> I don't think I ever tried to do a docker build with this repo on my M1. I'm testing now.
> 
> I'm wondering if we should instead use:
> 
> ```
> FROM --platform=linux/amd64 python:${PYTHON_VERSION}-slim-buster AS base
> ```
> 
> which I believe would consistently make the image build for x86.

ah cool, didn't realise you could do that. I'll modify this now and create a PR. Thanks 🙇

kik-kik:
Hey @ANich, apologies. I just realised that it was you who originally opened the draft and not me 🙈 I pushed the changes that Jeff suggested. Looks like this should work.

relud (approved):

jklukas (approved):

jklukas (changes requested):
Looks like CI is still failing with this change: https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/12679/workflows/31f14c6c-0fd2-4b43-a2b3-5e5a33471473/jobs/94605

.circleci/config.yml
```diff
@@ -447,22 +447,26 @@ jobs:
               || echo "Skipping push since it looks like there were no changes"
           # yamllint enable rule:line-length
   deploy-to-private-gcr:
-    executor: gcp-gcr/default
+    executor: docker/docker
```
jklukas, line 450: The default gcp-gcr executor looks to be `machine` with defaults: https://circleci.com/developer/orbs/orb/circleci/gcp-gcr#executors-default

To keep things as similar as possible, I think this could be:

```
executors:
  my-executor:
    machine:
      image: ubuntu-2004:202111-02
```

Perhaps that would avoid having to make some of the changes below?

If this all gets too complicated, I wouldn't be opposed to reverting back to adding jdk to the Dockerfile.

kik-kik:
@jklukas so this seems to work for both push to gcr and deploy ci jobs. Does that work? If you're happy with this I will remove the commented out line and squash the commits.

.circleci/config.yml
```diff
@@ -5,6 +5,11 @@ orbs:
   gcp-gcr: circleci/gcp-gcr@0.13.0
   docker: circleci/docker@1.5
 
+executors:
+  custom-ubuntu-executor:
```
jklukas, line 9: Maybe call this `ubuntu-machine-executor` since the `machine` part is significant to what this is doing for us.

Dockerfile
```diff
@@ -5,7 +5,8 @@ ARG PYTHON_VERSION=3.8
 # has updated coreutils that require a newer linux kernel than provided by CircleCI, per
 # https://forums.docker.com/t/multiple-projects-stopped-building-on-docker-hub-operation-not-permitted/92570/6
 # and https://forums.docker.com/t/multiple-projects-stopped-building-on-docker-hub-operation-not-permitted/92570/11
-FROM python:${PYTHON_VERSION}-slim-buster AS base
+# --platform=linux/amd64 added for Apple Silicon support
```
jklukas, line 8:
```suggestion
# --platform=linux/amd64 added to prevent pulling ARM images when run on Apple Silicon
```

.circleci/config.yml
```diff
@@ -361,11 +366,9 @@ jobs:
               || echo "Skipping push since it looks like there were no changes"
 
   deploy:
-    executor: docker/docker
+    executor: custom-ubuntu-executor
     steps:
       - checkout
-      - setup_remote_docker:
-          docker_layer_caching: true
```
jklukas, line 368: This is a nice simplification 👍

jklukas (approved):
r+wc

PR #2489, opened 2021-11-10
Add an events unnested table for Mozilla VPN

wlach:
This is hand-generated for now, because the Mozilla VPN "events" ping
is actually named main, which the current glean generation machinery
doesn't quite understand.

For https://github.com/mozilla/lookml-generator/issues/299


sql/moz-fx-data-shared-prod/mozilla_vpn/events_unnested/view.sql
```diff
@@ -0,0 +1,29 @@
...
+      ping_info.parsed_start_time,
+      ping_info.parsed_end_time
+    ) AS ping_info
+  ),
+  event.timestamp AS event_timestamp,
+  event.category AS event_category,
+  event.name AS event_name,
+  event.extra AS event_extra,
+FROM
+  `moz-fx-data-shared-prod.mozillavpn.main` AS e
```
scholtzan, line 27:
```suggestion
  `moz-fx-data-shared-prod.mozilla_vpn.main` AS e
```

I guess `mozilla_vpn.main` does reference `mozillavpn.main`, but it might be less confusing to use `mozilla_vpn` here. I thought at first `mozillavpn` is a typo. I guess for some historical reasons there are two vpn datasets.

scholtzan (approved):

sql/moz-fx-data-shared-prod/mozilla_vpn/events_unnested/view.sql
wlach: Agreed. `mozillavpn` refers to the app id but using the application-wide dataset makes more sense here, I think. 

https://github.com/mozilla/probe-scraper/blob/e6725dca9ed9b97d8cb27762a69b98498e8d8552/repositories.yaml#L691

PR #2494, opened 2021-11-16
Bug 1740392 - Add static iana_tls_cipher_suites table

whd:
Checklist for reviewer:


sql/moz-fx-data-shared-prod/static/iana_tls_cipher_suites/schema.json
```diff
@@ -0,0 +1,27 @@
+[
+  {
+    "mode": "NULLABLE",
+    "name": "Value",
```
jklukas, line 4: Can we make these field names all lower_snake_case? We're skipping the header row in the file, so I don't think there's any reason to be restricted to the names in the CSV itself.

I also think it's worth adding descriptions to each of these fields here.

sql/moz-fx-data-shared-prod/static/iana_tls_cipher_suites/schema.json
```diff
@@ -0,0 +1,32 @@
+[
+  {
+    "mode": "NULLABLE",
+    "description": "Hex value assigned to the TLS cipher",
```
jklukas, line 4:
```suggestion
    "description": "Hex value assigned to the TLS cipher in format like \"0x00,0x84\"; note some values are ranges or contain wildcards",
```

jklukas (approved):
Thanks for the changes. One suggested update to docs.

PR #2495, opened 2021-11-16
Bug 1740392 - Add x_lb_tags udf normalization

whd:
Checklist for reviewer:


sql/moz-fx-data-shared-prod/regrets_reporter/regrets_reporter_update/view.sql
```diff
@@ -20,7 +20,13 @@ SELECT
...
                 FROM
                   UNNEST(SPLIT(metadata.header.x_source_tags, ',')) t
-              ) AS parsed_x_source_tags
+              ) AS parsed_x_source_tags,
+              ARRAY(
+                SELECT
+                  TRIM(t)
+                FROM
+                  UNNEST(SPLIT(metadata.header.x_lb_tags, ',')) t
+              ) AS parsed_x_lb_tags
```
jklukas, line 29: Given that order is significant here, I think we could make this a struct with named fields rather than a list. Something like:

```
STRUCT(
  TRIM(SPLIT(metadata.header.x_lb_tags, ','))[SAFE_OFFSET(0)]) AS tls_version,
  TRIM(SPLIT(metadata.header.x_lb_tags, ','))[SAFE_OFFSET(1)]) AS ciphersuite_code
) AS parsed_x_lb_tags
```

whd: Yeah I mentioned this in https://bugzilla.mozilla.org/show_bug.cgi?id=1736466#c2 / https://bugzilla.mozilla.org/show_bug.cgi?id=1740392 but didn't get any feedback so I decided to implement the "least different" thing. A struct does make more sense here so I'll update the PR.

jklukas: As to naming, my best suggestion is `tls_version` and `tls_cipher_hex` for these values we extract.

whd: Do you think it' worth normalizing to `0xXX,0xXX` for `tls_cipher_hex` here to match the RFC format in https://github.com/mozilla/bigquery-etl/pull/2494/ or leave that for for downstream queries?

jklukas: The `0xXX,0xXX` format seems really odd, and some of those entries are ranges including `-`. I think we can leave that to downstream queries. We may in the future want to do some post-processing of the official CSV to produce a static table that includes the four-char hex string and would be directly joinable to this value.

jklukas (approved):

PR #2498, opened 2021-11-16
Don't exit early when references for a table already exist in CI

wlach:
We were exiting from the entire process of reference-generation if we
found an existing reference (and `--skip-existing` was true) rather
than just skipping this directory as we should have been.


wlach:
Verified that this fixes the missing references (even without #2497) by re-running the generate-sql job with ssh.

relud (approved):

PR #2501, opened 2021-11-16
Skip generating glean datasets for regrets_reporter (fixes #2499)

### LINKED ISSUE #2499: Issue with view suffix on views Generated via bigquery_etl.glean_usage.GleanAppPingViews
https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/695ca208a5dc7a8abfdee0d6b9f9013d41db2560

This errored (but for possibly an unrelated reason):

```

    |    .    |    .    |    .    |    .    |    .    |    .    |    .    |
   1:-- Generated by Generated via bigquery_etl.glean_usage.GleanAppPingViews
   2:CREATE OR REPLACE VIEW
   3:  `moz-fx-data-shared-prod.regrets_reporter.deletion_request`
   4:AS
   5:SELECT
   6:  *
   7:FROM
   8:  `moz-fx-data-shared-prod.regrets_reporter_ucs.deletion_request`
   9:
```

I think this is wrong and the computation of suffix for `regrets_reporter_ucs` is thinking `regrets_reporter` is its user-facing view, when in actuality `regrets_reporter` and `regrets_reporter_ucs` are two different user-facing namespaces.

wlach:


bigquery_etl/cli/glean_usage.py
```diff
@@ -31,7 +31,11 @@
     metrics_clients_last_seen.MetricsClientsLastSeen(),
     clients_last_seen_joined.ClientsLastSeenJoined(),
 ]
-SKIP_APPS = ["mlhackweek_search"]
+
+# regrets_reporter currently refers to two applications, skip the glean
+# one to avoid confusion: https://github.com/mozilla/bigquery-etl/issues/2499
+# mlhackweek_search was an experiment that we don't want to generate tables for
```
wlach, line 37: @jklukas do you have any idea why mlhackweek_search is here? It looks like it was added in https://github.com/mozilla/bigquery-etl/pull/2416

jklukas: This is probably due to permissions. `mlhackweek_search` data was considered sensitive and locked down to a specific google group. But I can't exactly remember what the failure scenario was. Likely airflow would have been able to run this, but local testing with an individual dev's account would have failed.

jklukas (approved):

bigquery_etl/cli/glean_usage.py
wlach: Ok, I think the current comment probably gives enough context.

PR #2503, opened 2021-11-16
Add regrets-reporter-ucs dataset and ACL metadata

whd:
https://github.com/mozilla-services/cloudops-infra/pull/3546 was sufficient to allow these schemas to be deployed, but once bqetl automation started to generate metadata as well we wound up in an inconsistent generated state.

This seems like a reasonable place to mention https://bugzilla.mozilla.org/show_bug.cgi?id=1740483#c5 / https://bugzilla.mozilla.org/show_bug.cgi?id=1654078#c45. I didn't include views here as I'm primarily concerned with unblocking the schemas pipeline, but maybe that should be considered.

Checklist for reviewer:


jklukas (approved):
Agreed that we should add geo restrictions at the view level, but can handle that in a separate PR.

PR #2507, opened 2021-11-17
Bug 1731310 - Added newtab related fields to feature_usage

ANich:
RE: https://bugzilla.mozilla.org/show_bug.cgi?id=1731310 - Added newtab related fields to `feature_usage` table

Do this look reasonable? I wasn't able to add:
- The shortcut rows pref because it isn't sent by activity stream at the moment and will probably require a data review
- Pocket impression data because it doesn't have client ids so can't be joined here.

Related probe dictionary link for contextual services - [link](https://probes.telemetry.mozilla.org/?search=contextual.services&view=detail&probeId=scalar%2Fcontextual.services.topsites.click).

Also updated the template DAG  to `bqetl_feature_usage`

Checklist for reviewer:

### DIFF
--- a/bigquery_etl/feature_usage/templates/metadata.yaml
+++ b/bigquery_etl/feature_usage/templates/metadata.yaml
@@ -9,12 +9,13 @@ owners:
   - shong@mozilla.com
   - cdowhygelund@mozilla.com
   - ascholtz@mozilla.com
+  - anicholson@mozilla.com
 labels:
   incremental: true
   application: firefox
   schedule: daily
 scheduling:
-  dag_name: bqetl_main_summary
+  dag_name: bqetl_feature_usage
 bigquery:
   time_partitioning:
     field: submission_date
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/feature_usage_v2/query.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/feature_usage_v2/query.sql
@@ -53,6 +53,19 @@ WITH user_type AS (
     search_count_tagged_sap AS search_count_tagged_sap,
     search_count_urlbar AS search_count_urlbar,
     search_with_ads_count_all AS search_with_ads_count_all,
+    SAFE_CAST(user_pref_browser_newtabpage_enabled AS BOOL) AS newtabpage_disabled,
+    (
+      SELECT
+        SUM(IF(SPLIT(key, '_')[SAFE_OFFSET(0)] = 'newtab', value, 0))
+      FROM
+        UNNEST(contextual_services_topsites_impression_sum)
+    ) AS num_topsites_new_tab_impressions_sponsored,
+    (
+      SELECT
+        SUM(IF(SPLIT(key, '_')[SAFE_OFFSET(0)] = 'newtab', value, 0))
+      FROM
+        UNNEST(contextual_services_topsites_click_sum)
+    ) AS num_new_tab_topsites_clicks_sponsored,
   FROM
     telemetry.clients_last_seen
   WHERE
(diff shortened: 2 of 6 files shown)


sql/moz-fx-data-shared-prod/telemetry_derived/feature_usage_v2/query.sql
```diff
@@ -477,6 +510,7 @@ activity_stream_sessions AS (
     MAX(user_prefs & 2 = 0) AS activitystream_reported_topsites_off,
     MAX(user_prefs & 4 = 0) AS activitystream_reported_pocket_off,
     MAX(user_prefs & 8 = 0) AS activitystream_reported_highlights_off,
+    MAX(user_prefs & 32 = 0) AS activitystream_reported_sponsored_topstories_off,
```
teonbrooks, line 513: iiuc, if topstories is turned on at during any of the sessions on the submission_date, then it would be register as true. it's not the final state of the day?

bigquery_etl/feature_usage/templates/metadata.yaml
```diff
@@ -14,7 +14,7 @@ labels:
   application: firefox
```
wlach, line 14: I wonder if it's worth updating the description with anything new here  (e.g. the links to the probe dictionary)? You may also want to add yourself to `owners`

sql/moz-fx-data-shared-prod/telemetry_derived/feature_usage_v2/query.sql
```diff
@@ -714,6 +748,17 @@ SELECT
     COALESCE(search_with_ads_count_all, CAST(0 AS INT64)),
     NULL
   ) AS search_with_ads_count_all,
+  newtabpage_disabled,
+  IF(
+    '87' < app_version,
```
wlach, line 753: I guess this isn't a new issue, but this check seems likely to fail when we hit Firefox 100? This can be fixed in a followup, but IMO best that we do it soon so this doesn't catch us off guard.

sql/moz-fx-data-shared-prod/telemetry_derived/feature_usage_v2/schema.yaml
```diff
@@ -398,3 +398,27 @@ fields:
 - mode: NULLABLE
   name: activitystream_sessions_both
   type: INTEGER
+- mode: NULLABLE
+  name: newtabpage_disabled
+  type: BOOLEAN
```
wlach, line 403: Since we're here, can we add descriptions for these new fields?

Prior art: https://github.com/mozilla/bigquery-etl/blob/67119cff58df85accb0bfbf7c864a835ad5bc16d/sql/moz-fx-data-marketing-prod/iprospect/adspend_meta_v1/schema.yaml

bigquery_etl/feature_usage/templates/metadata.yaml
```diff
@@ -14,7 +14,7 @@ labels:
...
   schedule: daily
 scheduling:
-  dag_name: bqetl_main_summary
+  dag_name: bqetl_feature_usage
```
wlach, line 17: I'm slightly confused by this, it looks like we're already scheduling this as of the part `bqetl_feature_usage` dag? Is this just updating the definition to be in sync?

wlach (approved):
Looks like good to me, although I'm not familiar with the underlying data so am relying on @teonbrooks and @mmccorks opinion there. CI is failing due to a schema mismatch but I assume that's due to the fact that we need to do a new schema deploy before/when we merge this (I haven't personally done this before).

Could you add a note about that to common workflows?

https://mozilla.github.io/bigquery-etl/cookbooks/common_workflows/

bigquery_etl/feature_usage/templates/metadata.yaml
ANich: Yeah exactly, `bqetl_feature_usage` is the production dag. I think the prod dag was updated manually without the template being updated.

sql/moz-fx-data-shared-prod/telemetry_derived/feature_usage_v2/query.sql
ANich: Yeah I think that's currently the case with these prefs because of the `MAX`.. We could take the last ping value instead if you think that will be a more accurate representation. cc @irrationalagent who I think uses this dataset

sql/moz-fx-data-shared-prod/telemetry_derived/feature_usage_v2/schema.yaml
```diff
@@ -398,3 +398,49 @@ fields:
...
+  description: |-
+    Whether the newtabpage is disabled or not.
+    Based on the browser.newtabpage.enabled pref
+- mode: NULLABLE
+  name: num_topsites_new_tab_impressions_sponsored
```
ANich, line 408: Rename

ANich: I can't update this after a schema deploy 😔

sql/moz-fx-data-shared-prod/telemetry_derived/feature_usage_v2/query.sql
teonbrooks: I created https://github.com/mozilla/bigquery-etl/issues/2517 to track this

ANich:
> Looks like good to me, although I'm not familiar with the underlying data so am relying on @teonbrooks and @mmccorks opinion there. CI is failing due to a schema mismatch but I assume that's due to the fact that we need to do a new schema deploy before/when we merge this (I haven't personally done this before).
> 
> Could you add a note about that to common workflows?
> 
> https://mozilla.github.io/bigquery-etl/cookbooks/common_workflows/

Whoops! I merged without responding to this. I see this is noted here - https://mozilla.github.io/bigquery-etl/cookbooks/common_workflows/#update-an-existing-query

PR #2510, opened 2021-11-18
Road 134 - Add the proper permissions to the RS quicksuggest table

Dexterp37:
Checklist for reviewer:


sql/moz-fx-data-shared-prod/search_derived/remotesettings_suggestions_v1/metadata.yaml
```diff
@@ -0,0 +1,5 @@
+---
+workgroup_access:
```
jklukas, line 2: Let's fill in the friendly_name, description, and owners fields here too. We may have logic that requires them, and they're good minimal documentation to give future users context on what this is.

sql/moz-fx-data-shared-prod/search_derived/remotesettings_suggestions_v1/metadata.yaml
```diff
@@ -0,0 +1,11 @@
+---
+friendly_name: Remote Settings Suggestions
+description: >
+  A table containing the suggestions importend from Remote Settings,
+  from the same list powering Quicksuggest.
```
jklukas, line 5:
```suggestion
  from the same list powering Quicksuggest.
  
  Populated by the `quicksuggest2bq` scheduled task in telemetry-airflow.
```

jklukas:
cc @whd This table is created and populated by a custom task in Airflow, so this metadata is intended primarily to grant appropriate permissions on the table. Does that all sound correct to you?

sql/moz-fx-data-shared-prod/search_derived/remotesettings_suggestions_v1/metadata.yaml
```diff
@@ -0,0 +1,13 @@
...
+
```
Dexterp37, line 6:
```suggestion

```

whd:
I don't think this is specifically necessary since `search_derived` at the dataset-level should be accessible to `mozilla-confidential`. As far as I can tell, this table doesn't exist yet.

Of note though, we have `search_derived` and `search_terms_derived` datasets, perhaps this is meant for the latter instead of the former.

jklukas:
> I don't think this is specifically necessary since `search_derived` at the dataset-level should be accessible to `mozilla-confidential`. As far as I can tell, this table doesn't exist yet.
> 
> Of note though, we have `search_derived` and `search_terms_derived` datasets, perhaps this is meant for the latter instead of the former.

You are correct! Thank you for catching that!

whd:
> You are correct! Thank you for catching that!

Ok, assuming this is changed to `search_terms_derived`, this looks correct  given what I see in https://mozilla-hub.atlassian.net/browse/ROAD-134?focusedCommentId=482791. We'll likely start enforcing increased change procedures on this dataset next week, which would have some additional procedural overhead (e.g. formal managerial approval).

jklukas (approved):

jklukas:
I pushed an update to change the path. The Airflow job successfully created the table, so it exists now and I verified that I'm currently unable to access it.

sql/moz-fx-data-shared-prod/search_terms_derived/remotesettings_suggestions_v1/metadata.yaml
```diff
@@ -0,0 +1,12 @@
+---
+friendly_name: Remote Settings Suggestions
+description: >
+  A table containing the suggestions importend from Remote Settings,
```
whd, line 4:
```suggestion
  A table containing the suggestions imported from Remote Settings,
```

whd (approved):
I can deploy this once it makes its way to `private-generated-sql`. Since this is a non-restricted table in a restricted dataset, it feels like it would be good to have some note of that in the description or comments. Were I to look at this at some point in the future I'd probably `git log` to see the Jira issue where `workgroup:mozilla-confidential` is noted as the appropriate classification. I'm not sure what classification `Remote Settings` data generally falls under but my assumption is that it's `mozilla-confidential` which is why this can be `mozilla-confidential`.

jklukas:
> I'm not sure what classification Remote Settings data generally falls under but my assumption is that it's mozilla-confidential which is why this can be mozilla-confidential.

Yes, that is correct. Given the coming change control process, it may be worth considering moving it into a non-restricted datasets, even though it's conceptually related to search terms data.

I'll file a follow up with the docs fixups.

sql/moz-fx-data-shared-prod/search_terms_derived/remotesettings_suggestions_v1/metadata.yaml
jklukas: Addressed in #2511

whd:
It appears I can't comment on the Jira issue but this ACL change has been applied.

PR #2512, opened 2021-11-18
Add derived table and view for fxa stdout events

relud:
so that mozilla vpn funnel analysis can use events like `fxa_pay_setup - *` that only appear in the stdout tables.


scholtzan (approved):
lgtm

PR #2513, opened 2021-11-18
Add start_date filter and aggregation by submission_date.

emtwo:
I've made some changes to support monitoring a rollout in release where we want to look at data by submission date rather than build ID. I've also added a readme to provide context/explanation on how this works

### DIFF
--- a/bigquery_etl/operational_monitoring/README.md
+++ b/bigquery_etl/operational_monitoring/README.md
@@ -1,0 +1,93 @@
+Operational Monitoring
+===
+
+Operational Monitoring is a general term that refers to monitoring the health of software.
+
+For the purpose of this project, there will be a couple of specific monitoring use cases that are supported:
+1. Monitoring build over build. This is typically for Nightly where one build may contain changes that a previous build doesn't and we want to see if those changes affected certain metrics.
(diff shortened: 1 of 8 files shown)


bigquery_etl/operational_monitoring/README.md
```diff
@@ -0,0 +1,93 @@
...
+}
+```
+
+Below is the description of each field in the project definition and its allowable values:
+* `name`: Name that will be used as the generated Looker dashboard title
+* `slug`: The slug associated with the rollout that is being monitored
+* `channel`: `release`, `beta`, or `nightly`. The channel the rollout is running in
+* `boolean_pref`: A sql snippet that results in a boolean representing whether a user is included in the rollout or not (note: this will soon be deprecated in favour of using the slug directly to check)
+* `xaxis`: `submission_date` or `build_id`. specifies the type of monitoring desired as described above.
+* `start_date`: `yyyy-mm-dd`, specifies the oldest submission date or build id to be processed (where build id is cast to date). If not set, defaults to the previou 30 days.
```
scholtzan, line 83:
```suggestion
* `start_date`: `yyyy-mm-dd`, specifies the oldest submission date or build id to be processed (where build id is cast to date). If not set, defaults to the previous 30 days.
```

scholtzan (approved):

PR #2518, opened 2021-11-19
Simplify experiment_enrollment_cumulative_population_estimate query

scholtzan:
Fixes https://github.com/mozilla/jetstream/issues/924

We've been running into `Resources exceeded during query execution: Not enough resources for query planning - too many subqueries or query is too complex.` errors when querying from `mozdata.telemetry.experiment_enrollment_cumulative_population_estimate`. This change fixes the issue. I already manually deployed the view to do some testing

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_cumulative_population_estimate_v1/view.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_cumulative_population_estimate_v1/view.sql
@@ -1,41 +1,16 @@
 CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.telemetry_derived.experiment_enrollment_cumulative_population_estimate_v1`
 AS
-WITH all_branches AS (
-  -- We need to determine all available branches for this experiment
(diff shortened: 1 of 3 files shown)


sql/moz-fx-data-shared-prod/telemetry_derived/experiment_enrollment_cumulative_population_estimate_v1/view.sql
```diff
@@ -1,41 +1,16 @@
...
-  SELECT
-    experiment,
-    IF(
-      all_branches.branch IS NULL,
-      IF(non_null_count = 0, 'null', all_branches.branch),
-      all_branches.branch
-    ) AS branch
-  FROM
-    all_branches
+    `moz-fx-data-experiments.monitoring.experimenter_experiments_v1`
```
scholtzan, line 9: `moz-fx-data-experiments.monitoring.experimenter_experiments_v1` periodically fetches experiment information, like configured branches, from Experimenter. So we can just use that data instead of doing some expensive computation to determine what branches exist for an experiment.

jklukas (approved):

PR #2522, opened 2021-11-23
ROAD-85 Simple sanitization job for Merino logs

jklukas:
See https://mozilla-hub.atlassian.net/browse/ROAD-85

This uses the adM allowlist of queries for sanitization, so can be expressed entirely in a single query. Future iterations will involve python logic and will likely need to be held elsewhere.

Checklist for reviewer:


sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2/query.sql
```diff
@@ -0,0 +1,81 @@
...
+)
+SELECT
+  * EXCEPT (request_id, sanitized_query, telemetry_query),
+  COALESCE(sanitized_query, telemetry_query) AS sanitized_query,
+FROM
+  impressions
+LEFT JOIN
+  merino_sanitized
+USING
+  (request_id)
```
jklukas, line 81: I think this table needs to be one row per impression, with a single merino request potentially generating several impression pings, and hence matching multiple impression rows on `request_id`. This is implicitly dropping all Merino requests that did not produce matching telemetry.

Does @rebecca-burwei have feedback on this design?

sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2/query.sql
```diff
@@ -0,0 +1,83 @@
...
+    is_clicked,
+    locale,
+    metadata.geo.country,
+    metadata.geo.subdivision1 AS region,
+    normalized_os,
+    normalized_os_version,
+    release_channel,
+    position,
+    reporting_url,
+    scenario,
```
rebecca-burwei, line 31: could we also include `version` or `version` truncated to major version? might help with troubleshooting data in/consistency questions

sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2/query.sql
```diff
@@ -0,0 +1,83 @@
...
+    jsonPayload.fields.rid AS request_id,
+    jsonPayload.fields.query,
+    NULLIF(jsonPayload.fields.country, 'none') AS merino_country,
+    NULLIF(jsonPayload.fields.region, 'none') AS merino_region,
+    NULLIF(jsonPayload.fields.dma, 'none') AS merino_dma,
+    -- -- We are not propagating city-level data to sanitized table.
+    -- NULLIF(jsonPayload.fields.city, 'none') AS merino_city,
+    NULLIF(jsonPayload.fields.form_factor, 'none') AS merino_form_factor,
+    NULLIF(jsonPayload.fields.browser, 'none') AS merino_browser,
+    NULLIF(jsonPayload.fields.os_family, 'none') AS merino_os_family,
```
rebecca-burwei, line 49: does merino have a version that we would want to include here? @ncloudioj (couldn't tag mythmon, not sure why)

sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2/query.sql
```diff
@@ -0,0 +1,83 @@
...
+  FROM
+    merino_logs
+  LEFT JOIN
+    allowed_queries
+  USING
+    (query)
+)
+SELECT
+  * EXCEPT (request_id, sanitized_query, telemetry_query),
+  COALESCE(sanitized_query, telemetry_query) AS sanitized_query,
```
rebecca-burwei, line 77: Do we still want to stop queries from going through telemetry / the impressions ping?

sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2/query.sql
ncloudioj: Merino does have a version, but I don't think we're reporting it in the operational log now. We can add it if that's preferred. 

Hmm, I can't tag mythmon either.

sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2/query.sql
rebecca-burwei: It's ok with me to drop all Merino requests that don't match on an impression.

Yes, this is ok. In the future, we might want to reduce this to one row per `request_id` by aggregating the matching impressions into a struct keyed on `position`. But we wouldn't get much use out of this aggregation until [this other new tab telemetry work](https://docs.google.com/document/d/1ZlKOQYyhlWzlbRKW9HhdmbekKZ9W6jENXmER8IKOTKE/edit#bookmark=id.u3fya4a3qkdl) is done.

sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2/query.sql
rebecca-burwei: ok - including a version would be helpful if you want to experiment with different suggestion logic in merino and see which ones were clicked on more.

sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2/query.sql
ncloudioj: +1. Firefox version is one of the frequently used dimensions, let's include it here.

sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2/query.sql
ncloudioj: Based on the current implementation in Firefox, each merino request would only generate at most 1 impression in Firefox. The Firefox Suggest impression ping is recorded with the following two conditions (see more details in this [doc](https://docs.google.com/document/d/1qLb4hUwR8YQj5QnjJtwxQIoDCPLQ6XuAmJPQ6_WmS4E/edit?usp=sharing)):
* there was a Firefox Suggest served at the end of the search session
* the user picked one of the search results from the search dropdown list

Firefox won't record the impressions for those "flicked-by" suggestions while the user is typing in the URL bar.

So I don't think we will have multiple impression rows corresponding to the same `request_id`.

sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2/query.sql
jklukas: I do want to enforce dropping `search_query` from telemetry in the pipeline at some point, but we haven't negotiated any timeline constraints on it. We could keep it until the opt-in population has been transitioned to merino.

rebecca-burwei: that sounds good to me

sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2/query.sql
jklukas: Added `version` (truncated to major version) below.

rebecca-burwei (approved):
looks good to me

sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2/query.sql
jklukas: @rebecca-burwei and I discussed this a bit more, and we've concluded there's more nuance and definition needed for this. I'm going to write up a doc to discuss the risks and benefits of continuing to allow `search_query` through telemetry in some cases.

jklukas:
I've created https://mozilla-hub.atlassian.net/browse/DSRE-333 to ask @whd to create the destination table.

sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2/query.sql
```diff
@@ -0,0 +1,47 @@
...
+  FROM
+    sanitized_impressions_count
+  LEFT JOIN
+    sanitized_impressions
+  ON
+    TRUE
+  WHERE
+    IF(
+      _n < 1,
+      ERROR("Source table is not yet populated; retry later or investigate upstream issues"),
```
jklukas, line 40: @scholtzan @relud This query introduces a new concept of checking for the source being empty and raising an error as part of the query itself. The intention here is to allow us to define the query within the bigquery-etl dag generation machinery rather than having to introduce a separate DAG in telemetry-airflow just to introduce a step that checks for content.

I'd appreciate your feedback on whether this seems reasonable and is sufficiently documented within this query file.

scholtzan: Yup, this looks like a good idea 👍 
For the error message, would it make sense to mention `suggest-searches-prod-a30f.sanitized.suggest_impression_sanitized_v2` explicitly as the source table, just that it might be good to know for people debugging Airflow errors.

relud: I agree, it's a good approach.

VPN also does something similar to this with [sql/moz-fx-data-shared-prod/mozilla_vpn_derived/site_metrics_empty_check_v1/query.sql](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/mozilla_vpn_derived/site_metrics_empty_check_v1/query.sql) and defining an explicit reference to the empty check:
https://github.com/mozilla/bigquery-etl/blob/101fe5f8e3a0c1ce4fdd3525c37df34ceeb61c51/sql/moz-fx-data-shared-prod/mozilla_vpn_derived/site_metrics_summary_v1/metadata.yaml#L13-L16

Putting it in one query, like you do here, is a nice improvement.

jklukas: Thanks for the review, @scholtzan and @relud!

PR #2528, opened 2021-11-25
Install mkdocs via requirements files

wlach:
This makes doc generation "work out of the box" and simplifies CI
somewhat.


ANich (approved):
LGTM seems like it streamlines local docs development too. Looks like CI is complaining about re-running `pip compile`?

wlach:
> LGTM seems like it streamlines local docs development too. Looks like CI is complaining about re-running `pip compile`?

Yup, looks like it needed some extra comments to match (since I used a different version of python to generate). A bit of a gotcha.

PR #2533, opened 2021-11-29
added changes to minimum client count logic

alekhyamoz:
This PR to correct the logic for filtering on minimum client counts per build-id


emtwo:
I'm glad these changes worked! I think we should also remove the filters from here in the same PR: https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/telemetry_derived/glam_client_probe_counts_extract_v1/query.sql#L21

emtwo (approved):

Issue #2536, opened 2021-11-30
Support partition_expiration_days and require_partition_filter in metadata.yaml


jklukas:
Our `bqetl query schema deploy` command is able to create a table if it doesn't exist. But we use two common options that are not supported yet. We should add support for `partition_expiration_days` and `require_partition_filter` under the `bigquery` section.

This will involve at least:

- Updating the `Metadata` class to include these fields
- Updating the default `metadata.yaml` to include `require_partition_filter: true` as a default
- Update the `deploy` command logic to support these

cc @whd

Issue #2537, opened 2021-11-30
Support updating tables to match metadata.yaml


jklukas:
Related to https://github.com/mozilla/bigquery-etl/issues/2536

We have a `bqetl query schema deploy` command which handles  should have a way to apply changes to metadata.yaml to the table. Some fields, like partitioning, cannot be updated. But other fields like expiration, require_partition_filter, and clustering fields can be changed through a table update operation.

`bqetl query schema deploy` currently calls `client.update_table`, but only passes in the new schema. We could consider having that command also apply options. Or perhaps it should be a new command. I do worry that it could be surprising to have metadata changes propagate via a command called `schema deploy`.

jklukas:
cc @whd @scholtzan

scholtzan:
I think it is fine for `schema deploy` to apply metadata changes. Actually, it already does use some of the metadata, such as clustering and partitioning information, when creating a new table. It might also be more confusing to have separate commands that do similar things.

ANich:
@scholtzan with the changes to `bqetl query schema deploy` and the artifact deployment DAG this is now the case right?

scholtzan:
yes

PR #2541, opened 2021-12-01
Allow to specify partition expiration in metadata

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/2536 and https://github.com/mozilla/bigquery-etl/issues/2537

This adds support for `expiration_ms` to the metadata. We can already specify `require_partition_filter` (see https://github.com/mozilla/bigquery-etl/blob/9bca48821a8a0d40b1700cc14ecd8068d132ed06/sql/moz-fx-data-shared-prod/telemetry_derived/desktop_funnel_installs_v1/metadata.yaml#L15). When running `schema deploy` this partitioning information will also be updated in the source table.

Checklist for reviewer:

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -957,6 +957,7 @@ def _update_query_schema(
                 field=table.time_partitioning.field,
                 partition_type=table.time_partitioning.type_.lower(),
                 required=table.time_partitioning.require_partition_filter,
+                expiration=table.time_partitioning.expiration_ms,
             )
             click.echo(f"Partitioning metadata added to {metadata_file_path}")
 
@@ -1108,6 +1109,7 @@ def deploy(
                         require_partition_filter=(
                             metadata.bigquery.time_partitioning.require_partition_filter
                         ),
+                        expiration_ms=metadata.bigquery.time_partitioning.expiration_ms,
                     )
 
                 if metadata.bigquery and metadata.bigquery.clustering:
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -28,7 +28,15 @@ from ..dependency import get_dependency_graph
 from ..dryrun import SKIP, DryRun
 from ..format_sql.formatter import reformat
 from ..metadata import validate_metadata
-from ..metadata.parse_metadata import METADATA_FILE, DatasetMetadata, Metadata
+from ..metadata.parse_metadata import (
+    METADATA_FILE,
+    DatasetMetadata,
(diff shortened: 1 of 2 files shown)


bigquery_etl/metadata/parse_metadata.py
```diff
@@ -59,7 +59,8 @@ class PartitionMetadata:
 
     field: str
     type: PartitionType
-    require_partition_filter: bool = attr.ib(False)
+    require_partition_filter: bool = attr.ib(True)
```
scholtzan, line 62: The issue mentions a preference to set `require_partition_filter` to true by default. Until now, false was the default.

bigquery_etl/cli/query.py
```diff
@@ -163,6 +171,10 @@ def create(name, sql_dir, project_id, owner, init):
         description="Please provide a description for the query",
         owners=[owner],
         labels={"incremental": True},
+        bigquery=BigQueryMetadata(
+            time_partitioning=PartitionMetadata(field="", type=PartitionType.DAY),
+            clustering=ClusteringMetadata(fields=[]),
+        ),
```
scholtzan, line 177: For explicitly adding time_partitioning and clustering configuration fields to newly created `metadata.yaml` files

bigquery_etl/metadata/parse_metadata.py
scholtzan: I made the changes to explicitly list the configuration fields in the metadata files. I think it makes sense to set it to `true` by default. I guess the default values could be seen as recommendations and we usually recommend requiring partition filters.

jklukas (approved):
Changes look good.

bigquery_etl/metadata/parse_metadata.py
```diff
@@ -59,7 +59,8 @@ class PartitionMetadata:
...
+    expiration_ms: Optional[int] = attr.ib(None)
```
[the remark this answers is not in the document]
jklukas: This is already merged, but additional wrinkle is that CREATE TABLE DDL uses days rather than ms, which is fascinating: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#table_option_list

That takes a float argument, so you can specify partial days, and I assume it ends up turning that into ms behind the scenes.

So, there's an argument for having this be days as a float, and handle the math to turn into millis.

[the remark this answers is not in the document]
scholtzan: Yes, agreed. We should use days here instead.

PR #2544, opened 2021-12-02
correcting minimum client count - desktop and fenix

alekhyamoz:
Correct the logic for filtering minimum client counts per app_build_id

Also removed the test for multiple clients (histogram) as we are generating around 900 client ids with test_minimal script


sql/moz-fx-data-shared-prod/telemetry_derived/clients_scalar_probe_counts_v1/query.sql
```diff
@@ -122,20 +122,51 @@ user_aggregates AS (
...
     user_count,
-    os = 'Windows' and channel = 'release' AS sampled,
+    os = 'Windows'
+    AND channel = 'release' AS sampled,
     udf_boolean_buckets(scalar_aggregates) AS scalar_aggregates
   FROM
-    user_aggregates),
-
+    user_aggregates
+  INNER JOIN
```
emtwo, line 165: I think we don't need the inner join here because we have it below right?

bigquery_etl/glam/generate.py
```diff
@@ -242,13 +250,15 @@ def main():
         table(
             "scalar_bucket_counts_v1",
             **models.scalar_bucket_counts(
-                source_table=f"glam_etl.{args.prefix}__clients_scalar_aggregates_v1"
+                source_table=f"glam_etl.{args.prefix}__clients_scalar_aggregates_v1",
+                **config[args.prefix],
```
emtwo, line 254: May I ask why this extra `**config[args.prefix]` was needed?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_histogram_bucket_counts_v1/query.sql
```diff
@@ -44,13 +44,44 @@ static_combos AS (
     '*' AS os,
     '*' AS app_build_id
 ),
+build_ids AS (
```
emtwo, line 47: Was there a test for this one? Just wondering if there was a test that should have failed but didn't

bigquery_etl/glam/generate.py
```diff
@@ -159,19 +164,22 @@ def main():
             "build_date_udf": "mozfun.glam.build_hour_to_datetime",
             "filter_version": True,
             "num_versions_to_keep": 3,
-            "total_users": 800,
+            "total_users": 10,
```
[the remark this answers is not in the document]
alekhyamoz: Yes, that was my initial thought. But the total_users was being used to filter on the aggregates counts. I retained it to its original state. 
If not needed we can remove it.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_scalar_probe_counts_v1/query.sql
alekhyamoz: Good catch. Will remove it

bigquery_etl/glam/generate.py
alekhyamoz: The scalar_bucket_counts will need the minimum_client_count param from the generate.py

sql/moz-fx-data-shared-prod/telemetry_derived/clients_histogram_bucket_counts_v1/query.sql
alekhyamoz: There is test for `histogram_probe_counts` and not for bucket_counts

bigquery_etl/glam/generate.py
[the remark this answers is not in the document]
alekhyamoz: It is used here : https://github.com/mozilla/bigquery-etl/blob/main/bigquery_etl/glam/templates/extract_probe_counts_v1.sql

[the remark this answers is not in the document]
emtwo: Thank you for the link. I know this PR is getting bigger than intended. I think the correct thing to do would be to remove this line: https://github.com/mozilla/bigquery-etl/blob/main/bigquery_etl/glam/templates/extract_probe_counts_v1.sql#L23 in the same way we removed it from desktop Glam here: https://github.com/mozilla/bigquery-etl/pull/2544/files#diff-ee7135c34b4c4ae3d2419b17328620e740ca07d052e41a0d0e892403db7a8ee1L21

Then `total_users` here in `generate.py` can be removed too. You can do this in a follow-up PR if you'd like.

emtwo (approved):
r+ pending removal of the extra inner join

PR #2551, opened 2021-12-06
DVPN-75 - Create aggregate tables for vpn metrics

relud:
[DVPN-75](https://mozilla-hub.atlassian.net/browse/DVPN-75)

After this is merged I'll post a PR to update looker-spoke-default to use the aggregate tables instead of directly accessing the all_subscriptions table


wwyc (approved):
I don't have the knowledge on how the bigquery etl updates should be like but the aggregated table queries look good to me.

Thanks for the heads up.  I am thinking maybe we can wait to merge the PR to update looker-spoke-default until some time after the official launch of the VPN SaaSboard on Dec9?  This is in case there are any impact to the visualizations and I need to make adjustments.  We can discuss more at the vpn data working group mtg tmr.

PR #2555, opened 2021-12-07
Bug 1744840 - Client probe processes public dataset

wlach:
This will power some upcoming ETL in the Glean Dictionary. See
bug for details.

### DIFF
--- a/dags/bqetl_main_summary.py
+++ b/dags/bqetl_main_summary.py
@@ -45,6 +45,24 @@ with DAG(
     doc_md=docs,
 ) as dag:
 
+    client_probe_processes__v1 = bigquery_etl_query(
+        task_id="client_probe_processes__v1",
+        destination_table="client_probe_processes_v1",
+        dataset_id="telemetry_derived",
+        project_id="moz-fx-data-shared-prod",
+        owner="wlachance@mozilla.com",
+        email=[
+            "dthorn@mozilla.com",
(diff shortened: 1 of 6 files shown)


wlach:
@scholtzan I think this is ok except for https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/12178/workflows/0b05d27b-9725-48c5-a9db-f7bf20c40cf4/jobs/88124 which I don't understand. Is there an extra step I need to do for public data publishing? If so we/I should probably update https://mozilla.github.io/bigquery-etl/cookbooks/creating_a_derived_dataset/

scholtzan (approved):
I'm guessing for generating the DAGs you ran `./bqetl dag generate bqetl_main_summary`.

For the newly scheduled task `public_json` is set to `true`. Exporting data as JSON to our publicly accessible endpoint is scheduled in the `bqetl_public_data_json` DAG. Running `./bqetl dag generate` will add the task there.

Once I start implementing the proposal for generating contents (some time in the next few days, though it will only land after the code freeze), this will happen automatically. In the meantime we can update the docs if you think that would be helpful.

wlach:
> Once I start implementing the proposal for generating contents (some time in the next few days, though it will land after the code freeze), this will happen automatically. In the meantime we can update the docs if you think that would be helpful.

Ah that's great. I think we can skip updating the docs then, thanks!

PR #2556, opened 2021-12-07
Add user-agent to experimenter request

scholtzan:
See https://github.com/mozilla/jetstream/issues/972

It's easier for the Experimenter folks to debug issues if they can see where requests from the API are coming from.

cc @jaredlockhart

### DIFF
--- a/sql/moz-fx-data-experiments/monitoring/experimenter_experiments_v1/query.py
+++ b/sql/moz-fx-data-experiments/monitoring/experimenter_experiments_v1/query.py
@@ -145,7 +145,9 @@ class ExperimentV6:
         converter = cattr.Converter()
         converter.register_structure_hook(
             datetime.datetime,
-            lambda num, _: datetime.datetime.fromisoformat(num.replace("Z", "+00:00")),
+            lambda num, _: pytz.utc.localize(
+                datetime.datetime.fromisoformat(num.replace("Z", "+00:00"))
+            ),
         )
         return converter.structure(d, cls)
 
@@ -174,7 +176,11 @@ class ExperimentV6:
 def fetch(url):
     for _ in range(2):
         try:
-            return requests.get(url, timeout=30).json()
+            return requests.get(
+                url,
+                timeout=30,
+                headers={"user-agent": "https://github.com/mozilla/bigquery-etl"},
+            ).json()
         except Exception as e:
             last_exception = e
             time.sleep(1)
@@ -241,9 +247,7 @@ def main():
         bigquery.SchemaField("channel", "STRING"),
     )
 
-    job_config = bigquery.LoadJobConfig(
-        write_disposition="WRITE_TRUNCATE"
-    )
+    job_config = bigquery.LoadJobConfig(write_disposition="WRITE_TRUNCATE")
     job_config.schema = bq_schema
 
     converter = cattr.Converter()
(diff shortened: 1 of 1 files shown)


relud (approved):

sql/moz-fx-data-experiments/monitoring/experimenter_experiments_v1/query.py
```diff
@@ -145,7 +145,9 @@ def from_dict(cls, d) -> "ExperimentV6":
         converter = cattr.Converter()
         converter.register_structure_hook(
             datetime.datetime,
-            lambda num, _: datetime.datetime.fromisoformat(num.replace("Z", "+00:00")),
+            lambda num, _: pytz.utc.localize(
+                datetime.datetime.fromisoformat(num.replace("Z", "+00:00"))
+            ),
```
relud, line 150: when i test this locally I get `ValueError: Not naive datetime (tzinfo is already set)`

relud: to reproduce:
```python
num = datetime.datetime.utcnow().isoformat() + "Z"
pytz.utc.localize(datetime.datetime.fromisoformat(num.replace("Z", "+00:00")))
```

relud, line 150: this needs to use `normalize`, instead of `localize` to apply the timezone only if it was missing:
```suggestion
            lambda num, _: pytz.utc.normalize(
                datetime.datetime.fromisoformat(num.replace("Z", "+00:00"))
            ),
```

relud (approved):
r+wc

sql/moz-fx-data-experiments/monitoring/experimenter_experiments_v1/query.py
scholtzan: I got `ValueError: Not naive datetime (tzinfo is already set)` before adding `localize()`. With `normalize()` I'm also getting `ValueError` 🤔

relud: ah, you're right, i messed up my local test, normalize does nothing of interest. I do, however, get `ValueError: Not naive datetime (tzinfo is already set)` for `num = datetime.datetime.utcnow().isoformat() + "Z"`

relud: this one works for me regardless of whether a `"Z"` suffix is present, but wouldn't handle a `"+00:00"` suffix:
```suggestion
            lambda num, _: datetime.datetime.fromisoformat(num.replace("Z", "") + "+00:00"),
```

relud: this one works for me with or without `"Z"` or `"+00:00"` suffix:
```suggestion
            lambda num, _: datetime.datetime.fromisoformat(num.replace("Z", "+00:00")).astimezone(pytz.utc),
```

scholtzan: Yup this works. Thanks!

sql/moz-fx-data-experiments/monitoring/experimenter_experiments_v1/query.py
```diff
@@ -145,7 +145,7 @@ def from_dict(cls, d) -> "ExperimentV6":
         converter = cattr.Converter()
         converter.register_structure_hook(
             datetime.datetime,
-            lambda num, _: datetime.datetime.fromisoformat(num.replace("Z", "+00:00")),
+            lambda num, _: datetime.datetime.fromisoformat(num.replace("Z", "+00:00")).astimezone(pytz.utc),
```
relud, line 148: for black:
```suggestion
            lambda num, _: datetime.datetime.fromisoformat(
                num.replace("Z", "+00:00")
            ).astimezone(pytz.utc),
```

PR #2565, opened 2021-12-10
Revert "correcting the issue due to unicode (#2554)"

alekhyamoz:
The glam imports are taking longer than usual and I would like to revert this code.


wlach (approved):
It seems somewhat unlikely to me that this is causing the problem (since it could only reduce the amount of data processed) but I'm fine with reverting the change if it helps investigate the root cause.

PR #2567, opened 2021-12-10
Move generators to sql_generators/

scholtzan:
For https://github.com/mozilla/bigquery-etl/issues/2313

This moves some of the SQL generators from `bigquery_etl/` to `sql_generators/`. I'll open a separate PR for moving `glam` to `sql_generators` (for people working on GLAM to review). The changes here also result in a new CLI command `generate` being added, which will replace the existing custom CLI commands for generating SQL queries. For example, `./bqetl glean_usage generate` is now `./bqetl generate glean_usage`.
Minor code changes were necessary.


bigquery_etl/util/common.py
```diff
@@ -50,9 +51,8 @@ def random_str(length: int = 12) -> str:
 
 def render(sql_filename, format=True, template_folder="glean_usage", **kwargs) -> str:
     """Render a given template query using Jinja."""
-    env = Environment(
-        loader=PackageLoader("bigquery_etl", f"{template_folder}/templates")
-    )
+    file_loader = FileSystemLoader(f"{template_folder}/templates")
+    env = Environment(loader=file_loader)
```
scholtzan, line 55: Moving this into `sql_generators` means it's not part of the `bigquery_etl` package anymore, so use `FileSystemLoader` instead.

sql_generators/glean_usage/__init__.py
```diff
@@ -144,10 +136,5 @@ def generate(project_id, output_dir, parallelism, exclude, only, app_name):
...
 
-    with Pool(parallelism) as pool:
-        pool.starmap(run_generate, generate_per_app_id + generate_per_app)
-
-
-def run_generate(func, params):
-    """Use in `generate()` for generating glean datasets in parallel."""
-    func(params)
+    with ProcessingPool(parallelism) as pool:
+        pool.map(lambda f: f[0](f[1]), generate_per_app_id + generate_per_app)
```
scholtzan, line 140: Use `pathos` for parallel processing. The standard `multiprocessing` module has problems with executing functions in parallel that are not part of a package.

bigquery_etl/cli/generate.py
```diff
@@ -0,0 +1,44 @@
...
+    generator_path = ROOT / SQL_GENERATORS_DIR
+
+    for path in generator_path.iterdir():
+        if path.is_dir():
+            # get Python modules for generators
+            spec = importlib.util.spec_from_file_location(
+                path.name, (path / "__init__.py").absolute()
+            )
+            module = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(module)
```
jklukas, line 25: This is a bit icky, but I'm actually surprised at how little code is needed. I think it's a reasonable thing to do, as the contents of `sql_generators` are essentially plugins, and it does make sense to organize them outside the module itself.

But I think it would be worth having some more commentary right here to discuss that. I also think there should be a `sql_generators/README.md` to explain some context of what's going on in that dir and what the contract is.

sql_generators/glean_usage/baseline_clients_daily.py
```diff
@@ -1,5 +1,5 @@
 """Generating and run baseline_clients_daily queries for Glean apps."""
-from bigquery_etl.glean_usage.common import GleanTable
+from common import GleanTable
```
jklukas, line 2: This file and many following it have permissions changed from `644` to `755` (executable) in this PR. Was that change intentional?

jklukas (changes requested):
The approach here looks great. I have just a few questions, mostly around making sure this is sufficiently documented.

sql_generators/glean_usage/baseline_clients_daily.py
scholtzan: Mh interesting, changed it back. I thought that maybe the way these files are added as part of the CLI might have caused a permission change, but testing it locally that doesn't seem to be the case. So not sure what happened, but should be fixed now

bigquery_etl/cli/generate.py
scholtzan: I added a few more comments and documentation to the README for this. Let me know if I should get more into details.

jklukas:
I still see quite a few yaml and sql files in this PR with permissions changed to 755. This may be and edge case that git doesn't handle well? It may not automatically stage permissions changes without content changes?

scholtzan:
Permissions should be fixed now.

jklukas (approved):

scholtzan:
Let's hold off on merging until next year

scholtzan:
Time to get this merged \o/

PR #2581, opened 2021-12-15
Lowercase and trim in suggest_impressions_sanitized_v2

jklukas:
This better matches the current client behavior for matching.
We're currently getting `<disallowed>` in results from Merino and mixed case in results from telemetry.
I don't think preserving the casing or leading whitespace has analytical value, and it makes the results harder to work with.

Note that we'll need to also make these changes in cloudops-infra to have them take effect.

Checklist for reviewer:

### DIFF
--- a/sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2_external/query.sql
+++ b/sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2_external/query.sql
@@ -17,7 +17,9 @@ WITH impressions AS (
     -- but we reapply truncation to be explicit about granularity.
     TIMESTAMP_TRUNC(submission_timestamp, SECOND) AS submission_timestamp,
     request_id,
-    search_query AS telemetry_query,
+    -- Firefox allows casing and whitespace differences when matching to the
+    -- list of suggestions in RemoteSettings.
+    LTRIM(LOWER(search_query)) AS telemetry_query,
     advertiser,
     block_id,
     context_id,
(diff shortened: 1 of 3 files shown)


sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2_external/query.sql
```diff
@@ -17,7 +17,9 @@ WITH impressions AS (
     -- but we reapply truncation to be explicit about granularity.
     TIMESTAMP_TRUNC(submission_timestamp, SECOND) AS submission_timestamp,
     request_id,
-    search_query AS telemetry_query,
+    -- Firefox allows casing and whitespace differences when matching to the
+    -- list of suggestions in RemoteSettings.
+    LTRIM(LOWER(search_query)) AS telemetry_query,
```
jklukas, line 22: @ncloudioj Can you point me to the specific Firefox code that does the matching to the RemoteSettings list? Does this transformation look similar to what's done in-client?

ncloudioj: Yes, Firefox does [left trimming](https://searchfox.org/mozilla-central/source/browser/components/urlbar/UrlbarProviderQuickSuggest.jsm#136) and [lowercasing](https://searchfox.org/mozilla-central/source/browser/components/urlbar/UrlbarQuickSuggest.jsm#114) before the match.

rebecca-burwei (approved):
looks good to me

PR #2584, opened 2021-12-16
Update google-api-core dependency

Iinh:
Checklist for reviewer:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -212,9 +212,9 @@ gitpython==3.1.24 \
     # via
     #   -r requirements.in
     #   mozilla-schema-generator
-google-api-core[grpc]==1.31.3 \
-    --hash=sha256:4b7ad965865aef22afa4aded3318b8fa09b20bcc7e8dbb639a3753cf60af08ea \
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,8 +1,8 @@
 #
-# This file is autogenerated by pip-compile with python 3.8
+# This file is autogenerated by pip-compile with python 3.9
```
Iinh, line 2: I followed the steps laid out in the docs [here](https://mozilla.github.io/bigquery-etl/cookbooks/common_workflows/#adding-new-python-requirements) to manually update this dependency, but the content of the generated change in this file worries me a little bit... @wlach does this look ok to you?

wlach: Yeah that's normal given differences in the python environment. I usually add these types of changes with `git add -p` to *just* add the changes that upgrade the dependency.

If you want to fix this without filing a whole new PR I would run something like `git revert HEAD^` in the branch, make the changes again (or copy them manually from the PR), and then make a new commit.

Iinh: Thanks Will for the detailed instruction, as always!

wlach (approved):
Thanks for working through this one!

PR #2594, opened 2021-12-21
CI SQL and DAG generation

scholtzan:
Depends on https://github.com/mozilla/bigquery-etl/pull/2597
For For #2313

This updates the Circle CI configuration to run tests on generated SQL and automatically run the DAG generation. Currently, DAGs are only written to `generated-sql`.

Checklist for reviewer:


.circleci/config.yml
```diff
@@ -230,28 +265,14 @@ jobs:
...
-      # Additional query generation commands may be added here in the future.
-      - run:
-          name: Generate events_daily Queries
-          command: PATH="venv/bin:$PATH" script/bqetl events_daily generate
-      - run:
-          name: Generate example glean usage queries
-          command: PATH="venv/bin:$PATH" script/update_example_glean_usage
-      - run:
-          name: Verify that queries were correctly generated and are up-to-date
-          command: git diff --exit-code
```
scholtzan, line 248: This has never actually been executed

.circleci/config.yml
```diff
@@ -53,7 +53,15 @@ jobs:
...
-      - *build
+      - &attach_generated_sql
+        attach_workspace:
+          at: /tmp/workspace
+      - &copy_generated_sql
+        run:
+          name: Move generated-sql into place
+          command: |
+            rm -rf sql/
+            cp -r /tmp/workspace/generated-sql/sql sql
```
scholtzan, line 64: A bunch of CI jobs now depend on the generated-sql. Creating aliases for these steps here to make them reusable

.circleci/config.yml
```diff
@@ -216,6 +249,7 @@ jobs:
             cd telemetry-airflow
             bash bin/test-parse
   verify-dags-up-to-date:
+    # todo: remove this step once we rely on generate-dags completely
```
scholtzan, line 252: This step can be removed once we copy DAGs from `generated-sql` instead of from the `main` branch

bigquery_etl/docs/validate_docs.py
```diff
@@ -38,20 +38,21 @@ def validate(project_dirs):
             for root, dirs, files in os.walk(project_dir):
                 if os.path.basename(root) == EXAMPLE_DIR:
                     for file in files:
-                        dry_run_sql = sub_local_routines(
-                            (Path(root) / file).read_text(),
-                            project_dir,
-                            parsed_routines,
-                        )
+                        if os.path.splitext(file)[1] == ".sql":
```
scholtzan, line 41: Run validation only on SQL files

sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -77,9 +78,11 @@ def generate_per_app(self, project_id, app_info, output_dir=None):
...
                     full_view_id,
                     "view.sql",
-                    VIEW_QUERY_TEMPLATE.format(
-                        full_view_id=full_view_id,
-                        target=underlying_view_id,
+                    reformat(
+                        VIEW_QUERY_TEMPLATE.format(
+                            full_view_id=full_view_id,
+                            target=underlying_view_id,
+                        )
```
scholtzan, line 85: This is causing most of the changes in the diff. But necessary to make `verify-sql-format` happy

scholtzan:
In the diff I can see some changes like: 
```diff
diff -bur --no-dereference bigquery-etl-main/sql/moz-fx-data-shared-prod/firefox_desktop/dataset_metadata.yaml /tmp/workspace/generated-sql/sql/moz-fx-data-shared-prod/firefox_desktop/dataset_metadata.yaml
--- bigquery-etl-main/sql/moz-fx-data-shared-prod/firefox_desktop/dataset_metadata.yaml	2022-01-05 22:43:43.162126032 +0000
+++ /tmp/workspace/generated-sql/sql/moz-fx-data-shared-prod/firefox_desktop/dataset_metadata.yaml	2022-01-05 22:40:12.000000000 +0000
@@ -1,6 +1,6 @@
-friendly_name: firefox-desktop
+friendly_name: Firefox Desktop
 description: |-
-  User-facing views related to document namespace firefox-desktop; see https://github.com/mozilla-services/mozilla-pipeline-schemas/tree/generated-schemas/schemas/firefox-desktop
+  Views on data from firefox_desktop
 dataset_base_acl: view
 user_facing: true
 labels: {}
```

This happens because both for generating stable tables and glean_usage the `dataset_metadata.yaml` file is updated. Previously stable tables have been generated after glean_usage but with the new setup glean usage generation actually happens after generating stable tables.

scholtzan:
New CI workflow:
Screen Shot 2022-01-05 at 15 34 05

.circleci/config.yml
```diff
@@ -174,6 +188,23 @@ jobs:
...
+      - *java_deps
+      - *attach_generated_sql
+      - *copy_generated_sql
+      - run:
+          name: Generate DAGs
+          command: PATH="venv/bin:$PATH" script/generate_airflow_dags
+      - persist_to_workspace:
+          root: /tmp/workspace
+          paths:
+            - generated-sql
```
jklukas, line 207:
```suggestion
      - persist_to_workspace:
          root: /tmp/workspace
          paths:
            - generated-sql
```

We already have the `generate-sql` task persisting to this exact same location. Do we know how that behaves? I suppose if one of the tasks that depends on `generate-sql` were to pick up the content produced by `generate-dags`, it would still be fine because it's just additive.

It would be worth adding a comment here that this is overwriting the content produced by `generated-sql`, though, since that behavior might be surprising.

bigquery_etl/docs/validate_docs.py
jklukas: We could avoid this extra level of nesting by moving the extension check into the `for` above:

```
sql_files = (f for f in files if os.path.splitext(f)[1] == ".sql")
for file in sql_files:
```

setup.py
```diff
@@ -35,6 +35,7 @@ def get_version():
         "google-cloud-bigquery",
         "google-cloud-storage",
         "Jinja2",
+        "pyjnius",
```
jklukas, line 38: We'e so far kept pyjnius out of `setup.py` and `requirements.txt`, instead keeping it in `java-requirements.txt`. Would it be possible to keep that separation?

cc @relud

jklukas:
From the bit of the diff that's not truncated, I see some changes to generated metadata that would be nice to avoid. Some examples:

```
-  User-facing views related to document namespace org-mozilla-firefox-beta; see https://github.com/mozilla-services/mozilla-pipeline-schemas/tree/generated-schemas/schemas/org-mozilla-firefox-beta
+  Views on data from org_mozilla_firefox_beta

-friendly_name: org-mozilla-firefoxreality
+friendly_name: Org Mozilla Firefoxreality
```

If it's fairly low-effort to fix those up, let's include in scope here. But don't block on those regressions.

jklukas (commented):
Some initial comments. It's probably worth having somebody else look through the full diff, which I haven't had time to do yet. I can put that on my schedule tomorrow, though.

scholtzan:
> From the bit of the diff that's not truncated, I see some changes to generated metadata that would be nice to avoid. [...]

Yes this is related to https://github.com/mozilla/bigquery-etl/pull/2594#issuecomment-1006141580

The easy fix here would be to ensure stable tables are generated after glean usage. Or to consolidate the metadata templates for both processes.

.circleci/config.yml
scholtzan: I added a comment for clarification.

setup.py
scholtzan: yup, this is actually not necessary. removed.

bigquery_etl/cli/generate.py
```diff
@@ -45,3 +46,26 @@ def generate_group():
...
+    "--target-project",
+    "--target_project",
+    help="GCP project ID",
+    default="moz-fx-data-shared-prod",
+    callback=is_valid_project,
+)
+@click.pass_context
+def generate_all(ctx, output_dir, target_project):
+    """Run all SQL generators."""
+    for _, cmd in reversed(generate.commands.items()):
```
scholtzan, line 69: reversing the order of commands here, so `glean_usage` is executed before `stable_views` to keep the diff small

jklukas (approved):
The diff looks much cleaner now, and I think I understand all the differences that appear there. Looks good to go 👍

PR #2614, opened 2022-01-04
Added schemas to derived dataset docs

ANich:
Get the schemas for each view from the generated `schema.yaml` file. Currently only works for stable views, will make sure these get generated for non-stable views in a follow up https://github.com/mozilla/bigquery-etl/pull/2614#discussion_r779028390

Checklist for reviewer:

### DIFF
--- a/bigquery_etl/docs/derived_datasets/generate_derived_dataset_docs.py
+++ b/bigquery_etl/docs/derived_datasets/generate_derived_dataset_docs.py
@@ -6,11 +6,13 @@ from jinja2 import Environment, FileSystemLoader
 
 from bigquery_etl.dependency import extract_table_references
 from bigquery_etl.metadata.parse_metadata import DatasetMetadata, Metadata
+from bigquery_etl.schema import Schema
 
 logging.basicConfig(format="%(levelname)s (%(filename)s:%(lineno)d) - %(message)s")
 
 VIEW_FILE = "view.sql"
 METADATA_FILE = "metadata.yaml"
+SCHEMA_FILE = "schema.yaml"
 DATASET_METADATA_FILE = "dataset_metadata.yaml"
 README_FILE = "README.md"
 NON_USER_FACING_DATASET_SUFFIXES = (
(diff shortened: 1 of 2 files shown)


bigquery_etl/docs/derived_datasets/generate_derived_dataset_docs.py
```diff
@@ -68,6 +70,22 @@ def _get_referenced_tables_from_view(table_path):
...
+    project = table_path.parent.parent.name
+    dataset = table_path.parent.name
+    table = table_path.name
+    table_id = f"{project}.{dataset}.{table}"
+
+    try:
+        table = client.get_table(table_id)
+        return table.schema
+    except NotFound as e:
+        logging.warning(f"Unable to find schema. {e}")
```
ANich, line 86: It would be nice to replace this with a call to the existing schema API (`Schema.for_table` for eg.) but dry-running doesn't return schema descriptions. Might be an idea to use the client directly instead of dry-running [here](https://github.com/mozilla/bigquery-etl/blob/main/bigquery_etl/schema/__init__.py#L55-L70) thoughts @scholtzan?

ANich:
(Just realizing nested fields still aren't supported here, but it's a start 😅)

bigquery_etl/docs/derived_datasets/generate_derived_dataset_docs.py
scholtzan: Since we generate the docs as part of the CI we don't have access to a service account that can be used with the client directly.
We could think about extending the dry run function to return schema descriptions: https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/tf/modules/cloudfunctions/src/bigquery_etl_dryrun/index.js

ANich: Gotcha.. Yeah I was trying to avoid relying on referenced tables since any query other than `SELECT * FROM view` gets weird and we'll have to do name matching, etc. And I guess we really wouldn't want to add a service account to CI with limited permissions?

scholtzan: Not with access to `shared-prod`. I guess there is a risk of accidentally exposing data. We do have a service account to the integration testing project, but that one doesn't have access to shared-prod, so not really helpful for this.

ANich: I think we might be able to get creative with a custom IAM role and permissions that only allows listing and getting metadata if it's worth it.. Will put my thoughts down in an issue/proposal or something.

For now though, I think dry running is gonna require a couple changes in different places.. recording all the view partition keys plus the cloud function changes so I'll leave this as a draft for now

scholtzan (approved):
Having this for stable views is already a good improvement

PR #2621, opened 2022-01-05
Fix broken opmon dag.

emtwo:
After this commit https://github.com/mozilla/bigquery-etl/commit/3f0b5373ce23734c3bc9cf7487107685f8c19fa2, the way that `render()` would be called has changed a tad and is currently causing failures in the operational monitoring airflow.

These changes should address the failures


alekhyamoz (approved):
Looks good

PR #2640, opened 2022-01-11
Support Fivetran import task generation

scholtzan:
This adds some convenience around generating Fivetran import tasks.
With this change it is possible to specify Fivetran import tasks in the `metadata.yaml` file as follows:

```yaml
---
friendly_name: "Test table for a non-incremental query"
description: "Test table for a non-incremental query"
owners:
  - test@mozilla.com
scheduling:
  dag_name: "bqetl_core"
  depends_on_fivetran:
    - task_id: fivetran_bugzilla
    - task_id: fivetran_another_import_task
```

This will add the Fivetran import tasks to the generated DAGs. This is quite useful for work on case and revenue, eliminating the need to manually write DAGs for these tasks (e.g. https://github.com/mozilla/telemetry-airflow/blob/main/dags/casa.py) 

Checklist for reviewer:


bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -5,6 +5,15 @@ from operators.task_sensor import ExternalTaskCompletedSensor
 import datetime
 from utils.gcp import bigquery_etl_query, gke_command
 
+{% set ns = namespace(uses_fivetran=False) -%}
+{% for task in tasks -%}
+{% if task.depends_on_fivetran != None and task.depends_on_fivetran|length > 1 and ns.uses_fivetran == False -%}
+{% set ns.uses_fivetran = True -%}
```
ANich, line 11: Just curious, what's `ns` used for here? I haven't used jinja namespaces

ANich (approved):
Makes sense to me that these tasks live together. We have a couple separate DAGs in telemetry-airflow, should we add some todos to migrate these so people stick to one approach?

bigquery_etl/query_scheduling/templates/airflow_dag.j2
scholtzan: I wanted to make sure that the Fivetran related imports are added exactly once if there actually Fivetran dependencies. Simply using `{% set uses_fivetran = false -%}` does not work due to Jinjas [scoping behaviour](https://jinja.palletsprojects.com/en/2.10.x/templates/#assignments). Even when setting `{% set uses_fivetran = true -%}` inside the loop, it will still show up as `false` outside of the loop. Using namespaces propagates changes across scopes.

PR #2643, opened 2022-01-11
[Bug 1749312] metrics_clients_last_seen join on channel

scholtzan:
For reference: https://mozilla.slack.com/archives/C4D5ZA91B/p1641927084020700

It is possible that clients switch between channels (for example from nightly -> beta) and keep their `client_id`. This problem is currently affecting the most recent runs of `metrics_clients_last_seen` for Fenix:
https://workflow.telemetry.mozilla.org/log?dag_id=copy_deduplicate&task_id=metrics_clients_last_seen&execution_date=2022-01-08T01%3A00%3A00%2B00%3A00

We'll need to backfill `metrics_clients_last_seen` once this has been merged.

Checklist for reviewer:


sql_generators/glean_usage/templates/metrics_clients_last_seen.query.sql
```diff
@@ -41,4 +41,5 @@ FULL JOIN
   _current
 USING
   (client_id,
-    sample_id)
+    sample_id,
+    normalized_channel)
```
jklukas, line 45: Are we guaranteed that `normalized_channel` is populated? IIRC, a row with a null in any of these join fields will get dropped.

jklukas (approved):
r+wc

sql_generators/glean_usage/templates/metrics_clients_last_seen.query.sql
scholtzan: I don't think it's guaranteed. Though looking at some selected tables (Fenix) it does not seem to be `null`. But I guess we should handle the `null` cases.

sql_generators/glean_usage/templates/metrics_clients_last_seen.query.sql
```diff
@@ -39,6 +39,10 @@ FROM
...
-USING
-  (client_id,
-    sample_id)
+ON
+  _previous.client_id = _current.client_id AND
+  _previous.sample_id = _current.sample_id AND
+  (
+    _previous.normalized_channel = _current.normalized_channel OR 
+    (_previous.normalized_channel IS NULL AND _current.normalized_channel IS NULL) 
+  )
```
scholtzan, line 48: @jklukas does this seem reasonable?

jklukas: This looks reasonable, though I'm not 100% sure without running it and validating the output. It's worth double-checking that we get the right result here before merging.

PR #2649, opened 2022-01-12
reduced the minimum count values for firefox desktop

alekhyamoz:
Adjusting the minimum client count and total_users values, so as to not filter all the data.


scholtzan (approved):
I'm not totally familiar with GLAM, but I guess you have checked that the parameters will return the expected results?

PR #2656, opened 2022-01-13
[Bug 1750005] submission_date parameter for stable_table_column_counts

scholtzan:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1750005

Checklist for reviewer:


dags/bqetl_monitoring.py
```diff
@@ -110,8 +110,9 @@
         project_id="moz-fx-data-shared-prod",
         owner="ascholtz@mozilla.com",
         email=["ascholtz@mozilla.com"],
-        date_partition_parameter="submission_date",
+        date_partition_parameter=None,
         depends_on_past=False,
+        parameters=["submission_date:DATE:{{ds}}"],
```
scholtzan, line 115: We cannot use `date_partition_parameter` here since there is an `INSERT INTO` statement as part of the script instead for writing selected data into a partition.

sql/moz-fx-data-shared-prod/monitoring_derived/stable_table_column_counts_v1/script.sql
```diff
@@ -41,6 +41,12 @@ LOOP
     ''';
 END LOOP;
 
+-- Delete existing records for day.
+-- Prevents having duplicates when re-running query.
+DELETE FROM
+  monitoring_derived.stable_table_column_counts_v1
+WHERE
+  submission_date = DATE(@submission_date)
```
scholtzan, line 49: One thing we should keep in mind, when backfilling we'll have the column numbers of the day when the backfill is executed for each `submission_date`. I think this is fine since we are generally interested in knowing whether we have reached the max columns. Keeping a record for historical dates is just a nice to have

ANich (approved):

PR #2663, opened 2022-01-14
Skip udfs and views to stable tables; dynamically extract view refs

ANich:
Some schemas are not being written because of dry-running with no partition key even though the partition key is there. I think this is because `bqetl dependency record` runs after schema generators. We can dynamically get the references from the view file instead of waiting for metadata to be written to fix that here.

Also skip views to stable tables since there's another generator for that and skip udfs.

Missing schemas on https://mozilla.github.io/bigquery-etl/mozdata/introduction/ should fall into these categories now: 
- Views to tables that don't have defined partition keys
- Views that can't easily be dry-run (eg. views to multiple tables each with partition keys)
- Nested views (view -> view -> table)

Will be tackling these in follow-ups (as much as possible)

Checklist for reviewer:

### DIFF
--- a/sql_generators/derived_view_schemas/__init__.py
+++ b/sql_generators/derived_view_schemas/__init__.py
@@ -6,17 +6,19 @@ from pathlib import Path
 import click
 from pathos.multiprocessing import ProcessingPool
 
-NON_USER_FACING_DATASET_SUFFIXES = (
+NON_USER_FACING_DATASET_SUBSTRINGS = (
     "_derived",
     "_external",
     "_bi",
     "_restricted",
+    "udf",
 )
 
 
 def _generate_view_schema(sql_dir, view_directory):
     import logging
 
+    from bigquery_etl.dependency import extract_table_references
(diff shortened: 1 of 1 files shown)


sql_generators/derived_view_schemas/__init__.py
```diff
@@ -11,12 +11,14 @@
     "_external",
     "_bi",
     "_restricted",
+    "udf",
```
scholtzan, line 14: does `udf_js` also need to be added here?

scholtzan (approved):

sql_generators/derived_view_schemas/__init__.py
ANich: The way the check is written: https://github.com/mozilla/bigquery-etl/blob/9677405044571c40b3490861eacbcc9d86d6d917/sql_generators/derived_view_schemas/__init__.py#L166-L169 should cover the `udf` substring in `udf_js`, `udf_js_ilb` and `udf_legacy` as well. (I guess that makes "suffix" a little misleading here haha, updated)

PR #2667, opened 2022-01-17
Update bqetl query command to handle generated SQL

scholtzan:
For https://github.com/mozilla/bigquery-etl/issues/2313

Checklist for reviewer:


bigquery_etl/cli/utils.py
```diff
@@ -104,6 +104,14 @@ def paths_matching_name_pattern(pattern, sql_path, project_id, files=("*.sql")):
     callback=is_valid_dir,
 )
 
+tmp_dir_option = click.option(
+    "--tmp_dir",
+    "--tmp-dir",
+    help="Path to temporary directory generated SQL is written to.",
+    type=click.Path(file_okay=False),
+    default=Path("/tmp/sql"),
```
jklukas, line 112: Can we consider generating the tmp directory name rather than a static path?

scholtzan: A temporary directory is now generated unless explicitly specified

bigquery_etl/cli/utils.py
```diff
@@ -104,6 +105,25 @@ def paths_matching_name_pattern(pattern, sql_path, project_id, files=("*.sql")):
     callback=is_valid_dir,
 )
 
+
+def generate_tmp_dir(ctx, param, value):
+    """Generate temporary directory if no default directory is provided."""
+    if value is None:
+        tmp_dir = Path(tempfile.mkdtemp()) / "sql"
```
jklukas, line 112: Do we ever clean up this directory? If we're generating a random tmp dir, I'd expect this code to also handle cleaning up the directory when it's done.

jklukas: Can we use the pattern described in https://click.palletsprojects.com/en/8.0.x/advanced/#managing-resources ?

scholtzan: Interesting, yes! It seems like the context object cannot be renamed, e.g. I was hoping `ctx.tmp_dir` would be allowed, but only `ctx.obj` works.

scholtzan: Ah actually `ctx.obj["TMP_DIR"]` might be a good alternative here

jklukas (approved):
Changes look great!

PR #2672, opened 2022-01-19
Update provider names for VPN subs

relud:
as requested by @birdsarah, because at some point this year everything will be `FxA`, and `IAP` is redundant.


sql/moz-fx-data-shared-prod/mozilla_vpn_derived/funnel_ga_to_subscriptions_v1/query.sql
```diff
@@ -83,7 +83,7 @@ subscriptions AS (
   WHERE
     DATE(subscription_start_date) >= '2020-07-01'
     AND product_name = "Mozilla VPN"
-    AND provider LIKE "FxA %"
+    AND provider IN ("Stripe", "Paypal")
```
sean-rose, line 86: This accurately preserves the existing logic, but obscures the intent (why Stripe and Paypal?), which seems like it would make it less likely this logic will be maintained correctly going forward.  An in-code comment mentioning this is filtering to FxA-based providers could help with that.

sean-rose: Also, per @birdsarah's comments [in Slack](https://mozilla.slack.com/archives/C018SNLQP6C/p1642614349014400?thread_ts=1642614305.013900&cid=C018SNLQP6C):
> Can the prefix of FxA go away?
> Can the suffix of IAP go away?
> Both would also apply to Google Play making it FxA Google Play IAP
> And Apple subs will be moved onto FxA this year.

- Should Google Play be included in this filter as well, since FxA apparently applies to Google Play?
- Did FxA always apply to Google Play, or is that a change that happened at some point?
- Do we need to be able to distinguish between FxA vs non-FxA subscriptions in an easier/clearer way?  Otherwise it seems like making that distinction will require knowing which providers are FxA-based (and even more specifically during which time periods).

sean-rose (approved):
This will likely break some existing reports in Redash and Looker ([example](https://github.com/mozilla/looker-spoke-default/blob/04a9c095b2122432cff3f0606ac9bed80faebee9/mozilla_vpn/views/subscriptions.view.lkml#L115)), so it'll be important to notify people.

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/funnel_ga_to_subscriptions_v1/query.sql
relud: > * Should Google Play be included in this filter as well, since FxA apparently applies to Google Play?

no, this funnel only counts subscriptions that signed up through the web flow, and so does not include any type of IAP subscriptions.

> * Did FxA always apply to Google Play, or is that a change that happened at some point?

it always applied

> * Do we need to be able to distinguish between FxA vs non-FxA subscriptions in an easier/clearer way? Otherwise it seems like making that distinction will require knowing which providers are FxA-based (and even more specifically during which time periods).

no, because everything will be FxA.

relud: > An in-code comment mentioning this is filtering to FxA-based providers could help with that.

how's this:

```suggestion
    -- only count subscriptions that can have UTMs matching GA, which currently maps to non-IAP providers
    AND provider NOT IN ("Apple Store", "Google Play")
```

sean-rose: Looks good, thanks!

relud:
waiting until monday to merge this, so that i can send out a notification and be around to fix anything that breaks

PR #2681, opened 2022-01-24
Add Suggestion fields to aggregated_search_terms_daily

rebecca-burwei:
Search-term-level analysis is not useful to the business. They want to see things like "how many clicks do we get on ebay watches?" or "how many clicks do we get in shopping (iab_category)?". 

I'm proposing to add these additional fields to aggregated_search_terms_daily. It's possible this PR is rejoining on fields that were previously dropped further up the sanitization pipeline (not sure - got a tip from Alessio that this might be the case).

Checklist for reviewer:


whd (commented):
@rebecca-burwei I just merged https://github.com/mozilla/bigquery-etl/pull/3017 which should hopefully fix the test if you rebase this, but this PR also looks like the SQL needs to be formatted.

jklukas (changes requested):
I don't think we're currently set up to accept the new fields in the table, so I expect this job will error out if merged as is.

We can add the following to the `scheduling` section in `metadata.yaml`:

```
```

Or we could do this join in the user-facing view rather than in this query directly.

Otherwise, we'd need to define a `schema.yaml` and deploy that.

jklukas:
I think I'm going to go ahead and push a change here to add ALLOW_FIELD_ADDITION.

jklukas (approved):
I pushed the scheduling config update. This should be good to go once CI passes.

PR #2682, opened 2022-01-24
Android Focus search ETL - DO-824, Bug 1749833

ANich:
RE: DO-824, Bug 1749833.

Update `mobile_search_clients_daily_v1` to include Glean data. 

Most of the diff here is adding test schemas. The most relevant changes are in `sql/` and `sql_generators/` 

Here's the test query I ran that returned no results as expected:

```
SELECT
  submission_date,
  engine,
  source,
  app_name,
  normalized_app_name,
  COUNT(*) AS client_count,
  SUM(search_count) AS search_count,
  SUM(organic) AS organic,
  SUM(tagged_sap) AS tagged_sap,
  SUM(tagged_follow_on) AS tagged_follow_on,
  SUM(ad_click) AS ad_click,
  SUM(ad_click_organic) AS ad_click_organic,
  SUM(search_with_ads) AS search_with_ads,
  SUM(unknown) AS unknown,
FROM
  search_derived.mobile_search_clients_daily_v1
WHERE
  submission_date = DATE('2022-07-01')
  AND engine IS NOT NULL
GROUP BY
  submission_date,
  engine,
  source,
  app_name,
  normalized_app_name
EXCEPT DISTINCT
SELECT
  submission_date,
  engine,
  source,
  app_name,
  normalized_app_name,
  COUNT(*) AS client_count,
  SUM(search_count) AS search_count,
  SUM(organic) AS organic,
  SUM(tagged_sap) AS tagged_sap,
  SUM(tagged_follow_on) AS tagged_follow_on,
  SUM(ad_click) AS ad_click,
  SUM(ad_click_organic) AS ad_click_organic,
  SUM(search_with_ads) AS search_with_ads,
  SUM(unknown) AS unknown,
FROM
    moz-fx-data-shared-prod._029ee9dbef54afd5652c9d7993ab170a90c5caea.anonf60774ed_ff6c_4865_8c77_c583894a4469
WHERE
  submission_date = DATE('2022-07-01')
  AND engine IS NOT NULL
GROUP BY
  submission_date,
  engine,
  source,
  app_name,
  normalized_app_name
```

As a follow-up it would be great if the SQL pytest plugin supported a single schema for each set of table tests (inputs schemas shouldn't change between tests).


sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/query.sql
```diff
@@ -344,6 +344,138 @@ metrics_org_mozilla_ios_fennec AS (
   WHERE
     mozfun.norm.truncate_version(client_info.app_display_version, 'major') >= 28
 ),
+-- metrics for Focus Android release
+metrics_org_mozilla_focus AS (
+  SELECT
+    DATE(submission_timestamp) AS submission_date,
+    client_info.client_id,
+    normalized_country_code AS country,
+    'Focus' AS app_name,
```
scholtzan, line 353: In `nondesktop_clients_last_seen` the app names for Glean Focus/Klar products contain a `Baseline` to distinguish them from core pings. So `Android Focus Baseline`, `iOS Focus Baseline` etc. I think we might want to be consistent here
https://github.com/mozilla/bigquery-etl/pull/3040

xluo-ds:
@ANich  hey just for my own understanding, so this PR doesn't change anything that's currently in `mobile_search_clients_daily` table but only add rows under app_name `Android Focus Baseline` instead, right? Another question, what's the impact onto downstream tables `mobile_search_aggregates`, and `mobile_search_clients_last_seen`?

ANich:
> @ANich hey just for my own understanding, so this PR doesn't change anything that's currently in `mobile_search_clients_daily` table but only add rows under app_name `Android Focus Baseline` instead, right? Another question, what's the impact onto downstream tables `mobile_search_aggregates`, and `mobile_search_clients_last_seen`?

Yes exactly - since there's no schema change, those downstreams can be back-filled as well to include this data. It's probably only necessary to back-fill to the beginning of the Glean data, I can check what the earliest date for that is.

xluo-ds:
> > @ANich hey just for my own understanding, so this PR doesn't change anything that's currently in `mobile_search_clients_daily` table but only add rows under app_name `Android Focus Baseline` instead, right? Another question, what's the impact onto downstream tables `mobile_search_aggregates`, and `mobile_search_clients_last_seen`?
> 
> Yes exactly - since there's no schema change, those downstreams can be back-filled as well to include this data. It's probably only necessary to back-fill to the beginning of the Glean data, I can check what the earliest date for that is.

gotcha. Thanks!

sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/query.sql
Dexterp37: @scholtzan @ANich although in #3040 it was called `Baseline` because it was coming from the `baseline` ping. Here we're looking at the `Metrics` ping. Maybe we should have gone by "Android Focus Glean" :)

sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/query.sql
```diff
@@ -344,6 +344,138 @@ metrics_org_mozilla_ios_fennec AS (
...
+    client_info.locale,
+  FROM
+    org_mozilla_focus_nightly.metrics AS org_mozilla_focus_nightly_metrics
+  WHERE
+    mozfun.norm.truncate_version(
+      client_info.android_sdk_version,
+      'minor'
+    ) >= 98  -- TODO: Update this with the actual version
+),
+-- metrics for Focus Android release
```
Dexterp37: Should this be `Klar Android release`?

sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/query.sql
```diff
@@ -412,6 +544,27 @@ ios_metrics AS (
...
+  UNION ALL
+  SELECT
+    *
+  FROM
+    metrics_org_mozilla_focus_nightly
+  UNION ALL
+  SELECT
+    *
+  FROM
+    metrics_org_mozilla_klar
```
Dexterp37, line 542: Since we're here, should we also add Focus iOS and Klar iOS?

sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/query.sql
```diff
@@ -344,6 +344,114 @@ metrics_org_mozilla_ios_fennec AS (
...
+-- metrics for Focus Android Baseline nightly
+metrics_org_mozilla_focus_nightly AS (
+  SELECT
+    DATE(submission_timestamp) AS submission_date,
+    client_info.client_id,
+    normalized_country_code AS country,
+    'Focus Android Baseline' AS app_name,
+    'Focus' AS normalized_app_name,
+    client_info.app_display_version AS app_version,
+    'Release' AS channel,
```
Dexterp37, line 410:
```suggestion
    'Nightly' AS channel,
```

?

sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/query.sql
```diff
@@ -344,6 +344,114 @@ metrics_org_mozilla_ios_fennec AS (
...
+-- metrics for Focus Android Baseline beta
+metrics_org_mozilla_focus_beta AS (
+  SELECT
+    DATE(submission_timestamp) AS submission_date,
+    client_info.client_id,
+    normalized_country_code AS country,
+    'Focus Android Baseline' AS app_name,
+    'Focus' AS normalized_app_name,
+    client_info.app_display_version AS app_version,
+    'Release' AS channel,
```
Dexterp37, line 383:
```suggestion
    'Beta' AS channel,
```

Should this be Beta?

ANich: Good catch, this is a templated query I'll update the template

sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/query.sql
Dexterp37: @ANich are you planning on adding the iOS Focus/Klar as part of this PR, too? If not, would you mind to file a DO ticket about this?

scholtzan (approved):
lgtm

gkabbz:
Catching up on this… I have a concern with this proposed solution. In my view the `mobile_search_clients_daily_v1` table is a record of truth type table. It is the table being used in looker to power our search reporting in the [mobile_search_clients_engines_sources_daily](https://mozilla.cloud.looker.com/projects/looker-hub/files/search/views/mobile_search_clients_engines_sources_daily.view.lkml) view (see [bq-etl](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/search/mobile_search_clients_engines_sources_daily/view.sql)). The current proposal will add the glean data into the table and use app_name to differentiate between the glean and legacy data. However, there’s no distinction in the normalized_app_name = Focus which is shared between the data coming from both glean and legacy. 

My concern is that we are creating an opportunity for incorrect interpretation of data downstream. If the proposed solution is implemented, I can see a situation where an end user of an explore might simply sum a search metric not realizing that Focus Android is being counted twice (approximately). It is not uncommon for people to use the normalize_app_name field to report on different products (we do on our team). The same issue would happen here as well where the Focus Android would be double counting. In addition a few other tables also depend on this table namely: `mobile_search_clients_last_seen`, `mobile_search_aggregates` and `mobile_search_aggregates_for_search_report`. With the exception of the latter, none of them have any filtering on app_name and os to perform their aggregations. So we will additionally cascade this issue downstream as well. If we had an easy way to communicate this that would be memorable and 2 months from now people would remember the caveat when using the data then that could be a way to limit the impact but I don’t think we do. 

But that doesn’t change that we need to support the experiment in the short term. My understanding is that we can tell jetstream which table to read from. My question is:

1. Can we ask jetstream to read the glean data where it is right now? 
Or
2. Can we build a similar derived table for the glean data that’s ready and usable but is not ready yet to be included in our record of truth tables? This way people can use it as in this instance but we do not introduce opportunities for error. When we are ready to start using that data in our record of truth we’d do the migration and backfills then deprecate the data for that product in the temp table.

bani:
> But that doesn’t change that we need to support the experiment in the short term. My understanding is that we can tell jetstream which table to read from. My question is:
> 1.  Can we ask jetstream to read the glean data where it is right now?

It'd be possible to implement the metric with raw data, but since this will be a guardrail metric for all Focus experiments and not just a single experiment config it'd be preferable to implement it consistently with how it's done in the other mobile browsers. So if we can have the data on `mobile_search_clients_daily_v1` in the next ~3 weeks that's what I'd rather use in Jetstream.

Dexterp37:
> Catching up on this… I have a concern with this proposed solution. In my view the `mobile_search_clients_daily_v1` table is a record of truth type table. It is the table being used in looker to power our search reporting in the [mobile_search_clients_engines_sources_daily](https://mozilla.cloud.looker.com/projects/looker-hub/files/search/views/mobile_search_clients_engines_sources_daily.view.lkml) view (see [bq-etl](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/search/mobile_search_clients_engines_sources_daily/view.sql)). The current proposal will add the glean data into the table and use app_name to differentiate between the glean and legacy data. However, there’s no distinction in the normalized_app_name = Focus which is shared between the data coming from both glean and legacy.

Given the [pipeline spec](https://docs.telemetry.mozilla.org/cookbooks/new_ping.html?highlight=normalized_app_name#ingestion-metadata), using `normalized_app_name` seems like the appropriate path: regardless of the data collection system, that's indeed "Focus".

@gkabbz , I understand your concerns. Is there any reason why we shouldn't instead fix the search reporting dashboards/tables/tools? (e.g. do the appropriate filtering where the search tables are used)?

I'm also fine with changing the `normalized_app_name` (although that's more of a @scholtzan question, whether or not this is acceptable).

> My concern is that we are creating an opportunity for incorrect interpretation of data downstream. If the proposed solution is implemented, I can see a situation where an end user of an explore might simply sum a search metric not realizing that Focus Android is being counted twice (approximately). It is not uncommon for people to use the normalize_app_name field to report on different products (we do on our team). The same issue would happen here as well where the Focus Android would be double counting. In addition a few other tables also depend on this table namely: `mobile_search_clients_last_seen`, `mobile_search_aggregates` and `mobile_search_aggregates_for_search_report`. With the exception of the latter, none of them have any filtering on app_name and os to perform their aggregations. So we will additionally cascade this issue downstream as well. If we had an easy way to communicate this that would be memorable and 2 months from now people would remember the caveat when using the data then that could be a way to limit the impact but I don’t think we do.

I agree with the concern you outlined. From a quick look, though, seems that [mobile_search_clients_last_seen](https://github.com/mozilla/bigquery-etl/blob/9bca48821a8a0d40b1700cc14ecd8068d132ed06/sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_last_seen_v1/query.sql#L35-L36) does group by `app_name` and `normalized_app_name`. And so does [mobile_search_aggregates](https://github.com/mozilla/bigquery-etl/blob/9bca48821a8a0d40b1700cc14ecd8068d132ed06/sql/moz-fx-data-shared-prod/search_derived/mobile_search_aggregates_v1/query.sql#L33-L34) and [`mobile_search_aggregates_for_searchreport`](https://github.com/mozilla/bigquery-etl/blob/9bca48821a8a0d40b1700cc14ecd8068d132ed06/sql/moz-fx-data-shared-prod/search_derived/mobile_search_aggregates_for_searchreport_v1/query.sql#L59). Given the lack of sleep, @ANich and/or @scholtzan should probably double check, given the importance of these tables!

As for how to communicate this, looks like both [dataset documentation](https://docs.telemetry.mozilla.org/datasets/search.html) and the data-dev mailing list might be good places to communicate this. The plan is to remove the legacy data from the derived datasets hopefully somewhen in 2023, so the risk of problems should be solved by then.

@gkabbz would this be an acceptable path forward? Happy to meet and talk about this over zoom!

> But that doesn’t change that we need to support the experiment in the short term. My understanding is that we can tell jetstream which table to read from. My question is:
> 
>     1. Can we ask jetstream to read the glean data where it is right now?

The derived datasets are are interface to our data collection and Glean data must be present in these datasets, along with the legacy data, as we move along with decomissioning legacy telemetry.

>        Or
> 
>     2. Can we build a similar derived table for the glean data that’s ready and usable but is not ready yet to be included in our record of truth tables? This way people can use it as in this instance but we do not introduce opportunities for error. When we are ready to start using that data in our record of truth we’d do the migration and backfills then deprecate the data for that product in the temp table.

This would be an exception only for Focus. While I agree that might be less confusing in the short term, this might become more confusing and problematic in the long run.

sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/query.sql
```diff
@@ -344,6 +344,114 @@ metrics_org_mozilla_ios_fennec AS (
   WHERE
     mozfun.norm.truncate_version(client_info.app_display_version, 'major') >= 28
 ),
+-- metrics for Focus Android Baseline release
+metrics_org_mozilla_focus AS (
+  SELECT
+    DATE(submission_timestamp) AS submission_date,
+    client_info.client_id,
+    normalized_country_code AS country,
+    'Focus Android Baseline' AS app_name,
```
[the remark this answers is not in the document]
alekhyamoz: as per the sql-generators[readme](https://github.com/mozilla/bigquery-etl/tree/main/sql_generators#readme), we do not require generating to merge the query.sql to main as it will be generated via CI

alekhyamoz (commented):
as per the sql-generators [readme](https://github.com/mozilla/bigquery-etl/tree/main/sql_generators#readme), we do not require generating to merge the query.sql to main as it will be generated via CI

sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/query.sql
scholtzan: We could do the renaming of "Baseline" to "Glean" in the views. Or we backfill the tables with the new name

sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/query.sql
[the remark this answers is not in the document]
ANich: @alekhyamoz good catch - I think this can be done in a follow-up PR along with any other generated queries

sql_generators/search/__init__.py
```diff
@@ -18,15 +18,28 @@
...
+]
+
+FIREFOX_IOS_TUPLES = [
+    ("org_mozilla_ios_firefox",     "Fennec", "release"),  # noqa E241 E501
+    ("org_mozilla_ios_firefoxbeta", "Fennec", "beta"),  # noqa E241 E501
+    ("org_mozilla_ios_fennec",      "Fennec", "nightly"),  # noqa E241 E501
+]
+
+FOCUS_ANDROID_TUPLES = [
+    ("org_mozilla_focus",           "Focus Android Baseline",    "release"),  # noqa E241 E501
```
scholtzan, line 36: Let's stick with calling them "Glean" and change the "Baseline" ones we have in other tables in a follow-up PR (we'll need to make sure not to break existing dashboards etc here though)
```suggestion
    ("org_mozilla_focus",           "Focus Android Glean",    "release"),  # noqa E241 E501
```

scholtzan (commented):
We should add a view for `mobile_search_clients_daily`

ANich:
> We should add a view for `mobile_search_clients_daily`

@scholtzan There's currently a view, it just doesn't follow the naming convention - https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/search/mobile_search_clients_engines_sources_daily/view.sql

sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/query.sql
ANich: Updated

scholtzan:
> @scholtzan There's currently a view, it just doesn't follow the naming convention - https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/search/mobile_search_clients_engines_sources_daily/view.sql

Ok, awesome. Didn't realize that.

PR #2686, opened 2022-01-25
Bug 1747068 - Increase the maximum parallelism of shredder

relud:
by adding a flag to not use DML

Checklist for reviewer:


bigquery_etl/shredder/delete.py
```diff
@@ -207,18 +218,16 @@ def create_job(client):
         )
         query = reformat(
             f"""
-            {"SELECT * FROM" if dry_run and read_only else "DELETE"}
+            {"DELETE" if use_dml else "SELECT * FROM"}
               `{sql_table_id(target)}`
             WHERE
-              ({field_condition})
-              AND {partition_condition}
+              {"" if use_dml else "NOT "}({field_condition})
```
sean-rose, line 224: Is it possible the field condition could evaluate to null?  Because simply applying `NOT` won't select the exact inverse if nulls are involved.

If nulls are involved you could do something like
```suggestion
              COALESCE(({field_condition}), FALSE) = {"TRUE" if use_dml else "FALSE"}
```

bigquery_etl/shredder/delete.py
```diff
@@ -249,8 +259,8 @@ def get_partition(table, partition_expr, end_date, id_=None) -> Optional[Partiti
         return Partition("TRUE")
     if id_ == NULL_PARTITION_ID:
         if table.time_partitioning:
-            return Partition(f"{table.time_partitioning.field} IS NULL", id_)
-        return Partition(f"{partition_expr} IS NULL", id_)
+            return Partition(f"{table.time_partitioning.field} IS NULL", id_, True)
+        return Partition(f"{partition_expr} IS NULL", id_, True)
```
sean-rose, line 263: IMO passing literal values like `True` as positional arguments makes code less readable as you have to track down the function definition to see what they mean or rely on code introspection (which isn't always available), so I'd suggest specifying these as keyword arguments.
```suggestion
            return Partition(f"{table.time_partitioning.field} IS NULL", id_, is_special=True)
        return Partition(f"{partition_expr} IS NULL", id_, is_special=True)
```

bigquery_etl/shredder/delete.py
```diff
@@ -262,6 +272,7 @@ def get_partition(table, partition_expr, end_date, id_=None) -> Optional[Partiti
                 f"{partition_expr} < {table.range_partitioning.range_.start} "
                 f"OR {partition_expr} >= {table.range_partitioning.range_.end}",
                 id_,
+                True,
```
sean-rose, line 275:
```suggestion
                is_special=True,
```

bigquery_etl/shredder/delete.py
relud: nulls are not a concern here.

sean-rose: OK, so it looks like the assumption is the deletion requests (or similar events) used as sources never have null values in the relevant ID columns.

Personally I'd still be inclined to code defensively against possible nulls, as the consequence of a possible null value there would be to accidentally delete all data in the partition (`IN` returns null if there's no positive match in the search set and the search set contains any nulls).

relud: Okay, then how about:
```suggestion
             ({field_condition}) {"" if use_dml else " IS NOT TRUE"}
```

bigquery_etl/shredder/delete.py
```diff
@@ -221,7 +221,7 @@ def create_job(client):
             {"DELETE" if use_dml else "SELECT * FROM"}
               `{sql_table_id(target)}`
             WHERE
-              {"" if use_dml else "NOT "}({field_condition})
+             ({field_condition}){"" if use_dml else " IS NOT TRUE"}
```
sean-rose, line 224: Indentation slightly off.
```suggestion
              ({field_condition}){"" if use_dml else " IS NOT TRUE"}
```

sean-rose (approved):

PR #2690, opened 2022-01-28
Ensure dags/ exists on generated-sql and speed up CI

scholtzan:
Okay, I'm trying to solve the mystery around `dags/` not being in `generated-sql` once and for all. I wrote a CI test step to check the files exist in: https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/12979/workflows/3dad7a02-d518-49b7-b5e1-39cd6482ab02/jobs/98704
The problem was that `generate-diff` overwrote the `generated-sql` contents.

Checklist for reviewer:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -374,23 +374,20 @@ jobs:
     executor: ubuntu-machine-executor
     steps:
       - checkout
-      - *attach_generated_sql
-      - *copy_generated_sql
       - when:
           condition: &deploy-condition
             or:
               - equal: [main, << pipeline.git.branch >>]
               - << pipeline.git.tag >>
           steps:
+            - *attach_generated_sql
+            - *copy_generated_sql
             - docker/check:
                 docker-password: DOCKER_PASS
                 docker-username: DOCKER_USER
-      - docker/build: &public-image
-          image: ${CIRCLE_PROJECT_USERNAME+$CIRCLE_PROJECT_USERNAME/}${CIRCLE_PROJECT_REPONAME:-bigquery-etl}
-          tag: ${CIRCLE_TAG:-latest}
-      - when:
-          condition: *deploy-condition
-          steps:
+            - docker/build: &public-image
+                image: ${CIRCLE_PROJECT_USERNAME+$CIRCLE_PROJECT_USERNAME/}${CIRCLE_PROJECT_REPONAME:-bigquery-etl}
+                tag: ${CIRCLE_TAG:-latest}
             - docker/push: *public-image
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -374,23 +374,20 @@ jobs:
...
               - << pipeline.git.tag >>
           steps:
+            - *attach_generated_sql
+            - *copy_generated_sql
             - docker/check:
                 docker-password: DOCKER_PASS
                 docker-username: DOCKER_USER
-      - docker/build: &public-image
-          image: ${CIRCLE_PROJECT_USERNAME+$CIRCLE_PROJECT_USERNAME/}${CIRCLE_PROJECT_REPONAME:-bigquery-etl}
-          tag: ${CIRCLE_TAG:-latest}
```
scholtzan, line 390: It seems unnecessary to build the docker image for every PR. It also takes around 5 minutes. It should be fine just doing it on main or am I missing something here?

.circleci/config.yml
```diff
@@ -461,22 +458,19 @@ jobs:
...
+            - *attach_generated_sql
+            - run:
+                name: Move generated-sql into place
+                command: |
+                  rm -rf sql/
+                  cp -r /tmp/workspace/private-generated-sql/sql sql
             - gcp-gcr/gcr-auth
-      - gcp-gcr/build-image: &private-image
-          image: bigquery-etl
-          tag: ${CIRCLE_TAG:-latest}
```
scholtzan, line 476: Same as `deploy`, just building when the deploy condition is true should be sufficient

.circleci/config.yml
```diff
@@ -507,10 +501,15 @@ jobs:
             ./script/generate_sql \
               --target-project moz-fx-data-shared-prod
 
+            ./script/generate_airflow_dags
+
```
scholtzan, line 505: Adding diff for generated DAGs

.circleci/config.yml
jklukas: I assume the intent here was to catch if we've introduced any errors that will affect the stability of building the image. I agree it's a pretty expensive step that would be nice to skip, though.

jklukas (approved):
r+wc

I don't have strong feelings about whether the docker build steps should stay in or be taken out. But that decision should probably be made in a separate PR. I'd tag @relud for input on that decision.

.circleci/config.yml
relud: klukas has it right. I'm fine with this change.

relud (approved):

PR #2693, opened 2022-01-31
Bug 1752787 ID field type change in fxa_auth_events

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1752787

The source field `jsonPayload.fields.id` changed from STRING to FLOAT64.
This change allows us to handle either case and consistently produce a STRING.

Checklist for reviewer:


jklukas:
I think we should probably not merge this, as @jbuck has indicated that he'll be able to fix up the broken partitions tomorrow, in which case we won't need this code change. It's also not clear to me if this is potentially losing some UUID data.

PR #2700, opened 2022-02-03
Add missing probes

alekhyamoz:
This is related to the discussion of missing probe https://mozilla.slack.com/archives/CB1EQ437S/p1643899307478819
The probe video_dropped_frames_proportion_exponential is not seen in glam UI. 
The table `telemetry_derived.clients_daily_histogram_aggregates_v1` does not show the probe either
I have run the query based on https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_histogram_aggregates_content_v1/query.sql however including the missing probe. 
[Here is the sample query](https://gist.github.com/alekhyamoz/0c3a252f1d8b601beec39b1ccee82e5b) which outputs the missing probe related data 
This PR updates the sql by including the missing probe

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_histogram_aggregates_content_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_histogram_aggregates_content_v1/query.sql
@@ -16443,6 +16443,13 @@ histograms AS (
         (1, 100, 50)
       ),
       (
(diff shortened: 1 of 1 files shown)


scholtzan (approved):
Are all these other changes expected, or will this break something?

alekhyamoz:
> 
Just added the missing probe

PR #2718, opened 2022-02-11
Bug 1751979 Sanitize main_summary_v4 in shredder

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1751979#c9

We need to use Shredder to process `main_summary_v4` as opposed to rerunning
the query that populates it because there is significant history in
`main_summary` that cannot be captured from the modern query on top of
`main_v4`.

All the other affected tables we should be able to handle via other queries without using Shredder.

We can separately have a conversation about how to accelerate the shredder process on main_summary so that it doesn't take a full 28 days.

This reverses and then reapplies some obscure string parsing logic from https://github.com/mozilla/bigquery-etl/blob/222c4266f5a3f30f8232c503b5018bf71980fa61/sql/moz-fx-data-shared-prod/telemetry_derived/main_summary_v4/part1.sql#L266-L273

Checklist for reviewer:


bigquery_etl/shredder/bug1751979.py
```diff
@@ -249,3 +249,119 @@
...
+        CONCAT(engine, '.', source) AS key,
+        count
+      FROM
+        UNNEST(input)
+    ),
+    parsed AS (
+      SELECT
+        *,
+        REGEXP_EXTRACT(key, "([^.]+\\.in-content[:.][^:]+:).*") AS prefix,
+        REGEXP_EXTRACT(key, "[^.]+\\.in-content[:.][^:]+:(.*)") AS code,
```
relud, line 271: I think it would better not to rely on being escaped the right number of times:
```suggestion
        REGEXP_EXTRACT(key, "([^.]+[.]in-content[:.][^:]+:).*") AS prefix,
        REGEXP_EXTRACT(key, "[^.]+[.]in-content[:.][^:]+:(.*)") AS code,
```

bigquery_etl/shredder/bug1751979.py
```diff
@@ -249,3 +249,119 @@
     )
 ) AS payload)
 """
+
+BUG_1751979_MAIN_SUMMARY_V4_UDFS = r"""
+CREATE TEMP FUNCTION sanitize_search_counts_ms(
+  input ARRAY<STRUCT<engine STRING, source STRING, count INT64>>
```
jklukas, line 255: I am realizing now that this is potentially opening us up to duplicate keys, which could cause pain for future analyses. We could have two entries ("google", "in-content:foo", 1) and ("google", "in-content:bar", 1) which should both end up with the code at the end being scrubbed to `other.scrubbed`. We'll have this same issue in other derived tables, so it may be worth turning this into an aggregation as part of this function.

bigquery_etl/shredder/bug1751979.py
```diff
@@ -249,3 +249,119 @@
...
+) AS (
+  (
```
relud, line 257:
```suggestion
  ARRAY(
```

bigquery_etl/shredder/bug1751979.py
```diff
@@ -249,3 +249,119 @@
...
+      --     /moz-fx-data-shared-prod/telemetry_derived/main_summary_v4/part1.sql#L266-L273
+    SELECT
+      ARRAY_AGG(
+        STRUCT(SUBSTR(_key, 0, pos - 2) AS engine, SUBSTR(_key, pos) AS source, count)
+      )
+    FROM
+      parsed,
+      UNNEST([REPLACE(key, 'in-content.', 'in-content:')]) AS _key,
+      UNNEST([LENGTH(REGEXP_EXTRACT(_key, '.+[.].'))]) AS pos
+  )
```
relud, line 361: group by engine and source to avoid duplicate entries:
```suggestion
      SUBSTR(_key, 0, pos - 2) AS engine,
      SUBSTR(_key, pos) AS source,
      SUM(count) AS count,
    FROM
      parsed,
      UNNEST([REPLACE(key, 'in-content.', 'in-content:')]) AS _key,
      UNNEST([LENGTH(REGEXP_EXTRACT(_key, '.+[.].'))]) AS pos
    GROUP BY
      engine,
      source
  )
```

bigquery_etl/shredder/bug1751979.py
relud: added suggestions that accomplish this

relud (approved):
this is running `relud1:US.a805b87e-65d4-4914-88b9-4a952136c4a4` to sanitize `relud1.telemetry_derived.main_summary_v4$20220101` for @jklukas to do data validation.

r+ with data validation

jklukas:
@relud's approach to summing was better than mine, so I ended up incorporating that one and force pushing. My previous commit actually failed to use the `summed` CTE that I had added.

bigquery_etl/shredder/bug1751979.py
```diff
@@ -249,3 +249,124 @@
...
+    SELECT AS STRUCT
+      SUBSTR(_key, 0, pos - 2) AS engine,
+      SUBSTR(_key, pos) AS source,
+      -- We sum by engine and source just in case we have several non-conforming codes that
+      -- all get the same scrubbed value; we want to avoid having duplicate keys.
+      SUM(`count`) AS `count`
+    FROM
+      parsed,
+      UNNEST([REPLACE(key, 'in-content.', 'in-content:')]) AS _key,
+      UNNEST([LENGTH(REGEXP_EXTRACT(_key, '.+[.].'))]) AS pos
```
sean-rose, line 362: Won't the dot in `other.scrubbed` cause problems with this logic?

It seems like the `.+` will be greedy and you'll end up with the engine ending in `in-content:other` and the source being `scrubbed`.

bigquery_etl/shredder/bug1751979.py
```diff
@@ -249,3 +249,124 @@
...
+              "ffsb",
+              "fpas",
+              "fpsa",
+              "ftas",
+              "ftsa",
+              "newext",
+              "monline_dg",
+              "monline_3_dg",
+              "monline_4_dg",
+              "monline_7_dg"
```
sean-rose, line 341: The same comments about the source of these codes would be helpful here for clarity/posterity.

sean-rose (changes requested):

bigquery_etl/shredder/bug1751979.py
jklukas: I think you're right. The existing content already has this problem where the `engine` often has more than it should due to a dot in the latter part of the string.

This same logic is used in the clients_daily query. I will check clients_daily to see if the `other.scrubbed` values coming through the pipeline are already being parsed this way.

jklukas: Indeed, `clients_daily` includes entries like with engine like `bing.in-content:organic:other` and source `scrubbed`.

We need to update this logic in the source queries as well as here.

Thank you so much for scrutinizing this!

sean-rose: IMO using `other-scrubbed` (like was done for mobile) would be ideal, but I suppose that ship has sailed since changing now would require re-reprocessing `main_v4`.

jklukas: Without boiling the ocean here, I think we can achieve what we want by changing `+` to `+?` which according to the [re2 docs](https://github.com/google/re2/wiki/Syntax) means "match one or more, prefer fewer".

sean-rose (approved):

bigquery_etl/shredder/bug1751979.py
sean-rose: Sounds good, and I'm glad to see that non-greedy matching is supported in BigQuery! (coming from Snowflake, where it silently allowed non-greedy indicators but kept being greedy because it didn't really support that option)

jklukas:
> `relud1:US.65aa2c9c-b40f-4ab7-bee4-2fc98b7947d7` is the new job updating `relud1.telemetry_derived.main_summary_v4$20220101` for data validation

I am seeing some bad values for `source` in that partition, such as `in-content:sap:firefoxThe name of the musical ...` from query:

```
SELECT
  sc.*
FROM
  relud1.telemetry_derived.main_summary_v4,
  UNNEST(search_counts) AS sc
WHERE
  submission_date = '2022-01-01'
  AND (LENGTH(engine) > 100
    OR LENGTH(source) > 100)
LIMIT
  1000
```

bigquery_etl/shredder/delete.py
```diff
@@ -261,6 +266,14 @@ def create_job(client):
             ):
                 replace_clause = BUG_1751979_MAIN_V4_REPLACE_CLAUSE
                 temporary_udfs = BUG_1751979_MAIN_V4_UDFS
+            # And corresponding changes for main_summary_v4 which contains values
+            # from one of the affected fields.
+            elif (
+                sql_table_id(target)
+                == "moz-fx-data-shared-prod.telemetry_derived.main_summary_v4"
```
jklukas, line 273: Perhaps I should update these conditions to not specify project. I suspect @relud's test may not have picked up these logic changes due to being in a different project.

relud: ive been manually changing the project specified here, but maybe i don't have it right. i'll double check what query that job ran.

jklukas: Thank you. I will look more deeply into this.

bigquery_etl/shredder/bug1751979.py
```diff
@@ -249,3 +249,124 @@
...
+    -- see https://github.com/mozilla/bigquery-etl/blob/222c4266/sql
+    --     /moz-fx-data-shared-prod/telemetry_derived/main_summary_v4/part1.sql#L266-L273
+    SELECT AS STRUCT
+      SUBSTR(_key, 0, pos - 2) AS engine,
+      SUBSTR(_key, pos) AS source,
+      -- We sum by engine and source just in case we have several non-conforming codes that
+      -- all get the same scrubbed value; we want to avoid having duplicate keys.
+      SUM(`count`) AS `count`
+    FROM
+      parsed,
```
sean-rose, line 360: Here's the problem.
```suggestion
      scrubbed,
```

jklukas: Thank you!

relud:
new job `relud1:US.b0e0e3b5-9e4c-471e-bd89-ef1874a5eecc`

bigquery_etl/shredder/delete.py
jklukas: I believe @sean-rose found the root cause; the fix is now merged and @relud kicked off another test.

bigquery_etl/shredder/bug1751979.py
jklukas: The intention is to remove this entire file once we have a complete run, so we don't need to worry too much about posterity. The existence of this comment thread I expect to be specific enough to trace history when tracing things through GitHub,

sean-rose (approved):

jklukas:
> new job `relud1:US.b0e0e3b5-9e4c-471e-bd89-ef1874a5eecc`

Verified content and it looks good. Merging and I'll coordinate with relud about final plans for deploying this.

PR #2731, opened 2022-02-15
Support clients_daily and search_clients_daily sanitization

jklukas:
See discussion in https://bugzilla.mozilla.org/show_bug.cgi?id=1751979#c13

This should be the final set of tables we need to support in shredder for sanitization. And this will only be applied to the partitions earlier than 2019-11-22.

Checklist for reviewer:

### DIFF
--- a/bigquery_etl/shredder/bug1751979.py
+++ b/bigquery_etl/shredder/bug1751979.py
@@ -480,7 +480,7 @@ CREATE TEMP FUNCTION sanitize_engine_source(
               -- Add a hash of code to avoid multiple rows ending up with the same
               -- (client_id, sanitized_engine, sanitized_source) value, violating
               -- the table's contract.
-              SUBSTR(TO_HEX(SHA256(code)), 1, 4))
+              LEFT(TO_HEX(SHA256(code)), 8))
           ) AS key
       FROM
         parsed
--- a/bigquery_etl/shredder/bug1751979.py
+++ b/bigquery_etl/shredder/bug1751979.py
@@ -474,7 +474,13 @@ CREATE TEMP FUNCTION sanitize_engine_source(
               "monline_7_dg"
             ),
             key,
-            CONCAT(prefix, "other.scrubbed")
+            CONCAT(
+              prefix,
+              "other.scrubbed.",
+              -- Add a hash of code to avoid multiple rows ending up with the same
+              -- (client_id, sanitized_engine, sanitized_source) value, violating
+              -- the table's contract.
+              SUBSTR(TO_HEX(SHA256(code)), 1, 4))
           ) AS key
       FROM
         parsed
--- a/bigquery_etl/shredder/bug1751979.py
+++ b/bigquery_etl/shredder/bug1751979.py
@@ -370,3 +370,128 @@ CREATE TEMP FUNCTION sanitize_search_counts_ms(
 BUG_1751979_MAIN_SUMMARY_V4_REPLACE_CLAUSE = r"""
   REPLACE (sanitize_search_counts_ms(search_counts) AS search_counts)
 """
+
+BUG_1751979_CLIENTS_DAILY_V6_REPLACE_CLAUSE = r"""
+REPLACE (
+  sanitize_search_counts_ms(search_counts) AS search_counts,
+  sanitize_scalar(search_content_urlbar_sum) AS search_content_urlbar_sum,
+  sanitize_scalar(search_content_urlbar_handoff_sum) AS search_content_urlbar_handoff_sum,
+  sanitize_scalar(
+    search_content_urlbar_searchmode_sum
+  ) AS search_content_urlbar_searchmode_sum,
+  sanitize_scalar(search_content_searchbar_sum) AS search_content_searchbar_sum,
+  sanitize_scalar(search_content_about_home_sum) AS search_content_about_home_sum,
+  sanitize_scalar(search_content_about_newtab_sum) AS search_content_about_newtab_sum,
+  sanitize_scalar(search_content_contextmenu_sum) AS search_content_contextmenu_sum,
+  sanitize_scalar(search_content_webextension_sum) AS search_content_webextension_sum,
+  sanitize_scalar(search_content_system_sum) AS search_content_system_sum,
+  sanitize_scalar(search_content_tabhistory_sum) AS search_content_tabhistory_sum,
+  sanitize_scalar(search_content_reload_sum) AS search_content_reload_sum,
+  sanitize_scalar(search_content_unknown_sum) AS search_content_unknown_sum
+)
+"""
+
+BUG_1751979_SEARCH_CLIENTS_DAILY_V8_UDFS = r"""
+CREATE TEMP FUNCTION sanitize_engine_source(
+  engine STRING, `source` STRING
+) AS ((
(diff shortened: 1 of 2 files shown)


bigquery_etl/shredder/bug1751979.py
```diff
@@ -370,3 +370,128 @@
...
+  sanitize_scalar(search_content_reload_sum) AS search_content_reload_sum,
+  sanitize_scalar(search_content_unknown_sum) AS search_content_unknown_sum
+)
+"""
+
+BUG_1751979_SEARCH_CLIENTS_DAILY_V8_UDFS = r"""
+CREATE TEMP FUNCTION sanitize_engine_source(
+  engine STRING, `source` STRING
+) AS ((
+    WITH base AS (
```
jklukas, line 398: This logic is again copied from the existing `sanitize_search_counts_ms` UDF and refactored to apply to a flat `engine` and `source` input rather than a repeated array of values. Since `search_clients_daily` is basically clients_daily unnested on `search_counts`.

bigquery_etl/shredder/bug1751979.py
```diff
@@ -370,3 +370,128 @@
...
+  sanitize_scalar(search_content_tabhistory_sum) AS search_content_tabhistory_sum,
+  sanitize_scalar(search_content_reload_sum) AS search_content_reload_sum,
+  sanitize_scalar(search_content_unknown_sum) AS search_content_unknown_sum
+)
+"""
+
+BUG_1751979_SEARCH_CLIENTS_DAILY_V8_UDFS = r"""
+CREATE TEMP FUNCTION sanitize_engine_source(
+  engine STRING, `source` STRING
+) AS ((
```
sean-rose, line 397: Doubled parentheses?

bigquery_etl/shredder/bug1751979.py
```diff
@@ -370,3 +370,128 @@
...
+      UNNEST([REPLACE(key, 'in-content.', 'in-content:')]) AS _key,
+      UNNEST([LENGTH(REGEXP_EXTRACT(_key, '.+?[.].'))]) AS pos
+));
+"""
+
+BUG_1751979_SEARCH_CLIENTS_DAILY_V8_REPLACE_CLAUSE = r"""
+REPLACE (
+  sanitize_engine_source(engine, `source`).engine AS engine,
+  sanitize_engine_source(engine, `source`).source AS source
+)
```
sean-rose, line 496: This will result in multiple `search_clients_daily_v8` rows with the same date + engine + source = "other.scrubbed" values.  That doesn't necessarily seem like a dealbreaker to me, but I think people should be made aware of that.

sean-rose (approved):

bigquery_etl/shredder/bug1751979.py
relud: they are necessary to indicate a subqquery expression within the definition of a sql function

bigquery_etl/shredder/bug1751979.py
jklukas: I agree this is a concern. I will bring it up with the revenue WG

jklukas:
Asking for review again to vet the hashed `code` solution to preventing rows with duplicate primary keys.

bigquery_etl/shredder/bug1751979.py
```diff
@@ -370,3 +370,134 @@
...
+      FROM
+        parsed
+    )
+    SELECT AS STRUCT
+      SUBSTR(_key, 0, pos - 2) AS engine,
+      SUBSTR(_key, pos) AS `source`,
+    FROM
+      scrubbed,
+      UNNEST([REPLACE(key, 'in-content.', 'in-content:')]) AS _key,
+      UNNEST([LENGTH(REGEXP_EXTRACT(_key, '.+?[.].'))]) AS pos
```
relud, line 494: seems like the `.+` vs `.+?` change could still result in duplicate client/engine/source entries, but we know it doesn't in practice, that's good enough.

relud (approved):
r+wc

bigquery_etl/shredder/bug1751979.py
```diff
@@ -370,3 +370,134 @@
...
+              "monline_7_dg"
+            ),
+            key,
+            CONCAT(
+              prefix,
+              "other.scrubbed.",
+              -- Add a hash of code to avoid multiple rows ending up with the same
+              -- (client_id, sanitized_engine, sanitized_source) value, violating
+              -- the table's contract.
+              SUBSTR(TO_HEX(SHA256(code)), 1, 4))
```
sean-rose, line 483: * How confident are we that a 4-character hash will be enough to not have any collisions? (65k isn't that big of a hash space)
* This won't entirely avoid some potential adverse effects of having a large number of `source` values (these "other.scrubbed.xxxx" values won't easily aggregate together, and having thousands of such values could overwhelm users if they're someplace like in a BI tool filter select list).
* `SUBSTR(value, 1, 4)` could be more succinctly/clearly done with `LEFT(value, 4)`.

sean-rose (approved):

bigquery_etl/shredder/bug1751979.py
jklukas, line 483:
```suggestion
              LEFT(TO_HEX(SHA256(code)), 8))
```

bigquery_etl/shredder/bug1751979.py
jklukas: I think likelihood is low, and this is a stopgap for now. I think we should fully rerun this in a month, after we're sure `clients_daily` is scrubbed through all history.

bigquery_etl/shredder/bug1751979.py
jklukas: I've updated to use LEFT and take 8 characters rather than 4.

> these "other.scrubbed.xxxx" values won't easily aggregate together, and having thousands of such values could overwhelm users if they're someplace like in a BI tool filter select list

Agreed, but the situation will be no worse than it is today with unscrubbed values. We've properly recreated the data from 2019-11-22 and later from `clients_daily`, so this logic will only be relevant for older history. We can do a full recreation in a month after we're sure the entirety of `clients_daily` history is sanitized.

bigquery_etl/shredder/delete.py
```diff
@@ -274,6 +277,24 @@ def create_job(client):
...
+                "moz-fx-data-shared-prod.telemetry_derived.clients_last_seen_joined_v1",
+            ]:
+                replace_clause = BUG_1751979_CLIENTS_DAILY_V6_REPLACE_CLAUSE
+                temporary_udfs = (
+                    BUG_1751979_MAIN_SUMMARY_V4_UDFS + BUG_1751979_MAIN_V4_UDFS
+                )
+            # And corresponding changes for search_clients_daily-derived tables.
+            elif (
+                sql_table_id(target)
+                == "moz-fx-data-shared-prod.telemetry_derived.search_clients_daily_v8"
```
relud, line 294: @jklukas of these only `clients_daily_v6` is actually being scrubbed. `telemetry_derived.clients_daily_joined_v1` and `telemetry_derived.clients_last_seen_joined_v1` aren't configured for deletes, and `search_clients_daily_v8` is in the `search_derived` dataset, not the `telemetry_derived` dataset.

PR #2749, opened 2022-02-23
add process to sample counts glam tables

alekhyamoz:
add process to sample counts glam tables. 
Will need to create a subsequent PR to add to the extract tables.

### DIFF
--- a/bigquery_etl/glam/templates/extract_sample_counts_v1.sql
+++ b/bigquery_etl/glam/templates/extract_sample_counts_v1.sql
@@ -21,3 +21,13 @@ WHERE
   cp.app_version IS NOT NULL
   AND total_users > {{ total_users }}
   AND cp.client_agg_type not in ('sum','min','avg','max')
+GROUP BY
+  cp.channel,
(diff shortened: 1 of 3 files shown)


bigquery_etl/glam/templates/extract_sample_counts_v1.sql
```diff
@@ -20,4 +20,14 @@ AND total_sample IS NOT NULL
...
+GROUP BY
+  cp.channel,
+  cp.ping_type,
+  cp.app_version,
+  cp.app_build_id,
+  cp.os,
+  cp.metric,
+  cp.key,
+  client_agg_type,
+  total_sample
```
scholtzan, line 33: Is the grouping here used to filter out duplicate entries?

scholtzan (approved):
looks good to me 👍

bigquery_etl/glam/templates/extract_sample_counts_v1.sql
alekhyamoz: yes you are correct

PR #2753, opened 2022-02-24
[Bug 1756829] Fix clients_last_seen joins

scholtzan:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1756829

Checklist for reviewer:


sql_generators/glean_usage/templates/baseline_clients_last_seen_v1.query.sql
```diff
@@ -70,7 +70,12 @@ FROM
...
   _previous
-USING
-  (client_id)
+ON
+  _previous.client_id = _current.client_id AND
+  _previous.sample_id = _current.sample_id AND
+  (
+    _previous.normalized_channel = _current.normalized_channel OR
+    (_previous.normalized_channel IS NULL AND _current.normalized_channel IS NULL)
+  )
```
jklukas, line 79: The contract for this table should be that there's only one row per client_id per `submission_date`. Do we meet that guarantee using this modified join condition?


The baseline_clients_* ETL is per appId whereas metrics_clients_* is per-appName. I don't think this logic makes sense for per-appId logic. We should be able to guarantee only a single client_id in the daily table.

jklukas (changes requested):

sql_generators/glean_usage/templates/baseline_clients_last_seen_v1.query.sql
scholtzan: Okay, so I have been digging a little more and it appears that the client daily table actually does have multiple entries for a single client per `submission_date`. The reason for this is that in the `baseline` table some clients  have baseline pings with a `sample_id` and sometimes these same clients send baseline pings with no `sample_id` (`sample_id = NULL`). For these clients `clients_daily` has two entries.

It looks like this might have been an ingestion issue which has been fixed (I'm only seeing this for clients in 2019).
After 2019 `clients_daily` has indeed a single entry per client.

`clients_last_seen` still have duplicate entries to this date however, since it always joins on itself for the previous date. I think the way to go is to filter out entries where `sample_id` is `NULL`.

sql_generators/glean_usage/templates/baseline_clients_last_seen_v1.query.sql
```diff
@@ -53,6 +54,7 @@ _previous AS (
     submission_date = DATE_SUB(@submission_date, INTERVAL 1 DAY)
     -- Filter out rows from yesterday that have now fallen outside the 28-day window.
     AND udf.shift_28_bits_one_day(days_seen_bits) > 0
+    AND sample_id IS NOT NULL
```
scholtzan, line 57: Some additional discussion and an example query illustrating the problem is also here: https://mozilla.slack.com/archives/C4D5ZA91B/p1646085773661049

sql_generators/glean_usage/templates/baseline_clients_last_seen_v1.query.sql
jklukas: > It looks like this might have been an ingestion issue which has been fixed (I'm only seeing this for clients in 2019).

Indeed, we didn't add support for parsing client_id out of Glean pings to populate sample_id until May 2019. Implemented in https://github.com/mozilla/gcp-ingestion/pull/612

jklukas: > I think the way to go is to filter out entries where sample_id is NULL

That would be fine. Even better would be to add sample_id in to the stable tables where it's missing, which we could do via query (and the data volume should be pretty tiny for 2019 data).

jklukas (approved):

PR #2757, opened 2022-02-28
Internet Outages DAG schedule interval update

kik-kik:
updated schedule interval to run later due to query running with missing data.

Relates to: https://bugzilla.mozilla.org/show_bug.cgi?id=1756985

If this is not enough to solve the problem then we'll need to do a deep dive into the dependencies and fully understand all processes involved in serving the data as well as when they occur. This is also something that could be useful in general.

Changes:
- Updated schedule in `dags.yaml`
- Re-run `bqetl dag generate` command


Dexterp37 (approved):
Thank you very much!

PR #2766, opened 2022-03-01
DVPN-111 - Wait even longer for stripe report

relud:
32 retries makes this try until 9:15am PT, so that we can at least see when it comes in before probably moving it to a separate dag.

also wait up to 30 minutes for the report to finish, because the most recent run didn't finish in 10 minutes.


sql/moz-fx-data-shared-prod/stripe_external/itemized_payout_reconciliation_v5/metadata.yaml
```diff
@@ -17,5 +17,5 @@ scheduling:
     - --time-partitioning-field=automatic_payout_effective_at
   # report data isn't generally ready when the dag starts, so retry for a while
   retry_delay: 30m
-  retries: 10
+  retries: 32
```
sean-rose, line 20: Since we don't yet know when the report becomes available, to avoid possibly having to do this again I'd consider increasing this to 47 to cover the entire day.

sean-rose (approved):

PR #2784, opened 2022-03-10
Update query.sql for DVPN-126

YeonjooSmith:


relud (approved):
r+ with using `JSON` to group by an `ARRAY<STRING>`

PR #2785, opened 2022-03-10
for adding promotion_codes colum DVPN-126

YeonjooSmith:


relud:
this was already accomplished by #2784

PR #2787, opened 2022-03-10
Bug 1759019 - Don't drop partitions when shredding whole tables

relud:
Checklist for reviewer:

### DIFF
--- a/bigquery_etl/shredder/delete.py
+++ b/bigquery_etl/shredder/delete.py
@@ -258,7 +258,7 @@ def delete_from_partition(
                 # always true partition condition to satisfy require_partition_filter
                 partition_condition = f"""
                 ({partition.condition}) IS NOT TRUE
-                OR ({partition.condition})"
+                OR ({partition.condition})
                 """
             else:
                 partition_condition = partition.condition
@@ -272,7 +272,7 @@ def delete_from_partition(
                 {field_joins}
                 WHERE
                   ({field_conditions})
-                  AND ({partition_condition})"
+                  AND ({partition_condition})
                 """
             )
         run_tense = "Would run" if dry_run else "Running"
--- a/bigquery_etl/shredder/delete.py
+++ b/bigquery_etl/shredder/delete.py
@@ -223,7 +223,7 @@ def delete_from_partition(
                   `{sql_table_id(target)}`
                 WHERE
                   ({field_condition})
-                  AND {partition.condition}
+                  AND ({partition.condition})
                 """
             )
         else:
@@ -249,6 +249,18 @@ def delete_from_partition(
                 f"_source_{index} IS NULL" for index, _ in enumerate(sources)
             )
 
+            if partition.id is None:
+                condition = f"""
+                (
+                  ({partition.condition}) IS NOT TRUE
+                  OR ({field_conditions})
+                )
+                """
+                if partition.required_partition_filter:
+                    condition += f" AND ({partition.required_partition_filter})"
+            else:
+                condition = f"({field_conditions}) AND ({partition.condition})"
+
             query = reformat(
                 f"""
                 SELECT
@@ -257,8 +269,7 @@ def delete_from_partition(
                   `{sql_table_id(target)}` AS _target
                 {field_joins}
                 WHERE
-                  ({field_conditions})
-                  AND {partition.condition}
+                  {condition}
                 """
             )
         run_tense = "Would run" if dry_run else "Running"
@@ -285,13 +296,24 @@ class Partition:
     condition: str
     id: Optional[str] = None
     is_special: bool = False
+    required_partition_filter: Optional[str] = None
 
 
 def get_partition(table, partition_expr, end_date, id_=None) -> Optional[Partition]:
     """Return a Partition for id_ unless it is a date on or after end_date."""
     if id_ is None:
         if table.time_partitioning:
-            return Partition(condition=f"{partition_expr} < '{end_date}'")
+            return Partition(
+                condition=f"{partition_expr} < '{end_date}'",
+                required_partition_filter=(
+                    # explicitly select all partitions
+                    f"{partition_expr} IS NULL"
+                    f" OR {partition_expr} < CURRENT_DATE"
+                    f" OR {partition_expr} >= CURRENT_DATE"
+                )
+                if table.require_partition_filter
+                else None,
+            )
         return Partition(condition="TRUE")
     if id_ == NULL_PARTITION_ID:
         if table.time_partitioning:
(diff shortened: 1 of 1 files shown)


bigquery_etl/shredder/delete.py
```diff
@@ -251,6 +251,7 @@ def create_job(client):
                   )
                 ON
                   {field} = _source_{index}
+                  AND {partition.on_condition}
```
sean-rose, line 254: This initially struck me as odd, since the partition condition only references the target table and not the other side of the join, but I finally grokked that this is forcing the join to fail for target rows that don't meet the partition condition so that `_source_{index} IS NULL` is true and all such rows are still included.

Another way to do that would be `OR {partition.condition} IS NOT TRUE` in the where clause, which I'm guessing would be a bit more computationally efficient, but having the logic to switch between `AND`/`OR` for the different cases seems like it'd be super awkward, so what you've got here is probably better in terms of readability/maintainability.

bigquery_etl/shredder/delete.py
```diff
@@ -334,6 +335,24 @@ class Partition:
...
+    def on_condition(self):
+        """Condition to use in the ON clause of joins.
+
+        Use partition condition when shredding a whole table and TRUE when
+        shredding a single partition.
+        """
+        return self.condition if self.id is None else "TRUE"
+
+    @property
+    def where_condition(self):
```
sean-rose, line 348: `select_where_condition` might be a slightly more accurate name.  At first I thought maybe `partition.condition` still being used in the `DELETE` statement was an oversight, but then I came back and read the description here more carefully (kudos for having docstrings).

sean-rose (approved):

sean-rose:
On a semi-related note, I think there's a bug lurking [here](https://github.com/mozilla/bigquery-etl/blob/b2b7a330c0f11ecdade27dbf840aa8e0a867bcf6/bigquery_etl/shredder/delete.py#L360-L361), where one of the possible partition conditions is a compound `OR` expression but isn't wrapped in parentheses:
```python
                condition=f"{partition_expr} < {table.range_partitioning.range_.start} "
                f"OR {partition_expr} >= {table.range_partitioning.range_.end}",
```
So when inserted into the where clause with `AND` expressions it won't do what's intended because `AND` has higher precedence than `OR`.

I was considering suggesting we wrap all _usages_ of the partition conditions in parentheses to guard against that, but since all the other partition conditions are single expressions it's probably easiest/sufficient to wrap that one compound expression in parentheses at the source.

relud:
> I was considering suggesting we wrap all _usages_ of the partition conditions in parentheses to guard against that, but since all the other partition conditions are single expressions it's probably easiest/sufficient to wrap that one compound expression in parentheses at the source.

I think that's the right solution, and i've applied it in this PR

relud:
waiting to merge this until I have a moment to verify that it works as expected, because shredder has already run all of the whole-table deletes for this cycle.

bigquery_etl/shredder/delete.py
```diff
@@ -249,6 +249,18 @@ def create_job(client):
...
+                condition = f"""
+                (
+                  ({partition.condition}) IS NOT TRUE
+                  OR ({field_conditions})
+                )
+                """
+                if partition.required_partition_filter:
+                    condition += f" AND ({partition.required_partition_filter})"
+            else:
+                condition = f"({field_conditions}) AND ({partition.condition})"
```
sean-rose, line 262: The `partition.required_partition_filter` stuff seems a bit kludgy.  If the problem is that the `({field_conditions})` branch of the `OR` doesn't have a partition condition then I think we could simply include `AND ({partition.condition})` there, and potentially simplify this into something like:
```suggestion
            condition = f"({field_conditions}) AND ({partition.condition})"
            if partition.id is None:
                condition = f"({condition}) OR (({partition.condition}) IS NOT TRUE)"
```
(the parentheses around `({partition.condition}) IS NOT TRUE` wouldn't technically be required, but I think it'd help avoid potential confusion about whether the `IS NOT TRUE` applied to the result of the `OR` expression)

bigquery_etl/shredder/delete.py
```diff
@@ -285,13 +296,24 @@ class Partition:
...
+            return Partition(
+                condition=f"{partition_expr} < '{end_date}'",
+                required_partition_filter=(
+                    # explicitly select all partitions
+                    f"{partition_expr} IS NULL"
+                    f" OR {partition_expr} < CURRENT_DATE"
+                    f" OR {partition_expr} >= CURRENT_DATE"
+                )
+                if table.require_partition_filter
+                else None,
```
sean-rose, line 315: IMO this would be more easily understandable if the `if`/`else` were within the parentheses.
```suggestion
                required_partition_filter=(
                    # explicitly select all partitions
                    f"{partition_expr} IS NULL"
                    f" OR {partition_expr} < CURRENT_DATE"
                    f" OR {partition_expr} >= CURRENT_DATE"
                    if table.require_partition_filter
                    else None,
                )
```

sean-rose (approved):
`r+wc`

bigquery_etl/shredder/delete.py
relud: I just checked, and this doesn't satisfy `require_partition_filter`

relud: how about this solution that doesn't use `required_partition_filter`, and satisfies `require_partition_filter`:
```suggestion
            if partition.id is None:
                # only apply field conditions on partition.condition
                field_conditions = f"""
                ({partition.condition}) IS NOT TRUE
                OR ({field_conditions})
                """
                # always true partition condition to satisfy require_partition_filter
                partition_condition = f"""
                ({partition.condition}) IS NOT TRUE
                OR ({partition.condition})"
                """
            else:
                partition_condition = partition.condition
                
            condition = f"({field_conditions}) AND ({partition_condition})"
```

bigquery_etl/shredder/delete.py
```diff
@@ -250,16 +250,18 @@ def create_job(client):
...
-                )
+                # only apply field conditions on partition.condition
+                field_conditions = f"""
+                ({partition.condition}) IS NOT TRUE
+                OR ({field_conditions})
+                """
+                # always true partition condition to satisfy require_partition_filter
+                partition_condition = f"""
+                ({partition.condition}) IS NOT TRUE
+                OR ({partition.condition})"
```
sean-rose, line 261:
```suggestion
                OR ({partition.condition})
```

bigquery_etl/shredder/delete.py
```diff
@@ -269,7 +271,8 @@ def create_job(client):
                   `{sql_table_id(target)}` AS _target
                 {field_joins}
                 WHERE
-                  {condition}
+                  ({field_conditions})
+                  AND ({partition_condition})"
```
sean-rose, line 275:
```suggestion
                  AND ({partition_condition})
```

sean-rose (approved):
`r+wc`

PR #2793, opened 2022-03-14
Exclude probes that cause error on GLAM from ETL (#2792)

### LINKED ISSUE #2792: Remove probes that cause error on GLAM from ETL
Implement a quickfix to stop calculating probes that cause error on GLAM
[More context here](https://github.com/mozilla/glam/issues/1865).

edugfilho:
Fixes #2792

This implements a quick-fix to stop probes that cause error on GLAM from being aggregated.
There's upcoming work fixing [another issue](https://github.com/mozilla/glam/issues/1847) that will also allow us to replace the probe names hardcoded in this PR with api calls that will return them.

Checklist for reviewer:

### DIFF
--- a/bigquery_etl/glam/clients_daily_histogram_aggregates.py
+++ b/bigquery_etl/glam/clients_daily_histogram_aggregates.py
@@ -6,6 +6,7 @@ from typing import Dict, List
 from jinja2 import Environment, PackageLoader
 
 from bigquery_etl.format_sql.formatter import reformat
+from bigquery_etl.util.probe_filters import get_etl_excluded_probes_quickfix
 
 from .utils import get_schema, ping_type_from_table
 
@@ -42,6 +43,7 @@ def get_distribution_metrics(schema: Dict) -> Dict[str, List[str]]:
         "custom_distribution",
     }
     metrics: Dict[str, List[str]] = {metric_type: [] for metric_type in metric_type_set}
+    excluded_metrics = get_etl_excluded_probes_quickfix()
 
     # Iterate over every element in the schema under the metrics section and
     # collect a list of metric names.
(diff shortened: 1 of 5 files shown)


edugfilho:
@alekhyamoz Do you think we need to add a filter to [here](https://github.com/mozilla/bigquery-etl/blob/6b0b918cc6c205ff133818696f253d9f76338159/bigquery_etl/glam/clients_daily_histogram_aggregates.py#L60) and [here](https://github.com/mozilla/bigquery-etl/blob/6b0b918cc6c205ff133818696f253d9f76338159/bigquery_etl/glam/clients_daily_scalar_aggregates.py#L88) as well? I think those functions are used by the `bqetl` cmd, right?

alekhyamoz:
Great job on the PR! Thank you for working on this @edugfilho 
Just one thing, though I have not come across a scenario of both Fenix and Firefox Desktop having the same list of probes. But at the same time, I'm not confident to rule out the possibility that the same metric name can be shared between the products. In such scenarios, the probe_lists might need to be separate lists. Again, need to verify. 
Also, this PR is going to tackle the particular issue mentioned. Otherwise, I am assuming that this logic will go away once we have the API call in place as it would list the probes to be included in the calculation. (not excluded)

edugfilho:
> Great job on the PR! Thank you for working on this @edugfilho Just one thing, though I have not come across a scenario of both Fenix and Firefox Desktop having the same list of probes. But at the same time, I'm not confident to rule out the possibility that the same metric name can be shared between the products. In such scenarios, the probe_lists might need to be separate lists. Again, need to verify. Also, this PR is going to tackle the particular issue mentioned. Otherwise, I am assuming that this logic will go away once we have the API call in place as it would list the probes to be included in the calculation. (not excluded)

Thank you, @alekhyamoz. That's a valid concern, and you're also right about the permanent solution to this being an allow-list instead of a block-list. Now in that case, when metric names repeat across products, the ETL might aggregate metrics that aren't required for one product just because those metrics have the same name as required metrics for another. But there's a chance we actually want that to be the case.

I want to add a `product` parameter to the function that provides the list. What do you think?

akkomar (approved):
LGTM, I'd maybe rewrite commit message to something like `Exclude probes that cause error in GLAM from ETL"

alekhyamoz (approved):
lgtm

PR #2807, opened 2022-03-17
DVPN-83 - Migrate stripe to fivetran

relud:
I have validated that this results in 0 rows changed in `itemized_payout_reconciliation` and `funnel_product_page_to_subscribed_v1`, and that the only changes to `mozdata.subscription_platform.stripe_subscriptions` are due to fivetran having more recent updates to the tables because fivetran ETL runs at 0:58 UTC and collects all changes up to that time, while the old ETL collected changes up to UTC midnight.


relud:
I have validated all fields in prod match except status, customer_start_date, fxa_uid, country, and provider, and i'm converting this to a draft until I work out why those fields aren't matching.

sql/moz-fx-data-shared-prod/subscription_platform/stripe_subscriptions/view.sql
```diff
@@ -1,121 +1,151 @@
...
+    TO_HEX(SHA256(JSON_VALUE(metadata, "$.userid"))) AS fxa_uid,
+    address_country,
   FROM
-    mozdata.stripe.customers
+    `moz-fx-data-bq-fivetran`.stripe.customer
+  UNION ALL
+  -- include customer records that were deleted before the initial fivetran stripe import
+  SELECT
+    id AS customer_id,
+    JSON_VALUE(metadata, "$.fxa_uid") AS fxa_uid,
```
relud, line 48: The difference between `TO_HEX(SHA256(JSON_VALUE(metadata, "$.userid"))) AS fxa_uid,` and `JSON_VALUE(metadata, "$.fxa_uid") AS fxa_uid,` on either side of this `UNION ALL` is because of https://github.com/mozilla/bigquery-etl/blob/78885a77dd17bc26585e2f4ad7d0c2199da02ee2/bigquery_etl/stripe/allowed_fields.py#L113-L117

sql/moz-fx-data-shared-prod/subscription_platform/stripe_subscriptions/view.sql
```diff
@@ -1,121 +1,151 @@
...
+    address_country,
+  FROM
+    `moz-fx-data-shared-prod`.stripe_external.pre_fivetran_customer
```
relud, line 51: This table was generated to capture customer records that are no longer in stripe, via this query:
```sh
bq query --destination_table=moz-fx-data-shared-prod:stripe_external.pre_fivetran_customer --replace '
CREATE TEMP FUNCTION map_to_json(metadata ARRAY<STRUCT<key STRING, value STRING>>)
RETURNS STRING
LANGUAGE js
AS
"""
result = {};
for (const element of metadata) {
  result[element.key] = element.value;
};
return JSON.stringify(result);
""";

SELECT
  customers_v1.id,
  customers_v1.address.country AS address_country,
  customers_v1.address.postal_code AS address_postal_code,
  customers_v1.address.state AS address_state,
  TRUE AS is_deleted,
  map_to_json(customers_v1.metadata) AS metadata,
FROM
  `moz-fx-data-shared-prod`.stripe_external.customers_v1
LEFT JOIN
  `moz-fx-data-bq-fivetran`.stripe.customer
USING
  (id)
WHERE
  customer.id IS NULL
'
```

and adding this populated all of the `fxa_uid` values that were missing when validating the results of this PR.

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/all_subscriptions_v1/query.sql
```diff
@@ -368,6 +349,7 @@ vpn_subscriptions AS (
 vpn_subscriptions_with_end_date AS (
   SELECT
     *,
+    MIN(subscription_start_date) OVER (PARTITION BY customer_id) AS customer_start_date,
```
relud, line 352: `customer_start_date` moved here from the `subscription_platform.stripe_subscriptions` view because it shouldn't be calculated across all products.

sql/moz-fx-data-shared-prod/subscription_platform/stripe_subscriptions/view.sql
```diff
@@ -1,121 +1,151 @@
...
-      -- FxA copies paypal billing agreement location to customer_address
-      STRUCT("Paypal" AS provider, invoices_v1.customer_address.country),
-      ("Stripe", stripe_charges.country)
+      JSON_VALUE(invoice.metadata, "$.paypalTransactionId") IS NOT NULL,
+      -- FxA copies paypal billing agreement country to customer address
+      STRUCT("Paypal" AS provider, customer.address_country AS country),
+      ("Stripe", charge.country)
     ).*,
-    invoices_v1.event_timestamp,
+    invoice.finalized_at,
```
sean-rose, line 75: `finalized_at` may be null, so using `created` might be preferable.

sean-rose (approved):
`r+wc`

A couple of miscellaneous questions:
- This preserves customers that may have been deleted from Stripe before the Fivetran sync started.  Does the same perhaps need to be done for any other entities?
- Right now the sync is scheduled in Fivetran, but would it be better to use `depends_on_fivetran` to have it triggered by Airflow?

sql/moz-fx-data-shared-prod/subscription_platform/stripe_subscriptions/view.sql
sean-rose: Good thinking.  I'd be inclined to also include `created`.

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/funnel_product_page_to_subscribed_v1/query.sql
```diff
@@ -1,8 +1,8 @@
 -- this will be updated once the service parameter get passed for "fxa_rp_button - view"
-WITH pricing_plans AS (
+WITH stripe_plan AS (
```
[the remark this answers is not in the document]
relud: okay, will replace with plurals.

sql/moz-fx-data-shared-prod/subscription_platform/nonprod_stripe_subscriptions/view.sql
```diff
@@ -1,121 +1,144 @@
...
+),
+customer AS (
   SELECT
     id AS customer_id,
-    mozfun.map.get_key(metadata, "fxa_uid") AS fxa_uid,
+    TO_HEX(SHA256(JSON_VALUE(metadata, "$.userid"))) AS fxa_uid,
+    address_country,
   FROM
-    `moz-fx-data-shared-prod`.stripe_derived.nonprod_customers_v1
+    `dev-fivetran`.stripe_nonprod.customer
```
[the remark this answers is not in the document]
relud: no, nonprod doesn't have any history previously imported to BQ that isn't still available.

sql/moz-fx-data-shared-prod/subscription_platform/nonprod_stripe_subscriptions/view.sql
```diff
@@ -1,121 +1,144 @@
...
 SELECT
   customer_id,
   subscription_id,
+  subscription_item_id,
   plan_id,
   status,
-  event_timestamp,
-  customer_start_date,
+  _fivetran_synced AS event_timestamp,
+  MIN(subscription_start_date) OVER (PARTITION BY customer_id) AS customer_start_date,
```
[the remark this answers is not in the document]
relud: yes

sql/moz-fx-data-shared-prod/subscription_platform/stripe_subscriptions/view.sql
relud: okay

relud:
> This preserves customers that may have been deleted from Stripe before the Fivetran sync started. Does the same perhaps need to be done for any other entities?

i won't immediately delete the non-fivetran tables, but validation didn't show any other entities missing.

relud:
> Right now the sync is scheduled in Fivetran, but would it be better to use depends_on_fivetran to have it triggered by Airflow?

let's make that a follow-up PR to unblock things depending on this.

sql/moz-fx-data-shared-prod/subscription_platform/stripe_subscriptions/view.sql
```diff
@@ -1,121 +1,151 @@
...
+    -- current_period_end,
+    -- current_period_start,
     canceled_at,
-    mozfun.map.get_key(metadata, "cancelled_for_customer_at") AS canceled_for_customer_at,
+    JSON_VALUE(metadata, "$.cancelled_for_customer_at") AS canceled_for_customer_at,
     cancel_at,
     cancel_at_period_end,
-    ended_at,
-    discount.promotion_code AS promotion_code_id,
+    IF(ended_at <= _fivetran_synced, ended_at, NULL) AS ended_at,
```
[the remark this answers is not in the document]
relud: > just passing `ended_at` through as is and letting downstream logic handle it as necessary?

I'll go with this for now, and open a PR to change it later if necessary.

sql/moz-fx-data-shared-prod/subscription_platform/stripe_subscriptions/view.sql
relud: added `created`

PR #2823, opened 2022-03-24
DVPN-119 - Add ETL for Non-Prod CJMS

relud:
~blocked on #2807~

### DIFF
--- a/dags/bqetl_cjms_nonprod.py
+++ b/dags/bqetl_cjms_nonprod.py
@@ -1,0 +1,84 @@
+# Generated via https://github.com/mozilla/bigquery-etl/blob/main/bigquery_etl/query_scheduling/generate_airflow_dags.py
+
+from airflow import DAG
+from operators.task_sensor import ExternalTaskCompletedSensor
+import datetime
+from utils.gcp import bigquery_etl_query, gke_command
+
+docs = """
+### bqetl_cjms_nonprod
+
+Built from bigquery-etl repo, [`dags/bqetl_cjms_nonprod.py`](https://github.com/mozilla/bigquery-etl/blob/main/dags/bqetl_cjms_nonprod.py)
+
+#### Description
+
+Hourly ETL for cjms nonprod.
+
+Scheduled to run after fivetran stripe_nonprod, which runs hourly at minute
+17 and takes less than 5 minutes.
+
+#### Owner
+
+dthorn@mozilla.com
+"""
+
+
+default_args = {
+    "owner": "dthorn@mozilla.com",
+    "start_date": datetime.datetime(2022, 3, 24, 0, 0),
+    "end_date": None,
+    "email": ["telemetry-alerts@mozilla.com", "dthorn@mozilla.com"],
+    "depends_on_past": False,
+    "retry_delay": datetime.timedelta(seconds=300),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
+}
+
+tags = ["impact/tier_3", "repo/bigquery-etl"]
+
+with DAG(
+    "bqetl_cjms_nonprod",
+    default_args=default_args,
+    schedule_interval="22 * * * *",
+    doc_md=docs,
+    tags=tags,
+) as dag:
+
+    cjms_bigquery__flows__v1 = bigquery_etl_query(
+        task_id="cjms_bigquery__flows__v1",
+        destination_table="flows_v1",
+        dataset_id="cjms_bigquery",
+        project_id="moz-fx-cjms-nonprod-9a36",
+        owner="dthorn@mozilla.com",
+        email=["dthorn@mozilla.com", "telemetry-alerts@mozilla.com"],
+        date_partition_parameter="submission_date",
+        depends_on_past=False,
+        dag=dag,
+    )
+
+    cjms_bigquery__refunds__v1 = bigquery_etl_query(
+        task_id="cjms_bigquery__refunds__v1",
+        destination_table="refunds_v1",
+        dataset_id="cjms_bigquery",
+        project_id="moz-fx-cjms-nonprod-9a36",
+        owner="dthorn@mozilla.com",
+        email=["dthorn@mozilla.com", "telemetry-alerts@mozilla.com"],
+        date_partition_parameter=None,
+        depends_on_past=False,
+        dag=dag,
+    )
+
+    cjms_bigquery__subscriptions__v1 = bigquery_etl_query(
+        task_id="cjms_bigquery__subscriptions__v1",
+        destination_table="subscriptions_v1",
+        dataset_id="cjms_bigquery",
+        project_id="moz-fx-cjms-nonprod-9a36",
+        owner="dthorn@mozilla.com",
+        email=["dthorn@mozilla.com", "telemetry-alerts@mozilla.com"],
+        date_partition_parameter=None,
+        depends_on_past=False,
+        dag=dag,
+    )
+
+    cjms_bigquery__subscriptions__v1.set_upstream(cjms_bigquery__flows__v1)
--- a/sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/flows_v1/query.sql
+++ b/sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/flows_v1/query.sql
@@ -1,0 +1,54 @@
+WITH fxa_content_auth_stdout_events_live AS (
+  SELECT
+    PARSE_DATE('%y%m%d', _TABLE_SUFFIX) AS submission_date,
+    JSON_VALUE(jsonPayload.fields.user_properties, '$.flow_id') AS flow_id,
+    `timestamp`,
+    TO_HEX(SHA256(jsonPayload.fields.user_id)) AS fxa_uid,
+  FROM
+    `moz-fx-fxa-nonprod-375e.fxa_stage_logs.docker_fxa_auth_20*`
+  WHERE
+    jsonPayload.type = 'amplitudeEvent'
+    AND jsonPayload.fields.event_type IS NOT NULL
+    AND jsonPayload.fields.user_id IS NOT NULL
+  UNION ALL
+  SELECT
+    PARSE_DATE('%y%m%d', _TABLE_SUFFIX) AS submission_date,
+    JSON_VALUE(jsonPayload.fields.user_properties, '$.flow_id') AS flow_id,
+    `timestamp`,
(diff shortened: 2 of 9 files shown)


dags/bqetl_cjms_nonprod.py
```diff
@@ -0,0 +1,172 @@
...
+        project_id="moz-fx-cjms-nonprod-9a36",
+        owner="dthorn@mozilla.com",
+        email=["dthorn@mozilla.com", "telemetry-alerts@mozilla.com"],
+        date_partition_parameter=None,
+        depends_on_past=False,
+        dag=dag,
+    )
+
+    cjms_bigquery__subscription__v1.set_upstream(cjms_bigquery__flow__v1)
+    wait_for_stripe_derived__nonprod_customers__v1 = ExternalTaskCompletedSensor(
```
relud, line 82: these dependencies will disappear after #2807 is merged

sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/flows_v1/query.sql
```diff
@@ -0,0 +1,54 @@
...
+  submission_date,
+  flow_id,
+  MIN(`timestamp`) AS flow_started,
+  ARRAY_AGG(
+    IF(fxa_uid IS NULL, NULL, STRUCT(fxa_uid, `timestamp` AS fxa_uid_timestamp)) IGNORE NULLS
+    ORDER BY
+      `timestamp` DESC
+    LIMIT
+      1
+  )[SAFE_OFFSET(0)].*,
```
sean-rose, line 46: I think only taking the last FxA per flow might miss some edge cases.

For example, if someone subscribes to VPN, but then in the same session logs into a different FxA after that for some reason (e.g. shared computer) this would currently only get the FxA that didn't subscribe.

sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/subscriptions_v1/query.sql
```diff
@@ -0,0 +1,72 @@
...
+),
+attributed_flows AS (
+  -- last flow_id in aic for each fxa_uid
+  SELECT
+    fxa_uid,
+    ARRAY_AGG(STRUCT(flow_id, flow_started) ORDER BY flow_started DESC LIMIT 1)[SAFE_OFFSET(0)].*,
+  FROM
+    flows
+  JOIN
+    EXTERNAL_QUERY("moz-fx-cjms-nonprod-9a36.us.cjms-sql", "SELECT flow_id FROM aic")
```
sean-rose, line 25: The [architecture/flow diagram](https://www.figma.com/file/6jnLCLzclBN0uyS4nJp57d/Affiliate-Marketing-(CJ)-Architecture-%2F-Flow?node-id=0%3A1) seems to indicate some data may be removed from the CJMS database:
> Has the time frame for affiliate credit passed?
> Yes:  CJMS removes the row from the unique id table

Since we're rebuilding this whole table every time, are we sure that the `aic` table will retain all the historical data we need to do that accurately?

sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/subscriptions_v1/query.sql
```diff
@@ -0,0 +1,72 @@
...
+  USING
+    (flow_id)
+  GROUP BY
+    fxa_uid
+),
```
sean-rose, line 30: I think taking the last flow per FxA before knowing the subscription created cutoff might miss some edge cases.

For example, if someone subscribes to VPN and then later clicks a VPN affiliate link (or has a page with affiliate URL params reloaded/navigated in history), this would only include the later flow and the subscription won't be counted as attributed in the `attributed_subs` CTE below.

sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/subscriptions_v1/query.sql
```diff
@@ -0,0 +1,72 @@
...
+  FROM
+    attributed_flows
+  JOIN
+    mozdata.subscription_platform.nonprod_stripe_subscriptions
+  ON
+    attributed_flows.fxa_uid = nonprod_stripe_subscriptions.fxa_uid
+    AND attributed_flows.flow_started < nonprod_stripe_subscriptions.created
+  GROUP BY
+    fxa_uid
+)
```
sean-rose, line 51: - This seems backward to me, as I'd think you'd want the last attributed flow for the FxA before the subscription was created (i.e. subscription should be the base/grain/group-by of this CTE).
- There's no filter specific to VPN, so wouldn't this also pick up other FxA-related subscriptions like Relay premium?

sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/subscriptions_v1/query.sql
```diff
@@ -0,0 +1,72 @@
...
+  CURRENT_TIMESTAMP AS report_timestamp,
+  created AS subscription_created,
+  subscription_id, -- transaction id
+  fxa_uid,
+  1 AS quantity,
+  plan_id, -- sku
+  plan_currency,
+  plan_amount,
+  country,
+  flow_id,
```
sean-rose, line 62: When joining multiple tables my preference would be to qualify all columns with the table/CTE name so there's no ambiguity about where columns are coming from, and it guards against columns of the same name being added to tables later and breaking queries.

sean-rose (approved):
`r+wc`

I brought up a number of edge cases, but since this is non-prod I don't necessarily think they all need to be addressed immediately.

sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/flows_v1/query.sql
```diff
@@ -0,0 +1,54 @@
+WITH fxa_content_auth_stdout_events_live AS (
+  SELECT
+    PARSE_DATE('%y%m%d', _TABLE_SUFFIX) AS submission_date,
+    JSON_VALUE(jsonPayload.fields.user_properties, '$.flow_id') AS flow_id,
+    `timestamp`,
+    TO_HEX(SHA256(jsonPayload.fields.user_id)) AS fxa_uid,
```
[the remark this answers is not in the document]
relud: yep, stakeholders know

sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/flows_v1/query.sql
```diff
@@ -0,0 +1,54 @@
...
+    ORDER BY
+      `timestamp` DESC
+    LIMIT
+      1
+  )[SAFE_OFFSET(0)].*,
+FROM
+  fxa_content_auth_stdout_events_live
+WHERE
+  submission_date = @submission_date
+  AND flow_id IS NOT NULL
```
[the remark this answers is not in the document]
relud: we need events where `fxa_uid IS NULL` to correctly determine `flow_started` when aggregating across multiple days.

sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/refunds_v1/query.sql
```diff
@@ -0,0 +1,27 @@
+CREATE OR REPLACE TABLE
+  `moz-fx-cjms-nonprod-9a36`.cjms_bigquery.refund_v1
+AS
+SELECT
```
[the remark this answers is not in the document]
relud: yes

sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/flows_v1/query.sql
relud: I will pass that on, but this is the compromise I worked out with subplat eng for selecting a single fxa_uid per flow.

sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/subscriptions_v1/query.sql
relud: this table is not expected to contain historical results, only current results based on the `aic` table.

sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/subscriptions_v1/query.sql
relud: I will pass that on, but this is the compromise I worked out with subplat eng

relud: good point, i'll discuss this with sarah bird and file a follow up if necessary

sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/subscriptions_v1/query.sql
relud: picking up non-vpn subscriptions is intentional

sean-rose: But by only picking a single subscription, if someone subscribes to multiple services in one flow they won't all get attributed.

And if only a single subscription should be attributed per flow then I'd think the flow's first subscription is probably the one that deserves attribution the most (what kicked off their buying spree).

relud: yeah, i think i had a bug here, I reworked this and `attributed_flows`, the end goal is to have a 1:1 relationship between flow_id and subscription_id, and i got caught up on fxa_uid.

sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/subscriptions_v1/query.sql
relud: will rework this to take the last flow after joining subscription

sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/subscriptions_v1/query.sql
relud: > And if only a single subscription should be attributed per flow then I'd think the flow's first subscription is probably the one that deserves attribution the most (what kicked off their buying spree).

I will bring it up with subplat eng, but choosing one subscription per flow, and choosing the last subscription, are both intentional.

sean-rose (approved):

PR #2829, opened 2022-03-25
Added generator for country_names 

ANich:
Updated the source for `country_names_v1` to a yaml file and sql generator (that doesn't actually generate SQL 😝) to make it easier to verify and add new aliases (names) for country codes. 
I've also included the aliases from `sql/moz-fx-data-shared-prod/static/country_names_alternate.csv` so I removed it and the accompanying script.

I matched the (name, code) pairs from the existing table to the new table and found no misses on the existing table. See: `static.country_names_tmp_v1` (Although there was one mismatched code on the existing table: Bolivia -> both BO and CO).

Hoping to use this as the single lookup table across tables for unifying aliases to country codes. The schema is the same but the source is different, should this require a version bump on `country_names`?


jklukas (changes requested):
> Hoping to use this as the single lookup table across tables for unifying aliases to country codes. The schema is the same but the source is different, should this require a version bump on country_names?

As long as the schema and the uniqueness contracts are satisfied, no need for a version bump. IIUC, all that's happening here is that we get some additional rows with additional unique `name` values.

It's unclear to me right now how one is expected to deploy changes to this list. Now that the country names list is not held directly in the repository, is the expectation that deployment would involve a DE running the generator script locally before running the static dataset deployment script? If so, we should update the README in `static` to reflect this.

ANich:
> It's unclear to me right now how one is expected to deploy changes to this list. Now that the country names list is not held directly in the repository, is the expectation that deployment would involve a DE running the generator script locally before running the static dataset deployment script? If so, we should update the README in `static` to reflect this.

Yeah I think `bqetl generate country_code_lookup` ... `python publish_static.py` is the way to do this (the country_names `data.csv` is still version controlled here, it's just a large diff). I'll go ahead and update the README.

It would be nice to get this automated all the way through. CI does currently run all the generators under `sql_generator`'s via a `bqetl generate all` command with the output in the `generated-sql` folder: https://github.com/mozilla/bigquery-etl/blob/main/.circleci/config.yml#L341 but adding the `publish_static` deploy script to CI wouldn't work since no service account IIRC.. Open to ideas here

jklukas:
> Maybe an airflow task would be appropriate here

Yes, I think if we want to fully automate this, setting up Airflow to publish would be the way to go. It could be a new DAG with a single task that simply calls `publish_static.py`.

sql/moz-fx-data-shared-prod/static/README.md
```diff
@@ -16,3 +16,6 @@ ETL and analysis:
 ./script/publish_static --project-id mozdata
 ./script/publish_static --project-id moz-fx-data-shared-prod
```
jklukas, line 17:
```suggestion
# Generate data.csv for country_names
bqetl generate country_code_lookup

./script/publish_static --project-id mozdata
./script/publish_static --project-id moz-fx-data-shared-prod
```

sql/moz-fx-data-shared-prod/static/README.md
```diff
@@ -16,3 +16,6 @@ ETL and analysis:
...
 ```
+
+Note the `country_names` `data.csv` is generated/updated using an SQL generator:
+`bqetl generate country_code_lookup`.
```
jklukas, line 21:
```suggestion
```

I think we should move this into the bash snippet above to make it less likely to get skipped.

jklukas (approved):

PR #2834, opened 2022-03-28
Bug 1761553 - Fix Rally study name label prefix in Shredder

akkomar:
Previously Shredder was expecting study name labels on analysis datasets to have dashes converted to underscores, however this seems to not be the case in production.

Checklist for reviewer:

### DIFF
--- a/bigquery_etl/shredder/config.py
+++ b/bigquery_etl/shredder/config.py
@@ -510,7 +510,7 @@ def find_pioneer_targets(pool, client, project=PIONEER_PROD, study_projects=[]):
     def __get_client_id_field__(table, deletion_request_view=False, study_name=None):
         """Determine which column should be used as client id for a given table."""
         if table.dataset_id.startswith("rally_") or (
-            study_name and study_name.startswith("rally_")
+            study_name and study_name.startswith("rally-")
         ):
             # `rally_zero_one` is a special case where top-level rally_id is used
             # both in the ping tables and the deletion_requests view


whd (approved):
r+wc 

> Previously Shredder was expecting study name labels on analysis datasets to have dashes converted to underscores, however this seems to not be the case in production.

I went back and looked at https://bugzilla.mozilla.org/show_bug.cgi?id=1680173#c5 since we definitely had this logic working to shred analysis datasets at some point. See also https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-pioneer/tf/modules/study-project/variables.tf#L15 where an equivalence between namespace and study-name is documented.

So the study name was never normalized in the label, in fact I called out that some processing of it should probably explicitly normalize in https://github.com/mozilla/bigquery-etl/pull/1600#discussion_r537794381, but maybe that ended up sending a mixed signal. It looks like the change that "broke" it might have been part of https://github.com/mozilla/bigquery-etl/pull/2139/files#diff-210f8f1b0c091d54f715a99c6a4fdbc2dd3fba9bc3595fd6a0c7a5e1f01a529cR488.

Regardless this is probably the right fix, though I think https://github.com/mozilla/bigquery-etl/blob/725c5147c9372de0b1086c8468a02e8c83fe9e45/bigquery_etl/shredder/config.py#L519 also technically should be updated to reflect the non-normalized string (though in practice, there are likely no tables in the analysis dataset for this project).

akkomar:
Good points, looks like I missed this when addressing https://github.com/mozilla/bigquery-etl/pull/1600#discussion_r537794381.

PR #2841, opened 2022-03-29
Use JSON_VALUE in fxa events views

relud:
this showed no changes for `2022-03-28` when comparing forward and backward using `EXCEPT DISTINCT` for `mozdata.firefox_accounts.fxa_content_auth_stdout_events`

and there were no `oauth` events for `2022-03-28` or `2022-03-27`


jklukas (approved):
LGTM! Thanks for doing the data comparison.

PR #2843, opened 2022-03-30
Bug 1761790 Point to v1 schema in impression_stats view

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1761790

We'll be adding a new v2 schema that's not yet used, so explicitly adding the view definition here so we don't automatically deploy a view that only looks at v2.

As we build this out, we'll need to make a coordinated change that unions the two tables and change downstream tables to accept a string-type tile `id` field.

Checklist for reviewer:

### DIFF
--- a/sql/moz-fx-data-shared-prod/activity_stream/impression_stats/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/activity_stream/impression_stats/metadata.yaml
@@ -1,0 +1,12 @@
+---
+friendly_name: Historical Pings for `activity-stream/impression-stats`
+description: |-
+  A historical view of pings sent for the
+  `activity-stream/impression-stats`
+  document type.
+
+  This view is guaranteed to contain only complete days
+  (per `submission_timestamp`)
+  and to contain only one row per distinct `document_id` within a given date.
+
+  Clustering fields: `normalized_channel`, `sample_id`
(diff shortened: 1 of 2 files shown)


sql/moz-fx-data-shared-prod/activity_stream/impression_stats/metadata.yaml
```diff
@@ -0,0 +1,12 @@
...
+description: |-
+  A historical view of pings sent for the
+  `activity-stream/impression-stats`
+  document type.
+
+  This view is guaranteed to contain only complete days
+  (per `submission_timestamp`)
+  and to contain only one row per distinct `document_id` within a given date.
+
+  Clustering fields: `normalized_channel`, `sample_id`
```
jklukas, line 12: Currently, this view is created automatically as part of bigquery-etl CI; see output in https://github.com/mozilla/bigquery-etl/tree/generated-sql/sql/moz-fx-data-shared-prod/activity_stream/impression_stats

Once we introduce `v2`, the automated view logic would choose to have the view point at `v2` rather than `v1`, which could break existing use cases. We check these definitions directly into the `main` branch here so that this takes precedence over the view generation logic.

ncloudioj (approved):

scholtzan (approved):

PR #2846, opened 2022-03-30
Wait for fivetran when reading stripe data

relud:


dags/bqetl_subplat.py
```diff
@@ -925,6 +928,19 @@
     mozilla_vpn_derived__all_subscriptions__v1.set_upstream(
         mozilla_vpn_derived__users__v1
     )
+    fivetran_stripe_sync_start = FivetranOperator(
+        connector_id="{{ var.value.fivetran_stripe_connector_id }}",
+        task_id="fivetran_stripe_task",
+    )
```
relud, line 934: not sure if there's something I need to do here to indicate when the fivetran connector runs relative to this DAG

bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -7,7 +7,7 @@ from utils.gcp import bigquery_etl_query, gke_command
 
 {% set ns = namespace(uses_fivetran=False) -%}
 {% for task in tasks -%}
-{% if task.depends_on_fivetran != None and task.depends_on_fivetran|length > 1 and ns.uses_fivetran == False -%}
+{% if task.depends_on_fivetran != None and task.depends_on_fivetran|length > 0 and ns.uses_fivetran == False -%}
```
relud, line 10: pretty sure this is supposed to be `0`, because without it my dag is missing necessary imports

dags/bqetl_subplat.py
scholtzan: No, the Fivetran connector will get triggered by this task. If you have configured a scheduling interval in the Fivetran UI, then it will automatically get changed saying something like that the connector is manually triggered.

scholtzan (approved):

dags/bqetl_subplat.py
```diff
@@ -925,6 +928,19 @@
...
+
+    fivetran_stripe_sync_wait = FivetranSensor(
+        connector_id="{{ var.value.fivetran_stripe_connector_id }}",
```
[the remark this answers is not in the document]
relud: yep, already done

PR #2859, opened 2022-04-04
DVPN-136 - Fix copypasta from stage in cjms flows

relud:


sean-rose (approved):
Are there any particular reasons why this doesn't make use of `mozdata.firefox_accounts.fxa_content_auth_stdout_events`?

relud:
> Are there any particular reasons why this doesn't make use of `mozdata.firefox_accounts.fxa_content_auth_stdout_events`?

good point, updated to do that

PR #2870, opened 2022-04-06
fixed param name to submission_date

kik-kik:
fixed param name to submission_date,

the template uses `@submission_date` param, it looks like we tried to set the value for param `date` which does not exist:
https://github.com/mozilla/bigquery-etl/blob/b9a2f77eb7b5b17e9c86d6b96c90e6ea48bdd3a0/sql/moz-fx-data-shared-prod/internet_outages/global_outages_staging_v1/query.sql#L89

error message:
```[2022-04-06 08:14:10,938] {{pod_launcher.py:149}} INFO - 'submission_date' not found at [89:23]```

### DIFF
--- a/dags/bqetl_internet_outages.py
+++ b/dags/bqetl_internet_outages.py
@@ -64,7 +64,7 @@ with DAG(
         email=["aplacitelli@mozilla.com"],
         date_partition_parameter=None,
         depends_on_past=False,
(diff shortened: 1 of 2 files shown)


Dexterp37 (approved):
Thanks. This makes sense to me, but let's wait for @jklukas since made the initial change to use 'date'

jklukas:
Looks great!

PR #2873, opened 2022-04-06
DVPN-89 - Catch all flows at end of day for cjms nonprod

relud:
flows_v1 was not getting events for the previous UTC date that arrive after 00:22 UTC, this pushes back that cutoff to 02:00 UTC, and uses a live view to catch not-yet-aggregated flows from the current date.

### DIFF
--- a/sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/flows_v1/metadata.yaml
+++ b/sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/flows_v1/metadata.yaml
@@ -9,3 +9,19 @@ labels:
 scheduling:
   dag_name: bqetl_cjms_nonprod
   query_project: moz-fx-data-shared-prod
+  # delay aggregates by 2 hours, to ensure data is complete
+  date_partition_parameter: null
+  destination_table:
+    >-
+    flows_v1${{
+    (execution_date - macros.timedelta(hours=2)).strftime("%Y%m%d")
+    }}
+  parameters:
+    - >-
+      submission_date:DATE:{{
+      (execution_date - macros.timedelta(hours=2)).strftime("%F")
+      }}
+  query_file_path:
+    # explicit query file path is necessary because the destination table
+    # includes a partition identifier that is not in the path
+    sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/flows_v1/query.sql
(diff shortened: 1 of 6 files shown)


sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/flows_v1/metadata.yaml
```diff
@@ -9,3 +9,19 @@ labels:
...
+  date_partition_parameter: null
+  destination_table:
+    >-
+    flows_v1${{
+    (execution_date - macros.timedelta(hours=2)).strftime("%Y%m%d")
+    }}
+  parameters:
+    - >-
+      submission_date:DATE:{{
+      (execution_date - macros.timedelta(hours=2)).strftime("%F")
```
sean-rose, line 22: IMO `%Y-%m-%d` would be more self-explanatory (e.g. the [Python docs](https://docs.python.org/3.8/library/datetime.html#strftime-and-strptime-format-codes) don't cover the `%F` format code).

sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/subscriptions_v1/query.sql
```diff
@@ -1,4 +1,17 @@
...
+    *
+  FROM
+    `moz-fx-cjms-nonprod-9a36`.cjms_bigquery.flows_v1
+  UNION ALL
+  SELECT
+    *
+  FROM
+    `moz-fx-cjms-nonprod-9a36`.cjms_bigquery.flows_live
+  WHERE
+    submission_date = CURRENT_DATE
```
sean-rose, line 12: This will result in duplicates between 03:00 - midnight when `flows_v1` contains rows for the current date.  Either the `flows_v1` query should be filtered to before the current date, or the `flows_live` query should be filtered based on the max value from `flows_v1`.

sean-rose (approved):
`r+wc`

sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/flows_v1/metadata.yaml
relud:
```suggestion
      (execution_date - macros.timedelta(hours=2)).strftime("%Y-%m-%d")
```

sql/moz-fx-cjms-nonprod-9a36/cjms_bigquery/subscriptions_v1/query.sql
relud:
```suggestion
  SELECT
    *
  FROM
    `moz-fx-cjms-nonprod-9a36`.cjms_bigquery.flows_v1
  WHERE
    submission_date < CURRENT_DATE
  UNION ALL
  SELECT
    *
  FROM
    `moz-fx-cjms-nonprod-9a36`.cjms_bigquery.flows_live
  WHERE
    submission_date = CURRENT_DATE
```

PR #2877, opened 2022-04-06
Added InvalidDag when generating a DAG with no tasks

ANich:
If you add a DAG to `dags.yaml` that doesn't appear in any metadata files it won't end up having any tasks. When trying to generate the dag python file it returns invalid python, something like:

```
...
with DAG(
  ...
) as dag:

<EOF>
```
which makes DAG generation fail at formatting the generated python file. The error raised from `Black` is pretty esoteric, so PR adds a clearer upstream error.

In future it might be worth changing this behaviour to support DAGs that don't appear in metadata files but are still bigquery-etl related, maybe with a `DummyOperator` or something. (For context this is why I put https://github.com/mozilla/telemetry-airflow/pull/1481 in telemetry-airflow)

Also updated an unrelated f-string.

### DIFF
--- a/bigquery_etl/query_scheduling/dag.py
+++ b/bigquery_etl/query_scheduling/dag.py
@@ -214,6 +214,11 @@ class Dag:
         dag_template = env.get_template(AIRFLOW_DAG_TEMPLATE)
 
         args = self.__dict__
+        if len(args["tasks"]) == 0:
+            raise InvalidDag(
+                f"DAG {self.name} has no tasks - cannot convert it to a valid .py DAG "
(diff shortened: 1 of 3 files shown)


scholtzan (approved):

ANich:
> Another potential route here would be to inject a single `pass` statement into the DAG and let it exist with no tasks. I don't have a strong opinion about whether that would be preferred over raising an error. If we create a DAG with no tasks, I suppose it is very likely that it was done unintentionally, so an error may indeed be best.

Yeah, will think on this. I suspect the answer is related to which DAGs we want to live in this repo and which ones we want in telemetry-airflow.

PR #2885, opened 2022-04-11
Work around VPN funnel issues (DVPN-94, DVPN-122)

sean-rose:
This should resolve:
- [DVPN-122](https://mozilla-hub.atlassian.net/browse/DVPN-122) "_the vpn product site visit counting from dashboard agg table is not matching with the fxa table_".
- [DVPN-94](https://mozilla-hub.atlassian.net/browse/DVPN-94) "_VPN attribution is missing some events_" from 2021-12-09 onward.

Some things worth noting:
- I haven't been able to discover the original cause of the issue from 2021-11-05 to 2022-01-06 where flows with VPN FxA button views (top of funnel) no longer had any events with `service = 'guardian-vpn'` ([I've asked for help in Slack](https://mozilla.slack.com/archives/C018SNLQP6C/p1649702727286539)).
- `fxa_attribution_v1` now considers _all_ events in VPN-related flows, rather than just the VPN-specific events.  This is what resolves [DVPN-94](https://mozilla-hub.atlassian.net/browse/DVPN-94), but if flows have both VPN-specific and non-VPN attributions then the non-VPN attributions could now take precedence if they occurred first.
- `funnel_product_page_to_subscribed_v1` will now be limited to >= 2021-08-25, because that's when the fully FxA-based funnel for VPN was enabled (the existing table goes back as far as 2019-10-10).
- I materialized the modified `funnel_product_page_to_subscribed_v1` ETL in the table `mozdata.tmp.srose_vpn_funnel_product_page_to_subscribed_v1` for QA/comparison purposes.

---

Checklist for reviewer:


sql/moz-fx-data-shared-prod/mozilla_vpn_derived/fxa_attribution_v1/query.sql
```diff
@@ -51,18 +51,41 @@ flows AS (
...
+      OR (
+        event_type = "fxa_rp_button - view"
+        AND (
+          -- The www.mozilla.org navbar CTA button was changed to link to VPN for Firefox users
+          -- on 2021-09-02, then attribution was implemented for it on 2021-09-15.
+          (
+            service IS NULL
+            AND utm_source = "www.mozilla.org"
+            AND utm_campaign = "navigation"
+            AND (partition_date BETWEEN "2021-09-15" AND "2021-12-08")
```
sean-rose, line 71: Previously all `fxa_rp_button - view` events with null `service` were being included, but the the vast majority of such events are from the "Join Firefox" button on the Firefox homepage and various links on the Firefox "What's New" pages, so this is significantly more targeted to the VPN CTA button.

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/fxa_attribution_v1/query.sql
```diff
@@ -51,18 +51,41 @@ flows AS (
...
+        )
+      )
+      -- Even when service attribution for VPN FxA links was fixed on 2022-01-06 there was still a
+      -- problem with service attribution for FxA payment events, which was fixed on 2022-03-09.
       OR (
-        (service IS NULL OR service = "undefined_oauth")
-        AND (event_type = "fxa_rp_button - view" OR event_type LIKE "fxa_pay_%")
+        event_type LIKE r"fxa\_pay\_%"
+        AND service = "undefined_oauth"
+        AND (partition_date BETWEEN "2021-12-08" AND "2022-03-09")
```
sean-rose, line 86: This no longer includes `fxa_pay_%` events with null `service`.  My thinking is that:
- Going forward VPN-related payment events should have `service = 'guardian-vpn'` (thanks to mozilla/fxa#11973).
- Prior to that, VPN payment events with null `service` should be still considered because we include all events in matching VPN-related flows, and we're now doing a better job of matching events earlier in the flow.

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/fxa_attribution_v1/query.sql
```diff
@@ -51,18 +51,41 @@ flows AS (
     fxa_content_auth_stdout_events
   WHERE
     IF(@date IS NULL, partition_date < CURRENT_DATE, partition_date = @date)
```
sean-rose, line 53: I presumed that this shouldn't be limited to >= 2021-08-25, because that would exclude the first year from when VPN launched with wave 1 in July 2020.

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/funnel_product_page_to_subscribed_v1/query.sql
```diff
@@ -107,15 +113,51 @@ flows AS (
...
+      OR (
+        event_type = "fxa_rp_button - view"
+        AND (
+          -- The www.mozilla.org navbar CTA button was changed to link to VPN for Firefox users
+          -- on 2021-09-02, then attribution was implemented for it on 2021-09-15.
+          (
+            service IS NULL
+            AND events.utm_source = "www.mozilla.org"
+            AND events.utm_campaign = "navigation"
+            AND (partition_date BETWEEN "2021-09-15" AND "2021-12-08")
```
sean-rose, line 143: I put parentheses around `BETWEEN` expressions to prevent `bqetl format` from formatting them confusingly.

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/funnel_product_page_to_subscribed_v1/query.sql
```diff
@@ -107,15 +113,51 @@ flows AS (
...
     LOGICAL_OR(event_type = "fxa_subscribe_coupon - success") AS subscribe_coupon_success,
   FROM
-    mozdata.firefox_accounts.fxa_content_auth_stdout_events
+    events
   WHERE
-    IF(@date IS NULL, DATE(`timestamp`) < CURRENT_DATE, DATE(`timestamp`) = @date)
+    IF(
+      @date IS NULL,
+      -- The fully FxA-based funnel for VPN was enabled on 2021-08-25.
+      (partition_date >= "2021-08-25" AND partition_date < CURRENT_DATE),
```
sean-rose, line 121: @YeonjooSmith I want to double check that limiting this to >= 2021-08-25 won't cause any issues for how this is being used in reports.

(we could allow previous data, but the top-of-funnel `vpn_site_hits` calculation only works from when the fully FxA-based funnel was enabled on 2021-08-25)

relud (approved):

sean-rose:
I only just realized `oauth_client_id` has meaningful values when `service = "undefined_oauth"`. 🤦 

I've used that to simplify things, and am asking for a re-review since it's a non-trivial change.

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/funnel_product_page_to_subscribed_v1/query.sql
YeonjooSmith: Yes the limiting dates to >=>= 2021-08-25 will be okay. Thank you for confirming!

YeonjooSmith (approved):

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/fxa_attribution_v1/query.sql
```diff
@@ -51,18 +51,41 @@ flows AS (
...
-    -- cannot filter service because
-    AND (
+    AND flow_id IS NOT NULL
+  GROUP BY
+    flow_id
+  HAVING
+    LOGICAL_OR(
```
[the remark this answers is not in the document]
sean-rose: > If I am not mistaken, would the multi touch approach resolve non-VPN attribution being selected?

This is still doing first touch attribution, so no change in that respect.

However, after further testing I think my previous assertion that this "_definitely broadens the pool of potential attribution parameters and may result in non-VPN attribution parameters being selected here if they came first_" is incorrect, and these changes will actually result in _fewer_ non-VPN attributions being selected:
- Flows started from www.mozilla.org FxA button views aren't persistent across www.mozilla.org page views like I originally thought.  I had assumed once a flow ID was assigned it would do something like set a cookie so that the same flow ID would be reused across www.mozilla.org page views within a limited time period, but it appears that each www.mozilla.org page view with FxA links generates a new flow ID for the FxA links on that page, and more events with that flow ID only occur if they click the FxA link.
- Previous logic was including all `fxa_rp_button - view` events with null `service`, but the vast majority of those events were for the "Join Firefox" button on the Firefox homepage and various FxA product links on the Firefox "What's New" pages.  The new logic won't include those non-VPN flows.

sean-rose:
@relud for `fxa_attribution_v1` I'm thinking we should rebuild it at least back to 2021-08-25, which is probably easiest to do with a manual query with a modified `flows` CTE like this:
```sql
flows AS (
  SELECT
    ...
  FROM
    fxa_content_auth_stdout_events
  WHERE
    partition_date BETWEEN "2021-08-25" AND (CURRENT_DATE - 1)
  ...
  UNION ALL
  SELECT
    *
  FROM
    fxa_attribution_v1
  WHERE
    DATE(flow_started) < "2021-08-25"
)
```
Or another option would be to rebuild the whole thing.

What do you think?

relud:
we should also preserve existing values where `flow_started IS NULL`, which were backfilled from guardian attribution data with `attribution.timestamp < "2021-06-01"`

sean-rose:
> we should also preserve existing values where `flow_started IS NULL`, which were backfilled from guardian attribution data with `attribution.timestamp < "2021-06-01"`

Great catch, thank you! 😅 

So that'd be like
```sql
  ...
  UNION ALL
  SELECT
    *
  FROM
    fxa_attribution_v1
  WHERE
    DATE(flow_started) < "2021-08-25"
    OR flow_started IS NULL
```

Do you think it'd be worth rebuilding further back than 2021-08-25, to 2021-06-01 or all-time? (preserving the Guardian attribution backfill data in any case)

relud:
probably better to only backfill to 2021-08-25

sean-rose:
Another round of changes:
- Fix for [issues with `funnel_product_page_to_subscribed_v1` that Yeonjoo found in QA](https://mozilla-hub.atlassian.net/browse/DVPN-122?focusedCommentId=536252).
- Rather than continuing to complicate `fxa_attributions_v1` with workarounds for the period before service attribution was implemented for VPN FxA links on 2021-12-08, since it doesn't need to be as precise about the `fxa_rp_button - view` events that get included I've opted to cut bait and only use the new attribution event filter logic after 2021-12-08, and thus only rebuild it back to 2021-12-09.
- I didn't like the idea of leaving the ETL in state where if someone rebuilt `fxa_attributions_v1` in the future it would change records we wanted left as is, so I've re-added the old logic with conditions to only apply it before the cutoff.

Hopefully this will be the last version/review needed.

relud (approved):

PR #2887, opened 2022-04-12
DENG-96 - Create browser_version_info UDF

ANich:
[DENG-96](https://mozilla-hub.atlassian.net/browse/DENG-96) - Creates a `browser_version_info` UDF that adds version metadata to a struct. This is primarily for views and is mimicking an external dataset that we could use for a version source of truth. This should show up in Looker as a dimension group. 

RE: the temporary part, I initially wanted to extend `telemetry.releases` but the API it's built on is missing release data for iOS and I'm not 100% sure whether it's maintained currently. So I think this is the next best thing for now.

### DIFF
--- a/sql/mozfun/norm/browser_version_info/udf.sql
+++ b/sql/mozfun/norm/browser_version_info/udf.sql
@@ -1,0 +1,76 @@
+CREATE OR REPLACE FUNCTION norm.browser_version_info(version_string STRING)
+RETURNS STRUCT<
+  version STRING,
+  major_version NUMERIC,
+  minor_version NUMERIC,
+  is_major_release BOOLEAN
+> AS (
+  STRUCT(
+    version_string AS version,
+    norm.truncate_version(version_string, 'major') AS major_version,
+    norm.truncate_version(version_string, 'minor') AS minor_version,
+    ARRAY_LENGTH(SPLIT(version_string, '.')) = 2
+    AND ENDS_WITH(version_string, '.0') AS is_major_release
+  )
+);
+
+-- Tests
+WITH browser_info AS (
+  SELECT AS VALUE
+    norm.browser_version_info('45.9.0')
+)
+SELECT
+  assert.equals('45.9.0', browser_info.version),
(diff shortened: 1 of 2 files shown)


sql/mozfun/norm/browser_version_info/udf.sql
```diff
@@ -0,0 +1,76 @@
+CREATE OR REPLACE FUNCTION norm.browser_version_info(version_string STRING)
+RETURNS STRUCT<
+  version STRING,
+  major_version NUMERIC,
+  minor_version NUMERIC,
+  is_major_release BOOLEAN
+> AS (
+  STRUCT(
+    version_string AS version,
+    norm.truncate_version(version_string, 'major') AS major_version,
```
kik-kik, line 10: I'd say `_version` could be omitted from the field name since it's already part of the `version` strict. `version.major` vs `version.major_version` doesn't add much in terms of readability IMHO.

sql/mozfun/norm/browser_version_info/udf.sql
```diff
@@ -0,0 +1,76 @@
...
+    norm.truncate_version(version_string, 'minor') AS minor_version,
+    ARRAY_LENGTH(SPLIT(version_string, '.')) = 2
+    AND ENDS_WITH(version_string, '.0') AS is_major_release
+  )
+);
+
+-- Tests
```
kik-kik, line 17: I actually love this. Are these queries / "tests" executed as part of CI in any way prior to deployment?

kik-kik (approved):

kik-kik:
general question, do all our apps follow the same versioning structure? If yes, maybe we could just make this a generic UDF for converting any app version string into a struct like this? @ANich

ANich:
> general question, do all our apps follow the same versioning structure? If yes, maybe we could just make this a generic UDF for converting any app version string into a struct like this? @ANich

The browsers do for sure, I suspect the SAAS products do not but version analytics is probably less of a concern there too (more rapid deployment and all that).

PR #2900, opened 2022-04-19
Feat/deng 17 acoustic data in bq

kik-kik:
Feat/[DENG-17](https://mozilla-hub.atlassian.net/browse/DENG-17) acoustic data in bq

JIRA: https://mozilla-hub.atlassian.net/browse/DENG-100

Adding ETL process for processing and cleaning Acoustic data to be consumed by Looker.

This DAG is related to another DAG: https://github.com/mozilla/telemetry-airflow/blob/main/dags/fivetran_acoustic.py

### DIFF
--- a/dags/bqetl_acoustic_raw_recipient_export.py
+++ b/dags/bqetl_acoustic_raw_recipient_export.py
@@ -1,0 +1,83 @@
+# Generated via https://github.com/mozilla/bigquery-etl/blob/main/bigquery_etl/query_scheduling/generate_airflow_dags.py
+
+from airflow import DAG
+from operators.task_sensor import ExternalTaskCompletedSensor
+import datetime
+from utils.gcp import bigquery_etl_query, gke_command
+
+docs = """
+### bqetl_acoustic_raw_recipient_export
+
+Built from bigquery-etl repo, [`dags/bqetl_acoustic_raw_recipient_export.py`](https://github.com/mozilla/bigquery-etl/blob/main/dags/bqetl_acoustic_raw_recipient_export.py)
+
+#### Description
+
+Processing data loaded by
+fivetran_acoustic_raw_recipient_export
+DAG to clean up the data loaded from Acoustic.
+
+#### Owner
+
+kignasiak@mozilla.com
+"""
+
+
+default_args = {
+    "owner": "kignasiak@mozilla.com",
+    "start_date": datetime.datetime(2022, 3, 1, 0, 0),
+    "end_date": None,
+    "email": ["telemetry-alerts@mozilla.com", "kignasiak@mozilla.com"],
+    "depends_on_past": False,
+    "retry_delay": datetime.timedelta(seconds=300),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
+}
+
+tags = ["impact/tier_3", "repo/bigquery-etl"]
+
+with DAG(
+    "bqetl_acoustic_raw_recipient_export",
+    default_args=default_args,
+    schedule_interval="0 8 * * *",
+    doc_md=docs,
+    tags=tags,
+) as dag:
+
+    acoustic__raw_recipient__v1 = bigquery_etl_query(
+        task_id="acoustic__raw_recipient__v1",
+        destination_table="raw_recipient_v1",
+        dataset_id="acoustic",
+        project_id="moz-fx-data-marketing-prod",
+        owner="kignasiak@mozilla.com",
+        email=["kignasiak@mozilla.com", "telemetry-alerts@mozilla.com"],
+        date_partition_parameter="submission_date",
+        depends_on_past=False,
+        arguments=["--submission_date", "{{ ds }}"],
+    )
+
+    acoustic__raw_recipient_raw__v1 = bigquery_etl_query(
+        task_id="acoustic__raw_recipient_raw__v1",
+        destination_table="raw_recipient_raw_v1",
+        dataset_id="acoustic",
+        project_id="moz-fx-data-marketing-prod",
+        owner="kignasiak@mozilla.com",
+        email=["kignasiak@mozilla.com", "telemetry-alerts@mozilla.com"],
+        date_partition_parameter="submission_date",
+        depends_on_past=False,
+        arguments=["--submission_date", "{{ ds }}"],
+    )
+
+    acoustic__raw_recipient__v1.set_upstream(acoustic__raw_recipient_raw__v1)
+
+    wait_for_fivetran_load_completed = ExternalTaskCompletedSensor(
+        task_id="wait_for_fivetran_load_completed",
+        external_dag_id="fivetran_acoustic_raw_recipient_export",
+        external_task_id="fivetran_load_completed",
+        execution_delta=datetime.timedelta(seconds=3600),
+        check_existence=True,
+        mode="reschedule",
+        pool="DATA_ENG_EXTERNALTASKSENSOR",
+    )
+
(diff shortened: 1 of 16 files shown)


kik-kik:
I will remove all the comments before merging.

sql/moz-fx-data-marketing-prod/acoustic/contact_v1/query.sql
```diff
@@ -0,0 +1,155 @@
...
+  )
+
+  , YESTERDAY_LOADED_CONTACTS AS (
+      SELECT * EXCEPT(job_date, processed_at)
+      FROM `moz-fx-data-marketing-prod.acoustic.contact_v1`  -- possible that this may not exist on first run
+      WHERE job_date = DATE_SUB(@submission_date, INTERVAL 1 DAY)
+  )
+
+
+  SELECT * FROM JUST_LOADED_CONTACTS
```
kik-kik, line 145: @ANich what do you think about an approach like this? so that over time we just collect the "current" state of all customers as of today. This should enable us to do nice aggregation of number of subscribers to each product over time?

sql/moz-fx-data-marketing-prod/acoustic/contact_v1/query.sql
```diff
@@ -0,0 +1,155 @@
...
+  UNION ALL
+  SELECT
+    *,
+    @submission_date AS job_date,
+    CURRENT_DATETIME() AS processed_at,
+  FROM YESTERDAY_LOADED_CONTACTS
+  WHERE email_id NOT IN (SELECT DISTINCT email_id FROM JUST_LOADED_CONTACTS);
+
+  TRUNCATE TABLE `moz-fx-data-bq-fivetran.acoustic_sftp.contact_export`;
```
ANich, line 154: Maybe add a comment here saying why it's necessary to truncate the source table after running the ETL? A link is probably fine

sql/moz-fx-data-marketing-prod/acoustic/contact_v1/query.sql
ANich: Is the final table supposed to be one row per contact per day? I'm not clear on how the processing/job date affects the granularity/partitioning. My intuition is that it shouldn't unless we have to but I don't have all the context. -- Will it be possible to recreate the table from scratch if necessary with this approach?

sql/moz-fx-data-marketing-prod/acoustic/contact_v1/query.sql
kik-kik: Added a comment to provide an explanation. Please let me know if it provides enough context.

sql/moz-fx-data-marketing-prod/acoustic/contact_v1/query.sql
kik-kik: okay, as per our discussion I updated the logic to just capture changes as we get them from Acoustic. We can always create another process on top of this "changelog" to generate the current version of contacts dataset.

sql/moz-fx-data-marketing-prod/acoustic/contact_raw_v1/query.sql
```diff
@@ -0,0 +1,76 @@
...
+    RECIPIENT_ID AS recipient_id,
+    DATE(@submission_date) AS submission_date,
+  FROM
+    `moz-fx-data-bq-fivetran.acoustic_sftp.contact_export`
+  -- The table created by Fivetran is not partitioned, and the data is not always great in quality
+  -- the data loaded into the Fivetran table is cleaned up and then loaded into a downstream table.
+  -- Once the data is cleaned and loaded in to the target table (partioning and cluster fields enabled)
+  -- this temporary/staging table (data) is no longer needed.
+    TRUNCATE TABLE `moz-fx-data-bq-fivetran.acoustic_sftp.contact_export`;
+END;
```
ANich, line 76: Discussed this a bit offline. I'm still unsure about the pattern of truncating the source though.. A couple of  questions/thoughts:
1. As it is now rebuilding this table means re-running the connector from the beginning of time instead of just the queries. It would be nice to break this dependency.. 
2. How does this interact with Fivetran billing? If I understood correctly from your description the connector only returns changed entries so the net Monthly Active Rows is the same?
3. Is it possible to add partitioning to these tables somehow? If not in the [fivetran native way](https://fivetran.com/docs/destinations/bigquery/partition-table), potentially through an intermediate append-only table?

sql/moz-fx-data-marketing-prod/acoustic/contact_raw_v1/metadata.yaml
```diff
@@ -0,0 +1,40 @@
+friendly_name: Contact Database (Acoustic data)
```
ANich, line 1: Unrelated to file but I believe you'll need a `dataset_metadata.yaml` in the `acoustic` dir

sql/moz-fx-data-marketing-prod/acoustic/contact_raw_v1/query.sql
```diff
@@ -0,0 +1,76 @@
...
+    CAST(
+      COALESCE(CAST(NULLIF(sub_firefox_sweepstakes, '') AS INTEGER), 0) AS BOOLEAN
+    ) AS sub_firefox_sweepstakes,
+    LOWER(NULLIF(vpn_waitlist_geo, '')) AS vpn_waitlist_geo,
+    SPLIT(NULLIF(vpn_waitlist_platform, ''), ",") AS vpn_waitlist_platform,
+    LOWER(NULLIF(relay_waitlist_geo, '')),
+    RECIPIENT_ID AS recipient_id,
+    DATE(@submission_date) AS submission_date,
+  FROM
+    `moz-fx-data-bq-fivetran.acoustic_sftp.contact_export`
```
ANich, line 70:
```suggestion
    `moz-fx-data-bq-fivetran.acoustic_sftp.contact_export`;
```

sql/moz-fx-data-marketing-prod/acoustic/contact_raw_v1/query.sql
kik-kik: okay, I modified the workflow a bit. Now we create a raw table that is partitioned by submission date which then is used when creating a downstream table with the data being processed.

dags/bqetl_acoustic_raw_recipient_export.py
```diff
@@ -0,0 +1,83 @@
...
+        task_id="wait_for_fivetran_load_completed",
+        external_dag_id="fivetran_acoustic_raw_recipient_export",
+        external_task_id="fivetran_load_completed",
+        execution_delta=datetime.timedelta(seconds=3600),
+        check_existence=True,
+        mode="reschedule",
+        pool="DATA_ENG_EXTERNALTASKSENSOR",
+    )
+
+    acoustic__raw_recipient__v1.set_upstream(wait_for_fivetran_load_completed)
```
ANich, line 81: Should this be:
```suggestion
    acoustic__raw_recipient__v1.set_upstream(acoustic__raw_recipient_raw__v1)
```
?

kik-kik: no, this seems correct. `wait_for_fivetran_load_completed` is a sensor task waiting for the task inside `fivetran_acoustic_raw_recipient_export` DAG to complete. Once it completes then the first BQETL process `acoustic__raw_recipient__v1` kicks off for loading the raw data into another raw table.

kik-kik: Fixed it, thanks a lot and well spotted!

PR #2913, opened 2022-04-25
Added extract_version UDF, modified browser_version_info UDF

ANich:
Added a new UDF to extract (rather than truncate) numerical versions - This allows analysis by patch numbers as well, given that truncating to the patch level doesn't support valid numbers. 

Also updated the `browser_version_info` UDF to use the new UDF. 

Is there any reason we shouldn't represent versions like this?

### DIFF
--- a/sql/mozfun/norm/browser_version_info/udf.sql
+++ b/sql/mozfun/norm/browser_version_info/udf.sql
@@ -3,14 +3,18 @@ RETURNS STRUCT<
   version STRING,
   major_version NUMERIC,
   minor_version NUMERIC,
+  patch_revision NUMERIC,
   is_major_release BOOLEAN
 > AS (
   STRUCT(
     version_string AS version,
     norm.truncate_version(version_string, 'major') AS major_version,
     norm.truncate_version(version_string, 'minor') AS minor_version,
-    ARRAY_LENGTH(SPLIT(version_string, '.')) = 2
-    AND ENDS_WITH(version_string, '.0') AS is_major_release
+    norm.extract_version(version_string, 'patch') AS patch_revision,
+    (
+      ARRAY_LENGTH(SPLIT(version_string, '.')) = 2
+      AND ENDS_WITH(version_string, '.0')
+    ) AS is_major_release
   )
 );
 
@@ -23,6 +27,7 @@ SELECT
   assert.equals('45.9.0', browser_info.version),
   assert.equals(45, browser_info.major_version),
   assert.equals(45.9, browser_info.minor_version),
+  assert.equals(0, browser_info.patch_revision),
   assert.false(browser_info.is_major_release)
 FROM
   browser_info;
@@ -35,6 +40,7 @@ SELECT
   assert.equals('96.0', browser_info.version),
   assert.equals(96, browser_info.major_version),
   assert.equals(96, browser_info.minor_version),
+  assert.null(browser_info.patch_revision),
   assert.true(browser_info.is_major_release)
 FROM
   browser_info;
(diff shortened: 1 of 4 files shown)


sql/mozfun/norm/browser_version_info/udf.sql
```diff
@@ -3,14 +3,18 @@ RETURNS STRUCT<
...
 > AS (
   STRUCT(
     version_string AS version,
-    norm.truncate_version(version_string, 'major') AS major_version,
-    norm.truncate_version(version_string, 'minor') AS minor_version,
-    ARRAY_LENGTH(SPLIT(version_string, '.')) = 2
-    AND ENDS_WITH(version_string, '.0') AS is_major_release
+    norm.extract_version(version_string, 'major') AS major_version,
+    norm.extract_version(version_string, 'minor') AS minor_version,
+    norm.extract_version(version_string, 'patch') AS patch_version,
```
scholtzan, line 13: I checked where this is used. Looks like this is only used in views so adding `patch_version` should not cause schemas to break

scholtzan (approved):

kik-kik (approved):

sql/mozfun/norm/browser_version_info/udf.sql
```diff
@@ -22,7 +26,8 @@ WITH browser_info AS (
 SELECT
   assert.equals('45.9.0', browser_info.version),
   assert.equals(45, browser_info.major_version),
-  assert.equals(45.9, browser_info.minor_version),
+  assert.equals(.9, browser_info.minor_version),
```
sean-rose, line 29: This is a significant change for how `minor_version` works and seems likely to affect downstream dashboards/analyses.

sean-rose: I might suggest:
- Keep `minor_version` using `truncate_version()`, and add a separate field with the new behavior (e.g. `minor_revision`).
- Have `patch_version` be a string in `<major>.<minor>.<patch>` format for consistency with `major_version` and `minor_version`, and add a separate field with just the new numeric `.<patch>` value (e.g. `patch_revision`).

sql/mozfun/norm/extract_version/udf.sql
```diff
@@ -0,0 +1,32 @@
...
+  THEN
+    CAST(REGEXP_EXTRACT(version_string, r"^[0-9]+[.][0-9]+([.][0-9]+).*") AS NUMERIC)
+  WHEN
+    extraction_level = "minor"
+  THEN
+    CAST(REGEXP_EXTRACT(version_string, r"^[0-9]+([.][0-9]+).*") AS NUMERIC)
+  WHEN
+    extraction_level = "major"
+  THEN
+    CAST(REGEXP_EXTRACT(version_string, r"^([0-9]+).*") AS NUMERIC)
```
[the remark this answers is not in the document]
ANich: For sure, it doesn't always have to be but the goal is to extract the numeric part(s) but it's a good point and why in `browser_version_info` we retain the original string

sql/mozfun/norm/browser_version_info/udf.sql
sean-rose: Or I see from another comment that `browser_version_info` was only added recently, so maybe not much will break?

I do like the consistency of the new behavior, having all the parts numeric.

ANich: Yeah some good points here, I initially thought this was worth the breaking change with some due diligence. I'll go ahead and call the patch: `patch_revision` and remove `patch_version` so no-one expects it to be a full version number.  --  If we do that we won't have to introduce a `patch_version` field at all (`.version` already contains the original version string)

sean-rose: That sounds reasonable, though `patch_version` might still be useful in certain scenarios, as the full `.version` string might contain more than just the numeric parts `<major>.<minor>.<patch>` (e.g. `96.0.2b1`).

Issue #1854, opened 2021-02-26
Dependabot broken due to pyjnius


relud:
pyjnius can't be imported if `which javac` fails, and dependabot tries to import pyjnius when checking for updates.

relud:
fixed by #1990

PR #2924, opened 2022-04-29
Specify pyjnius in `requirements.in`

sean-rose:
pynjius originally had to be split out into a separate `java-requirements.in` file because it ran `which javac` when imported, which failed when dependabot tried to import pyjnius when checking for updates (#1854).

Now apparently pynjius no longer does `which javac` so it can be put back in `requirements.in`, simplifying things.

In particular, this will allow the `bqetl dependency record` command in Netlify CI builds to succeed without having to add an extra step to install packages from `java-requirements.txt`.

---

Checklist for reviewer:


requirements.txt
```diff
@@ -952,9 +984,10 @@ six==1.15.0 \
     #   oauth2client
     #   ppft
     #   protobuf
+    #   pyjnius
     #   python-dateutil
     #   virtualenv
-smart-open==6.0.0 \
+smart_open==6.0.0 \
```
sean-rose, line 990: This is weird.  Running pip-compile locally using Python 3.8 changed this to `smart_open`, which seems correct to me, but the [CircleCI verify-requirments check](https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/14333/workflows/bef41822-d4ab-4bb5-880b-ba071acb6567/jobs/116973) is insisting this should be `smart-open`.

setup.py
```diff
@@ -56,7 +56,6 @@ def get_version():
         "stripe",
         "authlib",
     ],
-    extras_require={"java": ["pyjnius"]},
```
sean-rose, line 59: Not entirely sure about removing this, but since it was added in #1990 I'm defaulting to taking it out.

sean-rose:
[generate-diff](https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/14333/workflows/bef41822-d4ab-4bb5-880b-ba071acb6567/jobs/116986) is failing because it runs commands against the `main` branch, which doesn't yet have pyjnius in `requirements.txt`, so it doesn't get installed and the `jnius` module can't be found.

requirements.txt
relud: might be related to osx vs linux, but we should stick with `smart-open` if that's what CI gives us.

relud (approved):
r+wc

requirements.txt
sean-rose: Never mind, looks like I had pip-tools 6.4.0 installed.  Upgrading to pip-tools 6.6.0 and running pip-compile locally changes this to `smart-open` like the CI.

sean-rose:
[integration test](https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/14338/workflows/8ed6c362-2449-4dc4-aee1-0b50a3de36fe/jobs/117024) failed with the random, always fun `File /tmp/gcp.json is not a valid json file` error.  Ignoring.

sean-rose:
I've verified that the Netlify CI build can now run `bqetl dependency record` successfully without getting `ModuleNotFoundError: No module named 'jnius'`.

PR #2939, opened 2022-05-04
made some additional tweaks to solve errors when attempting to run bqetl for acoustic

kik-kik:
made some additional tweaks to solve errors when attempting to run bqetl for acoustic.

Primarly fixed source table and added `init.sql`


sql/moz-fx-data-marketing-prod/acoustic/raw_recipient_raw_v1/query.sql
```diff
@@ -2,4 +2,6 @@ SELECT
   * EXCEPT (_file, _line, _fivetran_synced, _modified),
   DATE(@submission_date) AS submission_date,
 FROM
-  `moz-fx-data-bq-fivetran.acoustic_sftp.raw_recipient_export_v1`
+  `moz-fx-data-bq-fivetran.acoustic_sftp.raw_recipient_export_v_1`
+WHERE
+  DATE(@submission_date) = PARSE_DATETIME("%m/%d/%Y %H:%M:%S", event_timestamp)
```
jklukas, line 7:
```suggestion
  DATE(@submission_date) = PARSE_DATETIME("%m/%d/%Y %H:%M:%S", event_timestamp)
```

It looks like we're comparing a DATE to a DATETIME here. Is that what we want? Can we wrap the datetime with `DATE` to make this more explicit?

sql/moz-fx-data-marketing-prod/acoustic/raw_recipient_v1/query.sql
```diff
@@ -13,6 +13,6 @@ SELECT
   LOWER(COALESCE(NULLIF(suppression_reason, ''), "unknown")) AS suppression_reason,
   submission_date,
 FROM
-  `moz-fx-data-marketing-prod.acoustic.raw_recipient_export_raw_v1`
+  `moz-fx-data-marketing-prod.acoustic.raw_recipient_raw_v1`
```
jklukas, line 16: I don't have context to understand the naming convention here. Why does `raw` appear twice in this name? Can you provide that context in descriptions in the relevant `metadata.yaml` files?

jklukas (approved):
r+ with comments

sql/moz-fx-data-marketing-prod/acoustic/raw_recipient_v1/query.sql
kik-kik: I'll add this as a separate change next week when I'm back.

kik-kik: Also, just to give some context here, raw_recipient is name of the actual type of object for which I pull data from Acoustic `raw_v1` is to represent this is raw data (raw layer) as coming from Acoustic.

PR #2941, opened 2022-05-05
Only publish static tables in user-facing datasets to `mozdata` (bug 1768018)

sean-rose:
This will resolve [bug 1768018](https://bugzilla.mozilla.org/show_bug.cgi?id=1768018) "_Airflow task publish_bqetl_static.publish_static_mozdata failing on 2022-05-03_".

Also:
- Allows the `publish_static` script to work for any project rather than just `moz-fx-data-shared-prod`.
- Fixes a couple of bugs.
- Uses a more efficient approach for finding `data.csv` files.

---
Checklist for reviewer:

### DIFF
--- a/bigquery_etl/publish_static.py
+++ b/bigquery_etl/publish_static.py
@@ -34,7 +34,7 @@ def _load_table(
     data_file_path, schema_file_path=None, description_file_path=None, project=None
 ):
     # Assume path is ...project/dataset/table/data.csv
-    path_split = os.path.normcase(data_file_path).split("/")
+    path_split = os.path.normcase(data_file_path).split(os.path.sep)
     dataset_id = path_split[-3]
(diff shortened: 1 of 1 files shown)


sean-rose:
NB:  I've intentionally allowed for running `publish_static` with no `--project_id` argument, which will publish static tables found in all project directories to the associated project.

The [current Airflow tasks](https://github.com/mozilla/telemetry-airflow/blob/bfb691662cdbd57462eea8f4206b775a245cb337/dags/publish_bqetl_static.py#L38-L54) all specify a `--project_id` argument, so this won't be used at the moment, but we could replace the `publish_static_shared_prod` task with a `publish_static_all` task if we wanted (`mozdata` would still have to be a special case with a dedicated task).

jklukas (approved):

PR #2960, opened 2022-05-11
Update docs to remove the default mozilla-public-data as default in docs

lucia-vargas-a:
The instructions to create a new table (which also populates them) uses the mozilla-public-data project as default to publish the data. This is a risk that can lead to potentially publish non-public data in this project.

Checklist for reviewer:


docs/cookbooks/creating_a_derived_dataset.md
```diff
@@ -188,16 +188,19 @@ GROUP BY
   os_version;
 ```
 
-As the dataset will be published in both Mozilla's public BigQuery project, the table will need to be created in the `mozilla-public-data` project. Once the ETL is running, a view `org_mozilla_mozregression_derived.mozregression_aggregates` to the public table will be automatically created in `moz-fx-data-shared-prod`.
```
lucia-vargas-a, line 191: This should make it more clear how to use the init.sql file. However I still find that the documentation here is disaligned with the documentation to [add a new scheduled query](https://mozilla.github.io/bigquery-etl/cookbooks/common_workflows/#adding-a-new-scheduled-query) which suggest to deploy a schema and run a backfill instead of using an init file. Is there a reason to keep both?

akkomar: There's definitely some duplication between this doc and common_workflows cookbook which was added later than this one. However this cookbook looks like a description of some specific use case and publishing a public dataset is one of the goals stated above: https://github.com/mozilla/bigquery-etl/blob/f1648b842dadfa11f165021cd2c8c03c22019c62/docs/cookbooks/creating_a_derived_dataset.md?plain=1#L11
Maybe the title here could be clearer?

jklukas: > suggest to deploy a schema and run a backfill instead of using an init file

The concept of an `init.sql` file was introduced in this codebase before the concept of `schema.yaml` files and the `bqetl backfill` machinery. The latter solves a wider range of problems and is preferred over `init.sql` now, so it would be great to update this example to reflect that.

lucia-vargas-a: @akkomar I added it to the title but also kept generic commands so that the instructions can be used for any dataset and highlighted the specifics of public datasets.

lucia-vargas-a: @jklukas Updated! I removed the use of the initialization process completely. We could also keep it as a second alternative to be used in a very particular case (if any).

akkomar: @lucia-vargas-a thank you for improving this, now it's immediately clear to me what to expect from this article.

scholtzan (approved):

PR #2961, opened 2022-05-11
Unschedule import of vpn waitlist table that was removed upstream

relud:


sean-rose (approved):

sean-rose:
The deleted files can be removed from the dry-run exclusions.

PR #2970, opened 2022-05-16
Updated metadata and DAG for cohort_daily_statistics

ANich:
Updates similar to #2969 and #2968. PR explicitly adds `submission_date` as a `bq` parameter to get activity for that day.


sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v1/metadata.yaml
```diff
@@ -14,8 +14,13 @@ labels:
 scheduling:
   dag_name: bqetl_analytics_aggregations
   date_partition_parameter: cohort_date
-  arguments: ['--append_table']
+  arguments: ["--append_table"]
+  parameters: ["submission_date:DATE:{{ds}}"]
```
lucia-vargas-a, line 18: Append also makes sense for active_users_aggregates. I will add that in a separate PR.

lucia-vargas-a (approved):

PR #2972, opened 2022-05-16
Use the query metadata's defined date_partition_parameter for CLI backfills

ANich:
Similar to the dryrun module. For queries that have date partitions other than `submission-date`


bigquery_etl/cli/query.py
```diff
@@ -445,6 +445,7 @@ def info(ctx, name, sql_dir, project_id, cost, last_updated):
 def _backfill_query(
     query_file_path,
     project_id,
+    date_partition_parameter,
```
kik-kik, line 448: maybe we should have set a default for this `date_partition_parameter="submission_date` for backwards compatibility? @ANich

PR #2973, opened 2022-05-16
Remove redundant GOOGLE_APPLICATION_CREDENTIALS in integration

relud:
maybe this can solve the issue with flaky integration tests? seems unlikely, but needs fixing regardless

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -176,16 +176,9 @@ jobs:
             venv/bin/pip-sync --pip-args=--no-deps requirements.txt
       - run:
           name: PyTest Integration Test
-          # Google's client libraries will check for
-          # GOOGLE_APPLICATION_CREDENTIALS
-          # and use a file in that location for credentials if present;
-          # See https://cloud.google.com/docs/authentication/production
           # yamllint disable rule:line-length
           command: |
-            export GOOGLE_APPLICATION_CREDENTIALS="/tmp/gcp.json"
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -176,16 +176,9 @@ jobs:
...
           name: PyTest Integration Test
-          # Google's client libraries will check for
-          # GOOGLE_APPLICATION_CREDENTIALS
-          # and use a file in that location for credentials if present;
-          # See https://cloud.google.com/docs/authentication/production
           # yamllint disable rule:line-length
           command: |
-            export GOOGLE_APPLICATION_CREDENTIALS="/tmp/gcp.json"
-            echo "$GCLOUD_SERVICE_KEY" > "$GOOGLE_APPLICATION_CREDENTIALS"
             PATH="venv/bin:$PATH" script/entrypoint -m 'integration or java' -n 8
```
relud, line 181: `script/entrypoint` already handles creating `/tmp/gcp.json` and setting `GOOGLE_APPLICATION_CREDENTIALS`

scholtzan (approved):

sean-rose:
> maybe this can solve the issue with flaky integration tests? seems unlikely, but needs fixing regardless

`script/entrypoint` has a check that `GCLOUD_SERVICE_KEY` isn't empty before using it, so if the problem is that sometimes that environment variable isn't being set/propagated properly in CI then at least we'll get a different error message now.

PR #2975, opened 2022-05-16
Bug 1767487 - mitigate conflicting writes between shredder and ETL

relud:


bigquery_etl/shredder/delete.py
```diff
@@ -192,9 +192,11 @@ def delete_from_partition(
...
-    # special partitions can't be set as a destination, so DML must be used except
-    # in dry runs because destination isn't required and DML may not be allowed
-    if partition.is_special and not dry_run:
+    # whole table operations must use DML to protect against dropping partitions in the
+    # case of conflicting write operations in ETL, and special partitions must use DML
+    # because they can't be set as a query destination, except in dry runs because DML
+    # may not be allowed and destination isn't required
+    if (partition.id is None or partition.is_special) and not dry_run:
         use_dml = True
     elif not use_dml and not partition.is_special:
```
sean-rose, line 201: This might be appropriate now:
```suggestion
    elif not use_dml and not (partition.id is None or partition.is_special):
```
And then the `job_config.destination =` expression below could be simplified to `sql_table_id(target) + f"${partition.id}"` (GitHub's annoying limit on which lines you can comment on prevents me from making that part of the suggested change).

sean-rose (approved):

bigquery_etl/shredder/delete.py
relud: `job_config.destination` should still be set if `not use_dml and partition.id is None`

sean-rose: But with this change that'd only be possible during a dry run, and the comment says "in dry runs ... destination isn't _required_".  Though I guess that doesn't necessarily mean it doesn't _matter_ in some cases.

relud: yeah, select statements used to be equivalent and never set destination because they were only for dry runs (which are not used in tests, it's a flag for manual use). it would make more sense to remove the `and not dry_run` flag entirely.

relud, line 201: per discussion above
```suggestion
    if partition.id is None or partition.is_special:
        use_dml = True
    elif not use_dml:
```

sean-rose (approved):

PR #2977, opened 2022-05-17
Calculate is_new_profile using first_seen_date for Fenix, Focus and F…

lucia-vargas-a:
There are many cases in the `fenix.clients_last_seen_joined` in which is_new_profile IS TRUE but first_seen_date <> submission_date.

Purpose is to:
- Unify the calculation of is_new_profile in all sub-queries
- Correct the calculation of is_new_profile in unified_metrics, which inherits to correct the count of new_profiles in active_users_aggregates

This query returns an example of the problem:
```
SELECT
submission_date,
first_seen_date,
days_since_seen,
is_new_profile,
first_seen_date=submission_date AS submission_date_equals_first_seen_date
FROM
  fenix.clients_last_seen_joined
WHERE
  submission_date = '2022-04-01'
  AND is_new_profile IS TRUE
  AND first_seen_date <> submission_date
```
image


kik-kik:
Maybe one question, what exactly is the `submission_date` field in this case?

lucia-vargas-a:
@kik-kik  From the first_seen_date of a client and onwards, we keep track of the activity per date ( that is the submission_date), which -as we saw last friday- doesn't necessarily mean that a ping was received. Only when `submission_date` matches the `first_seen_date` a client can be considered new.

kik-kik:
> @kik-kik From the first_seen_date of a client and onwards, we keep track of the activity per date ( that is the submission_date), which -as we saw last friday- doesn't necessarily mean that a ping was received. Only when `submission_date` matches the `first_seen_date` a client can be considered new.

Gotcha, I wonder if this would be a good opportunity to rename the `submission_date` field to something a bit more descriptive / representative of the meaning.

kik-kik (approved):

gkabbz (approved):

ANich (approved):
> There are many cases in the fenix.clients_last_seen_joined in which is_new_profile IS TRUE but first_seen_date <> submission_date.

I think you mentioned doing this offline already but just a reminder, this is definitely worth a bug/issue I think

PR #2982, opened 2022-05-19
Fix experiments status

scholtzan:
We had some experiments causing problems after they ended:

So after every daily run jetstream checks for all active experiments whether an external config has changed (both experiment-specific configs and platform-specific configs). If a config has been changed then a rerun is triggered for the affected experiment. Each result table in BigQuery has a timestamp, so these timestamps are used to compare them to the last_modified timestamp of the external configs.
It appears that ending the experiments changed their statuses from "Complete" to "Live" 2 days ago due to the end date changing from null to a date. The experiments have been running for about 197 days. Within that timeframe the fenix.toml config got updated (24 days ago). So while the experiments were considered by jetstream as complete, the external config change went unnoticed. But once the experiments got set to "Live" jetstream checked the timestamps of the result tables and noticed that the external fenix.toml config got updated (24 days ago) after some of the result tables got created (197 days ago). So it triggered a rerun.

Checklist for reviewer:


sql/moz-fx-data-experiments/monitoring/experimenter_experiments_v1/query.py
```diff
@@ -158,8 +158,8 @@ def to_experiment(self) -> "Experiment":
             experimenter_slug=None,
             type="v6",
             status="Live"
-            if self.endDate
-            and self.endDate <= pytz.utc.localize(datetime.datetime.now())
+            if (self.endDate is None or (self.endDate
+            and self.endDate <= pytz.utc.localize(datetime.datetime.now())))
```
scholtzan, line 162: Live experiments might not have an endDate set.

danielkberry: Copying my feedback from slack:
> I don't have great familiarity with the experimenter API, but this seems logical to me (that live experiments may not have an end date or that the end date is in the future).

LGTM

danielkberry (approved):

PR #2987, opened 2022-05-23
Use static dataset to convert locale into the corresponding ISO 639-1…

lucia-vargas-a:
The `locale` in active_users_aggregates currently stores the locale code, which is more meaningful for analysis id the corresponding language name is shown. Also using language names helps decreasing the size of the aggregation.

In order to translate the local to a language name there are a few options:

1. UDF - Since there are 185 language codes this might not be the best fit for a UDF.
2. Lookup table - This may be a good option even though there would be only one level (code - language name). Data can be retrieved from [DataHub](https://datahub.io/core/language-codes#data-cli) or from [CLDR](https://cldr.unicode.org/index/cldr-spec/core-data-for-new-locales) and would require an sql-generator.
3. Use the existing dataset in bigquery `mozdata.static.csa_gblmkt_languages` which contains 128 codes and was a one time load as part of this [private-bigquery-etl](https://github.com/mozilla/private-bigquery-etl/tree/main/sql/moz-fx-data-shared-prod/static/csa_gblmkt). This would require permissions to use it, and would potentially require an update to retrieve all the 185 codes.

This PR is a proposal for Option 3. The missing codes would be grouped as Other and the possibility to update it should be confirmed with [@cocomo](https://mozilla.slack.com/team/U4DSB7ZJM)

Checklist for reviewer:


sql/moz-fx-data-shared-prod/telemetry_derived/active_users_aggregates_device_v1/query.sql
```diff
@@ -0,0 +1,46 @@
...
+  device_model,
+  EXTRACT(YEAR FROM first_seen_date) AS first_seen_year,
+  is_default_browser,
+  normalized_app_name AS app_name,
+  normalized_channel AS channel,
+  normalized_os AS os,
+  normalized_os_version AS os_version,
+  os_version_major,
+  os_version_minor,
+  submission_date,
```
gkabbz, line 17: I would add country in as well.

PR #2999, opened 2022-05-31
Bug 1770814 adm_daily_aggregates

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1770814

Checklist for reviewer:

### DIFF
--- a/dags/bqetl_search_terms_daily.py
+++ b/dags/bqetl_search_terms_daily.py
@@ -53,6 +53,22 @@ with DAG(
     tags=tags,
 ) as dag:
 
+    search_terms_derived__adm_daily_aggregates__v1 = bigquery_etl_query(
+        task_id="search_terms_derived__adm_daily_aggregates__v1",
+        destination_table="adm_daily_aggregates_v1",
+        dataset_id="search_terms_derived",
+        project_id="moz-fx-data-shared-prod",
+        owner="jklukas@mozilla.com",
+        email=[
+            "jklukas@mozilla.com",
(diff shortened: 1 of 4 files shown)


rebecca-burwei (commented):
Do we need something like [this](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/search_terms_derived/aggregated_search_terms_daily_v1/query.sql#L15) as we have implemented for ourselves?

jklukas:
> Do we need something like [this](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/search_terms_derived/aggregated_search_terms_daily_v1/query.sql#L15) as we have implemented for ourselves?

I would assume that the partner expectation here is to have complete data without a filter. Assuming that's the case, then I agree we should relax the internal policy.

jklukas:
I am seeking more clarity on the filter question in https://bugzilla.mozilla.org/show_bug.cgi?id=1770814

whd (approved):
I will work with @jklukas to deploy this when/if it's ready

rebecca-burwei:
Sounds good. Thank you for sorting those things out @jklukas

rebecca-burwei (approved):

jklukas:
> I will work with @jklukas to deploy this when/if it's ready

This is ready to deploy at will. We won't actually start sending the data until we merge the Airflow change.

whd:
I've deployed this, including the ACL grant even though that hasn't propagated to generated branches yet.

PR #3002, opened 2022-06-01
pushing bqetl acoustic dags by 1 hour to see if that resolves sensor issues

kik-kik:
pushing bqetl acoustic dags by 1 hour to see if that resolves sensor issues

for some reason the external task sensor fails to realise task completed. Noticed that other DAGs using this pattern have different intervals for upstream and downstream DAGs. Trying to see if doing the same here helps resolve the issue.

### DIFF
--- a/dags.yaml
+++ b/dags.yaml
@@ -719,7 +719,7 @@ bqetl_cjms_nonprod:
     - impact/tier_3
 
 bqetl_acoustic_contact_export:
-  schedule_interval: 0 8 * * *
+  schedule_interval: 0 9 * * *
   description: |
     Processing data loaded by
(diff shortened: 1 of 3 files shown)


dags.yaml
```diff
@@ -719,7 +719,7 @@ bqetl_cjms_nonprod:
     - impact/tier_3
 
 bqetl_acoustic_contact_export:
-  schedule_interval: 0 8 * * *
+  schedule_interval: 0 9 * * *
```
lucia-vargas-a, line 722: LGTM. Also, no other DAGS scheduled at this time.

lucia-vargas-a (approved):

PR #3003, opened 2022-06-01
Revert "Bug 1770762 - Add Focus, Firefox iOS, Klar to `nondesktop_clients_last_seen`"

jklukas:
Reverts mozilla/bigquery-etl#2998

This is failing to deploy with `Column 28 in UNION ALL has incompatible types: DATE, DATE, DATE, DATE, DATE, DATE, DATE, STRING, DATE, STRING, DATE at [10:3]` so there is a difference in column ordering in the union we'll need to deal with.


jklukas:
See fix in https://github.com/mozilla/bigquery-etl/pull/3005

PR #3008, opened 2022-06-02
added sql and table definition for getting "current/latest" version acoustic contacts

kik-kik:
added sql and table definition for getting "current/latest" version acoustic contacts


sql/moz-fx-data-marketing-prod/acoustic/contact_current_snapshot_v1/init.sql
```diff
@@ -0,0 +1,48 @@
+CREATE TABLE IF NOT EXISTS
```
ANich, line 1: Do we actually need this given we have the schema and metadata?

ANich (approved):
Left one comment

sql/moz-fx-data-marketing-prod/acoustic/contact_current_snapshot_v1/init.sql
kik-kik: I think you're right about this. I'll remove that file.

PR #3028, opened 2022-06-14
[Bug 1762430] Generate ExternalTaskMarkers for Airflow downstream dependencies

scholtzan:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1762430

This updates the Airflow DAG generation logic to generate `ExternalTaskMarker`s for downstream tasks. This ensures that when clearing a task for a rerun, all downstream tasks get automatically cleared as well. Newer versions of Airflow (those with a grid view) nicely group all the tasks part of a `TaskGroup` instead of showing the individual tasks.

This update also makes a change to use Airflow's `ExternalTaskSensor` instead of the custom `ExternalTaskCompletedSensor`.

The change diff looks scarily large, but the actual code changes are pretty minimal. Most changes are due to the Airflow DAGs having been regenerated.

Checklist for reviewer:


bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -1,7 +1,9 @@
 # Generated via https://github.com/mozilla/bigquery-etl/blob/main/bigquery_etl/query_scheduling/generate_airflow_dags.py
 
 from airflow import DAG
-from operators.task_sensor import ExternalTaskCompletedSensor
```
scholtzan, line 4: Using `ExternalTaskSensor` instead. This is causing almost all of the DAGs to have some changes. If this is annoying for reviewing this, then I can also open a separate PR for this change.

whd (approved):
I'm approving this from a strictly CODEOWNERS/change control perspective as this change need not be subject to it

jklukas (approved):
The code changes here look good. 

Do we have a written plan for communicating the change and gameplan for rolling it out and testing or rolling back as needed? In particular, I think this should be coordinated with whoever is on Airflow triage in case it causes some unintended breakage or needs to be rolled back.

scholtzan:
I was checking if there is an option to disable "Downstream Recursive" clears by default, to avoid clearing downstream tasks unnecessarily. However, there is no such setting in Airflow.

PR #3034, opened 2022-06-17
Add scheduling for the merino log sanitized table population

chelseatroy:
This time I learned my lesson about not pushing to a fork!


jklukas (changes requested):
These changes should be on `suggest_impression_sanitized_v3` rather than this table. We'll need to move this change to the other `metadata.yaml` file and then run `bqetl dag generate`.

See https://mozilla.github.io/bigquery-etl/cookbooks/common_workflows/#adding-a-new-scheduled-query for broader context on the workflow and when `bqetl dag generate` is needed.

jklukas (approved):
👍

whd (approved):

PR #3037, opened 2022-06-20
 Bug 1772532 - Use HLL for user counts in GLAM scalar_bucker_counts_v1 

akkomar:
`scalar_bucket_counts_v1` queries started failing in 2022-06 for FOG and Fenix.
They started exceeding BQ disk and memory limits available for shuffle operations, most likely because of growing number of clients, metrics, and values.

This commit introduces HyperLogLog for estimating number of clients in `scalar_bucket_counts`. In my tests, after this change, query was finishing in 30-70 minutes while the current production query fails after 4-6 hours.

I have run this query before and after the change on a 50% sample of input table.
In terms of BQ slots used, this uses ~700 vs. ~2100 before - although not exact, this can be treated as an approximation of the run time.
In terms of errors, this notebook compares both outputs and user counts: https://colab.research.google.com/drive/1uilzQcFn1ppFMTTpi-RXj2EFLjCxX_mN#scrollTo=VfRZpBCjxgQM. It shows that 99.8% of counts estimated with HLL have an error smaller than 0.1%, 81% are below 0.01%. This makes this approach pretty good compared to sampling that we use in the legacy pipeline as it does not introduce risk of sampling-out some populations.

Checklist for reviewer:

### DIFF
--- a/bigquery_etl/glam/templates/scalar_bucket_counts_v1.sql
+++ b/bigquery_etl/glam/templates/scalar_bucket_counts_v1.sql
@@ -3,14 +3,28 @@
 {% from 'macros.sql' import enumerate_table_combinations %}
 
 WITH
-{{
-    enumerate_table_combinations(
-        source_table,
-        "all_combos",
-        cubed_attributes,
-        attribute_combinations
-    )
-}},
+build_ids AS (
+  SELECT
+    app_build_id,
+    channel,
+  FROM
+    {{ source_table }}
+  GROUP BY
+    app_build_id,
+    channel
+  HAVING
+    COUNT(DISTINCT client_id) > {{ minimum_client_count }}
+),
+valid_clients_scalar_aggregates AS (
+  SELECT
+    *
+  FROM
+    {{ source_table }}
+  INNER JOIN
+    build_ids
+  USING
+    (app_build_id, channel)
+),
 bucketed_booleans AS (
   SELECT
     client_id,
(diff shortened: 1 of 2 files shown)


akkomar:
After this is merged and confirmed to be running stable in production for a couple of days we'll need to update GLAM UI and make a note that scalar user counts are an approximation.

bigquery_etl/glam/templates/scalar_bucket_counts_v1.sql
```diff
@@ -3,14 +3,28 @@
...
-    )
-}},
+build_ids AS (
+  SELECT
+    app_build_id,
+    channel,
+  FROM
+    {{ source_table }}
+  GROUP BY
+    1,
```
edugfilho, line 13: it's usually good to write the column names explicitly to avoid bugs in case the selected columns get changed

sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__scalar_bucket_counts_v1/query.sql
```diff
@@ -89,39 +89,27 @@ RETURNS ARRAY<
...
-        ("*", NULL, NULL),
-        (NULL, "*", "*"),
-        ("*", NULL, "*"),
-        ("*", "*", NULL),
-        ("*", "*", "*")
-      ]
-    ) AS combos
+    glam_etl.org_mozilla_fenix_glam_nightly__clients_scalar_aggregates_v1
+  GROUP BY
+    1,
```
edugfilho, line 99: it's usually good to write the column names explicitly to avoid bugs in case the selected columns get changed

edugfilho (approved):
Great idea to apply hll count, and I really appreciate the work done around it to show the outcome.
I left suggestions regarding writing column names specifically as opposed to their positions. Approved

PR #3039, opened 2022-06-21
Use approximate client count in GLAM scalar_percentiles_v1

akkomar:
This is a follow-up to https://github.com/mozilla/bigquery-etl/pull/3037 which unblocked `scalar_bucket_counts_v1`.
`scalar_percentiles_v1` uses the same source table (`clients_scalar_aggregates_v1`) and started failing today with the same error (disk/memory limits exceeded for shuffle operations). We haven't noticed it earlier because It is a downstream query to `scalar_bucket_counts_v1` in Airflow.

`APPROX_COUNT_DISTINCT` used here runs HLL under the hood. The reason for using it here is that we can't split the aggregation here into two stages as in the aforementioned PR due to quantiles calculation.

I have run this query locally and confirmed that it works.

Checklist for reviewer:

### DIFF
--- a/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__scalar_percentiles_v1/query.sql
+++ b/sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__scalar_percentiles_v1/query.sql
@@ -52,8 +52,8 @@ percentiles AS (
     key,
     agg_type AS client_agg_type,
     'percentiles' AS agg_type,
-    COUNT(DISTINCT(client_id)) AS total_users,
-    APPROX_QUANTILES(value, 100) AS aggregates
+    APPROX_COUNT_DISTINCT(client_id) AS total_users,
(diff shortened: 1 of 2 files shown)


sql/glam-fenix-dev/glam_etl/org_mozilla_fenix_glam_nightly__scalar_percentiles_v1/query.sql
```diff
@@ -69,7 +69,10 @@ percentiles AS (
 )
 SELECT
   * REPLACE (
-    mozfun.glam.map_from_array_offsets([5.0, 25.0, 50.0, 75.0, 95.0], aggregates) AS aggregates
+    mozfun.glam.map_from_array_offsets_precise(
+      [0.1, 1.0, 5.0, 25.0, 50.0, 75.0, 95.0, 99.0, 99.9],
```
akkomar, line 73: This is changing because this query hasn't been recreated after https://github.com/mozilla/bigquery-etl/pull/2723/files

PR #3040, opened 2022-06-21
[Bug 1775216] Fix app names in nondesktop_clients_last_seen

scholtzan:
https://bugzilla.mozilla.org/show_bug.cgi?id=1775216

The app names in `nondesktop_clients_last_seen` had a trailing 'Baseline' which resulted in `mozfun.norm.product_info` transforming them to 'Other'. Removing 'Baseline' results in the product details getting transformed correctly and data appearing in downstream tables, such as `telemetry.smooth_usage_nondesktop` (which is what the bug was asking for). We'll need to run a backfill.


sql/moz-fx-data-shared-prod/telemetry/nondesktop_clients_last_seen_v1/view.sql
```diff
@@ -34,7 +34,7 @@ WITH glean_final AS (
     country,
     locale,
     app_display_version,
-    'Lockwise Baseline' AS app_name,
+    'Lockwise Android' AS app_name,
```
jklukas, line 37: These Baseline names were introduced for apps that are also present in `core` pings. We may introduce double-counting of clients from Glean and legacy telemetry by making these name changes. Focus is one of those apps.

scholtzan: I see. Would it make sense to ignore Focus/Klar clients from `core` pings by filtering them out here: https://github.com/mozilla/bigquery-etl/blob/98549e3cb8b4f81d03a08a33aedaf6687ea5f769/sql/moz-fx-data-shared-prod/telemetry/nondesktop_clients_last_seen_v1/view.sql#L223 and instead use their Glean data?

sql/moz-fx-data-shared-prod/telemetry/nondesktop_clients_last_seen_v1/view.sql
```diff
@@ -239,7 +239,7 @@ unioned AS (
     locale,
     NULL AS distribution_id,
     app_display_version AS app_version,
-    mozfun.norm.product_info(app_name, normalized_os) AS pinfo,
+    udf.product_info_with_baseline(app_name, normalized_os) AS pinfo,
```
scholtzan, line 242: This results in all the "Baseline" apps showing up in the smoot tables as "Baseline" as well.

sql/moz-fx-data-shared-prod/udf/product_info_with_baseline/udf.sql
```diff
@@ -0,0 +1,95 @@
...
+  contributes_to_2020_kpi BOOLEAN,
+  contributes_to_2021_kpi BOOLEAN
+> AS (
+  CASE
+  WHEN
+    legacy_app_name LIKE "Focus iOS Baseline"
+    AND normalized_os LIKE "iOS"
+  THEN
+    STRUCT(
+      "focus_ios" AS app_name,
```
scholtzan, line 17: This could also be `focus_ios_baseline`, but since this UDF is only used for smoot, which ignores `app_name`, it doesn't really matter. All that matter are `product` and `canonical_name`

jklukas (approved):
r+wc

PR #3041, opened 2022-06-21
#1775029 Update the suggest_impression_sanitized_v3 query

quiiver:
* exclude `region` and `country` when preparing for the join
* filter `impressions.request_id` to non null to drop queries without a
  corresponding impression.
* Add tests? Haven't figured out bootstraping issues on my M1 yet so not
  sure how well these will work. TO CI!

Checklist for reviewer:


quiiver:
I'm a little unsure how to resolve the test failure here. In the v2_external test we explicitly allow impressions without a corresponding query to make it through to the impressions table. however in this query we are only outputting queries with a 1-1 match (ie Inner join)

sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v3/query.sql
```diff
@@ -80,3 +80,5 @@ LEFT JOIN
   impressions
 USING
   (request_id)
+WHERE
```
quiiver, line 83: if we want to continue with this. i'm going to change this to `INNER JOIN` and drop the conditional stmt 🤦

quiiver:
> I'm a little unsure how to resolve the test failure here.

I had a conversation with @jklukas about this in slack. I've copied the relevant parts of the conversation into here for posterity.

> [wil]
> So I think what we are saying is that the impressions_sanitized table should contain all impressions and the subset of queries that match them
> 
> [klukas]
> Yes, I agree with that
> :+1:
> 
> [klukas]
> We now null out the search_query field in the pipeline for the small number of pings we still receive from older online clients that aren’t interacting with Merino, so this particular situation can’t occur anymore. But I still think the best contract for this table is that it has the same number of rows as the source quicksuggest_impression_v1 table

sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v3/query.sql
```diff
@@ -37,7 +37,9 @@ WITH impressions AS (
 ),
 sanitized_queries AS (
   SELECT
-    *
+    TIMESTAMP_TRUNC(timestamp, SECOND) as timestamp,
+    LTRIM(LOWER(query)) as query,
+    * except (timestamp, query)
```
jklukas, line 42:
```suggestion
    * except (timestamp, query, region, country)
```

I think it would be cleaner to remove these in this initial query rather than in the join.

sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v3/query.sql
```diff
@@ -56,7 +58,7 @@ sanitized_queries_count AS (
 -- output, which allows us to raise an error in the WHERE clause.
 validated_queries AS (
   SELECT
-    * EXCEPT (_n, _n_with_query),
+    * EXCEPT (_n, _n_with_query, region, country),
```
jklukas, line 61: I think these should be removed earlier.

```suggestion
    * EXCEPT (_n, _n_with_query),
```

jklukas:
> Haven't figured out bootstraping issues on my M1 yet

We'll also need to run `bqetl format sql/` on this to keep the linter happy, so let me know if you need me to run that and push results. Or we can schedule time to try to figure out bootstrapping issues.

quiiver:
i was able to build the docker image and i'm using that to format with the following command:

```
docker run --mount type=bind,source=$(pwd)/sql,target=/sql bqetl bqetl format /sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v3/query.sql
```

quiiver:
@whd is there anything else remaining that I need to do to get this merged?

jklukas:
@whd is OOO today. I'll ping @jasonthomas about codeowner approval.

jklukas:
> I'll ping @jasonthomas about codeowner approval.

Who appears to also be OOO today. So this will need to wait for tomorrow.

PR #3048, opened 2022-06-23
Set owners for Contextual Services tasks

jklukas:
I made some guesses here about who would make the most sense to own individual pieces; treat this as a starting point and we can discuss in PR comments to get this to final state.


dags.yaml
```diff
@@ -571,25 +572,26 @@ bqetl_search_terms_daily:
...
     Derived tables on top of search terms data.
 
-    Note that the task for populating `suggest_impression_sanitized_v2` is
+    Note that the tasks for populating `suggest_impression_sanitized_v*` are
     particularly important because the source unsanitized dataset has only
     a 2-day retention period, so errors fairly quickly become unrecoverable
     and can impact reporting to partners. If this task errors out, it could
     indicate trouble with an upstream task that runs in a restricted project
-    outside of Airflow. Contact `jklukas` and/or `jbuck`.
+    outside of Airflow. Contact `ctroy`, `whd`, and `jbuck`.
```
quiiver, line 594: we can list me as a contact here as well

quiiver (commented):
This looks good to me. I'm not entirely sure what ownership means exactly but I assume it means you're the primary contact for these tables. I'm happy to take ownership of any of these if folks are concerned about ownership overload.

rebecca-burwei:
I'm fine with owning event_aggregates and aggregated_search_terms. I would be okay with owning or being alerted on the sanitization job as well. I don't know that I should own adm_*_aggregates since I don't work with these tables' stakeholders.

Same question as Wil about what ownership means. I can fix SQL and config files, but not airflow jobs.

jklukas:
The primary way that ownership info gets used is in Airflow triage. When a DAG has failing tasks, whoever in DE is checking on Airflow tasks in a given week will file a bug and assign it to the person listed as owner for that DAG. See [Airflow Triage confluence page](https://mana.mozilla.org/wiki/x/eTn5C). So the person owning a given task or DAG should be someone who will be able to take some action to respond to errors.

Ownership of individual queries is less actionable. It may also ultimately be used to route communication if there are issues with running a task. But mostly the owner and email list are used to figure out who to talk to if someone has questions about what the data is used for.

rebecca-burwei:
Ok, seems like I do have access to the airflow infrastructure. I'm comfortable owning event_aggregates and aggregated_search_terms. I'm less comfortable owning adm_*_aggregates.

jklukas:
I've made updates based on the specific comments here. I'd definitely like to make sure @chelseatroy has a chance to weigh in and potentially transfer some of these things to @quiiver. So will leave open for now.

whd (approved):
r+ from a change control perspective, @chelseatroy to review before merge

jasonthomas (approved):

dags.yaml
```diff
@@ -571,25 +572,26 @@ bqetl_search_terms_daily:
...
     Derived tables on top of search terms data.
 
-    Note that the task for populating `suggest_impression_sanitized_v2` is
+    Note that the tasks for populating `suggest_impression_sanitized_v*` are
     particularly important because the source unsanitized dataset has only
     a 2-day retention period, so errors fairly quickly become unrecoverable
     and can impact reporting to partners. If this task errors out, it could
     indicate trouble with an upstream task that runs in a restricted project
-    outside of Airflow. Contact `jklukas` and/or `jbuck`.
+    outside of Airflow. Contact `ctroy`, `wstuckey`, `whd`, and `jbuck`.
```
chelseatroy, line 594: Please forgive my ignorance: what is/happens in this job?

sql/moz-fx-data-shared-prod/search_terms_derived/adm_daily_aggregates_v1/metadata.yaml
```diff
@@ -7,7 +7,7 @@ description: |-
   The sending to adMarketplace takes place in a separate DAG defined in
   the telemetry-airflow repository directly.
 owners:
-  - jklukas@mozilla.com
+  - ctroy@mozilla.com
```
chelseatroy, line 10: Just to make sure I understand it all: what is/happens in this job?

sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2/metadata.yaml
```diff
@@ -9,7 +9,7 @@ description: |-
 
   Fulfills https://mozilla-hub.atlassian.net/browse/ROAD-85
 owners:
-  - jklukas@mozilla.com
+  - ctroy@mozilla.com
```
chelseatroy, line 12: Happy to overtake this. For my edification, under what circumstances are we using v2 of this, versus when should we use v3?

chelseatroy: - And I assume v1 is deprecated/not in use?

chelseatroy (commented):
Added some questions!

dags/bqetl_ctxsvc_derived.py
```diff
@@ -18,15 +18,19 @@
 Contextual services derived tables
 #### Owner
 
-jklukas@mozilla.com
+ctroy@mozilla.com
```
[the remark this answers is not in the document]
jklukas: See https://workflow.telemetry.mozilla.org/tree?dag_id=bqetl_ctxsvc_derived

There's only one query that currently runs in this DAG. It's the `event_aggregates_v1` table that is used for internal analysis of impression and click counts by dimensions.

sql/moz-fx-data-shared-prod/search_terms_derived/adm_daily_aggregates_v1/metadata.yaml
jklukas: This query populates a table of daily count of engagement per allowlisted search term. There's a separate DAG defined in telemetry-airflow that sends the contents of this table to a partner SFTP server: https://github.com/mozilla/telemetry-airflow/blob/main/dags/adm_export.py

sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2/metadata.yaml
jklukas: The v1 query did nothing more than remove city-level geo data from the impressions. It has been removed.

v2 can go away once we are sufficiently comfortable with v3 as production-grade solution and we have adapted any downstream queries to use v3.

whd:
This may need some refinement but I'm going to merge as-is and we can iterate

PR #3056, opened 2022-06-24
[Bug 1770015] Allowlist for glean apps in glean_usage

scholtzan:
Temporary workaround for https://bugzilla.mozilla.org/show_bug.cgi?id=1770015

cc @whd


scholtzan:
Good call! We probably should document this somewhere, however this is only meant to be a temporary workaround until we make some modifications to schema deploys which will then not require us to manually add Glean apps here.

sql_generators/glean_usage/__init__.py
```diff
@@ -103,7 +153,7 @@ def generate(target_project, output_dir, parallelism, exclude, only, app_name):
     if app_name:
         app_info = {name: info for name, info in app_info.items() if name == app_name}
 
-    app_info = [info for name, info in app_info.items() if name not in SKIP_APPS]
+    app_info = [info for name, info in app_info.items() if name in ALLOWED_APPS]
```
jklukas, line 156: Do we know if the baseline tables are all deploying correctly automatically? If it's specifically metrics ETL that has been causing issues, then perhaps we can still allow all the baseline ping ETL to flow for new apps?

jklukas (approved):
r+wc It would be great if we can still allow baseline ETL to flow, but I'm okay with merging this as-is to avoid the current state of breakage for all apps every time a new Glean app is deployed.

sql_generators/glean_usage/__init__.py
scholtzan: Right, baseline tables were unaffected by this. I'll make some updates

PR #3064, opened 2022-06-27
Fix `mozilla_vpn_derived.active_subscription_ids_live` view

sean-rose:
As a side-effect of #3031 the `mozilla_vpn_derived.active_subscription_ids_live` view is now including a few subscriptions from the current date, which is causing the `mozilla_vpn_derived.subscription_events_live` view to consider most active subscriptions as being cancelled on the current date.

To fix this `mozilla_vpn_derived.active_subscription_ids_live` will now exclude subscriptions that started after the last ETL'd date (we can't rely on `CURRENT_DATE`, otherwise there would be a time period between UTC midnight and when the ETL runs where the same problem would occur).

---
Checklist for reviewer:


sean-rose:
The `validate-views` CI failed due to unrelated permission errors.

relud (approved):

PR #3067, opened 2022-06-28
Add a virtual view for the language data from the sanitization job

chelseatroy:
This PR adds a virtual view that offers a tabular representation, for each sanitation job, of:

- Which languages the NLP model detected in the search terms
- How many search terms got categorized as each language

We will use this on the job's Looker dashboard to visualize the language distribution in our search data, and to create an alarm if changes in the language distribution indicate that our constituency for search features might be changing.

Checklist for reviewer:

### DIFF
--- a/sql/moz-fx-data-shared-prod/search_terms/sanitization_job_languages/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/search_terms/sanitization_job_languages/metadata.yaml
@@ -1,0 +1,13 @@
+---
+friendly_name: Sanitization Job Languages
+description: |-
+  How many search terms got each language tag, by job.
+
+  We use this + other metrics to assess
+  whether the constituency using our search features is changing.
+owners:
+  - ctroy@mozilla.com
+workgroup_access:
+  - role: roles/bigquery.dataViewer
+    members:
+      - workgroup:search-terms/aggregated
--- a/sql/moz-fx-data-shared-prod/search_terms/sanitization_job_languages/view.sql
+++ b/sql/moz-fx-data-shared-prod/search_terms/sanitization_job_languages/view.sql
@@ -1,0 +1,11 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.search_terms.sanitization_job_languages`
+AS
+SELECT
+  started_at as job_start_time,
+  lang.key AS language_code,
+  CAST(lang.value AS int) AS search_term_count
+FROM
+  `moz-fx-data-shared-prod.search_terms_derived.sanitization_job_metadata`
+CROSS JOIN
+  UNNEST(mozfun.json.js_extract_string_map(approximate_language_proportions_json)) AS lang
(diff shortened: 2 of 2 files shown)


sql/moz-fx-data-shared-prod/search_terms/sanitization_job_languages/metadata.yaml
```diff
@@ -0,0 +1,13 @@
...
+  How many search terms got each language tag, by job.
+
+  We use this + other metrics to assess
+  whether the constituency using our search features is changing.
+owners:
+  - ctroy@mozilla.com
+workgroup_access:
+  - role: roles/bigquery.dataViewer
+    members:
+      - workgroup:search-terms/aggregated
```
jklukas, line 13: We've had some discussions in the past about whether to apply access restrictions to virtual views. We could technically have this view be accessible to all of Mozilla since queries will still throw errors if the user doesn't have access to the underlying table.

But having the access enforced at the view level as well is logical and somewhat self-documenting. So I like it this way, but cc'ing @whd as to whether he has a strong opinion on it.

sql/moz-fx-data-shared-prod/search_terms/sanitization_job_languages/view.sql
```diff
@@ -0,0 +1,11 @@
...
+  `moz-fx-data-shared-prod.search_terms.sanitization_job_languages`
+AS
+SELECT
+  started_at as job_start_time,
+  lang.key AS language_code,
+  CAST(lang.value AS int) AS search_term_count
+FROM
+  `moz-fx-data-shared-prod.search_terms_derived.sanitization_job_metadata`
+CROSS JOIN
+  UNNEST(mozfun.json.js_extract_string_map(approximate_language_proportions_json)) AS lang
```
jklukas, line 11: Per https://mozilla.github.io/bigquery-etl/mozfun/json/#js_extract_string_map-udf this is a JS UDF that will not have great performance, but I don't expect that to be a significant issue for the scale of data we have here. If were to become a problem, then we can consider materializing this logic to a derived table and updating the view to point there instead.

sql/moz-fx-data-shared-prod/search_terms/sanitization_job_languages/view.sql
```diff
@@ -0,0 +1,11 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.search_terms.sanitization_job_languages`
+AS
+SELECT
+  started_at as job_start_time,
+  lang.key AS language_code,
+  CAST(lang.value AS int) AS search_term_count
```
jklukas, line 7:
```suggestion
  SAFE_CAST(lang.value AS int) AS search_term_count
```

In case `lang.value` ever has a non-int value.

jklukas (approved):
Looks great!

I just left one question for @whd and a suggestion to use a safe cast.

sql/moz-fx-data-shared-prod/search_terms/sanitization_job_languages/view.sql
chelseatroy: ^^ Definitely. Because we're imagining this running daily on a dictionary with maybe two dozen keys in it, I figured it was okay for now. But we could totally do that.

Another option is to insert these sums into this new table from the job that creates the JSON in the first place.

jklukas:
I think all this needs is a `./bqetl format sql/` and it should be good to go. Views are automatically deployed within ~30 minutes of merging the PR, although as @whd mentions DSRE will need to perform the ACL update.

PR #3093, opened 2022-07-13
Calculate VPN promotion discounts

sean-rose:
Checklist for reviewer:


sql/moz-fx-data-shared-prod/subscription_platform/nonprod_stripe_subscriptions/view.sql
```diff
@@ -77,10 +77,16 @@ invoices_provider_country AS (
...
     invoices.status = "paid"
 ),
-subscriptions_promotion_codes AS (
+subscriptions_promotions AS (
   SELECT
     invoices.subscription_id,
     ARRAY_AGG(DISTINCT promotion_codes.code IGNORE NULLS) AS promotion_codes,
+    SUM(
+      COALESCE(coupons.amount_off, 0) + COALESCE(
+        CAST((invoices.subtotal * coupons.percent_off / 100) AS INT64),
```
scholtzan, line 86: Not sure if this should be a `SAFE_CAST` in case any of these values are null?

sql/moz-fx-data-shared-prod/subscription_platform/stripe_subscriptions/view.sql
```diff
@@ -85,10 +85,16 @@ invoices_provider_country AS (
...
     invoices.status = "paid"
 ),
-subscriptions_promotion_codes AS (
+subscriptions_promotions AS (
   SELECT
     invoices.subscription_id,
     ARRAY_AGG(DISTINCT promotion_codes.code IGNORE NULLS) AS promotion_codes,
+    SUM(
+      COALESCE(coupons.amount_off, 0) + COALESCE(
+        CAST((invoices.subtotal * coupons.percent_off / 100) AS INT64),
```
scholtzan, line 94: Same here, maybe?

scholtzan (approved):
The table schema will need to get updated as well when merging to add the new field.

sql/moz-fx-data-shared-prod/subscription_platform/nonprod_stripe_subscriptions/view.sql
sean-rose: Nulls are fine with normal `CAST()`.  What would be an issue is if there were any incompatible values like non-numeric strings, but here all the values are already numbers and this is just to avoid the calculation resulting in floats.

sean-rose:
> @scholtzan 
> The table schema will need to get updated as well when merging to add the new field.

This does update `all_subscriptions_v1/schema.yaml`, and the `all_subscriptions_v1` ETL recreates the entire table every time, so I don't think any manual schema changes will be necessary.  Or is there something I'm missing?

scholtzan:
As long as you deployed the new schemas via `bqetl query schema deploy ...` nothing else needs to be done

PR #3105, opened 2022-07-22
Move run query scripts to CLI

scholtzan:
This PR moves the old `script/run_query` and `script/run_multipart_query` functionality into the CLI. This change adds two new CLI commands:
```
./bqetl query run
./bqetl query run-multipart
```
The code has been copied from the existing scripts.

Requires https://github.com/mozilla/telemetry-airflow/pull/1529 to be merged as well.


bigquery_etl/util/common.py
```diff
@@ -80,3 +84,40 @@ def write_sql(output_dir, full_table_id, basename, sql):
     with target.open("w") as f:
         f.write(sql)
         f.write("\n")
+
+
+class TempDatasetReference(bigquery.DatasetReference):
```
scholtzan, line 89: Moving to `common` from `standard_args` since `standard_args` will eventually get removed once all scripts have been migrated to CLI (which will use click instead for argparse)

akkomar (approved):
👍

PR #3113, opened 2022-07-25
Normalize search engine for newtab interactions

ANich:
--title

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry/newtab_interactions/view.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry/newtab_interactions/view.sql
@@ -2,7 +2,7 @@ CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.telemetry.newtab_interactions`
 AS
 SELECT
-   * EXCEPT (search_engine),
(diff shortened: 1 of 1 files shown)


scholtzan (approved):

sql/moz-fx-data-shared-prod/telemetry/newtab_interactions/view.sql
```diff
@@ -2,6 +2,7 @@ CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.telemetry.newtab_interactions`
 AS
 SELECT
-  *
+  * EXCEPT (search_engine),
+  `moz-fx-data-shared-prod`.udf.normalize_search_engine(search_engine) AS search_engine,
```
sean-rose, line 6: I know this has already been approved & merged, but I had some thoughts after taking a peek:
- `REPLACE` is intended for situations like this, and would preserve the column's position in the schema.
- Wouldn't it be better to do this in the table ETL? Always doing it on the fly will affect query performance, especially if there's any filtering or grouping on this column.
- Should the same normalization also be applied to the `default_private_search_engine` and `default_search_engine` columns?

Issue #2878, opened 2022-04-07
Views deploy ImportError: failed to import java class via jni, please build java dependencies with: mvn package


whd:
This failed for the most recent views deploy, using the private GCR container.

/CC @ANich (most recent code change) and @relud (I vaguely recall something mvn related landing this week)

```
multiprocess.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/app/bigquery_etl/dependency.py", line 35, in extract_table_references
    ZetaSqlHelper = jnius.autoclass("com.mozilla.telemetry.ZetaSqlHelper")
  File "/usr/local/lib/python3.8/site-packages/jnius/reflect.py", line 211, in autoclass
    c = find_javaclass(clsname)
  File "jnius/jnius_export_func.pxi", line 26, in jnius.find_javaclass
  File "jnius/jnius_utils.pxi", line 91, in jnius.check_exception
jnius.JavaException: JVM exception occurred: com.google.zetasql.ClientChannelProvider: Provider com.google.zetasql.JniChannelProvider could not be instantiated java.util.ServiceConfigurationError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/multiprocess/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/usr/local/lib/python3.8/site-packages/multiprocess/pool.py", line 48, in mapstar
    return list(map(*args))
  File "/usr/local/lib/python3.8/site-packages/pathos/helpers/mp_helper.py", line 15, in <lambda>
    func = lambda args: f(*args)
  File "/app/sql_generators/derived_view_schemas/__init__.py", line 94, in _generate_view_schema
    reference_path = _get_reference_dir_path(view_directory)
  File "/app/sql_generators/derived_view_schemas/__init__.py", line 41, in _get_reference_dir_path
    view_references = extract_table_references(view_file.read_text())
  File "/app/bigquery_etl/dependency.py", line 38, in extract_table_references
    raise ImportError(
ImportError: failed to import java class via jni, please build java dependencies with: mvn package
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/app/bigquery_etl/cli/__init__.py", line 52, in cli
    group(prog_name=prog_name)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1130, in __call__
[...]

PR #3119, opened 2022-07-28
Data Validation Derived View

chelseatroy:
### This PR creates a derived view that runs daily.

The view requires tow input tables:

1. the search sanitization metadata table
2. the table that records the detected languages for each sanitization job run. 

The output table contains aggregate information describing what people searched for (total number of characters, total number of words, and total number of probably-in-English search terms versus unknown or probably-some-other-language, for example). 

We will check how these proportions vary over time to identify if our search volume is _changing_ such that the assumptions about it that we used to develop our sanitization strategy no longer apply.

**Example:** Right now, only U.S and Canada-based constituents can opt into providing their search data. Since most addresses in the U.S. and Canada contain numbers, our sanitization strategy of excluding all terms containing numbers is likely to remove all addresses. Suppose we start allowing constituents from other places to opt into providing their search data and that assumption stops holding: we might need to reevaluate our sanitization criteria to make sure we are still excluding all addresses from sanitized data.

This derived view will recalculate the whole data validation metrics table, but it's done for a job that runs daily and produces one row per run. So it's a table that grows extremely slowly (it'd take more than 25 years to hit 10,000 rows, and the calculations are all two-column arithmetic). The `sanitization_job_languages` view takes the same approach.

Checklist for reviewer:

### DIFF
--- a/sql/moz-fx-data-shared-prod/search_terms/sanitization_job_data_validation_metrics/view.sql
+++ b/sql/moz-fx-data-shared-prod/search_terms/sanitization_job_data_validation_metrics/view.sql
@@ -1,0 +1,26 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.search_terms.sanitization_job_data_validation_metrics`
+AS
+SELECT
+        finished_at,
+        total_search_terms_removed_by_sanitization_job / total_search_terms_analyzed AS pct_sanitized_search_terms,
+        contained_at / total_search_terms_analyzed AS pct_sanitized_contained_at,
+        contained_numbers / total_search_terms_analyzed AS pct_sanitized_contained_numbers,
+        contained_name / total_search_terms_analyzed AS pct_sanitized_contained_name,
+        sum_terms_containing_us_census_surname / total_search_terms_analyzed AS pct_terms_containing_us_census_surname,
+        sum_uppercase_chars_all_search_terms / sum_chars_all_search_terms AS pct_uppercase_chars_all_search_terms,
+        sum_words_all_search_terms / total_search_terms_analyzed AS avg_words_all_search_terms,
+        1 - languages.english_count / languages.all_languages_count AS pct_terms_non_english
+        FROM `moz-fx-data-shared-prod.search_terms_derived.sanitization_job_metadata_v2` AS metadata
+    JOIN
+    (
+        SELECT
+            job_start_time,
+            max(case when language_code = 'en' then search_term_count end) english_count,
+            sum(search_term_count) as all_languages_count,
+        FROM `moz-fx-data-shared-prod.search_terms.sanitization_job_languages`
+        GROUP BY job_start_time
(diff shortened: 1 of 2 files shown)


whd:
This view appears to represent a data classification change from `workgroup:search-terms/sanitized` to `workgroup:search-terms/aggregated`:

https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/search_terms_derived/sanitization_job_metadata_v2/metadata.yaml#L18
https://github.com/mozilla/bigquery-etl/pull/3119/files#diff-04a153ee15179fa859f25d246bbf1027057c32dd3cb70de5af3e6145ed7077e4R14

Therefore it's probably subject to the change control procedure even though it's a view. Please file a ticket like https://mozilla-hub.atlassian.net/browse/DSRE-899 for Mark to approve before we land this.

sql/moz-fx-data-shared-prod/search_terms/sanitization_job_data_validation_metrics/view.sql
```diff
@@ -0,0 +1,26 @@
...
+SELECT
+        finished_at,
+        total_search_terms_removed_by_sanitization_job / total_search_terms_analyzed AS pct_sanitized_search_terms,
+        contained_at / total_search_terms_analyzed AS pct_sanitized_contained_at,
+        contained_numbers / total_search_terms_analyzed AS pct_sanitized_contained_numbers,
+        contained_name / total_search_terms_analyzed AS pct_sanitized_contained_name,
+        sum_terms_containing_us_census_surname / total_search_terms_analyzed AS pct_terms_containing_us_census_surname,
+        sum_uppercase_chars_all_search_terms / sum_chars_all_search_terms AS pct_uppercase_chars_all_search_terms,
+        sum_words_all_search_terms / total_search_terms_analyzed AS avg_words_all_search_terms,
+        1 - languages.english_count / languages.all_languages_count AS pct_terms_non_english
```
scholtzan, line 13: If any of the divisors can be 0 then I would recommend using `SAFE_DIVE` (to cast to null) otherwise querying the view will fail

scholtzan (approved):

scholtzan:
You'll also need to run `./bqetl format` to fix the formatting issues

sql/moz-fx-data-shared-prod/search_terms/sanitization_job_data_validation_metrics/view.sql
chelseatroy: Good call. Thank you! CHange applied.

chelseatroy:
@whd I think we're ready here: https://mozilla-hub.atlassian.net/browse/DSRE-929

whd (approved):

whd:
This might not deploy automatically due to https://github.com/mozilla/bigquery-etl/issues/2878, but I will apply the table ACL once it's been created

PR #3124, opened 2022-07-29
Add derived tables to table size monitoring stats 

alekhyamoz:
This should resolve https://bugzilla.mozilla.org/show_bug.cgi?id=1779053. 
There were 2 approaches to solve this: 
1. To modify the existing folder to something like stable_and_derived_table_sizes_v1  and have both stats (derived and  stable) in the same o/p table. Modify the query.sql accordingly. Single airflow task.
2. Add the similar query.sql and metadata.yaml for the derived table (as stable tables) in different folders to create a new table in BQ. Add a new airflow task

Chose approach 2 to avoid duplication of query.py script.


sql/moz-fx-data-shared-prod/monitoring/stable_table_sizes/view.sql
```diff
@@ -4,4 +4,4 @@ AS
 SELECT
   *
 FROM
-  `moz-fx-data-shared-prod.monitoring_derived.stable_table_sizes_v1`
+  `moz-fx-data-shared-prod.monitoring_derived.stable_and_derived_table_sizes_v1`
```
scholtzan, line 7: It would make sense for this view (`stable_table_sizes`) to only explores stable tables and then add a new view `stable_and_derived_table_sizes` which will return both. Once added the `stable_and_derived_table_sizes` is what we want the Looker dashboard to use.

scholtzan, line 7: I would do this here
```suggestion
  `moz-fx-data-shared-prod.monitoring_derived.stable_and_derived_table_sizes_v1`
  WHERE destination_dataset LIKE '%_stable'
```

And a new view in a separate file in `monitoring/stable_and_derived_table_sizes/view.sql`

```
SELECT
  *
FROM
  `moz-fx-data-shared-prod.monitoring_derived.stable_and_derived_table_sizes_v1`
```

scholtzan (approved):

PR #3141, opened 2022-08-11
Fix the error caused due to date filter (submission_date vs submission_timestamp)

alekhyamoz:
Resolves the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1784105) which is causing the `monitoring_derived__stable_and_derived_table_sizes__v1 ` task to fail in airflow. 
Error logs are available [here](https://workflow.telemetry.mozilla.org/log?dag_id=bqetl_monitoring&task_id=monitoring_derived__stable_and_derived_table_sizes__v1&execution_date=2022-08-10T02%3A00%3A00%2B00%3A00)


edugfilho (approved):
lgtm!

PR #3142, opened 2022-08-11
Recursively look for upstream date partition offsets (bug 1784372)

sean-rose:
`telemetry_derived.rolling_cohorts_v1` and `telemetry_derived.cohort_daily_statistics_v1` have been missing data after 2022-07-04 ([bug 1784372](https://bugzilla.mozilla.org/show_bug.cgi?id=1784372)), which turns out to be due to the `bqetl_unified.telemetry_derived__rolling_cohorts__v1` task not getting the same `date_partition_offset` as its upstream tables in #3069.

This PR changes the `Task.with_upstream_dependencies()` logic to recursively look for upstream tasks with `date_partition_offset`, so we don't have to do manual config workarounds when upstream dependencies of upstream dependencies have a `date_partition_offset`.

---
Checklist for reviewer:


sql/moz-fx-data-shared-prod/telemetry_derived/active_users_aggregates_attribution_v1/metadata.yaml
```diff
@@ -15,7 +15,6 @@ labels:
 scheduling:
   dag_name: bqetl_analytics_aggregations
   task_name: active_users_aggregates_attribution_v1
-  date_partition_offset: -1
```
sean-rose, line 18: These `bqetl_analytics_aggregations` tasks still have the same `date_partition_offset` (`/dags/bqetl_analytics_aggregations.py` isn't modified), but that setting is now determined automatically based on the recursive upstream dependencies.

dags/bqetl_unified.py
```diff
@@ -101,6 +102,13 @@
     with TaskGroup(
         "telemetry_derived__unified_metrics__v1_external"
     ) as telemetry_derived__unified_metrics__v1_external:
+        ExternalTaskMarker(
+            task_id="bqetl_analytics_aggregations__wait_for_telemetry_derived__unified_metrics__v1",
+            external_dag_id="bqetl_analytics_aggregations",
+            external_task_id="wait_for_telemetry_derived__unified_metrics__v1",
+            execution_date="{{ (execution_date - macros.timedelta(seconds=7200)).isoformat() }}",
+        )
```
sean-rose, line 110: I'm unclear why this `ExternalTaskMarker` (which looks correct) is only now getting generated, as I don't think the changes I made should make a difference with that.

sean-rose:
I went ahead and resolved [bug 1784372](https://bugzilla.mozilla.org/show_bug.cgi?id=1784372) with a manual `date_partition_offset` config workaround in #3146 to get that done today.  I'll update this PR to remove that `date_partition_offset` config workaround when this is merged.

dags/bqetl_unified.py
scholtzan: mmh that is odd...

scholtzan (approved):

dags/bqetl_unified.py
sean-rose: I managed to figure this out:  it's due to [using `TaskRef` objects as dictionary keys in `DagCollection.downstream_tasks`](https://github.com/mozilla/bigquery-etl/blob/332c569bed4be3c3896c8cd6daa0d40ed7acc83f/bigquery_etl/query_scheduling/dag_collection.py#L37-L40) and ending up with a different `TaskRef` after the task has its `date_partition_offset` set (and that was exacerbated by the [upstream dependencies being calculated in parallel](https://github.com/mozilla/bigquery-etl/blob/332c569bed4be3c3896c8cd6daa0d40ed7acc83f/bigquery_etl/query_scheduling/dag_collection.py#L161-L162)).

Also, while doing test runs with/without [parallelization of calculating upstream & downstream dependencies](https://github.com/mozilla/bigquery-etl/blob/332c569bed4be3c3896c8cd6daa0d40ed7acc83f/bigquery_etl/query_scheduling/dag_collection.py#L161-L165) I noticed without parallelization is actually a bit faster and lets you see the log messages.

I'm going to make some changes to address what I found and re-request review.

bigquery_etl/query_scheduling/dag_collection.py
```diff
@@ -121,6 +101,24 @@ def with_tasks(self, tasks):
...
+                    _task.with_upstream_dependencies(self)
+                    for upstream_dependency in (
+                        _task.depends_on + _task.upstream_dependencies
+                    ):
+                        downstream_dependencies[upstream_dependency.task_key].append(
+                            _task.to_ref(self)
+                        )
+            self._downstream_dependencies = downstream_dependencies
+
+        return self._downstream_dependencies[task.task_key]
```
sean-rose, line 120: I changed this from a dictionary property to a method so that callers don't have to know about the implementation detail of what dictionary key is being used for the cache.

scholtzan (approved):

PR #3145, opened 2022-08-12
Hoist schemas in SQL tests up to table dir

ANich:
This is one step towards having only one source of truth for a table's schema. 

(Note the only diffs are (1) the change to `bigquery_etl/pytest_plugin/sql_test.py` and (2) moving schema files and deleting duplicates)

For SQL tests we currently define the schema for each input table in each test. This PR changes the pytest plugin to detect the schema for each input table test in the table's directory rather than the test directory. This cuts down on the duplication necessary for test-changing PR's (and hopefully will encourage more tests).

```
Before:

tests/
  └ ...            
    └ table
      └ test_1/     
        └ input_table.schema.json (or yaml)         
        └ input_table.json(or yaml)       
      └ test_2/
        └ input_table.schema.json (or yaml)         
        └ input_table.json(or yaml)       

After:

tests/
  └ ...            
    └ table
      └ input_table.schema.json (or yaml)
      └ test_1/     
        └ input_table.json(or yaml)       
      └ test_2/
        └ input_table.json(or yaml)  
```


fbertsch (changes requested):
This is a nice change, but there might be a little additional work to do! The GLAM tests are are generated, and if you run [this command](https://github.com/mozilla/bigquery-etl/tree/main/tests/sql/glam-fenix-dev/glam_etl#usage) you'll see that it updates all the schemas again. It might be worth getting someone who knows more about GLAM than me to take a look after that's done.

We should also update the docs here: https://mozilla.github.io/bigquery-etl/cookbooks/testing/

ANich:
> This is a nice change, but there might be a little additional work to do! The GLAM tests are are generated, and if you run [this command](https://github.com/mozilla/bigquery-etl/tree/main/tests/sql/glam-fenix-dev/glam_etl#usage) you'll see that it updates all the schemas again. It might be worth getting someone who knows more about GLAM than me to take a look after that's done.

I think this script is slightly outdated, it generates additional and changed files on main as well. I don't think it was meant to be kept up to date though, given that the doc paragraph says "It is not expected to be run again." Not sure it's worth bringing this up to date, what do you think @akkomar @alekhyamoz?

> We should also update the docs here: https://mozilla.github.io/bigquery-etl/cookbooks/testing/

Updated in 0642e2dc1418e89dec72a51d20498a4918089cf4

akkomar:
I'm not really too familiar with these tests and I'd be careful with losing the option to regenerate them if there are scripts for that. @alekhyamoz might know more.

Can this script be updated to generate schemas in line with the new convention?

I can look into this soon-ish if needed (have some other stuff on my plate now).

fbertsch:
> I think this script is slightly outdated, it generates additional and changed files on main as well. I don't think it was meant to be kept up to date though, given that the doc paragraph says "It is not expected to be run again." Not sure it's worth bringing this up to date, what do you think @akkomar @alekhyamoz?

I _believe_ that is referring to the `bootstrap.py`, but what's running in the test cmd is [this](https://github.com/mozilla/bigquery-etl/blob/main/bigquery_etl/glam/cli.py#L187) (with some scripting around it). It looks to me like it was meant to be kept up-to-date since it has e.g. testing, a cli, etc.

fbertsch (approved):
Thinking it through, this will be at least backwards compatible with any test generation, since any generation that worked previously will continue to work.

PR #3152, opened 2022-08-17
Add desktop mobile monthly search agg

alekhyamoz:
The [Looker dashboards](https://mozilla.cloud.looker.com/dashboards/430) are very slow and some are broken as each tile processes around 20k GB of data.

As part of an effort to resolve the issue, it was decided to create a derived table in BQ with all relevant dimensions and measures. The table will be populated with the previous month's aggregated data at the beginning of every month.


sql/moz-fx-data-shared-prod/search_derived/desktop_mobile_monthly_search_aggregates_v1/query.sql
```diff
@@ -0,0 +1,60 @@
+-- Query for search_derived.desktop_mobile_monthly_search_aggregates_v1
```
ANich, line 1: Nit: I'd call this something like `desktop_mobile_search_clients_monthly` since it's per-client per-month. The "aggregates" language I think is used elsewhere for non-client level data

sql/moz-fx-data-shared-prod/search_derived/desktop_mobile_monthly_search_aggregates_v1/metadata.yaml
```diff
@@ -0,0 +1,13 @@
+friendly_name: Desktop and Mobile Monthly Search Aggregates
```
ANich, line 1: Nit: see below comment on naming, this might cause some confusion

sql/moz-fx-data-shared-prod/search_derived/desktop_mobile_monthly_search_aggregates_v1/query.sql
```diff
@@ -0,0 +1,60 @@
...
+  COALESCE(SUM(ad_click), 0) AS ad_click,
+  COALESCE(SUM(tagged_follow_on), 0) AS tagged_follow_on,
+  COALESCE(SUM(tagged_sap), 0) AS tagged_sap,
+FROM
+  search.mobile_search_clients_engines_sources_daily
+WHERE
+  submission_date
+  BETWEEN date_trunc(date_sub(@submission_date, INTERVAL 1 month), month)
+  AND last_day(date_sub(@submission_date, INTERVAL 1 month), month)
+GROUP BY
```
fbertsch, line 24: Can these columns be explicitly named? e.g. `client_id` rather than `1`

fbertsch (changes requested):
If we partition by `month` and have the job overwrite the `month` partition, then that would make this job idempotent. I'm trying to remember of a bigquery_etl query will, by default, overwrite a partition, or append to it

sql/moz-fx-data-shared-prod/search_derived/desktop_mobile_monthly_search_aggregates_v1/query.sql
```diff
@@ -0,0 +1,60 @@
...
+  1,
+  2,
+  3,
+  4,
+  5,
+  6,
+  7
```
ANich, line 31: I'd suggest naming the GROUP BY columns here and below
```suggestion
client_id,
month,
device,
normalized_engine,
normalized_app_name,
os,
country
```

sql/moz-fx-data-shared-prod/search_derived/desktop_mobile_monthly_search_aggregates_v1/query.sql
```diff
@@ -0,0 +1,60 @@
...
```
ANich, line 60: Similar to above
```suggestion
client_id,
month,
device,
normalized_engine,
normalized_app_name,
os,
country
```

sql/moz-fx-data-shared-prod/search_derived/desktop_mobile_monthly_search_aggregates_v1/query.sql
```diff
@@ -0,0 +1,60 @@
...
+  COALESCE(SUM(search_with_ads), 0) AS search_with_ads,
+  COALESCE(SUM(ad_click), 0) AS ad_click,
+  COALESCE(SUM(tagged_follow_on), 0) AS tagged_follow_on,
+  COALESCE(SUM(tagged_sap), 0) AS tagged_sap,
+FROM
+  search.mobile_search_clients_engines_sources_daily
+WHERE
+  submission_date
+  BETWEEN date_trunc(date_sub(@submission_date, INTERVAL 1 month), month)
+  AND last_day(date_sub(@submission_date, INTERVAL 1 month), month)
```
ANich, line 23: (TIL `last_day` is a thing)

dags/bqetl_desktop_mobile_search_aggregates_monthly.py
```diff
@@ -0,0 +1,87 @@
...
+        email=["akommasani@mozilla.com", "telemetry-alerts@mozilla.com"],
+        date_partition_parameter="submission_date",
+        depends_on_past=False,
+    )
+
+    wait_for_search_derived__mobile_search_clients_daily__v1 = ExternalTaskSensor(
+        task_id="wait_for_search_derived__mobile_search_clients_daily__v1",
+        external_dag_id="bqetl_mobile_search",
+        external_task_id="search_derived__mobile_search_clients_daily__v1",
+        execution_delta=datetime.timedelta(seconds=79200),
```
fbertsch, line 62: Do we need these task sensors?

alekhyamoz: These were added by default while generating the dag, believe these are required for the search_daily table to be populated (dependency)

dags.yaml
```diff
@@ -786,3 +786,23 @@ bqetl_newtab:
...
+    - akommasani@mozilla.com
+    email_on_failure: true
+    email_on_retry: true
+    end_date: null
+    owner: akommasani@mozilla.com
+    retries: 2
+    retry_delay: 30m
+    start_date: '2019-01-01'
+  description: Populate monthly aggregates based on daily clients table for desktop
+  schedule_interval: monthly
```
ANich, line 804: I believe this translates to 0 0 1 * * for 2nd of the month, seems like this should be:
```suggestion
  schedule_interval: 0 0 2 * *
```

alekhyamoz: I should have updated it here as well. But in the .py folder I set it to something like `schedule_interval="0 5 2 * *"` 5 am to allow time for upstream to run

dags/bqetl_desktop_mobile_search_aggregates_monthly.py
fbertsch: Hmm, they're not really doing much since we would actually want the entire past month to be complete. It might be possible to remove them by adding an empty `referenced_tables`

fbertsch (approved):
LGTM with removing the task dependencies

PR #3170, opened 2022-08-19
Allow dry runs from mozdata

scholtzan:
This will allow dry runs from `mozdata`, like:

`./bqetl query validate --use-cloud-function=false --project-id=mozdata moz-fx-data-shared-prod.contextual_services.event_aggregates`

For example, DS don't have permissions for running queries in moz-fx-data-shared-prod, so they need to use mozdata.


alekhyamoz (approved):
lgtm

Issue #3080, opened 2022-07-05
Split bigquery-etl queries and tooling


scholtzan:
Proposal: https://docs.google.com/document/d/1I7_z578Fsq6CAMVBaIqCBTLFSNKhrzeAja2BthqN1HY/edit

To implement the proposal and separate out bqetl as a separate tool we'll need to:
- [x] Remove `stored_procedures/` (https://github.com/mozilla/bigquery-etl/pull/3081)
- [ ] Remove the remaining scripts and move the functionality to the CLI tool (https://github.com/mozilla/bigquery-etl/pull/3083)
  - make sure that cloudops-infra doesn't reference any scripts
  - see https://github.com/mozilla/bigquery-etl/issues/1648
- [x] ~Create a `bqetl` repository and copy the tooling logic~
- [ ] Setup bqetl to get automatically published to pypi
- [ ] Update CI etc in private-bigquery-etl and bigquery-etl to use `bqetl` package
- [x] ~remove `bigquery_etl/` from the `bigquery-etl` repo~
- [ ] Add support to bqetl to generate CircleCI config
- [ ] Update bigquery-etl and private-bigquery-etl to pull in shared/generated CircleCI config
- [ ] Document workflow for how to have a local build of bqetl and coordinate that with new ETL definitions you're making in bigquery-etl
- [ ] Update docs

scholtzan:
Instead of moving the tooling into a separate repo, it is likely more convenient to keep in the bigquery-etl repo. Often, adding new functionality for new queries requires changes to the tooling. Having to update two places will get quite annoying.

We still want to separate out the tooling and publish it as a package though to make it reusable across different repos.

PR #3174, opened 2022-08-23
Move static publish script to CLI

scholtzan:
Depends on https://github.com/mozilla/telemetry-airflow/pull/1545

Related to https://github.com/mozilla/bigquery-etl/issues/3080

This moves the `publish_static` script into the bqetl CLI. The script logic is left unchanged, but this is necessary for splitting out the bqetl tooling from the query definitions in the future.

### DIFF
--- a/bigquery_etl/cli/__init__.py
+++ b/bigquery_etl/cli/__init__.py
@@ -17,6 +17,7 @@ from ..cli.routine import mozfun, routine
 from ..cli.view import view
 from ..dependency import dependency
 from ..glam.cli import glam
+from ..static import static_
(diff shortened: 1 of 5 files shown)


bigquery_etl/static/__init__.py
```diff
@@ -1,36 +1,66 @@
-"""Publish csv files as BigQuery tables."""
```
ANich, line 1: Pet peeve/nit: not a fan of having logic in `__init__.py` files 😝 would prefer a `static.py` file in this case or another file with a 1-line import here

ANich (approved):
👍

PR #3177, opened 2022-08-24
Prototype dataset authorization metadata

whd:
See https://mozilla-hub.atlassian.net/browse/DSRE-959

This might make sense under `labels` as that's where we put metadata about authorizations [(example)](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/monitoring_derived/topsites_click_rate_live_v1/metadata.yaml#L9) but it's a list so I'm not sure how that would look.

This metadata would currently be unactionable by infrastructure but is useful for humans to reason about. It could therefore potentially be a comment as well [(example)](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/telemetry/buildhub2/view.sql#L1). I suspect we'd eventually want to support this as an alternative to specifically authorized views [DSRE-947] but I'm happy to defer proper codification of this for the implementation in a specific use case.

Checklist for reviewer:

[DSRE-947]: https://mozilla-hub.atlassian.net/browse/DSRE-947?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_accounts/dataset_metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts/dataset_metadata.yaml
@@ -4,6 +4,8 @@ description: |-
 dataset_base_acl: view
 user_facing: true
 labels: {}
+dataset_authorized_by:
+- moz-fx-fxa-prod-0712.fxa_prod_logs
 workgroup_access:
 - role: roles/bigquery.dataViewer
   members:


scholtzan (commented):
I think it makes sense to have this separate from `labels` especially if we might want to make this actionable in the future. `labels` are usually just used as metadata that gets added to the BigQuery table, but doesn't trigger an action.

fbertsch (approved):
+1, agreed with Anna. This seems reasonable

PR #3178, opened 2022-08-24
Fix dataset.get permission check in dryrun

scholtzan:
This is a followup to a previous change to ensure that DS can run `./bqetl query validate` even without having `dataset.get` permissions in shared-prod.

### DIFF
--- a/bigquery_etl/dryrun.py
+++ b/bigquery_etl/dryrun.py
@@ -21,7 +21,6 @@ from urllib.request import Request, urlopen
 
 import click
 from google.cloud import bigquery
-from google.api_core.exceptions import PermissionDenied
 
 from .metadata.parse_metadata import Metadata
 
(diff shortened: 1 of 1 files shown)


bigquery_etl/dryrun.py
```diff
@@ -350,8 +349,15 @@ def dry_run_result(self):
...
-                    dataset_labels = []
+                except Exception as e:
+                    # Most users do not have bigquery.datasets.get permission in
+                    # moz-fx-data-shared-prod
+                    # This should not prevent the dry run from running since the dataset
+                    # labels are usually not required
+                    if "Permission bigquery.datasets.get denied on dataset" in str(e):
+                        dataset_labels = []
+                    else:
+                        raise e
```
scholtzan, line 360: I was able to reproduce this in my sandbox project by creating a service account with limited permissions. Turns out here only a general `Exception` is thrown instead of a `PermissionDenied` exception

relud (approved):

PR #3183, opened 2022-08-26
DENG-179 dataset for investigation of incidents.

lucia-vargas-a:
This PR adds two attribution dimensions to active_users_aggregates_attribution: install_source and adjust_network, for the purpose of facilitating the investigation of incidents.

Health checks ran on subset `moz-fx-data-shared-prod.analysis.lvargas_active_users_aggregates_attribution_v1` in comparison with the current dataset `telemetry_derived.active_users_aggregates_attribution_v1`, checking the counts of DAU, New Profiles, Search counts and distinct values expected for both dimensions. No differences in counts and values as expected. Backfill works correct.

Results validated by stakeholder.

This PR is a medium term solution. For the long term and more robust solution we have this [Jira story](https://mozilla-hub.atlassian.net/browse/DENG-178?atlOrigin=eyJpIjoiM2Q0MzliZTUxOGYyNDI2ZDhmNTc5ZGRkMWQyOWVhMTMiLCJwIjoiaiJ9).

Checklist for reviewer:


sql/moz-fx-data-shared-prod/telemetry_derived/investigation_aggregates_v1/query.sql
```diff
@@ -0,0 +1,256 @@
...
+        'Untrusted Devices',
+        'Product Marketing (Owned media)',
+        'Google Ads ACI'
+      )
+      AND ARRAY_AGG(metrics.string.metrics_adjust_network)[SAFE_OFFSET(0)] IS NOT NULL
+    THEN
+      'Other'
+    ELSE
+      ARRAY_AGG(metrics.string.metrics_adjust_network)[SAFE_OFFSET(0)]
+    END
```
kik-kik, line 199: Maybe to reduce the verbosity we could move this logic into a UDF? Seems like the logic is used multiple times?

sql/moz-fx-data-shared-prod/telemetry_derived/investigation_aggregates_v1/query.sql
```diff
@@ -0,0 +1,256 @@
...
+    AS install_source
+  FROM
+    org_mozilla_firefox.metrics AS org_mozilla_firefox_metrics
+  WHERE
+    DATE(submission_timestamp) >= '2021-01-01'
+  GROUP BY
+    1
+),
+fenix AS (
```
kik-kik, line 220: Is this data specific only to fenix? If that is the case perhaps we should rename the table to reflect that?

lucia-vargas-a: It only contains data for Fenix now, but my understanding is that it is in the board to collect attribution for other browsers.
So my plan is to keep this as a unique aggregate for investigations and add data as it is made available, instead of creating separate datasets or change the models build on top of it if we rename the table later.
Another option is to add the browser_name column already, which would be an unnecessary storage cost but mitigates future changes.

What is your preference?

kik-kik: Investigations as in debugging? I wonder if maybe in that case it should live in a different dataset? I think we should avoiding production tables and debug tables together.

sql/moz-fx-data-shared-prod/telemetry_derived/active_users_aggregates_attribution_v1/query.sql
```diff
@@ -1,16 +1,193 @@
...
+  SELECT
+    base.client_id,
+    base.attributed,
+    base.attribution_campaign,
+    base.attribution_content,
+    base.attribution_experiment,
+    base.attribution_medium,
+    base.attribution_source,
+    base.attribution_variation,
+    normalize_adjust_network(att.adjust_network) AS adjust_network,
```
lucia-vargas-a, line 155: Filtering data first and normalizing with at the end.

sql/moz-fx-data-shared-prod/telemetry_derived/active_users_aggregates_attribution_v1/query.sql
```diff
@@ -1,16 +1,193 @@
 -- Query for telemetry_derived.active_users_aggregates_attribution_v1
+CREATE TEMP FUNCTION normalize_adjust_network(key STRING) AS (
+  (
+    CASE
+    WHEN
+      key NOT IN (
+        '',
```
kik-kik, line 7: Should we be keeping an empty string here?

kik-kik, line 7: Are we sure we want to keep an empty space for the field here?

sql/moz-fx-data-shared-prod/telemetry_derived/active_users_aggregates_attribution_v1/query.sql
```diff
@@ -1,16 +1,193 @@
...
+    DATE(submission_timestamp) >= '2021-01-01'
+    AND metrics.string.metrics_install_source IS NOT NULL
+    AND metrics.string.metrics_adjust_network IS NOT NULL
+  GROUP BY
+    1
+  UNION ALL
+  -- Fenix release
+  SELECT
+    client_info.client_id AS client_id,
+    MIN(submission_timestamp) AS submission_timestamp,
```
kik-kik, line 121: this would always be the lowest value found for `submission_timestamp` in the specific date range used? `submission_timestamp` field name could be misleading here.

sql/moz-fx-data-shared-prod/telemetry_derived/active_users_aggregates_attribution_v1/schema.yaml
```diff
@@ -68,3 +68,9 @@ fields:
 - mode: NULLABLE
   name: attributed
   type: BOOLEAN
+- mode: NULLABLE
```
kik-kik, line 71: maybe it would be defining `description` field here along with the description for any field which might not be self explanatory? For example, `install_source` what does this actually mean?

sql/moz-fx-data-shared-prod/telemetry_derived/active_users_aggregates_attribution_v1/query.sql
lucia-vargas-a: The reason to keep it here is so that it doesn't classify as `Other` which would not be correct.

sql/moz-fx-data-shared-prod/telemetry_derived/active_users_aggregates_attribution_v1/query.sql
lucia-vargas-a: Field name updated to avoid misleading!

sql/moz-fx-data-shared-prod/telemetry_derived/active_users_aggregates_attribution_v1/query.sql
lucia-vargas-a: @kik, I talked to the stakeholder, they want to keep the field empty for analysis.

sql/moz-fx-data-shared-prod/telemetry_derived/active_users_aggregates_attribution_v1/query.sql
```diff
@@ -1,16 +1,193 @@
...
+    ARRAY_AGG(metrics.string.metrics_adjust_network)[SAFE_OFFSET(0)] AS adjust_network,
+    ARRAY_AGG(metrics.string.metrics_install_source)[SAFE_OFFSET(0)] AS install_source
+  FROM
+    org_mozilla_fennec_aurora.metrics AS org_mozilla_fennec_aurora_metrics
+  WHERE
+    DATE(submission_timestamp) >= '2021-01-01'
+    AND metrics.string.metrics_install_source IS NOT NULL
+    AND metrics.string.metrics_adjust_network IS NOT NULL
+  GROUP BY
+    1
```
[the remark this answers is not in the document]
kik-kik: no sorry, I meant using column name rather than just saying `1`

sql/moz-fx-data-shared-prod/telemetry_derived/active_users_aggregates_attribution_v1/query.sql
kik-kik: 👍

kik-kik: Then let's leave it for now, but in general I think it would be good for us as analytics team to have a discussion how we think these kind of values (empty string) should be handled across the org.

sql/moz-fx-data-shared-prod/telemetry_derived/active_users_aggregates_attribution_v1/query.sql
```diff
@@ -1,16 +1,193 @@
...
+  -- Fenix release
+  SELECT
+    client_info.client_id AS client_id,
+    MIN(submission_timestamp) AS min_submission_timestamp,
+    ARRAY_AGG(metrics.string.metrics_adjust_network)[SAFE_OFFSET(0)] AS adjust_network,
+    ARRAY_AGG(metrics.string.metrics_install_source)[SAFE_OFFSET(0)] AS install_source
+  FROM
+    org_mozilla_firefox.metrics AS org_mozilla_firefox_metrics
+  WHERE
+    DATE(submission_timestamp) >= '2021-01-01'
```
kik-kik, line 127: From what I can see this is meant to be an incremental job. Is it always necessary to query data from the beginning of 2021 here? I assume the three fields should remain static?

kik-kik:
@lucia-vargas-a should there also be an `init.sql` file for creating the table along with data prior to the current date?

sql/moz-fx-data-shared-prod/telemetry_derived/active_users_aggregates_attribution_v1/query.sql
lucia-vargas-a: It is necessary, because we need to get the first attribution, unfortunately a variable is not supported here.
I am putting together a plan to add the first attribution to clients_first_seen, for a more robust solution.

kik-kik: I was thinking that maybe it would be sufficient to check if client_id already exists in the target table to decide whether to include it or not if there is an entry in the current submission_date.

A bit like we do with `clients_first_seen` logic:
https://github.com/mozilla/bigquery-etl/blob/generated-sql/sql/moz-fx-data-shared-prod/firefox_desktop_background_update_derived/baseline_clients_first_seen_v1/query.sql#L15-L26

But either way, I won't let this be a block for now.

lucia-vargas-a:
> should there also be an init.sql file for creating the table along with data prior to the current date?

The idea is to create the table by deploying the schema, there would be no need for .ini files, unless I am missing something?

kik-kik:
> > should there also be an init.sql file for creating the table along with data prior to the current date?
> 
> The idea is to create the table by deploying the schema, there would be no need for .ini files, unless I am missing something?

`init.sql` is not just to create the table skeleton. The idea is that we can run a single query to create the table until "now" without having to run any additional backfill tasks.

lucia-vargas-a:
> > > should there also be an init.sql file for creating the table along with data prior to the current date?
> > 
> > 
> > The idea is to create the table by deploying the schema, there would be no need for .ini files, unless I am missing something?
> 
> `init.sql` is not just to create the table skeleton. The idea is that we can run a single query to create the table until "now" without having to run any additional backfill tasks.

I see what you mean. The recommendation is to use [the schema and backfill ](https://github.com/mozilla/bigquery-etl/pull/2960/files#r871404413) method,

kik-kik (approved):

PR #3184, opened 2022-08-26
trim whitespace

Marlene-M-Hirose:
Checklist for reviewer:

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/unified_metrics_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/unified_metrics_v1/query.sql
@@ -1,4 +1,18 @@
-WITH unioned_source AS (
+WITH ios_is_default_browser AS (
+SELECT client_info.client_id AS client_id,
+          DATE(submission_timestamp) AS submission_date,
+          SUM(COALESCE(metrics.counter.app_opened_as_default_browser, 0)) > 0 AS is_default_browser,
+
+   FROM org_mozilla_ios_firefox.metrics m
+   WHERE DATE(submission_timestamp) >= DATE(2021, 1, 1)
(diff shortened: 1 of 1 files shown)


Marlene-M-Hirose:
@irrationalagent I'm updating the unified metrics table's Firefox iOS is_default_browser to deal with the update in version 14 where we could see if default_browser was set to ON

irrationalagent:
@Marlene-M-Hirose What is the reasoning here for why we aren't changing the `is_default_browser` field directly in `firefox_ios.clients_last_seen_joined`? 

My concern is that if we change the behavior here but not there, then that could lead to confusion about why the two tables don't align

Marlene-M-Hirose:
> 

@irrationalagent That makes sense.  Lucia and I thought this was a good place to make the changes. Though, I did think it would be make the query more complex than needed.

@lucia-vargas-a , it seems like it is better to make the changes in the firefox_ios.clients_last_seen_joined as Leif suggests. What do you think?

lucia-vargas-a:
Sure I agree that data needs to be consistent!
My reasoning behind recommending adding it to unified_metrics is because I understood this case as marketing-specific logic that could be applied here in a similar way as we find [user segments](https://github.com/mozilla/bigquery-etl/blob/5f1a04335c0d94ff5c5caeb405a2aeeb8b9bee98/sql/moz-fx-data-shared-prod/telemetry_derived/unified_metrics_v1/query.sql#L195), so I'd check if there is inconsistency with the upstream and if the logic applies overall.

Btw, didn't we get this potential difference in the health checks or user validation on the analysis dataset?

Marlene-M-Hirose:
> Btw, didn't we get this potential difference in the health checks or user validation on the analysis dataset?

I think we did.

PR #3201, opened 2022-09-02
bug(1788650): Added dedup step at the end of the query to ensure we only end up with a single entry per client_id

kik-kik:
# bug([1788650](https://bugzilla.mozilla.org/show_bug.cgi?id=1788650)): Added dedup step at the end of the query to ensure we only end up with a single entry per client_id

We started seeing a large number of duplicates inside one of our datasets which started to cause ETL failures. This should ensure that this specific table only has a single entry per user.

Checklist for reviewer:

### DIFF
--- a/sql_generators/glean_usage/templates/baseline_clients_first_seen_v1.init.sql
+++ b/sql_generators/glean_usage/templates/baseline_clients_first_seen_v1.init.sql
@@ -34,13 +34,12 @@ WITH
   {{ core_clients_first_seen(migration_table) }}
   SELECT
     client_id,
-    MIN(submission_date) AS submission_date,
-    MIN(COALESCE(core.first_seen_date, baseline.first_seen_date)) AS first_seen_date,
+    submission_date,
+    COALESCE(core.first_seen_date, baseline.first_seen_date) AS first_seen_date,
     sample_id,
   FROM baseline
-  LEFT JOIN _core_clients_first_seen core
+  LEFT JOIN _core_clients_first_seen AS core
   USING (client_id)
-  GROUP BY client_id, sample_id
 {% else %}
   SELECT * FROM baseline
 {% endif %}
--- a/sql_generators/glean_usage/templates/baseline_clients_first_seen_v1.init.sql
+++ b/sql_generators/glean_usage/templates/baseline_clients_first_seen_v1.init.sql
@@ -34,12 +34,13 @@ WITH
   {{ core_clients_first_seen(migration_table) }}
   SELECT
     client_id,
-    submission_date,
-    COALESCE(core.first_seen_date, baseline.first_seen_date) as first_seen_date,
-    sample_id
+    MIN(submission_date) AS submission_date,
+    MIN(COALESCE(core.first_seen_date, baseline.first_seen_date)) AS first_seen_date,
+    sample_id,
   FROM baseline
   LEFT JOIN _core_clients_first_seen core
   USING (client_id)
+  GROUP BY client_id, sample_id
 {% else %}
   SELECT * FROM baseline
 {% endif %}
--- a/sql_generators/glean_usage/templates/baseline_clients_first_seen_v1.query.sql
+++ b/sql_generators/glean_usage/templates/baseline_clients_first_seen_v1.query.sql
@@ -95,8 +95,8 @@ _previous AS (
 
 -- added this as the result of bug#1788650
 SELECT
-  MIN(submission_date) AS submission_date,
-  MIN(first_seen_date) AS first_seen_date,
+  submission_date,
+  first_seen_date,
   sample_id,
   client_id
 FROM _joined
--- a/sql_generators/glean_usage/templates/baseline_clients_first_seen_v1.query.sql
+++ b/sql_generators/glean_usage/templates/baseline_clients_first_seen_v1.query.sql
@@ -76,17 +76,28 @@ _previous AS (
     AND first_seen_date < @submission_date
 )
 {% endif %}
-  --
-SELECT
-  IF(
-    _previous.client_id IS NULL
-    OR _previous.first_seen_date >= _current.first_seen_date,
-    _current,
+
+, _joined AS (
+  SELECT
+    IF(
+      _previous.client_id IS NULL
+      OR _previous.first_seen_date >= _current.first_seen_date,
+      _current,
+      _previous
+    ).*
+  FROM
+    _current
+  FULL JOIN
     _previous
-  ).*
-FROM
-  _current
-FULL JOIN
-  _previous
-USING
-  (client_id)
+  USING
+    (client_id)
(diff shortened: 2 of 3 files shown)


kik-kik:
@fbertsch alternatively I could do this:
```
SELECT
  MIN(submission_date),
  MIN(first_seen_date),
  sample_id,
  client_id
FROM _joined
GROUP BY sample_id, client_id
```

kik-kik:
Decided to go with group by as I felt it was easier to read

sql_generators/glean_usage/templates/baseline_clients_first_seen_v1.init.sql
```diff
@@ -34,12 +34,13 @@ WITH
...
-    submission_date,
-    COALESCE(core.first_seen_date, baseline.first_seen_date) as first_seen_date,
-    sample_id
+    MIN(submission_date) AS submission_date,
+    MIN(COALESCE(core.first_seen_date, baseline.first_seen_date)) AS first_seen_date,
+    sample_id,
   FROM baseline
   LEFT JOIN _core_clients_first_seen core
   USING (client_id)
+  GROUP BY client_id, sample_id
```
relud, line 43: isntead of adding this here, we should ensure that we aren't getting duplicates from `_core_clients_first_seen`

relud: what we could do here is have the query error out if there's a duplicate, because the inputs to this join must be unique on client id,

kik-kik: Do you mean we should ensure a single row is returned here:
https://github.com/mozilla/bigquery-etl/blob/0528cbe6c5ba3205e14e9fb2115a9641a260828b/sql_generators/glean_usage/templates/macros.sql#L21-L30
?

kik-kik: @relud could you please share how you'd approach this? I'd like to close down related issue.

relud: yeah, that seems like a reasonable place to put the check

kik-kik:
@relud added qualify in the macro to ensure we only ever have one entry per user. Do you think that works or should we do something different here?

sql_generators/glean_usage/templates/baseline_clients_first_seen_v1.init.sql
relud, line 43: This should not have a `GROUP BY` clause, because both sides of the join are already unique on `client_id` (and `sample_id` is derived from `client_id`)
```suggestion
    submission_date,
    COALESCE(core.first_seen_date, baseline.first_seen_date) AS first_seen_date,
    sample_id,
  FROM baseline
  LEFT JOIN _core_clients_first_seen AS core
  USING (client_id)
```

If we want to ensure uniqueness here we should add a erroring qualify clause as seen in my other comments

sql_generators/glean_usage/templates/macros.sql
```diff
@@ -27,5 +27,7 @@ _core_clients_first_seen AS (
   JOIN
      _core
   ON _fennec_id_lookup.fennec_client_id = _core.client_id
+  QUALIFY
+    ROW_NUMBER() OVER (PARTITION BY client_id ORDER BY first_seen_date ASC) = 1
```
relud, line 31: CI is failing because zetasql requires `QUALIFY` to be used in conjunction with `WHERE` or `HAVING`.

also, per [this comment](https://github.com/mozilla/bigquery-etl/pull/3201/files#r961907636), this should throw an error if `client_id` (old or new) is not unique:

```suggestion
  WHERE
    TRUE
  QUALIFY
    IF(
      COUNT(*) OVER (PARTITION BY _fennec_id_lookup.client_id) > 1
      OR COUNT(*) OVER (PARTITION BY _core.client_id) > 1,
      ERROR("duplicate client_id detected"),
      TRUE
    )
```

sql_generators/glean_usage/templates/baseline_clients_first_seen_v1.query.sql
```diff
@@ -76,17 +76,28 @@ _previous AS (
...
+)
+
+-- added this as the result of bug#1788650
+SELECT
+  MIN(submission_date) AS submission_date,
+  MIN(first_seen_date) AS first_seen_date,
+  sample_id,
+  client_id
+FROM _joined
+GROUP BY sample_id, client_id
```
relud, line 103: this should also throw an error if `client_id` is not unique:

```suggestion
SELECT
  MIN(submission_date) AS submission_date,
  MIN(first_seen_date) AS first_seen_date,
  sample_id,
  client_id
FROM _joined
WHERE
  TRUE
QUALIFY
  IF(
    COUNT(*) OVER (PARTITION BY client_id) > 1,
    ERROR("duplicate client_id detected"),
    TRUE
  )
```

relud (approved):
r+ only if my suggestions are applied

we should throw an error when client id is not unique, rather than silently make it unique, because it means something is broken, probably a source table, and needs to be fixed upstream.

sql_generators/glean_usage/templates/baseline_clients_first_seen_v1.query.sql
kik-kik: this is neat, thanks @relud. You just became my go-to person for SQL questions 😄 

Could you explain the `WHERE TRUE` part?

relud: zetasql requires `QUALIFY` to be used in conjunction with `WHERE` or `HAVING`. BigQuery itself does not care, but we need zetasql to work because we use it to analyze the sql to find what tables are referenced. The `WHERE` clause doesn't need to do anything, it just needs to be present, so we use `WHERE TRUE` as a no-op to satisfy zetasql.

kik-kik: thank you for the explanation! 🙇

sql_generators/glean_usage/templates/baseline_clients_first_seen_v1.query.sql
```diff
@@ -76,17 +76,35 @@ _previous AS (
...
+    _current
+  FULL JOIN
+    _previous
+  USING
+    (client_id)
+)
+
+-- added this as the result of bug#1788650
 SELECT
+  MIN(submission_date) AS submission_date,
```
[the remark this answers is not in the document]
relud: indeed, both here and on the next line

kik-kik:
Just double checking, happy for me to merge @relud ?

relud (approved):

PR #3202, opened 2022-09-02
Fix exception for requiring auth with no default project

relud:
and run isort on touched files

Checklist for reviewer:

### DIFF
--- a/bigquery_etl/cli/utils.py
+++ b/bigquery_etl/cli/utils.py
@@ -1,15 +1,16 @@
 """Utility functions used by the CLI."""
 
-import os
 import fnmatch
+import os
+import re
 from fnmatch import fnmatchcase
 from pathlib import Path
 
 import click
-import re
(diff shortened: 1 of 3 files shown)


bigquery_etl/cli/utils.py
```diff
@@ -33,13 +34,12 @@ def is_valid_file(ctx, param, value):
...
-
-    if project_id:
-        return client.project == project_id
-
+def is_authenticated():
+    """Check if the user is authenticated to GCP."""
+    try:
+        bigquery.Client(project="")
+    except DefaultCredentialsError:
+        return False
```
relud, line 42: this did not actually check if the user could access the project, and was never being called with project_id outside of tests.

additionally, the way it was being called implies that it should return `False` when not authenticated, rather than throwing an exception.

scholtzan (approved):

PR #3204, opened 2022-09-04
feat(): Adding ETL for monitoring_airflow datasets

kik-kik:
# feat(): Adding ETL for monitoring_airflow datasets

The objective is to create a Looker dashboard so we can see how DAGs and tasks are doing.

Follow up PR:
- https://github.com/mozilla/lookml-generator/pull/514

Checklist for reviewer:


fbertsch (commented):
I like the idea of being able to build our own monitoring dashboards rather than rely on the decidedly clunky Airflow UI. @kik-kik what do you think about using _only_ views here? That would ensure we always have the latest data when running queries, and would greatly reduce the complexity of this pipeline.

Q: How is the data getting loaded into BQ in the first place?

kik-kik:
@fbertsch: the data is being loaded from Postgres using Fivetran. There's a connector set up to pull the data every 24 hours (since most of our jobs run at that interval anyways). I think in this case we need the intermediate step of loading the data into `moz-fx-data-shared-prod` as I don't believe Looker would have direct access to the fivetran project.

fbertsch:
> I don't believe Looker would have direct access to the fivetran project.

We can use authorized views then, or even authorize the entire fivetran dataset to `moz-fx-data-shared-prod` / `mozdata`. That's the commonly accepted path

whd:
What's the rationale for creating a new dataset here? Could we add these to `monitoring`/`monitoring_derived` with an `airflow_` prefx instead?

kik-kik:
@whd we could do that, I just thought it might make sense to keep those separate since they both hold different type of data.

whd:
> @whd we could do that, I just thought it might make sense to keep those separate since they both hold different type of data.

In my (not well-researched) opinion, `monitoring_airflow` logically feels like a subset of `monitoring`. The current `monitoring`/`monitoring_derived` datasets currently hold couple of different types of data from disparate source e.g. views on payload_bytes, derived views of live/stable tales, and shredder status. I think having a single monitoring dataset is preferable to having multiple domain-specific ones unless there are specific e.g. security or access characteristics we're trying to achieve by separating it into its own dataset.

I don't hold this opinion strongly though. If you want to use a separate dataset I think this PR is structured correctly and would automatically create the datasets upon merge.

sql/moz-fx-data-shared-prod/monitoring_airflow/dataset_metadata.yaml
```diff
@@ -0,0 +1,11 @@
+friendly_name: Monitoring - Airflow
+description: |-
+  User-facing views for pipeline monitoring,
```
whd, line 3: This looks like copypasta from the standard monitoring views dataset

kik-kik:
@whd as requested I moved over the airflow specific sql code into `monitoring` and `monitoring_derived` folders.

2 more questions:
1. since this is all incremental `false`, would `schema.yaml` be preferred over `init.sql`?
2. Regarding Frank's suggestion of using authorized views in order to remove the `_derived` step for this data. Is this something we could do and still have it usable by Looker? (Frank's comment: https://github.com/mozilla/bigquery-etl/pull/3204#issuecomment-1239533091)

Thanks!

whd:
> since this is all incremental false, would schema.yaml be preferred over init.sql?

I think so but I would check with @scholtzan 

> Regarding Frank's suggestion of using authorized views in order to remove the _derived step for this data. Is this something we could do and still have it usable by Looker? (Frank's comment:
https://github.com/mozilla/bigquery-etl/pull/3204#issuecomment-1239533091)

It would be possible but we generally treat fivetran datasets as an implementation detail and I think keeping the copy step into `monitoring_derived` step is the correct course of action here. The main reason for this is that we don't control fivetran dataset management to the same degree that we control e.g. `monitoring_derived`, since the Fivetran application is largely responsible for administering its own datasets. We technically have another mechanism called [dataset syndication](https://docs.google.com/document/d/13EKQIB6P2bca8gAKtZ43DjNY1Iyqvl9oYoNi4vB9yto/edit#) that maps approximately into the same problem space but because DSRE has limited management of fivetran datasets the same problem of managing authorizations arises and I think `_derived` copies of the data is the most straightforward solution.

If we did go with authorizations into the fivetran dataset, they would indeed be accessible via Looker because the purpose of an authorization is to delegate ACL management to the authorized view, which in this case would be `monitoring_derived`'s dataset ACL of `workgroup:mozilla-confidential`, itself accessible via the standard Looker SA.

scholtzan:
`schema.yaml` is generally preferred over `init.sql`

bigquery_etl/dryrun.py
```diff
@@ -154,6 +154,10 @@
         "sql/moz-fx-data-marketing-prod/acoustic/**/*.sql",
         recursive=True,
     ),
+    *glob.glob(
+        "sql/moz-fx-data-shared-prod/monitoring_derived/**/airflow_*.sql",
+        recursive=True,
+    ),  # noqa E501
```
scholtzan, line 160: why is access denied to these tables?

sql/moz-fx-data-shared-prod/monitoring_derived/airflow_dag_run_v1/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: Airflow DAG run instances
+description: List of DAG run instances
+labels:
+  incremental: false
+owners:
+  - kignasiak@mozilla.com
+scheduling:
+  dag_name: bqetl_monitoring_airflow
+  referenced_tables:
+    - ['moz-fx-data-bq-fivetran', 'airflow_metadata_airflow_db', 'dag_run']
```
scholtzan, line 10: There is also a configuration option to trigger the Airflow import in this DAG: https://github.com/mozilla/bigquery-etl/blob/f6b20988b5f19c9ce5ce04eda405a7bae89e6ff7/docs/reference/scheduling.md?plain=1#L64-L67

bigquery_etl/dryrun.py
kik-kik: I believe that's because the SQL is trying to access Fivetran tables which the user used by the CI does not have access to.

sql/moz-fx-data-shared-prod/monitoring_derived/airflow_dag_run_v1/metadata.yaml
kik-kik: made that tweak.

kik-kik:
@fbertsch as per whd's comment, I decided to leave the ETL in place instead of trying to use authorised views.

@scholtzan thanks a lot for the review, I tweaked the metadata to set dependency to manually trigger the Fivetran load at the beginning of the DAg and schema files added. I was having some issues trying to create the tables using the schema files so for now decided to keep the `init.sql` files. We should try and update the schemas to propagate the field descriptions over to BQ.

@whd thanks a lot for your comments, I believe all of them should now be addressed.

Thanks everyone!

scholtzan (approved):

PR #3209, opened 2022-09-07
Fix table pattern matching logic

akkomar:
This is a follow up to https://github.com/mozilla/bigquery-etl/pull/3186

Checklist for reviewer:

### DIFF
--- a/tests/cli/test_cli_utils.py
+++ b/tests/cli/test_cli_utils.py
@@ -8,6 +8,7 @@ from bigquery_etl.cli.utils import (
     is_valid_dir,
     is_valid_file,
     is_valid_project,
+    table_matches_patterns,
 )
 
 TEST_DIR = Path(__file__).parent.parent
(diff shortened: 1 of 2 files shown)


tests/cli/test_cli_utils.py
```diff
@@ -39,3 +40,26 @@ def test_is_valid_project(self):
         assert is_valid_project(None, None, "moz-fx-data-shared-prod")
         with pytest.raises(BadParameter):
             assert is_valid_project(None, None, "not-existing")
+
+    def test_table_matches_patterns(self):
+        assert not table_matches_patterns(
+            table="telemetry_live.main_v4",
+            pattern=["telemetry_live.main_v4", "telemetry_live.event_v4"],
+            invert=True,
+        )
```
akkomar, line 49: This is the case that's fixed here. This error was causing tables that we wanted to be excluded in https://github.com/mozilla/telemetry-airflow/blob/main/dags/copy_deduplicate.py#L81-L85 to get processed in `copy_deduplicate_all`.

fbertsch (approved):
It would be good to also have a test for `invert = False`.

tests/cli/test_cli_utils.py
kik-kik: 😱  Thanks for fixing this!

PR #3229, opened 2022-09-19
Ignore `jsonPayload.fields.id` column for `fxa_auth_events` (bugs 1791061, 1791062)

sean-rose:
It changed type from string to float in recent `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_auth_20*` tables, which is breaking the `fxa_auth_events` ETL (bugs [1791061](https://bugzilla.mozilla.org/show_bug.cgi?id=1791061), [1791062](https://bugzilla.mozilla.org/show_bug.cgi?id=1791062)), and it doesn't appear to be populated for `amplitudeEvent` events anyway.

---
Checklist for reviewer:


kik-kik:
Are you sure we can null this field @sean-rose ?

relud (approved):

relud:
CI failure looks unrelated

sean-rose:
> Are you sure we can null this field @sean-rose ?

Pretty sure.

Unfortunately due to the nature of the schema change and how BigQuery wildcard table queries work I can't easily query all matching tables to verify `jsonPayload.fields.id` is always null for `amplitudeEvent` events, but that has been the case for all of the ~10 days I've spot-checked.

Also, as far as I can tell `jsonPayload.fields.id` isn't currently used in any downstream ETL.

relud:
> Unfortunately due to the nature of the schema change and how BigQuery wildcard table queries work I can't easily query all matching tables to verify `jsonPayload.fields.id` is always null for `amplitudeEvent` events, but that has been the case for all of the ~10 days I've spot-checked.

you should only need to spot-check days after the ETL started failing, because you can easily check the derived table for all days prior to that.

sean-rose:
> > Unfortunately due to the nature of the schema change and how BigQuery wildcard table queries work I can't easily query all matching tables to verify `jsonPayload.fields.id` is always null for `amplitudeEvent` events, but that has been the case for all of the ~10 days I've spot-checked.
> 
> you should only need to spot-check days after the ETL started failing, because you can easily check the derived table for all days prior to that.

Good point!  Using that approach I've verified that `jsonPayload.fields.id` has been null for all `amplitudeEvent` events thus far.

PR #3234, opened 2022-09-20
feat(DENG-351): fxa_devices_first_seen_by_users_services_v1 model added

kik-kik:
# feat(DENG-351): fxa_users_services_first_seen_v1 remodeled

Checklist for reviewer:

### DIFF
--- a/dags/bqetl_fxa_events.py
+++ b/dags/bqetl_fxa_events.py
@@ -336,6 +336,23 @@ with DAG(
         arguments=["--schema_update_option=ALLOW_FIELD_ADDITION"],
     )
 
+    x_accounts_derived__fxa_devices_first_seen_by_user_service__v1 = bigquery_etl_query(
+        task_id="x_accounts_derived__fxa_devices_first_seen_by_user_service__v1",
+        destination_table="fxa_devices_first_seen_by_user_service_v1",
+        dataset_id="firefox_accounts_derived",
+        project_id="moz-fx-data-shared-prod",
+        owner="kignasiak@mozilla.com",
+        email=[
+            "dthorn@mozilla.com",
+            "kignasiak@mozilla.com",
+            "telemetry-alerts@mozilla.com",
+        ],
+        date_partition_parameter=None,
+        depends_on_past=False,
+        task_concurrency=1,
+        parameters=["submission_date:DATE:{{ds}}"],
+    )
+
     firefox_accounts_derived__exact_mau28__v1.set_upstream(
         firefox_accounts_derived__fxa_users_last_seen__v1
     )
@@ -365,5 +382,13 @@ with DAG(
     )
 
     firefox_accounts_derived__fxa_users_services_daily__v1.set_upstream(
+        firefox_accounts_derived__fxa_content_events__v1
+    )
+
+    x_accounts_derived__fxa_devices_first_seen_by_user_service__v1.set_upstream(
+        firefox_accounts_derived__fxa_auth_events__v1
+    )
+
+    x_accounts_derived__fxa_devices_first_seen_by_user_service__v1.set_upstream(
         firefox_accounts_derived__fxa_content_events__v1
     )
--- a/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_devices_first_seen_by_user_service_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_devices_first_seen_by_user_service_v1/query.sql
@@ -1,0 +1,137 @@
+WITH fxa_events AS (
+  SELECT
+    `timestamp`,
+    user_id,
+    IF(service IS NULL AND event_type = 'fxa_activity - cert_signed', 'sync', service) AS service,
+    device_id,
+    os_name,
+    flow_id,
+    event_type,
+    country,
+    `language`,
+    entrypoint,
+    utm_term,
+    utm_medium,
+    utm_source,
+    utm_campaign,
+    utm_content,
+    ua_version,
+    ua_browser,
+  FROM
+    `moz-fx-data-shared-prod.firefox_accounts.fxa_content_auth_oauth_events`
+  WHERE
+    DATE(`timestamp`)
+    BETWEEN DATE_SUB(@submission_date, INTERVAL 1 DAY)
+    AND @submission_date
+),
+entrypoints AS (
+  SELECT DISTINCT
+    flow_id,
+    entrypoint
+  FROM
+    fxa_events
+  WHERE
+    entrypoint IS NOT NULL
+  QUALIFY
+    ROW_NUMBER() OVER (PARTITION BY flow_id ORDER BY `timestamp` ASC) = 1
+),
+utms AS (
+  SELECT DISTINCT
+    flow_id,
+    FIRST_VALUE(utm_term) OVER (_window) AS utm_term,
+    FIRST_VALUE(utm_medium) OVER (_window) AS utm_medium,
+    FIRST_VALUE(utm_source) OVER (_window) AS utm_source,
+    FIRST_VALUE(utm_campaign) OVER (_window) AS utm_campaign,
+    FIRST_VALUE(utm_content) OVER (_window) AS utm_content,
+  FROM
+    fxa_events
+  WHERE
+    flow_id IS NOT NULL
+  WINDOW
+    _window AS (
+      PARTITION BY
+        flow_id
+      ORDER BY
+        `timestamp` ASC
+      ROWS BETWEEN
+        UNBOUNDED PRECEDING
+        AND UNBOUNDED FOLLOWING
+    )
+),
+new_user_service_device_entries AS (
+  SELECT DISTINCT
+    flow_id,
+    device_id,
+    `timestamp`,
+    user_id,
+    service,
+    os_name,
+    country,
+    `language`,
+    ua_version,
+    ua_browser,
+  FROM
+    fxa_events
+  WHERE
+    DATE(`timestamp`) = @submission_date
+    AND flow_id IS NOT NULL
+    AND ((event_type IN ('fxa_login - complete', 'fxa_reg - complete') AND service IS NOT NULL))
+    AND CONCAT(user_id, service, device_id) NOT IN (
+      SELECT
+        CONCAT(user_id, service, device_id)
+      FROM
+        `moz-fx-data-shared-prod.firefox_accounts_derived.fxa_devices_first_seen_by_user_service_v1`
+      WHERE
+        DATE(first_seen_date) < @submission_date
+    )
+),
+first_values AS (
+  SELECT DISTINCT
+    device_id,
+    FIRST_VALUE(flow_id) OVER (_window) AS first_flow_id,
+    FIRST_VALUE(`timestamp`) OVER (_window) AS first_seen_date,
+    ANY_VALUE(user_id) OVER (_window) AS user_id,
+    ANY_VALUE(service) OVER (_window) AS service,
+    ANY_VALUE(os_name) OVER (_window) AS os_name,
+    ANY_VALUE(country) OVER (_window) AS country,
+    ANY_VALUE(`language`) OVER (_window) AS `language`,
+    ANY_VALUE(ua_version) OVER (_window) AS ua_version,
+    ANY_VALUE(ua_browser) OVER (_window) AS ua_browser,
+  FROM
+    new_user_service_device_entries
+  WINDOW
+    _window AS (
+      PARTITION BY
+        device_id
+      ORDER BY
+        `timestamp` ASC
+      ROWS BETWEEN
+        UNBOUNDED PRECEDING
+        AND UNBOUNDED FOLLOWING
+    )
+)
+SELECT
+  first_values.*,
+  entrypoints.* EXCEPT (flow_id),
+  utms.* EXCEPT (flow_id),
+FROM
+  first_values
+LEFT JOIN
+  entrypoints
+ON
+  first_values.first_flow_id = entrypoints.flow_id
+LEFT JOIN
+  utms
+ON
+  first_values.first_flow_id = utms.flow_id
+WHERE
+  user_id IS NOT NULL
+QUALIFY
+  ROW_NUMBER() OVER (
+    PARTITION BY
+      user_id,
+      service,
+      device_id
+    ORDER BY
+      first_seen_date ASC
+  ) = 1  -- this could be partitioned by first_flow_id to handle multiple device_ids being attached to a single flow_id
--- a/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_oauth_events_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_oauth_events_v1/query.sql
@@ -28,7 +28,7 @@ WITH base AS (
       ) AS jsonPayload
     )
   FROM
-    `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_oauth_20*`
+    `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_oauth_20*`  -- # TODO: this table does not seem to exist anymore, should this query be deprecated?
   WHERE
     _TABLE_SUFFIX = FORMAT_DATE('%y%m%d', @submission_date)
 )
--- a/sql/moz-fx-data-shared-prod/firefox_accounts/fxa_devices_first_seen_by_user_service/view.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts/fxa_devices_first_seen_by_user_service/view.sql
@@ -1,0 +1,7 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.firefox_accounts.fxa_devices_first_seen_by_user_service`
+AS
+SELECT
+  *
+FROM
+  `moz-fx-data-shared-prod.firefox_accounts_derived.fxa_devices_first_seen_by_user_service_v1`
--- a/sql/moz-fx-data-shared-prod/firefox_accounts_derived/funnel_events_source_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts_derived/funnel_events_source_v1/query.sql
@@ -17,7 +17,7 @@ SELECT
     STRUCT('email_version' AS key, email_version AS value)
   ] AS extra,
   CAST([] AS ARRAY<STRUCT<key STRING, value STRING>>) AS experiments,
-  * EXCEPT (device_id),
+  *,
 FROM
   `moz-fx-data-shared-prod.firefox_accounts.fxa_content_auth_oauth_events`
 WHERE
--- a/sql/moz-fx-data-shared-prod/firefox_accounts_derived/funnel_events_source_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts_derived/funnel_events_source_v1/query.sql
@@ -17,7 +17,7 @@ SELECT
     STRUCT('email_version' AS key, email_version AS value)
   ] AS extra,
   CAST([] AS ARRAY<STRUCT<key STRING, value STRING>>) AS experiments,
-  *,
+  * EXCEPT (device_id),
 FROM
   `moz-fx-data-shared-prod.firefox_accounts.fxa_content_auth_oauth_events`
 WHERE
--- a/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_devices_first_seen_by_user_service_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_devices_first_seen_by_user_service_v1/metadata.yaml
@@ -1,0 +1,50 @@
+---
+friendly_name: FxA device(_id) first seen by user, service combination
+
+description: |
+  Contains information when a device_id per
+  user and service combination was first seen by us.
+
+  Additional attributes get captured to provide insight into
+  what os and browser is used by the new device, as well as,
+  the device's entrypoint and utm information.
+
+  This information is static and represents as was at the time
+  (does not change).
+
+  Lowest granularity field: device_id
+
+  Notes:
+    - There are very few cases when a single flow_id
+      can contain multiple device_ids.
+    - utm fields contain very high levels of null values.
+
+  More information about fxa data can be found here:
+  https://docs.telemetry.mozilla.org/datasets/fxa.html
+
+owners:
+  - kignasiak@mozilla.com
+labels:
+  application: fxa
+  incremental: true
+  schedule: daily
+scheduling:
+  dag_name: bqetl_fxa_events
+  # We completely recreate this table every day;
+  # making it incremental is possible but nuanced since it windows over
+  # events that may cross the midnight boundary.
+  date_partition_parameter: null
+  parameters:
+    - submission_date:DATE:{{ds}}
+bigquery:
+  time_partitioning:
+    field: first_seen_date
+    type: day
+    require_partition_filter: true
+    expiration_days: null
+  clustering:
+    fields:
+      - service
+      - os_name
+      - country
+references: {}
--- a/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_devices_first_seen_by_user_service_v1/schema.yaml
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_devices_first_seen_by_user_service_v1/schema.yaml
@@ -1,0 +1,109 @@
+fields:
+
+- mode: NULLABLE
+  name: flow_id
+  type: STRING
+  description: |
+    The flow identifier. A randomly-generated opaque id.
+
+- mode: NULLABLE
+  name: user_id
+  type: STRING
+  description:
+
+- mode: NULLABLE
+  name: service
+  type: STRING
+  description: |
+    The service identifier. For Sync it may be empty or sync.\
+    For OAuth reliers it is their hex client id.
+
+- mode: NULLABLE
+  name: device_id
+  type: STRING
+  description: |
+    The most granual field.\
+    The id of the device record.\
+    This does correlate back to a record the FxA user db.
+
+- mode: NULLABLE
+  name: os_name
+  type: STRING
+  description:
+
+- mode: NULLABLE
+  name: country
+  type: STRING
+  description:
+
+- mode: NULLABLE
+  name: language
+  type: STRING
+  description:
+
+- mode: NULLABLE
+  name: entrypoint
+  type: STRING
+  description: |
+    The entrypoint of the first\
+    flow in the session.\
+    Typically a UI touchpoint like "preferences".
+
+- mode: NULLABLE
+  name: first_seen_date
+  type: TIMESTAMP
+  description: |
+    Timestamp when the first event occured\
+    for a specific device_id
+
+- mode: NULLABLE
+  name: utm_term
+  type: STRING
+  description: |
+    Marketing campaign search term\
+    for the first flow in the session.\
+    Not stored if the DNT request header was 1.
+
+- mode: NULLABLE
+  name: utm_medium
+  type: STRING
+  description: |
+    Marketing campaign medium for the first\
+    flow in the session.\
+    Not stored if the DNT request header was 1.
+
+- mode: NULLABLE
+  name: utm_source
+  type: STRING
+  description: |
+    Marketing campaign source for\
+    the first flow in the session.\
+    Not stored if the DNT request header was 1.
+
+- mode: NULLABLE
+  name: utm_campaign
+  type: STRING
+  description: |
+    Marketing campaign identifier\
+    for the first flow in the session.\
+    Not stored if the DNT request header was 1.
+
+- mode: NULLABLE
+  name: utm_content
+  type: STRING
+  description: |
+    Marketing campaign content identifier\
+    for the first flow in the session.\
+    Not stored if the DNT request header was 1.
+
+- mode: NULLABLE
+  name: ua_version
+  type: STRING
+  description: |
+    The user's browser version.
+
+- mode: NULLABLE
+  name: ua_browser
+  type: STRING
+  description: |
+    The user's web browser, e.g. 'Firefox' or 'Chrome'.


gkabbz (changes requested):

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/init.sql
```diff
@@ -1,144 +1,33 @@
...
+  DATE(MIN(timestamp)) AS first_seen_date,
+  user_id,
+  service,
+  ANY_VALUE(country) AS country,
+  ANY_VALUE(os_name) AS os_name,
+FROM fxa_events
 WHERE
-  first_service_flow IS NOT NULL
+  user_id IS NOT NULL
+  AND service IS NOT NULL
```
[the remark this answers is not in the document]
kik-kik: honestly, I'm not sure. This piece of logic I copied over from the original model. It might not be needed, I will double check tomorrow.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/init.sql
```diff
@@ -1,144 +1,33 @@
...
-  first_services_g s
-LEFT JOIN
-  flows f
-USING
-  (first_service_flow)
+  DATE(MIN(timestamp)) AS first_seen_date,
+  user_id,
+  service,
+  ANY_VALUE(country) AS country,
+  ANY_VALUE(os_name) AS os_name,
```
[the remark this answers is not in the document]
kik-kik: my assumption here is that on the same day all fields will have the same value as a country and os_name. So any not null value here should result in the same result (assumption).

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/init.sql
```diff
@@ -1,144 +1,33 @@
...
-FROM
-  first_services_g s
-LEFT JOIN
-  flows f
-USING
-  (first_service_flow)
+  DATE(MIN(timestamp)) AS first_seen_date,
+  user_id,
+  service,
+  ANY_VALUE(country) AS country,
```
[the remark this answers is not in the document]
kik-kik: same response as for `os_name` comment below

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/init.sql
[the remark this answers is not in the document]
kik-kik: with the current fxa_events CTE seems to imply that services could contain null values. I'd say we should keep this filter out because if no service is listed in the row then we can really use it for this model. Same with user_id, user_id implies it's a record for a registered user.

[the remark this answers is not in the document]
kik-kik: just confirmed, we have `61788819529` records with user_id and service fields being null in the fxa_events CTE since `2021-01-01`

dags/bqetl_fxa_events.py
```diff
@@ -298,6 +298,18 @@
         depends_on_past=False,
     )
 
+    firefox_accounts_derived__fxa_users_services_first_seen__v1 = bigquery_etl_query(
+        task_id="firefox_accounts_derived__fxa_users_services_first_seen__v1",
+        destination_table="fxa_users_services_first_seen_v1",
+        dataset_id="firefox_accounts_derived",
+        project_id="moz-fx-data-shared-prod",
+        owner="dthorn@mozilla.com",
```
gkabbz, line 306: @kik-kik you don't want to add yourself as an owner too?

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/init.sql
```diff
@@ -1,144 +1,18 @@
...
-LEFT JOIN
-  flows f
-USING
-  (first_service_flow)
+SELECT DISTINCT
+  CAST(NULL AS STRING) AS user_id,
+  CAST(NULL AS STRING) AS service,
+  CAST(NULL AS STRING) AS os_name,
+  CAST(NULL AS DATE) AS first_seen_date,
+FROM fxa_events
```
gkabbz, line 15: @kik-kik are there additional dimensions we can add to this table that provide some useful information for first seen date?

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
```diff
@@ -1,136 +1,44 @@
...
-      IF(service IS NULL AND event_type = 'fxa_activity - cert_signed', 'sync', service) AS service
-    )
+    timestamp,
+    user_id,
+    IF(service IS NULL AND event_type = 'fxa_activity - cert_signed', 'sync', service) AS service,
+    os_name,
   FROM
-    firefox_accounts.fxa_content_auth_oauth_events
+    `moz-fx-data-shared-prod.firefox_accounts.fxa_content_auth_oauth_events`
+  WHERE DATE(timestamp) = DATE(@submission_date)
```
gkabbz, line 9: @kik-kik it looks like the init file creates an empty table and this is running a one day pull. I'm wondering how we do the initial backfill. For the [browser first seen tables](https://github.com/mozilla/bigquery-etl/blob/9bca48821a8a0d40b1700cc14ecd8068d132ed06/sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v1/query.sql) the process is usually a backfill that scans the entire table then the daily job checks to see if the client_id value already exists in the first seen table then appends the new data only.

dags/bqetl_fxa_events.py
kik-kik: This value come from `metadata.yaml` in the dataset. I'll make an update there to make me the owner of this.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/init.sql
kik-kik: I'm considering adding flow_id so that we could use it to map a specific user to dimensions like entrypoint and country instead of adding this logic here.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
kik-kik: I'm happy to explain the backfill options to you if you'd like in a call. This is fine.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/init.sql
gkabbz: I think the flow_id is a good ide but I don't think using it to pull commonly used dimensions in will be very performant, especially in looker. I looked at the lineage for this table and we'll likely have a user_id with a 2019 start date. If we setup the model to look for the country for this client, then we'll have to query the `fxa_content_auth_oauth_events` view which is based on `fxa_events_v1`(4.85TB) and `fxa_auth_events_v1` (52TB) derived tables. I think it will be very slow. What do you think about pulling in what we think might be commonly used dimensions then including flow_id so that if an analyst wants to look at something that isn't pre aggregated then they have a much simpler join to make?

kik-kik: let's chat about it next time we meet.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/init.sql
```diff
@@ -1,144 +1,18 @@
...
+CREATE TABLE IF NOT EXISTS
+  `moz-fx-data-shared-prod.firefox_accounts_derived.fxa_users_services_first_seen_v1`
 PARTITION BY
-  DATE(first_service_timestamp)
+  first_seen_date
 CLUSTER BY
   service,
-  user_id
+  user_id,
+  os_name
```
relud, line 8: order matters for clustering fields, and because user_id is unique per row, adding any fields after it should have no useful impact

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
```diff
@@ -1,136 +1,93 @@
...
+    fxa_events
   WHERE
-    (
-      (event_type IN ('fxa_login - complete', 'fxa_reg - complete') AND service IS NOT NULL)
-      OR (event_type LIKE 'fxa_activity%')
+    flow_id IS NOT NULL
+    AND ((event_type IN ('fxa_login - complete', 'fxa_reg - complete') AND service IS NOT NULL))
+    AND CONCAT(user_id, service, os_name) NOT IN (
+      SELECT CONCAT(user_id, service, os_name)
+      FROM `moz-fx-data-shared-prod.firefox_accounts_derived.fxa_users_services_first_seen_v1`
```
relud, line 51: This should only scan prior days, so that we can run backfills:
```suggestion
      FROM `moz-fx-data-shared-prod.firefox_accounts_derived.fxa_users_services_first_seen_v1`
      WHERE first_seen_date < @submission_date
```

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
```diff
@@ -1,136 +1,93 @@
...
-  first_service_flow IS NOT NULL
+  * EXCEPT(row_id)
+FROM (
+  SELECT
+    *,
+    ROW_NUMBER() OVER(PARTITION BY user_id, service, os_name ORDER BY first_seen_date ASC) AS row_id
+  FROM first_values
+  WHERE user_id IS NOT NULL
+)
+WHERE row_id = 1
```
relud, line 93: this can be done without a subquery using `QUALIFY`:
```suggestion
  *
FROM first_values
WHERE user_id IS NOT NULL
QUALIFY
    ROW_NUMBER() OVER(PARTITION BY user_id, service, os_name ORDER BY first_seen_date ASC) = 1
```

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
```diff
@@ -1,136 +1,93 @@
...
+    utm_medium,
+    utm_campaign,
+    utm_content,
+    ua_version,
+    ua_browser,
+    entrypoint,
   FROM
-    firefox_accounts.fxa_content_auth_oauth_events
+    `moz-fx-data-shared-prod.firefox_accounts.fxa_content_auth_oauth_events`
+  WHERE DATE(timestamp) = DATE(@submission_date)  -- # TODO: need to verify how long the flow session lasts. Might have to make this window a bit bigger
```
relud, line 22: `@submission_date` should already be a `DATE`:
```suggestion
  WHERE DATE(timestamp) = @submission_date  -- # TODO: need to verify how long the flow session lasts. Might have to make this window a bit bigger
```

relud (commented):
sql files need to be formatted, e.g. 
```sh
PATH="venv/bin:$PATH" script/bqetl format \
  $(git ls-tree -d HEAD --name-only)
```

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/init.sql
kik-kik: it's not unique, but there should be very few entries per user. Thanks for pointing this out, I will remove it now.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
kik-kik: that's a very good point. Thank you for pointing this out! 🙇

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
kik-kik: This is indeed much cleaner, thanks for the suggestion 🙇

kik-kik:
> sql files need to be formatted, e.g.
> 
> ```shell
> PATH="venv/bin:$PATH" script/bqetl format \
>   $(git ls-tree -d HEAD --name-only)
> ```

Yes, I was going to do it at the end once the logic is finalised.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_content_events_v1/query.sql
```diff
@@ -6,7 +6,8 @@ SELECT
           (
             SELECT AS STRUCT
               jsonPayload.fields.* EXCEPT (device_id, user_id),
-              TO_HEX(SHA256(jsonPayload.fields.user_id)) AS user_id
+              TO_HEX(SHA256(jsonPayload.fields.user_id)) AS user_id,
```
gkabbz, line 9: @kik-kik do we need to hash the user_id? I know for device_id you have an open question out but not sure I understand the user_id part.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/init.sql
```diff
@@ -1,144 +1,24 @@
...
-  -- I've verified that `date(first_service_timestamp), count(distinct user_id) where did_register = true group by 1`  matches the counts of registrations per day in amplitude.
-first_services_g AS (
-  SELECT
-    * EXCEPT (_n)
-  FROM
-    first_services
-  WHERE
-    _n = 1
-),
-  -- sadly, `entrypoint` is null on registration complete and login complete events.
```
gkabbz, line 72: @kik-kik just looking through the parts we are not keeping. Given this comment, does this mean that the entrypoint field in the table we currently have will always be null?

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/init.sql
```diff
@@ -1,144 +1,24 @@
...
+CREATE TABLE IF NOT EXISTS
+  `moz-fx-data-shared-prod.firefox_accounts_derived.fxa_users_services_first_seen_v1`
 PARTITION BY
-  DATE(first_service_timestamp)
+  first_seen_date
 CLUSTER BY
   service,
-  user_id
+  os_name,
+  country
```
gkabbz, line 8: I think clustering by user_id vs country would be more beneficial

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
```diff
@@ -1,136 +1,74 @@
...
     user_id,
     service,
-      -- using mode_last with w1_reversed to get mode_first
-    udf.mode_last(ARRAY_AGG(`timestamp`) OVER w1_reversed) AS first_service_timestamp_last,
-    udf.mode_last(ARRAY_AGG(flow_id) OVER w1_reversed) AS first_service_flow,
-    LOGICAL_OR(IFNULL(event_type = 'fxa_reg - complete', FALSE)) OVER w1_reversed AS did_register
+    device_id,
+    os_name,
+    country,
+    LANGUAGE,
```
gkabbz, line 30: Can we include the utm_term, utm_medium, utm_source, utm_campaign, utm_content, ua_version, ua_browser in a sample table? I think we should try and run some queries and see what the analysis would feel like if we needed the fields when in the table vs if we had to do joins to get the fields.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_content_events_v1/query.sql
kik-kik: this logic has always been here, I only added a comma here. I would not modify this.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/init.sql
kik-kik: on the point @relud made, because we will have so few entries per user_id clustering on that field makes no sense here.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
kik-kik: sure I can add them back in, based on the slack conversation my impression was that we should remove it.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_content_events_v1/query.sql
gkabbz: oops missed that

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
```diff
@@ -1,136 +1,99 @@
...
+    utm_term,
+    utm_medium,
+    utm_source,
+    utm_campaign,
+    utm_content,
+    ua_version,
+    ua_browser,
   FROM
-    firefox_accounts.fxa_content_auth_oauth_events
+    `moz-fx-data-shared-prod.firefox_accounts.fxa_content_auth_oauth_events`  -- # TODO: there seems to be fxa_content_auth_events, maybe we should use that instead since oauth has been deprecated for a long time now?
```
gkabbz, line 21: @kik-kik I think you are the best person to make this call on which table to use.

kik-kik:
sorry had some commit issues, not quite sure what happened. I've recreated the branch and pushed it anew.

kik-kik:
@gkabbz given the introduction of `device_id` I think we should consider renaming the `fxa_users_services_first_seen_v1` model to `fxa_users_services_devices_first_seen_v1` or something like that. What do you think?

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
kik-kik: For now let's use the same view, but I think we should see how many "similar" views and tables we have and reduce them down to one to avoid having to maintain multiple things that are "basically" the same.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
```diff
@@ -1,136 +1,99 @@
...
+    utm_source,
+    utm_campaign,
+    utm_content,
+    ua_version,
+    ua_browser,
   FROM
-    base
+    fxa_events
+  LEFT JOIN
+    (SELECT DISTINCT flow_id, entrypoint FROM fxa_events) AS entrypoints
```
irrationalagent, line 48: ~i could be wrong on this, but if there are flow_ids that have multiple values for `entrypoint`, then this will result in multiple rows per flow_id?~ edit: nevermind, I think windowed FIRST_VALUE function below takes care of this.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
```diff
@@ -1,136 +1,99 @@
...
+  USING
+    (flow_id)
   WHERE
-    (
-      (event_type IN ('fxa_login - complete', 'fxa_reg - complete') AND service IS NOT NULL)
-      OR (event_type LIKE 'fxa_activity%')
+    DATE(timestamp) = @submission_date
+    AND flow_id IS NOT NULL
+    AND ((event_type IN ('fxa_login - complete', 'fxa_reg - complete') AND service IS NOT NULL))
+    AND CONCAT(user_id, service, device_id) NOT IN (
```
irrationalagent, line 55: IIUC this is to ensure we are not adding additional entries for a previously-seen user-service-device combination. this means that if there ARE previous entries for **a previously seen service, on a different DEVICE** then we WILL insert another row. 

i can see the utility of this from the sync perspective, since it (kind of) tells us when a new device was added to sync (it is not 100% accurate since IIUC in some cases a device that is disconnected from sync but then reconnected will get a fresh device_id - i am not sure of the details here).

assuming we want to go this route, should we rename the table `fxa_users_services_devices_first_seen_v1` ?

irrationalagent (approved):
r+ once we work out what to do about the user-device-service issue in my previous comment

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
```diff
@@ -1,136 +1,99 @@
...
-  -- [kimmy] the variable first_service_timestamp_last is named so because it is actually the last timestamp recorder in the user's first flow,
-  -- NOT the first timestamp in their first flow.
-  -- it's used later on to order by service, so i'm keeping it here and just renaming it.
-first_services AS (
-  SELECT
-    ROW_NUMBER() OVER w1_unframed AS _n,
+new_user_service_os_entries AS (
+  SELECT DISTINCT
+    flow_id,
+    timestamp,
```
relud, line 30: fields whose name matches a keyword should be quoted with back-ticks:
```suggestion
    `timestamp`,
```

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
```diff
@@ -1,136 +1,99 @@
...
     user_id,
     service,
-      -- using mode_last with w1_reversed to get mode_first
-    udf.mode_last(ARRAY_AGG(`timestamp`) OVER w1_reversed) AS first_service_timestamp_last,
-    udf.mode_last(ARRAY_AGG(flow_id) OVER w1_reversed) AS first_service_flow,
-    LOGICAL_OR(IFNULL(event_type = 'fxa_reg - complete', FALSE)) OVER w1_reversed AS did_register
+    device_id,
+    os_name,
+    country,
+    LANGUAGE,
```
relud, line 36:
```suggestion
    `language`,
```

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_oauth_events_v1/query.sql
```diff
@@ -28,7 +28,7 @@ WITH base AS (
       ) AS jsonPayload
     )
   FROM
-    `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_oauth_20*`
+    `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_oauth_20*`  -- # TODO: this table does not seem to exist anymore, should this query be deprecated?
```
relud, line 31: @jbuck can you confirm that fxa oauth events are gone, and we shouldn't expect them somewhere else or something like that?

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/metadata.yaml
```diff
@@ -3,15 +3,30 @@ friendly_name: FxA Users Services First Seen
 description: Usage aggregations describing when each FxA user was first seen
 owners:
   - dthorn@mozilla.com
+  - kignasiak@mozilla.com
```
relud, line 6: there can only be one owner (it's not an array field in airflow). IDK why it's an array field here, but for the scope of this PR, we'll just remove me from the list for this task:

```suggestion
owners:
  - kignasiak@mozilla.com
```

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/metadata.yaml
```diff
@@ -3,15 +3,30 @@ friendly_name: FxA Users Services First Seen
...
 labels:
   application: fxa
   incremental: true
   schedule: daily
 # # Descheduled. Last run was 2021-01-23.
 # # See https://github.com/mozilla/bigquery-etl/issues/1713
```
relud, line 12: not descheduled anymore
```suggestion
```

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
relud: assuming that oauth is gone, i'd recommend the `fxa_content_auth_stdout_events` view here

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
```diff
@@ -1,136 +1,99 @@
...
+    fxa_events
+  LEFT JOIN
+    (SELECT DISTINCT flow_id, entrypoint FROM fxa_events) AS entrypoints
+  USING
+    (flow_id)
   WHERE
-    (
-      (event_type IN ('fxa_login - complete', 'fxa_reg - complete') AND service IS NOT NULL)
-      OR (event_type LIKE 'fxa_activity%')
+    DATE(timestamp) = @submission_date
```
relud, line 52:
```suggestion
    DATE(`timestamp`) = @submission_date
```

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
```diff
@@ -1,136 +1,99 @@
...
-      PARTITION BY
-        f.flow_id
-      ORDER BY
-        f.`timestamp`
-      ROWS BETWEEN
-        UNBOUNDED PRECEDING
-        AND UNBOUNDED FOLLOWING
-    ) AS first_service_country,
-    FIRST_VALUE(f.os_name) OVER (
+    FIRST_VALUE(flow_id) OVER (PARTITION BY user_id ORDER BY timestamp ASC) AS flow_id,
```
relud, line 66:
```suggestion
    FIRST_VALUE(flow_id) OVER (PARTITION BY user_id ORDER BY `timestamp` ASC) AS flow_id,
```

relud (approved):
r+wc

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
relud: i agree, we should rename if using device_id like this

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
kik-kik: Thanks @relud, another view to add to a list to review. I've noticed we have a few different views that seem to be doing the same thing. I will create a separate issue for investigating that and deprecating duplicate logic.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
kik-kik: ok, marking this conversation as resolved then. Let me know if you think something else should be done here.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
kik-kik: 100% agree with this, already mentioned this in one of my previous comments. The dataset has been renamed to `fxa_users_services_devices_first_seen_v1` as suggested by @irrationalagent

kik-kik: ok so naming the query results in the following task name `firefox_accounts_derived__fxa_users_services_devices_first_seen__v1` (67 chars) which exceeds the char limit of `62` and makes the DAG generation fail...

@relud I wonder if we should make a change to bqetl dag generation to truncate the task name to `62` characters during DAG generation but allow actual dataset names to be longer?

kik-kik: @relud would something like this be appropriate?
https://github.com/mozilla/bigquery-etl/pull/3279

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_oauth_events_v1/query.sql
relud: I can confirm that part, the source table is no longer present in BigQuery at all.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/query.sql
relud: yes, see review there for more details

kik-kik: thanks @relud 🙇 The change in bqetl has been applied.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_devices_first_seen_v1/query.sql
```diff
@@ -0,0 +1,99 @@
...
+),
+first_values AS (
+  SELECT DISTINCT
+    FIRST_VALUE(flow_id) OVER (PARTITION BY device_id ORDER BY `timestamp` ASC) AS flow_id,
+    FIRST_VALUE(user_id) OVER (_window) AS user_id,
+    FIRST_VALUE(service) OVER (_window) AS service,
+    FIRST_VALUE(device_id) OVER (_window) AS device_id,
+    FIRST_VALUE(os_name) OVER (_window) AS os_name,
+    FIRST_VALUE(country) OVER (_window) AS country,
+    FIRST_VALUE(LANGUAGE) OVER (_window) AS language,
```
relud, line 72:
```suggestion
    FIRST_VALUE(`language`) OVER (_window) AS `language`,
```

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_devices_first_seen_v1/query.sql
```diff
@@ -0,0 +1,99 @@
...
+  SELECT
+    timestamp,
+    user_id,
+    IF(service IS NULL AND event_type = 'fxa_activity - cert_signed', 'sync', service) AS service,
+    device_id,
+    os_name,
+    flow_id,
+    event_type,
+    country,
+    LANGUAGE,
```
relud, line 11:
```suggestion
    `language`,
```

dags/bqetl_fxa_events.py
```diff
@@ -336,6 +336,23 @@
         arguments=["--schema_update_option=ALLOW_FIELD_ADDITION"],
     )
 
+    ox_accounts_derived__fxa_users_services_devices_first_seen__v1 = bigquery_etl_query(
```
gkabbz, line 339: says `ox` instead of `fx_accounts...` is that intentional or a typo?

dags/bqetl_fxa_events.py
```diff
@@ -336,6 +336,23 @@
...
+        task_id="ox_accounts_derived__fxa_users_services_devices_first_seen__v1",
```
gkabbz, line 340: Same here: says `ox` instead of `fx_accounts...` is that intentional or a typo?

dags/bqetl_fxa_events.py
```diff
@@ -367,3 +384,23 @@
     firefox_accounts_derived__fxa_users_services_daily__v1.set_upstream(
         firefox_accounts_derived__fxa_content_events__v1
     )
+
+    ox_accounts_derived__fxa_users_services_devices_first_seen__v1.set_upstream(
```
gkabbz, line 388: Same here and lines below: says `ox` instead of `fx_accounts...` is that intentional or a typo?

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_devices_first_seen_v1/metadata.yaml
```diff
@@ -0,0 +1,29 @@
+---
+friendly_name: FxA Users Services Devices First Seen
+description: Usage aggregations describing when each FxA device was first seen
```
gkabbz, line 3: Maybe we can expand on the definition a little bit. See descriptions in [PR-3309](https://github.com/mozilla/bigquery-etl/pull/3309/files#)

gkabbz (approved):
Looks good. A couple of comments:
- The name `fxa_users_services_devices_first_seen` seems long. This is probably a preference thing on my part but maybe `fxa_users_first_seen_by_service_device`. Again probably just me being nit picky.
- Looks like you are removing the `fxa_users_services_first_seen` view without a replacement. You should check if this might cause breakage for queries. I still think we should have that view but reading from this table you are creating. I've included a rough sample query in [the notebook](https://colab.research.google.com/drive/15g4eXXeHNVy6dy9cysQ_b5oncL7XFiMm#scrollTo=TKOLjFH3pQML&line=5&uniqifier=1). I'd check that before merging. Winnie has a query you can use to see who is querying. I'll ping her in channel. It doesn't look like there are any additional tables downstream of the views (thanks data hub)
image

dags/bqetl_fxa_events.py
kik-kik: intentional, same as the explanation to Airlfow task name.

kik-kik:
Hey @gkabbz ,

Thanks for the review and the comments. I've re-added `fxa_users_services_first_seen_v1`.

In terms of naming, I personally have nothing against longer names as long as they provide a good summary of the contents. In this case, I do agree though that the name should be changed. I think we should name this model to: `fxa_devices_first_seen_by_user_service` because this reflects the contents and the granularity of the table (device). 

To make this data easier to use per user service we could have another model build on top of this one which just aggregates to first_seen by user / service combination (and avoid recomputing everything).

dags/bqetl_fxa_events.py
```diff
@@ -336,6 +336,23 @@
         arguments=["--schema_update_option=ALLOW_FIELD_ADDITION"],
     )
 
+    x_accounts_derived__fxa_devices_first_seen_by_user_service__v1 = bigquery_etl_query(
```
gkabbz, line 339: missing the 'f'??

dags/bqetl_fxa_events.py
```diff
@@ -367,3 +384,11 @@
     firefox_accounts_derived__fxa_users_services_daily__v1.set_upstream(
         firefox_accounts_derived__fxa_content_events__v1
     )
+
+    x_accounts_derived__fxa_devices_first_seen_by_user_service__v1.set_upstream(
```
gkabbz, line 388: missing the leading `f`???

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_devices_first_seen_by_user_service_v1/query.sql
```diff
@@ -0,0 +1,130 @@
...
+    entrypoint,
+    utm_term,
+    utm_medium,
+    utm_source,
+    utm_campaign,
+    utm_content,
+    ua_version,
+    ua_browser,
+  FROM
+    `moz-fx-data-shared-prod.firefox_accounts.fxa_content_auth_oauth_events`  -- # TODO: there seems to be fxa_content_auth_events, maybe we should use that instead since oauth has been deprecated for a long time now?
```
gkabbz, line 21: Is the TODO here still needed??

kik-kik:
This PR has been superseded by: https://github.com/mozilla/bigquery-etl/pull/3388

PR #2852, opened 2022-04-01
Bug 1761553 - Fix client_id field discovery in Rally analysis tables

akkomar:
Checklist for reviewer:

- [ ] Commits should reference a bug or github issue, if relevant (if a bug is referenced, the pull request should include the bug number in the title)
- [ ] If the PR comes from a fork, trigger integration CI tests by running the [Push to upstream workflow](https://github.com/mozilla/bigquery-etl/actions/workflows/push-to-upstream.yml) and provide the `<username>:<branch>` of the fork as parameter. The parameter will also show up
in the logs of the `manual-trigger-required-for-fork` CI task together with more detailed instructions.
- [ ] If adding a new field to a query, ensure that the schema and dependent downstream schemas have been updated

For modifications to schemas in restricted namespaces (see [`CODEOWNERS`](/CODEOWNERS)):
- [ ] Follow the [change control procedure](https://docs.google.com/document/d/1TTJi4ht7NuzX6BPG_KTr6omaZg70cEpxe9xlpfnHj9k/edit#heading=h.ttegrcfy18ck)

PR #3240, opened 2022-09-22
Update Shredder Pioneer config to account for shared views

fbertsch:
Attention stream datasets are shared to other projects. When analysts create derived datasets from these shared views, it failed Shredder, which was looking for a PIONEER_ID to delete.

This change will give the correct RALLY_ID to delete, but will not actually delete the corresponding rows. As a follow-up to this, we will delete with the appropriate shared-view deletion request.

Checklist for reviewer:


fbertsch:
I have no clue how to go about testing this.

whd:
> I have no clue how to go about testing this.

I'm not sure if we have an easy way of testing this. Even if this wasn't a pioneer-specific change, I'm not sure how shredder changes are usually tested but /CC @relud in case he has thoughts. I believe when setting up the stage environment for pioneer I opted not to configure shredder because we hard-code some identifiers like https://github.com/mozilla/bigquery-etl/blob/main/bigquery_etl/shredder/config.py#L481 and shredder wasn't working there as a result.

Looking through history the last pioneer-specific change we made was https://github.com/mozilla/bigquery-etl/pull/2852. I think we just YOLO'd it. The good news is that since shredder is already broken it's doubtful this particular change could make it worse.

fbertsch:
@whd do you want to yolo this today, or wait until tomorrow and let Arkadiusz look at it in the morning?

whd (approved):
YOLO today

PR #3236, opened 2022-09-21
Move to python 3.10

fbertsch:
Checklist for reviewer:

- [ ] Commits should reference a bug or github issue, if relevant (if a bug is referenced, the pull request should include the bug number in the title)
- [ ] If the PR comes from a fork, trigger integration CI tests by running the [Push to upstream workflow](https://github.com/mozilla/bigquery-etl/actions/workflows/push-to-upstream.yml) and provide the `<username>:<branch>` of the fork as parameter. The parameter will also show up
in the logs of the `manual-trigger-required-for-fork` CI task together with more detailed instructions.
- [ ] If adding a new field to a query, ensure that the schema and dependent downstream schemas have been updated

For modifications to schemas in restricted namespaces (see [`CODEOWNERS`](/CODEOWNERS)):
- [ ] Follow the [change control procedure](https://docs.google.com/document/d/1TTJi4ht7NuzX6BPG_KTr6omaZg70cEpxe9xlpfnHj9k/edit#heading=h.ttegrcfy18ck)

PR #3248, opened 2022-10-05
Update CI to python 3.10

relud:
follow-up from #3236
blocks #3246

updates mypy to pick up fix for https://github.com/python/mypy/issues/13627


.circleci/config.yml
```diff
@@ -58,6 +58,7 @@ jobs:
     steps:
       - checkout
       - *restore_venv_cache
+      - *build
```
relud, line 61: updating the cache keys revealed that this was missing, but `restore_venv_cache` was providing the required dependencies

.circleci/config.yml
```diff
@@ -23,16 +23,16 @@ jobs:
             # patterns to restore cache
             - &python_cache_key
               # yamllint disable-line rule:line-length
-              python-3.8-packages-v1-{{ .Branch }}-{{ checksum "requirements.in" }}-{{ checksum "requirements.txt" }}
+              python-3.10-packages-v1-{{ .Branch }}-{{ checksum "requirements.in" }}-{{ checksum "requirements.txt" }}
```
fbertsch, line 26: Can we parameterize the python version here?

fbertsch (approved):
r+wc, thank you for the followup!

PR #3266, opened 2022-10-11
Add typing stubs to pre-commit mypy run

akkomar:
This is a follow-up to https://mozilla.slack.com/archives/C01E8GDG80N/p1665434432654689

We've seen mypy pre-commit hook failing and complaining about missing library stubs for some of the dependencies. Per [pre-commit mirror docs](https://github.com/pre-commit/mirrors-mypy#using-mypy-with-pre-commit) it runs in an isolated environment so it seems to be making sense.

The issue might be related to https://github.com/mozilla/bigquery-etl/pull/3248.

Changes proposed here do fix the mypy hook run for me on Python 3.10.7. However since I don't understand why we're seeing these errors I am not entirely sure if the local environment (mine and @kik-kik ) isn't at fault here.

Checklist for reviewer:


akkomar (commented):
Lint

.pre-commit-config.yaml
```diff
@@ -32,3 +32,9 @@ repos:
   rev: v0.982
   hooks:
   - id: mypy
+    additional_dependencies: [types-pytz==2022.4.0.0,
+                              types-ujson==5.5.0,
+                              types-python-dateutil==2.8.19,
+                              types-requests==2.28.11.2,
+                              types-attrs==19.1.0,
+                              types-PyYAML==6.0.12]
```
akkomar, line 40: These are actually duplicates of values from requirements.in which is not great.

kik-kik:
I can confirm using this config solves the errors on my machine

.pre-commit-config.yaml
kik-kik: looks like we need to specify the additional_dependencies...

relud (approved):

PR #3267, opened 2022-10-11
Bug 1788650 - relax client_id uniqueness check in baseline_clients_first_seen_v1

akkomar:
This is a follow-up to https://github.com/mozilla/bigquery-etl/pull/3201 where client_id uniqueness checks were introduced in `baseline_clients_first_seen` queries. Afterwards, `moz-fx-data-shared-prod.org_mozilla_firefox_beta_derived.baseline_clients_first_seen_v1` is failing with `@submission_date=2022-10-10`.

It looks like the source of duplicates in this case is [mapping from the new client_id to the old fennec_client_id](https://github.com/mozilla/bigquery-etl/blob/307dc7ffad4c9b1b596c5619e020e665c9f77af9/sql_generators/glean_usage/templates/macros.sql#L4-L13) that [is not 1-to-1 for 300 clients](https://sql.telemetry.mozilla.org/queries/88003/source)(@chutten is this expected?). Looking at the following CTEs in this query, it seems like `SELECT DISTINCT` [here](https://github.com/mozilla/bigquery-etl/blob/307dc7ffad4c9b1b596c5619e020e665c9f77af9/sql_generators/glean_usage/templates/baseline_clients_first_seen_v1.query.sql#L20-L30) might have been introduced to work around the issue we're hitting now. I think it's relatively safe to remove the first uniqueness check as it relates to legacy subset of data that we no longer receive.

Checklist for reviewer:

### DIFF
--- a/sql_generators/glean_usage/templates/macros.sql
+++ b/sql_generators/glean_usage/templates/macros.sql
@@ -27,14 +27,5 @@ _core_clients_first_seen AS (
   JOIN
      _core
   ON _fennec_id_lookup.fennec_client_id = _core.client_id
-  WHERE
-    TRUE
-  QUALIFY
-    IF(
-      COUNT(*) OVER (PARTITION BY _fennec_id_lookup.client_id) > 1
-      OR COUNT(*) OVER (PARTITION BY _core.client_id) > 1,
-      ERROR("duplicate client_id detected"),
-      TRUE
-    )
 )
 {% endmacro %}


chutten:
First and foremost, this record of Fennec's client_id [expired last November](https://dictionary.telemetry.mozilla.org/apps/fenix/metrics/migration_telemetry_identifiers_fennec_client_id) and all code that did migration on the client was removed this year. We can probably stop caring about it on the server side.

Duplicate client_ids are possible by collision (statistically impossible given the age of the universe, but depends on the source of randomness) or by copying. It's not impossible to copy a profiledir on Android, but it's unlikely to happen. If these ancient (because they're running code at least a year old) clients are still reporting data this year, then it could stand to reason that they'd have the weirder setups.

Or, in short: if we wish to keep this info, this amount of dupe is within my gut's range of plausible. But maybe we should stop doing anything with it.

kik-kik:
This seems to be a relatively small table, should we manually dedup it as a one-time fix and retain this logic? If this issue comes up again my understanding would be that there is an issue somewhere.

relud (commented):
if we don't check for duplicates, then we need to remove duplicates, so I propose: #3268

relud:
yes

PR #3272, opened 2022-10-12
Update marketing_attributable_metrics to match dashboard

bani:
This updated marketing_attributable_metrics to match the data in the current dashboard.
Main changes are using the country from first seen table and doing an outer join for search.

Checklist for reviewer:

### DIFF
--- a/sql/moz-fx-data-shared-prod/fenix/marketing_attributable_metrics/view.sql
+++ b/sql/moz-fx-data-shared-prod/fenix/marketing_attributable_metrics/view.sql
@@ -5,8 +5,6 @@ WITH dau_client AS (
   SELECT
     submission_date,
     client_id,
-    first_seen_date,
-    country,
     COUNTIF(days_since_seen < 1) AS dau,
     COUNTIF(first_seen_date = submission_date) AS new_profiles
   FROM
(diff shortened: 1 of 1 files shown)


skahmann3:
Not sure if this PR is meant to be the final stage dataset for marketing analyses or if that's another PR, but it would be good to incorporate some of the changes from [Xuan's analysis](https://colab.research.google.com/drive/1JlOwWRx_HzEtjyeBqls7UKf8L2d6Bdya#scrollTo=UfydwRaYX4dy). She has indicators for new, single day, and multiday profiles which would be useful when setting up future monitoring analyses

bani:
@skahmann3 makes sense! I'd like to merge this version so that we can have the correct numbers on the new Looker explore, but we can iterate on it based on feedback!

bani:
I think the main issue regarding single day / multi day is that we shouldn't report on that for clients that are too new, and usually Marketing wants the full month results right after the month ends. Maybe we could add a "TBD" category.

skahmann3 (approved):
Approving these changes. Though as a note for dashboard development, we need "average clicks per retained multiday client." This will require merging on an aggregated dataset which identifies the single day clients so they can be removed. This can be easily done in Looker and redash

PR #3282, opened 2022-10-14
[DENG-358] New dataset for collecting domain metadata and new top_domains table

quiiver:
Checklist for reviewer:


whd:
I think you may need to split this into two PRs: one to land the dataset per https://mozilla.github.io/bigquery-etl/cookbooks/common_workflows/#creating-a-new-bigquery-dataset, then the following PR to add the tables and jobs once the dataset has been provisioned.

whd:
The dataset has been provisioned via https://github.com/mozilla/bigquery-etl/pull/3292

sql/moz-fx-data-shared-prod/domain_metadata_derived/top_domains_v1/query.sql
```diff
@@ -0,0 +1,78 @@
...
+    h.host,
+    origin,
+    suffix,
+    country_code
+  FROM
+    (SELECT * FROM host_join WHERE host_rank IS NULL) h
+  LEFT JOIN
+    daily_tranco t
+  USING
+    (domain)
```
fbertsch, line 57: Could be:
```suggestion
    host_join
  LEFT JOIN
    daily_tranco t
  USING
    (domain)
  WHERE
    host_join.host_rank IS NULL
```

fbertsch: Could these two joins be just one `ON` clause?
```
  FROM
    crux_country_summary c
  LEFT JOIN
    daily_tranco t
  ON
    c.host = d.host OR (c.host IS NULL AND c.domain = d.domain)
```

sql/moz-fx-data-shared-prod/domain_metadata_derived/top_domains_v1/query.sql
```diff
@@ -0,0 +1,78 @@
+WITH daily_tranco AS (
+  SELECT
+    rank,
+    NET.REG_DOMAIN(domain) AS domain,
+    NET.HOST(domain) AS host
+  FROM
+    `tranco.daily.daily`
+  WHERE
+    date IN (SELECT MAX(date) FROM `tranco.daily.daily`)
```
fbertsch, line 9: Should this (and below, line 21) use `@submission_date`? IIUC these are not idempotent and may result in different values when re-run for the same day.

If this isn't supposed to be idempotent, consider dropping the `@submission_date` altogether (not sure it makes sense for something that's not being submitted to the pipeline).

fbertsch (commented):
The join thing doesn't matter, but the idempotence piece probably needs addressing

sql/moz-fx-data-shared-prod/domain_metadata_derived/top_domains_v1/query.sql
quiiver: I *think* so, however we're leveraging the separate joins to produce rankings for each join so we would lose that if we combine them

sql/moz-fx-data-shared-prod/domain_metadata_derived/top_domains_v1/query.sql
quiiver: I've included submission_date in the output so we can partition the results by scheduled run

fbertsch: Could do for Tranco: `WHERE date = @submission_date`, and for CrUX `WHERE PARSE_DATE('%Y%m', CAST(yyyymm AS STRING)) = @submission_date`, however I have no clue how this dataset is being used so take that with a grain of salt :)

If it's mostly "need to find the most recent value for any domain/host", should this just overwrite every time with the most recent daily/monthly data from tranco/crux?

sql/moz-fx-data-shared-prod/domain_metadata_derived/top_domains_v1/query.sql
fbertsch: Right, I missed that. Carry on

sql/moz-fx-data-shared-prod/domain_metadata_derived/top_domains_v1/query.sql
quiiver: yeah, I was thinking that idempotency is less importance here than reliability. Given that these datasets are outside our control we can't guarantee what the partition dates will be at any given time. I would vote to leave them dynamically computed and know that we lose idempotency with that.

fbertsch (approved):

PR #3289, opened 2022-10-18
Handle Stripe subscription plan changes (DS-2160)

sean-rose:
## [DS-2160](https://mozilla-hub.atlassian.net/browse/DS-2160): Handle and surface plan upgrade events

When the subscription platform handles a plan upgrade for a customer, the associated Stripe `subscription_item` record is updated in place with no indication it was ever different, and `previous_plan_id` and `plan_change_date` metadata fields are set in the Stripe `subscription` record.  Luckily Fivetran stores a history of `subscription` record changes as it syncs so we can use that history of subscription metadata to determine if/how a subscription's plan changed over time.

Notable changes:
- Adds `subscription_platform.stripe_subscriptions_history` and `subscription_platform.nonprod_stripe_subscriptions_history` views which take Fivetran's Stripe `subscription_history` records, interprets the subscription plan metadata changes, and produces records very much like `subscription_platform.stripe_subscriptions` but with `valid_from`, `valid_to`, `plan_started_at`, and `plan_ended_at` columns.
  - `valid_from` is inclusive and will never be null.
  - `valid_to` is non-inclusive and will be null for the current/active record for each subscription.
  - `plan_started_at` is inclusive and will be null for the original plan.
  - `plan_ended_at` is non-inclusive and will only be set on the last history record for each plan.
- Changes the `subscription_platform.stripe_subscriptions` and `subscription_platform.nonprod_stripe_subscriptions` views to select from the associated history views and includes new `plan_started_at`, `plan_ended_at`, and `plan_name` columns.
  - To avoid breaking things these views continue to only output a single record for each Stripe subscription representing its current/active state, but the new `plan_started_at` and `plan_ended_at` columns can be used to differentiate subscriptions whose plan has changed from when the subscription was created.
- Changes the `mozilla_vpn_derived.all_subscriptions_v1` ETL to treat periods when Stripe subscriptions had different plans as separate subscriptions (each subsequent period has `-2`, `-3`, etc. appended to the subscription ID), and adds `original_subscription_id`, `subscription_start_reason`, and `ended_reason` columns.
  - `original_subscription_id` will only be set for subscription records that have been synthesized due to a plan change or a gap in the subscription (it will be null for the original subscription record).
  - `subscription_start_reason` will only be set if the subscription has started (i.e. not in trial period) and its possible values are "Converted Trial", "New", "Resurrected", and "Plan Change".
  - `ended_reason` will only be set if the subscription ended and its possible values are "Cancelled by Customer", "Cancelled by IAP", "Payment Failed", and "Plan Change".
- Changes the `mozilla_vpn_derived.subscription_events_live` view to use the `subscription_start_reason` and `subscription_start_reason` columns from `mozilla_vpn_derived.all_subscriptions_v1` for the `granular_event_type` column logic, so `granular_event_type` will be "Plan Change" for events that occurred due to plan upgrade.

Example views & tables are available in the `moz-srose-dev.ds_2160` dataset.

---
Checklist for reviewer:


sql/moz-fx-data-shared-prod/mozilla_vpn_derived/subscription_events_live/view.sql
```diff
@@ -98,39 +98,21 @@ SELECT
...
-    all_subscriptions.canceled_for_customer_at IS NOT NULL
-    OR all_subscriptions.cancel_at_period_end
-  THEN
-    "Cancelled by Customer"
-  ELSE
-    "Payment Failed"
+    COALESCE(
+      all_subscriptions.ended_reason,
+      IF(provider = "Apple Store", "Cancelled by IAP", "Payment Failed")
+    )
```
relud, line 115: it seems like ended_reason already includes the logic for `IF(provider = "Apple Store", "Cancelled by IAP", "Payment Failed")`, why is it being duplicated here?

sql/moz-fx-data-shared-prod/subscription_platform/nonprod_stripe_subscriptions_history/view.sql
```diff
@@ -0,0 +1,294 @@
...
+      ORDER BY
+        valid_from
+    ) AS lead_previous_plan_id,
+  FROM
+    subscriptions_history
+),
+subscriptions_history_with_plan_ends AS (
+  SELECT
+    *,
+    -- A new `previous_plan_id` value means the previous row was the last row with that plan.
```
relud, line 49: This comment matches the behavior for `LAG`, but query logic here and in subsequent CTEs is consitent with `LEAD`, so:
```suggestion
    -- A new `previous_plan_id` value means the current row is the last row with that plan.
```

sql/moz-fx-data-shared-prod/subscription_platform/nonprod_stripe_subscriptions_history/view.sql
```diff
@@ -0,0 +1,294 @@
...
+    plan_id,
+  FROM
+    `dev-fivetran`.stripe_nonprod.subscription_item
+  -- ZetaSQL requires QUALIFY to be used in conjunction with WHERE, GROUP BY, or HAVING.
+  WHERE
+    TRUE
+  QUALIFY
+    -- With how our subscription platform currently works each Stripe subscription should
+    -- only have one subscription item, and we enforce that so the ETL can rely on it.
+    1 = COUNT(*) OVER (PARTITION BY subscription_id)
```
relud, line 96: Rather than throwing out rows that violate our condition I think we should throw an error, and thus get notified of ETL failing because this assumption is incorrect.
```suggestion
    IF(COUNT(*) OVER (PARTITION BY subscription_id) > 1, ERROR("stripe subscription contains multiple items"), TRUE)
```

relud (approved):
r+wc

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/subscription_events_live/view.sql
sean-rose: Because if an Apple subscription was cancelled and later resurrected its `all_subscriptions.ended_reason` could be null when we're storing the cancelled event here (a situation made more probable by the 7 day delay for VPN ETLs).

That should become less of an issue in the future once your new Apple subscription periods logic from #3329 is in effect that'll produce separate `all_subscriptions` records for the disjoint subscription periods going forward.

sql/moz-fx-data-shared-prod/subscription_platform/nonprod_stripe_subscriptions_history/view.sql
sean-rose: I think the original comment is accurate, though I'll admit the logic of using `LEAD()` with a field named `previous_...` is difficult to explain unambiguously.  The crux is that when a new `previous_plan_id` is set the row has already been updated for the new plan, so we want the previous row to that change happening.

Maybe it'd be clearer to rephrase it from the perspective of the current row?
```suggestion
    -- If the next row has a different `previous_plan_id` then this was the last row with that plan.
```

sql/moz-fx-data-shared-prod/subscription_platform/nonprod_stripe_subscriptions_history/view.sql
sean-rose: Unfortunately that would already fail with the error because a test subscription with multiple subscription items exists in the non-prod environment.  And while we could currently add the error logic in the production view, I'm a little leery of some future test/experimental/accidental Stripe setup breaking the ETL, forcing us to immediately scramble to address it.

Other ideas would be to create a "check" ETL that would run independently and error under the same conditions, or try using something like Great Expectations or the [data integrity monitoring PoC](https://github.com/mozilla/dim/).  Thoughts?

sql/moz-fx-data-shared-prod/subscription_platform/nonprod_stripe_subscriptions_history/view.sql
relud: your suggestion is much more clear, thank you

sql/moz-fx-data-shared-prod/subscription_platform/nonprod_stripe_subscriptions_history/view.sql
relud: no change for now is fine then.

PR #3297, opened 2022-10-19
RS 477 - Change app name from Fennec to Firefox iOS Glean

alekhyamoz:
Changed the app name from Fennec to Firefox iOS Glean
Based on discussions here: https://mozilla-hub.atlassian.net/browse/RS-477
and the proposal https://docs.google.com/document/d/1ZM2gt4dSj5scOcdgQO5ca2vpP7IJBOyVj6HuFfGLd4s/edit#


sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/query.sql
```diff
@@ -260,7 +260,7 @@ metrics_org_mozilla_ios_firefox AS (
     DATE(submission_timestamp) AS submission_date,
     client_info.client_id,
     normalized_country_code AS country,
-    'Fennec' AS app_name,
```
Dexterp37, line 263: Is this dataset the only one "Fennec" is being used in?

kik-kik:
From looking into the code, I believe this is where you will need to make a change:
https://github.com/mozilla/bigquery-etl/blob/main/sql_generators/search/__init__.py#L32

This is because this particular query file gets generated from a templates:
https://github.com/mozilla/bigquery-etl/tree/main/sql_generators/search/templates

sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/query.sql
Dexterp37: @alekhyamoz we should probably check with @scholtzan what happens to the normalized app name if you haven't already, given that this might have downstream consequences.

alekhyamoz: Checked with @scholtzan  and it was confirmed that `normalized_app_name` can remain 'Fennec'

Dexterp37: Thank you for checking!

scholtzan:
Did we check that there is no impact on downstream dependencies, like:
* https://github.com/mozilla/bigquery-etl/blob/9d650106d6ff8bef19e22210c912c8df1ae4ee4a/sql/moz-fx-data-shared-prod/telemetry_derived/unified_metrics_v1/query.sql#L141

Do we need to update something here?
https://github.com/mozilla/bigquery-etl/blob/9d650106d6ff8bef19e22210c912c8df1ae4ee4a/sql/moz-fx-data-shared-prod/search/mobile_search_clients_engines_sources_daily/view.sql#L13

I think @ANich should probably take a look here since he did something very similar for Android earlier.

alekhyamoz:
@scholtzan @ANich 
1. All the downstream tables have `app_name` in the `select` clause but for `unified_metrics`. Though his table filters on `app_name`, I did not see it filter on `Fennec`. 

2. my understanding was that the cut-off for Firefox iOS Glean has already happened. @ANich  please correct me here

Dexterp37:
> @scholtzan @ANich
> 
>     1. All the downstream tables have `app_name` in the `select` clause but for `unified_metrics`. Though this table filters on `app_name`, I did not see it filter on `Fennec`.
> 
>     2. my understanding was that the cut-off for Firefox iOS Glean has already happened. @ANich  please correct me here

Yes, this already happened a long while back. We're [filtering out legacy telemetry data for Firefox iOS version < 28](https://github.com/mozilla/bigquery-etl/blob/1ff69d25f7208603021043687178c71627bf3849/sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/query.sql#L75-L81). Honestly, the ideal course of action here would be, imho, the following:

1. Remove this filtering
2. Filter out Firefox iOS Glean from the view
3. Update the queries to make sure to point to the glean data (and communicate this out!).

This way Xuan et al will be able to check historical data for legacy as well.

In practice, the above thing might not be needed, since we've been filtering out legacy since #2219 and [the proposal](https://docs.google.com/document/d/10-E3-oy4_sqmuTv2IwOxJ3CxI5-YTrjDOFqzwMUGUlg/edit) was approved by @xluo-ds , among the others.

ANich:
> I think @ANich should probably take a look here since he did something very similar for Android earlier.

I don't follow why `app_name` would be renamed for FxiOS? The linked proposal seems to be for apps that haven't yet migrated to Glean. FxiOS data in `mobile_search_clients_daily_v1` has been sourced from only Glean data for a while. Unless that's being revisited?

Besides that, just noting that this will require backfills for this table and its downstreams, from 2021 data to data

Dexterp37:
> I don't follow why `app_name` would be renamed for FxiOS? The linked proposal seems to be for apps that haven't yet migrated to Glean. FxiOS data in `mobile_search_clients_daily_v1` has been sourced from only Glean data for a while. Unless that's being revisited?

Firefox iOS hasn't yet migrated to Glean. As other apps (Focus/Klar Android, Focus/Klar iOS), it's sending both legacy and glean data. A decision was made a while back to filter legacy data out of the search table, but that's pretty much the extent of it.

xluo-ds:
Talked to @alekhyamoz , right now for mobile search tables, I am fine with leaving it under the name of `Fennec` at this moment. If we want to keep naming consistent to other places, we can do it as we officially migrate mobile search metrics to Glean.

Dexterp37:
> Talked to @alekhyamoz , right now for mobile search tables, I am fine with leaving it under the name of `Fennec` at this moment. If we want to keep naming consistent to other places, we can do it as we officially migrate mobile search metrics to Glean.

@xluo-ds what do you mean by officially migrate mobile search metrics to Glean?

xluo-ds:
> 

@Dexterp37  I mean when we make the change to user-facing view of showing only search metrics in Glean so that users can update their queries in one batch if needed.

PR #3302, opened 2022-10-24
Add M1 Mac setup guide

curtismorales:
I've documented here the steps I took to get bqetl working on my M1 Macbook. I'm not sure whether this is the right place for this to live--I imagine there are similar setup steps needed to get other repos running, and it would be nice to have a more general M1 setup guide we can point to. But I've only run through this process specifically to get `bigquery-etl` running, so I thought it made sense to put it here to start.

Any and all feedback is welcome! I've tried not to make this *too* specific to my individual setup, but it is mostly a log of the steps I took to get things working. And I haven't done a whole lot with `bqetl`; there are likely more steps that belong in this guide.

Checklist for reviewer:


M1_MAC_SETUP.md
```diff
@@ -0,0 +1,92 @@
...
+1. Install Rosetta if you don't have it yet
+
+   ```zsh
+   softwareupdate --install-rosetta
+   ```
+
+2. Set Terminal to run in Rosetta mode:
+   1. In Finder, go to Applications -> Utilities
+   2. Right-click on Terminal and choose Get Info
+   3. Check the box "Open using Rosetta"
```
lucia-vargas-a, line 26: Is great to have a step by step guide. If we have the capacity, we could move all these commands to a makefile. We'd need to maintain it with version changes, but it'd save a lot of time during the onboarding setup, and when upgrading the MAC.

M1_MAC_SETUP.md
```diff
@@ -0,0 +1,92 @@
+# M1 Mac Setup Guide
+
+This guide should get you set up to run `bigquery-etl` on an M1 or M2 Mac. The
```
lucia-vargas-a, line 3: Are there other repositories that require this setup? Perhaps a link in the [Data onboarding documentation](https://mana.mozilla.org/wiki/display/DATA/Onboarding) would give more visibility and benefit?

kik-kik: I second this, I believe this is the doc that gets distributed to joiners. I don't think we need a copy of this guide there, but maybe a disclaimer/warning on top of the doc to make all new joiners aware of general compatibility issues on Apple silicon devices so that they can avoid the setup frustrations later.

M1_MAC_SETUP.md
```diff
@@ -0,0 +1,92 @@
...
+   ```zsh
+   brew86 install pyenv openjdk maven
+   brew86 install --cask temurin
+   ```
+
+6. Install python; you'll need to install `xz` first
+
+   ```zsh
+   brew86 install xz
+   pyenv install 3.10.6  # (or whichever version)
```
perrymcmanis144, line 89: 🤔 hmm, I am not sure about this as a default suggestion. I think for folks who are doing work primarily with bq-etl or other repos that need rosetta, it could make a lot of sense & be simpler, but personally I need the arm64 stuff for most of my work (granted I am in a bit of a unique position). 

I *think* most of these tools being in rosetta isn't a problem, the way I had solved this problem was pyenv-alias, but I have had a bit of a running challenge with the sql formatter.

PR #3318, opened 2022-10-31
Add check for contextual_services_derived.event_aggregates_v1

scholtzan:
This adds a check to make sure there are no empty partitions for `contextual_services_derived.event_aggregates_v1`. If empty partitions occur, for example after a backfill goes wrong then the Airflow task will fail and list affected partitions.

I can't test it on the actual dataset since I don't have permissions, so this might be a good time to hand this over to someone on the contextual services team who can test this.

### DIFF
--- a/dags/bqetl_ctxsvc_derived.py
+++ b/dags/bqetl_ctxsvc_derived.py
@@ -82,6 +82,22 @@ with DAG(
         arguments=["--schema_update_option=ALLOW_FIELD_ADDITION"],
     )
 
+    contextual_services_derived__event_aggregates_check__v1 = bigquery_etl_query(
+        task_id="contextual_services_derived__event_aggregates_check__v1",
+        destination_table=None,
+        dataset_id="contextual_services_derived",
+        project_id="moz-fx-data-shared-prod",
+        owner="wstuckey@mozilla.com",
+        email=[
+            "ctroy@mozilla.com",
+            "telemetry-alerts@mozilla.com",
+            "wstuckey@mozilla.com",
+        ],
+        date_partition_parameter="submission_date",
+        depends_on_past=False,
+        sql_file_path="sql/moz-fx-data-shared-prod/contextual_services_derived/event_aggregates_check_v1/query.sql",
+    )
+
     contextual_services_derived__event_aggregates_spons_tiles__v1 = bigquery_etl_query(
         task_id="contextual_services_derived__event_aggregates_spons_tiles__v1",
         destination_table="event_aggregates_spons_tiles_v1",
@@ -161,6 +177,10 @@ with DAG(
 
     contextual_services_derived__event_aggregates__v1.set_upstream(
         wait_for_copy_deduplicate_all
+    )
+
+    contextual_services_derived__event_aggregates_check__v1.set_upstream(
+        contextual_services_derived__event_aggregates__v1
     )
 
     contextual_services_derived__event_aggregates_spons_tiles__v1.set_upstream(
--- a/sql/moz-fx-data-shared-prod/contextual_services_derived/event_aggregates_check_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/contextual_services_derived/event_aggregates_check_v1/metadata.yaml
@@ -1,0 +1,26 @@
+friendly_name: Event Aggregates Check
+description: |-
+  This query checks whether there are any empty partitions in
+  contextual_services_derived.event_aggregates_check_v1 which might
+  indicate that data is missing.
+owners:
+- wstuckey@mozilla.com
+email:
+- rburwei@mozilla.com
+- ctroy@mozilla.com
+- wstuckey@mozilla.com
+- skahmann@mozilla.com
+- xluo@mozilla.com
+labels:
+  incremental: false
+scheduling:
+  dag_name: bqetl_ctxsvc_derived
+  destination_table: null
+  query_file_path:
+    sql/moz-fx-data-shared-prod/contextual_services_derived/event_aggregates_check_v1/query.sql
+  referenced_tables:
+  - [
+    'moz-fx-data-shared-prod',
+    'contextual_services_derived',
+    'event_aggregates_v1'
+  ]
--- a/sql/moz-fx-data-shared-prod/contextual_services_derived/event_aggregates_check_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/contextual_services_derived/event_aggregates_check_v1/query.sql
@@ -1,0 +1,48 @@
+WITH dates AS (
+  -- Generate the date range to check for missing data
+  SELECT
+    *
+  FROM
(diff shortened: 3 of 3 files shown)


sql/moz-fx-data-shared-prod/contextual_services_derived/event_aggregates_check_v1/metadata.yaml
```diff
@@ -0,0 +1,26 @@
...
+  contextual_services_derived.event_aggregates_check_v1 which might
+  indicate that data is missing.
+owners:
+- wstuckey@mozilla.com
+email:
+- rburwei@mozilla.com
+- ctroy@mozilla.com
+- wstuckey@mozilla.com
+- skahmann@mozilla.com
+- xluo@mozilla.com
```
scholtzan, line 13: I just selected these email adresses based on the ones listed for existing tasks.

sql/moz-fx-data-shared-prod/contextual_services_derived/event_aggregates_check_v1/query.sql
```diff
@@ -0,0 +1,48 @@
...
+    -- Based on the information we have about existing partitions
+    -- determine which partitions are either missing or have 0 rows.
+    -- This is cheaper than querying the table every day to get the row count.
+      SELECT
+        *
+      FROM
+        `contextual_services_derived.INFORMATION_SCHEMA.PARTITIONS`
+      WHERE
+        table_name = 'event_aggregates_v1'
+        AND total_rows > 0
```
scholtzan, line 24: So this checks for partitions that are entirely missing or have 0 rows of data (we could also do something like `total_rows > 10`). Not sure if that is the check we want, it would have at least caught the backfill issues we saw

skahmann3: Is it common that a backfill issue can cause <10 rows to show up instead of 100% missingness?

scholtzan: No, that would not indicate a backfill issue, but some issue with the downstream data

skahmann3:
Thank you for doing this, @scholtzan! Re: needing a contextual services DE to run this and check, @quiiver, will you test this solution?

dags/bqetl_ctxsvc_derived.py
```diff
@@ -82,6 +82,22 @@
         arguments=["--schema_update_option=ALLOW_FIELD_ADDITION"],
     )
 
+    contextual_services_derived__event_aggregates_check__v1 = bigquery_etl_query(
```
rebecca-burwei, line 85: We had the incident because two tasks were in the same DAG and one affected the other. Does putting the check also in the same DAG make the check vulnerable to the same issues?

dags/bqetl_ctxsvc_derived.py
```diff
@@ -82,6 +82,22 @@
...
+            "ctroy@mozilla.com",
+            "telemetry-alerts@mozilla.com",
+            "wstuckey@mozilla.com",
+        ],
+        date_partition_parameter="submission_date",
+        depends_on_past=False,
+        sql_file_path="sql/moz-fx-data-shared-prod/contextual_services_derived/event_aggregates_check_v1/query.sql",
+    )
+
     contextual_services_derived__event_aggregates_spons_tiles__v1 = bigquery_etl_query(
```
rebecca-burwei, line 101: @skahmann3 do you think we should have a check for this table and for `contextual_services_derived__event_aggregates_suggest__v1` as well? or do you think just having a check for event aggregates is sufficient? We're going to start deprecating event aggregates soon (not removing, but strongly encouraging users to not use this table).

skahmann3:
@scholtzan - Would the permissions you're talking about mean all contextual services working group DEs would be able to edit/ test this Airflow task? Or would permissions be granted temporarily and on a case-by-case basis? It would be preferred if any CS DE could edit/ test this task. 

Also, related to a comment from @rebecca-burwei above, once we sort out the best way to write this check for more blanket permissions, could we expand the check to rows >0 for tiles and rows >0 for suggest data? event_aggregates is split by product with source ("topsites", "suggest") or we have the individual tables (event_aggregates_spons_tiles, event_aggregates_suggest). Not sure which way is easier for expanding the check

dags/bqetl_ctxsvc_derived.py
rebecca-burwei: @skahmann3 thanks for your thoughts!

@scholtzan We'd like to check that the event aggregates table has >0 rows for each partition and each value of `source` (a field in the event aggregates table that can either be `topsites` or `suggest`). Would it be possible to do that?

whd:
> The permissions I was referring to were related to having access to INFORMATION_SCHEMA.PARTITIONS which is getting worked on in https://mozilla-hub.atlassian.net/browse/DSRE-1094 it seems

While the problem doesn't have an obvious solution in general I did file https://mozilla-hub.atlassian.net/browse/DSRE-1095 for fixing this specific use case. I expect that to land some time tomorrow Pacific.

whd:
https://github.com/mozilla/bigquery-etl/pull/3323 landed which should grant `workgroup:contextual-services` access to run the query on `INFORMATION_SCHEMA` that was previously failing.

skahmann3:
Thanks whd and Anna! @quiiver will you take this on? The permissions issue should be resolved, but we'll need someone with contextual services permission to test that Anna's solution works. Also, Rebecca suggested an extension in the comments above for warnings for each product (suggest vs tiles) that needs to be implemented/ tested

quiiver:
I tested out the query with the following for `@submission_date`: `date_sub(current_date(), interval 1 day)`

It returned the following error:
```
Partitions with data missing: 2021-12-13 
```

Is that what we expect @skahmann3 @scholtzan

skahmann3:
LOL @quiiver we are actually missing December 13, 2021, so seems like it works to me! Here's Looker showing the [missing date](https://mozilla.cloud.looker.com/x/uVzWNDwunDBHAhK2gQe3mw). I'll sync with the other DEs to see if we have that date in any of their backup copies so we can re-populate. Otherwise, should be good to go to extend?

dags/bqetl_ctxsvc_derived.py
scholtzan: It will be easier to check the `event_aggregates_suggest` and `event_aggregates_tiles` separately since the partition information which we use here doesn't have access to the `source` field values.

I'll update the check to test both of these datasets.

dags/bqetl_ctxsvc_derived.py
scholtzan: I don't think it will... The check is configured to run after the ETL for processing these events has completed.

scholtzan:
I did some testing and added `event_aggregates_spons_tiles_v1` and `event_aggregates_suggest_v1` to the check. I also added `2021-12-13` for `event_aggregates_v1` to be ignored by the checks since it is known to have no data.

Should be good to go

quiiver (approved):
looks good. thanks @scholtzan for picking this back up

Issue #3299, opened 2022-10-20
Add Nimbus branch value to moz-fx-data-experiments.monitoring.experimenter_experiments_v1


scholtzan:
Follow-up to: https://mozilla.slack.com/archives/C4D5ZA91B/p1666195091035389

Add the Nimbus branch value field here: https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-experiments/monitoring/experimenter_experiments_v1/query.py

cc @rebecca-burwei

PR #3325, opened 2022-11-01
Add branch values for experiments

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/3299

This adds a new `value` field to each branch. Branch values pulled from the API can either be strings or JSON objects, so the field type here is JSON which allows for queries like:

```sql 
select 
  branch.value.firefoxSuggestLabelsEnabled
from `moz-fx-data-experiments.monitoring.experimenter_experiments_v1`, UNNEST(branches) AS branch
where normandy_slug = "smarter-suggestions-in-the-awesomebar"
```


ANich (approved):
LGTM, left one suggestion. Caveat: I don't have much expmt context 😝)

PR #3331, opened 2022-11-02
separate the test windows for mean and range data validation checks

chelseatroy:
OKAY!

We are replacing one column, which used to dictate the test windows for both mean and range, with two columns, that allow those to be adjusted separately.

We are removing the old column because keeping it would be confusing. We can move the old data into the new columns to preserve clarity and accuracy.

This is a new table with zero dependencies that currently only is used for monitoring by three people who are all aware this change is being made. So, though it is technically backward-breaking due to.the removal of a column, it is "safe" and should not require a change request or a new version of the table.

There is some SQL that needs to be run in prod to get the schema there to match this change. Here it is, divided up into teeny parts for legibility and risk mitigation:

```
ALTER TABLE `moz-fx-data-shared-prod.search_terms_derived.search_term_data_validation_reports_v1`
ADD COLUMN range_test_window_num_days integer;

UPDATE `moz-fx-data-shared-prod.search_terms_derived.search_term_data_validation_reports_v1`
SET range_test_window_num_days = test_window_num_days;

ALTER TABLE `moz-fx-data-shared-prod.search_terms_derived.search_term_data_validation_reports_v1`
ADD COLUMN mean_test_window_num_days integer;

UPDATE `moz-fx-data-shared-prod.search_terms_derived.search_term_data_validation_reports_v1`
SET mean_test_window_num_days = test_window_num_days;

ALTER TABLE `moz-fx-data-shared-prod.search_terms_derived.search_term_data_validation_reports_v1`
DROP COLUMN test_window_num_days;
```

Thank you! 

Checklist for reviewer:


whd (commented):
I'm comfortable given the context with running the commands in this PR but would like @mreid-moz to confirm or weigh in on the slack thread or here before proceeding. That will also mostly serve as change control as the ultimate approval is from the same source (but without a corresponding jira ticket).

I'll note that using [DELETE](https://cloud.google.com/bigquery/docs/managing-table-schemas#delete_a_column) is a beta feature of BQ and therefore our usage should be governed by https://docs.google.com/document/d/1pvIiyk7Kwr6C66LlPzOhjlIq8gQkZCw8I4OIYohazD8/edit. If this table were in a more critical path I would have some concerns about using it, but for this case I think it's fine.

mreid-moz:
This sounds good go me 👍

chelseatroy:
> I assume `WHERE true` is the intent but want to double check before I run it.

@whd correct! Thank you :)

whd (approved):
All SQL statements have been run.

PR #3343, opened 2022-11-07
Have Glean stable views parse datetime metrics as timestamps (bug 1771733)

sean-rose:
## [Bug 1771733](https://bugzilla.mozilla.org/show_bug.cgi?id=1771733): Map Glean 'datetime' type to a 'datetime' in bigquery views

Datetime metrics are stored in BigQuery tables as strings, which is not very user friendly.  This will change the generated stable views to replace those string fields with timestamp fields, while still providing the original string values in fields prefixed with `raw_`.  For example, a `metrics.datetime.foo` string field would be replaced with a timestamp field, and the string value would be available in an additional `metrics.datetime.raw_foo` field.

Note that despite the title of the bug, we're parsing the datetime metrics as _timestamps_ to preserve their accuracy in representing an exact point in time.  However, all BigQuery timestamps are inherently in UTC, so the original time zone isn't readily available, but could be retrieved from the original string value if needed.

---
Checklist for reviewer:

### DIFF
--- a/sql/mozfun/glean/parse_datetime/metadata.yaml
+++ b/sql/mozfun/glean/parse_datetime/metadata.yaml
@@ -1,0 +1,6 @@
+---
+description: |
+  Parses a Glean datetime metric string value as a BigQuery timestamp.
+
+  See https://mozilla.github.io/glean/book/reference/metrics/datetime.html
+friendly_name: Parse Datetime
--- a/sql/mozfun/glean/parse_datetime/udf.sql
+++ b/sql/mozfun/glean/parse_datetime/udf.sql
@@ -23,6 +23,7 @@ SELECT
     glean.parse_datetime('2000-01-02T03:04:05.123+00:00')
   ),
   assert.equals(TIMESTAMP '2000-01-02 03:04:05', glean.parse_datetime('2000-01-02T03:04:05+00:00')),
+  assert.equals(TIMESTAMP '2000-01-02 09:04:05', glean.parse_datetime('2000-01-02T03:04:05-06:00')),
   assert.equals(TIMESTAMP '2000-01-02 03:04:00', glean.parse_datetime('2000-01-02T03:04+00:00')),
   assert.equals(TIMESTAMP '2000-01-02 03:00:00', glean.parse_datetime('2000-01-02T03+00:00')),
   assert.equals(TIMESTAMP '2000-01-02 00:00:00', glean.parse_datetime('2000-01-02+00:00')),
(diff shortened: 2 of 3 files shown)


sean-rose:
Still TBD:
- Whether there are any downstream impacts (e.g. ETLs, Looker).
- Whether a special deployment process is needed given that this introduces and relies on a new UDF.
  - Maybe the fact that it's referenced in generated SQL rather than checked in SQL means the CI/deployment won't choke on the UDF not already existing?

badboy:
Thanks for this! I was still dealing with understanding a bit more about bigquery-etl before I would have achieved at a solution.

sean-rose:
> Still TBD:
> * Whether there are any downstream impacts (e.g. ETLs, Looker).

As best I can tell from searching the code in [bigquery-etl](https://github.com/mozilla/bigquery-etl/tree/generated-sql), [looker-hub](https://github.com/mozilla/looker-hub), and [looker-spoke-default](https://github.com/mozilla/looker-spoke-default) no datetime metrics are currently being used in ETLs or Looker (at least Looker content that's saved in GitHub).

scholtzan (approved):

sean-rose:
> > Still TBD:
> > 
> > * Whether there are any downstream impacts (e.g. ETLs, Looker).
> 
> As best I can tell from searching the code in [bigquery-etl](https://github.com/mozilla/bigquery-etl/tree/generated-sql), [looker-hub](https://github.com/mozilla/looker-hub), and [looker-spoke-default](https://github.com/mozilla/looker-spoke-default) no datetime metrics are currently being used in ETLs or Looker (at least Looker content that's saved in GitHub).

Since not all Looker content is saved in GitHub I looked at the `mozdata` BigQuery project's query history for the last month and only found two instances of reporting queries referencing datetime metrics:
- Nov 15: a couple of Looker queries on `fenix.addresses_sync` and `fenix.bookmarks_sync` by Looker user ID 777.
- Nov 10: a series of Redash queries on `firefox_ios.baseline` by @badboy that ended up saved as [STMO query 88603](https://sql.telemetry.mozilla.org/queries/88603/source), but which is now archived.

Based on that I feel fairly confident this change won't break existing reporting, and will proceed with posting a PSA in the [#data](https://mozilla.slack.com/archives/CKFKC7Y1Y) channel and then merge this ~24 hours later.

sean-rose:
> Still TBD:
>   * Whether a special deployment process is needed given that this introduces and relies on a new UDF.
>     * Maybe the fact that it's referenced in generated SQL rather than checked in SQL means the CI/deployment won't choke on the UDF not already existing?

It turns out that although the `generate-sql` CI step is considered passing, it's actually failing to generate `schema.yaml` files for the affected views because dry-running the view SQL fails due to the `mozfun.glean.parse_datetime` UDF not existing yet.

Therefore my plan is to manually create the `mozfun.glean.parse_datetime` UDF just prior to merging this PR so the SQL generation and views deployment will succeed when it's merged.

PR #3350, opened 2022-11-09
Created dervied table for big query job usage reporting

wwyc:
Checklist for reviewer:


sql/moz-fx-data-shared-prod/usage_reporting/view.sql
```diff
@@ -0,0 +1,7 @@
+CREATE OR REPLACE VIEW
```
scholtzan, line 1: This view should be in `sql/moz-fx-data-shared-prod/monitoring/bigquery_usage`

sql/moz-fx-data-shared-prod/usage_reporting_derived/bq_usage_v1/query.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python3
```
scholtzan, line 1: Instead of creating a new dataset `usage_reporting_derived`, I think this query fits well into the `monitoring_derived` dataset. In fact there is already one that looks like it's doing something very similar: https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/monitoring_derived/bigquery_etl_scheduled_query_usage_v1/query.py
Maybe we can update the existing query to add the fields that have been added here (like the errors)?

wwyc: Sounds good, I can moved it into the `monitoring_derived` dataset.

However, the `bigquery_etl_scheduled_query_usage_v1` query loops through datasets and tables that only exist in the repo.  I am hoping to look at usage across everything that is available.  It is also limited to statement_type = "SELECT" and includes the reference table field without the destination table field.  It seems like the difference would impact those who are currently using the table.

scholtzan: I'm not aware of anyone using the table, so I think it's fine to update the query. If anything the same information plus some additional tables that are not part of the repo should be available in the table.

wwyc: Currently there isn’t a reliable way to confirm if anyone is using `bigquery_etl_scheduled_query_usage_v1`.  Since my query is quite different from it, I hesitate to make changes and affect any users who may be using this table.  The work I am doing is to eventually be able to capture all usage information.  I can first create this new table and once I have the usage data modelling work completed, I can then come back to deprecate `bigquery_etl_scheduled_query_usage_v1`?

gkabbz: To avoid blocking on deprecating this particular table, I've made a jira ticket [DENG-503](https://mozilla-hub.atlassian.net/browse/DENG-503?atlOrigin=eyJpIjoiNDg3MTZjMDA3MmEwNGM1MDlhZmI4YTVjNjViOWY3NTgiLCJwIjoiaiJ9) under the epic to ensure we don't miss the step to deprecate the old table. Given the naming of that particular table, I do think we would need to update it's name as this new table is meant to look at all our querying of BQ not just the scheduled queries.

gkabbz (approved):

alekhyamoz (approved):
lgtm

sql/moz-fx-data-shared-prod/monitoring_derived/bigquery_tables_inventory_v1/query.py
```diff
@@ -0,0 +1,63 @@
...
+
+    # remove old partition in case of re-run
+    client = bigquery.Client(args.project)
+    client.delete_table(destination_table, not_found_ok=True)
+
+    for project in args.source_projects:
+        client = bigquery.Client(project)
+        query = create_query(args.date, project)
+        job_config = bigquery.QueryJobConfig(
+            destination=destination_table, write_disposition="WRITE_APPEND"
```
scholtzan, line 57: It looks like every execution is based on a date. Instead of manually removing the partition and then appending to it, we could simplify it by using `WRITE_TRUNCATE` to just overwrite the partition

Issue #3351, opened 2022-11-10
Implement checks for retention policies in derived table queries


scholtzan:
We have different retention policies for different source data sets. After a certain amount of time data can get deleted. This could pose a problem when running a backfill. Currently, derived table queries do not have any checks implemented to prevent backfills from running on dates where data has already been deleted. This can result in data being deleted if not detected in time.

There might be a way to implement automatic checks by pulling in the retention duration from the schema metadata. A more short-term workaround is to manually add these checks to the ETL queries by adding something like this: 

```sql
IF (DATE_DIFF(DATE('@submission_date'), CURRENT_DATE, DAY) > 30, ERROR("Data older than 30 days has been removed"), NULL)
```

Errors being thrown will stop the query from running and will prevent partitions getting overwritten

scholtzan:
cc @alekhyamoz

PR #3352, opened 2022-11-10
Add check to contextual_services_derived.event_aggregates_v1 to prevent backfills on deleted partitions

scholtzan:
I just had an idea to prevent backfills that accidentally remove old partitions from running. This is related to what is described in https://github.com/mozilla/bigquery-etl/issues/3351

Some of our source data sets only keep data for a certain period of time (contextual services 30 days??), so running a backfill on derived tables on anything older than that will overwrite partitions with no data. It is not always apparent to those who run the backfills what the retention policies are, so we should be explicit about them in the query and even prevent the query from running if upstream data has been deleted. This PR shows how this can be done in the case of  `contextual_services_derived.event_aggregates_v1`. I don't know if there are other derived queries that might be affected where we should implement this?


sql/moz-fx-data-shared-prod/contextual_services_derived/event_aggregates_v1/query.sql
```diff
@@ -278,6 +278,11 @@ WHERE
   submission_date = @submission_date
   -- Filter out events associated with suspiciously active clients.
   AND NOT (user_event_count > 50 AND event_type = 'click')
+  AND IF(
+    DATE_DIFF(CURRENT_DATE(), DATE('@submission_date'), DAY) > 30,
```
scholtzan, line 282: Someone should check whether 30 days is the retention policy here

sql/moz-fx-data-shared-prod/contextual_services_derived/event_aggregates_v1/query.sql
```diff
@@ -278,6 +278,11 @@ WHERE
...
+    ERROR("Data older than 30 days has been removed"),
```
scholtzan, line 283: once an error is thrown, the query will stop executing and not write anything to the destination table

sql/moz-fx-data-shared-prod/contextual_services_derived/event_aggregates_v1/query.sql
whd: It is: https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/main/templates/contextual-services/defaults.schema.json#L4

skahmann3:
@scholtzan love this idea! And thanks for continuing to think about how we could potentially lose all our tiles/ suggest data #yikes

rebecca-burwei:
@scholtzan love this. I added it to 2 more tables.

PR #3363, opened 2022-11-16
[Bug 1800538] Workgroup access for workgroup:pocket/external

scholtzan:
For https://bugzilla.mozilla.org/show_bug.cgi?id=1800538
I'm not the owner of these datasets, but Kirill asked me for help here.

cc @kirill-demtchouk

### DIFF
--- a/sql/moz-fx-data-shared-prod/activity_stream/impression_stats_by_experiment/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/activity_stream/impression_stats_by_experiment/metadata.yaml
@@ -1,0 +1,5 @@
+friendly_name: Impressions stats by experiment
+workgroup_access:
+  - role: roles/bigquery.dataViewer
+    members:
+      - workgroup:pocket/external
--- a/sql/moz-fx-data-shared-prod/activity_stream/impression_stats_flat/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/activity_stream/impression_stats_flat/metadata.yaml
@@ -1,0 +1,5 @@
+friendly_name: Impressions stats flat
+workgroup_access:
+  - role: roles/bigquery.dataViewer
+    members:
+      - workgroup:pocket/external
--- a/sql/moz-fx-data-shared-prod/pocket/spoc_tile_ids/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/pocket/spoc_tile_ids/metadata.yaml
@@ -3,3 +3,7 @@ description: |-
   Lookup table of the currently registered tile IDs for Pocket sponsored content
 owners:
   - jklukas@mozilla.com
+workgroup_access:
+  - role: roles/bigquery.dataViewer
+    members:
+      - workgroup:pocket/external
--- a/sql/moz-fx-data-shared-prod/static/normal_distribution/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/static/normal_distribution/metadata.yaml
@@ -1,0 +1,5 @@
+friendly_name: Normal distribution
+workgroup_access:
+  - role: roles/bigquery.dataViewer
+    members:
+      - workgroup:pocket/external
--- a/sql/moz-fx-data-shared-prod/activity_stream/events/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/activity_stream/events/metadata.yaml
@@ -1,14 +1,11 @@
----
-friendly_name: Historical Pings for `activity-stream/impression-stats`
+friendly_name: Historical Pings for `activity-stream/sessions`
 description: |-
   A historical view of pings sent for the
-  `activity-stream/impression-stats`
+  `activity-stream/sessions`
   document type.
-
   This view is guaranteed to contain only complete days
   (per `submission_timestamp`)
   and to contain only one row per distinct `document_id` within a given date.
-
   Clustering fields: `normalized_channel`, `sample_id`
 labels:
   authorized: true
--- a/sql/moz-fx-data-shared-prod/activity_stream/on_save_recs/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/activity_stream/on_save_recs/metadata.yaml
@@ -1,14 +1,11 @@
----
-friendly_name: Historical Pings for `activity-stream/impression-stats`
+friendly_name: Historical Pings for `activity-stream/on-save-recs`
 description: |-
   A historical view of pings sent for the
-  `activity-stream/impression-stats`
+  `activity-stream/on-save-recs`
   document type.
-
   This view is guaranteed to contain only complete days
   (per `submission_timestamp`)
   and to contain only one row per distinct `document_id` within a given date.
-
   Clustering fields: `normalized_channel`, `sample_id`
 labels:
   authorized: true
--- a/sql/moz-fx-data-shared-prod/activity_stream/pocket_button/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/activity_stream/pocket_button/metadata.yaml
@@ -1,14 +1,11 @@
----
-friendly_name: Historical Pings for `activity-stream/impression-stats`
+friendly_name: Historical Pings for `activity-stream/pocket-button`
 description: |-
   A historical view of pings sent for the
-  `activity-stream/impression-stats`
+  `activity-stream/pocket-button`
   document type.
-
   This view is guaranteed to contain only complete days
   (per `submission_timestamp`)
   and to contain only one row per distinct `document_id` within a given date.
-
   Clustering fields: `normalized_channel`, `sample_id`
 labels:
   authorized: true
--- a/sql/moz-fx-data-shared-prod/activity_stream/sessions/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/activity_stream/sessions/metadata.yaml
@@ -1,14 +1,11 @@
----
-friendly_name: Historical Pings for `activity-stream/impression-stats`
+friendly_name: Historical Pings for `activity-stream/sessions`
 description: |-
   A historical view of pings sent for the
-  `activity-stream/impression-stats`
+  `activity-stream/sessions`
   document type.
-
   This view is guaranteed to contain only complete days
   (per `submission_timestamp`)
   and to contain only one row per distinct `document_id` within a given date.
-
   Clustering fields: `normalized_channel`, `sample_id`
 labels:
   authorized: true
--- a/sql/moz-fx-data-shared-prod/activity_stream/spoc_fills/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/activity_stream/spoc_fills/metadata.yaml
@@ -1,14 +1,11 @@
----
-friendly_name: Historical Pings for `activity-stream/impression-stats`
+friendly_name: Historical Pings for `activity-stream/spoc-fills`
 description: |-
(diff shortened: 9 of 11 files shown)


sql/moz-fx-data-shared-prod/activity_stream/impression_stats_by_experiment/metadata.yaml
```diff
@@ -0,0 +1,5 @@
+friendly_name: Impressions stats by experiment
+workgroup_access:
+  - role: roles/bigquery.dataViewer
+    members:
+      - workgroup:pocket/external
```
scholtzan, line 5: `workgroup_access` has been manually added for the downstream views `pocket.spoc_tile_ids` and `activity_stream_bi.impression_stats_by_experiment_v1`

sql/moz-fx-data-shared-prod/activity_stream/impression_stats_flat/metadata.yaml
```diff
@@ -0,0 +1,5 @@
+friendly_name: Impressions stats flat
+workgroup_access:
+  - role: roles/bigquery.dataViewer
+    members:
+      - workgroup:pocket/external
```
scholtzan, line 5: `workgroup_access` has been manually added for the downstream views `pocket.spoc_tile_ids` and `activity_stream_bi.impression_stats_by_experiment_v1`

sql/moz-fx-data-shared-prod/static/normal_distribution/metadata.yaml
```diff
@@ -0,0 +1,6 @@
+friendly_name: Normal distribution
+workgroup_access:
+  - role: roles/bigquery.dataViewer
+    members:
+      - workgroup:mozilla-confidential
+      - workgroup:pocket/external
```
scholtzan, line 6: Is metadata and `workgroup_access` supported for static datasets? Or should I add this in the cloudops-infra repo instead? https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/tf/prod/envs/prod/bigquery-new/namespaces.auto.tfvars.json

sql/moz-fx-data-shared-prod/activity_stream/dataset_metadata.yaml
```diff
@@ -10,3 +10,4 @@ workgroup_access:
   members:
   - workgroup:mozilla-confidential
   - workgroup:pocket/prefect
+  - workgroup:pocket/external
```
scholtzan, line 13: The docs mention: 

> If a table does not have workgroup_access metadata specified, its access will be determined by the workgroup_access on its dataset. Datasets are guaranteed to have workgroup_access metadata.

So would it make sense to have 

```yaml
labels:
  authorized: true
workgroup_access:
  - role: roles/bigquery.dataViewer
    members:
      - workgroup:pocket/external
```
in the `dataset_metadata.yaml` and removing the manually added metadata files? Or does this cause problems here? Also does the same apply to `authorized: true` being applied to the tables?

whd: The docs are a bit confusing here. Dataset `workgroup_access` and configuration are not inherited by underlying tables i.e. dataset-level `workgroup_access` is not a default value for a table's `workgroup_access`. Table ACLs are only deployed when they are specified, and are additive to dataset access, so the docs are saying that if table ACLs are not specified a table will be accessible only by its dataset ACLs, and that these ACLs will always exist.

`labels: authorized: true` is only for authorized views currently. Authorized datasets exist but we don't currently support them in general: https://mozilla-hub.atlassian.net/browse/DSRE-947, but that metadata would look something like https://github.com/mozilla/bigquery-etl/pull/3177/files.

The issue with granting this access at the dataset level is that the original purpose of `activity_stream_bi` as a separate dataset was to avoid granting blanket access to all tables in the `activity_stream` dataset: https://bugzilla.mozilla.org/show_bug.cgi?id=1616895#c1. Any dataset-level grant will apply to all tables regardless of table configuration. See the top-level review comment for more on that.

In practice it would be much simpler to switch to [uniform dataset access](https://mozilla-hub.atlassian.net/wiki/spaces/SRE/pages/27924789/Data+Access+Workgroups#DataAccessWorkgroups-Exampledatasetwithuniformspecificworkgroupaccess(creatingadatasetthatisaccessibletoaspecificworkgroup)) which is what this line does. This appears to be how we implemented pocket/prefect access (DSRE-532). I don't know what prefect is but this may be the time to revisit the policy decision on what data can/should be accessible to pocket's various systems. However, we don't really have an owner of this data that I'm aware of to make these kinds of policy decisions, and I probably should have sorted that out before implementing DSRE-532.

All to say: if we can switch to a single dataset grant for `pocket/external` then we can greatly simplify our configuration, at the expense of mode having access to a broader set of data than it had before. This would require somebody to make an ownership/policy decision and I don't know who that person is.

sql/moz-fx-data-shared-prod/activity_stream/impression_stats_by_experiment/metadata.yaml
whd: downstream views and downstream tables looks like :+1:

sql/moz-fx-data-shared-prod/static/normal_distribution/metadata.yaml
whd: `workgroup_access` is supported for static datasets.

sql/moz-fx-data-shared-prod/pocket/spoc_tile_ids/metadata.yaml
```diff
@@ -3,3 +3,8 @@ description: |-
   Lookup table of the currently registered tile IDs for Pocket sponsored content
 owners:
   - jklukas@mozilla.com
+workgroup_access:
+  - role: roles/bigquery.dataViewer
+    members:
+      - workgroup:mozilla-confidential
```
whd, line 9: A note about `workgroup:mozilla-confidential` in table ACLs: we generally prefer not to specify these when the underlying dataset is accessible to `mozilla-confidential`. This used to be because table ACLs weren't automated and `mozilla-confidential` changes often but it also makes it clear that table ACLs are additive and avoiding the redundancy is desirable. `workgroup:mozilla-confidential` in table ACLs should be used when granting access to a specific table in a dataset that is not accessible to `mozilla-confidential`, but we have comparatively few of these cases.

So I recommend removing `workgroup:mozilla-confidential` from the table ACLs in this PR.

sql/moz-fx-data-shared-prod/activity_stream/dataset_metadata.yaml
whd: Additional note: this requires a corresponding PR to https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/tf/prod/envs/prod/bigquery-new/namespaces.auto.tfvars.json#L11 so that metadata is synchronized between ops logic and bqetl.

whd (commented):
I think we should probably step back and figure out if we can switch to uniform dataset access for this SA. If we do need to handle this access with the current granularity, then I think this PR looks correct, but I think we would probably want to avoid the dataset-level grant on the views dataset and instead specifically grant it on the views that are also authorized (already part of this PR). Otherwise we're relying on lack of authorizations on some tables in a way that can be semantically confusing (i.e. we're granting `pocket/external` access to all `activity_stream` views, relying on the fact that there are no tables in the dataset and that it does not have access to the underlying data for non-authorized views).

sql/moz-fx-data-shared-prod/activity_stream/dataset_metadata.yaml
scholtzan: > This would require somebody to make an ownership/policy decision and I don't know who that person is.
I brought this up to Rob yesterday, who will look for a new owner

mreid-moz: > All to say: if we can switch to a single dataset grant for pocket/external then we can greatly simplify our configuration, at the expense of mode having access to a broader set of data than it had before. This would require somebody to make an ownership/policy decision and I don't know who that person is.

Sounds like the original concern from [Bug 1616895](https://bugzilla.mozilla.org/show_bug.cgi?id=1616895#c1) was around granting Mode access to data containing `client_id` values. Would that become the case if we switch to a single dataset grant?
If **not**, we can safely switch to that approach. 
If **so**, I think we should land the more complex config as it's currently being proposed, and open a separate conversation about whether sharing `client_id`s with Mode should be allowed. As part of that conversation, we can also track where that requirement stems from (looks like an RRA from quite a few years ago).

Are other policy concerns that need to be addressed here?

scholtzan: The `events` and `sessions` tables under `activity_stream` do look to have telemetry `client_id`. So I guess we'll go ahead without switching to uniform data access for now.

kirill-demtchouk: @mreid-moz @scholtzan How can I help further the conversation on whether Mode is allowed to use `client_id`? Kenny Long and I would love to see if we can revisit that RRA from a long time ago.

kirill-demtchouk: Also, when you refer to the "complex config", I understand that means granting Mode access at the table/view level vs. the schema level, but does that have any bearing on whether Mode gets access to the client_id?

whd: > The events and sessions tables under activity_stream do look to have telemetry client_id. 

> Sounds like the original concern from [Bug 1616895](https://bugzilla.mozilla.org/show_bug.cgi?id=1616895#c1) was around granting Mode access to data containing client_id values.

This feels like a policy blocker to me. I can review the PR as if we're requiring granular access, but the granular access includes `events` which contains `client_id`. I agree with @kirill-demtchouk that furthering the conversation on whether Mode is allowed to use `client_id` is the way to proceed here.

So the options as I see them are:

1. have a policy conversation that ends in allowing `client_id` access for Mode, in which case this PR can be simplified to use only uniform dataset access
2. have a policy conversation that ends in retaining the same policy about `client_id` access for Mode, in which case this PR should not include tables that have that column. If derived versions e.g. `select * EXCEPT (client_id)` of those tables are desirable, they could be added to `activity_stream_bi` as authorized views

mreid-moz: @kirill-demtchouk Yes, I think we should revisit that decision. Do you have access to the original RRA doc (linked in [Bug 1486889](https://bugzilla.mozilla.org/show_bug.cgi?id=1486889))? If so please share it with me. If not, please [request a new RRA](https://infosec.mozilla.org/guidelines/risk/rapid_risk_assessment) and include me in the meeting. Thanks!

kirill-demtchouk: @mreid-moz Done! I shared 1486889 with you. Is there anything else you need from me to revisit the RRA?

mreid-moz: @kirill-demtchouk I still haven't been able to access the original RRA document linked from Bug 1486889. Please request a new RRA, and add me to the attendee list.

PR #3370, opened 2022-11-21
feat(DENG-537): tweaked fxa queries to support partitioned tables instead of sharded tables - PROD

kik-kik:
# feat(DENG-537): tweaked fxa queries to support partitioned tables instead of sharded tables

This is to address out transition from sharded tables to partitioned tables for our FxA log tables.

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_accounts/fxa_all_events/view.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts/fxa_all_events/view.sql
@@ -28,8 +28,7 @@ WITH fxa_auth_events AS (
   -- but should always be included for a complete raw event log.
 fxa_auth_bounce_events AS (
   SELECT
-    `timestamp`,
-    receiveTimestamp,
+    timestamp AS submission_timestamp,
     jsonPayload.fields.user_id,
     CAST(
       NULL AS STRING
@@ -48,8 +47,7 @@ fxa_auth_bounce_events AS (
 ),
 fxa_content_events AS (
   SELECT
-    `timestamp`,
-    receiveTimestamp,
+    timestamp AS submission_timestamp,
     jsonPayload.fields.user_id,
     jsonPayload.fields.country,
     jsonPayload.fields.language,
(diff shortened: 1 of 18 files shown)


sql/moz-fx-data-shared-prod/firefox_accounts_derived/docker_fxa_admin_server_sanitized_v1/inti.sql
```diff
@@ -23,4 +23,4 @@ SELECT
 FROM
   `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_admin_server_20*`
 WHERE
-  PARSE_DATE('%y%m%d', _TABLE_SUFFIX) >= DATE("2022-08-01")
+  DATE(`timestamp`) = "2022-08-01"
```
fbertsch, line 26: this file is called `inti.sql`, can you change it while we're here?

sql/moz-fx-data-shared-prod/firefox_accounts_derived/docker_fxa_admin_server_sanitized_v1/query.sql
```diff
@@ -18,4 +18,4 @@ SELECT
 FROM
   `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_admin_server_20*`
```
fbertsch, line 19: These all need to be changed as well to e.g. `docker_fxa_admin_server` (or `docker_fxa_admin_server_v1` if there's going to be versioning, not sure)

sql/moz-fx-data-shared-prod/firefox_accounts/fxa_all_events/view.sql
```diff
@@ -19,7 +19,7 @@ WITH fxa_auth_events AS (
   -- but should always be included for a complete raw event log.
 fxa_auth_bounce_events AS (
   SELECT
-    timestamp AS submission_timestamp,
+    `timestamp` AS submission_timestamp,
```
fbertsch, line 22: Do we need backticks around every timestamp?

kik-kik: I believe this is a good practice since `timestamp` would also be a keyword so this way we make sure it is not treated as such.

fbertsch (approved):
r+ assuming the upstream tables are ready

PR #3371, opened 2022-11-21
feat(DENG-537): tweaked fxa queries to support partitioned tables instead of sharded tables - STAGING

kik-kik:
# feat(DENG-537): tweaked fxa queries to support partitioned tables instead of sharded tables - STAGING


fbertsch (changes requested):
- needs table names changed
- optional, remove backticks around timestamp

fbertsch (approved):
LGTM

PR #3373, opened 2022-11-21
Add experimentation information to unified metrics

### LINKED ISSUE #3095: Adding experiments info to unified_metrics
With unified_metrics, we have a standard way to calculate CDOU. However, unified_metrics cannot be directly used to calculate AdM CDOU as AdM usage requires subsetting to clients within specific experiments. 

For this use case, unified_metrics requires experiment key + branch information to be included at the client-level.

alekhyamoz:
fixes #3095

To do:
Deploy the schema updated to BQ
Create a PR to update the view


skahmann3 (changes requested):
Thanks for working on this, Alekhya! I can't query the validation tables you have linked, but I expect that if clients can have multiple entries per day, this will inflate downstream CDOU calculations. Desired output would be similar to clients_daily where one client can have multiple experiments contained within the singular entry for client x day. 

Example query: 
SELECT *
FROM `moz-fx-data-shared-prod.telemetry.clients_daily`
WHERE submission_date = "2022-11-01" AND sample_id = 1 AND client_id = "bc007846-0a00-44d2-835a-2139bf5f22e5"

alekhyamoz:
@skahmann3  I have moved the datasets to `moz-fx-data-shared-prod.tmp` and the tables are` ak_test_um_experiments ` and `ak_test_um_withoutexperiments`. Please let me know if this is still an issue. Will move them to mozdata analysis dataset.

skahmann3:
@alekhyamoz I get permissions errors when I try to query `moz-fx-data-shared-prod.tmp.ak_test_um_experiments` and `moz-fx-data-shared-prod.tmp.ak_test_um_withoutexperiments`. I know I'm able to query mozdata.analysis, so can you move the tables there?

skahmann3 (changes requested):
This mostly looks good though. One thing to change: the naming in the experiments field needs to be key-value and not key-branch to be compatible with the [mozfun.map](https://mozilla.github.io/bigquery-etl/mozfun/map/) functions.

 A couple comments from validation:
- Desktop exactly matches when running this [test query](https://console.cloud.google.com/bigquery?sq=687003905036:024fd2de0b1747d6af271e9a3daa9e7f) against unified metrics. Mobile CDOU is up 2.3% in the new data.
- After unnesting experiments and filtering to experiment.key = "firefox-android-sponsored-shortcuts-experiment", the client count totals are similar to those reported in that [experiment's monitoring dashboard](https://mozilla.cloud.looker.com/dashboards/216?Experiment=firefox-android-sponsored-shortcuts-experiment&Time+Range+%5BUTC%5D=2022%2F10%2F10). Client counts are up 2.6% in the new data.

This deviation seems within the realm of normal. I'll approve the merge after the change to the naming!

skahmann3 (approved):
Thank you, Alekhya! Appreciate you resolving this issue!

sql/moz-fx-data-shared-prod/telemetry_derived/unified_metrics_v1/query.sql
```diff
@@ -329,13 +329,305 @@ desktop AS (
...
+      FROM
+        exp_org_mozilla_ios_klar
+      UNION ALL
+      SELECT
+        *
+      FROM
+        exp_org_mozilla_ios_focus
+    )
+  WHERE
+    submission_date = @submission_date
```
lucia-vargas-a, line 573: This filter is required? The subqueries already filter the data for that submission_date.

alekhyamoz:
@skahmann3 tagging Sydney to provide inputs on the comments by @lucia-vargas-a. Do you have any suggestions?

skahmann3:
An activity statement would be convenient @lucia-vargas-a, since we need to filter to experiments x country x version number for AdM eligibility. What's the best way to filter to experiments info to create the activity segment if the experiments info is to be removed from the table?

lucia-vargas-a:
@alekhyamoz to your request, this is what it would look like:

A new column (int or string) in unified_metrics i.e. `subset` that would tell to which subset the client belongs to.
What formulas would you need to calculate it?

@skahmann3 do you think that would be enough for the CDOU calculation ?

lucia-vargas-a:
The other option is to identify the data required for CDOU -and potentially QCDOU- and create a dedicated derived dataset, since this is a key metric it should justify a dedicated table. What do you think?

alekhyamoz:
I would prefer precalculating and just adding the `subset` column to UM. I will leave it to @skahmann3  to take a call.

skahmann3:
I agree that adding a column to unified_metrics is easiest. This is also more sustainable, since we could do it for Tiles and Suggest and unknown future products. We have a couple new markets which were just released, so I'll check in with Product, update the query that we currently use to subset to the correct clients, and then let you know @alekhyamoz so we can create the column. This isn't super urgent, so no pressure to get this done before the holidays

PR #3374, opened 2022-11-22
Make ./bqetl query schema deploy work for Python scripts without --force

### LINKED ISSUE #3368: Make `./bqetl query schema deploy` work for Python scripts without --force
Deploying schemas for Python scripts only works with the `--force` option: 

`./bqetl query schema deploy --force ...`

Without `--force` the validation that is triggered before the deploy will fail with `query.py is not a valid SQL file.` 
We should ignore the validation by default when deploying schemas for Python files since this is otherwise confusing to users.

alekhyamoz:
Fixes #3368

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -1526,6 +1526,10 @@ def deploy(
             click.echo(f"{query_file} dry runs are skipped. Cannot validate schemas.")
             continue
 
+        if str(query_file).endswith('.py'):
+            click.echo(f" Cannot validate schemas for python file {query_file}")
+            continue
+
         query_file_path = Path(query_file)
         existing_schema_path = query_file_path.parent / SCHEMA_FILE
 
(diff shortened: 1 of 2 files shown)


bigquery_etl/cli/query.py
```diff
@@ -1526,6 +1526,10 @@ def deploy(
             click.echo(f"{query_file} dry runs are skipped. Cannot validate schemas.")
             continue
 
+        if str(query_file).endswith(".py"):
+            click.echo(f" Cannot validate schemas for python file {query_file}")
+            continue
+
```
scholtzan, line 1532: We'd still like to be able to deploy `schema.yaml` files for python scripts, but skip the validation part for them. I think this would skip the schema deploy for python files in general.
We'll probably need to add a check around the `if not force:` line to see if a python script has a `schema.yaml` file that could be deployed

scholtzan (approved):

PR #3377, opened 2022-11-24
bug(DENG-462): letting through some of ios sync events in the fxa_auth_events query

kik-kik:
# bug(DENG-462): letting through some of ios sync events in the fxa_auth_events query

cc: @clouserw


gkabbz (approved):

kik-kik:
To be merged on Monday, once merged we should run a backfill for the last 30 days.

PR #3379, opened 2022-11-26
Create table, view, DAG for Firefox Android Clients.

lucia-vargas-a:
This PR creates the table, view & DAG for Firefox Android Clients first reported attributes.

The initialization script scans the first_session and metrics pings to find the earliest not null attribution details.
The query for incremental updates queries the date being processed and updates late-reported data as well as data for new clients.

[Jira story](https://mozilla-hub.atlassian.net/browse/DENG-178).
[Proposal for this new table](https://docs.google.com/document/d/12bj4DhCybelqHVgOVq8KJlzgtbbUw3f68palNrv-gaM/edit).

**Tests**
- Initialization script, query and health checks ran on dataset `analysis.lvargas_firefox_android_clients_v1`
- Unitest implemented for UDF to find earliest attribution values.

Checklist for reviewer:


sql/moz-fx-data-shared-prod/telemetry/firefox_android_clients/view.sql
```diff
@@ -0,0 +1,34 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.firefox_android_clients`
+AS
+SELECT
+  * EXCEPT (adjust_network, install_source),
+  CASE
+  WHEN
+    adjust_network NOT IN (
+      '',
```
fbertsch, line 9: Can we normalize the empty and null Adjust networks? Not sure if they mean e.g. Adjust unavailable, or unknown attribution, etc.

sql/moz-fx-data-shared-prod/telemetry/firefox_android_clients/view.sql
```diff
@@ -0,0 +1,34 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.firefox_android_clients`
```
fbertsch, line 2: I think these belong in the `fenix` dataset.

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/init.sql
```diff
@@ -0,0 +1,312 @@
+-- Initialization query first observations for Firefox Android Clients.
+CREATE OR REPLACE TABLE
+  `moz-fx-data-shared-prod.telemetry_derived.firefox_android_clients_v1`(
+    client_id STRING NOT NULL
+    OPTIONS
+      (description = "Unique ID for the client installation."),
```
fbertsch, line 6: I know you wanted to keep the descriptions so you have the schema here, but they're also down below in the `schema.yaml`. Does that file not propagate descriptions to BQ?

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/init.sql
```diff
@@ -0,0 +1,312 @@
...
+  first_reported_country,
+  adjust_network,
+  install_source
+OPTIONS
+  (
+    description = "First observations for Firefox Android clients retrieved from the earliest pings: baseline, first_session and metrics. The attributes stored in this table include the first attribution, device, OS version and ISP. This table should be accessed through the user-facing view `telemetry.firefox_android_clients`. Proposal: https://docs.google.com/document/d/12bj4DhCybelqHVgOVq8KJlzgtbbUw3f68palNrv-gaM/. For more details about attribution and campaign structure see https://help.adjust.com/en/article/tracker-urls#campaign-structure-parameters."
+  );
+
+INSERT
+  analysis.lvargas_firefox_android_clients_v1
```
fbertsch, line 83: dev table

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/init.sql
```diff
@@ -0,0 +1,312 @@
...
+WITH first_seen AS (
+  SELECT
+    client_id,
+    first_seen_date
+  FROM
+    `mozdata.org_mozilla_firefox.baseline_clients_first_seen`
+  WHERE
+    submission_date >= '2021-01-01'
+    AND first_seen_date = submission_date
```
fbertsch, line 92: What's going on with this statement? Looking over the defn this is only always true for Fenix clients (not migrated Fennec clients)

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/query.sql
```diff
@@ -0,0 +1,482 @@
+-- Query first observations for Firefox Android Clients.
+-- Proposal:
+-- https://docs.google.com/document/d/12bj4DhCybelqHVgOVq8KJlzgtbbUw3f68palNrv-gaM/edit#
+CREATE TEMP FUNCTION get_first_not_null_value(
+  existing_value STRING,
+  existing_date DATE,
+  first_session_value STRING,
+  first_session_date DATE,
+  metrics_value STRING,
+  metrics_date DATE
```
fbertsch, line 10: Could we instead take this is a priority sorted list? Something like:
```
CREATE TEMP FUNCTION get_first_not_null_value(
    date_values ARRAY<STRUCT<value, date>>
)
```
with an unnest and taking the first non-null value for the lowest date. Then for e.g. `source` we'd be able to avoid the `NULL, NULL` arguments.

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/query.sql
```diff
@@ -0,0 +1,482 @@
...
+    AND metrics_value IS NOT NULL
+    AND metrics_date < existing_date
+  THEN
+    metrics_value
+  ELSE
+    existing_value
+  END
+);
+
+CREATE TEMP FUNCTION get_first_not_null_date(
```
fbertsch, line 36: We can have the above function also return the `DATE` so the single function can serve both purposes

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/query.sql
```diff
@@ -0,0 +1,482 @@
+-- Query first observations for Firefox Android Clients.
+-- Proposal:
+-- https://docs.google.com/document/d/12bj4DhCybelqHVgOVq8KJlzgtbbUw3f68palNrv-gaM/edit#
+CREATE TEMP FUNCTION get_first_not_null_value(
```
fbertsch, line 4: If we move this to mozfun, it would be easy to write some tests against it (as compared to writing tests for the entire sql query which is much more tedious)

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/query.sql
```diff
@@ -0,0 +1,482 @@
...
+      AS install_source__source_ping_date
+    ) AS metadata
+  FROM
+    first_seen
+  LEFT JOIN
+    first_session_ping first_session
+  USING
+    (client_id)
+  LEFT JOIN
+    metrics_ping_unique AS metrics
```
fbertsch, line 464: Can we verify the synchrony of client_ids in metrics -> baseline, and first_session -> baseline? Just to see how many we'll leave out with this left join

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/query.sql
```diff
@@ -0,0 +1,482 @@
...
+SELECT
+  first_seen_with_pings_new_data.*
+FROM
+  first_seen_with_pings_new_data
+LEFT JOIN
+  existing_with_pings_new_data
+USING
+  (client_id)
+WHERE
+  existing_with_pings_new_data.client_id IS NULL
```
fbertsch, line 482: I'm not following what's going on here. Why are we unioning then joining, if the client_id is going to be null after the join (and that is the join key?)

fbertsch (commented):
This is a real tricky bit of code. I'm not finished reviewing but this should give you some idea of where my heads at so far

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/query.sql
```diff
@@ -0,0 +1,482 @@
...
+    ANY_VALUE(normalized_channel) AS channel,
+    ANY_VALUE(client_info.device_manufacturer) AS device_manufacturer,
+    ANY_VALUE(client_info.device_model) AS device_model,
+    ANY_VALUE(client_info.os_version) AS os_version,
+    ANY_VALUE(metrics.string.first_session_campaign) AS adjust_campaign,
+    ANY_VALUE(metrics.string.first_session_network) AS adjust_network,
+    ANY_VALUE(metrics.string.first_session_adgroup) AS adjust_ad_group,
+    ANY_VALUE(metrics.string.first_session_creative) AS adjust_creative
+  FROM
+    `mozdata.org_mozilla_firefox.first_session`
```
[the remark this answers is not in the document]
fbertsch: If we do this, we could all it `fenix.release_clients` and `fenix.release_clients_v1`

sql/moz-fx-data-shared-prod/telemetry/firefox_android_clients/view.sql
lucia-vargas-a: @bani how should we normalize this value?
Specially because it's different scenarios, NULL = not received and empty = received without any value.

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/init.sql
lucia-vargas-a: That's a good point, and I'm thinking about removing the schema to avoid confusion becuase a) the process to create or re-create this table is to run the init script, so the schema.yml would not be used and b) Because of the structure of the query, running `bqetl schema update` doesn't work so this file needs to be maintained manually. Probably safer to remove it. What do you think?

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/init.sql
lucia-vargas-a: thanks!

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/schema.yaml
```diff
@@ -0,0 +1,89 @@
...
+  type: STRING
+- description: Manufacturer of device where the client was init. installed.
+  mode: NULLABLE
+  name: device_manufacturer
+  type: STRING
+- description: Model of device where the client was initially installed.
+  mode: NULLABLE
+  name: device_model
+  type: STRING
+- description: OS versino where the client was initially installed.
```
lucia-vargas-a, line 34: typo

sql/moz-fx-data-shared-prod/telemetry/firefox_android_clients/view.sql
lucia-vargas-a: Makes sense, moved it.
How about renaming this table to `fenix_clients`, would it bring more clarity?

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/init.sql
bani: Note that we have duplicate client_ids on this table, and in those cases the same client ids have multiple first_seen_date and submission_date, but they are equal in a given row.

We should fix upstream or accept the duplicates, as they have been generated in error.

sql/moz-fx-data-shared-prod/telemetry/firefox_android_clients/view.sql
bani: In the context that this table will be used for data incidents, not marketing attribution, It's ok to group empty and null into a single value. I think `Unknown` is a good value for the purpose of Looker visualizations.

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/init.sql
lucia-vargas-a: Hey Frank, could you point me to the definition you mention?
For what I see in the [metadata of the baseline_clients_first_seen table](https://github.com/mozilla/bigquery-etl/blob/afad4d84378266774dcc0d745fda2feea6ed46d3/sql/moz-fx-data-shared-prod/org_mozilla_firefox_derived/baseline_clients_first_seen_v1/metadata.yaml#L7),  migrated clients are not captured in this table.

sql/moz-fx-data-shared-prod/telemetry/firefox_android_clients/view.sql
lucia-vargas-a: Thank you. So I've normalized in the view and we have the raw data in the table for the scenario where data is used by Marketing.

sql/moz-fx-data-shared-prod/telemetry/firefox_android_clients/view.sql
```diff
@@ -0,0 +1,34 @@
...
+  THEN
+    'Other'
+  ELSE
+    adjust_network
+  END
+  AS adjust_network,
+  CASE
+  WHEN
+    install_source NOT IN ('com.android.vending')
+    AND install_source IS NOT NULL
```
lucia-vargas-a, line 26: Normalized NULL and empty values as `Unknown` to align with the adjust_network logic.

lucia-vargas-a:
Vanessa, @bani could you please run your health checks and approve if it's satisfactory? The test dataset is analysis.lvargas_firefox_android_clients_v1.

sql/moz-fx-data-shared-prod/telemetry/firefox_android_clients/view.sql
fbertsch: @bani can you describe what you mean by "data incidents"?

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/init.sql
fbertsch: It does seem like we only need one or the other. In the bqetl paradigm we would probably opt to drop the `init.sql`, recreate the table from the `schema.yaml`, and backfill the data using `query.sql` one day at a time. That is difficult and slow, however, so if BQ is capable of loading in one go I like using `init.sql` to recreate all at once and dropping `schema.yaml`.

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/init.sql
fbertsch: It shouldn't be the case that we have duplicates any longer, see [this code that Kik added](https://github.com/mozilla/bigquery-etl/blob/generated-sql/sql/moz-fx-data-shared-prod/org_mozilla_firefox_derived/baseline_clients_first_seen_v1/query.sql#L109). @bani let us know if you've seen dupes since then - we will try and figure out why they're still happening.

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/query.sql
lucia-vargas-a: Sound like a good idea, then we wouldn't need the table schema. Let me have a look,

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/query.sql
lucia-vargas-a: This condition avoids duplicated records for existing clients.
E.g. If `client1` is already in the table and there is a new valid attribution on @submission_date, the record is updated in `first_seen_with_pings_new_data` but it will also appear in `first_seen_with_pings_new_data`, so it would appear twice in the result.
This line excludes the existing records from the `first_seen_with_pings_new_data` set.

sql/moz-fx-data-shared-prod/telemetry/firefox_android_clients/view.sql
bani: @fbertsch data incidents is when one or more metrics we care about deviate from the expected values and we start getting questions "why", and then we start the investigation by segmenting the metric in various ways. The idea from this table was a follow up from [this incident](https://docs.google.com/document/d/170I6yaaSws8LJEsMmXqDQ0U4IChVmRmyfgXmIhGUkrs/edit?usp=sharing)

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/query.sql
fbertsch: I see. You're essentially recreating a `FULL OUTER JOIN` that could encompass both of the CTEs (`first_seen_with_pings_new_data` and `existing_pings_with_new_data`). What do you think about switching this to an outer join and handling the three possibilities there?
- Existing data, but no new data
- Existing and new data
- New data, but no existing data

sql/moz-fx-data-shared-prod/telemetry/firefox_android_clients/view.sql
lucia-vargas-a: For reference, I'd add that in addition to analysis of incidents, this table was extended from the original requirement to add more value:

- Enhanced the existing Firefox Android new_profile ping so that it can be used for clients first attributes and not require the metrics ping anymore to retrieve attribution details.
- Use this structure as a baseline to create new_profile pings for apps that don't have one yet, specifically Firefox Desktop.
- Integrate additional "first seen" dimensions to enable further analysis: first seen country, ISP, device manufacturer, device model.

And it is also a reason to keep the raw data in the table, and implement the normalization and calculations in the view.

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
```diff
@@ -0,0 +1,312 @@
...
+  SELECT
+    client_info.client_id AS client_id,
+    MIN(SAFE.PARSE_DATE('%F', SUBSTR(client_info.first_run_date, 1, 10))) AS first_run_date,
+    ANY_VALUE(normalized_country_code) AS first_reported_country,
+    ANY_VALUE(metadata.isp.name) AS first_reported_isp,
+    ANY_VALUE(normalized_channel) AS channel,
+    ANY_VALUE(client_info.device_manufacturer) AS device_manufacturer,
+    ANY_VALUE(client_info.device_model) AS device_model,
+    ANY_VALUE(client_info.os_version) AS os_version,
+    ANY_VALUE(metrics.string.first_session_campaign) AS adjust_campaign,
```
bani, line 105: I have been using `ARRAY_AGG(field)[SAFE_OFFSET(0)]` instead of ANY_VALUE. I think we should keep things as deterministic as possible.

sql/moz-fx-data-shared-prod/telemetry/firefox_android_clients/view.sql
bani: Good point!

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
fbertsch: +1 on determinism

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
```diff
@@ -0,0 +1,312 @@
...
+  WHERE
+    DATE(submission_timestamp) >= '2021-01-01'
+  GROUP BY
+    client_id
+),
+-- Get first adjust data per client_id per date
+metrics_unique_attribution AS (
+  SELECT
+    client_id,
+    DATE(min_submission_timestamp) AS metrics_ping_submission_date,
```
lucia-vargas-a, line 220: Required as a timestamp to be able to compare which ping comes first btw. metrics and first_session.

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
```diff
@@ -0,0 +1,312 @@
...
+    `mozdata.org_mozilla_firefox.baseline_clients_first_seen`
+  WHERE
+    submission_date >= '2021-01-01'
+    AND first_seen_date = submission_date
+),
+-- Find earliest first_session ping per client.
+first_session_ping AS (
+  SELECT
+    client_info.client_id AS client_id,
+    MIN(SAFE.PARSE_DATE('%F', SUBSTR(client_info.first_run_date, 1, 10))) AS first_run_date,
```
lucia-vargas-a, line 98: This field is DATE. Update to timestamp field to allow comparing with metrics ping date for the first value reported.

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
```diff
@@ -0,0 +1,312 @@
...
+  );
+
+INSERT
+  `moz-fx-data-shared-prod.fenix_derived.firefox_android_clients_v1`
+WITH first_seen AS (
+  SELECT
+    client_id,
+    first_seen_date
+  FROM
+    `mozdata.org_mozilla_firefox.baseline_clients_first_seen`
```
lucia-vargas-a, line 89: `mozdata.org_mozilla_firefox.baseline_clients_first_seen` contains data for the Release channel only.
Use view `mozdata.fenix.baseline_clients_first_seen` for all channels.

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
```diff
@@ -0,0 +1,312 @@
...
+    ANY_VALUE(normalized_channel) AS channel,
+    ANY_VALUE(client_info.device_manufacturer) AS device_manufacturer,
+    ANY_VALUE(client_info.device_model) AS device_model,
+    ANY_VALUE(client_info.os_version) AS os_version,
+    ANY_VALUE(metrics.string.first_session_campaign) AS adjust_campaign,
+    ANY_VALUE(metrics.string.first_session_network) AS adjust_network,
+    ANY_VALUE(metrics.string.first_session_adgroup) AS adjust_ad_group,
+    ANY_VALUE(metrics.string.first_session_creative) AS adjust_creative
+  FROM
+    `mozdata.org_mozilla_firefox.first_session`
```
lucia-vargas-a, line 110: Use `mozdata.fenix.first_session`(based on `org_mozilla_firefox.first_session`) for channels Release, Beta, Nightly. Aurora dataset `mozdata.org_mozilla_fennec_aurora.first_session` is empty for 2021-2022.

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/init.sql
lucia-vargas-a: Removed to include migrated clients too.

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/query.sql
lucia-vargas-a: Moved to `mozfun.norm.get_earliest_value`.

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/query.sql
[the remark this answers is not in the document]
lucia-vargas-a: Well, there is data for all channels not only Release, and the channel is one of the required dimensions to enable investigations.
But it's true this source needed to be updated to `fenix.first_session` which is the correct one as it contains the data for all channels.

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
lucia-vargas-a: Agree, it's updated!

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
lucia-vargas-a: Updated to datetime.

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
```diff
@@ -0,0 +1,312 @@
...
+  `moz-fx-data-shared-prod.fenix_derived.firefox_android_clients_v1`
+WITH first_seen AS (
+  SELECT
+    client_id,
+    first_seen_date
+  FROM
+    `mozdata.org_mozilla_firefox.baseline_clients_first_seen`
+  WHERE
+    submission_date >= '2021-01-01'
+    AND first_seen_date = submission_date
```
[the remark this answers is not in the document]
lucia-vargas-a: Thank you, I removed this condition to collect migrated clients.

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/query.sql
lucia-vargas-a: For ref. This logic is moved to mozfun.norm.get_earliest_value.

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
```diff
@@ -0,0 +1,410 @@
...
+SELECT
+  first_seen_with_pings_new_data.*
+FROM
+  first_seen_with_pings_new_data
+LEFT JOIN
+  existing_with_pings_new_data
+USING
+  (client_id)
+WHERE
+  existing_with_pings_new_data.client_id IS NULL
```
fbertsch, line 410: I really think this should be an outer join. That is simpler, easier to understand, and matches our existing patterns (e.g. `clients_last_seen`).
```
WITH current AS (
  SELECT $fields
  FROM
    existing
  LEFT JOIN
    first_session_ping first_session
  USING
    (client_id)
  LEFT JOIN
    metrics_ping AS metrics
  USING
    (client_id)
  LEFT JOIN
    first_seen
)

SELECT
  IF(previous.client_id IS NULL, previous.*, current.*) REPLACE (
    LEAST(previous.first_seen_date, current.first_seen_date) AS first_seen_date,
    COALESCE(previous.adjust_campaign, current.adjust_campaign) AS adjust_campaign,
    COALESCE(previous.adjust_network, current.adjust_network) AS adjust_network,
    ...other fields...
  )
FROM
  current
FULL OUTER JOIN
  previous
  USING (client_id)
  ```

File: sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql [hunk 21]
```diff
@@ -0,0 +1,410 @@
+-- Query first observations for Firefox Android Clients.
+WITH first_seen AS (
+  SELECT
+    client_id,
+    first_seen_date
+  FROM
+    `mozdata.fenix.baseline_clients_first_seen`
+  WHERE
+    submission_date = @submission_date
```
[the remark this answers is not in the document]
lucia-vargas-a: Yes, for that @submission_date. And joining with the existing clients at the end.

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
```diff
@@ -0,0 +1,410 @@
...
+),
+existing AS (
+  SELECT
+    client_id,
```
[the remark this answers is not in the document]
lucia-vargas-a: Yes

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
```diff
@@ -0,0 +1,316 @@
...
+      ORDER BY
+        metrics.datetime.first_session_timestamp ASC
+    )[SAFE_OFFSET(0)] AS adjust_ad_group,
+    ARRAY_AGG(
+      metrics.string.first_session_creative
+      ORDER BY
+        metrics.datetime.first_session_timestamp ASC
+    )[SAFE_OFFSET(0)] AS adjust_creative
+  FROM
+    `mozdata.fenix.first_session`
```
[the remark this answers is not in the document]
bani: yes, release is sufficient.

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
lucia-vargas-a: Great, let's align with the current implementations.

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/query.sql
```diff
@@ -0,0 +1,482 @@
...
+    existing_source
+  END
+);
+
+WITH first_seen AS (
+  SELECT
+    client_id,
+    first_seen_date
+  FROM
+    `mozdata.org_mozilla_firefox.baseline_clients_first_seen`
```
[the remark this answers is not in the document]
lucia-vargas-a: Looks like it's there, runs with [copy_deduplicate](https://workflow.telemetry.mozilla.org/graph?dag_id=copy_deduplicate)

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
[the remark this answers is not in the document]
lucia-vargas-a: Ok, so the changes would be:
- Filter only Release here i.e. exclude data classified in channel 'Other' / 'beta
- Collect only metrics ping data for Release channel
- Update the naming for the table to specify that is only for Release

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
```diff
@@ -0,0 +1,371 @@
...
+    `mozdata.fenix.baseline_clients_first_seen`
+  WHERE
+    submission_date = @submission_date
+),
+-- Find earliest first_session ping per client.
+first_session_ping AS (
+  SELECT
+    client_info.client_id AS client_id,
+    MIN(SAFE.PARSE_DATETIME('%F', SUBSTR(client_info.first_run_date, 1, 10))) AS first_run_datetime,
+    ARRAY_AGG(normalized_country_code ORDER BY metrics.datetime.first_session_timestamp ASC)[
```
lucia-vargas-a, line 16: If the `first_session` ping is not reported, the dimensions (first_reported_country, first_reported_isp, channel, device_manufacturer, device_model, os_version) will be NULL.
To solve it, collect these dimensions from `mozdata.fenix.baseline_clients_first_seen` .

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
```diff
@@ -0,0 +1,244 @@
...
+          (description = "Datetime of the ping that reported the first install_source value.")
+      >
+  )
+PARTITION BY
+  first_seen_date
+CLUSTER BY
+  channel,
+  first_reported_country,
+  first_reported_isp,
+  device_model
```
fbertsch, line 76: I would add `sample_id` as the first or second clustering field

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
```diff
@@ -0,0 +1,299 @@
...
+),
+-- Find earliest data per client from the first_session ping.
+first_session_ping AS (
+  SELECT
+    client_info.client_id AS client_id,
+    MIN(SAFE.PARSE_DATETIME('%F', SUBSTR(client_info.first_run_date, 1, 10))) AS first_run_datetime,
+    ARRAY_AGG(
+      metrics.string.first_session_campaign
+      ORDER BY
+        metrics.datetime.first_session_timestamp ASC
```
fbertsch, line 27: This will be the same for every install, so if we see two of these from the same client, the order will be nondeterministic. Let's stick with `submission_timestamp` for ordering.

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
```diff
@@ -0,0 +1,299 @@
...
+    )[SAFE_OFFSET(0)] AS adjust_creative
+  FROM
+    `mozdata.fenix.first_session`
+  WHERE
+    DATE(submission_timestamp) = @submission_date
+    AND SAFE.PARSE_DATE('%F', SUBSTR(client_info.first_run_date, 1, 10)) >= DATE(
+      submission_timestamp
+    ) -- first_session ping received after the client is first seen.
+    AND ping_info.seq = 0 -- Pings are sent in sequence, this guarantees that the first one is returned.
```
fbertsch, line 51: If we're doing the array_agg ordering above, do we need this?

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
```diff
@@ -0,0 +1,299 @@
...
+  FROM
+    first_seen
+  LEFT JOIN
+    first_session_ping first_session
+  USING
+    (client_id)
+  LEFT JOIN
+    metrics_ping AS metrics
+  USING
+    (client_id)
```
fbertsch, line 165: This will eliminate any clients who are not new on this day. That means if we e.g. don't get a `first_session` ping on the same day as their first `baseline` ping, they won't be present in this table.

This should be an `OUTER JOIN` to capture all possible new data

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
```diff
@@ -0,0 +1,299 @@
...
+)
+SELECT
+  IF(_current.client_id IS NOT NULL, _current, _previous).* REPLACE (
+    COALESCE(_previous.device_manufacturer, _current.device_manufacturer) AS device_manufacturer,
+    COALESCE(_previous.device_model, _current.device_model) AS device_model,
+    COALESCE(_previous.os_version, _current.os_version) AS os_version,
+    COALESCE(_previous.adjust_campaign, _current.adjust_campaign) AS adjust_campaign,
+    COALESCE(_previous.adjust_ad_group, _current.adjust_ad_group) AS adjust_ad_group,
+    COALESCE(_previous.adjust_creative, _current.adjust_creative) AS adjust_creative,
+    mozfun.norm.get_earliest_value(
```
fbertsch, line 182: there's a lot of repeated logic - can we either have it in `_current`, or here?

sql/moz-fx-data-shared-prod/telemetry_derived/firefox_android_clients_v1/query.sql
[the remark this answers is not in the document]
fbertsch: Yes, but [with a different task_id](https://github.com/mozilla/telemetry-airflow/blob/main/dags/copy_deduplicate.py#L305-L326). The view for `baseline_clients_first_seen` is actually over the `baseline_clients_daily` table, so that will have to be the dependency.

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
lucia-vargas-a: Good point, thanks.

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
lucia-vargas-a: The submission_timestamp is when the ping reaches the server and the seq is the order in which pings are generated. Not necessarily the pings are received in the server in the same order they were generated, so this guarantees that we use the first one, is important to keep it. However the OFFSET adds a level of safety in case more than one ping is generated with the same seq (shouldn't happen tho)

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
lucia-vargas-a: True, this query gets only new clients data and the  late-arriving attribution is captured in the final query so we wouldn't miss it. But we do need to replicated logic for that. Let me check adding the logic here.

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
```diff
@@ -0,0 +1,241 @@
...
+          adjust_network__source_ping_datetime DATETIME
+        OPTIONS
+          (description = "Datetime of the ping that reported the first adjust_network value."),
+          install_source__source_ping_datetime DATETIME
+        OPTIONS
+          (description = "Datetime of the ping that reported the first install_source value.")
+      >
+  )
+PARTITION BY
+  first_seen_date
```
fbertsch, line 74:
```suggestion
```

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
```diff
@@ -0,0 +1,241 @@
...
+    isp AS first_reported_isp,
+    DATETIME(first_run_date) AS first_run_datetime,
+    normalized_channel AS channel,
+    device_manufacturer,
+    device_model,
+    normalized_os_version AS os_version
+  FROM
+    `mozdata.fenix.baseline_clients_first_seen`
+  WHERE
+    submission_date >= '2021-01-01'
```
fbertsch, line 102: We should include more history than this. At least until beginning of 2019.
```suggestion
    submission_date >= '2019-01-01'
```

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
```diff
@@ -0,0 +1,241 @@
...
+    ARRAY_AGG(metrics.string.metrics_adjust_network ORDER BY submission_timestamp ASC)[
+      SAFE_OFFSET(0)
+    ] AS adjust_network,
+    ARRAY_AGG(metrics.string.metrics_install_source ORDER BY submission_timestamp ASC)[
+      SAFE_OFFSET(0)
+    ] AS install_source
+  FROM
+    org_mozilla_firefox.metrics AS org_mozilla_firefox_metrics
+  WHERE
+    DATE(submission_timestamp) >= '2021-01-01'
```
fbertsch, line 148:
```suggestion
    DATE(submission_timestamp) >= '2019-01-01'
```

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
```diff
@@ -0,0 +1,222 @@
...
+    COALESCE(_previous.adjust_campaign, _current.adjust_campaign) AS adjust_campaign,
+    COALESCE(_previous.adjust_ad_group, _current.adjust_ad_group) AS adjust_ad_group,
+    COALESCE(_previous.adjust_creative, _current.adjust_creative) AS adjust_creative,
+    COALESCE(_previous.adjust_network, _current.adjust_network) AS adjust_network,
+    COALESCE(_previous.install_source, _current.install_source) AS install_source,
+    STRUCT(
+      COALESCE(
+        _previous.metadata.reported_first_session_ping,
+        _current.metadata.reported_first_session_ping
+      ) AS reported_first_session_ping,
```
fbertsch, line 182: These should be ORs, since they could have reported these pings previously or currently:
```suggestion
      _previous.metadata.reported_first_session_ping
        OR _current.metadata.reported_first_session_ping
        AS reported_first_session_ping,
```

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
```diff
@@ -0,0 +1,241 @@
...
+FROM
+  first_seen
+LEFT JOIN
+  first_session_ping AS first_session
+USING
+  (client_id)
+LEFT JOIN
+  metrics_ping AS metrics
+USING
+  (client_id)
```
fbertsch, line 241: Should we outer join these as well?

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
```diff
@@ -0,0 +1,222 @@
...
+),
+--existing clients in firefox_android_clients_v1
+_previous AS (
+  SELECT
+    *
+  FROM
+    `moz-fx-data-shared-prod.fenix_derived.firefox_android_clients_v1`
+)
+SELECT
+  IF(_current.client_id IS NOT NULL, _current, _previous).* REPLACE (
```
fbertsch, line 169: Should we default to `_previous`? Not sure offhand, just want to make sure this is right

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
lucia-vargas-a: The init runs once and updates the whole table so it captures all data. Do you see it necessary?

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
lucia-vargas-a: So if either previous or current reported (not null) it would be captured, it'd be equivalent right?

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
lucia-vargas-a: I see it correct.

Do you mean
    `_previous.client_id IS NULL, _current, _previous` or
    `_previous.client_id IS NOT NULL, _previous, _current` ?

fbertsch (approved):
:shipit:

PR #3382, opened 2022-11-30
Proof of concept: Move to poetry and include mac m1 native zetasql jars

quiiver:
The story here is that I have been enjoying using poetry over pip when working in the merino-py environment. It consolidates virtual environments (`venv`), packaging (`setuptools`), dependency management (`pip`) and locking (`pip-tools`) into a single `npm` like interface. It also leverages the newer `pyproject.toml` format that aims to consolidate the various tool configurations into a single file (flake8 being the exception). 

This particular exploration started after being frustrated trying to get pip to build dependencies in the m1 native arm64 arch. Out of curiosity, I copied the dependencies from `requirements.in` to `pyproject.toml` and `poetry` was able to build all the dependencies with minimal work. Encouraged by this result I started looking into what it would take to get a fully native m1 build working and identified that the java lib `zetasql` was not yet publishing a version that supports arm64. This led to a serious detour that resulted in forking [zetasql](https://github.com/quiiver/zetasql), reading through various issues and learning enough `bazel` to get it building jars that work on m1 macs. 

Encouraged again by the results of this experiment I wanted to throw this PR together to get feedback on whether or not this could be a viable path forward for native m1 support.

The following changes were done in service of this proof of concept:
1. Migrated requirements.in -> pyproject.toml
2. Migrated setup.py -> pyproject.toml
3. Migrated pytest.ini -> pyproject.toml
4. Migrated isort.cfg -> pyproject.toml
5. Updated Dockerfile to support poetry
6. Updated circleci/config.yaml to support poetry
7. Removed pytest lint plugins in favor of `script/lint`
8. Added `lib/m2` as a local maven repo
9. Added zetasql jars to the repository in `lib/m2`
10. Updated pom.xml to use the local repository and `LOCAL-SNAPSHOT` version of zetasql when on an M1 mac


lib/m2/com/google/zetasql/zetasql-client/LOCAL-SNAPSHOT/maven-metadata-local.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
quiiver: I created this manually to store the custom zetasql jars. Is it confusing because it looks like the local cache repo at $HOME/.m2 ?

pyproject.toml
```diff
@@ -0,0 +1,100 @@
...
+  { include = "sql_generators" },
+]
+include = ["target/*.jar"]
+
+[tool.poetry.scripts]
+bqetl = 'bigquery_etl.cli:cli'
+
+
+[tool.poetry.dependencies]
+python = "^3.10,<3.11"
```
relud, line 18: unless poetry doesn't support it, this should probably use `~=3.10`, which means the same thing

script/lint
```diff
@@ -0,0 +1,12 @@
...
+set -x
+
+APP_DIRS="bigquery_etl sql_generators"
+TEST_DIR="tests"
+APP_AND_TEST_DIRS="$APP_DIRS $TEST_DIR"
+
+black --quiet --diff --check $APP_AND_TEST_DIRS
+flake8 $APP_AND_TEST_DIRS
+pydocstyle $APP_AND_TEST_DIRS
+mypy $APP_DIRS  --config-file="pyproject.toml"
```
relud, line 12: this is missing isort

script/lint
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
relud, line 1: why this instead of pytest plugins?

pom.xml
```diff
@@ -17,6 +17,27 @@
...
+          </properties>
+
+          <repositories>
+            <repository>
+              <id>local-maven-repo</id>
+              <url>file://${basedir}/lib/m2/</url>
+            </repository>
+          </repositories>
+
+        </profile>
```
relud, line 40: `include mac m1 native zetasql jars` should be broken out into a separate PR

pom.xml
```diff
@@ -17,6 +17,27 @@
...
+          </activation>
+
+          <properties>
+            <zetasql.version>LOCAL-SNAPSHOT</zetasql.version>
+          </properties>
+
+          <repositories>
+            <repository>
+              <id>local-maven-repo</id>
+              <url>file://${basedir}/lib/m2/</url>
```
relud, line 36: this is traditionally in `~/.m2`, why is it in the git repo?

.circleci/config.yml
```diff
@@ -573,7 +559,7 @@ workflows:
       - verify-format-sql:
           requires:
             - generate-sql
-      - verify-requirements
+      # - verify-requirements
```
relud, line 562:
```suggestion
```

bqetl
```diff
@@ -17,27 +17,13 @@ if ! command -v python3 &> /dev/null; then
...
 fi
 
-if [ ! -d "venv" ]; then
-    echo "Please run ./bqetl bootstrap"
-    exit 1
-fi
-
-venv/bin/bqetl "$@"
+export CLOUDSDK_PYTHON=$(which python)
+poetry run bqetl "$@"
```
relud, line 29: script/bqetl needs to be updated accordingly

Dockerfile
```diff
@@ -12,7 +12,16 @@ WORKDIR /app
...
-COPY requirements.txt ./
+
+# install poetry
+RUN pip install --no-cache-dir --quiet poetry
+
+# copy poetry files over
+COPY ./pyproject.toml ./poetry.lock ./
+
+# Just need the requirements.txt from Poetry
+RUN poetry export --no-interaction --output requirements.txt --without-hashes
```
relud, line 23: why without hashes?

Dockerfile
```diff
@@ -12,7 +12,16 @@ WORKDIR /app
 # build typed-ast in separate stage because it requires gcc and libc-dev
 FROM base AS python-deps
 RUN apt-get update -qqy && apt-get install -qqy gcc libc-dev
-COPY requirements.txt ./
+
+# install poetry
+RUN pip install --no-cache-dir --quiet poetry
```
mikaeld, line 17: [preferred way of installing](https://python-poetry.org/docs/#installing-with-the-official-installer) `poetry` is 
```shell
curl -sSL https://install.python-poetry.org | python - --version <<poetry_version>>
```

relud (commented):
This requires buy-in from a broader audience. I'd like to see it discussed at the data platform working group meeting, and if we do decide to use poetry we should also migrate data eng's other pip-tools repos.

script/lint
quiiver: I believe we only use isort for pre-commit, i believe that's why i didn't include it here

script/lint
quiiver: I was having issues with using them but I think that turned out to be a red herring. Also I thought it would be nice to have a script to run specifically for linting without having to include the full test suite

quiiver: I just checked and re-enabling the pytest plugins works fine locally, so if folks prefer we could go back to that

lib/m2/com/google/zetasql/zetasql-client/LOCAL-SNAPSHOT/maven-metadata-local.xml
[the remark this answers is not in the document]
relud: ah, i see. this is a local repo (not cache) storing your build of the m1 jars. I still don't think this belongs in git.

script/lint
relud: i do prefer

script/lint
relud: definitely not just for pre-commit, isort should be included with all our other linters

quiiver: I'm just saying that previously we only used it for pre-commit in this repository and I wanted to limit additional changes with this. can certainly be added but will require changing source files as a few don't pass

lib/m2/com/google/zetasql/zetasql-client/LOCAL-SNAPSHOT/maven-metadata-local.xml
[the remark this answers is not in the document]
relud: to be clear, the main reason i don't think this belongs in git is that it includes a binary jar. instead i think this should be hosted in a maven repo outside of git, and pulled in as an artifact from there.

Dockerfile
```diff
@@ -12,7 +12,16 @@ WORKDIR /app
...
+
+# copy poetry files over
+COPY ./pyproject.toml ./poetry.lock ./
+
+# Just need the requirements.txt from Poetry
+RUN poetry export --no-interaction --output requirements.txt --without-hashes
+
 # use --no-deps to work around https://github.com/pypa/pip/issues/9644
 RUN pip install --no-deps -r requirements.txt
```
mikaeld, line 26: In my experience, `pip install` with requirements output from poetry is very shaky. Here's how you can do this without pip.

```Dockerfile
ARG PYTHON_VERSION=3.10

FROM python:${PYTHON_VERSION}-slim-buster as base

ENV PYTHONUNBUFFERED=1 \
    PYTHONFAULTHANDLER=1 \
    PYSETUP_PATH="/opt/pysetup" \
    VENV_PATH="/opt/pysetup/.venv" \
    POETRY_HOME="/opt/poetry" \
    POETRY_VIRTUALENVS_IN_PROJECT=true \
    POETRY_NO_INTERACTION=1

ENV PATH="$POETRY_HOME/bin:$VENV_PATH/bin:$PATH"

FROM base as builder

ARG POETRY_VERSION=1.2.1

WORKDIR $PYSETUP_PATH
COPY . .

RUN apt-get update &&\
    apt-get install -y --no-install-recommends curl
RUN curl -sSL https://install.python-poetry.org | python - --version $POETRY_VERSION &&\
    poetry install -vvv --no-dev

FROM base as final
ARG USER=example
ARG HOME_DIR=/home/$USER
ARG PROJECT_DIR=$HOME_DIR/src

# create user with home directory and bash as the default shell
RUN groupadd --gid 5000 $USER
RUN useradd --home-dir $HOME_DIR --create-home --uid 5000 \
      --gid 5000 --shell /bin/bash --skel /dev/null $USER

WORKDIR $PROJECT_DIR

# copy python dependencies from builder layer
COPY --from=builder --chown=5000:5000 $VENV_PATH $VENV_PATH

RUN chown 5000:5000 $PROJECT_DIR

# use container as non-root
USER 5000
```

script/lint
relud: ah, we had pytest-isort installed, but it was only being enforced via pre-commit, because it was missing from script/entrypoint. that's a separate bug, yeah.

relud: 302 -> https://github.com/mozilla/bigquery-etl/pull/3384

lib/m2/com/google/zetasql/zetasql-client/LOCAL-SNAPSHOT/maven-metadata-local.xml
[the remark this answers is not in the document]
quiiver: Moved this out to #3400

pyproject.toml
quiiver: I don't think poetry supports that, at least it thinks the python constraint is `>=3.10,<4.0` when i set it to `~=3.10`

Dockerfile
quiiver: This will be resolved when we update the Dockerfile

mikaeld:
@relud how can we test if the Dockerfile still works as intended?

relud:
use it to run a query and publish a view

bqetl
mikaeld: prefixing everything with `poetry run` will be an issue inside the docker container since poetry should not be in the final layer.

quiiver: The reason I only prefix this command with `poetry run` is because (I believe) this is supposed to be the main developer interface. For scripts in the `scripts/*` directory I explicitly did not add the `poetry run` prefix because the context could be within a container.

`poetry run` just runs the command in a subshell with the virtual environment activated. if a dev wanted to run a script in `scripts/` they could do so by either prefixing with `poetry run` or activating the virtual env (`poetry shell` will do that

PR #3384, opened 2022-11-30
Enforce isort via pytest

relud:
we are already installing `pytest-isort`, but we weren't actually running it


fbertsch (approved):
Definitely thought we already did this. Looks like it was just precommit

quiiver:
Just a heads up, I don't believe this will cause `isort` to be enforced in CI since we always call `script/entrypoint` with command line flags

PR #3385, opened 2022-11-30
DENG-561 UDF to find earliest of a set of values

lucia-vargas-a:
UDF that receives an array and returns the earliest not null value.
Required for the [implementation of the Firefox Android Clients table](https://github.com/mozilla/bigquery-etl/pull/3379).

Checklist for reviewer:


sql/mozfun/norm/get_earliest_value/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: Get Earliest Value
+description: |
+  This UDF returns the earliest not-null value from a set of 3 values.
```
fbertsch, line 3:
```suggestion
  This UDF returns the earliest not-null value from a list of values and their dates.
```

sql/mozfun/norm/get_earliest_value/metadata.yaml
```diff
@@ -0,0 +1,10 @@
...
+
+  The order is not important as the function will return the first value
```
fbertsch, line 5: This isn't the case for us, right? We want to take the first value for a given date. For example, we want to first get the value from the `first_session` ping on a day, and if unavailable, get it from the `metrics` ping on that day.

sql/mozfun/norm/get_earliest_value/udf.sql
```diff
@@ -0,0 +1,65 @@
...
+      ORDER BY
+        offset ASC
+    )
+    SELECT
+      ARRAY_AGG(STRUCT(value, value_date) ORDER BY value_date ASC)[SAFE_OFFSET(0)]
+    FROM
+      unnested
+    WHERE
+      value IS NOT NULL
+      AND value_date IS NOT NULL
```
lucia-vargas-a, line 21: This value is not expected to be null as pings have a timestamp, so it's meant to avoid returning values without dates.

sql/mozfun/norm/get_earliest_value/metadata.yaml
lucia-vargas-a: @bani here is the scenario I explained, what do you think?

bani: Usually I expect first_session to come before metrics, but if it happens that metrics comes first, I'm fine with that value, we can use whatever ping is available earliest that contains non-NULL data.

lucia-vargas-a: After checking with Vanessa, we confirmed about collecting the first reported (not null) value, from any ping,

fbertsch (approved):

PR #3391, opened 2022-12-01
Only merge descriptions from Glean stable table schemas to views

sean-rose:
After #3343 `schema.yaml` files are failing to be generated for the Glean stable views because the step to merge Glean stable table schemas with their view schemas is failing due to the datetime metric column types being different.  That schema merge is only being done to copy the column descriptions to the view, so this tweaks that schema merge logic to only copy the `description` attributes (ignoring other attributes like `type`).

---
Checklist for reviewer:

### DIFF
--- a/bigquery_etl/schema/__init__.py
+++ b/bigquery_etl/schema/__init__.py
@@ -90,6 +90,7 @@ class Schema:
         other: "Schema",
         exclude: Optional[List[str]] = None,
         add_missing_fields=True,
+        attributes: Optional[List[str]] = None,
     ):
         """Merge another schema into the schema."""
         self._traverse(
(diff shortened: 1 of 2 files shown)


bigquery_etl/schema/__init__.py
```diff
@@ -230,7 +236,11 @@ def _traverse(
                         f"{prefix}.{field_path}",
                         nodes[node_name]["fields"],
                         node["fields"],
-                        update,
+                        update=update,
+                        add_missing_fields=add_missing_fields,
+                        ignore_missing_fields=ignore_missing_fields,
+                        exclude=exclude,
```
sean-rose, line 242: I'm unclear whether `exclude` is intended to only be for top-level columns.  If so, we wouldn't want to pass it along to the recursive calls.

scholtzan:
```suggestion
```
Yeah, it was only for top-level columns

scholtzan (approved):

Issue #3397, opened 2022-12-02
Parser to extract Windows OS name, version, and build from an OS version string


jcchuang:
# Need for a new SQL user-defined function (UDF)
Our ETL tools currently don't have an efficient way to extract the OS name, version name, version number, and build number from a Microsoft Windows OS version string.

The closest existing UDF `mozfun.norm.windows_version_info` can produce a _standardized_ OS name when given an OS name, OS version number, and OS build number. However, all these values can actually be extracted directly from an OS version string without additional knowledge. This UDF requires additional input about OS names and build numbers in order to run. This UDF only normalizes the OS name and does not extract the other fields.

There are no efficient built-in SQL operations, either. The Windows OS version string has evolved over the years -- from a 3-number string (`x.y.z`) to a 4-number string (`w.x.y.z`), from embedding the version number in the first two digits (`x.y` of `x.y.z`) to embedding the version number in the third digit (`y` of `w.x.y.z`), among other idiosyncratic characteristics. Therefore, parsing cannot be done via simple array manipulations, string operations, or type conversions.

# Proposed UDF
The first version of the proposed UDF is designed to extract the OS name, version name, version number, and build number from the Microsoft Windows OS version string collected in field `os_version` of table `mozdata.default_browser_agent.default_browser`.

The proposed UDF contains sufficient contextual information about Microsoft Windows systems running in the past 26 months (from Windows Vista to Windows 11), and is expected to parse approximately 99.89% of the OS version string values observed in the `os_version` field as of November 2022.

wwyc:
@jcchuang it looks like the new udf has been merged.  Should this issue be closed?

jcchuang:
Resolved in https://github.com/mozilla/bigquery-etl/pull/3398

PR #3398, opened 2022-12-02
New UDF: mozfun.norm.get_windows_info

jcchuang:
Background (on the need for a new UDF and the proposed UDF) is described in issue #3397.

Checklist for reviewer:


sql/mozfun/norm/get_windows_info/README.md
```diff
@@ -0,0 +1,49 @@
+# Windows Names, Versions, and Builds
+
+### Summary
+This function is primarily designed to parse the field `os_version` in table `mozdata.default_browser_agent.default_browser`. Though untested, the function should also parse any generic Microsoft Windows OS version strings.
```
perrymcmanis144, line 4: might suggest changing any to most, and dating when this was written, e.g.

"most generic Microsoft Windows OS version strings as of December 2022"

I realize this is mentioned later, but being explicit here might help

sql/mozfun/norm/get_windows_info/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+As of November 2022, the expected valid values of `os_version` are either `x.y.z` or `w.x.y.z` where `w`, `x`, `y`, and `z` are integers.
+
+As of November 2022, the return values for Windows 10 and Windows 11 are based on [Windows 10 release information](https://learn.microsoft.com/en-us/windows/release-health/release-information) and [Windows 11 release information](https://learn.microsoft.com/en-us/windows/release-health/windows11-release-information). For 3-number version strings, the parser assumes the valid values of `z` in `x.y.z` are at most 5 digits in length. For 4-number version strings, the parser assumes the valid values of `z` in `w.x.y.z` are at most 6 digits in length. The function makes an educated effort to handle Windows Vista, Windows 7, Windows 8, and Windows 8.1 information, but does not guarantee the return values are absolutely accurate. The function assumes the presence of undocumented non-release versions of Windows 10 and Windows 11, and will return an estimated name, version number, build number but not the version name. The function does not handle other versions of Windows.
```
perrymcmanis144, line 11: I think the comment that accuracy is not guaranteed buries the lead slightly, could this be moved to the top? I think it's absolutely fine that this UDF cannot guarantee that it will be correct 100% of the time, but IMO if this is the case, that caveat should probably be the actual first line of the readme

sql/mozfun/norm/get_windows_info/metadata.yaml
```diff
@@ -0,0 +1,6 @@
+---
+friendly_name: Get Microsoft Windows OS information
+description: |
+  Exract the name, the version name, the version number, and the build number
```
perrymcmanis144, line 4: The accuracy comment should be here as well, in the description

perrymcmanis144 (changes requested):
I'm satisfied with the JS code. It's pretty deeply nested, but because of the nature of UDFs and their interaction with regex compilations, this is probably a reasonable way to go. This assumes that it passes all of the assertions, of course.

Of my requested changes to the readme/description, it is emphasizing that it may not be 100% consistent that is the only blocker for me here. To reiterate, I think it's fine that you may get an unknown or there may be some sharp edges in the UDF itself, but anyone perusing mozfun should see that first when they consider invoking this for their own edification.

PR #3406, opened 2022-12-06
Replace ZetaSQL with sqlglot, a pure Python library

badboy:
**This is a WIP and nowhere near ready. I just figured I should publish my initial work here already.**

We use ZetaSQL for exactly one thing: parsing SQL and extracting the referenced table names.
ZetaSQL is a Java project and currently not available for aarch64 (such as Apple M1 macbooks), making it harder than necessary to develop bigquery-etl on those machines than we would like.

An non-Java alternative could be a solution.
SQLGlot is a pure-Python SQL Parser and Transpiler.

Unfortunately it's not yet able to parse all our BigQuery SQL.
But they seem to be eager to fix bugs, the [one I reported yesterday was addressed within hours](https://github.com/tobymao/sqlglot/issues/797). Maybe we'll get it working good enough for our use case in the not-so-distant future.


bigquery_etl/dependency.py
```diff
@@ -13,44 +14,17 @@
...
-        # fails, because for scripts zetasql incorrectly includes CTE references from
-        # subquery expressions
-        try:
-            result = ZetaSqlHelper.extractTableNamesFromScript(sql)
-        except jnius.JavaException as e:
-            # replace jnius.JavaException because it's not available outside this function
-            raise ValueError(*e.args)
-    return [".".join(table.toArray()) for table in result.toArray()]
+
+    parsed = sqlglot.parse_one(sql, read="bigquery")
```
tobymao, line 21: it might be more robust to do 

```python

from sqlglot import exp
from sqlglot.optimizer.scope import traverse_scope

def find_tables(query: exp.Expression) -> t.Set[str]:
    """Find all tables referenced in a query.

    Args:
        query: The expression to find tables for.

    Returns:
        A Set of all the table names.
    """
    return {
        exp.table_name(table)
        for scope in traverse_scope(query)
        for table in scope.tables
        if isinstance(table.this, exp.Identifier)
        and exp.table_name(table) not in scope.cte_sources
    }
```

relud:
fixed in #3755

PR #3410, opened 2022-12-06
Add table for Fenix attributable clients

fbertsch:
We will use this table to monitor the new Fenix campaigns rolling out. This should give us daily data (I don't think we need any more timely than that, but if we do I will enable it in a view).

Checklist for reviewer:


sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v1/init.sql
```diff
@@ -0,0 +1,115 @@
...
+    WITH client_day AS (
+      SELECT
+        DATE(submission_timestamp) AS submission_date,
+        sample_id,
+        client_info.client_id,
+        LOGICAL_OR(
+          mozfun.norm.extract_version(client_info.app_display_version, 'major') >= 107
+        ) AS has_search_data,
+        SUM(sum_map_values(metrics.labeled_counter.browser_search_with_ads)) AS searches_with_ads,
+        SUM(sum_map_values(metrics.labeled_counter.browser_search_ad_clicks)) AS ad_clicks,
```
fbertsch, line 23: This bit is to pull in the data from baseline ping for search only when the client is above 107 on that day.

sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v1/init.sql
```diff
@@ -0,0 +1,115 @@
...
+          SAFE_OFFSET(0)
+        ] AS adjust_campaign,
+        ARRAY_AGG(metrics.string.first_session_creative ORDER BY submission_timestamp)[
+          SAFE_OFFSET(0)
+        ] AS adjust_creative,
+        MIN(DATE(submission_timestamp)) AS first_session_date
+      FROM
+        `mozdata.fenix.first_session`
+      WHERE
+        DATE(submission_timestamp) >= '2022-08-01'
```
fbertsch, line 78: Limited date range. If we feel we need to backfill further, we can.

bani: are we rebuilding the entire table every day? Shouldn't we make it incremental?

bani:
This table seems to fill the same role as the view created at https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/fenix/marketing_attributable_metrics/view.sql

We should either deprecate that view or make some changes on this table to enable the view to be constructed over this table.

sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v1/init.sql
bani: nevermind, I see this is the init and not the query

sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v1/init.sql
```diff
@@ -0,0 +1,115 @@
...
+  (SELECT SUM(value) FROM UNNEST(map))
+);
+
+CREATE OR REPLACE TABLE
+  fenix_derived.attributable_clients_v1
+PARTITION BY
+  submission_date
+CLUSTER BY
+  adjust_network,
+  adjust_campaign,
```
bani, line 12: we have almost 10k campaign values with installs since August, so that may not be the best clustering field. I'd leave it out and maybe add country.

sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v1/init.sql
```diff
@@ -0,0 +1,115 @@
...
+        metrics_searches.searches_with_ads
+      ) AS searches_with_ads,
+      IF(client_day.has_search_data, client_day.ad_clicks, metrics_searches.ad_clicks) AS ad_clicks
+    FROM
+      adjust_client
+    JOIN
+      client_day
+    USING
+      (client_id)
+    LEFT JOIN
```
bani, line 107: previously we had to switch this to outer join because our search numbers were incomplete otherwise.

sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v1/query.sql
```diff
@@ -0,0 +1,106 @@
...
+
+WITH client_day AS (
+  SELECT
+    DATE(submission_timestamp) AS submission_date,
+    sample_id,
+    client_info.client_id,
+    LOGICAL_OR(
+      mozfun.norm.extract_version(client_info.app_display_version, 'major') >= 107
+    ) AS has_search_data,
+    SUM(sum_map_values(metrics.labeled_counter.browser_search_with_ads)) AS searches_with_ads,
```
skahmann3, line 14: Add search counts here. Search with ads can be unreliable so better to have both if necessary

sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v1/query.sql
```diff
@@ -0,0 +1,106 @@
...
+  country,
+  adjust_network,
+  adjust_adgroup,
+  adjust_campaign,
+  adjust_creative,
+  first_seen_date = submission_date AS is_new_profile,
+  IF(
+    client_day.has_search_data,
+    client_day.searches_with_ads,
+    metrics_searches.searches_with_ads
```
skahmann3, line 90: Why are we keeping both? Are we planning on backfilling past the baseline search implementation? If so, we might need an outer join below as we'll have mismatched dates between baseline and metrics

sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v1/query.sql
```diff
@@ -0,0 +1,106 @@
...
+  GROUP BY
+    submission_date,
+    sample_id,
+    client_id
+),
+searches AS (
+  SELECT
+    submission_date,
+    client_id,
+    SUM(search_with_ads) AS searches_with_ads,
```
skahmann3, line 29: Add search counts here. Search with ads can be unreliable so better to have both

sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v1/init.sql
skahmann3: +1 and IIRC this had a big impact

sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v1/init.sql
fbertsch: Even with a long tail it can still be useful. I'll be using this one for filtering so I'd rather keep it.

bani: how about adding country? We always use that one.

fbertsch: Sure, we can add it in

fbertsch:
@bani @skahmann3 Dealing with search_counts from two pings is tricky, but I believe this way will prevent us from double counting (only taking it from `baseline` if the version is >= 107, only taking it from metrics if the version of that client on that day is < 107).
- Clustering fields are updated
- Views are updated (including redirecting `marketing_attributable_clients`, this will be much faster to query)
- This table is available in shared-prod now for testing

I'm going to run some tests in Looker on the two tables.

bani (approved):

sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v1/init.sql
bani: I'd suggest having country as the first or second field since the order matters (it won't leverage it unless the query is also filtering on campaign if country comes after it), but I'm good with it as it is 👍

fbertsch: Usually, secondary/tertiary defined clustered fields will be used in filters or joins, but they just won't be as effective at filtering out partitions.

I think in general this table is going to be such an improvement on `marketing_attributable_metrics` that it won't matter - however if we end up with a higher need on fast queries for countries over campaigns, we can switch this up.

PR #3415, opened 2022-12-07
Allow deploys for tables without metadata.yaml

scholtzan:
Checklist for reviewer:

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -1218,7 +1218,7 @@ def schema():
     default="analysis",
 )
 @use_cloud_function_option
-@respect_dryrun_skip_option(default=False)
+@respect_dryrun_skip_option(default=True)
 def update(
     name,
     sql_dir,
@@ -1578,8 +1578,6 @@ def deploy(
             click.echo(f"No schema file found for {query_file}")
             continue
 
-        click.echo(f"Deploying schema for {query_file}")
-
         table_name = query_file_path.parent.name
         dataset_name = query_file_path.parent.parent.name
         project_name = query_file_path.parent.parent.parent.name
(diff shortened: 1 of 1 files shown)


bigquery_etl/cli/query.py
```diff
@@ -1578,8 +1578,6 @@ def deploy(
             click.echo(f"No schema file found for {query_file}")
             continue
 
-        click.echo(f"Deploying schema for {query_file}")
```
scholtzan, line 1581: Removing this since the message might be confusing since the schema might not actually get deployed (e.g. when using `--skip-existing`)

bigquery_etl/cli/query.py
```diff
@@ -1636,7 +1634,10 @@ def deploy(
 
 def _attach_metadata(query_file_path: Path, table: bigquery.Table) -> None:
     """Add metadata from query file's metadata.yaml to table object."""
-    metadata = Metadata.of_query_file(query_file_path)
+    try:
+        metadata = Metadata.of_query_file(query_file_path)
+    except FileNotFoundError:
+        return
```
scholtzan, line 1640: some older queries might not have `metadata.yaml` files

relud (approved):

bigquery_etl/cli/query.py
kik-kik: Should we add them instead of making `metadata.yaml` optional for future tables? I'm just thinking we should encourage people to add descriptions and context around the tables they create and what purpose it serves.

kik-kik: Or maybe there should be a place where explicitly we have to allow no `metadata.yaml` for a specific table? Like we do with skip-dry-run. So it's more "intentional".

scholtzan: We should probably add metadata.yaml files instead of investing time in curating another skip list.

scholtzan: I opened https://github.com/mozilla/bigquery-etl/issues/3416 to follow up on this

kik-kik: thanks!

PR #3423, opened 2022-12-07
Automatically generate tables for Glean apps

scholtzan:
We should hold off from merging this until we have moved the view deployment logic out of Jenkins into Airflow.


relud:
this is causing CI to fail because dry run doesn't have permissions to read regrets_reporter_ucs datasets

relud:
https://github.com/mozilla/bigquery-etl/pull/3434

PR #3424, opened 2022-12-08
Publish views serially and default to only publishing changes

relud:


bigquery_etl/view/__init__.py
```diff
@@ -216,6 +222,58 @@ def _valid_view_naming(self):
...
+        ).strip()
+        actual_view_query = sqlparse.format(
+            table.view_query, strip_comments=True
+        ).strip()
+        if expected_view_query != actual_view_query:
+            print(f"view {target_view_id} will change: query does not match")
+            return True
+
+        # check schema
+        schema_file = Path(self.path).parent / "schema.yaml"
```
scholtzan, line 262: Unless `schema.yaml` is added manually or at some point `bqetl query schema update` is called, most views won't have a schema file.

scholtzan: Simply deploying them might be faster though than dry running the view in order to get the schema information

relud: yeah, but if it's there we apply it during the publish step

scholtzan (approved):

Issue #2445, opened 2021-10-20
Support for creating initial derived tables


whd:
Occasionally a derived table is created in a restricted access dataset that ops needs to manually run `init.sql` for. Airflow/ops views deploy could theoretically be responsible for this table initialization as well.

whd:
fixed demonstrated by linked PR

Issue #2374, opened 2021-09-23
Views deploys failures with views that have dependencies on other views that have not been created


whd:
e.g. https://github.com/mozilla/bigquery-etl/pull/2367

A simple re-run is the easy workaround, but if we had robust dependency metadata we could probably avoid this.

whd:
fixed demonstrated by linked PR

PR #3440, opened 2022-12-08
Add test query to test BigQuery artifact deployment

scholtzan:
This is only for testing the deployment and every dataset and view created will be deleted


relud (approved):

whd (approved):
should demonstrate fixes for https://github.com/mozilla/bigquery-etl/issues/2445 and https://github.com/mozilla/bigquery-etl/issues/2374

PR #3444, opened 2022-12-09
bug: fixing partition field containing value for multiple partitions error

kik-kik:
`timestamp` field can contain a value which corresponds to a day in the past. Fixing this by setting the value of the field by which the table is partitioned to be equal the the partition processed by bqetl.


alekhyamoz (approved):
lgtm

PR #3449, opened 2022-12-12
Don't stop updating and deploying schemas if one has failed

scholtzan:
This PR makes some improvements to unblock table deploys:
If a deploy fails for some table, then this changes ensures that the deploys for the remaining tables will keep running instead of stopping immediately.

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -13,6 +13,7 @@ from graphlib import TopologicalSorter
 from multiprocessing.pool import Pool, ThreadPool
 from pathlib import Path
 from tempfile import NamedTemporaryFile
+from traceback import print_exc
 
 import click
 import yaml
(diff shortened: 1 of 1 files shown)


bigquery_etl/cli/query.py
```diff
@@ -1587,60 +1594,75 @@ def deploy(
...
-                    "time_partitioning",
-                    "clustering_fields",
-                    "labels",
-                ],
-            )
-            click.echo(f"Schema (and metadata) updated for {full_table_id}.")
+    if len(failed_deploys) > 0:
+        click.echo("The following tables could not be deployed:")
+        for failed_deploy in failed_deploys:
+            click.echo(failed_deploy)
```
scholtzan, line 1662: might be handy to have a summary of failed deploys at the end of the log

relud (approved):

PR #3456, opened 2022-12-13
Fix Firefox iOS `unified_metrics` view (bug 1805485)

sean-rose:
As a result of #3453 the `org_mozilla_ios_firefox.unified_metrics` ETL is failing ([bug 1805485](https://bugzilla.mozilla.org/show_bug.cgi?id=1805485)) because the generated legacy telemetry subqueries contain a Glean datetime metric column as a string (based on the stable table schema) while the Glean metrics view now selects the datetime metric column as a timestamp.

This changes the generated legacy telemetry subqueries to apply the `mozfun.glean.parse_datetime()` function to the column (even though the column contains no data), like the Glean metrics view does.

---
Checklist for reviewer:

### DIFF
--- a/sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox/unified_metrics_v1/query.py
+++ b/sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox/unified_metrics_v1/query.py
@@ -3,11 +3,13 @@
 This approach is required because the order of columns is significant in order
 to be able to union two tables, even if the set of columns are the same."""
 
-import requests
-import json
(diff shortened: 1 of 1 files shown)


sean-rose:
An alternative approach would be to have it use the schema of the Glean metrics view it's querying rather than the underlying Glean metrics stable table, but I wasn't quite sure about how best to do that and wanted to get this fixed today.  If a reviewer knows how to do that and thinks it would be a better approach I'm happy to change the approach.

relud:
CI failure is unrelated

PR #3462, opened 2022-12-16
feat(DENG-609): Added extra note in fxa_all_events around oauth events

kik-kik:
# feat(DENG-609): Added extra note in fxa_all_events around oauth events

This is due to the fact that oauth log event table no longer exists and has been merged into auth server/event log. This change aims to reflect this change.


relud (approved):
r+wc

The oauth events table still has 162 days of data from 2019H2, that's farther back than we usually care, but I want to make sure this is known before we remove access to it through this view.

looks like this breaks a test, which should be fixed:
tests/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_daily_v1/test_aggregation

kik-kik:
@relud thanks for the comment, decided to make a change to this PR and keep `oauth` events accessible via the view, but added a not at the top of the view definition. Please let me know what you think.

relud (approved):

PR #3464, opened 2022-12-16
feat(DENG-610): added view for fxa_users_daily and fxa_users_first_seen built from fxa_users_daily

kik-kik:
# feat(DENG-610): added view for fxa_users_daily and fxa_users_first_seen built from fxa_users_daily


sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/query.sql
```diff
@@ -1,20 +1,20 @@
 WITH _current AS (
-  SELECT
+  SELECT DISTINCT
     user_id,
-    ARRAY_AGG(DISTINCT service IGNORE NULLS) AS services_used,
+    -- TOOD: should this should be dropped? We have fxa_users_services_first_seen to answer this kind of question.
+    -- ARRAY_AGG(DISTINCT service IGNORE NULLS) AS services_used,ß
```
fbertsch, line 5: The way to answer this is three parts:
1. Are people using this field? If so, keep it.

2. Is it possible queries that would target this field can be answered with this new table + `fxa_users_services_first_seen`. For example, asking for the first seen date for users of a specific service:
```
SELECT ...
FROM fxa_user_first_seen_v1
JOIN fxa_users_services_first_seen USING (user_id)
WHERE fxa_users_services_first_seen.service = '$service'
```

3. Does this field add an incremental ease of querying for common query patterns over (2.)? If so, consider keeping it.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/query.sql
```diff
@@ -23,22 +23,8 @@ _previous AS (
...
-  ) AS services_used,
+  DATE(@submission_date) AS first_seen_date,
 FROM
-  _previous
-FULL OUTER JOIN
   _current
-USING
-  (user_id)
+WHERE
+  user_id NOT IN (SELECT user_id FROM _previous)
```
fbertsch, line 30: Can we do this with a join instead?

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/query.sql
kik-kik: I think my main concern is that logically I do not think `services_used` belongs in this model and should live in some kind of `fxa_users_profile` or `fxa_users_usage_overview` model instead. To me first_seen implies we record the state as it at the creation time and is immutable.

Also, this model to me can be a bit counter-intuitive. In that we have a date with a corresponding list of services that user has used in general and not on that date specifically. Unless the end user of this model has that knowledge the model could be easily misinterpreted. 

I hope that explained the rationale behind proposing this change.

fbertsch: I'll leave it to you then :) Misuse b/c of misunderstanding is a very good thing at avoid. Just be cautious about changing the schema in case people are using that field.

kik-kik: Will do, and I do think this field as a quick access field could have value. So we still should provide it, but I think it should live elsewhere. I'll make the rest of the changes to have this PR ready and will only merge once I'm confident it won't break anything.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/query.sql
```diff
@@ -1,44 +1,37 @@
...
-    firefox_accounts_derived.fxa_users_first_seen_v1
+    `firefox_accounts.fxa_users_first_seen`
   WHERE
     -- In reprocessing scenarios, we must always backfill from the first affected date
     -- all the way to the present; to enforce that, we explicitly drop any data after
     -- the target @submission_date
     first_seen_date < @submission_date
+    -- it appears we have a user_id will null value exist inside fxa_users_first_seen
+    -- excluding it here to avoid it being recreated for each new submission_date
+    AND user_id IS NOT NULL
```
kik-kik, line 25: This is something I noticed in the current fxa_users_first_seen table, I assume this is a bug. What do you think @fbertsch ?

fbertsch: IMO this should be dropped upstream. Not bad to also include it here though.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/metadata.yaml
```diff
@@ -13,17 +12,9 @@ scheduling:
   start_date: '2021-07-09'
   priority: 80
   depends_on_past: true
-  # This query updates the entire existing table every day rather than appending
-  # a new partition, so we need to disable date_partition_parameter and instead
-  # pass submission_date as a generic param.
-  date_partition_parameter: null
-  parameters:
-  - submission_date:DATE:{{ds}}
+  date_partition_parameter: submission_date
```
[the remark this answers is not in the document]
kik-kik: Let me explain, this is referring to `@submission_date` macro in the corresponding SQL template/query. Here's an example where it's being used:
https://github.com/mozilla/bigquery-etl/pull/3464/files#diff-54b80d676f877f5692e9ed1363cb9c7c8747e35ed414ad24f575939ab737f3acR9

As far as the naming goes, I decided to leave it as this seems to be "de facto" a standard name of the parameter for passing in Airflow's execution_date. I'm open to alternative suggestions.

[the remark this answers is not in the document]
fbertsch: Right, but that will also create a partition and a new column called `submission_date`. See https://github.com/mozilla/telemetry-airflow/blob/main/dags/utils/gcp.py#L233

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/query.sql
kik-kik: As far as I can see we drop it now:
https://github.com/mozilla/bigquery-etl/blob/945c4ee0301020d98567e7a9f3cd8d8c8d37f1c5/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_daily_v1/query.sql#L55

This might be a left over from a very long time, as the way currently this model is structured that each time we include all previous records which results in user_id null entry to exist for each date:

```sql
SELECT MIN(first_seen_date)
FROM `firefox_accounts_derived.fxa_users_first_seen_v1`
WHERE
  first_seen_date >= "2021-01-01"
  AND user_id IS NULL
```
Returns: `2021-07-14`.

My guess is that ever since we see 1 entry per date with user_id being null since then because we just keep including that row due to how the logic is structured. I'm not sure if it's worth to re-compile the table since then though.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/metadata.yaml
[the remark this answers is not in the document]
kik-kik: gotcha, I see now. I made a tweak to the `metadata.yaml` which should fixed the issue.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/query.sql
kik-kik: So the only place where I can see that field (`services_used`) being referenced is `fxa_amplitude_export_v1`. Do you happen to know @fbertsch if we still need this table? I did not find any usage when I did a global search for it inside bqetl, looker-spoke and looker-hub.

Also, tried an org wide search on Github: https://github.com/search?q=org%3Amozilla%20fxa_amplitude_export_v1&type=code

In Looker it seems to be a hidden field:
https://github.com/search?q=org%3Amozilla+services_used&type=code

My impression is it only existed to support `fxa_amplitude_export_v1`, is there anything you think I missed?

fbertsch: We _don't_ need the amplitude export anymore, since we don't export to Amplitude. If `services_user` isn't otherwise used, then we can probably drop it.

fbertsch (approved):

gkabbz (approved):

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v1/query.sql
akkomar: Rebuilding this table from fxa_users_daily will cost ~$5 (1.11 TB) so I'd say it's worth it. Note that the init script still points to fxa_all_events which would process 7.91 TB.

kik-kik: IMHO we should be using the daily table for building the first seen table. The question is whether we should keep the init.sql at all. If I'm correct in saying it our old pattern for initialising a table. The new approach is to have a schema.yaml and using our backfill process for populating the table.

kik-kik:
ok, I decided to abandon this change and instead make `fxa_users_first_seen_v2` which will supersede `fxa_users_first_seen_v1`. This way both of them can run in parallel and once `fxa_users_first_seen_v2` catches up we can make a comparison and if all is in order proceed with deprecating `fxa_users_first_seen_v1`.

PR #3481, opened 2022-12-30
Increase Fivetran sensors' poke interval to 30 seconds

sean-rose:
Using a poke interval of 5 seconds means only 7 Fivetran sensors polling for an hour would exceed [Fivetran's rate limit](https://fivetran.com/docs/rest-api/rate-limiting) of 5,000 API requests per hour.  Increasing the poke interval to 30 seconds will allow for up to 41 Fivetran sensors polling for an hour.

This should wait until #3480 is merged (and this PR will be re-targeted to the `main` branch at that point).

CCing other DAG owners for visibility: @kik-kik

---
Checklist for reviewer:

### DIFF
--- a/bigquery_etl/query_scheduling/templates/airflow_dag.j2
+++ b/bigquery_etl/query_scheduling/templates/airflow_dag.j2
@@ -218,7 +218,7 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
     {{ fivetran_task.task_id }}_sync_wait = FivetranSensor(
         connector_id='{% raw %}{{ var.value.{% endraw %}{{ fivetran_task.task_id }}{% raw %}_connector_id }}{% endraw %}',
         task_id='{{ fivetran_task.task_id }}_sensor',
-        poke_interval=5,
+        poke_interval=30,
         xcom="{% raw %}{{{% endraw %} task_instance.xcom_pull('{{ fivetran_task.task_id }}_task') {% raw %}}}{% endraw %}",
         on_retry_callback=retry_tasks_callback,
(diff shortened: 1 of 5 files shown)


bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -218,7 +218,7 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
     {{ fivetran_task.task_id }}_sync_wait = FivetranSensor(
         connector_id='{% raw %}{{ var.value.{% endraw %}{{ fivetran_task.task_id }}{% raw %}_connector_id }}{% endraw %}',
         task_id='{{ fivetran_task.task_id }}_sensor',
-        poke_interval=5,
+        poke_interval=30,
```
sean-rose, line 221: I chose 30 seconds because some Fivetran sensors' poke interval had already been increased to 30 seconds and it seemed best to keep things consistent.

However, the [default Fivetran sensor poke interval](https://github.com/fivetran/airflow-provider-fivetran/blob/59f046ead0b81824e72b438b5475be2d3322de26/fivetran_provider/sensors/fivetran.py#L50) is 60 seconds, so another option would be to remove the custom poke interval settings and just go with the default.  That would give us even more headroom, with the downside being a bit more latency in detecting sync completions/failures.

@scholtzan it looks like you added the previous poke interval setting of 5 seconds in #2640.  Did you have a particular reason for choosing 5 seconds then, and do you have an opinion on what to increase it to now?

scholtzan (approved):

PR #3496, opened 2023-01-11
Remove dependencies on the `mozdata` project from ETL

sean-rose:
Having dependencies on things in `mozdata` can cause issues with deployments, as deploying things to `mozdata` is usually a separate secondary step.

---
Checklist for reviewer:

- [ ] Commits should reference a bug or github issue, if relevant (if a bug is referenced, the pull request should include the bug number in the title).
- [ ] If the PR comes from a fork, trigger integration CI tests by running the [Push to upstream workflow](https://github.com/mozilla/bigquery-etl/actions/workflows/push-to-upstream.yml) and provide the `<username>:<branch>` of the fork as parameter. The parameter will also show up
in the logs of the `manual-trigger-required-for-fork` CI task together with more detailed instructions.
- [ ] If adding a new field to a query, ensure that the schema and dependent downstream schemas have been updated.
- [ ] When adding a new derived dataset, ensure that data is not available already (fully or partially) and recommend extending an existing dataset in favor of creating new ones. Data can be available in the [bigquery-etl repository](https://github.com/mozilla/bigquery-etl), [looker-hub](https://github.com/mozilla/looker-hub) or in [looker-spoke-default](https://github.com/mozilla/looker-spoke-default/tree/e1315853507fc1ac6e78d252d53dc8df5f5f322b).

For modifications to schemas in restricted namespaces (see [`CODEOWNERS`](/CODEOWNERS)):
- [ ] Follow the [change control procedure](https://docs.google.com/document/d/1TTJi4ht7NuzX6BPG_KTr6omaZg70cEpxe9xlpfnHj9k/edit#heading=h.ttegrcfy18ck)

PR #3504, opened 2023-01-13
Fix bug generating task dependencies when queries reference stable views in `mozdata`

sean-rose:
While we don't want people to reference such views in `mozdata` and will be removing existing references to such views (#3496), if people do reference them then the task dependencies should still be generated correctly.

---
Checklist for reviewer:

### DIFF
--- a/dags/bqetl_main_summary.py
+++ b/dags/bqetl_main_summary.py
@@ -529,6 +529,20 @@ with DAG(
         depends_on_past=False,
     )
 
+    wait_for_copy_deduplicate_all = ExternalTaskSensor(
+        task_id="wait_for_copy_deduplicate_all",
+        external_dag_id="copy_deduplicate",
+        external_task_id="copy_deduplicate_all",
+        execution_delta=datetime.timedelta(seconds=3600),
+        check_existence=True,
+        mode="reschedule",
+        allowed_states=ALLOWED_STATES,
+        failed_states=FAILED_STATES,
+        pool="DATA_ENG_EXTERNALTASKSENSOR",
+    )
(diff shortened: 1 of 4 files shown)


dags/bqetl_main_summary.py
```diff
@@ -529,6 +529,20 @@
...
+        external_task_id="copy_deduplicate_all",
+        execution_delta=datetime.timedelta(seconds=3600),
+        check_existence=True,
+        mode="reschedule",
+        allowed_states=ALLOWED_STATES,
+        failed_states=FAILED_STATES,
+        pool="DATA_ENG_EXTERNALTASKSENSOR",
+    )
+
+    crashes_daily_v1.set_upstream(wait_for_copy_deduplicate_all)
```
sean-rose, line 544: In all the other cases the affected tasks had other dependencies on tasks downstream from `copy_deduplicate_all`, but this `telemetry_derived.crashes_daily_v1` task had no other dependencies so it could have ended up running before `copy_deduplicate_all` finished successfully (and wouldn't be re-run if `copy_deduplicate_all` was cleared recursively downstream).

It looks like that did happen at least once, as the `telemetry_derived.crashes_daily_v1` table has zero rows for 2022-12-01, so we should re-run this ETL for that date.

scholtzan (approved):

dags/bqetl_main_summary.py
sean-rose: I have re-run the `telemetry_derived.crashes_daily_v1` ETL for 2022-12-01 along with all downstream tasks.

PR #3505, opened 2023-01-13
Add --skip-dryrun option for CLI validation commands

scholtzan:
Adding `--skip-dryrun` options to some validation commands. This is for adding CI tests to private-bigquery-etl: https://github.com/mozilla/private-bigquery-etl/pull/179 Most of the views/queries in that repo are restricted, so dryruns will always fail. Running some checks is still useful however, e.g. to detect incorrect references in views (which has caused some issues in the past).

Checklist for reviewer:

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -25,6 +25,7 @@ from ..cli.format import format
 from ..cli.utils import (
     is_authenticated,
     is_valid_project,
+    no_dryrun_option,
     paths_matching_name_pattern,
     project_id_option,
     respect_dryrun_skip_option,
@@ -1085,6 +1086,7 @@ def _run_part(
     default=False,
 )
 @respect_dryrun_skip_option(default=False)
+@no_dryrun_option(default=False)
 @click.pass_context
 def validate(
     ctx,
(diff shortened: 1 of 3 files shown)


bigquery_etl/cli/view.py
```diff
@@ -121,14 +124,17 @@ def validate(
...
+        ctx.invoke(
+            dryrun,
+            paths=[str(f) for f in view_files],
+            use_cloud_function=use_cloud_function,
+            project=project_id,
+            validate_schemas=validate_schemas,
+            respect_skip=respect_dryrun_skip,
+        )
+    else:
+        click.echo("Skip view dryruns.")
```
sean-rose, line 137: For consistency with the similar message when validating queries.
```suggestion
        click.echo("Dry run skipped for view files.")
```

bigquery_etl/cli/query.py
```diff
@@ -1085,6 +1086,7 @@ def _run_part(
     default=False,
 )
 @respect_dryrun_skip_option(default=False)
+@skip_dryrun_option(default=False)
```
sean-rose, line 1089: It might be a bit confusing to people how the existing `--ignore-dryrun-skip` option and this new `--skip-dryrun` option would relate/interact.  Maybe having the new option avoid the term "skip" would help? (e.g. `--no-dryrun`).

sean-rose (approved):
r+wc

bigquery_etl/cli/query.py
scholtzan: good point, I'll rename it

PR #3508, opened 2023-01-16
Move check for dags being up to date after DAGs have been generated

scholtzan:
We currently have a separate `verify-dags-up-to-date` task which runs separately from `generate-dags`. We can move the up-to-date check into the `generate-dags` (this way we'll only need to run the generation once) and with this DAGs can now also reference and schedule generated content (which wasn't available in `verify-dags-up-to-date`)

Checklist for reviewer:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -206,7 +206,12 @@ jobs:
           name: Generate DAGs
           command: |
             PATH="venv/bin:$PATH" script/bqetl dag generate
-            mv dags/ /tmp/workspace/generated-sql
+            cp -R dags/ /tmp/workspace/generated-sql
+      - run:
+          name: Verify that DAGs were correctly generated and are up-to-date
+          command: |
+            git diff --exit-code dags/
+            diff <(git ls-files dags/*.py) <(ls dags/*.py)
       # this task is overwriting the content produced by generate-sql;
       # the behaviour here is additive, generated DAGs are just added to
       # the generated-sql output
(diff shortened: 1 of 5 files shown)


sql/moz-fx-data-shared-prod/fenix_derived/new_profile_activation_v1/metadata.yaml
```diff
@@ -16,3 +16,16 @@ bigquery:
 scheduling:
   dag_name: bqetl_mobile_activation
   date_partition_parameter: submission_date
+  referenced_tables:
+  - - 'moz-fx-data-shared-prod'
+    - 'org_mozilla_firefox_stable'
+    - 'first_session_v1'
```
scholtzan, line 22: This is only available after running the SQL generators. Specifying it here explicitly so that if users run generating the DAGs without generating SQL, it is still correctly picked up as dependency

dags/bqetl_mobile_activation.py
```diff
@@ -81,6 +81,21 @@
...
+        check_existence=True,
+        mode="reschedule",
+        allowed_states=ALLOWED_STATES,
+        failed_states=FAILED_STATES,
+        pool="DATA_ENG_EXTERNALTASKSENSOR",
+    )
+
+    fenix_derived__new_profile_activation__v1.set_upstream(
+        wait_for_copy_deduplicate_all
+    )
```
scholtzan, line 98: Added due to `org_mozilla_firefox_stable.first_session_v1`

kik-kik (approved):
The change looks fine to me.

One thing I think we could consider is to wrap some of the CI commands in a Makefile to make them easily reproducible locally. For example, verify-dags-up-to-date could be executed using something like `make dag-verify-up-to-date` which would execute all corresponding shell commands and produce the same output as CI would (this is outside of the scope of this PR).

.circleci/config.yml
```diff
@@ -206,7 +206,12 @@ jobs:
           name: Generate DAGs
           command: |
             PATH="venv/bin:$PATH" script/bqetl dag generate
-            mv dags/ /tmp/workspace/generated-sql
+            cp -R dags/ /tmp/workspace/generated-sql
```
sean-rose, line 209: `cp` does different things depending on how trailing slashes are used, and this is going to put all the DAG files directly in `/tmp/workspace/generated-sql`.

Fix:
```suggestion
            cp -R dags /tmp/workspace/generated-sql
```

PR #3513, opened 2023-01-18
Clients Daily: Replace keyed scalars with flat sums for urbar metrics

rebecca-burwei:
This PR:
1. Adds a new UDF `mozfun.map.extract_keyed_scalar_sum` that sums all values in a keyed scalar.
2. Uses the new UDF in clients daily to sum the `urlbar_picked_*` and `contextual_services_*` keyed scalars.
3. Removes the `urlbar_picked_*` and `contextual_services_*` as keyed scalars from clients daily, leaving on the flat sums.

Reason for this change: The keyed scalars break out counts by urlbar position. I have never used this breakout in clients daily. When such breakouts are needed, I use urlbar_clients_daily. Since I am often summing these keyed scalars in clients daily, I think it would be more efficient to replace the keyed scalars with flat sums. Also, I need the flat sums in Looker.

@skahmann3 , are these changes okay with you? I think you might use:
- contextual_services_topsites_impression
- contextual_services_topsites_click

Checklist for reviewer:


rebecca-burwei:
Hi @curtismorales , could you help with merging this PR and backfilling?

skahmann3:
@rebecca-burwei +1 that it would be great to have a sum of clicks and impressions in clients_daily. I see that the `contextual_services_topsites_impression_sum` keys include `urlbar_1` and `urlbar_2`. They currently have very low counts, but was that a Suggest field at some point? My only hesitation would be if that field starts being used again and we don't realize that's being added in with topsites. Otherwise, we aren't currently planning on breaking out counts by the remaining keys, `newtab_1` or `newtab_2` at this time, and we can always add later if necessary

rebecca-burwei:
Hi @curtismorales -- This PR is now ready for your review and backfill.

sql/moz-fx-data-shared-prod/telemetry/clients_daily_v6/view.sql
```diff
@@ -13,7 +13,161 @@ SELECT
...
+    ) AS scalar_parent_urlbar_picked_topsite_sum,
+    `mozfun.map.extract_keyed_scalar_sum`(
+      scalar_parent_urlbar_picked_unknown_sum
+    ) AS scalar_parent_urlbar_picked_unknown_sum,
+    `mozfun.map.extract_keyed_scalar_sum`(
+      scalar_parent_urlbar_picked_visiturl_sum
+    ) AS scalar_parent_urlbar_picked_visiturl_sum,
+    -- `mozfun.map.extract_keyed_scalar_sum`(
+    --   contextual_services_quicksuggest_block_dynamic_wikipedia_sum
+    -- ) AS contextual_services_quicksuggest_block_dynamic_wikipedia_sum,
```
curtismorales, line 85: The commented out fields are the fields that don't exist yet. Since these are new, I'm thinking we can do the extract in the table and then delete these commented out lines?

rebecca-burwei:
Hi @curtismorales -- I have updated query.sql so that it just adds the new fields.

curtismorales (approved):

ANich:
cc @rebecca-burwei @curtismorales how should these changes be reflected in `telemetry_derived.urlbar_clients_daily_v1`? There's a downstream failure due to the change in types for the `scalar..sum` columns.

rebecca-burwei:
@ANich , Curtis should double check this is right, but I think you can pull from the clients daily table instead of the view.

curtismorales:
Sorry, missed this one, most of the other downstream tables pull from the table rather than the view.

Pointing at the table is an option, but maybe we can also just use the transformed value? @ANich I'll message you and see if I can help

PR #3517, opened 2023-01-19
Add bootstrap_percentile_ci UDF

bochocki:
This PR adds a UDF to calculate more accurate bootstrap samples from histograms.

Improvements over the existing UDF (`jackknife_percentile_ci`):

- [Uses the boostrap confidence intervals method described here](https://engineering.atspotify.com/2022/03/comparing-quantiles-at-scale-in-online-a-b-testing/).
- Constant time and space complexity
- Uses empirical confidence intervals -- guaranteed to provide sensible results.
- Accounts for histogram bin widths when reporting confidence intervals -- guaranteed to not provide zero-width bins.

Ticket: https://mozilla-hub.atlassian.net/browse/DS-2128
Slides: https://docs.google.com/presentation/d/1uUR0i9ox0K6IkmG68tlNASnGloXNSi_ZZKThyVRM7tg

Checklist for reviewer:

- [ ] ~If adding a new field to a query, ensure that the schema and dependent downstream schemas have been updated.~
- [ ] ~When adding a new derived dataset, ensure that data is not available already (fully or partially) and recommend extending an existing dataset in favor of creating new ones. Data can be available in the [bigquery-etl repository](https://github.com/mozilla/bigquery-etl), [looker-hub](https://github.com/mozilla/looker-hub) or in [looker-spoke-default](https://github.com/mozilla/looker-spoke-default/tree/e1315853507fc1ac6e78d252d53dc8df5f5f322b).~

- [ ] ~Follow the [change control procedure](https://docs.google.com/document/d/1TTJi4ht7NuzX6BPG_KTr6omaZg70cEpxe9xlpfnHj9k/edit#heading=h.ttegrcfy18ck)!~


sql/moz-fx-data-shared-prod/udf_js/bootstrap_percentile_ci/udf.sql
```diff
@@ -0,0 +1,212 @@
...
+)
+RETURNS ARRAY<STRUCT<
+    metric STRING,
+    statistic STRING,
+    point FLOAT64,
+    lower FLOAT64,
+    upper FLOAT64,
+    parameter STRING
+>> DETERMINISTIC
+LANGUAGE js OPTIONS ( library=['gs://moz-data-science-bochocki-bucket/bq_udf_library/qbinom.js'])
```
bochocki, line 26: This library needs to be moved to another location, but I'm not sure where to put it.

sql/moz-fx-data-shared-prod/udf_js/bootstrap_percentile_ci/udf.sql
```diff
@@ -0,0 +1,212 @@
...
+    if (percentile < 0 || percentile > 100) {
+      throw "percentile must be a value between 0 and 100";
+    }
+  }
+  
+  function getQuantileSamples(percentile, histogram, n_samples) {
+    validatePercentile(percentile);
+
+    /* expand the histogram */
+    var full_histogram = histogram.map(bin => Array(bin.value).fill(parseFloat(bin.key))).reduce((prev, curr) => prev.concat(curr));
```
scholtzan, line 43: Just noticing that currently we compute the normalized sum of the histograms and pass them into the forkknife percentile UDF. This wouldn't work here (since `fill` on floats will fail). I am not 100% sure why we compute the normalized sums in the first place, but it seems like with this we can just pass the histograms into the UDFs and not worry about it? Or will this have some implications?

scholtzan (commented):
I think CI is just failing because you haven't logged in to CircleCI? Or maybe you need to log out and log in again.

sql/moz-fx-data-shared-prod/udf_js/bootstrap_percentile_ci/udf.sql
bochocki: Good catch! The previous version normalized the histogram twice I think. Once in the UDF [here](https://github.com/mozilla/bigquery-etl/blob/4ed4bd07be3281c3571720c19172c4acee117489/sql/moz-fx-data-shared-prod/udf_js/jackknife_percentile_ci/udf.sql#L25), and once (I _think_) for the UDF input, [based on the templating here](https://github.com/mozilla/opmon/blob/3291261a49d01efff00a55aedf1e0abb996a1fb7/opmon/templates/statistics_query.sql#L3).

This proposed approach skips the normalization step by simply using the location of values in a sorted array (e.g. `1` is the median value in the array `[0, 1, 1, 2]`). The result is the same, but the proposed approach has better time complexity (it's just a lookup instead of performing the normalization and looping through the histogram).

The simplicity comes with the memory cost of having to reshape the histogram into an array, but the proposed approach requires the data to be in an array form anyway, so we're really just taking advantage of the shape of the data.

One positive side-effect of this is that we could modify this UDF (or create a slightly different one) to accept a scalar array as input. In that case, we wouldn't have to convert scalar columns to histograms before computing confidence intervals, we could just use the columns as-is.

bochocki: To answer your question more directly, we'll need to make sure that we don't normalize the histograms that get passed in.

sql/moz-fx-data-shared-prod/udf_js/bootstrap_percentile_ci/udf.sql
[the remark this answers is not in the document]
bochocki: Thanks! I added `sql/moz-fx-data-shared-prod/udf_js_lib/qbinom.js`

sql/moz-fx-data-shared-prod/udf_js/bootstrap_percentile_ci/udf.sql
```diff
@@ -0,0 +1,212 @@
...
+    numericSort(full_histogram);
```
scholtzan, line 44: As discussed, this is causing `UDF memory exceeded` errors for certain histograms like `gc_ms`. But it _should_ be safe to remove at least this operation since it seems to be guaranteed that histograms remain sorted.

We should add some kind of check though to throw an error if that is not the case.

bochocki: https://github.com/mozilla/bigquery-etl/pull/3517/commits/79997ec3355412759165b1dd58ed736d2e235813 changes the order of operations so that we first sort the histogram based on the keys, then we build the `full_histogram`. As we talked about offline, sorting the histogram is much faster and the `map` method will process the keys in order.

sql/moz-fx-data-shared-prod/udf_js/bootstrap_percentile_ci/udf.sql
```diff
@@ -0,0 +1,243 @@
...
+          }
+        };
+      }
+    }
+
+    /* ensure the histogram is sorted */
+    histogramSort(samples);
+
+    /* expand the output */
+    var output = samples.map(bin => Array(bin.value).fill(parseFloat(bin.key))).reduce((prev, curr) => prev.concat(curr));
```
scholtzan, line 77: I'm getting `Reduce of empty array with no initial value`. Does this happen for bins that are 0?

scholtzan: something like 
```
if (arr.length > 0) {
      var output = arr.reduce((prev, curr) => prev.concat(curr));
    } else {
      var output = [];
    }
```
does cause it to run successfully, but I'm not sure if the results are still correct with this?

bochocki: Thanks, great catch! I think you're right, this is caused when all of the histogram bins contain 0 counts (e.g. `[{'key': 1, 'value': 0, 'key': 2, 'value': 0}]`). I am now explicitly checking for that case with `histogramIsNonZero`. I added a test for this case, which failed under the previous version and is passing now.

For these cases, the UDF will now return `NULL` for `lower` / `point` / `upper`.

scholtzan (approved):
This is great! 🎉 

I tested it on a couple of existing rollouts and it worked fine. Numbers are looking good as well.

I re-formatted the UDF so that CI tests will pass. From my side, once CI is happy this should be good to get merged.

sql/moz-fx-data-shared-prod/udf_js/bootstrap_percentile_ci/udf.sql
```diff
@@ -0,0 +1,337 @@
...
+SELECT
+  assert.equals('test', (SELECT DISTINCT metric FROM all_zero_results)),
+  assert.equals('percentile', (SELECT DISTINCT statistic FROM all_zero_results)),
+  assert.array_equals(
+    (SELECT ['5', '25', '50', '75', '95']),
+    (SELECT ARRAY(SELECT parameter FROM all_zero_results))
+  ),
+  assert.null((SELECT ARRAY(SELECT lower FROM all_zero_results WHERE lower IS NOT NULL))),
+  assert.null((SELECT ARRAY(SELECT point FROM all_zero_results WHERE point IS NOT NULL))),
+  assert.null((SELECT ARRAY(SELECT upper FROM all_zero_results WHERE upper IS NOT NULL))),
```
scholtzan, line 269: I made an update to these tests since they were failing with ` Array cannot have a null element; error in writing field f0_ ` previously.

sql/moz-fx-data-shared-prod/udf_js/bootstrap_percentile_ci/udf.sql
```diff
@@ -0,0 +1,337 @@
...
+          /* ignore this empty bin when drawing samples */
+          bin.cumulative_sum = -1;
+        }
+    };
+
+    var samples = [];
+    /* sample the quantile of interest */
+    for ( var i = 0; i < n_samples; i++ ) {
+      /* draw a random variable from the binomial distribution */
+      var x = qbinom(Math.random(), total, parseFloat(percentile) / 100);
```
scholtzan, line 72:
```suggestion
      var x = qbinom(Math.random(), Math.ceil(total), parseFloat(percentile) / 100);
```

`qbinom` doesn't seem to be able to handle floats. When normalizing histograms values usually become floats, so we need to handle this somehow. I'm not sure if simply rounding it to the closest integer is good enough here?

bochocki: This was a great catch. The fix is here: (https://github.com/mozilla/bigquery-etl/pull/3517/commits/e17e136dd53e6ace1a4dff1c1379e8b73ad0b3bc). The quick fix you proposed was most of the solution, but we needed to make some other changes to get the sampling to behave properly. Essentially:
```js
const scale = 1000;
total = Math.ceil(total * scale);
var x = qbinom(Math.random(), total, parseFloat(percentile) / 100) / total;
```
plus a few other changes in how we handle the samples, since they can now take on non-integer values.

scholtzan (approved):
CI errors should be fixed now. Good to get merged 👍

PR #3520, opened 2023-01-23
Edits to search_revenue_levers_daily

curtismorales:
I made two sets of changes to your query:

1. The first couple commits are just fixes and getting the query running
2. With the rest, I rewrote the query in a way that I think makes it easier to read and work with it

There are two things I wanted to go over with you -- maybe we can discuss in our 1 on 1?

1. I believe I've mostly recreated what you're trying to do in your query, except for the desktop DAU -- I'm not sure exactly what number we want to use there
2. I'm still wondering if the extra metrics here could just be added to the `search_aggregates` tables and this table built on top of that (sorry if I'm still not understanding why that doesn't work)

### DIFF
--- a/dags/bqetl_search_dashboard.py
+++ b/dags/bqetl_search_dashboard.py
@@ -87,6 +87,54 @@ with DAG(
         depends_on_past=False,
     )
 
+    wait_for_telemetry_derived__clients_last_seen__v1 = ExternalTaskSensor(
+        task_id="wait_for_telemetry_derived__clients_last_seen__v1",
+        external_dag_id="bqetl_main_summary",
+        external_task_id="telemetry_derived__clients_last_seen__v1",
+        execution_delta=datetime.timedelta(seconds=7200),
+        check_existence=True,
+        mode="reschedule",
+        allowed_states=ALLOWED_STATES,
+        failed_states=FAILED_STATES,
+        pool="DATA_ENG_EXTERNALTASKSENSOR",
+    )
+
+    search_derived__desktop_search_aggregates_by_userstate__v1.set_upstream(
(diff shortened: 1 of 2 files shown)


xluo-ds:
> There are two things I wanted to go over with you -- maybe we can discuss in our 1 on 1?
> 
>     1. I believe I've mostly recreated what you're trying to do in your query, except for the desktop DAU -- I'm not sure exactly what number we want to use there

For both desktop and mobile, I think we have two DAU related metrics here, one is called `dau` which is just to measure the total user base on that day, and the other is called `dau_engaged_w_sap` which is to measure the size of user engage with the search access points from browser that day. They measure different stuff, and we would like to keep both in the data table. 

> 
>     2. I'm still wondering if the extra metrics here could just be added to the `search_aggregates` tables and this table built on top of that (sorry if I'm still not understanding why that doesn't work)

I think it's possible for the desktop as it's solely based on `search_clients_engines_sources_daily`, but for mobile it might be tricky as it's based on other table to count `DOU/DAU` besides just `mobile_search_clients_engines_sources_daily`. Then we will need to combine them into one that I can use in Looker, when we do this it might be difficult as we will put things from different places (like desktop could be just from `search_aggregates` while for mobile you start the logic again), it might be difficult for people to capture the logic? While it we pull it out this way and have the logic exists in one place. That's my original thinking, but open to your feedback.

xluo-ds:
@curtismorales  for Desktop Bing, we shall update the logic to remove these clients in `mozdata.analysis.acer_cohort` from all calculation of `dou, dou_engaged_w_sap, sap, tagged_sap, tagged_follow_on, search_with_ads, ad_click`.

curtismorales:
Closing in favor of #3534

PR #3521, opened 2023-01-23
feat(DENG-649): deriving fxa users services first seen from users services daily

kik-kik:
# feat(DENG-649): deriving fxa users services first seen from users services daily

There are some tweaks to `fxa_users_services_daily_v1` so that it does most of the "lift" when it comes to the logic and accessing the events data. This enables for `fxa_users_services_first_seen_v1` to get all the information it needs from it and simplifying the FxA data model a bit.

The next step will be to have `fxa_users_v1` be derived from `fxa_users_services_daily_v1` as it should contain all needed information and avoid accessing the base events table directly.


sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/init.sql
```diff
@@ -1,3 +1,25 @@
...
+-- Expects an array of event_types and service name
+-- Iterates and checks if an event_type, service combination
+-- should be considered when determining if this is the first
+-- time it has been observed.
+-- Example usage:
+-- contains_qualified_fxa_activity_event(['fxa_activity'], NULL),  # returns TRUE
+-- contains_qualified_fxa_activity_event(['fxa_activity'], 'firefox-ios'),  # returns TRUE
+-- contains_qualified_fxa_activity_event(['fxa_login - complete'], NULL),  # returns FALSE
+-- contains_qualified_fxa_activity_event(['fxa_login - complete'], 'firefox-ios'),  # returns TRUE
+CREATE TEMP FUNCTION contains_qualified_fxa_activity_event(events ANY TYPE, `service` ANY TYPE) AS (
```
relud, line 11: rather than duplicate all of the logic from query.sql to init this table, it's probably worth making this query something that would only init the schema, such as a `SELECT ... WHERE FALSE` type query, like here https://github.com/mozilla/bigquery-etl/blob/d01ce2fc8a6193121dc961951c850cf4bf8bcc3e/sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v1/init.sql#L3-L18

relud (changes requested):

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/init.sql
kik-kik: If we take the approach to init the schema only then `init.sql` is probably redundant assuming schema.yaml is accurately reflecting the table structure?

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_daily_v1/query.sql
```diff
@@ -68,50 +91,123 @@ WITH base AS (
...
+  WINDOW
+    _window AS (
+      PARTITION BY
+        flow_id
+      ORDER BY
+        `timestamp` ASC
+      ROWS BETWEEN
+        UNBOUNDED PRECEDING
+        AND UNBOUNDED FOLLOWING
+    )
```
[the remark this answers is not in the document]
kik-kik: I see, so if any of those have a value, all values from that row should be considered not first value of each individually even if it comes from multiple rows.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/init.sql
relud: if i understand correctly, having an `init.sql` that produces an accurate schema allows the tooling to automatically update the `schema.yaml` when new fields are added to recursive tables (like clients_last_seen). not sure if that's relevant here, if it's not recursive probably leave out init.sql entirely.

kik-kik: Thanks, I just decided to delete it completely (at least for now).

kik-kik:
relud: I hope I made all the changes as you requested. Please let me know if there's anything else I should change.

---

A couple of other things I'd like to pick your brain with:

I wonder if it would make sense for me to update this to `v2`. I know the schema changes here should be fine, but wonder this is enough to justify doing that.

Also, do you think it would make more sense to have a separate table for containing flow_ids and their attributes like entrypoint, etc. And instead of the daily table having the "first_see" flow info only having an array of all flows the user / service combination has on each day in ASC order like the events field and in first seen table use `mode_last` to get the first flow_id and the mapping table to get the flows attributes? 

---

P.S. Sorry for changing the owner without asking you first. I assumed that'd be fine since I'm making changes here and it looks like I'll be working more with FxA data I thought it'd make sense to do so. But I should have checked first.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_daily_v1/query.sql
```diff
@@ -17,41 +20,44 @@ CREATE TEMP FUNCTION udf_contains_tier1_country(
...
   WHERE
-    event_category IN ('content', 'auth', 'oauth')
-    AND user_id IS NOT NULL
+    DATE(`timestamp`)
+    -- 2 day time window used to make sure we can get user session attribution information
+    -- which will not always be available in the same partition as active user activity
+    -- ('fxa_login - complete', 'fxa_reg - complete').
+    -- this includes fields such as entrypoint, utm's etc.
+    BETWEEN DATE_SUB(@submission_date, INTERVAL 1 DAY)
+    AND @submission_date
```
relud, line 59: I think ```DATE(`timestamp`)``` is on the wrong side of this comment
```suggestion
    -- 2 day time window used to make sure we can get user session attribution information
    -- which will not always be available in the same partition as active user activity
    -- ('fxa_login - complete', 'fxa_reg - complete').
    -- this includes fields such as entrypoint, utm's etc.
    DATE(`timestamp`)
    BETWEEN DATE_SUB(@submission_date, INTERVAL 1 DAY)
    AND @submission_date
```

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_daily_v1/query.sql
```diff
@@ -1,3 +1,6 @@
+-- Should this be a persisted UDF? tier1_country
+-- definition probably does not change between
+-- specific queries.
```
relud, line 3: let's leave this as is, tier1 country can have different definitions based on context (vpn's tier1 is not necessarily firefox desktop's tier1) and formatting (e.g. country name vs country code)

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_daily_v1/query.sql
```diff
@@ -68,50 +74,166 @@ WITH base AS (
...
+user_service_flow_entrypoints AS (
+  SELECT
+    user_id,
+    service,
+    ARRAY_AGG(
+      STRUCT(
+        flow_id,
+        flow_entrypoint_info.`timestamp`,
+        flow_entrypoint_info.entrypoint
+      ) IGNORE NULLS
```
relud, line 105: A structed defined this way will never be `NULL`, so `IGNORE NULLS` won't do anything unless you do something like this:
```suggestion
      IF(
        flow_entrypoint_info.entrypoint IS NOT NULL,
        STRUCT(
          flow_id,
          flow_entrypoint_info.`timestamp`,
          flow_entrypoint_info.entrypoint
        ),
        NULL
      ) IGNORE NULLS
```

relud: or you could add `flow_id` to `flow_entrypoint_info` in `flow_entrypoints`, so that here you just have:
```suggestion
      flow_entrypoint_info IGNORE NULLS
```

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_first_seen_v1/metadata.yaml
```diff
@@ -1,20 +1,23 @@
...
-# scheduling:
-#   dag_name: bqetl_fxa_events
-#   # We completely recreate this table every day;
-#   # making it incremental is possible but nuanced since it windows over
-#   # events that may cross the midnight boundary.
-#   date_partition_parameter: null
+scheduling:
+  dag_name: bqetl_fxa_events
+  date_partition_parameter: submission_date
+  referenced_tables:
```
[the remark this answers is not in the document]
kik-kik: It appears if you include the project id the auto part fails. I just removed the project_id from the reference and it appears the generation now works correctly.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_daily_v1/query.sql
```diff
@@ -68,50 +74,166 @@ WITH base AS (
...
+    user_service_flow_entrypoints.flow_entrypoint_info.flow_id,
+    user_service_flow_entrypoints.flow_entrypoint_info.`timestamp` AS flow_timestamp,
+    user_service_flow_entrypoints.flow_entrypoint_info.entrypoint,
+    -- flow utms
+    user_service_flow_utms.utm_info.utm_term,
+    user_service_flow_utms.utm_info.utm_medium,
+    user_service_flow_utms.utm_info.utm_source,
+    user_service_flow_utms.utm_info.utm_campaign,
+    user_service_flow_utms.utm_info.utm_content
+  ) AS first_daily_service_flow_info,
```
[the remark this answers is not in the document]
kik-kik: I think I can see what you mean now. For some reason I thought I was joining the utm info using flow_id rather than user, service combination for joining the data.

I guess the question would be here if that's what I should be doing?

kik-kik:
ok, I think I made all requested changes. Could you please let me know if there's anything else you think needs to be updated relud?

relud (approved):

PR #3524, opened 2023-01-24
DENG-514 Add windows version to active users aggregates

wwyc:
[DENG-514](https://mozilla-hub.atlassian.net/browse/DENG-514):  Updated `normalized_os_version` column to  use [windows_version_info udf](https://github.com/mozilla/bigquery-etl/blob/main/sql/mozfun/norm/windows_version_info/udf.sql) to capture the correct windows version in the `unified_metrics` table (upstream of `active_users_aggregrates`).  

Checklist for reviewer:

[DENG-514]: https://mozilla-hub.atlassian.net/browse/DENG-514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lucia-vargas-a:
@wwyc and I have agreed that the best solution is to calculate the OS Version in the view `telemetry.unified_metrics`.

When OS != Windows, then use os_version withouth any changes
When OS = Windows, apply the UDF to get the correct version. This UDF uses `windows_build_number` so data will change for Windows.

The benefits with this approach:
- We avoid adding a new column to `unified_metrics`, so not increase its wide/size. This data point is ultimately not required by final users.
- We don't need to modify any of the aggregates because the data is inherited.
- We don't need to backfill any data.

@irrationalagent are you happy with this change to the OS Version for Windows, or you prefer that we add a new column e.g. `windows_version` in the view?

sql/moz-fx-data-shared-prod/telemetry_derived/active_users_aggregates_v1/query.sql
```diff
@@ -29,6 +29,7 @@ WITH todays_metrics AS (
     search_with_ads,
     uri_count,
     active_hours_sum,
+    windows_build_number
```
lucia-vargas-a, line 32: We wouldn't need this field. but the UDF applied to the OS Version only when the OS = Windows, right?

lucia-vargas-a:
@winnie to solve the CI failure in verify-format.sql you can run `./bqetl format $(git ls-tree -d HEAD --name-only)`
- I think is good to have some eyes from Data Science as this changes the data os version datapoint and any place where it's used. Probably Leif as one of the table owners.

sql/moz-fx-data-shared-prod/telemetry_derived/unified_metrics_v1/query.sql
```diff
@@ -321,7 +321,14 @@ desktop AS (
     first_seen_date,
     days_since_seen,
     os AS normalized_os,
-    normalized_os_version,
```
lucia-vargas-a, line 324: Shall we have some validations of how data changes after this change? Different versions and counts. Maybe in analysis or temp.

wwyc: Thanks. Can you provide more context on how this validation would be done?  Would I create a table in a particular folder to compare counts before and after the change?

wwyc: New tables created for validations:  `analysis.um_windows_version_wchan`, `analysis.um_windows_version_wchan_1` for WHERE `submission_date = 2023-01-28`
AND `normalized_app_name = Firefox Desktop`

irrationalagent:
Sorry to be slow to comment here. 

**I think** we don't do this in the upstream clients_last_seen table, correct? There is an idiom used in other queries to identify "modern" versions of windows in e.g. clients_daily that is something like:

`CAST(LEFT(normalized_os_version,2) as float64) >= 10` 

([here's an example using clients_daily](https://sql.telemetry.mozilla.org/queries/84768/source).

I can't verify whether queries of UM use this... I'm pretty sure  that no "important" queries that I own do, but if someone else is using this pattern on unified metrics (it seems plausible that there are some queries out there), then their queries will break. Also, if someone is used to a certain behavior of `normalized_os_version` from clients_daily/clients_last_seen then they will be surprised when querying the same column in unified metrics. 

I do think this is an improvement however, so I also don't want to block it on query breakage. Maybe we can make an announcement via email and/or slack about the change?

bochocki (commented):
This looks good to me, but @irrationalagent brings up a good point. Is it possible to move this transformation upstream so that the column is represented consistently across tables?

PR #3530, opened 2023-01-25
[Bug 1812301] Publish only string typed labels

scholtzan:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1812301 and should unblock artifact deploys

### DIFF
--- a/bigquery_etl/metadata/publish_metadata.py
+++ b/bigquery_etl/metadata/publish_metadata.py
@@ -41,7 +41,11 @@ def publish_metadata(client, dataset, table, metadata):
         if metadata.description is not None:
             table.description = metadata.description
 
-        table.labels = metadata.labels
+        table.labels = {
+            key: value
+            for key, value in metadata.labels.items()
+            if isinstance(value, str)
+        }
 
         client.update_table(table, ["friendly_name", "description", "labels"])
     except yaml.YAMLError as e:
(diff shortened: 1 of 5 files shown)


akkomar (approved):
LGTM, but wouldn't it make sense to document that non-string labels are ignored? An inline comment might save future us some time figuring out what's going on.

bigquery_etl/metadata/publish_metadata.py
```diff
@@ -41,7 +41,11 @@ def publish_metadata(client, dataset, table, metadata):
         if metadata.description is not None:
             table.description = metadata.description
 
-        table.labels = metadata.labels
+        table.labels = {
+            key: value
+            for key, value in metadata.labels.items()
+            if isinstance(value, str)
```
kik-kik, line 47: Is there a specific reason to filter stuff out rather than convert to a string?

kik-kik: Or at the very least should we log that something has been filtered out?

scholtzan: there are too many restrictions on what characters are allowed in labels (doesn't allow `,`, for example). I don't think it's worth to implement any more complex logic just to get these custom labels which aren't likely be looked at anyway

PR #3538, opened 2023-01-27
Bug 1813015 - Account for new (optional) client_info field in VPN view

badboy:
This currently causes queries against this view to fail because of a field mismatch in the `client_info` subfields.
This field was recently added, but will only ever be available on Windows (for obvious reason).
Nonetheless the column exists everywhere.

Checklist for reviewer:

### DIFF
--- a/sql/moz-fx-data-shared-prod/mozilla_vpn/deletion_request/view.sql
+++ b/sql/moz-fx-data-shared-prod/mozilla_vpn/deletion_request/view.sql
@@ -25,7 +25,8 @@ SELECT
       client_info.os AS os,
       client_info.os_version AS os_version,
       client_info.telemetry_sdk_build AS telemetry_sdk_build,
-      client_info.build_date AS build_date
+      client_info.build_date AS build_date,
+      NULL AS windows_build_number -- will never exist on Android
     ) AS client_info,
     (
       SELECT AS STRUCT
--- a/sql/moz-fx-data-shared-prod/mozilla_vpn/deletion_request/view.sql
+++ b/sql/moz-fx-data-shared-prod/mozilla_vpn/deletion_request/view.sql
@@ -26,7 +26,9 @@ SELECT
       client_info.os_version AS os_version,
       client_info.telemetry_sdk_build AS telemetry_sdk_build,
       client_info.build_date AS build_date,
-      CAST(NULL AS INT64) AS windows_build_number  -- will never exist on Android
+      CAST(
+        NULL AS INT64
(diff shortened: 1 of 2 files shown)


scholtzan:
This fields is also missing in `moz-fx-data-shared-prod.mozilla_vpn.deletion_request`, could you add it there as well?

scholtzan (approved):

scholtzan:
This should also fix view deploys 🎉

sql/moz-fx-data-shared-prod/mozilla_vpn/deletion_request/view.sql
```diff
@@ -25,7 +25,8 @@ SELECT
       client_info.os AS os,
       client_info.os_version AS os_version,
       client_info.telemetry_sdk_build AS telemetry_sdk_build,
-      client_info.build_date AS build_date
+      client_info.build_date AS build_date,
+      NULL AS windows_build_number -- will never exist on Android
```
sean-rose, line 29:
```suggestion
      CAST(NULL AS INT64) AS windows_build_number  -- will never exist on Android
```

sql/moz-fx-data-shared-prod/mozilla_vpn/main/view.sql
```diff
@@ -25,7 +25,8 @@ SELECT
       client_info.os AS os,
       client_info.os_version AS os_version,
       client_info.telemetry_sdk_build AS telemetry_sdk_build,
-      client_info.build_date AS build_date
+      client_info.build_date AS build_date,
+      NULL AS windows_build_number -- will never exist on Android
```
sean-rose, line 29:
```suggestion
      CAST(NULL AS INT64) AS windows_build_number  -- will never exist on Android
```

sean-rose (approved):
r+wc

sean-rose:
Also, I just noticed that `mozilla_vpn.events` doesn't need to have a similar fix, because for some reason `org_mozilla_firefox_vpn.events` _does_ have a `client_info.windows_build_number` column, which seems weird given that it's Android.

badboy:
> Also, I just noticed that `mozilla_vpn.events` doesn't need to have a similar fix, because for some reason `org_mozilla_firefox_vpn.events` _does_ have a `client_info.windows_build_number` column, which seems weird given that it's Android.

I guess I was imprecise here:
`windows_build_number` is a field generally available on all tables as it is part of the `client_info`. However (for obvious reasons) this field is only ever non-null for Windows clients, which of course is not where Android apps run.

PR #3545, opened 2023-01-30
[Bug 1712332] Generate UNIONed app ping views

scholtzan:
This PR updates the `glean_usage` generator to create UNIONed views for Glean app-datasets across all channels. Previously these views were only created for release channels.

The tricky part is that schemas for the same pings are different across channels (in most cases the ordering is different, or fields are missing in one but present in another schema)

The approach here for each app and ping is to:
1. Merge schemas of ping tables across all channels into one schema. This ensures that this schema will have all fields that are present in at least one of the tables
2. Create a SELECT expression used when UNIONing the table of a specific channel with other channels. This select expression will set fields that aren't present in the source table (but exist for other channels) to `NULL` and will ensure that the ordering of fields is consistent across channels.


sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -66,26 +66,96 @@ def generate_per_app(self, project_id, app_info, output_dir=None):
...
+                )
+
+                try:
+                    unioned_schema.merge(schema, add_missing_fields=True)
+                except Exception as e:
+                    # if schema incompatibilities are detected, then only generate for release channel
+                    print(
+                        f"Cannot UNION 'moz-fx-data-shared-prod.{app_dataset}.{view_name}': {e}"
+                    )
+                    print("Generate view only for release channel.")
```
scholtzan, line 100: The only view I noticed this happening for is `fenix.migration` where type attributes for root.metrics.datetime.migration_telemetry_identifiers_fennec_profile_creation_date are incompatible. In some cases this field is a timestamp, in others it's a string

sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -66,26 +66,105 @@ def generate_per_app(self, project_id, app_info, output_dir=None):
...
+
+                    unioned_schema = Schema.for_table(
+                        "moz-fx-data-shared-prod",
+                        release_app["document_namespace"].replace("-", "_"),
+                        view_name,
+                        partitioned_by="submission_timestamp",
+                    )
+                    release_only = True
+
```
relud, line 109: seems like this loop should break here:
```suggestion
                    break
```

sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -66,26 +66,105 @@ def generate_per_app(self, project_id, app_info, output_dir=None):
...
+            # generate the SELECT expression used for UNIONing the stable tables;
+            # fields that are not part of a table, but exist in others, are set to NULL
+            datasets = []
+            for app in app_info if not release_only else [release_app]:
+                app_dataset = app["document_namespace"].replace("-", "_")
+
+                # compare table schema with unioned schema to determine fields that need to be NULL
+                schema = Schema.for_table(
```
relud, line 117: seems like a waste to pull this twice instead of re-using from the previous loop

sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -66,26 +66,105 @@ def generate_per_app(self, project_id, app_info, output_dir=None):
...
                     output_dir,
                     full_view_id,
                     "view.sql",
-                    reformat(
-                        VIEW_QUERY_TEMPLATE.format(
-                            full_view_id=full_view_id,
-                            target=underlying_view_id,
-                        )
-                    ),
+                    reformat(rendered_view),
```
relud, line 164: reformat is already applied to rendered_view
```suggestion
                    rendered_view,
```

relud (approved):
r+wc

sql_generators/glean_usage/glean_app_ping_views.py
scholtzan: you're right, I'll cache the schemas to reuse them here

sean-rose:
It looks like this will initially impact the `fenix`, `firefox_ios`, and `focus_android` datasets, and we'll want to give people a [heads up](https://mozilla-hub.atlassian.net/wiki/spaces/DATA/pages/6848559/Communicating+changes) beforehand so they're not caught off guard by non-release data showing up in all those ping views.

sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -66,26 +66,105 @@ def generate_per_app(self, project_id, app_info, output_dir=None):
         ):
             return
```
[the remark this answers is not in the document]
scholtzan: In these cases we don't have multiple channels. Also we wouldn't want for the `UNION` views to overwrite the views referencing `stable` tables (which in turn are needed by the UNION views.)

sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -1,60 +1,60 @@
...
+        Generate per-app ping views across channels.
 
-        For the release channel of a glean app *only*, generate a
-        pointer view to the app-id specific view for that channel/app
-        combination
+        If schemas are incompatible, then use release channel only.
         """
+
+        # get release channel info
         release_app = app_info[0]
```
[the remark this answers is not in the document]
scholtzan: I think we can wait until a problem arises here. For now it shouldn't be a problem if there are multiple release channels, we are merging everything and should be getting all fields from all channels.

sql_generators/derived_view_schemas/__init__.py
```diff
@@ -38,9 +38,17 @@ def _get_reference_dir_path(view_dir):
...
+
+
+        try:
+            view_references = extract_table_references(view_file.read_text())
+            if len(view_references) != 1:
+                return
+        except Exception as e:
+            print(f"extract from: {view_file}") 
+            print(view_file.read_text())
+            print(e)
```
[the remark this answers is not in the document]
scholtzan: I'll revert this, shouldn't be necessary

sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -66,52 +66,153 @@ def generate_per_app(self, project_id, app_info, output_dir=None):
...
+                            # unnest repeated record
+                            select_expr.append(f"""
+                                ARRAY_AGG(
+                                    (   SELECT
+                                        STRUCT(
+                                            {self._generate_select_expression(node['fields'], app_schema_nodes[node_name]['fields'], path + [node_name])}
+                                        )
+                                        FROM UNNEST({node_name}) AS {node_name}
+                                    )
+                                ) AS {node_name}
```
sean-rose, line 203: - With the aggregate function not in a subquery the overall query will complain about the other columns not being grouped or aggregated.
- The identifier within `UNNEST()` needs to be fully qualified.
- The inner select expressions need to be qualified with the `UNNEST()` alias rather than the full path.

```suggestion
                                (
                                    SELECT ARRAY_AGG(
                                        STRUCT(
                                            {self._generate_select_expression(node['fields'], app_schema_nodes[node_name]['fields'], [node_name])}
                                        )
                                    )
                                    FROM UNNEST({'.'.join(path + [node_name])}) AS {node_name}
                                ) AS {node_name}
```

scholtzan: good catch, I'll push a fix

PR #3546, opened 2023-01-31
Change `bqetl format` to improve readability of `CASE` statements

sean-rose:
* Indent `WHEN` and `ELSE` clauses one level more than `CASE`.
* Indent `THEN` clauses one level more than the corresponding `WHEN` clause.
* Have the content of `WHEN`, `THEN`, and `ELSE` clauses start on the same line as the clause keyword.
  * This is a divergence from our existing SQL style conventions where most structural keywords are on their own line, but I feel it's worth it to have more compact `CASE` statements that read more naturally (e.g. `WHEN ...` reads like a sentence).
  * If we were relying on people to manually format SQL I might have been concerned about folks having trouble getting the "hanging indent" for multi-line `WHEN`/`THEN`/`ELSE` values correct, but since `bqetl format` applies the correct formatting for us I don't think that's much of a concern.  Also, it appears that most such values are single-line (e.g. ~80% of `WHEN` values in this repo are single-line).
* Allow an alias, comma, or dot right after a `CASE` statement's `END`.

A basic example:
```sql
  -- Before (9 lines)
  CASE
  WHEN
    ...
  THEN
    ...
  ELSE
    ...
  END
  AS ...

  -- After (5 lines)
  CASE
    WHEN ...
      THEN ...
    ELSE ...
  END AS ...
```

CC @relud since it looks like you wrote most of the existing SQL formatter code.

---
Checklist for reviewer:


bigquery_etl/format_sql/formatter.py
```diff
@@ -281,7 +295,9 @@ def inline_block_format(tokens, max_line_length=100):
                 line_length += line.inline_length
                 if line_length > max_line_length:
                     break
-                if line.indent_level <= indent_level:
+                if line.starts_with_closing_bracket:
+                    open_brackets -= 1
+                if open_brackets == 0:
```
sean-rose, line 300: This change was necessary because the "hanging indent" for `WHEN`/`THEN`/`ELSE` values means we can't rely on the indent level to determine if we're done with a bracket block.

relud (approved):

PR #3553, opened 2023-02-01
Cast submission_date in ctxsvc check

scholtzan:
fixes the Airflow error


dags/bqetl_ctxsvc_derived.py
```diff
@@ -95,6 +95,7 @@
         ],
         date_partition_parameter="submission_date",
         depends_on_past=False,
+        parameters=["submission_date:DATE:{{ds}}"],
```
quiiver, line 98: not sure if it matters but it looks like `parameters` is supposed to be a tuple.

quiiver (approved):
thanks, just one comment on the type for the parameters argument otherwise looks good

PR #3559, opened 2023-02-02
Fix Glean Usage view generation for repeated nested fields

scholtzan:
Fix for https://github.com/mozilla/bigquery-etl/pull/3545#discussion_r1093854094

### DIFF
--- a/sql_generators/glean_usage/glean_app_ping_views.py
+++ b/sql_generators/glean_usage/glean_app_ping_views.py
@@ -193,13 +193,13 @@ class GleanAppPingViews(GleanTable):
                         if node.get("mode", None) == "REPEATED":
                             # unnest repeated record
                             select_expr.append(f"""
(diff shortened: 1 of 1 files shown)


relud (approved):

sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -193,13 +193,13 @@ def _generate_select_expression(
                         if node.get("mode", None) == "REPEATED":
                             # unnest repeated record
                             select_expr.append(f"""
-                                ARRAY_AGG(
-                                    (   SELECT
+                                (
+                                    SELECT ARRAY_AGG(
```
relud, line 197: this works, but so does `ARRAY(SELECT STRUCT(...) ...)`, which i think might have been what you were going for before

sean-rose (approved):

PR #3565, opened 2023-02-03
Add --use_cloud_function option when generating SQL queries

scholtzan:
For https://github.com/mozilla/telemetry-airflow/pull/1639

`fenix.topsites_impression` got removed because it was using the dry-run function

### DIFF
--- a/sql_generators/glean_usage/glean_app_ping_views.py
+++ b/sql_generators/glean_usage/glean_app_ping_views.py
@@ -44,7 +44,9 @@ class GleanAppPingViews(GleanTable):
         self.per_app_id_enabled = False
         self.per_app_enabled = True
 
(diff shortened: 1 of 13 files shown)


sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -202,31 +216,42 @@ def _generate_select_expression(
                                     )
                                     FROM UNNEST({'.'.join(path + [node_name])}) AS {node_name}
                                 ) AS {node_name}
-                            """)
+                            """
```
sean-rose, line 219: Is it Black making these formatting changes?  I'd think the closing quotes should be aligned at the same level of indentation as the opening quotes.
```suggestion
                                """
```
(ditto for the similar multi-line strings later in the file)

sean-rose (approved):

PR #3567, opened 2023-02-06
[Bug 1804256] Update views for moving org_mozilla_firefox_vpn to mozilla_vpn

scholtzan:
Depends on https://bugzilla.mozilla.org/show_bug.cgi?id=1804256

To unblock enabling `org.mozilla.ios.FirefoxVPN`

`events_daily` are generated and need to be updated separately once `org.mozilla.ios.FirefoxVPN` is actually available (to add it to the templates)


sql/moz-fx-data-shared-prod/mozilla_vpn/events_unnested/view.sql
```diff
@@ -58,64 +58,6 @@ SELECT
...
-    STRUCT(
-      ping_info.end_time,
-      ping_info.experiments,
-      ping_info.ping_type,
-      ping_info.seq,
-      ping_info.start_time,
-      ping_info.reason,
-      ping_info.parsed_start_time,
-      ping_info.parsed_end_time
-    ) AS ping_info
```
sean-rose, line 112: The copy of this code reordering subfields above could also be removed, as there will be only one source view now (which will itself be reordering subfields as necessary for unioning already).

sean-rose (approved):
r+wc

PR #3568, opened 2023-02-07
Fix: removed explicitly specying referenced tables in metadata.yaml for fxa_user_services_device queries

kik-kik:


akkomar (approved):
r+, please fix typos in the message (for example: `Remove explicitly specified referenced tables for fxa_users_services_devices_devices_ queries`).

PR #3574, opened 2023-02-08
Prevent glean_usage generation if referenced dataset doesn't exist

scholtzan:
I could reproduce the issue described in https://mozilla.slack.com/archives/C01E8GDG80N/p1675877827569069 and earlier with`treeherder`. The generation logic checks if referenced tables in generated queries exist and aborts if they don't. However, if a dataset that is referenced does not exist dryrun returns a different error code which was considered as "all references exist" (even though they didn't). This resulted in some tables and the `dataset_metadata.yaml` getting created, which resulted in the dataset actually being deployed. So in the next run the dryrun did get the correct error code for missing references and deleted the generated datasets again.

This should fix this issue.

cc @whd

### DIFF
--- a/sql_generators/glean_usage/common.py
+++ b/sql_generators/glean_usage/common.py
@@ -104,7 +104,14 @@ def table_names_from_baseline(baseline_table, include_project_id=True):
 def referenced_table_exists(view_sql):
     """Dry run the given view SQL to see if its referent exists."""
     dryrun = DryRun("foo/bar/view.sql", content=view_sql)
-    return 404 not in [e.get("code") for e in dryrun.errors()]
+    # 403 is returned if referenced dataset doesn't exist; we need to check that the 403 is due to dataset not existing
+    # since dryruns on views will also return 403 due to the table CREATE
+    # 404 is returned if referenced table or view doesn't exist
+    return not any([
+        404 == e.get("code") or 
+        (403 == e.get("code") and "does not exist in location" in e.get("message")) 
+        for e in dryrun.errors()
+    ])
 
 
 def _contains_glob(patterns):
--- a/sql_generators/glean_usage/common.py
+++ b/sql_generators/glean_usage/common.py
@@ -108,8 +108,8 @@ def referenced_table_exists(view_sql):
     # since dryruns on views will also return 403 due to the table CREATE
     # 404 is returned if referenced table or view doesn't exist
     return not any([
-        404 == e.get("code") or 
-        (403 == e.get("code") and "does not exist in location" in e.get("message")) 
+        404 == e.get("code")
+        or (403 == e.get("code") and "bigquery.tables.create denied" not in e.get("message")) 
         for e in dryrun.errors()
     ])
 


relud (approved):
r+wc

sql_generators/glean_usage/common.py
```diff
@@ -104,7 +104,11 @@ def table_names_from_baseline(baseline_table, include_project_id=True):
 def referenced_table_exists(view_sql):
     """Dry run the given view SQL to see if its referent exists."""
     dryrun = DryRun("foo/bar/view.sql", content=view_sql)
-    return 404 not in [e.get("code") for e in dryrun.errors()]
+    # 403 is returned if referenced dataset doesn't exist
+    # 404 is returned if referenced table or view doesn't exist
+    return 404 not in [e.get("code") for e in dryrun.errors()] and 403 not in [
+        e.get("code") for e in dryrun.errors()
+    ]
```
relud, line 111: rather than iterate over errors twice, this seems like a good place for `any`:
```suggestion
    return not any(e.get("code") in {403, 404} for e in dryrun.errors())
```

sql_generators/glean_usage/common.py
```diff
@@ -104,7 +104,11 @@ def table_names_from_baseline(baseline_table, include_project_id=True):
 def referenced_table_exists(view_sql):
     """Dry run the given view SQL to see if its referent exists."""
     dryrun = DryRun("foo/bar/view.sql", content=view_sql)
-    return 404 not in [e.get("code") for e in dryrun.errors()]
+    # 403 is returned if referenced dataset doesn't exist
```
sean-rose, line 107: I'd think 403 would also be returned if whatever account is doing the dry-running doesn't have permission to access a referenced table/view, but in that scenario I guess we also wouldn't want to generate Glean usage views.

sean-rose (approved):

sean-rose:
Wait, what's up with the failed tests saying some generated Fenix and Firefox iOS ETLs can't be found?
```
FAILED tests/sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/baseline_clients_daily_v1/test_aggregation::SQL - FileNotFoundError: [Errno 2] No such file or directory: '/root/project//sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/baseline_clients_daily_v1/query.sql'
FAILED tests/sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/baseline_clients_daily_v1/test_init::SQL - FileNotFoundError: [Errno 2] No such file or directory: '/root/project//sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/baseline_clients_daily_v1/init.sql'
FAILED tests/sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/baseline_clients_first_seen_v1/test_aggregation::SQL - FileNotFoundError: [Errno 2] No such file or directory: '/root/project//sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/baseline_clients_first_seen_v1/query.sql'
FAILED tests/sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/baseline_clients_first_seen_v1/test_init::SQL - FileNotFoundError: [Errno 2] No such file or directory: '/root/project//sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/baseline_clients_first_seen_v1/init.sql'
FAILED tests/sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox_derived/baseline_clients_first_seen_v1/test_aggregation::SQL - FileNotFoundError: [Errno 2] No such file or directory: '/root/project//sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox_derived/baseline_clients_first_seen_v1/query.sql'
FAILED tests/sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox_derived/baseline_clients_first_seen_v1/test_init::SQL - FileNotFoundError: [Errno 2] No such file or directory: '/root/project//sql/moz-fx-data-shared-prod/org_mozilla_ios_firefox_derived/baseline_clients_first_seen_v1/init.sql'
```
Are those not being generated due to 403 permission errors when dry-running?

sean-rose:
There are 46 `403` errors in the [generate SQL CI step](https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/18860/workflows/6eecd663-5b95-41d7-b296-1e8e727dccbf/jobs/179741/parallel-runs/0/steps/0-106).

sean-rose:
I'd be very happy if we could get rid of the `CREATE VIEW ... AS` from our `view.sql` files, as it seems like the tooling should be able to automatically add that just-in-time when deploying the views (and it would eliminate the possibility of the hard-coded view name not matching the file path).

scholtzan:
Agreed, though that change would probably require a proposal and opportunity for people to weigh in.

scholtzan:
For now I'd just check the error message which are different depending on whether the dataset doesn't exist or a `CREATE` statement is being run:

```
Permission bigquery.tables.create denied on dataset moz-fx-data-shared-prod:mozphab (or it may not exist)
```
vs
```
User does not have permission to query table moz-fx-data-shared-prod:msadf.metrics, or perhaps it does not exist in location US.
```

sql_generators/glean_usage/common.py
```diff
@@ -104,7 +104,14 @@ def table_names_from_baseline(baseline_table, include_project_id=True):
...
     dryrun = DryRun("foo/bar/view.sql", content=view_sql)
-    return 404 not in [e.get("code") for e in dryrun.errors()]
+    # 403 is returned if referenced dataset doesn't exist; we need to check that the 403 is due to dataset not existing
+    # since dryruns on views will also return 403 due to the table CREATE
+    # 404 is returned if referenced table or view doesn't exist
+    return not any([
+        404 == e.get("code") or 
+        (403 == e.get("code") and "does not exist in location" in e.get("message")) 
+        for e in dryrun.errors()
+    ])
```
sean-rose, line 114: - Relying on specific error messages seems a bit tenuous either way (e.g. they could arbitrarily change "does not" to "doesn't"), but I think excluding the unwanted error would be slightly safer.
- IMO boolean operators at the beginning of lines are better for code readability.
```suggestion
    return not any([
        404 == e.get("code")
        or (403 == e.get("code") and "bigquery.tables.create denied" not in e.get("message")) 
        for e in dryrun.errors()
    ])
```

PR #3581, opened 2023-02-09
Sidefill newtab.opened and filter invalid events in newtab_interactions

ANich:
This addresses two problems that led to over-counting visits in some cases:
1. Some valid interactions are not tagged with their newtab_open_source. This led to some visits having one interaction with a null `visit_id` and another with the actual `visit_id`
2. Some visits come from tab-preloading. We identify these by the fact that they do not have a `newtab.opened` event which has `newtab_open_source` and `newtab_visit_started_at` as extras OR any interactions. 

As a side note, we remove interactions without a `visit_id`, these won't be valid in any case.

See analysis of difference [here](https://sql.telemetry.mozilla.org/queries/90564/source) and broader context [here](https://docs.google.com/document/d/1chY8z6noN5CF3k8YX2TPThLRF3bTGuUR93zUZR8nX1Q/edit#) and [here](https://docs.google.com/document/d/1ShDAY-PfOakCwA00t_pvOwIYDJpYOifVEIPMTM30-a4/edit#).

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/newtab_interactions_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/newtab_interactions_v1/query.sql
@@ -244,6 +244,6 @@ FROM
   side_filled
 WHERE
    -- If we're not able to attach newtab_open_source and newtab_visit_started_at,
-   -- we haven't received a valid newtab.opened event:
+   -- we haven't received a valid newtab.opened event within the partition (submission date):
   newtab_open_source IS NOT NULL
   AND newtab_visit_started_at IS NOT NULL
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/newtab_interactions_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/newtab_interactions_v1/query.sql
@@ -184,6 +184,8 @@ aggregated_newtab_activity AS (
     COUNTIF(is_organic_pocket_save) AS organic_pocket_saves,
   FROM
     categorized_events
+  WHERE
+    newtab_visit_id IS NOT NULL
   GROUP BY
     newtab_visit_id,
     client_id,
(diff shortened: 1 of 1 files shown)


sql/moz-fx-data-shared-prod/telemetry_derived/newtab_interactions_v1/query.sql
```diff
@@ -206,12 +208,42 @@ client_profile_info AS (
...
 FROM
-  aggregated_newtab_activity
-LEFT JOIN
-  client_profile_info
-USING
-  (legacy_telemetry_client_id)
+  side_filled
+WHERE
+   -- If we're not able to attach newtab_open_source and newtab_visit_started_at,
+   -- we haven't received a valid newtab.opened event:
```
chutten, line 247:
```suggestion
   -- we haven't received a valid newtab.opened event within the partition (submission date):
```

chutten (commented):
Can't speak to its use (if this were a lower-level dataset I'd want the nullish stuff left in so I could filter them out myself or choose to broaden my query and coalesce from adjacent days if it matters) or correctness (though it does look fine to my eyes), but I was only asked to look at the descriptions of the input which appear correct.

Perhaps a good first thing to do would be to check how much data this loses for the various reasons. How much Bad Data are we saving people from having to deal with? How much Good But Ill Timed Data (ie, data for which there does exist a newtab.opened event but not within the partition) does it cost?

sql/moz-fx-data-shared-prod/telemetry_derived/newtab_interactions_v1/query.sql
```diff
@@ -201,17 +203,58 @@ client_profile_info AS (
...
+      ORDER BY
+        submission_date ASC
+    ) AS newtab_visit_ended_at,
+    LOGICAL_OR(
+      searches > 0
+      OR tagged_search_ad_clicks > 0
+      OR tagged_search_ad_impressions > 0
+      OR follow_on_search_ad_clicks > 0
+      OR follow_on_search_ad_impressions > 0
+      OR topsite_impressions > 0
```
irrationalagent, line 240: if we ever add more types of interactions, this list will have to be updated. that's probably OK. in theory another alternative would be to create a catch-all "any_interaction" higher up in the logic, looking for any event_name of `impression`, `click` or `issued`. but that might also be too liberal, if future events have that event_name but we don't want (for some reason) to count it as an interaction. so i guess im just calling it out for now as something we need to be aware of.

sql/moz-fx-data-shared-prod/telemetry_derived/newtab_interactions_v1/query.sql
```diff
@@ -201,17 +203,58 @@ client_profile_info AS (
...
+  * EXCEPT (visit_had_any_interaction)
 FROM
-  aggregated_newtab_activity
-LEFT JOIN
-  client_profile_info
-USING
-  (legacy_telemetry_client_id)
+  side_filled
+WHERE
+   -- Keep only rows with interactions, unless we receive a valid newtab.opened event.
```
irrationalagent, line 258: is it worth adding a comment here about what we specifically expect to be dropped? "we expect this to only drop newtab.closed events from newtab_visit_ids that don't have an non-null open source on the same partition" or something?

sql/moz-fx-data-shared-prod/telemetry_derived/newtab_interactions_v1/query.sql
ANich: Good call, I'll add a comment to reflect that

sql/moz-fx-data-shared-prod/telemetry_derived/newtab_interactions_v1/query.sql
ANich: Yup, I'll clarify that here as well

PR #3587, opened 2023-02-10
Deploy changes to stage env

scholtzan:
Implements https://docs.google.com/document/d/1BAGCjohkNPOqcrq1RrkkEtYdZyP6CoBEoeVlS4jWK4w/edit#

Depends on https://mozilla-hub.atlassian.net/browse/DSRE-1196

This change will result in the CI pushing schemas for new or updated tables, views and UDFs to a stage environment (`bigquery-etl-integration-test`). These artifacts will be published in temporary datasets. All references in SQL will be replaced with their new stage location. This way all checks (dryruns, view validation etc) will use the updated schemas and definitions. This eliminates the need to deploy changes to prod manually to get CI checks to run successfully.


bigquery_etl/cli/stage.py
```diff
@@ -0,0 +1,367 @@
...
+
+    dataset = client.update_dataset(dataset, ["default_table_expiration_ms", "labels"])
+
+
+@stage.command(
+    help="""
+        Remove deployed artifacts from stage environment
+
+    Examples:
+    ./bqetl deploy clean
```
akkomar, line 327:
```suggestion
    ./bqetl stage clean
```

bigquery_etl/cli/stage.py
```diff
@@ -0,0 +1,367 @@
...
+@sql_dir_option
+@click.option(
+    "--dataset-suffix",
+    "--dataset_suffix",
+    help="Suffix appended to the deployed dataset",
+)
+@click.option(
+    "--update-references",
+    "--update_references",
+    default=True,
```
akkomar, line 67:
```suggestion
    default=True,
    is_flag=True,
```

bigquery_etl/cli/stage.py
```diff
@@ -0,0 +1,367 @@
...
+        if test_path.exists():
+            shutil.copytree(
+                test_path, TEST_DIR / project_id / dataset / name, dirs_exist_ok=True
+            )
+            shutil.rmtree(test_path)
+
+    # remove artifacts from the "prod" folders
+    for artifact_file in artifact_files:
+        if artifact_file.parent.exists():
+            shutil.rmtree(artifact_file.parent)
```
akkomar, line 139: Does this remove stuff from `sql/` directory if `--copy_sql_to_tmp_dir` is not set?
If so then I'd be surprised by that if I ran this locally.

.circleci/config.yml
```diff
@@ -132,8 +139,7 @@ jobs:
                 "commit message"
               PATHS=sql
             else
-              echo "Check dry run for modified queries"
-              PATHS="$(git diff origin/main... --name-only --diff-filter=d -- sql)"
+              PATHS="sql/bigquery-etl-integration-test"
```
akkomar, line 142: It would probably take me a while to figure out where does `sql/bigquery-etl-integration-test` come from if I was reading this couple of months from now. Maybe it's worth adding a comment that this path got created in `bqetl stage deploy`?

akkomar (approved):
r+wc

bigquery_etl/cli/stage.py
scholtzan: Yes, it would. I guess we could have a flag to control this. However, if these files do not get deleted then all tests run on them (dryrun, view validation, etc) are likely going to fail since the referenced artifacts don't actually exist or the folder structure doesn't conform to the dataset and project names of the artifacts that got deployed to stage.

PR #3592, opened 2023-02-14
[Bug 1816621] Generate nested type information for unioned Glean ping views

scholtzan:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1816621

Casting nested records to NULL requires the exact type information. That information needs to be generated based on the unioned schema

### DIFF
--- a/sql_generators/glean_usage/glean_app_ping_views.py
+++ b/sql_generators/glean_usage/glean_app_ping_views.py
@@ -237,22 +237,24 @@ class GleanAppPingViews(GleanTable):
                                 f"SAFE_CAST(NULL AS {dtype}) AS {node_name}"
                             )
             else:
-                if dtype == "RECORD":
-                    # unwrap missing struct - workaround to prevent type incompatibilities; NULL is always INT in STRUCT
-                    select_expr.append(
-                        f"""
-                            STRUCT(
-                                {self._generate_select_expression(node['fields'], {}, path + [node_name])}
-                            ) AS {node_name}
-                        """
-                    )
-                else:
-                    # field doesn't exist in app schema, set to NULL
-                    if node.get("mode", None) == "REPEATED":
-                        select_expr.append(
-                            f"SAFE_CAST(NULL AS ARRAY<{dtype}>) AS {node_name}"
-                        )
-                    else:
-                        select_expr.append(f"SAFE_CAST(NULL AS {dtype}) AS {node_name}")
+                select_expr.append(
+                    f"CAST(NULL AS {self._type_info(node)}) AS {node_name}"
+                )
 
         return ", ".join(select_expr)
+
+    def _type_info(self, node):
+        """Determine the type information"""
+        dtype = node["type"]
+        if dtype == "RECORD":
+            dtype = (
+                "STRUCT<"
+                + ", ".join(
+                    f"{field['name']} {self._type_info(field)}"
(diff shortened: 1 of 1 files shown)


sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -239,13 +239,18 @@ def _generate_select_expression(
             else:
                 if dtype == "RECORD":
                     # unwrap missing struct - workaround to prevent type incompatibilities; NULL is always INT in STRUCT
```
relud, line 241: This entire conditional structure can be replaced with:
```python
                select_expr.append(
                    f"CAST(NULL AS {self._type_info(node)}) AS {node_name}"
                )
```

sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -256,3 +261,18 @@ def _generate_select_expression(
...
+            dtype = node["type"]
+            if dtype == "RECORD":
+                if node.get("mode", None) == "REPEATED":
+                    type_expr.append(f"ARRAY<STRUCT<{self._type_info(node['fields'])}>>")
+                else:
+                    type_expr.append(f"STRUCT<{self._type_info(node['fields'])}>")
+            else:
+                type_expr.append(dtype)
+
+        return ", ".join(type_expr)
```
relud, line 278: this will produce results like STRUCT<STRING, INT64>, which bigquery allows, but it makes the fields anonymous, which is undesirable. UNION ALL uses the field names from the first query for all subsequent queries, so if this happens in the first query it will anonymize field names from subsequent queries.

Instead fields should be named, and that combined with the above comment makes it so that this should probably accept a single node.

Also, mode needs to be checked regardless of whether the node is a record.

```suggestion
    def _type_info(self, node):
        """Determine the type information"""
        dtype = node["type"]
        if dtype == "RECORD":
            dtype = (
                "STRUCT<"
                + ", ".join(
                    f"{field['name']} {self._type_info(field)}"
                    for field in node["fields"]
                )
                + ">"
            )
        if node.get("mode", None) == "REPEATED":
            return f"ARRAY<{dtype}>"
        return dtype
```

relud (approved):
r+wc

sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -256,3 +261,18 @@ def _generate_select_expression(
...
+        type_expr = []
+        for node in nodes:
+            dtype = node["type"]
+            if dtype == "RECORD":
+                if node.get("mode", None) == "REPEATED":
+                    type_expr.append(f"ARRAY<STRUCT<{self._type_info(node['fields'])}>>")
+                else:
+                    type_expr.append(f"STRUCT<{self._type_info(node['fields'])}>")
+            else:
+                type_expr.append(dtype)
```
sean-rose, line 276: Need to handle possible arrays in this logic branch as well.
```suggestion
                if node.get("mode", None) == "REPEATED":
                    type_expr.append(f"ARRAY<{dtype}>")
                else:
                    type_expr.append(dtype)
```

sql_generators/glean_usage/glean_app_ping_views.py
sean-rose: This incorporates/supersedes my suggestions, so I'll mark mine as resolved.

PR #3598, opened 2023-02-20
feat(MP-267): Move popularities export to Airflow

kik-kik:
# feat(MP-267): Move popularities export to Airflow

Job to support a data export for MDN.


kik-kik:
@akkomar atm I have both query.py and query.sql approaches here. Please let me know which way you think we should take. Here's what the telemetry-airflow DAG would look like:
https://github.com/mozilla/telemetry-airflow/compare/main...Guyzeroth:telemetry-airflow:mp-267-move-popularities-generation-to-airflow

akkomar (commented):
As discussed on Slack:
I think you can use [this script](https://github.com/mozilla/bigquery-etl/pull/3598/files#diff-b514745028e25bbbb3d1795aa61b5d722652ef427bea48dcf28315c994ca6206R1-R87) and schedule it [like this](https://github.com/mozilla/bigquery-etl/blob/7b00dd676df06147009050e27d2ea45eba996abf/dags/bqetl_monitoring.py#L52).

dags.yaml
```diff
@@ -978,3 +978,25 @@ bqetl_fivetran_costs:
...
+  default_args:
+    depends_on_past: false
+    email:
+      - telemetry-alerts@mozilla.com
+      - mdn@mozilla.com
+      - fmerz@mozilla.com
+    email_on_failure: true
+    email_on_retry: false
+    end_date: null
+    owner: fmerz@mozilla.com
```
akkomar, line 992: We should have someone from Data Team to either own this or be on the `email` list as support.

sql/moz-fx-data-shared-prod/mdn_yari_derived/dataset_metadata.yaml
```diff
@@ -0,0 +1,12 @@
+friendly_name: MDN Yari Dataset
```
akkomar, line 1: Why is dataset metadata needed here? Can we omit it to avoid creation of an empty dataset?

sql/moz-fx-data-shared-prod/mdn_yari_derived/mdn_popularities_v1/query.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env python3
+
+"""Monthly data exports of MDN 'Popularities'. This aggregates and counts total page visits and normalizes them agains the max."""
+import logging
+from argparse import ArgumentParser
+from datetime import datetime
+
+from google.cloud import bigquery, storage
+
+QUERY_TEMPLATE = """\
```
akkomar, line 10: Have you tested this? I think this query might fail because:
1. It's not filtering over `submission_date`
2. `@submission_date` parameter is not passed in (see https://github.com/mozilla/bigquery-etl/blob/7b00dd676df06147009050e27d2ea45eba996abf/script/glam/run_scalar_agg_clustered_query.py#L70)

sql/moz-fx-data-shared-prod/mdn_yari_derived/mdn_popularities_v1/query.py
```diff
@@ -0,0 +1,117 @@
...
+    temp_blob.download_to_filename(local_path)
+
+    # upload to the target bucket
+    mdn_storage_client = storage.Client(args.destination_project)
+    mdn_bucket = mdn_storage_client.get_bucket(args.destination_bucket)
+    mdn_uri = f"{args.destination_path}/{target_file_name}.json"
+
+    logging.info("Uploading: %s to %s" % (local_path, mdn_uri))
+    blob = mdn_bucket.blob(mdn_uri)
+    blob.upload_from_filename(local_path)
```
akkomar, line 109: Why don't we write directly to the destination bucket?

akkomar (changes requested):
Does the Airflow service account have write access to `mdn-dev` bucket? It would need to since there's no custom connection configuration in the DAG.

More comments inline.

sql/moz-fx-data-shared-prod/mdn_yari_derived/mdn_popularities_v1/query.py
kik-kik: It's because we want the file to end up in a bucket within a different GCP project. I don't believe we can export the table directly from one project to a bucket inside another project.

dags.yaml
kik-kik: Added myself to email list.

sql/moz-fx-data-shared-prod/mdn_yari_derived/dataset_metadata.yaml
kik-kik: Removed the dataset_metadata.yaml file.

sql/moz-fx-data-shared-prod/mdn_yari_derived/mdn_popularities_v1/query.py
kik-kik: Thanks, I just noticed I was using both @submission_date and date placeholder. Updated to use the suggested approach.

sql/moz-fx-data-shared-prod/mdn_yari_derived/mdn_popularities_v1/query.py
kik-kik: At least I was not able to find a nice way to do this. Please let me know if you know a way to handle this better.

akkomar: I think you can extract directly to a bucket in another project. Compare the GLAM DAG where we export to a bucket in `moz-fx-data-glam-prod-fca7`: https://github.com/mozilla/telemetry-airflow/blob/f5977b176aa3ca903efd83c751a6431b0554a96f/dags/glam_subdags/extract.py#L144

kik-kik: ok turns out you can write to different projects, based on the docs it was not obvious to me and Google wasn't giving me any useful results. Made the change and removed the unnecessary middle steps.

akkomar (approved):
r+, as discussed on Slack we'll ask SRE to give Airflow SA write access to destination bucket.

PR #3600, opened 2023-02-21
chore(ci): use unit tests from telemetry-airflow

mikaeld:
# Description
Update CircleCI config to use unit tests from `telemetry-airflow` instead of running docker-compose. This significantly shortens CircleCI job `validate-dags` runtime from 5-6 minutes down to around 2 minutes.

Current implementation requires the job's executor to match `telemetry-airflow`'s python version. There's is a fairly low risk we encounter issues when updating `telemetry-airflow` to a new python version if we don't update the executor version in this CI config **_IF_** _there are breaking changes in key packages_. 

A possible alternative is to pull the latest `telemetry-airflow` image, mount the generated DAGs as a volume and run a pytest command. This will probably be the long-term solution as we would not need to pull a git repo and install python dependencies. I would probably switch to this solution after the related PRs are merged.

# Related PR
- https://github.com/mozilla/telemetry-airflow/pull/1638
- https://github.com/mozilla-services/cloudops-infra/pull/4705

Checklist for reviewer:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -250,7 +250,7 @@ jobs:
       - run:
           name: Replace telemetry-airflow DAGs with BigQuery ETL DAGs
           command: |
-            rm telemetry-airflow/dags/*
+            rm telemetry-airflow/dags/* -f || true
             cp -a dags/. telemetry-airflow/dags/
       - *attach_generated_sql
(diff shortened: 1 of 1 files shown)


scholtzan (approved):
Wow, the check is running so much faster! 🎉

PR #3601, opened 2023-02-21
Support glean apps with additional deletion requests table in shredder

relud:


bigquery_etl/shredder/config.py
```diff
@@ -406,19 +407,48 @@ def find_glean_targets(pool, client, project=SHARED_PROD):
...
-            re.sub("_stable$", "_derived", table.dataset_id),
-        ]
-        if dataset_id in datasets
-    }
+    sources = defaultdict(default_factory=list)
+    for table in glean_stable_tables:
+        if table.table_id.startswith(source_doctype):
+            source = DeleteSource(qualified_table_id(table), GLEAN_CLIENT_ID, project)
+            derived_dataset = re.sub("_stable$", "_derived", table.dataset_id)
+            # append to list to use every version of deletion request tables
```
relud, line 415: in practice, we only have one version, so this shouldn't impact behavior

tests/shredder/test_config.py
```diff
@@ -0,0 +1,111 @@
...
+        return [
+            bigquery.DatasetReference(project, f"{app}_{suffix}")
+            for app in ["fenix", "focus_android"]
+            for suffix in ["stable", "derived"]
+        ]
+
+    def list_tables(self, dataset_ref):
+        labels = {}
+        if dataset_ref.dataset_id.endswith("stable"):
+            table_ids = ["metrics_v1", "deletion_request_v1"]
```
sean-rose, line 19: For better test coverage it'd be good to have a table that starts with `migration` and verify it's excluded (i.e. no change to expected output).

tests/shredder/test_config.py
```diff
@@ -0,0 +1,111 @@
...
+            table.schema = [
+                bigquery.SchemaField(
+                    "client_info",
+                    "RECORD",
+                    "NULLABLE",
+                    [bigquery.SchemaField("client_id", "STRING")],
+                )
+            ]
+        elif table_ref.dataset_id.endswith("derived"):
+            table.schema = [bigquery.SchemaField("client_id", "STRING")]
```
sean-rose, line 51: For better test coverage it'd be good to also have a table that doesn't have a `client_id` column and verify it's excluded (i.e. no change to expected output).

sean-rose (approved):
r+wc

PR #3602, opened 2023-02-22
bug: added a debug message to entrypoint

kik-kik:
# bug: added a debug message to entrypoint

### DIFF
--- a/script/entrypoint
+++ b/script/entrypoint
@@ -4,6 +4,7 @@ set -e
 
 # Use credentials from environment
 if [ -n "$GCLOUD_SERVICE_KEY" ]; then
+    echo "DEBUG: SETTING UP CREDS BLOCK HIT"  # TODO: once the issue is solved this can be removed.
     # Google's client libraries will check for GOOGLE_APPLICATION_CREDENTIALS
     # and use a file in that location for credentials if present;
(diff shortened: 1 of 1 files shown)


mikaeld:
First step is knowing if the JSON creds are empty or simply malformed. The issue likely happens under the `if` statement starting at line 15. I would be surprised if the issue is with the env var not being present, i.e. where you added the echo command.

I would probably switch this:
```shell
gcloud --quiet auth activate-service-account --key-file "$GOOGLE_APPLICATION_CREDENTIALS"
```
to add a verbosity flag like this:
```shell
gcloud --verbosity=debug --quiet auth activate-service-account --key-file "$GOOGLE_APPLICATION_CREDENTIALS"
```

sean-rose:
> to add a verbosity flag like this:
> 
> ```shell
> gcloud --verbosity=debug --quiet auth activate-service-account --key-file "$GOOGLE_APPLICATION_CREDENTIALS"
> ```

When specifying both `--verbosity` and `--quiet` like that do we know that `--verbosity` would take precedence over `--quiet`?  If not, I'd suggest removing `--quiet`.

kik-kik:
@mikaeld should we merge this as a first attempt to hopefully get more info about the problem?

PR #3605, opened 2023-02-22
Fix deploy dataset naming

scholtzan:


bigquery_etl/cli/stage.py
```diff
@@ -317,9 +317,6 @@ def _deploy_artifacts(ctx, artifact_files, project_id, dataset_suffix, sql_dir):
 
 def create_dataset_if_not_exists(project_id, dataset, suffix=None):
     """Create a temporary dataset if not already exists."""
-    if suffix:
-        dataset += f"_{suffix}"
-
```
scholtzan, line 322: dataset already has suffixed appended

wwyc (approved):

PR #3606, opened 2023-02-23
Fix shredder queries skipping additional deletion requests tables

relud:


bigquery_etl/shredder/config.py
```diff
@@ -466,7 +470,11 @@ def find_glean_targets(
         },
         **{
             # glean derived tables that contain client_id
-            client_id_target(table=qualified_table_id(table)): sources[table.dataset_id]
+            DeleteTarget(
+                table=qualified_table_id(table),
+                # field must be repeated for each deletion source
+                field=(CLIENT_ID,) * len(sources[table.dataset_id]),
```
relud, line 476: this is because field gets zipped together with sources, and zip stops at the end of the shortest iterable, so without this change only the first source was actually being used.

sean-rose (approved):

PR #3607, opened 2023-02-23
feat(DENG-610): added fxa_users_daily_v2 model

kik-kik:
# feat(DENG-610): added fxa_users_daily_v2 model

## Context:
The primary change introduced by the v2 compared to v1 is that it is now derived from a lower granularity table which contains users/service level granularity (`fxa_users_services_daily_v2`) instead of accessing the events table directly. This should reduce the amount of data we have to process as part of the FxA pipeline and reduced code duplication across different models (for example, only the initial users services table needs to filter for specific events).

### Removed fields:
- `app_version` - as pointed out by Arkadiusz, this field does not seem to make sense on the user level as it is related to the service used and as the result provides no useful meaning here. Also, it can be accessed via the upstream table `fxa_users_services_daily_v2`.
- `monitor_only` - based on the conversation under this PR it appears this field is a legacy field and is no longer used/needed.

### Follow-up PR's:
- https://github.com/mozilla/bigquery-etl/pull/3610 - first_seen_v2 added
- https://github.com/mozilla/bigquery-etl/pull/3611 - last_seen_v2 added

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_daily_v2/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_daily_v2/metadata.yaml
@@ -1,14 +1,13 @@
 ---
 friendly_name: FxA Users Daily
 description: |
-  Usage aggregations per FxA user per day. Build from fxa_users_services_daily.
-  If the user had multiple different entries on a single day, then we grab
-  the attributes from the first service observed that day.
+  Derived from fxa_users_services (user/service level) and aggregated
+  to the user level. This includes information about which users were
+  active on a specific day along with some attributes related to that user.
+  See the schema for a list of attributes and their descriptions.
 
-  This lower level model uses the following category of events:
-  event_category IN ('auth', 'auth_bounce', 'content', 'oauth')
-
-  Partitioned by submission_date and clustered by country, os_name, and user_id
+  Partitioned by submission_date,
+  clustered by country, os_name, and user_id
 owners:
   - kignasiak@mozilla.com
 labels:
--- a/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_daily_v2/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_daily_v2/metadata.yaml
@@ -1,0 +1,29 @@
+---
+friendly_name: FxA Users Daily
+description: |
+  Usage aggregations per FxA user per day. Build from fxa_users_services_daily.
+  If the user had multiple different entries on a single day, then we grab
+  the attributes from the first service observed that day.
+
+  This lower level model uses the following category of events:
+  event_category IN ('auth', 'auth_bounce', 'content', 'oauth')
+
+  Partitioned by submission_date and clustered by country, os_name, and user_id
+owners:
+  - kignasiak@mozilla.com
+labels:
+  application: fxa
+  incremental: true
+  schedule: daily
+scheduling:
+  dag_name: bqetl_fxa_events
+bigquery:
+  time_partitioning:
+    type: day
+    field: submission_date
+    require_partition_filter: true
+  clustering:
+    fields:
+      - country
+      - os_name
+      - user_id
--- a/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_daily_v2/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_daily_v2/metadata.yaml
@@ -14,8 +14,9 @@ labels:
   application: fxa
   incremental: true
   schedule: daily
-scheduling:
-  dag_name: bqetl_fxa_events
+# TODO: once we agree on the model scheduling needs to be uncommented.
+# scheduling:
+#   dag_name: bqetl_fxa_events
 bigquery:
   time_partitioning:
     type: day
--- a/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_daily_v2/query.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_daily_v2/query.sql
@@ -1,17 +1,39 @@
+WITH base AS (
+  SELECT
+    user_id,
+    country,
+    `language`,
+    os_name,
+    os_version,
+    seen_in_tier1_country,
+    registered,
+    user_service_first_daily_flow_info,
+  FROM
+    `firefox_accounts_derived.fxa_users_services_daily_v2`
+  WHERE
+    submission_date = @submission_date
+),
+bool_fields AS (
+  SELECT
+    user_id,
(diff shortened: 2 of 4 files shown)


kik-kik:
One thing we need to check is if `monitor_only` field is needed which we have in the v1:
https://github.com/mozilla/bigquery-etl/blob/636cf6bf2597c5c3da8c64167618c20a8beeb179/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_daily_v1/query.sql#L50

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_daily_v2/query.sql
```diff
@@ -0,0 +1,25 @@
+SELECT
+  submission_date,
+  user_id,
+  udf.mode_last(ARRAY_AGG(country) OVER w1) AS country,
+  udf.mode_last(ARRAY_AGG(`language`) OVER w1) AS `language`,
+  udf.mode_last(ARRAY_AGG(app_version) OVER w1) AS app_version,
+  udf.mode_last(ARRAY_AGG(os_name) OVER w1) AS os_name,
+  udf.mode_last(ARRAY_AGG(os_version) OVER w1) AS os_version,
+  TRUE IN UNNEST(ARRAY_AGG(seen_in_tier1_country) OVER w1) AS seen_in_tier1_country,
```
kik-kik, line 9: @akkomar is there a nicer way to do this?

kik-kik: broke down the query to make this more readable.

akkomar (changes requested):
This PR adds `fxa_users_daily_v2` query but the title mentions `fxa_users_services_daily_v2`. Please clearly describe what's changed here.

===

> One thing we need to check is if `monitor_only` field is needed which we have in the v1:
> 
> https://github.com/mozilla/bigquery-etl/blob/636cf6bf2597c5c3da8c64167618c20a8beeb179/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_daily_v1/query.sql#L50

Per discussion with @irrationalagent  this was a hacky way to see how many users that were not signed in to FxA started the monitor signup flow from password manager.
We suspect that this code path no longer exists. However [looking at the data in `fxa_users_daily_v1`](https://sql.telemetry.mozilla.org/queries/90667/source#224508), number of such users seems roughly unchanged since Nov 2019. Although it's relatively low, at peaks it reaches 1% of the total user count.
I think we should filter out such entries from this table to avoid potential analysis surprises. @irrationalagent  - what do you think?

===

Compared to `fxa_users_daily_v1` we don't have an `app_version` field in `_v2` added in this PR. This makes sense to me since this field is relevant only in context of a service. However the same can be said about `os_name` and `os_version`. @irrationalagent  - would you agree?

irrationalagent:
@akkomar 

1. that's some strange data. i definitely dont see that path present in the recent version of Firefox and if this were being caused by old versions i would expect to see more of a gradual downward trend. so there may be something additional going on here that im not aware of. in any case, you are right, we should maintain the filter to ignore these non-authenticated users. 
2. agreed if the intent here is a user-day level of representation. when it comes to documenting this table, we should have a clear explanation for the uses of this table vs the service-level table since the absence of OS information here could confuse some people who "should be" using the service level tables.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_daily_v2/metadata.yaml
```diff
@@ -0,0 +1,29 @@
+---
+friendly_name: FxA Users Daily
+description: |
+  Usage aggregations per FxA user per day. Build from fxa_users_services_daily.
+  If the user had multiple different entries on a single day, then we grab
+  the attributes from the first service observed that day.
+
+  This lower level model uses the following category of events:
+  event_category IN ('auth', 'auth_bounce', 'content', 'oauth')
```
akkomar, line 9: This doesn't seem to match https://github.com/mozilla/bigquery-etl/blob/7b00dd676df06147009050e27d2ea45eba996abf/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_daily_v2/query.sql#L57-L72

kik-kik: Thanks,
I just double checked and I updated the description in the metadata and removed `auth_bounce`. This is fine to do as I checked the `auth_bounce` "event_type" and the only event_type we seem to have in there is `fxa_email - bounced` which is being filtered currently out anyways. So seems safe to not include at all.

kik-kik:
Thanks both @irrationalagent and @akkomar for the conversation. I updated the description and hope it is easier to understand what this data is.

akkomar (commented):
Thanks for the updates @kik-kik!

This looks pretty good to me, I have two questions:
1. What do you think about filtering out users that previously were in `monitor_only` group that I suggested in https://github.com/mozilla/bigquery-etl/pull/3607#pullrequestreview-1315384483? I think if we don't do that we should file a bug to investigate the scenario that yields only the `event_type = 'fxa_rp - engage' AND service = 'fx-monitor'` event.
2. I'm still not sure about `os_name` and `os_version`. They're described as `OS on which the app was running. For example: Android.` and `Version of the OS the device was using.`. But what does an "app" or "device" mean in context of this table since it's aggregated on the user level? Shall we remove these fields and maybe note in the description that for analyses requiring information about services and devices, other tables should be used?

kik-kik:
Hey @akkomar,

Thanks for your thoughts. So I did a quick check.

So it's not many, but there are also some entries that match in the last 7 days with filter:
`WHERE event_type = 'fxa_rp - engage' AND service = 'fx-monitor'`

Results:
```
date	        cnt
2023-03-12	2031
2023-03-11	9842
2023-03-10	993
2023-03-09	1373
2023-03-08	2505
2023-03-07	8247
2023-03-06	67050
```

Generally speaking, my impression from the previous conversation was that this monitor_only field is no longer needed. The main question remains if we should just filter out all entries for service `fxa_monitor` from the fxa user tables. I wonder if @irrationalagent have any thoughts about this.

In regards to the fields, I agree and as we agreed during our discussion I deleted the app and app version fields from this table as those are very service specific fields and that's where they can be found. Regarding to the os fields, I'm also not sure, once again maybe @irrationalagent has an opinion here. They need to be in the daily table if we want them to be available inside the first_seen table unless we change the flow of the data a bit, but I'd be against it as I feel it starts making the lineage more complicated again. If those fields is something we want to be able to see in the first_seen table, such as what os did a user is when they used fxa for the first time I think this is ok. Of course, this information could also be easily retrieved from the users_services table by just looking at the very first record for a user.

Either way, for now I will update the description for those two fields to make it clear it corresponds to the first user entry for that day.

kik-kik:
Will merge the changes to avoid the branch diverting too much from main. For now, the query is being descheduled and once we agree on what fields should be included we can make the final tweak and schedule the query.

akkomar (approved):

PR #3610, opened 2023-02-23
feat(DENG-704): added fxa_users_first_seen_v2

kik-kik:
# feat(DENG-704): added fxa_users_first_seen_v2

depends: https://github.com/mozilla/bigquery-etl/pull/3607

---

## Context

The main difference between v2 and v1 of this table is that v2 is created using `fxa_users_daily_v2` instead of using the events table directly. This change aims to reduce code duplication and to have a cleaner and more linear flow of data which is a bit easier to reason about. Avoiding going directly to the events table also reduced the amount of data that we need to process for each run.

### Field changes:

`Removed`:
- `services_used` - this was a field which was modified on each run and aggregates a list of all services the user has ever used. This feels like a wrong place for this information as it did not include only information related to the first instance of when a user was observed making the table more difficult to reason about. Currently, this information can be retrieved from `fxa_users_services_first_seen_v2` by querying for all entries for a specific user_id. This will return us a list of all services a specific user ever accessed with the corresponding dates.

`Added`:
- `country` - shows what country the user had assigned in their first interaction
- `language` - shows what language the user had assigned in their first interaction
- `os_name` - shows what os the user had assigned in their first interaction
- `os_version` - shows what os version the user had assigned in their first interaction
- `seen_in_tier1_country` - whows if the user has an activity event in one of the tier1 countries on their first day.
- `registered` - did a user register when first observed? I believe in our case this should always be true in this table. 

These fields were added to match the fields defined inside `fxa_users_daily_v2`

tier1 countries include: `United States`, `France`, `Germany`, `United Kingdom`, `Canada`

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v2/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v2/metadata.yaml
@@ -1,0 +1,31 @@
+friendly_name: FxA Users First Seen
+description: |-
+  Contains the first entry seen inside fxa_users_daily_v2
+  for each user along with some attributes.
+
+  Partitioned by submission_date,
+  clustered by country, os_name
+owners:
+- kignasiak@mozilla.com
+labels:
+  application: firefox_accounts
+  incremental: true
+  schedule: daily
+# TODO: once we agree on the model scheduling needs to be uncommented.
+# scheduling:
+#   dag_name: bqetl_fxa_events
+#   start_date: '2021-07-09'
+#   priority: 80
+#   depends_on_past: true
+#   date_partition_parameter: null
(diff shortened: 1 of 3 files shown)


akkomar (commented):
Query LGTM, I'm not sure about dimensions - see https://github.com/mozilla/bigquery-etl/pull/3611#pullrequestreview-1315475945

kik-kik:
@akkomar thanks for the comments. I see your point regarding the registered field. I'll double check to make sure this indeed is the case. Regarding the os_name and os_version fields. As per my comment in the other PR, I'll wait to see what @irrationalagent thinks about those fields and whether we should keep them in the user level table.

kik-kik:
Will merge the changes to avoid the branch diverting too much from main. For now, the query is being descheduled and once we agree on what fields should be included we can make the final tweak and schedule the query.

PR #3611, opened 2023-02-23
feat(DENG-711): Adding fxa_users_last_seen_v2

kik-kik:
# feat(DENG-711): Adding fxa_users_last_seen_v2

relates to: https://github.com/mozilla/bigquery-etl/pull/3607

---

## Context

This change is introducing `fxa_users_last_seen_v2` to build on top of the work done completed for `fxa_users_daily_v2`. The main reason for introducing v2 here is to enable both v1 and v2 to run in parallel until we validate that v2 is correctly generated at which point v1 if this table can be deprecated.

### Field Changes

`Removed`:
- `days_seen_no_monitor_bits` - based of a monitor field in the daily table which appear to be no longer needed and is being removed in `fxa_users_daily_v2` and hence it is being removed here.

Below fields are being removed as they do not appear to fit in this table, primarily due to how we track the activity fields in this table (each activity field contains information for the last 28 days):
- `country`
- `language`  
- `app_version` 
- `os_name`
- `os_version` 

Unless we do the same for the above fields, I'm not sure those fields should live in this table as they could not apply to all of the 28 days of recorded activity in each partition. This information could be alternatively accessed using the `fxa_users_daily_v2` table instead.

### DIFF
--- a/dags/bqetl_fxa_events.py
+++ b/dags/bqetl_fxa_events.py
@@ -54,6 +54,7 @@ with DAG(
     doc_md=docs,
     tags=tags,
 ) as dag:
+
     docker_fxa_admin_server_v1 = bigquery_etl_query(
         task_id="docker_fxa_admin_server_v1",
         destination_table="docker_fxa_admin_server_sanitized_v1",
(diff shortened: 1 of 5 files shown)


kik-kik:
@irrationalagent I know you use or have in the past used the last_seen tables. In this case do you think the fields mentioned by @akkomar should be in this table or as long as they are in the daily table it's enough?

In another words, should this table contain anything other than the bit fields representing the user activity over the last 28 days?

akkomar (approved):
> @irrationalagent I know you use or have in the past used the last_seen tables. In this case do you think the fields mentioned by @akkomar should be in this table or as long as they are in the daily table it's enough?
> 
> In another words, should this table contain anything other than the bit fields representing the user activity over the last 28 days?

I think it's fine to keep the last value of `country` and `language` here, similar to what we do in `telemetry_clients_last_seen`. Since these fields don't seem to be in use right now I'll let you decide @kik-kik.

akkomar:
@kik-kik note that `generate-dags` test is failing.

akkomar (approved):

PR #3612, opened 2023-02-23
Fix stage deploy test file names

scholtzan:
Followup to https://mozilla.slack.com/archives/C01E8GDG80N/p1677171113619619

Tested in https://github.com/mozilla/bigquery-etl/pull/3615

### DIFF
--- a/bigquery_etl/dryrun.py
+++ b/bigquery_etl/dryrun.py
@@ -30,7 +30,7 @@ except ImportError:
     # python 3.7 compatibility
     from backports.cached_property import cached_property  # type: ignore
 
-
+TEST_PROJECT = "bigquery-etl-integration-test"
 SKIP = {
     # Access Denied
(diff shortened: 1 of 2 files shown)


bigquery_etl/dryrun.py
```diff
@@ -253,6 +253,16 @@
...
+SKIP.update(
+    [
+        p
+        for f in [Path(s) for s in SKIP]
+        for p in glob.glob(
+            f"sql/{TEST_PROJECT}/{f.parent.parent.name}*/{f.parent.name}/{f.name}",
+            recursive=True,
+        )
+    ]
+)
```
scholtzan, line 265: We still want to skip dry runs even for those files that get copied to the stage project, so add their paths here

lelilia (approved):
not that I totally understand what happens here, but it does seem to run with my tests :)

PR #3621, opened 2023-02-28
Bug 1818505 - Monitor number of rows deleted by shredder

relud:


sql/moz-fx-data-shared-prod/monitoring_derived/shredder_rows_deleted_v1/schema.yaml
```diff
@@ -0,0 +1,21 @@
...
+- name: dataset_id
+  type: STRING
+  mode: REQUIRED
+- name: table_id
+  type: STRING
+  mode: REQUIRED
+- name: partition_id
+  type: STRING
+  mode: NULLABLE
+- name: deleted_row_count
```
lucia-vargas-a, line 19: Daniel, is there a way we could calculate in this table the `client_ids` deleted?

relud: doing so would require that we query each table twice, to count the client ids in it before and after, something that would likely incur considerable cost.

I think it would be better to approximate the number of client ids deleted by checking how many rows were deleted from an appropriate clients daily type table.

sql/moz-fx-data-shared-prod/monitoring_derived/shredder_rows_deleted_v1/query.py
```diff
@@ -0,0 +1,138 @@
...
+  -- https://cloud.google.com/bigquery/docs/information-schema-jobs
+  SELECT
+    *
+  FROM
+    `moz-fx-data-shredder.region-us.INFORMATION_SCHEMA.JOBS_BY_PROJECT`
+  UNION ALL
+  SELECT
+    *
+  FROM
+    `moz-fx-data-bq-batch-prod.region-us.INFORMATION_SCHEMA.JOBS_BY_PROJECT`
```
sean-rose, line 32: The `moz-fx-data-backfill-*` projects have also been used for Shredder jobs (as recently as last month), but I'm not sure unioning all those would work, as when I've tried to union a bunch of `INFORMATION_SCHEMA` queries together before I've sometimes run into a "too many metadata operations" kind of error.

There is an effort to make the `JOBS_BY_ORGANIZATION` view available ([DSRE-1106](https://mozilla-hub.atlassian.net/browse/DSRE-1106)) which would be useful for this, but I'm not sure what the ETA on that is.

sql/moz-fx-data-shared-prod/monitoring_derived/shredder_rows_deleted_v1/query.py
```diff
@@ -0,0 +1,138 @@
...
+    if job["deleted_row_count"] is None:
+        table = f'{job["project_id"]}.{job["dataset_id"]}.{job["table_id"]}'
+        try:
+            before = client.get_table(f'{table}@{job["start_time_millis"]}').num_rows
+            after = client.get_table(f'{table}@{job["end_time_millis"]}').num_rows
+        except BadRequest as e:
+            if "Invalid time travel timestamp" in e.message:
+                return None  # row count not available
+            raise
+        job["deleted_row_count"] = before - after
```
sean-rose, line 85: Is there any scenario we could imagine where the row count would be higher after (e.g. a backfill of the table happens during the same time)?  If so, we might want to guard against this being negative.

sql/moz-fx-data-shared-prod/monitoring_derived/shredder_rows_deleted_v1/query.py
```diff
@@ -0,0 +1,138 @@
...
+  WHERE
+    DATE(creation_time) BETWEEN @end_date - 1 AND @end_date
+    AND DATE(end_time) = @end_date
+    AND state = "DONE"
+    AND error_result IS NULL
+)
+SELECT
+  successful_jobs.end_time,
+  shredder.full_job_id AS job_id,
+  successful_jobs.destination_table.*,
```
sean-rose, line 48: I'd suggest being explicit here rather than using `*`, both for clarity and to guard against Google changing the output in the future.
```suggestion
  successful_jobs.destination_table.project_id,
  successful_jobs.destination_table.dataset_id,
  successful_jobs.destination_table.table_id,
```

sql/moz-fx-data-shared-prod/monitoring_derived/shredder_rows_deleted_v1/query.py
```diff
@@ -0,0 +1,138 @@
...
+        except BadRequest as e:
+            if "Invalid time travel timestamp" in e.message:
+                return None  # row count not available
+            raise
+        job["deleted_row_count"] = before - after
+    # move partition id to separate column
+    if "$" in job["table_id"]:
+        (job["table_id"], _, job["partition_id"],) = job[
+            "table_id"
+        ].partition("$")
```
sean-rose, line 90: Is the trailing comma in the assignment tuple necessary?
```suggestion
        (job["table_id"], _, job["partition_id"]) = job[
            "table_id"
        ].partition("$")
```

sql/moz-fx-data-shared-prod/monitoring_derived/shredder_rows_deleted_v1/query.py
```diff
@@ -0,0 +1,138 @@
...
+
+def add_job_metdata(client, job):
+    if job["deleted_row_count"] is None:
+        table = f'{job["project_id"]}.{job["dataset_id"]}.{job["table_id"]}'
+        try:
+            before = client.get_table(f'{table}@{job["start_time_millis"]}').num_rows
+            after = client.get_table(f'{table}@{job["end_time_millis"]}').num_rows
+        except BadRequest as e:
+            if "Invalid time travel timestamp" in e.message:
+                return None  # row count not available
```
sean-rose, line 83: Since this continues on when time travel fails I'd be concerned about the potential for accidentally re-running this for dates older than the 7-day time travel window (e.g. someone clears an entire past day of the `bqetl_monitoring` DAG) and the `shredder_rows_deleted_v1` data for that day getting overwritten without the `SELECT` queries results (which would be the bulk of them since that's the default approach now to avoid BigQuery's DML concurrency limits).

I'd suggest either removing this `except` clause and letting time travel errors cause the overall job to fail, or putting in some kind of safeguard against running this job beyond the time travel window (I'd personally lean toward the former).

sean-rose (approved):
r+wc

sql/moz-fx-data-shared-prod/monitoring_derived/shredder_rows_deleted_v1/query.py
relud: good point, I think I'll make it a cli arg and run the whole query once per project

sql/moz-fx-data-shared-prod/monitoring_derived/shredder_rows_deleted_v1/query.py
relud: good point, i'll throw a ValueError if that happens

sql/moz-fx-data-shared-prod/monitoring_derived/shredder_rows_deleted_v1/query.py
```diff
@@ -0,0 +1,138 @@
...
+)
+standard_args.add_parallelism(parser)
+
+
+def add_job_metdata(client, job):
+    if job["deleted_row_count"] is None:
+        table = f'{job["project_id"]}.{job["dataset_id"]}.{job["table_id"]}'
+        try:
+            before = client.get_table(f'{table}@{job["start_time_millis"]}').num_rows
+            after = client.get_table(f'{table}@{job["end_time_millis"]}').num_rows
```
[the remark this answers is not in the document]
relud: i'll double check, but in my tests the new table state was always available ~500ms before end_time_millis, so it works out

sql/moz-fx-data-shared-prod/monitoring_derived/shredder_rows_deleted_v1/query.py
relud: this was a result of black and variable name changes during development, i believe the parentheses are also unnecessary
```suggestion
        job["table_id"], _, job["partition_id"] = job[
            "table_id"
        ].partition("$")
```

sql/moz-fx-data-shared-prod/monitoring_derived/shredder_rows_deleted_v1/query.py
relud: actually, after thinking: I'll get make a query to get the list of projects, then run a query per project, ensuring i get all the projects and don't run into `too many metadata operations`

sql/moz-fx-data-shared-prod/monitoring_derived/shredder_rows_deleted_v1/query.py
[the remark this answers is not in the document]
relud: yeah, updated table state is consistently available before a job's end_time

PR #3627, opened 2023-03-02
RS-633 update ad click and search with ads

mmccorks:
the old probes expire in firefox 108, we want to use the new ad click and search with ads probes for the ad_click and search_with_ads fields after 108 and the old probes before. 
[Jira ticket](https://mozilla-hub.atlassian.net/browse/RS-333)

does this look like the right place to add in the version check logic or should that be done earlier?


curtismorales:
I moved things around a bit and put in a couple fixes, I think this will work, but I'm also going to get a second opinion if I can.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
```diff
@@ -1513,8 +1499,20 @@ SELECT
   map_sum_aggregates[OFFSET(35)].map AS scalar_parent_urlbar_picked_topsite_sum,
   map_sum_aggregates[OFFSET(36)].map AS scalar_parent_urlbar_picked_unknown_sum,
   map_sum_aggregates[OFFSET(37)].map AS scalar_parent_urlbar_picked_visiturl_sum,
-  map_sum_aggregates[OFFSET(38)].map AS search_with_ads,
-  map_sum_aggregates[OFFSET(39)].map AS ad_clicks,
+  CASE
+    WHEN mozfun.norm.truncate_version(app_version, "major") <= 108
+      THEN map_sum_aggregates[OFFSET(38)].map
+    WHEN mozfun.norm.truncate_version(app_version, "major") <= 108
```
fbertsch, line 1505:
```suggestion
    WHEN mozfun.norm.truncate_version(app_version, "major") > 108
```
?

fbertsch: I would use IF here:
```
IF(
  mozfun.norm.truncate_version(app_version, "major") > 108,
  map_sum_aggregates[OFFSET(113)].map,
  map_sum_aggregates[OFFSET(38)].map
) AS search_with_ads
```

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
```diff
@@ -1353,7 +1337,9 @@ aggregates AS (
       STRUCT(ARRAY_CONCAT_AGG(browser_search_adclicks_urlbar_persisted)),
       STRUCT(ARRAY_CONCAT_AGG(scalar_parent_sidebar_opened)),
       STRUCT(ARRAY_CONCAT_AGG(scalar_parent_sidebar_search)),
-      STRUCT(ARRAY_CONCAT_AGG(scalar_parent_sidebar_link))
+      STRUCT(ARRAY_CONCAT_AGG(scalar_parent_sidebar_link)),
+      STRUCT(ARRAY_CONCAT_AGG(browser_ad_clicks_combined)),
+      STRUCT(ARRAY_CONCAT_AGG(browser_search_with_ads_combined))
```
fbertsch, line 1342: nit: I would switch the order here, to keep it consistent with the old versions (`browser_search_with_ads_combined`, then `browser_ad_clicks_combined`)

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
fbertsch: On third thought, can we do this in the view? Less likelihood of problems

fbertsch:
Everything except for those final two fields (`ad_clicks`, `search_with_ads`) looks additive so should be fine to change. Moving those to the view _should_ make this relatively low risk.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
curtismorales: That seems reasonable to me, with 2 caveats:

1. `clients_first_seen` and `clients_last_seen` pull directly from the table here, so they're not going to pick up the new logic for these two fields. I don't have a super great understanding of how `clients_first_seen` and `clients_last_seen` are used, but should we also change the views for those 2 tables?
   
   For that matter, is there a reason that those tables query this table directly and not the view?

2. Is there a more descriptive name we can use to refer to these new fields and make it clear that they are the versions coming from the new probe? @mmccorks any thoughts?

curtismorales: Oh and thanks for catching that!

mmccorks: Thanks Frank. 
Curtis - I agree that we want `clients_first_seen` and `clients_last_seen` to also have the data from the new probes. If that means updating those views as well, that works. I'm fine with doing this in the view if that makes the most sense.  
For the filed names - I don't know that we want to change the name in the risk of breaking existing queries against this table, but we could add a note in the analysis gotchas and update the documentation for this table on DTMO to make sure that this change is documented properly.

curtismorales:
This is still failing a test `moz-fx-data-shared-prod.telemetry_derived.urlbar_clients_daily_v1:test_overactive_filter` for reasons that I can't make sense of -- I don't see what this change has to do with that. I also can't reproduce the error when I run it locally against a sandbox project -- the test passes for me in that case.

@fbertsch I wonder if you're able to take a quick look? Otherwise I'm out of town next week but can come back to this when I get back

curtismorales:
This is definitely not relevant anymore; I'm going to close it

PR #3518, opened 2023-01-19
RS-595

rebecca-burwei:
We are adding a new boolean field to search clients daily. This field indicates whether or not the SAP and tagged searches are monetizable. It uses the monetized_search UDF defined [here](https://github.com/mozilla/private-bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/udf/monetized_search/udf.sql#L123).

Jira: [RS-595](https://mozilla-hub.atlassian.net/browse/RS-595)

Checklist for reviewer:

- [ ] Commits should reference a bug or github issue, if relevant (if a bug is referenced, the pull request should include the bug number in the title).
- [ ] If the PR comes from a fork, trigger integration CI tests by running the [Push to upstream workflow](https://github.com/mozilla/bigquery-etl/actions/workflows/push-to-upstream.yml) and provide the `<username>:<branch>` of the fork as parameter. The parameter will also show up
in the logs of the `manual-trigger-required-for-fork` CI task together with more detailed instructions.
- [ ] If adding a new field to a query, ensure that the schema and dependent downstream schemas have been updated.
- [ ] When adding a new derived dataset, ensure that data is not available already (fully or partially) and recommend extending an existing dataset in favor of creating new ones. Data can be available in the [bigquery-etl repository](https://github.com/mozilla/bigquery-etl), [looker-hub](https://github.com/mozilla/looker-hub) or in [looker-spoke-default](https://github.com/mozilla/looker-spoke-default/tree/e1315853507fc1ac6e78d252d53dc8df5f5f322b).

For modifications to schemas in restricted namespaces (see [`CODEOWNERS`](/CODEOWNERS)):
- [ ] Follow the [change control procedure](https://docs.google.com/document/d/1TTJi4ht7NuzX6BPG_KTr6omaZg70cEpxe9xlpfnHj9k/edit#heading=h.ttegrcfy18ck)

[RS-595]: https://mozilla-hub.atlassian.net/browse/RS-595?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3631, opened 2023-03-02
CI fixes for supporting private UDFs in bigquery-etl - DENG-735

ANich:
This PR does two things:

1. Modifies the `normalize_search_engine` udf to be more minimal than it currently is. This UDF is a stub that is superseded by a private UDF. This PR modifies it so it has a (very limited) subset of the actual production logic rather than completely different logic. This is done so that tests locally and tests in CI (using production logic) both pass. The tests for the production UDF are next to its code and are more specific. The PR also modifies the `search_clients_last_seen` test to use the minimal logic. 

2. Copies UDFs from private bigquery-etl to the `sql` dir when tests run. For situations such as https://github.com/mozilla/bigquery-etl/pull/3518 where we want to use UDFs from private-bigquery-etl in bigquery-etl queries, tests should still pass. 

A side effect of this is that UDF tests for private UDFs now run in bigquery-etl CI (see: https://github.com/mozilla/private-bigquery-etl/pull/190).


.circleci/config.yml
```diff
@@ -361,6 +361,12 @@ jobs:
       - *java_deps
       - *attach_generated_sql
       - *copy_generated_sql
+      - run:
+          name: Copy files from private-generated-sql
+          command: |
+            apt update
+            apt install -y rsync
+            rsync --archive /tmp/workspace/private-generated-sql/sql/ sql/
```
scholtzan, line 369: I'd suggest moving the `rsync` between line 386 and 387, otherwise the private artifacts will get deployed to stage on every CI run since they are not part of `generated-sql`

scholtzan (approved):
🎉

Issue #3515, opened 2023-01-19
Support setting billing project when running queries and backfill


scholtzan:
For running longer backfills it would be nice if bqetl backfill allowed to specify a different billing project

BenWu:
I'm going to pick this up because I need the billing project for regular queries and I can add it to backfills as well since they depend on the same function.  I'm going to add the option but I'll leave it unused for now

BenWu:
This is to support usage of on-demand projects re:https://bugzilla.mozilla.org/show_bug.cgi?id=1892284.  Nearly all queries will continue to use the default

PR #3634, opened 2023-03-03
 Support setting billing project when running backfill

wwyc:
#3515: the billing project is set from the `project_id` param in the backill command.  Changes were made where the `destination_project_id` is extracted from the query file.


bigquery_etl/cli/query.py
```diff
@@ -1681,6 +1697,7 @@ def _attach_metadata(query_file_path: Path, table: bigquery.Table) -> None:
         return
 
     table.description = metadata.description
+
```
wwyc, line 1700:
```suggestion
```

bigquery_etl/cli/query.py
```diff
@@ -478,13 +482,11 @@ def _backfill_query(
 
     backfill_date = backfill_date.strftime("%Y-%m-%d")
     if backfill_date not in exclude:
-        if no_partition:
-            destination_table = table
-        else:
-            destination_table = f"{table}${partition}"
+        if not no_partition:
+            destination_table = f"{destination_table}${partition}"
```
ANich, line 486: I'd add a comment just because the double negative is a little weird
```suggestion
        # If table is partitioned, add partition decorator: 
        if not no_partition:
            destination_table = f"{destination_table}${partition}"
```

scholtzan (approved):

bigquery_etl/cli/query.py
```diff
@@ -835,11 +839,11 @@ def _run_query(
                     and destination_table is not None
                     and re.match(DESTINATION_TABLE_RE, destination_table)
                 ):
-                    destination_table = "{}:{}.{}".format(
+                    qualified_destination_table = "{}:{}.{}".format(
                         public_project_id, dataset_id, destination_table
                     )
                     query_arguments.append(
-                        "--destination_table={}".format(destination_table)
+                        "--destination_table={}".format(qualified_destination_table)
```
ANich, line 846: Awkward moment because I think I suggested this naming change during the session haha but I think we should probably leave this block as-is at least in this PR since it's likely unrelated to backfills

ANich (approved):
LGTM. The `_run_query` function is pretty central to a lot of things (which makes me think we should put in its own module) though so I'd test this once or twice in the context of a backfill and in the context of what a regular ETL run would look like. Also probably a morning test rather than an EOD test 😆

wwyc:
It was discovered during testing that a [search](https://github.com/mozilla/bigquery-etl/blob/f08729e12feb8398c074eecfc0834687a4186733/bigquery_etl/cli/query.py#L621) for the query file is using `project_id` instead of the `destination_project_id`.  One potential solution is to require a fully qualified destination table name in cli command.  Decision has been made to incorporate further changes with [Managed BigQuery Backfills Proposal](https://docs.google.com/document/d/1uE20CE2TkXSSap7xmEHV89lZntjxsZhUy50coQGBkaY/edit?usp=sharing) in order to consolidate the change process.

PR #3644, opened 2023-03-08
DENG-658 - Initial table definitions for dl_token processing. 

gleonard-m:
Checklist for reviewer:

**Notes:**

1.  See https://mozilla-hub.atlassian.net/browse/DENG-660
2. This code change will process stubattribution and Google Analytics data during 2023Q1, until https://mozilla-hub.atlassian.net/browse/DS-2467 is Completed.  A new version will then be released.
3. Includes update to sql pytest_plugin to account for tablenames with date suffixes. @relud Can you look at this change in `bigquery_etl/pytest_plugin/sql.py`
4. Tests are included in `tests/sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v1`


bigquery_etl/pytest_plugin/sql.py
```diff
@@ -123,6 +124,20 @@ def runtest(self):
...
+                # e.g. see moz-fx-data-marketing-prod.65789850.ga_sessions_20230214
+                # A query using that table uses moz-fx-data-marketing-prod.65789850.ga_sessions_*
+                # with the date appended to allow for daily processing.
+                try:
+                    datetime.datetime.strptime(table_name[-8:], "%Y%m%d")
+                    generic_table_name = table_name[:-8] + "*"
+                    generic_original = original[:-8] + "*"
+                    query = query.replace(generic_original, generic_table_name)
+                except ValueError:
+                    pass
```
relud, line 140: lgtm, but should probably take advantage of try/catch/else to only catch exceptions from the intended line:
```suggestion
                try:
                    datetime.datetime.strptime(table_name[-8:], "%Y%m%d")
                except ValueError:
                    pass
                else:
                    generic_table_name = table_name[:-8] + "*"
                    generic_original = original[:-8] + "*"
                    query = query.replace(generic_original, generic_table_name)
```

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v1/query.sql
```diff
@@ -0,0 +1,315 @@
...
+  WHERE
+    DATE(s.timestamp) = @submission_date
+    AND s.jsonPayload.fields.log_type = 'download_started'
+  GROUP BY
+    stub_visit_id,
+    dltoken,
+    DATE(s.timestamp)
+),
+-- This will drop all the ga_sessions w/o a DLtoken but keep DLtoken without a GA session.
+-- This will also result in multiple rows is the ga.client_id is not unique for the day
```
fbertsch, line 128: nit
```suggestion
-- This will also result in multiple rows as the ga.client_id is not unique for the day
```

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v1/query.sql
```diff
@@ -0,0 +1,315 @@
...
+-- since this visit_id is missing from the stub.
+downloads_and_ga_session AS (
+  SELECT
+    gs.client_id AS client_id,
```
fbertsch, line 132: We can `COALESCE(stub_visit_id, client_id) AS client_id` and remove the `stub_visit_id` from the SELECT and GROUP BY clauses

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v1/query.sql
```diff
@@ -0,0 +1,315 @@
...
+    WHEN stub_visit_id = 'something'
+      THEN NULL
+    WHEN client_id IS NULL
+      THEN NULL
+    --only set to null when we do not have a session.  The value can be null when we do have a session so set those to 0.
+    ELSE IFNULL(time_on_site, 0)
+  END
+  time_on_site,
+  CASE
+    WHEN nrows > 1
```
fbertsch, line 185: If I'm understanding correctly, these fields will be `NULL` for any users with more than one session on that day?

fbertsch:
With the query as-is, it looks to me like we'd only store attribution data for clients with GA a single session that has page_view and download events on the same calendar day as the dl_token log entry. Does that parallel your understanding?

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v1/query.sql
```diff
@@ -0,0 +1,315 @@
...
+    WHEN browser IN ('Edge')
+      THEN 'Edge'
+    WHEN browser IN ('Chrome')
+      THEN 'Chrome'
+    WHEN browser IN ('Safari')
+      THEN 'Safari'
+    WHEN browser IN ('(not set)')
+      THEN NULL
+    WHEN browser IS NULL
+      THEN NULL
```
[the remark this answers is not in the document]
gleonard-m: Unfortunately is does not behave in the same way.  The NULL check is not satisfied and the ELSE clause is executed.

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v1/query.sql
```diff
@@ -0,0 +1,315 @@
...
+    client_id,
+    visit_id
+),
+ga_sessions_with_hits AS (
+  SELECT
+    *
+  FROM
+    ga_sessions ga
+  JOIN
+    pageviews p
```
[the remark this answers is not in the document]
gleonard-m: You are correct, update will be made and test added.

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v1/query.sql
gleonard-m: I intentionally keep them as separate fields to identify cases where the `stub_visit_id` is not set vs no matching GA Session.

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v1/query.sql
```diff
@@ -0,0 +1,315 @@
...
+    all_hits
+  GROUP BY
+    client_id,
+    visit_id
+),
+ga_sessions AS (
+  SELECT
+    clientId AS client_id,
+    visitId AS visit_id,
+    ANY_VALUE(geoNetwork.country) AS country,
```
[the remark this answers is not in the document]
gleonard-m: Isn't `mozfun.stats.mode_last` used for array type, not as an aggregation function?

SuYoungHong (commented):
Hi Glenda, a few notes to add at the top level: 

1) I did some data QA, and it looks like some null values are popping up in cases where we expect values. So in rows that are exception=null (meaning has a matching client ID in GA, with a single session, hence should have all GA derived values), I'm seeing (some) cases where the following fields are null: 
* normalized_country_code
* normalized_browser
* browser_major_version
* pageviews
* unique_pageviews

data for reference: https://colab.research.google.com/drive/1zlVxOgGFAXOeV-Ei6JS90wpDC3RJAiCf#scrollTo=yEWwgo2Z7Q2K&line=6&uniqifier=1

2) I want to point out, the cases where a single GA session has multiple downloads (dl tokens) isn't captured in this dataset. We lose that information with the current ETL. 
* We could choose to leave it out of scope since this is going to be a temporary ETL, and deal with it in V2
* If we choose to add it to this, it could be a pretty simple fix, just add a self join to the DL log table, on stub_visit_id, to create a "num_dls_same_session" field (which counts, for each record, how many other downloads happened on the same GA session). 

3) It would be nice to have documentation giving a top level summary of the possible behaviors of downloads and what to expect in the dataset. Something like: 
* Each row in this table should be a unique dltoken
* A single stub_visit_id can have multiple dltokens (a session resulted in multiple downloads)
* The stub_visit_id is '', '(not set)', or 'something':
	- exception will be DOWNLOAD_SESSION_ID_EMPTY, DOWNLOAD_SESSION_ID_VALUE_NOT_SET, or DOWNLOAD_SESSION_ID_VALUE_SOMETHING
	- All GA dervied values will be null
* The stub_visit_id is valid but not found in that day's GA sessions: 
	- exception will be MISSING_GA_CLIENT
	- All GA dervied values will be null
* A GA session doesn't have a dltoken with a corresponding stub_visit_id/clientId match
	- excluded from this dataset
* The stub_visit_id has a clientId match but that clientId has multiple GA sessions that day:
	- exception will be GA_UNRESOLVABLE
	- All GA dervied values will be null
* A stub_visit_id has multiple dltokens associated with it
	- Cannot be deterimed from this dataset
* A single dltoken has multiple records
	- count_dltoken_duplicates will be > 1

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v1/query.sql
```diff
@@ -0,0 +1,313 @@
...
+    WHEN stub_visit_id = '(not set)'
+      THEN NULL
+    WHEN stub_visit_id = 'something'
+      THEN NULL
+    WHEN client_id IS NULL
+      THEN NULL
+    --only set to null when we do not have a session.  The value can be null when we do have a session so set those to 0.
+    ELSE IFNULL(time_on_site, 0)
+  END
+  time_on_site,
```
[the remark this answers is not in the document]
gleonard-m: The logic using stub_visit_id was needed to account for NULL vs 0 values.  Moving the code to impute the value when we first full the value removes the need for that custom logic.  End result, time_on_site is handled like other fields in the last stage of the query.

gleonard-m:
1.
> 
    I did some data QA, and it looks like some null values are popping up in cases where we expect values. So in rows that are exception=null (meaning has a matching client ID in GA, with a single session, hence should have all GA derived values), I'm seeing (some) cases where the following fields are null:

    normalized_country_code
    normalized_browser
    browser_major_version
    pageviews
    unique_pageviews

**normalized_country_code:**  2 causes
1. country = (not set)
2. The country code mapping table does not contain the values, country values will be added in a separate PR.

**normalized_browser:**  Caused by browser = (not set) therefore cannot be normalized.
**browser_major_version:** Caused by browser_version=(not set) therefore version cannot be extracted.
**pageviews:**  Caused by GA session not having a hit.type=PAGE
**unique_pageviews:** Caused by GA session not having a hit.type=PAGE

2. 

>     I want to point out, the cases where a single GA session has multiple downloads (dl tokens) isn't captured in this dataset. We lose that information with the current ETL.
> 
>     We could choose to leave it out of scope since this is going to be a temporary ETL, and deal with it in V2
>     If we choose to add it to this, it could be a pretty simple fix, just add a self join to the DL log table, on stub_visit_id, to create a "num_dls_same_session" field (which counts, for each record, how many other downloads happened on the same GA session).

I'm not seeing this issue.  I added a unit test for this case and it is passing.  The initial query for the DL log table does group by the visit_id and the dltoken

3. 

> It would be nice to have documentation giving a top level summary of the possible behaviors of downloads and what to expect in the dataset.

Will add a detailed description which will be viewable from the BigQuery table definition.

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v1/query.sql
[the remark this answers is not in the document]
fbertsch: It is. We use it like `mozfun.stats.mode_last(ARRAY_AGG($field ORDER BY x IGNORE NULL))`.

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v1/query.sql
fbertsch: Got it. Help me understand what we're going to be using the v1 table for then? Shouldn't we wait until session_id matching is enabled?

gleonard-m: The priority was to support the session_id matching however with the delayed implementation due to the inconsistent GA session value `_v1` will be released instead.  Since the source data is only stored for 90 days we can use `_v1` to have some representation of a full year of data.

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v1/query.sql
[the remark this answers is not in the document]
gleonard-m: Updated

PR #3646, opened 2023-03-09
 DENG-514 capture windows version in active users aggregates

wwyc:
[DENG-514](https://mozilla-hub.atlassian.net/browse/DENG-514): Updated `os_version` column to use [windows_version_info udf](https://github.com/mozilla/bigquery-etl/blob/main/sql/mozfun/norm/windows_version_info/udf.sql) to capture the correct windows version.

The following table is created for validations:  `moz-fx-data-shared-prod.analysis.wchan-deng-514-windows`

[DENG-514]: https://mozilla-hub.atlassian.net/browse/DENG-514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sql_generators/active_users/templates/desktop_query.sql
+++ b/sql_generators/active_users/templates/desktop_query.sql
@@ -13,7 +13,10 @@ WITH todays_metrics AS (
     EXTRACT(YEAR FROM first_seen_date) AS first_seen_year,
     days_since_seen,
     os,
-    normalized_os_version AS os_version,
+    COALESCE(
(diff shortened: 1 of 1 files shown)


lucia-vargas-a (approved):
LGTM.
Looking at Jira, it doesn't look like we need to keep both fields for analysis, but only the one with the updated calculation.

PR #3667, opened 2023-03-16
bug(): added org_mozilla_firefox_derived/client_deduplication_v1 to dry run skip

kik-kik:
# bug(): added org_mozilla_firefox_derived/client_deduplication_v1 to dry run skip


fbertsch (changes requested):
There's a few of these (in e.g. `org_mozilla_firefox_beta_derived`), so I think we should add it to the glob on line 184.
```
    *glob.glob(
        "sql/moz-fx-data-shared-prod/**/client_deduplication*/*.sql",
        recursive=True,
),
```
should work

PR #3685, opened 2023-03-24
Support Jinja templating in query files

scholtzan:
With this change any `.sql` file will be treated as a Jinja template. Rendering of templates happens before queries are executed, UDFs and views are deployed as well as before a dryrun. The `generated-sql` branch will contain the rendered SQL files.

There is also a new CLI command `bqetl query render` which renders the SQL file templates

Next steps:
* metric-hub integration
* update docs

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -52,7 +52,7 @@ from ..query_scheduling.generate_airflow_dags import get_dags
 from ..schema import SCHEMA_FILE, Schema
 from ..util import extract_from_query_path
 from ..util.bigquery_id import sql_table_id
-from ..util.common import random_str
+from ..util.common import random_str, render
 from .dryrun import dryrun
 from .generate import generate_all
 
@@ -860,9 +860,17 @@ def _run_query(
             # point to a public table it needs to be passed as parameter for the query
             query_arguments.append("--destination_table={}".format(destination_table))
 
-        with open(query_file) as query_stream:
-            # run the query as shell command so that passed parameters can be used as is
-            subprocess.check_call(["bq"] + query_arguments, stdin=query_stream)
+        query_arguments.append(
+            render(
+                query_file.name,
+                template_folder=str(query_file.parent),
+                templates_dir="",
+                format=False,
+            )
+        )
+
+        # run the query as shell command so that passed parameters can be used as is
+        subprocess.check_call(["bq"] + query_arguments)
 
 
 @query.command(
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -52,7 +52,8 @@ from ..query_scheduling.generate_airflow_dags import get_dags
 from ..schema import SCHEMA_FILE, Schema
 from ..util import extract_from_query_path
 from ..util.bigquery_id import sql_table_id
-from ..util.common import random_str, render
+from ..util.common import random_str
+from ..util.common import render as render_template
 from .dryrun import dryrun
 from .generate import generate_all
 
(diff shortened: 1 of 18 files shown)


Marlene-M-Hirose (approved):

bigquery_etl/cli/query.py
```diff
@@ -1186,6 +1195,55 @@ def initialize(name, sql_dir, project_id, dry_run):
...
+@query.command(
+    help="""Render a query Jinja template.
+
+    Examples:
+
+    \b
+    # Backfill for specific date range
+    # second comment line
+    ./bqetl query render telemetry_derived.ssl_ratios_v1 \\
+      --output-dir=/tmp
```
sean-rose, line 1207: The comments don't seem to fit with this example.

scholtzan: yikes

bigquery_etl/format_sql/tokenizer.py
```diff
@@ -736,6 +736,33 @@ class Literal(Token):
     )
 
 
+class JinjaExpression(Token):
+    """Jinja expression delimiters {{ }}.
+
+    May be followed by no whitespace or a new line and increased indent.
+    """
+
+    pattern = re.compile(r"{\s*{.*}\s*}")
```
sean-rose, line 745: I don't think allowing whitespace between the pairs of curly brackets makes sense.  Did you perhaps mean `{{\s*.*\s*}}` to allow newlines before/after the main content?

Also, even with `{{\s*.*\s*}}` that won't allow multiple lines of content because `.` doesn't match newlines by default.

Perhaps something like this instead?
```suggestion
    pattern = re.compile(r"{{.*?}}", re.DOTALL)
```

(these same concerns also apply to `JinjaStatement` and `JinjaComment`)

bigquery_etl/util/common.py
```diff
@@ -53,9 +53,15 @@ def random_str(length: int = 12) -> str:
     return "".join(random.choice(string.ascii_lowercase) for i in range(length))
 
 
-def render(sql_filename, format=True, template_folder="glean_usage", **kwargs) -> str:
+def render(
+    sql_filename,
+    format=True,
+    template_folder="glean_usage",
```
sean-rose, line 59: Since this is now being more generally used, IMO it'd be better to not have this default to `glean_usage`, but instead explicitly specify `template_folder="glean_usage"` where that's applicable.

bigquery_etl/util/common.py
```diff
@@ -53,9 +53,15 @@ def random_str(length: int = 12) -> str:
...
+    templates_dir="templates",
```
sean-rose, line 60: Maybe make `templates_dir` optional as well, set to `templates_dir="templates"` only where applicable?

Needing to override this repeatedly with `templates_dir=""` is a bit of a code smell to me.

bigquery_etl/format_sql/tokenizer.py
```diff
@@ -658,7 +658,7 @@ class AliasSeparator(SpaceBeforeBracketKeyword):
     """
 
     pattern = re.compile(
-        r"AS(?=\s+(?!(WITH|SELECT|STRUCT|ARRAY)\b)[a-z_`(])", re.IGNORECASE
+        r"AS(?=\s+(?!(WITH|SELECT|STRUCT|ARRAY)\b)[a-z_`({])", re.IGNORECASE
```
lucia-vargas-a, line 661: Might need the closing '}'?

tests/routine/test_parse_routine.py
```diff
@@ -52,37 +53,6 @@ def test_raw_routine_from_file(self):
         )
         assert result.dependencies == ["udf.test_bitmask_lowest_28"]
 
-    def test_raw_routine_from_text(self):
```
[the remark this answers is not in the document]
scholtzan: Yeah, I removed the `from_text` parameter from `RawRoutine.from_file` since it wasn't used anywhere in the code base except for this test, which is just checking that the `from_text` parameter works correctly.

PR #3687, opened 2023-03-28
DENG-774 Change Control for active_users_aggregates

lucia-vargas-a:
Implements [DENG-774](https://mozilla-hub.atlassian.net/browse/DENG-774)

[DENG-774]: https://mozilla-hub.atlassian.net/browse/DENG-774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bigquery_etl/cli/query.py
```diff
@@ -1116,9 +1116,46 @@ def validate(
...
-        validate_metadata.validate(query.parent)
+        validate_metadata.validate(query.parent, project_id, sql_dir)
         dataset_dirs.add(query.parent.parent)
 
+    if not query_files:
+        # run SQL generators if no matching query has been found.
+        ctx.invoke(
+            generate_all,
+            output_dir=ctx.obj["TMP_DIR"],
+            ignore=[
```
lucia-vargas-a, line 1127: Generating sql only for files with change control implemented.

bochocki (approved):
This LGTM! I don't think I fully understand how all the pieces fit together with the bigger picture (e.g. I wouldn't have caught @scholtzan's comment about the Airflow task), but from what I'm able to follow it looks great. I just had a few minor refactoring suggestions.

bigquery_etl/cli/query.py
```diff
@@ -1835,8 +1872,8 @@ def _validate_schema_from_path(
 def validate_schema(
     ctx, name, sql_dir, project_id, use_cloud_function, respect_dryrun_skip
 ):
-    """Validate the defined query schema against the query and destination table."""
-    query_files = paths_matching_name_pattern(name, sql_dir, project_id)
+    """Validate the defined querypaths_matching_name_pattern schema against the query and destination table."""
```
scholtzan, line 1875:
```suggestion
    """Validate the defined query schema against the query and destination table."""
```

bigquery_etl/cli/query.py
```diff
@@ -1835,8 +1872,8 @@ def _validate_schema_from_path(
 def validate_schema(
     ctx, name, sql_dir, project_id, use_cloud_function, respect_dryrun_skip
 ):
-    """Validate the defined query schema against the query and destination table."""
-    query_files = paths_matching_name_pattern(name, sql_dir, project_id)
```
scholtzan, line 1839: I think we want to keep this

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -27,7 +32,45 @@ def validate_public_data(metadata, path):
...
+    path_in_codeowners = os.path.join(query_files_path, project_id, path_to_add)
+
+    if CHANGE_CONTROL_LABEL in metadata.labels:
+        # This label requires to have at least one owner in the metadata file.
+        # And for any of the owners, at least one entry in the CODEOWNERS file.
+        counter = 0
+        rows_expected = []
+        row_to_search_for = ""
+        for owner in metadata.owners:
+            if not owner.__contains__("@"):
```
scholtzan, line 54:
```suggestion
            if "@" not in owner:
```

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -27,7 +32,45 @@ def validate_public_data(metadata, path):
...
+    if not query_files_path:
+        query_files_path = "sql"
+
+    path_to_add = file_path.partition(f"{project_id}/")[2]
+    path_in_codeowners = os.path.join(query_files_path, project_id, path_to_add)
+
+    if CHANGE_CONTROL_LABEL in metadata.labels:
+        # This label requires to have at least one owner in the metadata file.
+        # And for any of the owners, at least one entry in the CODEOWNERS file.
+        counter = 0
```
scholtzan, line 50: This could also be a boolean flag `has_codeowner` that gets set if the owner is in the CODEOWNERS file. Could make the condition a little more readable

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -27,7 +32,45 @@ def validate_public_data(metadata, path):
...
 
-def validate(target):
+def validate_change_control(
+    file_path, metadata, project_id, query_files_path, codeowners_file
+):
+    """Verify that a query is correctly setup for change control."""
+    if not project_id:
+        project_id = "moz-fx-data-shared-prod"
+    if not query_files_path:
+        query_files_path = "sql"
```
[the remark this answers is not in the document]
lucia-vargas-a: Only using it because the default values were not properly retrieved. Let me see what I can do to avoid these lines of code.

CODEOWNERS
```diff
@@ -2,3 +2,9 @@
 # described in https://docs.google.com/document/d/1TTJi4ht7NuzX6BPG_KTr6omaZg70cEpxe9xlpfnHj9k/
 /sql/moz-fx-data-shared-prod/search_terms @whd @jasonthomas
 /sql/moz-fx-data-shared-prod/search_terms_derived @whd @jasonthomas
+/sql/moz-fx-data-shared-prod/fenix_derived/active_users_aggregates_v1 @mozilla/kpi_table_reviewers
```
lucia-vargas-a, line 5: @bochocki preferences about the name of this group? I'll check when it's available before merging the PR.

tests/cli/test_cli_metadata.py
```diff
@@ -0,0 +1,163 @@
...
+
+from bigquery_etl.metadata.parse_metadata import Metadata
+from bigquery_etl.metadata.validate_metadata import validate_change_control
+
+
+class TestMetadata:
+    @pytest.fixture
+    def runner(self):
+        return CliRunner()
+
```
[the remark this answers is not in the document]
lucia-vargas-a: Great Brad, thank you. Implemented with some refactoring of the tests.

CODEOWNERS
bochocki: This sounds good to me!

whd (commented):
I had some trouble reading through the validation section. I think it would be simpler to reason about if we required all github identities listed as `owners` in metadata to be listed as CODEOWNERS for tables subject to this change control.

CODEOWNERS
lucia-vargas-a: Resolved.

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -27,6 +32,54 @@ def validate_public_data(metadata, path):
...
+    sql_dir="sql",
+):
+    """Verify that a query is correctly setup for change control."""
+    path_to_add = file_path.partition(f"{project_id}/")[2]
+    path_in_codeowners = os.path.join(sql_dir, project_id, path_to_add)
+    has_codeowner = CHANGE_CONTROL_LABEL in metadata.labels
+
+    if has_codeowner:
+        # This label requires to have at least one owner in the metadata file.
+        # And for any of the owners, at least one entry in the CODEOWNERS file.
```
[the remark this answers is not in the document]
lucia-vargas-a: Yes, in this case, the team includes people from both teams and that is aligned with the requirement.
Thank you for the link, noted in case the requirement changes in the future.

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -27,6 +32,54 @@ def validate_public_data(metadata, path):
...
+        rows_expected = []
+        row_to_search_for = ""
+        for owner in metadata.owners:
+            if "@" not in owner:
+                owner = f"@{owner}"
+            row_to_search_for = f"/{path_in_codeowners} {owner}"
+            rows_expected.append(row_to_search_for)
+
+            with open(codeowners_file) as owners:
+                if row_to_search_for in owners.read():
```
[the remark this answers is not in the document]
lucia-vargas-a: Slight change of syntax.

dags/bqetl_analytics_aggregations.py
```diff
@@ -118,6 +118,7 @@
         email=[
             "gkaberere@mozilla.com",
             "lvargas@mozilla.com",
+            "mozilla/kpi_table_reviewers",
```
scholtzan, line 121: Will this cause any issues when sending emails since it is not technically an email address?

scholtzan (approved):

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -27,6 +32,55 @@ def validate_public_data(metadata, path):
...
+            content = owners_file.readlines()
+        occurrences = 0
+        rows_expected = []
+        for meta_owner in metadata.owners:
+            if "@" not in meta_owner:
+                meta_owner = f"@{meta_owner}"
+            row_to_search_for = f"/{path_in_codeowners} {meta_owner}"
+            rows_expected.append(row_to_search_for)
+            for line in content:
+                if row_to_search_for == line.rstrip("\n"):
```
whd, line 70: I'm still somewhat confused about the use of `occurrences` here, which will always be `0` or `1`, since this check appears to require that there be at least one line matching the format `row_to_search_for`. That value is generated for each metadata owner, one at a time, so per the string comparison strictly requires that there exist a `CODEOWNERS` line containing only a single `CODEOWNER` from among those specified in metadata. Therefore I would expect the following to fail the check:

```
/{path_in_codeowners} {meta_owner} {other_meta_owner}
```

which is particularly confusing because  the error message that is emitted presumably looks like:

`The metadata includes label change_controlled, which requires CODEOWNERS to include at least one owner's record: [{meta_owner}, {other_meta_owner}]`

The requirements in the error message are met by my example (`two` relevant CODEOWNERS is at least `one`). The most trivial way to deal with this is to change the wording and specification to require `exactly one owner's record` instead of `at least one owner's record`, to match what the check implements, but I think we should relax the parsing of the CODEOWNERS line to match the natural language specification and error message.

Related but separate, due to [precedence semantics](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax), a CODEOWNERS file might pass the CI check but fail to enforce the expected change control via the following:

```
/{path_in_codeowners} @meta-owner
/{path_in_codeowners} @user-not-in-codeowners
```

This should be caught by a human reviewing the PR so it's perhaps not worth explicitly verifying in CI, but given that the we're implementing these checks to be very specific about change control I'm noting it as an edge case.

dags/bqetl_analytics_aggregations.py
lucia-vargas-a: I think it would. I added a change to [remove non-email owners ](https://github.com/mozilla/bigquery-etl/blob/08e997a2d80fd52434cba6584b4a6a64c1fa67fb/bigquery_etl/query_scheduling/task.py#L333-L341) when generating the email list for Airflow.

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -27,6 +32,65 @@ def validate_public_data(metadata, path):
...
+        with open(codeowners_file, "r") as owners_file:
+            content = owners_file.readlines()
+        owners_list = []
+        for owner in metadata.owners:
+            if "@" not in owner:
+                owner = f"@{owner}"
+            owners_list.append(owner)
+        sample_row_all_owners = f"/{path_in_codeowners} {(' '.join(owners_list))}"
+
+        if not [line for line in content if path_in_codeowners in line]:
```
whd, line 69: Since the logic has been refactored back to using substring comparisons, I think the comments from my first review are relevant again about accepting commented strings e.g. I would expect this to erroneously pass the check:

```
# {path_in_codeowners} {valid_owner}
```

Can you add a test for this, as well as a test verifying that from my second review a line like:

```
{path_in_codeowners} {valid_owner_1} {valid_owner_2}
```

is now accepted?

lucia-vargas-a: Agree, added a few test scenarios for github identities, one/none/many owners in CODEOWNERS and the handling of lines starting with '#"

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -27,6 +32,72 @@ def validate_public_data(metadata, path):
...
+            click.echo(
+                click.style(
+                    f"The metadata for {file_path} has the label"
+                    f" change_controlled but it's missing owners."
+                )
+            )
+            return False
+
+        with open(codeowners_file, "r") as owners_file:
+            content = owners_file.readlines()
```
whd, line 61: I suggest moving the `startswith` check here to avoid duplicating it below.

whd (approved):
r+wc

Thank you for adding the tests. I think it's important to decide whether to require only `owners` in CODEOWNERS, but I don't think that needs another round of review if a test is added.

After this is merged https://github.com/mozilla/bigquery-etl/settings/branch_protection_rules/4491948 should be updated to require the `validate-metadata` check since it isn't required presently.

bigquery_etl/metadata/validate_metadata.py
lucia-vargas-a: Yes, also to not validate lines unnecessarily.

PR #3690, opened 2023-03-30
DENG-762 Adjust mobile views to query Glean data

lucia-vargas-a:
Implements DENG-762, according to the in the [announcement email](https://groups.google.com/a/mozilla.org/g/fx-data-dev/c/jqbrzb1PTyE). As mentioned in the email, this PR should be merged on April 3rd.


sql_generators/active_users/templates/focus_android_view.sql
```diff
@@ -2,7 +2,8 @@ CREATE OR REPLACE VIEW
   `{{ project_id }}.{{ app_name }}.active_users_aggregates`
 AS
 SELECT
-  * EXCEPT (app_version),
+  * EXCEPT (app_version, app_name),
+  IF(app_name IN ('Focus Android Glean', 'Focus Android'), 'Focus Android', app_name) AS app_name,
```
wwyc, line 6: Is there a need to capture the case where app_name = 'Focus Android Glean BrowserStack'

wwyc (approved):
I'm curious what other app_name (s) are expected for `focus_android`.  It looks like the mobile_query has [filters](https://github.com/mozilla/bigquery-etl/blob/10dafcde0ee97ffa929b78b331cf7f4fd1cdf69a/sql_generators/active_users/templates/mobile_query.sql#L46) for app_name and I wonder if it would have any impact. 

Left one comment, otherwise lgtm!

sql_generators/active_users/templates/focus_android_view.sql
gkabbz: Given that this is a view update @wwyc let's merge and I've taken a note to ask Lucia about this when she is back.

sql_generators/active_users/templates/view.sql
```diff
@@ -3,7 +3,8 @@ CREATE OR REPLACE VIEW
   `{{ project_id }}.{{ app_name }}.active_users_aggregates`
 AS
 SELECT
-  * EXCEPT (app_version),
+  * EXCEPT (app_version, app_name),
```
kik-kik, line 6: We exclude `app_name` and then include it?

sql_generators/active_users/templates/focus_android_view.sql
lucia-vargas-a: No, BrowserStack needs to be separated because is bot generated clients.

sql_generators/active_users/templates/view.sql
lucia-vargas-a: This is to enable the UNIONED view of desktop + mobile.

PR #3691, opened 2023-03-30
Jinja queries support

scholtzan:
Reverts https://github.com/mozilla/bigquery-etl/pull/3689

The issues that caused queries to fail are fixed via https://github.com/mozilla/bigquery-etl/pull/3691/commits/bf8c51810f4cdfcb3d36288b041f17e8990ca1b9

Once merged I'll rerun some of the queries in Airflow that failed yesterday, just to make sure it's also running on Airflow

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -861,17 +861,21 @@ def _run_query(
             # point to a public table it needs to be passed as parameter for the query
             query_arguments.append("--destination_table={}".format(destination_table))
 
-        query_arguments.append(
-            render_template(
-                query_file.name,
-                template_folder=str(query_file.parent),
-                templates_dir="",
-                format=False,
+        # write rendered query to a temporary file;
+        # query string cannot be passed directly to bq as SQL comments will be interpreted as CLI arguments
+        with tempfile.NamedTemporaryFile(mode="w+") as query_stream:
+            query_stream.write(
+                render_template(
+                    query_file.name,
+                    template_folder=str(query_file.parent),
+                    templates_dir="",
+                    format=False,
+                )
             )
-        )
+            query_stream.seek(0)
 
-        # run the query as shell command so that passed parameters can be used as is
-        subprocess.check_call(["bq"] + query_arguments)
+            # run the query as shell command so that passed parameters can be used as is
+            subprocess.check_call(["bq"] + query_arguments, stdin=query_stream)
 
 
 @query.command(
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -52,7 +52,7 @@ from ..query_scheduling.generate_airflow_dags import get_dags
 from ..schema import SCHEMA_FILE, Schema
 from ..util import extract_from_query_path
 from ..util.bigquery_id import sql_table_id
-from ..util.common import random_str
+from ..util.common import random_str, render
 from .dryrun import dryrun
 from .generate import generate_all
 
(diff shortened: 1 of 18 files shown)


bigquery_etl/cli/query.py
```diff
@@ -860,7 +861,20 @@ def _run_query(
...
+            query_stream.write(
+                render_template(
+                    query_file.name,
+                    template_folder=str(query_file.parent),
+                    templates_dir="",
+                    format=False,
+                )
+            )
+
+        with open(tmp.name) as query_stream:
```
ANich, line 877:
```suggestion
            query_stream.seek(0)
```

kik-kik:
Are we able to run this in some kind of dev environment prior to merging to make sure we don't have mass failure again tomorrow morning?

bigquery_etl/cli/query.py
```diff
@@ -860,7 +861,20 @@ def _run_query(
             # point to a public table it needs to be passed as parameter for the query
             query_arguments.append("--destination_table={}".format(destination_table))
 
-        with open(query_file) as query_stream:
+        # write rendered query to a temporary file;
+        # query string cannot be passed directly to bq as SQL comments will be interpreted as CLI arguments
+        tmp = tempfile.NamedTemporaryFile()
+        with open(tmp.name, "w") as query_stream:
```
ANich, line 867:
```suggestion
        with tempfile.NamedTemporaryFile(prefix=query_file.name, mode="w+") as query_stream:
```

scholtzan:
We do run some queries through the `integration` test CI step, however it wasn't running any queries with comments... We can't run queries directly since CI doesn't have access to shared-prod

tests/test_run_query.py
```diff
@@ -34,9 +34,12 @@ def test_run_query(self, tmp_path):
...
 
             assert mock_call.call_args.args == (
-                ["bq", "--dataset_id=test", "--destination_table=query_v1"],
+                [
+                    "bq",
+                    "--dataset_id=test",
+                    "--destination_table=query_v1",
+                ],
             )
-            assert "stdin" in mock_call.call_args.kwargs
```
ANich, line 39: We're back to using `stdin` should we re-add these?

bigquery_etl/cli/query.py
```diff
@@ -860,7 +861,20 @@ def _run_query(
             # point to a public table it needs to be passed as parameter for the query
             query_arguments.append("--destination_table={}".format(destination_table))
 
-        with open(query_file) as query_stream:
+        # write rendered query to a temporary file;
+        # query string cannot be passed directly to bq as SQL comments will be interpreted as CLI arguments
+        tmp = tempfile.NamedTemporaryFile()
+        with tempfile.NamedTemporaryFile(prefix=tmp.name, mode="w+") as query_stream:
```
ANich, line 867: Do we still need `tmp` here? We could either omit the `prefix` or make it something meaningful for potential debugging

ANich (approved):
Left a few comments/suggestions but overall lgtm. It feels like there's some overlap in functionality now between SQL generators and "regular" queries though. This might help us remove some simpler generators

sql/mozfun/hist/string_to_json/udf.sql
```diff
@@ -1,3 +1,4 @@
+{% raw %}
```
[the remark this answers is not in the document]
scholtzan: It has some tests that have `{%s}` in some strings. Jinja is trying to interpret those and is failing

bigquery_etl/cli/query.py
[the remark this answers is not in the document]
sean-rose: There's still a question about text vs binary mode, but that's probably unlikely to bite us, as the default locale encoding should pretty much always be UTF-8.

PR #3693, opened 2023-03-30
Jinja related improvements

scholtzan:
Follow-up to https://github.com/mozilla/bigquery-etl/pull/3691#pullrequestreview-1365513402

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -1227,16 +1227,16 @@ def render(name, sql_dir, output_dir):
     if name is None:
         name = "*.*"
 
-    query_files = paths_matching_name_pattern(name, sql_dir, None)
+    query_files = paths_matching_name_pattern(name, sql_dir, project_id=None)
+    resolved_sql_dir = Path(sql_dir).resolve()
     for query_file in query_files:
         rendered_sql = render_template(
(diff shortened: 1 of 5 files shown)


bigquery_etl/cli/query.py
```diff
@@ -1227,17 +1227,16 @@ def render(name, sql_dir, output_dir):
...
-    query_files = paths_matching_name_pattern(name, sql_dir, None)
+    query_files = paths_matching_name_pattern(name, sql_dir, project_id=None)
+    sql_dir = sql_dir.resolve()
     for query_file in query_files:
         rendered_sql = render_template(
             query_file.name, template_folder=query_file.parent, templates_dir=""
         )
 
         if output_dir:
             sql_dir = Path(sql_dir)
```
sean-rose, line 1238: Need to convert it to `Path` first.
```suggestion
    sql_dir = Path(sql_dir).resolve()
    for query_file in query_files:
        rendered_sql = render_template(
            query_file.name, template_folder=query_file.parent, templates_dir=""
        )

        if output_dir:
```
I also sometimes worry about changing passed-in arguments too drastically, so if it were me I might use a separate variable like `resolved_sql_dir`.

sean-rose (approved):
r+wc

PR #3696, opened 2023-03-31
Metric-hub integration

scholtzan:
Depends on https://github.com/mozilla/metric-config-parser/pull/209

This adds a `metrics.calculate()` function to the Jinja context which allows to get the SQL for referenced metrics and use them in SQL queries.

`metrics.calculate()` is used here to rewrite the SSL ratios query to use available metrics: https://github.com/mozilla/bigquery-etl/pull/3698/files#r1156513857

Docs are under way here: https://github.com/mozilla/bigquery-etl/pull/3697 and will also be added to dtmo

### DIFF
--- a/bigquery_etl/metrics.py
+++ b/bigquery_etl/metrics.py
@@ -1,0 +1,33 @@
+"""Metric-hub integration."""
+
+from typing import Dict, List, Optional, Union
+
+import attr
+from metric_config_parser.config import ConfigCollection
+
+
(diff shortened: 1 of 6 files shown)


mikewilli (approved):
Looks good to me! I was thinking about whether it made sense to put the `MetricHub` class somewhere else (somewhere shared like metric-config-parser? mozanalysis?), but we can handle that down the line if it turns out to more broadly useful.

PR #3706, opened 2023-04-05
[Bug 1826618] bqetl status checks

scholtzan:
This implements two tasks that run every 30m to detect bqetl breakages faster. See https://bugzilla.mozilla.org/show_bug.cgi?id=1826618


sql/moz-fx-data-shared-prod/monitoring_derived/bigquery_etl_sql_run_check_v1/query.sql
```diff
@@ -0,0 +1,4 @@
+SELECT
+  CURRENT_TIMESTAMP() AS last_run_timestamp,
+  @task_instance AS task_instance,
+  @run_id AS run_id
```
scholtzan, line 4: Might be useful for debugging to know what time and Airflow task was last successful.

dags/bqetl_status_check.py
```diff
@@ -0,0 +1,89 @@
...
+
+    monitoring_derived__bigquery_etl_sql_run_check__v1 = bigquery_etl_query(
+        task_id="monitoring_derived__bigquery_etl_sql_run_check__v1",
+        destination_table="bigquery_etl_sql_run_check_v1",
+        dataset_id="monitoring_derived",
+        project_id="moz-fx-data-shared-prod",
+        owner="ascholtz@mozilla.com",
+        email=[
+            "ascholtz@mozilla.com",
+            "data-platform-infra-wg@mozilla.com",
```
scholtzan, line 79: Do we want to send emails to the work group? Or is telemetry-alerts enough?

sean-rose: I'd just send to telemetry-alerts and not data-platform-infra-wg, to keep the WG list as noise-free as possible (e.g. this could end up being noisy if there's a problem that takes a while to fix).

sql/moz-fx-data-shared-prod/monitoring_derived/bigquery_etl_sql_run_check_v1/query.sql
sean-rose: Makes sense.  I'd suggest just calling it `run_timestamp` (corollary to `run_id`).

sql/moz-fx-data-shared-prod/monitoring_derived/bigquery_etl_sql_run_check_v1/metadata.yaml
```diff
@@ -0,0 +1,15 @@
...
+description: |-
+  This query is scheduled frequently and used to detect any breakages in the
+  bigquery-etl tooling as soon as possible to ensure nightly ETL tasks can run.
+owners:
+- ascholtz@mozilla.com
+labels:
+  incremental: false
+scheduling:
+  dag_name: bqetl_status_check
+  date_partition_parameter: null
```
sean-rose, line 11: Might be worth a comment here explaining we're intentionally overwriting the entire table each time and keeping just the latest record.

sean-rose (approved):
r+wc

dags.yaml
```diff
@@ -1018,3 +1018,28 @@ bqetl_mdn_yari:
...
+  description: |
+    This DAG checks if bigquery-etl is working properly. Dummy ETL tasks are executed to detect
+    breakages as soon as possible.
+
+    *Triage notes*
+
+    None of these tasks should fail. If they do it is very likely that other/all ETL tasks will
+    subsequently fail as well. Any failures should be communicated to the Data Infra Working Group
+    as soon as possible.
+  schedule_interval: "30m"
```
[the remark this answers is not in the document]
sean-rose: Or alternatively, I'd be less worried about alerts happening over the weekend if this was a bit lower frequency (e.g. hourly).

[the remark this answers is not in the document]
scholtzan: I think during weekends is fine. Someone _happens_ to check during that time they'd know early on. Plus if something is failing, we'd be getting a thousand emails from all the other tasks already anyway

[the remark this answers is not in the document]
scholtzan: Yeah, we can set it to hourly for now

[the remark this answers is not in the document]
sean-rose: Thanks.  That also gives us a bit more leeway when a breakage happens to get a fix PR approved and deployed before the next check.

PR #3713, opened 2023-04-11
feat(): small tweaks to firefox_ios.new_profile_activation table

kik-kik:
feat(): small tweaks to firefox_ios.new_profile_activation table


sql/moz-fx-data-shared-prod/firefox_ios_derived/new_profile_activation_v1/query.sql
```diff
@@ -49,11 +49,11 @@ dou AS (
       mozfun.bits28.to_dates(mozfun.bits28.range(days_seen_bits, -5, 6), submission_date)
     ) AS days_2_7,
   FROM
-    `moz-fx-data-shared-prod.firefox_ios.baseline_clients_last_seen`
+    firefox_ios.baseline_clients_last_seen
   WHERE
     submission_date = @submission_date
     AND DATE_DIFF(submission_date, first_seen_date, DAY) = 6
-    AND normalized_channel = 'release'
+    AND normalized_channel = 'release'  -- # TODO: should we be filtering for a single channel here?
```
fbertsch, line 56: +1, let's include all channels

sql/moz-fx-data-shared-prod/firefox_ios_derived/new_profile_activation_v1/query.sql
```diff
@@ -70,8 +70,11 @@ SELECT
   country,
   first_seen_date,
   submission_date,
-  1 AS new_profile,
-  CAST(days_2_7 > 1 AND COALESCE(search_count, 0) > 0 AS integer) AS activated
+  1 AS new_profile,  -- # TODO: should this be a boolean value? Also, if this is always 1 should we just assume an entry in this table means new_profile?
```
fbertsch, line 73: I strongly prefer booleans. Counting is then `COUNTIF(is_new_profile)`, over `SUM(new_profile)`

sql/moz-fx-data-shared-prod/firefox_ios_derived/new_profile_activation_v1/query.sql
```diff
@@ -39,7 +39,7 @@ client_search AS (
     AND normalized_app_name = 'Fennec'
```
fbertsch, line 39: Wow, we really have to change this string for `normalized_app_name`

sql/moz-fx-data-shared-prod/firefox_ios_derived/new_profile_activation_v1/query.sql
bani: What is your reasoning for the boolean preference?
In terms of understanding the column as a metric instead of a dimension, I have a preference for numeric fields. It also makes it more straightforward to add to Looker.

fbertsch: Definitely personal preference rules here. But consistency matters, and we have largely used boolean over numeric. There's a few ancillary benefits (less storage overhead, schema limits to yes/no/null, easier to understand what the field represents), but in general we should stick with one approach for all of our tables.

bani: So here we can remove the new_profile column, as it was only intended as a shortcut to make it a metric and avoid people having to check if this table only contained new profiles, and convert activated to is_activated.

fbertsch: @kik-kik you should confirm this with the AERO team. Not sure if ya'll have standards / guidance for data modeling that we can reference, but it would be helpful for situations like this

kik-kik:
@bani so we'd only want to have  a `client_id` in this table?

sql/moz-fx-data-shared-prod/firefox_ios_derived/new_profile_activation_v1/query.sql
kik-kik: I'm not sure if filtering just for iOS is enough here or if we also need to filter for a specific app_name here?

kik-kik:
So technically, I am not changing the schema of the table. But considering we are making a change to include data for different release channels it would probably make sense to recreate this table and since we already have to do that I think it could be a perfectly reasonable time to change the structure of the table and only include the fields we need.

In this particular case, we could rename the table to `v2` and deprecate `v1`.

sql/moz-fx-data-shared-prod/firefox_ios_derived/new_profile_activation_v1/query.sql
bani: For added context, another reason I initially chose numeric was that this table was replacing https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/telemetry_derived/firefox_nondesktop_day_2_7_activation_v1/query.sql#L8-L12 - which was pre-aggregated and therefore numeric and queried with SUM, but yes it'd be great to hear what other people think and make an official standard out of it.

sql/moz-fx-data-shared-prod/firefox_ios_derived/new_profile_activation_v1/query.sql
bani: Given that eventually we have a join on client_id and those are "probably" unique it shouldn't give us wrong numbers, but we do have other iOS apps on the table (Klar, Focus, etc) so if we don't filter it's technically a less efficient join later.

fbertsch:
> So technically, I am not changing the schema of the table. But considering we are making a change to include data for different release channels it would probably make sense to recreate this table and since we already have to do that I think it could be a perfectly reasonable time to change the structure of the table and only include the fields we need.
> 
> In this particular case, we could rename the table to `v2` and deprecate `v1`.

Agreed with this plan, and it would be good to simplify the schema if possible.

sql/moz-fx-data-shared-prod/firefox_ios/new_profile_activation/view.sql
```diff
@@ -1,7 +1,7 @@
 CREATE OR REPLACE VIEW
-  `moz-fx-data-shared-prod.firefox_ios.new_profile_activation`
+  firefox_ios.new_profile_activation
 AS
 SELECT
   *
 FROM
-  `moz-fx-data-shared-prod.firefox_ios_derived.new_profile_activation_v1`
+  firefox_ios_derived.new_profile_activation_v1
```
kik-kik, line 7: Once the table is generated and the backfill catches up this will get changed to v2. Could be done during the v1 deprecation process.

sql/moz-fx-data-shared-prod/firefox_ios_derived/new_profile_activation_v1/query.sql
kik-kik: So what is still unclear to me, should we be filtering for a specific app here or are we doing a bag consisting of everything ios?

fbertsch: Just a single app. This should only be Firefox for iOS, not Focus for iOS.

sql/moz-fx-data-shared-prod/firefox_ios_derived/new_profile_activation_v2/query.sql
```diff
@@ -0,0 +1,60 @@
...
+    ARRAY_LENGTH(
+      mozfun.bits28.to_dates(mozfun.bits28.range(days_seen_bits, -5, 6), submission_date)
+    ) AS days_2_7,
+  FROM
+    firefox_ios.baseline_clients_last_seen
+  WHERE
+    submission_date = @submission_date
+    AND DATE_DIFF(submission_date, first_seen_date, DAY) = 6
+),
+client_first_seen AS (
```
fbertsch, line 20: We have `first_seen_date` in `baseline_clients_last_seen`, I believe we can drop this and just get it from there

sql/moz-fx-data-shared-prod/firefox_ios_derived/new_profile_activation_v2/query.sql
```diff
@@ -0,0 +1,60 @@
...
+    SUM(search_count) AS search_count
+  FROM
+    search_derived.mobile_search_clients_daily_v1
+  JOIN
+    client_first_seen
+  USING
+    (client_id)
+  WHERE
+    (submission_date BETWEEN DATE_SUB(@submission_date, INTERVAL 3 DAY) AND @submission_date)
+    -- AND normalized_app_name = 'Fennec'  -- # TODO: should this be filtering for a specific app or is os filter enough here?
```
fbertsch, line 41: I would include the filter, just to clarify the query and slightly improve performance. (it's not strictly necessary since we're joining on `client_id`)

fbertsch (approved):
r+wc. Logic here will work, just a few improvements

sql/moz-fx-data-shared-prod/firefox_ios_derived/new_profile_activation_v1/query.sql
kik-kik: ok, I added back the filter for Fennec

sql/moz-fx-data-shared-prod/firefox_ios_derived/new_profile_activation_v2/query.sql
```diff
@@ -0,0 +1,61 @@
...
+  WHERE
+    submission_date = @submission_date
+    AND DATE_DIFF(submission_date, first_seen_date, DAY) = 6
+),
+client_first_seen AS (
+  SELECT
+    client_id,
+    first_seen_date,
+  FROM
+    firefox_ios.baseline_clients_first_seen
```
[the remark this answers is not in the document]
fbertsch: Yup, right

sql/moz-fx-data-shared-prod/firefox_ios_derived/new_profile_activation_v2/query.sql
```diff
@@ -0,0 +1,61 @@
...
+  client_first_seen
+USING
+  (client_id)
+LEFT JOIN
+  client_search
+USING
+  (client_id)
+-- filter for users that activated
+WHERE
+  IF(days_2_7 > 1 AND COALESCE(search_count, 0) > 0, TRUE, FALSE)
```
fbertsch, line 61: I realized a big downside here with only including activated users, is it's hard to figure out if a client is either not activated or if they're simply not eligible for activation yet (i.e. not old enough). Given that I think we should actually include all clients here

fbertsch (changes requested):

sql/moz-fx-data-shared-prod/firefox_ios_derived/new_profile_activation_v2/query.sql
kik-kik: gotcha, making the change now.

sql/moz-fx-data-shared-prod/firefox_ios_derived/new_profile_activation_v2/query.sql
```diff
@@ -0,0 +1,47 @@
...
+    search_derived.mobile_search_clients_daily_v1
+  WHERE
+    (submission_date BETWEEN DATE_SUB(@submission_date, INTERVAL 3 DAY) AND @submission_date)
+    AND os = 'iOS'
+    AND normalized_app_name = 'Fennec'
+  GROUP BY
+    client_id
+)
+SELECT
+  @submission_date AS `date`,
```
fbertsch, line 36: This is `date`, but below in the schema it's `activation_date`. I think I prefer the latter

fbertsch (approved):
r+wc

sql/moz-fx-data-shared-prod/firefox_ios_derived/new_profile_activation_v2/schema.yaml
```diff
@@ -0,0 +1,31 @@
...
+  name: first_seen_date
+  type: DATE
+  description: |
+    Date when the app first reported a baseline ping for the client.
+
+# - mode: NULLABLE
+#   name: is_activated
+#   type: BOOLEAN
+#   description: |
+#     Determines if a client is activated based on the activation metric and a 7 day lag.
```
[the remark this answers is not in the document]
kik-kik: updated.

PR #3716, opened 2023-04-12
DENG-775 downloads_with_attribution_v2

gleonard-m:
Added session_id to JOIN between GA data and stub_attr.stdout.  Also expanded date range on GA session data to [download_date - 2 days, download_date + 1 day]

Checklist for reviewer:

1. See https://mozilla-hub.atlassian.net/browse/DENG-775
2. This code change builds on https://github.com/mozilla/bigquery-etl/pull/3686 and generates downloads_with_attribution_v2
3. The query now uses the custom session_id GA event to more accurately join the GA data and stub_attribution download data.
4. The joining of the stub attribution data has been expanded to include the 2 days prior and 1 day after the corresponding GA session data.
5. Tests are included in tests/sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2


sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,367 @@
...
+  CASE
+    WHEN nrows > 1
+      THEN NULL
+    ELSE browser
+  END
+  browser,
+  CASE
+    WHEN nrows > 1
+      THEN NULL
+    ELSE normalize_browser(browser)
```
fbertsch, line 297: Can we normalize in the view, and keep the raw value here (then we can later add e.g. opera, brave)

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,367 @@
...
+      ELSE NULL
+    END AS download_session_id
+  FROM
+    `moz-fx-data-marketing-prod.65789850.ga_sessions_*` AS ga
+  CROSS JOIN
+    UNNEST(hits) AS hit
+  WHERE
+    _TABLE_SUFFIX
+    BETWEEN FORMAT_DATE('%Y%m%d', DATE_SUB(@download_date, INTERVAL 2 DAY))
+    AND FORMAT_DATE('%Y%m%d', DATE_ADD(@download_date, INTERVAL 1 DAY))
```
fbertsch, line 51: Does this indicates that _no_ ga hits come in on the same day as the `session_id` presents itself in the stub attribution service logs?

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,367 @@
...
+    hit_type = 'PAGE'
+  GROUP BY
+    client_id,
+    visit_id
+),
+hits_fields AS (
+  SELECT
+    client_id AS client_id,
+    visit_id AS visit_id,
+    mozfun.stats.mode_last(ARRAY_AGG(download_session_id)) AS download_session_id,
```
fbertsch, line 71: nit: needs order by to be idempotent (these and all others)

fbertsch: This should `IGNORE NULL`s, right? Since most hits don't have `download_session_id`?

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,367 @@
...
+),
+stub_other_dl AS (
+  SELECT
+    stub_visit_id,
+    stub_download_session_id,
+    CASE
+      WHEN (COUNT(*) > 1)
+        THEN TRUE
+      ELSE FALSE
+    END AS additional_download_occurred
```
fbertsch, line 148: I don't understand how this is different than `count_dltoken_duplicates`, and why they get joined back together in `stub`

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,367 @@
...
+  FROM
+    stub_dl
+  GROUP BY
+    stub_visit_id,
+    stub_download_session_id
+),
+stub AS (
```
fbertsch, line 155: nit: some CTE names could be more descriptive (e.g. grain is useful when reading the query)

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,367 @@
...
+    mozfun.stats.mode_last_retain_nulls(ARRAY_AGG(`language`)) AS `language`,
+    s.stub_visit_id AS stub_visit_id,
+    s.stub_download_session_id AS stub_download_session_id,
+    s.dltoken AS dltoken,
+    mozfun.stats.mode_last_retain_nulls(ARRAY_AGG(landing_page)) AS landing_page,
+    mozfun.stats.mode_last_retain_nulls(ARRAY_AGG(pageviews)) AS pageviews,
+    mozfun.stats.mode_last_retain_nulls(ARRAY_AGG(unique_pageviews)) AS unique_pageviews,
+    LOGICAL_OR(
+      has_ga_download_event
+    ) AS has_ga_download_event,  -- this will be ignored if nrows >1
```
fbertsch, line 195: Is this comment still true?

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,367 @@
...
+    stub_download_session_id
+)
+SELECT
+  dltoken,
+  download_date,
+  CASE
+    WHEN nrows > 1
+      THEN NULL
+    ELSE time_on_site
+  END
```
fbertsch, line 222: nit: This could be simpler with e.g. `IF(nrows <= 1, itme_on_site, NULL)`

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,367 @@
...
+  USING
+    (client_id, visit_id)
+),
+stub_dl AS (
+  SELECT
+    s.jsonPayload.fields.visit_id AS stub_visit_id,
+    jsonPayload.fields.session_id AS stub_download_session_id,
+    s.jsonPayload.fields.dltoken AS dltoken,
+    (COUNT(*) - 1) AS count_dltoken_duplicates,
+    DATE(s.timestamp) AS download_date
```
fbertsch, line 128: nit: this doesn't need to be in the `GROUP BY`
```suggestion
    @download_date AS download_date
```

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,367 @@
...
+    mozfun.stats.mode_last_retain_nulls(ARRAY_AGG(pageviews)) AS pageviews,
+    mozfun.stats.mode_last_retain_nulls(ARRAY_AGG(unique_pageviews)) AS unique_pageviews,
+    LOGICAL_OR(
+      has_ga_download_event
+    ) AS has_ga_download_event,  -- this will be ignored if nrows >1
+    mozfun.stats.mode_last_retain_nulls(
+      ARRAY_AGG(count_dltoken_duplicates)
+    ) AS count_dltoken_duplicates,
+    LOGICAL_OR(additional_download_occurred) AS additional_download_occurred,
+    COUNT(*) AS nrows,
```
fbertsch, line 200: This count is different than v1, now it should be any dl_token / session_id / client_id dupes. Is that happening often?

fbertsch (commented):
Some initial comments, a lot around readability

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
gleonard-m: > nit: needs order by to be idempotent (these and all others)

You are referring to the `client_id` and `visit_id` at lines 63,64?

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
gleonard-m: You are asking about the difference between `count_dltoken_duplicates` and `additional_download_occurred`?

`additional_download_occurred': This column represents if there was at least one other download completed during the same session as this download. e.g.  the user downloaded more than once in the session.

`count_dltoken_duplicates`: This field is intended for diagnostics.  The dltoken should be unique however occasionally duplicates may be present in the raw data. 

They are joined back together in`stub` since they are determined per dl_token value.

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
gleonard-m: No it is not, however I am not certain that it will never happen.

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
gleonard-m: The raw value is maintained in the table as well.  There is no view currently available due to the changes in how the data is processed from v1 to v2.  The user of the table needs to know if the data is v1 or v2 to understand the results.

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
gleonard-m: > This should `IGNORE NULL`s, right? Since most hits don't have `download_session_id`?

If I add `download_session_id IS NOT NULL` then the hit rows indicating if a download event occurred will be removed resulting in `has_ga_download_event` being set to false.  I will add the filter `hit_type = 'EVENT' ` which will not drop the needed hit rows..

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
gleonard-m: In general the comment is not relevant.  I'll remove it.

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
gleonard-m: Yep - will update

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
gleonard-m: filter `hit_type = 'EVENT' ` has been added.

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
gleonard-m: I've updated the CTE names and added more comments to explain the purpose of the CTEs

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
gleonard-m: Removed.

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
gleonard-m: PR updated.

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
gleonard-m: Fixed.

SuYoungHong (changes requested):
Hi Glenda, thanks for your patience. 

Notebook reference: 
[ETL Review downloads_with_attribution_v2 (DENG-775)](https://colab.research.google.com/drive/19xJRLEeM_K9aIegRVJJpBHX84RU18qSi#scrollTo=lFOjGvR7LHmd)

Overall, looks good, except for a few comments and one change request: 

1: [Line 75](https://github.com/mozilla/bigquery-etl/blob/afd3c5a9a6a5da7265f9afbdd2e9d2e93aaa89d4/sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql#L75)
```
mozfun.stats.mode_last(ARRAY_AGG(download_session_id)) AS download_session_id,
```
Here, you're taking an arbitrary value from one of the GA row's download_session_id. I checked the data, there do seem to be cases of a single GA session having multiple unique download_session_ids. So this could result in some loss (i.e. when we roll up the GA hits into session rows, we keep the non-matching download_session_id and throw out the matching download_session_id). 

But this seems pretty rare. I ran some example data, and it looks like we go from around a 2.95% match rate (of DLs that have valid session id + client ids) to 3.00% match rate, if we *don't* throw out those cases.

I don't think it's a problem, but we should be aware of it. 

2: [Line 266-287](https://github.com/mozilla/bigquery-etl/blob/afd3c5a9a6a5da7265f9afbdd2e9d2e93aaa89d4/sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql#L266)
```
 CASE
    WHEN stub_visit_id IS NULL
      OR stub_download_session_id IS NULL
      THEN 'DOWNLOAD_SESSION_ID_NULL'
    WHEN stub_visit_id = ''
      OR stub_download_session_id = ''
      THEN 'DOWNLOAD_SESSION_ID_EMPTY'
    WHEN stub_visit_id = '(not set)'
      OR stub_download_session_id = '(not set)'
      THEN 'DOWNLOAD_SESSION_ID_VALUE_NOT_SET'
    WHEN stub_visit_id = 'something'
      OR stub_download_session_id = 'something'
      THEN 'DOWNLOAD_SESSION_ID_VALUE_SOMETHING'
    WHEN client_id IS NULL
      THEN 'MISSING_GA_CLIENT'
    WHEN dltoken IS NULL
      THEN 'MISSING_DL_TOKEN'
    WHEN nrows > 1
      THEN 'GA_UNRESOLVABLE'
    ELSE NULL
  END
  `exception`
```

For this exception block, I feel we should clean up the verbiage a little bit. 
* For the `stub_visit_id IS <> OR stub_download_session_id IS <>` patterns, can we describe these as "DOWNLOAD_CLIENT_OR_SESSION_ID_<>", since that's more accurate? 
    * Also, I want to note - so it does seem like there can be valid session ids present with invalid client ids, and vice versa in the DL logs, but it's so rare, that we can treat them like the same event. Might want to include that as a comment. 
* For `client_id IS NULL`, can we describe as "MISSING_GA_CLIENT_OR_SESSION_ID"
* Also, for `dltoken IS NULL`, this should never happen... are we writing this defensively? Maybe let's include a comment here that we should never actually expect this case. 

3: [Line 382](https://github.com/mozilla/bigquery-etl/blob/afd3c5a9a6a5da7265f9afbdd2e9d2e93aaa89d4/sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql#L382)

```
  CASE
    WHEN stub_visit_id IS NULL
      THEN 'DOWNLOAD_SESSION_ID_NULL'
    WHEN stub_visit_id = ''
      THEN 'DOWNLOAD_SESSION_ID_EMPTY'
    WHEN stub_visit_id = '(not set)'
      THEN 'DOWNLOAD_SESSION_ID_VALUE_NOT_SET'
    WHEN stub_visit_id = 'something'
      THEN 'DOWNLOAD_SESSION_ID_VALUE_SOMETHING'
    WHEN client_id IS NULL
      THEN 'MISSING_GA_CLIENT'
    WHEN dltoken IS NULL
      THEN 'MISSING_DL_TOKEN'
    WHEN nrows > 1
      THEN 'GA_UNRESOLVABLE'
    ELSE NULL
  END
  `exception`
```
* We can probably remove all the cases except for 'MISSING_GA_CLIENT', 'GA_UNRESOLVABLE', and NULL, since they should all have been caught in the V2 exceptions. 

4: Lastly, one major design change - I think we should tier the exceptions column from the v2 processing and the v1 processing. As in separate them into separate columns. 

It would be useful, I think, to preserve some data on how much of the data had matching client id + session id and how much had to go through the client id only processing. 

I think there's a straightforward solution, with something like this in the final query: 

```
SELECT * EXCEPT exception, null AS exception_v2, exception as exception_v1
FROM v1_table
UNION ALL
SELECT * EXCEPT exception, exception AS exception_v2, null as exception_v1
FROM v2_table
WHERE exception != 'MISSING_GA_CLIENT' OR exception is NULL
```

gleonard-m:
@SuYoungHong As we discussed today (referencing your list above):
1. Decided not to make any change
2. Code updated
3. Code updated
4. The exception column has been replaced by `join_result_v2` and `join_result_v1`. 
  - If the join was successful using the v2 version `join_result_v2` is assigned `CLIENT_ID_SESSION_ID_MATCH`
  - If the join required the v1 processing and was successful `join_result_v2`is assigned `MISSING_GA_CLIENT_OR_SESSION_ID` and `join_result_v1` is assigned `CLIENT_ID_ONLY_MATCH`

gleonard-m:
> If the join was successful using the v2 version join_result_v2 is assigned CLIENT_ID_SESSION_ID_MATCH
If the join required the v1 processing and was successful join_result_v2is assigned MISSING_GA_CLIENT_OR_SESSION_ID and join_result_v1 is assigned CLIENT_ID_ONLY_MATCH

Should the text be 'MATCH' since the column name indicates the type of match?

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,404 @@
+CREATE TEMP FUNCTION normalize_browser(browser STRING) AS (
```
kik-kik, line 1: I don't remember if I already asked this, would this UDF be useful anywhere outside of this query?

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,404 @@
...
+-- and hit.eventInfo.eventAction('Firefox Download'| 'Stub Session ID') are
+-- extracted in the following CTEs
+-- Those extracted fields are joined to the GA session level data, one session per day.
+WITH all_hits AS (
+  SELECT
+    clientId AS client_id,
+    visitId AS visit_id,
+    hit.appInfo.landingScreenName AS landing_page,
+    hit.page.pagePath AS pagepath,
+    CASE
```
kik-kik, line 32: Since this appears to be a true or false expression. Would it be easier to read if expressed using `if()`?

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,404 @@
...
+      WHEN (hit.type = 'EVENT' AND hit.eventInfo.eventAction = 'Stub Session ID')
+        THEN hit.eventInfo.eventLabel
+      ELSE NULL
+    END AS download_session_id
+  FROM
+    `moz-fx-data-marketing-prod.65789850.ga_sessions_*` AS ga
+  CROSS JOIN
+    UNNEST(hits) AS hit
+  WHERE
+    _TABLE_SUFFIX
```
kik-kik, line 53: Would this be easier to read?
```
WHERE
PARSE_DATE('%Y%m%d', _TABLE_SUFFIX) BETWEEN 
    BETWEEN DATE_SUB(@download_date, INTERVAL 2 DAY)
    AND DATE_ADD(@download_date, INTERVAL 1 DAY)
```

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,404 @@
...
+    BETWEEN FORMAT_DATE('%Y%m%d', DATE_SUB(@download_date, INTERVAL 2 DAY))
+    AND FORMAT_DATE('%Y%m%d', DATE_ADD(@download_date, INTERVAL 1 DAY))
+),
+page_hits AS (
+  SELECT
+    client_id AS client_id,
```
kik-kik, line 59:
```suggestion
    client_id,
```

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,404 @@
...
+    visit_id AS visit_id,
```
kik-kik, line 60:
```suggestion
    visit_id,
```

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,404 @@
...
+    all_hits
+  WHERE
+    hit_type = 'PAGE'
+  GROUP BY
+    client_id,
+    visit_id
+),
+event_hits AS (
+  SELECT
+    client_id AS client_id,
```
kik-kik, line 73:
```suggestion
    client_id,
```

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,404 @@
...
+    visit_id AS visit_id,
```
kik-kik, line 74:
```suggestion
    visit_id,
```

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,404 @@
...
+    mozfun.stats.mode_last_retain_nulls(ARRAY_AGG(device.deviceCategory)) AS device_category,
+    mozfun.stats.mode_last_retain_nulls(ARRAY_AGG(device.operatingSystem)) AS os,
+    mozfun.stats.mode_last_retain_nulls(ARRAY_AGG(device.browser)) AS browser,
+    mozfun.stats.mode_last_retain_nulls(ARRAY_AGG(device.browserVersion)) AS browser_version,
+    mozfun.stats.mode_last_retain_nulls(ARRAY_AGG(device.language)) AS `language`,
+    IFNULL(mozfun.stats.mode_last_retain_nulls(ARRAY_AGG(totals.timeOnSite)), 0) AS time_on_site
+  FROM
+    `moz-fx-data-marketing-prod.65789850.ga_sessions_*` AS ga
+  WHERE
+    _TABLE_SUFFIX
```
kik-kik, line 108: Would this be easier to read?
```
WHERE
PARSE_DATE('%Y%m%d', _TABLE_SUFFIX) BETWEEN 
    BETWEEN DATE_SUB(@download_date, INTERVAL 2 DAY)
    AND DATE_ADD(@download_date, INTERVAL 1 DAY)
```

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,404 @@
...
+    (client_id, visit_id)
+),
+
+-- Extract all the download rows, de-duping and tracking number of duplicates per download token.
+stub_downloads AS (
+  SELECT
+    s.jsonPayload.fields.visit_id AS stub_visit_id,
+    jsonPayload.fields.session_id AS stub_download_session_id,
+    s.jsonPayload.fields.dltoken AS dltoken,
+    (COUNT(*) - 1) AS count_dltoken_duplicates,
```
kik-kik, line 138: Not quite clear what `COUNT(*) - 1` means or why it's row count - 1?

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,404 @@
...
+    stub_visit_id,
+    stub_download_session_id,
+    dltoken
+),
+
+multiple_downloads_in_session AS (
+  SELECT
+    stub_visit_id,
+    stub_download_session_id,
+    CASE
```
kik-kik, line 155: Could this be expressed as `IF()`?

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,404 @@
...
+  CASE
+    WHEN nrows > 1
+      THEN NULL
+    WHEN os IS NULL
+      THEN NULL
+    WHEN os LIKE 'Macintosh%'
+      THEN 'Mac'  -- these values are coming from GA.
+    ELSE mozfun.norm.os(os)
+  END
+  normalized_os,
```
kik-kik, line 255:
```suggestion
 AS normalized_os,
```

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,404 @@
...
+  join_result_v1
+FROM
+  v1_downloads_with_ga_session
+LEFT JOIN
+  `moz-fx-data-shared-prod.static.country_names_v1` AS cn
+ON
+  cn.name = country
+)
+
+SELECT * EXCEPT (join_result_v1), 'MISSING_GA_CLIENT_OR_SESSION_ID' AS join_result_v2, join_result_v1 as join_result_v1
```
kik-kik, line 399:
```suggestion
SELECT * EXCEPT (join_result_v1), 'MISSING_GA_CLIENT_OR_SESSION_ID' AS join_result_v2, join_result_v1 AS join_result_v1
```

dags.yaml
```diff
@@ -1042,3 +1042,20 @@ bqetl_status_check:
...
+      - telemetry-alerts@mozilla.com
+    end_date: null
+    owner: gleonard@mozilla.com
+    retries: 2
+    retry_delay: 30m
+    start_date: '2023-04-10'
+  schedule_interval: 0 23 * * *
+  tags:
+    - impact/tier_1
+    - repo/bigquery-etl
```
kik-kik, line 1061:
```suggestion
```

if I recall correctly, for the DAGs inside the bqetl repo this tag gets added automatically during the dag generation process.

tests/sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v1/test_exception/expect.ndjson
```diff
@@ -1,20 +1,20 @@
-{"dltoken": "ga_unresolve",  "download_date": "2023-02-14", "count_dltoken_duplicates": 0, "additional_download_occurred": 0, "exception": "GA_UNRESOLVABLE"}
```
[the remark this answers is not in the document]
gleonard-m: Cleanup - switched 0 to false.

dags.yaml
gleonard-m: I followed the instructions here https://mozilla.github.io/bigquery-etl/cookbooks/common_workflows/ to create a new DAG file.  You're suggesting that I delete that tag and it will be re-added when the DAG file is deployed to Airflow?

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
kik-kik: Would it be easier to reason about if it was a boolean field and was called something like `has_dltoken_duplicated` or `contains_dltoken_duplicated`?

dags.yaml
kik-kik: The tag will be added to your DAG definition when you run the `./bqetl dag generate` command

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
gleonard-m: Yep  :-)  I've been in the weeds on this query for so long that I don't see these things.

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,404 @@
...
+  IF(nrows <= 1, ad_content, NULL)  ad_content,
+  IF(nrows <= 1, campaign, NULL)  campaign,
+  IF(nrows <= 1, medium, NULL)  medium,
+  IF(nrows <= 1, source, NULL)  source,
+  IF(nrows <= 1, landing_page, NULL)  landing_page,
+  IF(nrows <= 1, country, NULL)  country,
+  IF(nrows <= 1, cn.code, NULL)  normalized_country_code,
+  IF(nrows <= 1, device_category, NULL)  device_category,
+  IF(nrows <= 1, os, NULL) os,
+  CASE
```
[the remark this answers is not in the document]
gleonard-m: Replace `IF(nrows <= 1, <column>, NULL)` in each line with a function such as 
`nullify_value_if_extra_row(nrows, <column>) ` ?

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/schema.yaml
```diff
@@ -0,0 +1,152 @@
+fields:
+- description: Unique token generated for the Firefox download. Value should never
+    be NULL.
+  mode: NULLABLE
+  name: dltoken
+  type: STRING
+- description: Time visitor on site
```
wwyc, line 7: It may be good to elaborate more on this description.

gleonard-m: Unfortunately I don't any more details on how GA measures the value.

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
wwyc: +1

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
```diff
@@ -0,0 +1,404 @@
...
+          hit.type = 'EVENT'
+          AND hit.eventInfo.eventAction = 'Firefox Download'
+          AND hit.eventInfo.eventCategory IS NOT NULL
+          AND hit.eventInfo.eventLabel LIKE 'Firefox for Desktop%'
+        )
+        THEN TRUE
+      ELSE FALSE
+    END AS has_ga_download_event,
+    hit.type AS hit_type,
+    CASE
```
wwyc, line 43: can also be expressed using if()?

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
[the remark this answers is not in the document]
kik-kik: Sorry, in this case I was referring to the CASE statement below all the IF(nrows <= 1, <column>, NULL).

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/schema.yaml
kik-kik: Would `visit duration on the site` be easier to follow here? The description doesn't appear to be human "readable"? Maybe could also mention if this is in seconds, milliseconds, minutes, ...?

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/query.sql
gleonard-m: Boolean would result in lost information.  We want to know the number of duplicates.

sql/moz-fx-data-marketing-prod/ga_derived/downloads_with_attribution_v2/schema.yaml
gleonard-m: I'll add the units (seconds) and update the text to exactly what GA describes it as.  I do not want to deviate from that description since doing so may lead to a mis-representation of the meaning.

dags.yaml
gleonard-m: I honestly can't remember.  Is there an issue with leaving it as is?

SuYoungHong (approved):

PR #3718, opened 2023-04-14
bug(1826384): add telemetry latest releases

kik-kik:
# bug(1826384): add telemetry latest releases

## Changes:

- Added `major`, `minor`, `patch` fields to the `telemetry.releases` view
- Added `telemetry.releases_latest` view to provide an entry for each day for release and beta channels and what app versions were available that day.

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry/releases_latest/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/telemetry/releases_latest/metadata.yaml
@@ -1,9 +1,10 @@
 friendly_name: Releases Latest
 description: |-
-  This is built on top of telemtry.releases and aims
-  to provide an entry on each day for release and beta channels
-  along with the newest firefox app version being available
-  that day on those channels.
+  This is built on top of telemetry.releases and aims
+  to provide an entry on each day for on each day
+  for the major release channels ("release", "beta", "esr")
+  along with the latest Firefox version released on
+  that day or earlier for those channels.
 
   Firefox release information imported from
   https://product-details.mozilla.org/1.0/firefox.json
--- a/sql/moz-fx-data-shared-prod/telemetry/releases_latest/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/telemetry/releases_latest/metadata.yaml
@@ -1,0 +1,14 @@
+friendly_name: Releases Latest
+description: |-
+  This is built on top of telemtry.releases and aims
+  to provide an entry on each day for release and beta channels
+  along with the newest firefox app version being available
+  that day on those channels.
+
+  Firefox release information imported from
+  https://product-details.mozilla.org/1.0/firefox.json
+
+  For more context, see
+  https://wiki.mozilla.org/Release_Management/Product_details
+owners:
+- kignasiak@mozilla.com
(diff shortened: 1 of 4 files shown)


sql/moz-fx-data-shared-prod/telemetry/releases_latest/metadata.yaml
```diff
@@ -0,0 +1,14 @@
+friendly_name: Releases Latest
+description: |-
+  This is built on top of telemtry.releases and aims
```
ncalexan, line 3: nit: telemetry.releases

sql/moz-fx-data-shared-prod/telemetry/releases_latest/metadata.yaml
```diff
@@ -0,0 +1,14 @@
...
+  to provide an entry on each day for release and beta channels
+  along with the newest firefox app version being available
```
ncalexan, line 5: along with the latest Firefox version released on that day or earlier for those channels.

sql/moz-fx-data-shared-prod/telemetry/releases_latest/view.sql
```diff
@@ -0,0 +1,70 @@
...
+  SELECT
+  "release" AS channel
+  UNION ALL SELECT "beta"
+),
+dates_and_channels AS (
+  SELECT
+    `date`, channel
+  FROM
+    UNNEST(
+      (SELECT GENERATE_DATE_ARRAY(DATE_SUB(CURRENT_DATE, INTERVAL 1 YEAR), CURRENT_DATE, INTERVAL 1 DAY) AS dates)
```
ncalexan, line 11: This seems short-sighted, pun intended.  We have lots of queries that will want more than 1 year of release data, and there's just not that much of it -- I found ~10k rows with my queries.  Why not make it historical?

sql/moz-fx-data-shared-prod/telemetry/releases_latest/metadata.yaml
```diff
@@ -0,0 +1,14 @@
+friendly_name: Releases Latest
+description: |-
+  This is built on top of telemtry.releases and aims
+  to provide an entry on each day for release and beta channels
```
ncalexan, line 4: It appears that you're also handling ESR (which is good).  Can you put the values in double quotes so that it's clear that these are the literal values?  I.e.,
```
on each day for the major release channels ("release", "beta", "esr")
```

scholtzan:
There seems to be an issue in zetasql with using `QUALIFY`, which is why CI is failing: https://github.com/google/zetasql/issues/124 You might need to add a workaround be having a condition that doesn't really to anything, like `WHERE TRUE`

kik-kik:
@scholtzan ah thanks for catching that. I added a `WHERE` clause so hopefully we will not see this error anymore.

sql/moz-fx-data-shared-prod/telemetry/releases_latest/view.sql
```diff
@@ -0,0 +1,89 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.latest_releases`
```
scholtzan, line 2:
```suggestion
  `moz-fx-data-shared-prod.telemetry.releases_latest`
```
Make sure the directory name matches the view name

scholtzan (approved):

sql/moz-fx-data-shared-prod/telemetry/releases_latest/view.sql
kik-kik: oops 😬 good catch. Thanks!

PR #3732, opened 2023-04-24
DENG-774 Test change control.

lucia-vargas-a:
DENG-774. This is a PR to test change control.


lucia-vargas-a:
@whd I created this PR to test the change control process implemented in PR.
The reviewers are not added to the PR and that's likely because this PR doesn't change the query file directly but it updates the template that will be used by the sql_generators to create the final query file.
Would you say that adding the query and metadata templates to the CODEOWNERS is the correct implementation in the case of sql generators or is there something else to consider?

whd:
> Would you say that adding the query and metadata templates to the CODEOWNERS is the correct implementation in the case of sql generators or is there something else to consider?

There might be considerations around the ordering of operations in CI e.g. running https://github.com/mozilla/bigquery-etl/blob/main/.circleci/config.yml#L649 after https://github.com/mozilla/bigquery-etl/blob/main/.circleci/config.yml#L662 or other generation steps might be prudent.

Ultimately, unless you check the generated artifacts into the main branch, it doesn't seem like it would be easy to implement change control on essentially arbitrary generation logic. Manually annotating generators associated with change control might be the simplest option. Perhaps someone with a better understanding of the SQL generator logic can determine if there are some properties of them that make them easier to write automatic checks against, but you would need to do something like:

1. run generator logic, verify change control metadata is valid on generated files
2. check that an entry for CODEOWNERS exists for the paths / scripts involved in generating the added files

For instance, hypothetically in generator logic you could be running `curl | bash` which could result in changes to  generated SQL that aren't subject to change control.

But overall, in the case where we're implementing change control for the first time on tables, as long as generators that require change control map 1:1 to generating tables that require change control, manually annotating CODEOWNERS is a reasonably safe way to achieve that.

Requiring generated files that are change controlled to be checked into `main` is perhaps another option that would be easy to check for. Basically in the verify metadata step, assuming it's run after generation steps, if there are any added files subject to change control that aren't in git history, you could error. I suspect there are some standards in the repo on not checking in generated files though, and depending on how dynamic these generators are (e.g. for every new glean app) that could cause problems in CI. But I think it's worth considering specifically for change controlled tables.

lucia-vargas-a:
@whd do you mean that the change control will only be based on files in the main branch and not in the generated-sql branch?

lucia-vargas-a:
Indeed, [the CODEOWNER file must exist in the specific branch where the path exists](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-file-location).

whd:
At this point the potentially arbitrary generation logic we're using to define resulting artifacts in branches is beyond the cases CODEOWNERS is designed for. However, it's possible to develop bespoke CI to e.g. assign reviewers that aren't CODEOWNERS based on generated output and then fail CI until all or any of those reviewers approve the PR. There may even be something on https://github.com/marketplace?category=code-review&type=actions&query= to work from that does something like this.

lucia-vargas-a:
Tested and validated in [PR-3773](https://github.com/mozilla/bigquery-etl/pull/3773).

PR #3733, opened 2023-04-25
RS 682 : Set up data table for AMP impressions and clicks reporting

alekhyamoz:
Checklist for reviewer:


sql/moz-fx-data-shared-prod/contextual_services_derived/request_payload_suggest_v1/query.sql
```diff
@@ -0,0 +1,48 @@
...
+),
+parsed AS (
+  SELECT
+    si.source,
+    REGEXP_EXTRACT(url, r'form-factor=(\w+)', 1) AS form_factor,
+    CASE
+      WHEN (url LIKE '%impressions%')
+        THEN 'impression'
+      ELSE 'click'
+    END AS send_event,
```
quiiver, line 20: `impressions` are only a part of the url for topsites impressions. using it here would flag suggest impressions as clicks. instead you can use `si.interactionType`

sql/moz-fx-data-shared-prod/contextual_services_derived/request_payload_suggest_v1/query.sql
```diff
@@ -0,0 +1,48 @@
...
+    si.flaggedFraud,
+    DATE(submission_timestamp) AS submission_date,
+    CASE
+      WHEN REGEXP_CONTAINS(url, r'impressions=(\d+)')
+        THEN TIMESTAMP_SECONDS(CAST(REGEXP_EXTRACT(url, r'begin-timestamp=(\d+)') AS INT))
+      ELSE NULL
+    END AS begin_timestamp,
```
quiiver, line 27:
```sql
when 
  then timestamp_seconds(cast(json_value(si.parsedReportingUrl, '$.params.begin-timestamp') as int)
else null
```
```suggestion
    CASE
      WHEN si.source = 'topsites' and si.interactionType = 'impression'
        THEN TIMESTAMP_SECONDS(CAST(JSON_VALUE(si.parsedReporting.params.begin_timestamp) AS INT))
      ELSE NULL
    END AS begin_timestamp,
```

sql/moz-fx-data-shared-prod/contextual_services_derived/request_payload_suggest_v1/query.sql
```diff
@@ -0,0 +1,48 @@
...
+        THEN TIMESTAMP_SECONDS(CAST(REGEXP_EXTRACT(url, r'end-timestamp=(\d+)') AS INT))
+      ELSE NULL
+    END AS end_timestamp,
+    REGEXP_EXTRACT(url, r'country-code=(\w+)', 1) AS country_code,
+    REGEXP_EXTRACT(url, r'region-code=(\w+)', 1) AS region_code,
+    REGEXP_EXTRACT(url, r'os-family=(\w+)', 1) AS os_family,
+    REGEXP_EXTRACT(url, r'product-version=(\w+)', 1) AS product_version,
+    IFNULL(CAST(REGEXP_EXTRACT(url, r'impressions=(\d+)', 1) AS int), 0) AS impression_count,
+    IF(url LIKE '%ctag%', 1, 0) AS click_count,
+    url AS error_message,
```
quiiver, line 39: curious why we're keeping this.

sql/moz-fx-data-shared-prod/contextual_services_derived/request_payload_suggest_v1/query.sql
```diff
@@ -0,0 +1,48 @@
...
+      WHEN REGEXP_CONTAINS(url, r'impressions=(\d+)')
+        THEN TIMESTAMP_SECONDS(CAST(REGEXP_EXTRACT(url, r'end-timestamp=(\d+)') AS INT))
+      ELSE NULL
+    END AS end_timestamp,
+    REGEXP_EXTRACT(url, r'country-code=(\w+)', 1) AS country_code,
+    REGEXP_EXTRACT(url, r'region-code=(\w+)', 1) AS region_code,
+    REGEXP_EXTRACT(url, r'os-family=(\w+)', 1) AS os_family,
+    REGEXP_EXTRACT(url, r'product-version=(\w+)', 1) AS product_version,
+    IFNULL(CAST(REGEXP_EXTRACT(url, r'impressions=(\d+)', 1) AS int), 0) AS impression_count,
+    IF(url LIKE '%ctag%', 1, 0) AS click_count,
```
quiiver, line 38: because these can never exist at the same time is there a reason we don't generalize this to something like `interaction_count`?

if so then both rows should collapse to the following:

```sql
si.interactionCount as interaction_count
```

quiiver:
Looking good! I left some feedback on the `request_payload_suggest_v1` file. But the feedback applies to the `tiles` table as well. That leads to the following question: why these are split into two separate tables? It seems to me that the simplest thing would be to have a single `request_payload_v1` table and split them at the view level. Especially given that as of right now the table schemas are identical.

sql/moz-fx-data-shared-prod/contextual_services_derived/request_payload_suggest_v1/query.sql
alekhyamoz: @skahmann3 please let me know if `interaction_count` is good enough here.

alekhyamoz:
@quiiver based on the discussion in [the document](https://docs.google.com/document/d/1PO5sU2yGC6F74FLQowcFMwiaGW_DPGvDw7hJ9ev-XOE) the team decided to split the datasets into `tiles` and `suggest`. Though I believe this can be handled in Looker. 
@skahmann3  do you have any preference? If not, I will go by Wil's suggestion to have a single table `request_payload_v1`

quiiver:
> @quiiver based on the discussion in [the document](https://docs.google.com/document/d/1PO5sU2yGC6F74FLQowcFMwiaGW_DPGvDw7hJ9ev-XOE) the team decided to split the datasets into tiles and suggest. Though I believe this can be handled in Looker.

Yeah I think you can keep the data split in the views but share a single base table, especially if they have the same schema. The views would then filter between tiles and suggest:

```sql
CREATE OR REPLACE VIEW
  `moz-fx-data-shared-prod.contextual_services.request_payload_suggest`
AS
SELECT
  *
FROM
  `moz-fx-data-shared-prod.contextual_services_derived.request_payload_v1`
WHERE 
  source = 'quicksuggest'
```

sql/moz-fx-data-shared-prod/contextual_services_derived/request_payload_suggest_v1/query.sql
skahmann3: Let's go ahead and leave out. We can always add more fields from the error_message if necessary later

sql/moz-fx-data-shared-prod/contextual_services_derived/request_payload_suggest_v1/query.sql
skahmann3: Having them separate makes it easier for stakeholders to report on both. We originally had event_aggregates with just "event_count" but ended up adding click and impression counts separately since it was a pain to work with.

skahmann3:
We've been splitting all the contextual_services tables by Tiles vs. Suggest lately since the products are now evolving independently with reporting and billing going in different directions. Product has requested these tables also be split. We're expecting this will be used for direct sales and so partner request info will again start to diverge by product. So let's go ahead and separate, should save us some future work. So we don't need to add the "source" field in the table since they'll already be differentiated

alekhyamoz:
@skahmann3 just to clarify, you are asking the tables themselves to be separate and not just at the `view` level as Wil suggested correct?

skahmann3:
Yes @alekhyamoz that is correct

sql/moz-fx-data-shared-prod/contextual_services_derived/request_payload_suggest_v1/query.sql
```diff
@@ -0,0 +1,45 @@
...
+    AND error_type = 'SendRequest'
+),
+extracted AS (
+  SELECT
+    si.formFactor AS form_factor,
+    si.interactionType AS send_event,
+    si.flaggedFraud,
+    DATE(submission_timestamp) AS submission_date,
+    CASE
+      WHEN si.source = 'topsites'
```
quiiver, line 18: since we're splitting these. `source` will never be `topsites` so this will always be a null value. I would either remove the column from this table or do something like `null as begin_timestamp` with a comment for why it's set to null

sql/moz-fx-data-shared-prod/contextual_services_derived/request_payload_suggest_v1/query.sql
```diff
@@ -0,0 +1,45 @@
...
+        THEN TIMESTAMP_SECONDS(
+            CAST(JSON_VALUE(si.parsedReportingUrl.params.begin_timestamp) AS INT)
+          )
+      ELSE NULL
+    END AS begin_timestamp,
+    CASE
+      WHEN si.source = 'topsites'
+        AND si.interactionType = 'impression'
+        THEN TIMESTAMP_SECONDS(CAST(JSON_VALUE(si.parsedReportingUrl.params.end_timestamp) AS INT))
+      ELSE NULL
```
quiiver, line 29: same as `begin_timestamp`

sql/moz-fx-data-shared-prod/contextual_services_derived/request_payload_suggest_v1/query.sql
```diff
@@ -0,0 +1,45 @@
...
+    `moz-fx-data-shared-prod.payload_bytes_error.contextual_services`
+  WHERE
+    DATE(submission_timestamp) = @submission_date
+    AND error_type = 'SendRequest'
+),
+extracted AS (
+  SELECT
+    si.formFactor AS form_factor,
+    si.interactionType AS send_event,
+    si.flaggedFraud,
```
quiiver, line 15: in the schema there is no column for `flaggedFraud` or `flagged_fraud` so you'll want to add that.

sql/moz-fx-data-shared-prod/contextual_services_derived/request_payload_suggest_v1/schema.yaml
```diff
@@ -0,0 +1,37 @@
...
+  - mode: NULLABLE
+    name: product_version
+    type: STRING
+  - mode: NULLABLE
+    name: impression_count
+    type: INTEGER
+  - mode: NULLABLE
+    name: click_count
+    type: INTEGER
+  - mode: NULLABLE
```
quiiver, line 35: can remove this

sql/moz-fx-data-shared-prod/contextual_services_derived/request_payload_suggest_v1/schema.yaml
```diff
@@ -0,0 +1,37 @@
...
+    type: STRING
+  - mode: NULLABLE
+    name: submission_date
+    type: DATE
+  - mode: NULLABLE
+    name: begin_timestamp
+    type: TIMESTAMP
+  - mode: NULLABLE
+    name: end_timestamp
+    type: TIMESTAMP
```
quiiver, line 16: if you decide to remove these from the query you'll want to also take them out here.

sql/moz-fx-data-shared-prod/contextual_services_derived/request_payload_tiles_v1/query.sql
```diff
@@ -0,0 +1,45 @@
...
+    AND error_type = 'SendRequest'
+),
+extracted AS (
+  SELECT
+    si.formFactor AS form_factor,
+    si.interactionType AS send_event,
+    si.flaggedFraud,
+    DATE(submission_timestamp) AS submission_date,
+    CASE
+      WHEN si.source = 'topsites'
```
quiiver, line 18: `source` is always topsites now so you can remove this and on line 26

sql/moz-fx-data-shared-prod/contextual_services_derived/request_payload_tiles_v1/query.sql
```diff
@@ -0,0 +1,45 @@
...
+    `moz-fx-data-shared-prod.payload_bytes_error.contextual_services`
+  WHERE
+    DATE(submission_timestamp) = @submission_date
+    AND error_type = 'SendRequest'
+),
+extracted AS (
+  SELECT
+    si.formFactor AS form_factor,
+    si.interactionType AS send_event,
+    si.flaggedFraud,
```
quiiver, line 15: also need a fraud column in the schema

sql/moz-fx-data-shared-prod/contextual_services_derived/request_payload_tiles_v1/schema.yaml
```diff
@@ -0,0 +1,37 @@
...
+  - mode: NULLABLE
+    name: product_version
+    type: STRING
+  - mode: NULLABLE
+    name: impression_count
+    type: INTEGER
+  - mode: NULLABLE
+    name: click_count
+    type: INTEGER
+  - mode: NULLABLE
```
quiiver, line 35: can remove this as well

sql/moz-fx-data-shared-prod/udf_js/parse_sponsored_interaction/udf.sql
```diff
@@ -0,0 +1,102 @@
...
+  reportingUrl string, 
+  requestId string, 
+  submissionTimestamp timestamp,
+  parsedReportingUrl json,
+  originalDocType string,
+  originalNamespace string,
+  interactionCount integer,
+  flaggedFraud bool> 
+LANGUAGE js AS """
+    const parseURL = (url) => {
```
quiiver, line 21: for legibility we want to un-indent this line two spaces

sql/moz-fx-data-shared-prod/udf_js/parse_sponsored_interaction/udf.sql
```diff
@@ -0,0 +1,102 @@
...
+  assert.array_equals(null,e.scenario),
+  assert.array_equals("click",e.interactionType),
+  assert.array_equals("{10679079-b1cd-45a3-9e40-cdfb364d3476}",e.contextId),
+  assert.array_equals("https://bridge.sfo1.ap01.net/ctp?ci=1681139740815.12791&country-code=DE&ctag=pd_sl_08aeb79c14ac3da0f8e9116cdcb0afadec2e24da616da802ba033bf6&dma-code=&form-factor=desktop&key=1681139740400900002.1&os-family=Windows&product-version=firefox_111&region-code=NW&version=16.0.0",e.reportingUrl),
+  assert.array_equals(null,e.requestId),
+  assert.array_equals("2023-04-10 15:41:55 UTC",e.submissionTimestamp),
+  assert.array_equals("topsites-click",e.originalDocType),
+  assert.array_equals("contextual-services",e.originalNamespace),
+  assert.array_equals(1,e.interactionCount),
+  assert.array_equals(false,e.flaggedFraud)
```
quiiver, line 99: you might consider a test case for `parsedReportingUrl`

quiiver (approved):
lgtm

sql/moz-fx-data-shared-prod/contextual_services_derived/request_payload_suggest_v1/query.sql
```diff
@@ -0,0 +1,33 @@
...
+    DATE(submission_timestamp) = @submission_date
+    AND error_type = 'SendRequest'
+),
+extracted AS (
+  SELECT
+    si.formFactor AS form_factor,
+    si.interactionType AS send_event,
+    si.flaggedFraud AS flagged_fraud,
+    DATE(submission_timestamp) AS submission_date,
+    NULL AS begin_timestamp, /* we're splitting the tables based on 'source' and source will never be 'topsites' so this will always be a null value */
```
skahmann3, line 17: Let's go ahead and remove the begin_timestamp and end_timestamp for Suggest since they're always null

skahmann3 (approved):
Looks great, let's remove the null begin/end timestamp columns from the suggest table and then should be good to merge. Thanks for your work on this Alekhya and Wil!

PR #3755, opened 2023-05-03
Remove java dependency in favor of sqlglot

relud:
generate-diff CI is expected to fail


Dexterp37:
This rhymes with #3406 and #3400. @relud make sure to peek at least at the former to see why it wasn't merged ;-)

relud:
it wasn't merged because sqlglot couldn't handle certain things, all of which have a workaround in this PR. most of them were fixed upstream yesterday, but a couple of them i still need to file bugs for.

scholtzan (approved):
This is great! 

It looks like the dry run errors are coming from queries that reference views that haven't been deployed yet. Which is related to this unresolved issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1821789

relud:
> It looks like the dry run errors are coming from queries that reference views that haven't been deployed yet. Which is related to this unresolved issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1821789

turns out it was a bug introduced in this PR.

validate-views failure is a new table not yet deployed in prod, and is also failing on main.

PR #3761, opened 2023-05-05
feat(DENG-854?): added fenix app id to the fxa_activity event filter list

kik-kik:
feat(DENG-854?): added fenix app id to the fxa_activity event filter list

It was brought to our attention that we might be getting high volume event data from fenix and that we might want to exclude them from our derived tables. This PR suggests a similar approach as for `firefox-ios` of excluding the highest volume events from making it through.


kik-kik:
we decided that this filtering is wrong and would hinder our analytics. Closing this PR and the related issue has been closed.

PR #3763, opened 2023-05-05
Filter non email owners from labels 

ANich:
In some cases we have owners that are GitHub identities, e.g.: https://github.com/mozilla/bigquery-etl/blob/5b178559b3f7b9de4ed5f12763b685c29d09d67d/sql_generators/active_users/templates/metadata.yaml#L17-L19 

These aren't always valid labels and currently show up as empty strings in BQ, so we should skip them entirely. ~Also we use emails verbatim as owners in dags so it makes sense here too I think.~

### DIFF
--- a/bigquery_etl/metadata/parse_metadata.py
+++ b/bigquery_etl/metadata/parse_metadata.py
@@ -10,7 +10,7 @@ import cattrs
 import yaml
 from google.cloud import bigquery
 
-from bigquery_etl.query_scheduling.utils import is_email_or_github_identity
+from bigquery_etl.query_scheduling.utils import is_email, is_email_or_github_identity
 
 METADATA_FILE = "metadata.yaml"
 DATASET_METADATA_FILE = "dataset_metadata.yaml"
@@ -255,11 +255,12 @@ class Metadata:
 
                 if "owners" in metadata:
                     owners = metadata["owners"]
-                    for i, owner in enumerate(metadata["owners"]):
+                    owner_idx = 1
+                    for owner in filter(is_email, owners):
                         label = owner.split("@")[0]
-                        if not Metadata.is_valid_label(label):
-                            label = ""
-                        labels[f"owner{i+1}"] = label
+                        if Metadata.is_valid_label(label):
+                            labels[f"owner{owner_idx}"] = label
+                            owner_idx += 1
 
                 if "schema" in metadata:
                     converter = cattrs.BaseConverter()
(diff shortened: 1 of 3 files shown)


bigquery_etl/metadata/parse_metadata.py
```diff
@@ -256,10 +261,9 @@ def from_file(cls, metadata_file):
                 if "owners" in metadata:
                     owners = metadata["owners"]
                     for i, owner in enumerate(metadata["owners"]):
-                        label = owner.split("@")[0]
-                        if not Metadata.is_valid_label(label):
-                            label = ""
-                        labels[f"owner{i+1}"] = label
+                        if not is_email(owner):
```
wwyc, line 264: I think emails cannot be labels because "@" is not allowed?
https://cloud.google.com/bigquery/docs/labels-intro#requirements

ANich: Ahh fair enough, I'll update this to just skip non-email labels then

bigquery_etl/metadata/parse_metadata.py
```diff
@@ -166,6 +166,11 @@ def validate_labels(self, attribute, value):
                             raise ValueError(f"Invalid label format: {bug}")
                 else:
                     raise ValueError("Error: review_bugs needs to be a list.")
+            elif re.match(r"owner\d", key):
+                if not is_email(label):
+                    raise ValueError(
```
lucia-vargas-a, line 171: Should we skip Github identities instead of raising an exception? E.g.
if not is_email_or_github_identity(label):

tests/metadata/test_parse_metadata.py
```diff
@@ -84,9 +84,13 @@ def test_from_file(self):
...
-        assert metadata.owners == ["test1@mozilla.com", "test2@example.com"]
-        assert metadata.labels["owner1"] == "test1"
-        assert metadata.labels["owner2"] == "test2"
+        assert metadata.owners == [
+            "test1@mozilla.com",
+            "test2@example.com",
+            "mozilla/test_group",
+        ]
+        assert metadata.labels["owner1"] == "test1@mozilla.com"
+        assert metadata.labels["owner2"] == "test2@example.com"
```
lucia-vargas-a, line 93: Could we add a test to assert the case of owner "mozilla/test_group"? Expects an exception (or None).

lucia-vargas-a (approved):

PR #3765, opened 2023-05-08
Deng 847 change control for generated sql

lucia-vargas-a:
Test changing the view.
Codeowners should be added automatically to the PR.


lucia-vargas-a:
To be created in a fork.

PR #3777, opened 2023-05-11
Skip deploy jobs on non-main branches

ANich:
Skip the entire deploy job on non-main branches instead of just some steps. Also remove the tag condition since we're not doing that anymore

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -425,21 +425,15 @@ jobs:
     executor: ubuntu-machine-executor
     steps:
       - checkout
-      - when:
-          condition: &deploy-condition
-            or:
-              - equal: [main, << pipeline.git.branch >>]
-              - << pipeline.git.tag >>
-          steps:
-            - *attach_generated_sql
-            - *copy_generated_sql
-            - docker/check:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -425,21 +425,15 @@ jobs:
     executor: ubuntu-machine-executor
     steps:
       - checkout
-      - when:
-          condition: &deploy-condition
-            or:
-              - equal: [main, << pipeline.git.branch >>]
-              - << pipeline.git.tag >>
```
kik-kik, line 432: Are we sure we do not want to do a deploy when a new tag is introduced? I guess the question is, do we even tag things?

kik-kik (approved):

.circleci/config.yml
ANich: I'll defer to @relud on this but I think the idea was tagged releases for this repo at some point

relud (approved):

.circleci/config.yml
relud: yeah, this was built with tagging in mind, but we don't use them

PR #3778, opened 2023-05-11
fix #3756: parse inputs and outputs from UDFs for docs

### LINKED ISSUE #3756: `mozfun` UDF documentation should show parameters and return type
One problem with `mozfun` is that it doesn't show what the input or output types are. That makes it hard to use without opening up the UDF code directly.

mikewilli:
Checklist for reviewer:


bigquery_etl/docs/mozfun/generate_mozfun_docs.py
```diff
@@ -88,7 +88,134 @@ def generate_mozfun_docs(out_dir, project_dir):
...
+                            found_return = False
+                            return_lines = []
+                            with open(os.path.join(root, UDF_FILE), "r") as udf_sql:
+                                udf_content = [
+                                    line
+                                    for line in udf_sql.readlines()
+                                    if not line.startswith("#")
+                                ]
+
+                                FUNCTION_HEADER = (
```
lucia-vargas-a, line 103: Shall we use the [funtion to extract input parameters](https://github.com/mozilla/bigquery-etl/blob/3756_function_to_get_UDF_input_output/bigquery_etl/util/udf_functions.py) using regex? The [test scenarios](https://github.com/mozilla/bigquery-etl/blob/3756_function_to_get_UDF_input_output/tests/util/test_udf_functions.py) make it more robust and easy to read the expected headers. If a developer requires a different header then the function can be expanded.

bigquery_etl/docs/mozfun/generate_mozfun_docs.py
```diff
@@ -88,7 +88,134 @@ def generate_mozfun_docs(out_dir, project_dir):
...
+                                    ):
+                                        input_lines.append(line)
+                                    elif found_input and (
+                                        ") AS (" in line or "RETURNS" in line
+                                    ):
+                                        found_input = False
+
+                                    # TODO: find_block function? def find_block(start_word, end_word)
+
+                                    # RETURNS has the output types
```
lucia-vargas-a, line 130: We can extend the [function to extract parameters](https://github.com/mozilla/bigquery-etl/blob/3756_function_to_get_UDF_input_output/bigquery_etl/util/udf_functions.py) to get the output parameters using regex and add [test scenarios](https://github.com/mozilla/bigquery-etl/blob/3756_function_to_get_UDF_input_output/tests/util/test_udf_functions.py).

lucia-vargas-a (changes requested):
@mikewilli to reduce the lines of code and make it more readable we implemented a function and test scenarios to extract the input parameters. This was in our pairing session with @lelilia today. The code is in [this branch](https://github.com/mozilla/bigquery-etl/tree/3756_function_to_get_UDF_input_output) and still requires working in the output parameters. Feel free to merge it if you like the approach.

tests/util/test_udf_functions.py
```diff
@@ -126,6 +123,22 @@ def test_get_output_multiple_matches(self):
...
+                    """
+                    CREATE OR REPLACE FUNCTION norm.get_windows_info(os_version STRING)
+RETURNS STRUCT<name STRING, version_name STRING, version_number DECIMAL, build_number INT64>
+LANGUAGE js AS r\"\"\"
+  const test = "test";
+  return test;
+\"\"\";
+                    """
+                )
+                == "STRUCT<name STRING, version_name STRING, version_number DECIMAL, build_number INT64> LANGUAGE js"
```
lelilia, line 139: I'm not really happy with this test. It's not the result we would want, so I would change the `get_output` function to give us the correct result here

bigquery_etl/util/udf_functions.py
```diff
@@ -19,7 +19,7 @@ def get_input(sql_text: str) -> Optional[str]:
 
 def get_output(sql_text: str) -> Optional[str]:
     """Extract output parameters from UDF sql code."""
-    cleaned_sql_text = re.sub(r"\s{2,}", " ", sql_text)
+    cleaned_sql_text = re.sub(r"\s{2,}|\n", " ", sql_text)
     output_groups = re.search(
         r"(?:RETURNS |CAST\([a-z,A-Z,0-9,_,.]+ AS |, OUT )(.+?)(?: AS|\s?\))",
```
lelilia, line 24:
```suggestion
        r"(?:RETURNS |CAST\([a-z,A-Z,0-9,_,.]+ AS |, OUT )(.+?)(?: AS|\s?\)|\s?LANGUAGE)",
```
that way the edge case you found will give the correct output

bigquery_etl/docs/mozfun/generate_mozfun_docs.py
```diff
@@ -41,6 +44,28 @@ def add_source_and_edit(source_url, edit_url):
...
 
 
+def get_inputs_outputs_from_udf(root: str) -> Tuple[List[str], List[str]]:
+    """Parse UDF SQL file and return inputs and outputs lists."""
+    with open(os.path.join(root, UDF_FILE), "r") as udf_file:
+        input_str, output_str = get_udf_parameters(udf_file.read())
+
+    input_lines = []
+    if input_str is not None:
+        if "<" in input_str and ">" in input_str:
```
lucia-vargas-a, line 54: Since the function get_udf_parameters already returns the formatted parameters, perhaps we can avoid splitting and concatenating again by adding this validation and a test to that function?

lelilia (approved):
1 small change but otherwise it looks good

bigquery_etl/docs/mozfun/generate_mozfun_docs.py
```diff
@@ -88,6 +113,26 @@ def generate_mozfun_docs(out_dir, project_dir):
                             dataset_doc_file.write(f"{formated}\n\n")
                         # Inject the contents of the README.md
                         dataset_doc_file.write(docfile_content)
+                        if is_udf:
+                            input_lines, output_lines = get_inputs_outputs_from_udf(
+                                root
+                            )
+
+                            # assemble, format, write inputs and outputs
+                            if len(input_lines) > 0:
```
lucia-vargas-a, line 122: It's great to have the len() validation. I think we could simplify a lot the code by printing the parameters as returned in get_udf_parameters, (and moving any required logic there), wdyt?

lucia-vargas-a (commented):
Two comments to potentially simplify the code,

bigquery_etl/docs/mozfun/generate_mozfun_docs.py
mikewilli: So I agree that would simplify things here, but I think it depends on how we think the utility function might be used. I thought that doing the formatting here helped keep the utility more generic, but another issue is that I also couldn't immediately think of a solution to the various splitting issues I came across. I'm sure we could update the regex to pull out ARRAY/STRUCT/etc types, but I was trying to balance the time I'd need to spend on that with the priority of the issue. This solution seemed quickest to getting an acceptable (to me, at least) result. Another alternative would be to simply never split out inputs by line, but having one input per line seems nicer to me.

What do you think?

bigquery_etl/util/udf_functions.py
mikewilli: Awesome, committed this suggestion and updated the test, thanks!

bigquery_etl/docs/mozfun/generate_mozfun_docs.py
lelilia: I think I would opt for just using the output you get for now. Because having it sometimes in different lines and sometimes as one line might be confusing / inconsistent. 

If we get bored at some point we could try to tackle the parsing of ARRAY/STRUCT values .. it might be not regex but stacks? somehow every coding puzzle with parentheses always comes back to stacks 

But since I'm not sure how it should look like in the end, I can't think of how to parse it yet. 

So my take would be: let's move the formatting to the utils and have it return a string (with or without line breaks) and if we ever end up implementing it we wouldn't need to touch the `generate_mozfun_docs.py` again? 

I don't think the utils functions are generic enough to be used for anything else. It's very specific regex for exactly that purpose?

mikewilli: Ok, I will update to remove some of the formatting I'm doing now for consistency's sake. I think if the functions are not intended to be generic they may be more appropriately placed either in the `generate_mozfun_docs.py` file or alongside it.

lucia-vargas-a (approved):

PR #3785, opened 2023-05-16
utility functions to get udf inputs and outputs

mikewilli:
@lelilia and @lucia-vargas-a built some functions in support of #3778 that I would like to incorporate in that PR. This PR is to merge those utility functions to make them available.

Checklist for reviewer:


bigquery_etl/util/udf_functions.py
```diff
@@ -0,0 +1,29 @@
...
+    if res is None:
+        return None
+    return res.groups()[0]
+
+
+def get_udf_parameters(sql_text):
+    """Extract input and output parameters from UDF sql code."""
+    input_part = find_input(sql_text)
+    output_part = find_output(sql_text)
+    return {"input": input_part, "output": output_part}
```
mikewilli, line 29: Is it more standard to return a `dict` (as shown here) or a `tuple` (i.e., multiple values)? I think personally I would prefer to return a `tuple`, but I am open to arguments for the `dict`.

bigquery_etl/util/udf_functions.py
```diff
@@ -0,0 +1,29 @@
+"""Utility functions related to UDF parsing and documentation."""
+
+import re
+
+
+def find_input(sql_text):
+    """Extract input parameters from UDF sql code."""
+    sql_text = re.sub(r"\s{2,}|\\n", " ", sql_text)
+    return (
+        re.search(r"FUNCTION [a-z,0-9,_,.]+\s*\(\s*(.+?)\s*\)", sql_text)
```
mikewilli, line 10: Just want to check that function names are guaranteed to be lowercase?

I also think we can narrow the whitespace searching a little bit because we already removed all whitespace of 2+ characters in the line 8 (though I'm not actually sure if this is more efficient performance-wise?):
```suggestion
        re.search(r"FUNCTION [a-z,0-9,_,.]+\s?\(\s?(.+?)\s?\)", sql_text)
```

bigquery_etl/util/udf_functions.py
```diff
@@ -0,0 +1,29 @@
...
+        .groups()[0]
+        .strip()
```
mikewilli, line 12: This shouldn't be necessary, given that we are already stripping whitespace characters with the regex.

bigquery_etl/util/udf_functions.py
mikewilli: To clarify, here is what I am suggesting in the above comment:
```suggestion
    return input_part, output_part
```

mikewilli (commented):
I have a few comments, but overall looks great @lelilia! Let me know what you think of the comments, otherwise :+1: :shipit:

bigquery_etl/util/udf_functions.py
lelilia: we only strip double white spaces ... I added it for `test_find_input_multiline_input` where there is a line break before the closing parathesis

bigquery_etl/util/udf_functions.py
lelilia: They do have comments but I'm looking for a specific string pattern around the group I'm looking for. In the tests it worked with lot's of comments ... we would only run into problems here if someone would comment out an old udf and add a new one underneath I think

bigquery_etl/util/udf_functions.py
lelilia: oh no you are right ... it doesn't break any more ... thanks for catching that .. must have been a leftover from a previous iteration

bigquery_etl/util/udf_functions.py
lelilia: But if you have a specific problem in mind, we could add that to the tests and see if it works

lelilia: on the function names ... I'm not 100% sure that we are guaranteed it's lowercase but I think it's the convention .. nut sure if it's better to allow for everything or keep in line with the convention here, open for suggestions

lelilia:
this is currently not able to parse the output for outputs that do not have a `RETURNS` statement and will return None for that. Looking at https://github.com/mozilla/bigquery-etl/blob/main/sql/mozfun/bits28/retention/udf.sql I don't see any way I could ever parse that. 
So in those cases we might have to add a `RETURNS` to the UDF or just live with the fact that it won't be added automatically to the documentation

lucia-vargas-a (approved):
One great feature of these functions is to be able to parse the full UDF code. I added a line in the docstring to clarify it and some tests for this scenarios in case someone uses the find_input or find_input separately. Also a small comment on the functions names.
LGTM.

bigquery_etl/util/udf_functions.py
```diff
@@ -0,0 +1,35 @@
+"""Utility functions related to UDF parsing and documentation."""
+
+import re
+from typing import Optional, Tuple
+
+
+def find_input(sql_text: str) -> Optional[str]:
```
lucia-vargas-a, line 7: Out of standard, would it be better to name this function get_input()?

bigquery_etl/util/udf_functions.py
```diff
@@ -0,0 +1,35 @@
...
+    input_groups = re.search(
+        r"FUNCTION [a-z,0-9,_,.]+\s?\(\s?(.+?)\s?\)", cleaned_sql_text
+    )
+    if input_groups is None:
+        return None
+    input_string = input_groups.groups()[0]
+    return input_string
+
+
+def find_output(sql_text: str) -> Optional[str]:
```
lucia-vargas-a, line 19: Also for standard, get_output()?

PR #3791, opened 2023-05-17
Bug 1823627 - Normalize the channel based on probeinfo data in UNIONized views

scholtzan:
A couple of changes for #3787


scholtzan:
Mhh, stage deploys are taking very long...

badboy (approved):

scholtzan:
oh right, I forgot about this. I'll get it landed later today

PR #3802, opened 2023-05-18
Fixes to Android attributable_clients

fbertsch:
Checklist for reviewer:

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -811,6 +811,7 @@ def _run_query(
     destination_table,
     dataset_id,
     query_arguments,
+    addl_templates: dict,
 ):
     """Run a query."""
     if dataset_id is not None:
@@ -881,6 +882,7 @@ def _run_query(
                     template_folder=str(query_file.parent),
                     templates_dir="",
                     format=False,
+                    **addl_templates,
                 )
             )
             query_stream.seek(0)
@@ -1189,34 +1191,64 @@ def initialize(name, sql_dir, project_id, dry_run):
         sys.exit(1)
 
     for query_file in query_files:
-        init_files = Path(query_file.parent).rglob("init.sql")
-        client = bigquery.Client()
+        sql_content = query_file.read_text()
 
-        for init_file in init_files:
-            project = init_file.parent.parent.parent.name
+        # Enable init from query.sql files
+        # First deploys the schema, then runs the init
+        # This does not currently verify the accuracy of the schema
+        if "is_init()" in sql_content:
+            project = query_file.parent.parent.parent.name
+            dataset = query_file.parent.parent.name
+            destination_table = query_file.parent.name
+            Schema.from_schema_file(query_file.parent / SCHEMA_FILE).deploy(
+                f"{project}.{dataset}.{destination_table}"
+            )
+            arguments = [
+                "query",
+                "--use_legacy_sql=false",
+                "--replace",
+                "--format=none",
+            ]
+            _run_query(
+                query_files=[query_file],
+                project_id=project,
+                public_project_id=None,
+                destination_table=destination_table,
+                dataset_id=dataset,
+                query_arguments=arguments,
+                addl_templates={
+                    "is_init": lambda: True,
+                },
+            )
+        else:
+            init_files = Path(query_file.parent).rglob("init.sql")
+            client = bigquery.Client()
 
-            with open(init_file) as init_file_stream:
-                init_sql = init_file_stream.read()
-                dataset = Path(init_file).parent.parent.name
-                job_config = bigquery.QueryJobConfig(
-                    dry_run=dry_run,
-                    default_dataset=f"{project}.{dataset}",
-                )
+            for init_file in init_files:
+                project = init_file.parent.parent.parent.name
 
-                if "CREATE MATERIALIZED VIEW" in init_sql:
-                    click.echo(f"Create materialized view for {init_file}")
-                    # existing materialized view have to be deleted before re-creation
-                    view_name = query_file.parent.name
-                    client.delete_table(
-                        f"{project}.{dataset}.{view_name}", not_found_ok=True
+                with open(init_file) as init_file_stream:
+                    init_sql = init_file_stream.read()
+                    dataset = Path(init_file).parent.parent.name
+                    job_config = bigquery.QueryJobConfig(
+                        dry_run=dry_run,
+                        default_dataset=f"{project}.{dataset}",
                     )
-                else:
-                    click.echo(f"Create destination table for {init_file}")
 
-                job = client.query(init_sql, job_config=job_config)
+                    if "CREATE MATERIALIZED VIEW" in init_sql:
+                        click.echo(f"Create materialized view for {init_file}")
+                        # existing materialized view have to be deleted before re-creation
+                        view_name = query_file.parent.name
+                        client.delete_table(
+                            f"{project}.{dataset}.{view_name}", not_found_ok=True
+                        )
+                    else:
+                        click.echo(f"Create destination table for {init_file}")
 
-                if not dry_run:
-                    job.result()
+                    job = client.query(init_sql, job_config=job_config)
+
+                    if not dry_run:
+                        job.result()
 
 
 @query.command(
@@ -1534,7 +1566,12 @@ def _update_query_schema(
                 existing_schema.to_yaml_file(existing_schema_path)
 
     # replace temporary table references
-    sql_content = query_file_path.read_text()
+    sql_content = render_template(
+        query_file_path.name,
+        template_folder=str(query_file_path.parent),
+        templates_dir="",
+        format=False,
+    )
 
     for orig_table, tmp_table in tmp_tables.items():
         table_parts = orig_table.split(".")
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -9,6 +9,7 @@ import string
 import subprocess
 import sys
 import tempfile
+import typing
 from datetime import date, timedelta
 from functools import partial
 from graphlib import TopologicalSorter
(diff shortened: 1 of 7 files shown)


sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v1/query.sql
```diff
@@ -27,6 +27,7 @@ searches AS (
   SELECT
     submission_date,
     client_id,
+    sample_id,
```
fbertsch, line 30: This is so we have a valid `sample_id` for clients who only reported a search

sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v1/query.sql
```diff
@@ -109,21 +103,18 @@ SELECT
     ELSE 0
   END AS ad_clicks,
   COALESCE(activated, FALSE) AS activated,
+  COALESCE(activated AND activation_date = submission_date, FALSE) AS activated_on_this_day,
```
fbertsch, line 106: This may be controversial. I would like the ability to partition on activated users overall. This is still imperfect (first 6 days won't say whether they're activated), but in this way filtering on `activated` will give e.g. DAU for activated users, not the number of activations. For # of activations, we'd look at `activated_on_this_day`

bani: Yes, I think `activated_on_this_day` is important to have on the table. I still find it easier to understand the table and write queries when the things we want to count are metrics in a table designed for them to be [additive](https://www.kimballgroup.com/data-warehouse-business-intelligence-resources/kimball-techniques/dimensional-modeling-techniques/additive-semi-additive-non-additive-fact/), and this is a step in that direction if we don't want to actually have it as a number.

I'm not entirely sure we need `activated`. It's not one of our official segments and only once we performed an analysis on how activated users compare to non-activated ones. Do you see that request re-ocurring often enough to justify having a hard to understand table with the two different activated values?

sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v1/query.sql
```diff
@@ -59,6 +51,8 @@ adjust_client AS (
     adjust_campaign,
     adjust_creative,
     metadata.reported_first_session_ping,
+    first_seen_date,
+    first_reported_country AS country,
   FROM
     `moz-fx-data-shared-prod`.fenix.firefox_android_clients
   WHERE
```
bani, line 58: I think we should include all clients.

sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v1/query.sql
fbertsch: I guess I think of `activated` as a dimension, not a fact, especially in regards to this table. Maybe we have two fields:
- `is_activated`: Whether the user is activated, a dimension
- `activations_count`: Number of activations, a fact (and additive)

Looking at `is_actiated` across a number of users was kind of the whole reason I started this PR 😄 where as we're trying to understand how our activated users behave, it makes more sense to target them in certain ways (e.g. marketing)

bani: Yes, that would be perfect.

sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v1/query.sql
```diff
@@ -87,43 +90,52 @@ SELECT
   first_seen_date = submission_date
   AND reported_first_session_ping AS is_new_install,
   first_seen_date = submission_date AS is_new_profile,
+  COALESCE(activations.activated, FALSE) AS is_activated,
+  -- Metrics / Facts
+  IF(COALESCE(new_activations.activated, FALSE), 1, 0) AS activations_count,
+  IF(COALESCE(client_days.client_id IS NOT NULL, FALSE), 1, 0) AS client_count,
```
fbertsch, line 96: Quick update on what's happening here. The `activations_count` includes clients who were not necessarily active on this day; which means we now need to include a `client_count` to get the actual DAU (but maybe `client_count` is the wrong name -`active_client_count`? `DAU`? Open to suggestions)

sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v1/query.sql
```diff
@@ -87,43 +90,52 @@ SELECT
...
-    WHEN client_day.has_search_data
-      THEN client_day.ad_clicks
+    WHEN client_days.has_search_data
+      THEN client_days.ad_clicks
     WHEN metrics_searches.has_search_data
       THEN metrics_searches.ad_clicks
     ELSE 0
   END AS ad_clicks,
-  COALESCE(activated, FALSE) AS activated,
+  -- Metadata
```
fbertsch, line 118: Useful now for answering: "Why is this client present on this day?"

bani (approved):

bigquery_etl/cli/query.py
```diff
@@ -1177,34 +1179,64 @@ def initialize(name, sql_dir, project_id, dry_run):
...
-                    default_dataset=f"{project}.{dataset}",
-                )
-
-                if "CREATE MATERIALIZED VIEW" in init_sql:
-                    click.echo(f"Create materialized view for {init_file}")
-                    # existing materialized view have to be deleted before re-creation
-                    view_name = query_file.parent.name
-                    client.delete_table(
-                        f"{project}.{dataset}.{view_name}", not_found_ok=True
+        sql_content = query_file.read_text()
```
fbertsch, line 1182: @scholtzan this adds the capability to initialize a query file. It first deploys the schema, then runs the query with the templated `is_init` set to `True`.

This feels a little hacked together, and is probably not well tested.

bigquery_etl/cli/query.py
```diff
@@ -1488,7 +1520,12 @@ def _update_query_schema(
                 existing_schema.to_yaml_file(existing_schema_path)
 
     # replace temporary table references
-    sql_content = query_file_path.read_text()
+    sql_content = render_template(
```
fbertsch, line 1523: I had to add this so the `bqetl query schema update` worked for templated queries.

bigquery_etl/cli/query.py
```diff
@@ -1177,34 +1179,64 @@ def initialize(name, sql_dir, project_id, dry_run):
...
+                project_id=project,
+                public_project_id=None,
+                destination_table=destination_table,
+                dataset_id=dataset,
+                query_arguments=arguments,
+                addl_templates={
+                    "is_init": lambda: True,
+                },
+            )
+        else:
```
fbertsch, line 1211: Everything after this is the same, just indented one level

bigquery_etl/util/common.py
```diff
@@ -64,6 +64,7 @@ def render(
     sql_filename,
     template_folder=".",
     format=True,
+    is_init=lambda: False,
```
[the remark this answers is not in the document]
fbertsch: I actually like that better, b/c then we don't have to update things in 3 places

scholtzan (approved):
I'm fine with adding this to unblock this workflow. 
I'm not quite convinced this is the most ideal option. It shouldn't be used for queries where `depends_on_past: true`, for example. Ideally, we'd eventually just have `schema.yaml` files and an easy way trigger backfills when creating these new tables.

sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v2/query.sql
```diff
@@ -0,0 +1,106 @@
...
+    ) AS has_search_data,
+    SUM(sum_map_values(metrics.labeled_counter.browser_search_in_content)) AS searches,
+    SUM(sum_map_values(metrics.labeled_counter.browser_search_with_ads)) AS searches_with_ads,
+    SUM(sum_map_values(metrics.labeled_counter.browser_search_ad_clicks)) AS ad_clicks,
+  FROM
+    `moz-fx-data-shared-prod`.fenix.baseline
+  WHERE
+    {% if is_init() %}
+    DATE(submission_timestamp) >= "2021-08-01" {% else %}
+    DATE(submission_timestamp) = @submission_date {% endif %}
```
sean-rose, line 22: This Jinja formatting by `bqetl format` is unfortunate.  I just pushed #3952 which should improve such Jinja block formatting once it's merged.

Also, with the new Jinja block formatting, consistently trimming trailing spaces after Jinja block statements with `-%}` seems to produce the most readable SQL when rendered IMO.

sql/moz-fx-data-shared-prod/fenix/attributable_clients_v2/view.sql
```diff
@@ -0,0 +1,36 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod`.fenix.attributable_clients_v2
```
kik-kik, line 2: I thought views should not include version numbers?

fbertsch: This is a special view; it adds some fields via joins that I need downstream.

On the other hand, we're not quite ready to replace v1, due to some _different_ downstream dependencies. As such, I'm putting the logic in a v2 view, and when we're ready, we'll point the overall view `fenix.attributable_clients` at this view instead.

sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v2/query.sql
fbertsch: This is great, and looks _much_ better. Thank you @sean-rose!

bigquery_etl/cli/query.py
fbertsch: Ah missed this. Yes some tests would be good. I'll file a ticket.

PR #3828, opened 2023-05-22
generate sql and dags for main in a separate CI task for faster results

relud:


scholtzan (approved):

.circleci/config.yml
```diff
@@ -615,9 +624,14 @@ workflows:
...
             - deploy-changes-to-stage
       - generate-sql
+      - main-generate-sql-and-dags:
+          filters:
+            branches:
+              ignore: main
       - generate-diff:
           requires:
             - generate-dags
+            - main-generate-sql-and-dags
```
scholtzan, line 634: I'm guessing this task will still run on `main` even though `main-generate-sql-and-dags` gets skipped?

PR #3847, opened 2023-05-23
feat(DENG-789): making apple ads data accessible

kik-kik:
# feat(DENG-789): making apple ads data accessible


sql/moz-fx-data-shared-prod/apple_ads/ad_group_report/view.sql
```diff
@@ -1,9 +1,4 @@
--- This is an authorized view that allows us to read Apple Ads data
--- from a Fivetran project.
-CREATE OR REPLACE VIEW
-  `moz-fx-data-shared-prod.apple_ads.ad_group_report`
-AS
 SELECT
   *
 FROM
-  `moz-fx-data-bq-fivetran.dbt_fivetran_transformation_apple_search_ads.apple_search_ads__ad_group_report`
+  `moz-fx-data-bq-fivetran.apple_ads_derived.ad_group_report`
```
fbertsch, line 4:
```suggestion
  `moz-fx-data-shared-prod.apple_ads_derived.ad_group_report`
```

Here and the other views

dags.yaml
```diff
@@ -1097,3 +1097,24 @@ bqetl_fenix_external:
...
+    end_date: null
+    owner: kik@mozilla.com
+    retries: 2
+    retry_delay: 30m
+    start_date: '2023-01-01'
+    description: |
+      Copies over apple_ads data coming from Fivetran
+      into our data BQ project. Fivetran syncs this data
+      every hour. We copy the data every 3 hours to our project.
+  schedule_interval: 0 */3 * * *
```
fbertsch, line 1118: Seems a little excessive, but if the data is tiny it doesn't matter

sql/moz-fx-data-shared-prod/apple_ads_derived/ad_group_report_v1/query.sql
```diff
@@ -0,0 +1,4 @@
+SELECT
+  *
```
sean-rose, line 2: I'd recommend explicitly listing the columns to be selected in these mirroring ETLs, since by default Fivetran will automatically add new columns to its tables (`SELECT *` in the downstream views is fine).

sql/moz-fx-data-shared-prod/apple_ads_derived/ad_group_report_v1/query.sql
```diff
@@ -0,0 +1,4 @@
...
+FROM
+  `moz-fx-data-bq-fivetran.dbt_fivetran_transformation_apple_search_ads.apple_search_ads__ad_group_report`
```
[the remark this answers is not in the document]
kik-kik: As per your previous comment, I think in this case if we just list the columns this should be not necessary? I guess adding it to make sure descriptions would be included would be useful.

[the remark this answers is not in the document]
sean-rose: For the automated table deployment process to work it needs to be able to determine the table schema, either by dry-running it or having a `schema.yaml` file.  It won't be able to dry-run these queries, so `schema.yaml` files will be necessary.

sean-rose:
> Other thought too is to use `apple_ads_external` rather than `apple_ads_derived`.

Precedent: when relud did a [similar thing for the Fivetran Stripe tables](https://github.com/mozilla/bigquery-etl/pull/3660) they put the ETLs in a `stripe_external` dataset.

dags.yaml
sean-rose: Especially since the tables are being copied in their entirety each time.  Is the data tiny?

sql/moz-fx-data-shared-prod/apple_ads/ad_group_report/view.sql
```diff
@@ -1,9 +1,4 @@
--- This is an authorized view that allows us to read Apple Ads data
--- from a Fivetran project.
-CREATE OR REPLACE VIEW
-  `moz-fx-data-shared-prod.apple_ads.ad_group_report`
-AS
```
sean-rose, line 5: The `view.sql` files still need to have `CREATE OR REPLACE VIEW`.

sean-rose:
Normally when we have ETLs depend on Fivetran syncs we configure things so Airflow triggers the Fivetran sync and waits for it to complete before starting the ETL.  However, these ETLs depend on downstream Fivetran dbt jobs, not the raw syncs, so I guess we can't/shouldn't do that.

sql/moz-fx-data-shared-prod/apple_ads_derived/ad_group_report_v1/query.sql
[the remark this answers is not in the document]
kik-kik: Either way, I added schema.yaml files to each query so that we can also include the descriptions for each field.

sql/moz-fx-data-shared-prod/apple_ads/ad_group_report/view.sql
kik-kik: 🤦 yes, I'm not sure why I deleted this 😬

sql/moz-fx-data-shared-prod/apple_ads_derived/ad_group_report_v1/metadata.yaml
```diff
@@ -0,0 +1,23 @@
+friendly_name: Ad Group Report (Apple)
+description: |-
+  Each record represents the daily ad performance of each ad group.
+owners:
+- kik@mozilla.com
+labels:
+  application: firefox_ios
```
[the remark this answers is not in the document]
kik-kik: You're right, for some reason firefox_ios attribution was stuck in my head and put that in. I removed it now.

kik-kik:
ok, I renamed `apple_ads_derived` for `apple_ads_external` as requested @sean-rose and @fbertsch. All comments should now be addressed with one exception.

@fbertsch please let me know your thoughts on the scheduling and I can make adjustments if necessary and merge the PR.

sql/moz-fx-data-shared-prod/apple_ads/dataset_metadata.yaml
```diff
@@ -1,10 +0,0 @@
-friendly_name: Apple Ads
```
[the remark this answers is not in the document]
kik-kik: gotcha, thanks for catching this. I re-added the file.

sql/moz-fx-data-shared-prod/apple_ads_external/dataset_metadata.yaml
```diff
@@ -0,0 +1,19 @@
...
+  project over to our data BQ project.
+
+  The fivetran tables being copied over are not raw ingested tables,
+  but instead tables that have been transformed using fivetran transformations
+  power by our DBT project:
+  https://github.com/mozilla/dbt-fivetran-transformations/
+
+  More information on how these models are built can be found here:
+  https://github.com/fivetran/dbt_apple_search_ads/blob/main/models/apple_search_ads.yml
+# dataset_base_acl: restricted
```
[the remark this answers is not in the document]
kik-kik: Now should be fixed, thanks!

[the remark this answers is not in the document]
kik-kik: Now should be fixed, thanks!

bigquery_etl/dryrun.py
```diff
@@ -88,7 +88,7 @@
     "sql/moz-fx-data-shared-prod/fivetran_costs_derived/incremental_mar_v1/query.sql",
     "sql/moz-fx-data-shared-prod/fivetran_costs_derived/monthly_costs_v1/query.sql",
     *glob.glob(
-        "sql/moz-fx-data-shared-prod/apple_ads/**/view.sql",
+        "sql/moz-fx-data-shared-prod/apple_ads_external/**/query.sql",
```
kik-kik, line 91: This will not work until dryrun fix for staging CI step is merged. PR#3858 addresses this issue and will need to be merged prior to this PR.

dags.yaml
```diff
@@ -1097,3 +1097,24 @@ bqetl_fenix_external:
...
+    email_on_retry: true
+    end_date: null
+    owner: kik@mozilla.com
+    retries: 2
+    retry_delay: 30m
+    start_date: '2023-01-01'
+    description: |
+      Copies over apple_ads data coming from Fivetran
+      into our data BQ project. Fivetran syncs this data
+      every hour. We copy the data every 3 hours to our project.
```
sean-rose, line 1117: `description` shouldn't be under `default_args`.
```suggestion
  description: |
    Copies over apple_ads data coming from Fivetran
    into our data BQ project. Fivetran syncs this data
    every hour. We copy the data every 3 hours to our project.
```

dags.yaml
sean-rose: My inclination would be to start with a daily schedule and only sync more frequently if a concrete use case for that comes up.

sean-rose (approved):
r+wc

bigquery_etl/dryrun.py
kik-kik: the above is no longer correct. Explanation of the issue can be found here:
https://github.com/mozilla/bigquery-etl/pull/3856#issuecomment-1564318101

The currently agreed upon solution is to wildcard the project_id in the skip list.

dags.yaml
kik-kik: good catch, updated the yaml definition and regenerated the DAG.

PR #3854, opened 2023-05-24
Add normalized_app_id manually to ping view schemas

scholtzan:
Addresses parts of https://bugzilla.mozilla.org/show_bug.cgi?id=1834827

The view ping schemas are derived from the underlying ping tables. These tables do not include `normalized_app_id` however since it is added in the view definitions. So we have to add it manually to the schemas


sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -81,6 +81,12 @@ def generate_per_app(
             # generate a unioned schema that contains all fields of all ping tables across all channels
             unioned_schema = Schema.empty()
 
+            # normalized_app_id is not part of the underlying table the schemas are derived from,
+            # the field gets added as part of the view definition, so we have to add it manually to the schema
+            unioned_schema.schema["fields"] = [
+                {"name": "normalized_app_id", "mode": "NULLABLE", "type": "STRING"}
+            ]
```
sean-rose, line 88: I think adding `normalized_app_id` to the schema needs to be done after the SQL select statement is generated (e.g. right before generating the `schema.yaml` file below), otherwise the generated SQL select statement will try to select a (nonexistent) `normalized_app_id` column from the underlying tables (and it would also result in a duplicate select result column name).

Also, adding a column description might be nice.

sean-rose (approved):
r+wc

sean-rose (approved):

PR #3856, opened 2023-05-24
DO NOT MERGE! dryrun ci debugging

kik-kik:
# DO NOT MERGE! dryrun ci debugging


bigquery_etl/cli/dryrun.py
```diff
@@ -99,6 +99,8 @@ def dryrun(
         click.echo("Not authenticated to GCP. Run `gcloud auth login` to login.")
         sys.exit(1)
 
+    dryrun.add_test_project_to_skip()
```
scholtzan, line 102: we need to add this line

bigquery_etl/cli/dryrun.py
```diff
@@ -99,6 +99,8 @@ def dryrun(
         click.echo("Not authenticated to GCP. Run `gcloud auth login` to login.")
         sys.exit(1)
 
+    add_test_project_to_skip()
```
scholtzan, line 102: we need to add this

scholtzan:
it's still not working :(

kik-kik:
We believe we found the source of the issue. The issue lies in how the staging testing in CI is structured.

All sql files that have been modified get moved to the test_project before the SKIP list gets generated resulting in the first `glob.glob()` call returning an empty list for those sql files as they no longer exist in the directory specified in the SKIP "constant". The current decision is to just wildcard the project_id when adding sql files to the dryrun SKIP list.

cc: @scholtzan could you please confirm my above understanding of the issue and the solution are correct?

PR #3862, opened 2023-05-26
reassigning DAG ownership to kik as per "adopt a DAG" doc

kik-kik:
# reassigning DAG ownership to kik as per "adopt a DAG" doc


dags/bqetl_addons.py
```diff
@@ -52,7 +52,11 @@
         dataset_id="telemetry_derived",
         project_id="moz-fx-data-shared-prod",
         owner="jklukas@mozilla.com",
```
scholtzan, line 54: best to remove all `jklukas` mentions here

scholtzan: Everything his name touches... is yours now.

kik-kik: yeah, starting to realise my mistake now 😓

kik-kik: I think Jeff should be no more in this PR.

dags/bqetl_addons.py
```diff
@@ -101,7 +101,11 @@
...
```
scholtzan, line 103: One more Jeff is left here

kik-kik: 💩  oops. Should be fixed now. Thanks for being so thorough 🙇

scholtzan (approved):
🎉 thanks for adopting them

PR #3870, opened 2023-05-31
feat(DENG-980): fxa_users_services_first_seen view updated to use v2

kik-kik:
# feat(DENG-980): fxa_users_services_first_seen view updated to use v2

The v1 of this table has not been updated for more than a year now and it does not look like anything is using this view so there shouldn't be any impact.

Checked:

- bigquery-etl
- looker-spoke-default
- looker-hub (some minimal usage, but does not look like this should break anything)


akkomar (approved):
LGTM, it's good to check STMO too. I did that with the following query and it looks like the last query that touched this table was refreshed 3 years ago:
```sql
SELECT *
FROM queries
where query like '%fxa_users_services_first_seen%'
and is_archived=false
and is_draft=false
order by updated_at desc
LIMIT 10
```

PR #3882, opened 2023-06-01
feat(DENG-722): preparing for fxa AWS to GCP migration (nonprod)

kik-kik:
# feat(DENG-722): preparing for fxa migration

Updating nonprod (staging) set up to include the new fxa log event table which will be populated with `auth` and `content` server events once the application is migrated over to GCP.

Once this change is successfully rolled out, we will apply very similar changes to the production logic.


sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/metadata.yaml
```diff
@@ -0,0 +1,31 @@
...
+  FxA server events extracted from accounts server stdout logs.
+
+  This new table is the direct result of FxA migration from AWS to GCP
+  (see: OPST-296 for more context).
+
+  Effective x date the events from the following servers will land in this table:
+  - auth    (previous table: `nonprod_fxa_auth_events_v1`)
+  - content (previous table: `nonprod_fxa_content_events_v1`)
+
+  Payment server events will continue landing inside: `nonprod_fxa_stdout_events_v1`
```
akkomar, line 13: LGTM

akkomar (commented):
LGTM, however if we want to test a production scenario I think it would be worth to add a cutoff date in auth, content, and server_events CTEs in fxa_all_events view. In production view we'd set it to a date after migration is finished, here it would be a day after the staging environment was created. WDYT @kik-kik?

kik-kik:
@akkomar I'm actually not sure we should be doing that other than a note. When the switch over occurs to the GCP service my understanding is that we should not be receiving any events in the old tables unless there's a rollback due to migration issues. In which case we might see some events in the new table and then again in the old table. The reason I think this is not strictly necessary (apart from a note in the tables descriptions) is that based on what I gathered from @jbuck is that the traffic will be re-routed so no duplicates should occur.

akkomar:
Yeah, it is not strictly necessary if everything goes well and there will be no duplicates. But it makes it easier to figure out where's data coming from when reading the query + might make queries more efficient (not sure about this one, but I imagine if you're querying a view with a data filter set to after the cutoff, BQ might be able to do less work).

kik-kik:
okay, since I do not have strong feelings about this I'll add a date filter to the tables. However, we need to remember that in case we do have the scenario of services being rolled back we'll have a time window where we might have events in both the old and the new tables.

sql/moz-fx-data-shared-prod/firefox_accounts/nonprod_fxa_content_auth_stdout_events/view.sql
```diff
@@ -1,7 +1,7 @@
 CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.firefox_accounts.nonprod_fxa_content_auth_stdout_events`
 AS
-  --
+  -- TODO: we should look into deleting this view.
```
akkomar, line 4: Please add some context/links for the next person reading this comment.

akkomar (approved):
LGTM.
FYI @relud since they're listed on a notification list of `firefox_accounts_derived__nonprod_fxa_server_events__v1`.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/metadata.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+  This new table is the direct result of FxA migration from AWS to GCP
+  (see: OPST-296 for more context).
+
+  Effective 2023-05-26 the events from the following servers land in this table
+  (existing data will remain in those tables):
+  - auth    (previous table: `nonprod_fxa_auth_events_v1`)
+  - content (previous table: `nonprod_fxa_content_events_v1`)
+
+  Payment server events will continue landing inside: `nonprod_fxa_stdout_events_v1`
```
relud, line 14: cc @sean-rose

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/query.sql
```diff
@@ -0,0 +1,22 @@
...
+              jsonPayload.fields.* EXCEPT (device_id, user_id),
+              TO_HEX(SHA256(jsonPayload.fields.user_id)) AS user_id,
+              TO_HEX(SHA256(jsonPayload.fields.device_id)) AS device_id
+          ) AS fields
+        )
+    ) AS jsonPayload
+  ),
+  SPLIT(jsonPayload.logger, "-")[OFFSET(1)] fxa_log,
+FROM
+  `moz-fx-fxa-nonprod.gke_fxa_stage_log.stdout`
```
sean-rose, line 17: I was expecting this to end up in the existing `moz-fx-fxa-nonprod-375e` project.  It looks like I don't have access to that dataset in the (new?) `moz-fx-fxa-nonprod` project.  Will all DEs get access, or only on an as-needed basis?

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/metadata.yaml
```diff
@@ -0,0 +1,37 @@
...
+    field: timestamp
+    require_partition_filter: true
+  # cannot set clustering via this config file as no schema.yaml
+  # file due to how volatile the table schema can be.
+  # To mitigate this we create the table using the init.sql file
+  # which creates an empty table with the following fields and configuration:
+  # - timestamp TIMESTAMP - partition field
+  # - fxa_log STRING - cluster field
+  # the argument ALLOW_FIELD_ADDITION takes care of adding all the missing field
+  # as part of the ETL execution.
```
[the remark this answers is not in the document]
kik-kik: we need to add this to dry-run skip anyways due to the source data being in a project to which the CI does not have access to.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/query.sql
kik-kik: I think by default not all DEs get access to it, but I'm sure you could ask for it since you are involved in this specific domain.

The project is also changing for the production data and will not include the random suffix.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/query.sql
```diff
@@ -0,0 +1,22 @@
...
+          (
+            SELECT AS STRUCT
+              jsonPayload.fields.* EXCEPT (device_id, user_id),
+              TO_HEX(SHA256(jsonPayload.fields.user_id)) AS user_id,
+              TO_HEX(SHA256(jsonPayload.fields.device_id)) AS device_id
+          ) AS fields
+        )
+    ) AS jsonPayload
+  ),
+  SPLIT(jsonPayload.logger, "-")[OFFSET(1)] fxa_log,
```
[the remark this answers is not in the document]
kik-kik: Yes @aliasing, I just missed it.

Regarding OFFSET VS SAFE_OFFSET, my thinking here was that this should never fail and if it does then there's a problem and maybe breaking the ETL at this point would be the right thing to do to catch the issue as soon as possible @sean-rose .

sql/moz-fx-data-shared-prod/firefox_accounts/nonprod_fxa_all_events/view.sql
```diff
@@ -0,0 +1,158 @@
...
+    jsonPayload.fields.language,
+    jsonPayload.fields.app_version,
+    jsonPayload.fields.os_name,
+    jsonPayload.fields.os_version,
+    jsonPayload.fields.event_type,
+    jsonPayload.logger,
+    jsonPayload.fields.user_properties,
+    jsonPayload.fields.event_properties,
+    jsonPayload.fields.device_id,
+    "payments" AS fxa_log
```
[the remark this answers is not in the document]
kik-kik: We could rename the table, but I also think that it's not really necessary. I would update the fxa_log value to be `payments` and a note in the `metadata.yaml` to indicate this. As far as anything downstream from the view is concerned, it should rely on the `fxa_log` field (or whatever we decide to call it).

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/metadata.yaml
kik-kik: regarding the renaming to `auth_content_events` this would also not be desirable as in the future the payments events might also end up landing in this table and once again we end up with naming confusion.

Alternative here could be to have our ETL include the events from the existing `payments` (stdout) table and the data coming from multiple tables being an implementation detail. This is what it would look like: https://docs.google.com/document/d/1fe6aUXHPU3P4ZZItDB5Wc2r8Y6I_pI9m1BVcGQjIqs0/edit?disco=AAAAzAOKPg4

sql/moz-fx-data-shared-prod/firefox_accounts/nonprod_fxa_content_auth_stdout_events/view.sql
kik-kik: ok, added a bit more context to this comment. I hope that adds clarity.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/metadata.yaml
sean-rose: Given that design vision I'd be inclined to include payments server events in this ETL now (i.e. option 2).  You'll just need to be careful with having a hard cutoff (as opposed to the overlapping cutoff for the other logs that transitioned) in both the`server_events` ETL for when to start including existing `stdout` logs and in `nonprod_fxa_all_events` for when to stop including logs from the existing `stdout` ETL (otherwise `nonprod_fxa_all_events` will select duplicate `stdout` logs).

sql/moz-fx-data-shared-prod/firefox_accounts/nonprod_fxa_all_events/view.sql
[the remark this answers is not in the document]
sean-rose: I think calling it `fxa_service` would be a bit more intuitive/user friendly.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/metadata.yaml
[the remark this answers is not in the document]
sean-rose: Copying what I wrote in [Slack](https://mozilla.slack.com/archives/C01E8GDG80N/p1686756243631509?thread_ts=1686596901.565209&cid=C01E8GDG80N) here for posterity/continued discussion:

Adding it to dry-run skip doesn't solve the CI problem, which has to do with changes to ETLs/views immediately downstream of such ETLs.  When CI tries to deploy all the things the changed ETL/view is referencing to staging it can't do that for any un-dry-runnable ETLs referenced unless they have a `schema.yaml` file.

For example, the `fxa_all_events` view selects from un-dry-runnable ETLs, so for any changes to `fxa_all_events` to successfully pass CI in order to get deployed the FxA events ETLs it references must have `schema.yaml` files, otherwise CI fails to deploy them to staging in order to run the required `validate-metadata` CI check (as I found in #3926).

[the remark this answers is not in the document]
kik-kik: ok, I will add the remainder of the schema.yaml files. I will pull the nonprod ones from your PR into this one.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/metadata.yaml
akkomar: > You'll just need to be careful with having a hard cutoff (as opposed to the overlapping cutoff for the other logs that transitioned) in both theserver_events ETL for when to start including existing stdout logs and in nonprod_fxa_all_events for when to stop including logs from the existing stdout ETL (otherwise nonprod_fxa_all_events will select duplicate stdout logs).

IMO this is too complex, it's easier to reason about what's going on in these logs if there's 1-1 relationship between these tables and their sources.
I'd rather opt for either keeping this query as is (I agree that naming is not ideal and am open to suggestions) or separate `auth` and `content` (although this breaks 1-1 relationship mentioned above, but is simpler to reason about in case someone's e.g. debugging some missing data).

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/query.sql
```diff
@@ -0,0 +1,22 @@
...
+        jsonPayload.* REPLACE (
+          (
+            SELECT AS STRUCT
+              jsonPayload.fields.* EXCEPT (device_id, user_id),
+              TO_HEX(SHA256(jsonPayload.fields.user_id)) AS user_id,
+              TO_HEX(SHA256(jsonPayload.fields.device_id)) AS device_id
+          ) AS fields
+        )
+    ) AS jsonPayload
+  ),
```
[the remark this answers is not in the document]
kik-kik: Whilst `jsonPayload` is critical here, I'm not really sure about the other fields such as `labels` and `resource` since we do not use them anywhere in our analytics and I don't see how they could provide much value for this use case. I guess the main question to consider here would be if we actually should be preserving them at all. What are your thoughts @akkomar ?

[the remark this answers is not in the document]
akkomar: This is a good suggestion, but since there is a slight time pressure to get this work done I suggest we follow the convention used in production tables here and file a new PR to rework table schemas.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/metadata.yaml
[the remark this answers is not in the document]
kik-kik: schema files added. I consider this conversation resolved, please let me know if there's anything I missed.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/query.sql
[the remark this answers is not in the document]
kik-kik: All comments in this conversation have been addressed with the exception of this:

> Similar thought as a previous comment about whether we should differentiate between log/table vs server/service

@sean-rose could you please expand on this?

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/query.sql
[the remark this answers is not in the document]
kik-kik: At this stage I'd also argue that unless we implement this change for all the fxa log tables then maybe we should keep this one behaving the same way for consistency even if it's not ideal.

sql/moz-fx-data-shared-prod/firefox_accounts/nonprod_fxa_all_events/view.sql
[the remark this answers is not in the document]
kik-kik: I'm not necessarily opposed to this change. But I'm a bit worried that it will end up getting mixed up with this field: `jsonPayload.fields.event_properties.service` which seems to correspond to a Mozilla product.

Perhaps an alternative here would be something like `fxa_server` to avoid the confusion with mozilla services?

I do think it might be worth improving upon our naming, however, I'd suggest we treat it as a separate unit of work from the migration.

We could have a follow-up change/improvement to rename these fields since that would have to be done on the view level and would have downstream impact.

What are your thoughts @sean-rose ?

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/metadata.yaml
sean-rose: Yeah, there's no ideal solution here.

If there's a definite plan to switch to having payments events also land in `moz-fx-fxa-nonprod.gke_fxa_stage_log.stdout` (and similarly for production) then I think it'd make sense to keep this as `server_events`.  Otherwise I'd personally be inclined to have separate `auth` and `content` v2 ETLs (i.e. option 3).

sql/moz-fx-data-shared-prod/firefox_accounts/nonprod_fxa_all_events/view.sql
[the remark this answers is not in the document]
sean-rose: Sure, `fxa_server` seems like a reasonable name, especially since in the new ETL the value is being extracted from `logger` values like "fxa-auth-server".

And I'm fine with punting on the `fxa_log` column renaming and doing it later.  However, in the same spirit, maybe this should be `"stdout" AS fxa_log` to keep with the current convention for now, and only change it to `"payments"` at the same time the `fxa_log` column renaming is done, since changing the value also potentially has downstream impacts?

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/query.sql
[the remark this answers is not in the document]
sean-rose: > Similar thought as a previous comment about whether we should differentiate between log/table vs server/service

That's referring to the other comment thread where we discussed renaming `fxa_log` to `fxa_server`.

I also had an idea that including more explicit information about the underlying FxA log (e.g. passing through `logName` as `fxa_log`) or BigQuery table (e.g. having an `fxa_log_table` column with whatever the table happens to be, like "moz-fx-fxa-nonprod.gke_fxa_stage_log.stdout") might be useful, but that's probably not worth hassling about at this point.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/query.sql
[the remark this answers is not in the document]
sean-rose: That's fine with me, it was just a suggestion.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/schema.yaml
sean-rose, line 1: It looks like maybe you generated this `schema.yaml` file by hand or using some custom process (based on the order of the properties, indentation, and some blank lines in the middle of the file).

FYI, last week I merged #3923, and now `bqetl query schema update` produces a more reasonable `schema.yaml` file with the properties in a sensible order by default (`name`, `type`, `mode`, `fields`), so you might consider using that now (if you do, you'd probably need to delete/rename this file before running the command, otherwise it would preserve the existing properties order).

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/metadata.yaml
akkomar: > If there's a definite plan to switch to having payments events also land in `moz-fx-fxa-nonprod.gke_fxa_stage_log.stdout` (and similarly for production)

No concrete plan exists yet. It will probably happen in the future, next year at the earliest.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/query.sql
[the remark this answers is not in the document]
kik-kik: Thanks, we do appreciate it. 🙇

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/schema.yaml
kik-kik: When I try to update the schema for this table using the above command I keep getting permissions errors. For now I just ended up copying the schema file from the nonprod_stdout table, I think this might be good enough for now at least.

akkomar:
> Do we want to split auth and content events into two separate tables and then union them with `nonprod_fxa_all_events` as we have been doing it so far? Or have both of these server events landing in the same table and potentially rename it to reflect that it does not contain all the events? for example `nonprod_content_auth_events_v1`?
> @akkomar do you have any thoughts / preference here?

Either is fine by me. I don't expect us to see changes to the source tables soon and I do expect us to replace this telemetry with Glean over next year.

kik-kik:
Ok, so having thought about it a bit more. I think we should keep this as a single table basically 1:1 mapping between the source and target table. Whilst in staging indeed we only have `auth` and `content` events. In prod, we also get other events from other servers such as `fxa-bounce-server`. Instead of having to create a new ETL to capture events from these sources I think it's much easier to just include them all. It's unfortunate about `payments` events landing in a different table, but I don't think this should be the only reason why we take the more complicated approach (multiple ETL with different logs that need to be added for each new server source).

Especially, since we expose them downstream via a view which unions all of them together anyways.

sql/moz-fx-data-shared-prod/firefox_accounts/nonprod_fxa_all_events/view.sql
```diff
@@ -0,0 +1,159 @@
...
+    jsonPayload.logger,
+    jsonPayload.fields.user_properties,
+    jsonPayload.fields.event_properties,
+    jsonPayload.fields.device_id,
+    "payments" AS fxa_server
+  FROM
+    `moz-fx-data-shared-prod.firefox_accounts_derived.nonprod_fxa_stdout_events_v1`
+),
+-- New fxa event table (nonprod) includes, content and auth events
+-- this also includes "payments" events.
```
sean-rose, line 67: This doesn't agree with the ETL description metadata:
> Payment server events will continue landing inside: `nonprod_fxa_stdout_events_v1`

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/metadata.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - content (previous table: `nonprod_fxa_content_events_v1`)
+
+  Payment server events will continue landing inside: `nonprod_fxa_stdout_events_v1`
+owners:
+- kik@mozilla.com
+labels:
+  application: fxa
+  incremental: true
+  schedule: daily
+  owner: kik
```
sean-rose, line 20: I think typically we don't specify `owner` labels, and `owner#` labels get automatically generated based on the `owners` list above (e.g. as is this table would get an `owner1: kik` label).

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/schema.yaml
```diff
@@ -0,0 +1,379 @@
+fields:
+- name: fxa_server
+  type: STRING
+  mode: NULLABLE
```
sean-rose, line 4: The ETL query currently selects `fxa_server` as the last column, not the first.  Either the query or this schema should be changed so they match (since `fxa_server` is a clustering column I'd be inclined to change the query to select that column first).

sql/moz-fx-data-shared-prod/firefox_accounts/nonprod_fxa_content_auth_stdout_events/view.sql
```diff
@@ -1,7 +1,11 @@
 CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.firefox_accounts.nonprod_fxa_content_auth_stdout_events`
 AS
-  --
+  -- TODO: we should look into deleting this view
+  -- as we should use firefox_accounts.fxa_all_events instead
+  -- to access fxa event data.
```
sean-rose, line 6:
```suggestion
  -- as we should use firefox_accounts.nonprod_fxa_all_events instead
  -- to access nonprod fxa event data.
```

sean-rose (approved):
r+wc

sql/moz-fx-data-shared-prod/firefox_accounts_derived/nonprod_fxa_server_events_v1/schema.yaml
sean-rose: By default it'll use the cloud function with default Google credentials which won't have access to the underlying FxA log tables.  Also, since this ETL is in the dryrun skip list you have to specifically tell it to ignore that.

I think this should work (again, first renaming/deleting the existing `schema.yaml` file so you get a fresh one):
```bash
gcloud auth login --update-adc

bqetl query schema update --use-cloud-function=false --ignore-dryrun-skip firefox_accounts_derived.nonprod_fxa_server_events_v1
```

kik-kik:
Thanks for the comments @sean-rose , made adjustments as requested. Once the checks pass I will merge this PR.

PR #3885, opened 2023-06-01
DENG-941 initial impl of check rendering and execution.

gleonard-m:
Checklist for reviewer:

**Summary:**
1. Any parameters specified on the CLI are parsed and passed to BQ via the query_arguments.
2. Any parameters specified in the query_arguments (via Airflow) are parsed and passed to jinja template processing.
3. Instead of` ./bqetl query test` I implemented  `./bqetl check run`. 
4.  I used Jinja to render the query.  The issue is with the replacement of `download_date` in the AS clause.  `@download_date` was not replaced using the query parameters, it was only replaced in the actual SELECT statement.  I tried using a CTE however a CTE requires a SELECT statement immediately following it, ASSERT caused an error.
5. I included the text `ETL Data Check Failed:` in the error message (hack impl for now).  I think that once the macros are coded we should add consistent prefix to the message.
6.  The sample `checks.sql `file is for the downloads table I recently added.  The threshold is not valid and will have to be updated but by changing the partition being checked you can see the check passing or failing.
7.  Sample commands:
> Pass:  ./bqetl check run ga_derived.downloads_with_attribution_v2 --parameter download_date::2023-05-26
> Failure: ./bqetl check run ga_derived.downloads_with_attribution_v2 --parameter download_date::2023-05-28
8. I'll add more tests once we settle on the implementation (before merging).

**Questions:**
1. Does `project_id` need to be specified?
2. Does `sql_dir` need to be specified


bigquery_etl/cli/query.py
```diff
@@ -1262,6 +1265,151 @@ def render(name, sql_dir, output_dir):
...
+    if re.match(r"^\w+$", params[0]):
+        return {params[0]: params[2]}
+
+
+def _validate_partition_date(ctx, param, partition_date):
+    """Process the CLI parameter check_date and set the parameter for BigQuery."""
+    # Will be None if launched from Airflow.  Also ctx.args is not populated at this stage.
+    if partition_date is None:
+        return None
+    else:
```
kik-kik, line 1289: Redundant else, if the above condition is met then this block won't get executed so `else` can be removed.

bigquery_etl/cli/query.py
```diff
@@ -1262,6 +1265,151 @@ def render(name, sql_dir, output_dir):
...
+        parsed = _parse_partition_date(partition_date)
+        if parsed is None:
```
kik-kik, line 1291: Should we consider raising this exception inside `_parse_partition_date` instead?

bigquery_etl/cli/query.py
```diff
@@ -1262,6 +1265,151 @@ def render(name, sql_dir, output_dir):
...
+            raise click.BadParameter("Format must be <column-name>::<yyyy-mm-dd>")
+        return parsed
```
kik-kik, line 1293: Generally, if we want we could perhaps do something like this instead and outsource the validation logic to `click`:
https://github.com/mozilla/bigquery-etl/blob/main/bigquery_etl/cli/backfill.py#L56-L63

bigquery_etl/cli/query.py
```diff
@@ -1262,6 +1265,151 @@ def render(name, sql_dir, output_dir):
...
+    query_arguments.append("--use_legacy_sql=false")
+    if project_id is not None:
+        query_arguments.append(f"--project_id={project_id}")
+
+    # Partition will be None if from Airflow
+    if partition is None:
+        # Need to check the query_arguments for the value
+        for parameter in query_arguments:
+            if parameter.startswith("--parameter"):
+                param_value = parameter.split("=")[1]
```
kik-kik, line 1372: Maybe I'd add a comment here why this logic to go through all parameters? I assume this is because some of our queries use `submission_date` and some other might use a different field for as a partitioning field?

bigquery_etl/cli/query.py
```diff
@@ -1262,6 +1265,151 @@ def render(name, sql_dir, output_dir):
...
+        ignore_unknown_options=True,
+        allow_extra_args=True,
+    ),
+)
+@click.argument("name")
+@sql_dir_option
+@click.option(
+    "--partition",
+    "-p",
+    help="Partition to check, format <column-name>::<yyy-mm-dd>, must be provided if not executing in Airflow ",
```
kik-kik, line 1325: Should we consider `column-name` in this case being a separate command param?

bigquery_etl/cli/query.py
```diff
@@ -1262,6 +1265,151 @@ def render(name, sql_dir, output_dir):
...
+    table,
+    partition,
+    query_arguments,
+):
+    """Run the check."""
+    query_arguments.append("--use_legacy_sql=false")
+    if project_id is not None:
+        query_arguments.append(f"--project_id={project_id}")
+
+    # Partition will be None if from Airflow
```
kik-kik, line 1367: Why should it be null when running the command from Airlfow? Couldn't Airflow call to to this command just pass it as an argument?

bigquery_etl/cli/query.py
```diff
@@ -1262,6 +1265,151 @@ def render(name, sql_dir, output_dir):
...
+    if len(params) != 3:
+        return None
+
+    # Check date format
+    try:
+        datetime.datetime.strptime(params[2], "%Y-%m-%d").date()
+    except ValueError:
+        return None
+
+    # Check column name
```
kik-kik, line 1279: I'd consider splitting column name and date partition to be two separate params. Otherwise, maybe we should consider renaming this function and it's parameter since we're not working with just a partition_date but more something like partition_settings.

bigquery_etl/cli/query.py
```diff
@@ -1262,6 +1265,151 @@ def render(name, sql_dir, output_dir):
...
+
+    if partition is None:
+        raise ValueError("No partition specified to check.")
+
+    jinja_params = {
+        **{"project_id": project_id, "dataset_id": dataset_id, "table_name": table},
+        **partition,
+    }
+
+    with tempfile.NamedTemporaryFile(mode="w+") as query_stream:
```
kik-kik, line 1391: Is there a specific reason why we write the generated template into file vs just keeping it memory and passing it into `bq query` as text?

bigquery_etl/cli/utils.py
```diff
@@ -79,7 +83,25 @@ def table_matches_patterns(pattern, invert, table):
     return matching != invert
 
 
-def paths_matching_name_pattern(pattern, sql_path, project_id, files=["*.sql"]):
+def paths_matching_checks_pattern(pattern, sql_path, project_id):
+    """Return single path to checks.sql matching the name pattern."""
+    checks_files = paths_matching_name_pattern(
+        pattern, sql_path, project_id, ["checks.sql"], CHECKS_FILE_RE
+    )
+    assert len(checks_files) < 2  # should only have 0 or 1 file.
```
kik-kik, line 91: What if we pattern match on a whole namespace like `firefox_accounts_derived` because we want to run all tests? Is this something we may want to support?

kik-kik: We should be careful with using `assert` like this inside application code:
https://stackoverflow.com/a/1838411

bigquery_etl/cli/query.py
```diff
@@ -1,6 +1,7 @@
 """bigquery-etl CLI query command."""
```
kik-kik, line 1: Should we consider using a different namespace than `query` for this? It feels like it may cause some confusion since we're not checking the query, but were checking the data. Maybe a new namespace called `check` or `data_check` would help avoid this?

kik-kik:
Would it be possible to update the PR title to something a bit more descriptive like: `feat(DENG-941): Check rendering and execution (part of data checks impl.)`

bigquery_etl/cli/query.py
gleonard-m: This function is already specified as the validation callback for the partition option.

bigquery_etl/cli/query.py
gleonard-m: We could but I wanted to keep the command param the same as the ctx.arg format/value so only one validation path is needed.

bigquery_etl/cli/query.py
gleonard-m: The standard method is for Airflow to pass any partition params in the ctx.args.  It doesn't pass the partition info as a param to bqelt CLI unless I misunderstood the current implementation.

bigquery_etl/cli/utils.py
gleonard-m: > What if we pattern match on a whole namespace like firefox_accounts_derived because we want to run all tests? Is this something we may want to support?
This could be added as a Jira item and decided on later, not Phase 1.  Phase 1 is only at the table level for a controlled release.

bigquery_etl/cli/query.py
gleonard-m: It's how it is currently done in `_run_query`, no reason to not follow the same pattern.

bigquery_etl/cli/utils.py
gleonard-m: > We should be careful with using assert like this inside application code:

This link is specific to Python and using asserts instead of clean and narrow exception handling.  Using ASSERT here isn't the same use case.  Verification of using ASSERT vs ERROR is one of the many reasons for rolling this out to a small set of non-critical tables.

bigquery_etl/cli/query.py
gleonard-m: I followed what was in the proposal which indicated  `query` (which also includes the `run` command).  I agree `check` isn't a great name, see point 2 in the above Summary.  I can update it to data_check.

bigquery_etl/cli/query.py
gleonard-m: No because the function is also called when parsing the ctx.args from Airflow and raising a click exception in that processing path is not appropriate.

bigquery_etl/cli/query.py
gleonard-m: Renamed to `_parse_partition_setting`

bigquery_etl/cli/query.py
scholtzan: I'm wondering if we should have a more generic `--parameters` option here instead. There are some queries which have additional parameters besides the partition date which could be specified then as well.

bigquery_etl/cli/query.py
```diff
@@ -1262,6 +1265,150 @@ def render(name, sql_dir, output_dir):
...
+                if partition is not None:
+                    break
+    else:
+        # We have a partition value from the CLI so add to the query_arguments.
+        # There should only be 1
+        key, value = next(iter(partition.items()))
+        query_arguments.append(f"--parameter={key}::{value}")
+
+    if partition is None:
+        raise ValueError("No partition specified to check.")
```
scholtzan, line 1383: We do have some tables that are not partitioned, so we should support them as well

bigquery_etl/cli/query.py
gleonard-m: Sounds good,  I'll move it to a separate command.

bigquery_etl/cli/query.py
gleonard-m: can you provide an example table I can use for dev testing?

scholtzan: https://github.com/mozilla/bigquery-etl/blob/505c895f624ce7f9b13fc3c16752f0d9a9420456/sql/moz-fx-data-marketing-prod/adjust_derived/firefox_mobile_installs_v1/query.sql for example.

Usually, unpartitioned (non-incremental) queries are marked in the metadata with `incremental: false`. There are a couple more if you need more examples

kik-kik: I wonder if as part of the base work for this tool we should put some effort into ensuring all metadata follow at least some rules. I don't think currently all of them would have the `incremental` label so we can't really rely on it. Or maybe we could use something like the `date_partitioning_parameter` inside the `scheduling` [section](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/apple_ads_external/ad_group_report_v1/metadata.yaml#L10) to automatically determine which field to use and whether check should be against the whole table or just a partition?

bigquery_etl/cli/utils.py
kik-kik: I agree with the first comment, regarding the `assert`. Generally, you have to be careful about using it for logic because if make use of the PYTHONOPTIMIZE assert statements would be entirely removed from the execution code. Meaning if we rely on them to check something and stop the execution then this check could never happen.

bigquery_etl/cli/query.py
gleonard-m: I don't think that the Phase 1 implementation of BQETL checks is the place to enforce rules on the metadata.yaml files.  It should/could be a separate Issue in bigquery-etl itself.

bigquery_etl/cli/check.py
```diff
@@ -0,0 +1,199 @@
...
+    return [f"--parameter={key}::{value}" for key, value in parameters.items()]
+
+
+def _parse_query_arguments(query_args):
+    parameters = []
+    for query_arg in query_args:
+        if query_arg.startswith("--parameter=") and len(query_arg.split("=")) == 2:
+            parameters.append(query_arg.split("=")[1])
+        else:
+            raise ValueError("argument must start with --parameter")
```
scholtzan, line 32: does this mean `--use_legacy_sql=false` as query argument is not supported?

bigquery_etl/cli/check.py
```diff
@@ -0,0 +1,199 @@
...
+    query_arguments.append("--use_legacy_sql=false")
+    if project_id is not None:
+        query_arguments.append(f"--project_id={project_id}")
+
+    if parameters is None:
+        # Convert all the Airflow params to jinja usable dict.
+        parameters = _parse_query_arguments(query_arguments)
+    else:
+        # Convert all CLI parameters to BQ query args.
+        query_arguments += _build_query_arguments(parameters)
```
scholtzan, line 173: I'm wondering if parsing these parameters is actually necessary. For our scheduling configuration it is already possible to specify parameters as strings, and they just get used 1:1: https://github.com/mozilla/bigquery-etl/blob/d8f1dbf1ff56698a25e422ab60797e88c25cc875/sql/moz-fx-data-marketing-prod/ga_derived/www_site_empty_check_v1/metadata.yaml#L13

Could we do the same here? `--parameter` would just take a string and pass it into `bq query` without any parsing and or validation? If something is incorrect then it would get raised when running the check.

I'd imagine that once these checks get scheduled in Airflow we can either reuse the `parameters` under `scheduling` or add a new `checks` section in `metadata.yaml` where parameters can be specified that are used for these checks.

bigquery_etl/cli/query.py
```diff
@@ -1265,6 +1268,150 @@ def render(name, sql_dir, output_dir):
...
+        # run the query as shell command so that passed parameters can be used as is
+        try:
+            subprocess.check_output(
+                ["bq", "query"] + query_arguments, stdin=query_stream, encoding="UTF-8"
+            )
+        except CalledProcessError as e:
+            print(_parse_check_output(e.output))
+            sys.exit(1)
+
+
```
scholtzan, line 1414: it looks like this moved into `checks.py` and can be removed here?

bigquery_etl/cli/check.py
```diff
@@ -0,0 +1,199 @@
...
+
+    if parameters is None:
+        return {}
+
+    for param in parameters:
+        params = param.split(":")
+        if len(params) != 3:
+            raise ValueError(
+                f"parameter: {param} is not a valid parameter.  Please use format: <param-name>::<param-value>, exiting"
+            )
```
scholtzan, line 53: By splitting the parameter definition we lose the type information (`params[1]`)

scholtzan (commented):
My main concern is whether parsing and validating these parameters misses some edge cases and makes things slightly more complicated. For our scheduling configuration it is already possible to specify parameters as strings, and they just get used 1:1: https://github.com/mozilla/bigquery-etl/blob/d8f1dbf1ff56698a25e422ab60797e88c25cc875/sql/moz-fx-data-marketing-prod/ga_derived/www_site_empty_check_v1/metadata.yaml#L13

Could we do the same here? (some more context inline)

bigquery_etl/cli/check.py
```diff
@@ -0,0 +1,199 @@
...
+
+        # If the param value smells like a date then check if it is a valid date.
+        if re.match(r"^[0-9-]+$", params[2]):
```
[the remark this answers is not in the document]
kik-kik: To be honest, I kind of like the idea of having to specifically define the data type in the parameter.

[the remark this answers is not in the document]
gleonard-m: This is the validation I mentioned at the meeting, that anything parameter value with numbers and '-' would be confirmed to be a date.  I had concerns but left in in for discussion.  I'll remove it.

bigquery_etl/cli/query.py
gleonard-m: Oops, missed that.

bigquery_etl/cli/check.py
```diff
@@ -0,0 +1,199 @@
...
+    if parameters is None:
+        return []
+    return [f"--parameter={key}::{value}" for key, value in parameters.items()]
+
+
+def _parse_query_arguments(query_args):
+    parameters = []
+    for query_arg in query_args:
+        if query_arg.startswith("--parameter=") and len(query_arg.split("=")) == 2:
+            parameters.append(query_arg.split("=")[1])
```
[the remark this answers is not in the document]
gleonard-m: Yeah, I've made this a lot more complicated than it needs to be.

bigquery_etl/cli/check.py
gleonard-m: No longer doing this.

bigquery_etl/cli/check.py
gleonard-m: latest updates fully support all formats.

kik-kik: If we use query parameters this flag needs to be passed `--use_legacy_sql=false`, I wonder could we just implicitly always pass that flag without me as the user having to include it in the command?

gleonard-m: `--use_legacy_sql=false` is always added in `_run_check()`

bigquery_etl/cli/query.py
```diff
@@ -1265,6 +1266,40 @@ def render(name, sql_dir, output_dir):
...
+def _validate_partition_date(ctx, param, partition_date):
+    """Process the CLI parameter check_date and set the parameter for BigQuery."""
+    # Will be None if launched from Airflow.  Also ctx.args is not populated at this stage.
+    if partition_date:
+        parsed = _parse_partition_setting(partition_date)
+        if parsed is None:
+            raise click.BadParameter("Format must be <column-name>::<yyyy-mm-dd>")
+        return parsed
+    return None
```
scholtzan, line 1293:
```suggestion
```

These methods seem to be unused

scholtzan (approved):
🎉

PR #3891, opened 2023-06-02
feat(): added support for `--log-level` to bqetl query command and using logging instead of print()

kik-kik:
# feat(): added support for `--log-level` to bqetl ~query~ command and using logging instead of print() inside bqetl query cli module.

The idea behind adding the `--log-level` flag is to over time tweak some of the logging statements to be debug statements to reduce the amount of noise produced by the query command unless the user explicitly wants to see all output.

Moving away from `print()` to `logging` makes the command output easier to work with and identify the source of the messages.

Example command:
`./bqetl --log-level=debug query my.query query`

The log level setting should now apply across runtime for all of the cli commands.

(part of the great bug throwdown pairing session, so kudos for @lelilia @lucia-vargas-a @akkomar @Iinh for participating and coming up with ideas to improve the command's output)
----
An alternative approach would be to have a `--verbose` flag to print all messages which could be excluded in a non-verbose mode.

### DIFF
--- a/bigquery_etl/cli/__init__.py
+++ b/bigquery_etl/cli/__init__.py
@@ -24,6 +24,7 @@ from ..glam.cli import glam
 from ..static import static_
 from ..stripe import stripe_
 from ..subplat.apple import apple
+import logging
 
 
 def cli(prog_name=None):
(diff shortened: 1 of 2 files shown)


bigquery_etl/cli/__init__.py
```diff
@@ -51,9 +52,24 @@ def cli(prog_name=None):
...
+    @click.option("--log-level", "log_level", default="info", type=str)
+    def group(log_level):
         """CLI tools for working with bigquery-etl."""
-        pass
+        log_level_mapping = {
+            "debug": logging.DEBUG,
+            "info": logging.INFO,
+            "error": logging.ERROR,
+            "critical": logging.CRITICAL,
+        }
```
scholtzan, line 63:
```suggestion
```

scholtzan (approved):
🎉 just a few suggestions

kik-kik:
Thank you @scholtzan for your comments. Made the changes, will merge this tomorrow.

relud (approved):

PR #3893, opened 2023-06-02
Move schema and initialization logic for firefox_android_clients_v1 to metadata

ANich:
PR moves schema info to a `schema.yaml` file and removes initialization metadata (already in `metadata.yaml`). If still required, the `init.sql` can be used as a standalone script. 

Also stage files on changes to `init.sql`


sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
```diff
@@ -1,108 +1,4 @@
 -- Initialization query first observations for Firefox Android Clients.
-CREATE OR REPLACE TABLE
-  `moz-fx-data-shared-prod.fenix_derived.firefox_android_clients_v1`(
```
ANich, line 3: This could be kept if it makes things easier too

scholtzan (approved):

PR #3899, opened 2023-06-03
Fix `bqetl stage` table ID quoting

sean-rose:
#3791 changed `bqetl stage` so that replaced table IDs were entirely in quotes, but quoting the entire table ID breaks cases where an unaliased table name is used to qualify a column reference ([example](https://github.com/mozilla/bigquery-etl/blob/cfca76f9c498d6ae0eec4fe2dae08b642d7cd426/sql/moz-fx-data-shared-prod/mozilla_vpn/active_subscription_ids/view.sql#L10-L13)).

Quoting the individual table ID parts separately does not have the same problem.

---
Checklist for reviewer:

### DIFF
--- a/bigquery_etl/cli/stage.py
+++ b/bigquery_etl/cli/stage.py
@@ -301,14 +301,14 @@ def _update_references(artifact_files, project_id, dataset_suffix, sql_dir):
                 re.compile(
                     rf"(?<![\._])`?{original_dataset}`?\.`?{name_pattern}(?![a-zA-Z0-9_])`?"
                 ),
-                f"`{deployed_project}.{deployed_dataset}.{name}`",
+                f"`{deployed_project}`.`{deployed_dataset}`.`{name}`",
             )
         )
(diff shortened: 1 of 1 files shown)


sean-rose:
This issue is blocking #3898 from being deployed.

scholtzan:
The problem we have seen with separate quotes is when selecting from a table that has a field with the same name. We have seen this for `metrics` views, for example, where:

```
SELECT
  metrics,
  ...
FROM `moz-fx-data-shared-prod`.`fenix`.`metrics`
```

was causing an error. I'll have a closer look on Monday to see if there is a way where we don't run into issues

sean-rose:
> The problem we have seen with separate quotes is when selecting from a table that has a field with the same name.

So many edge cases. 🙃

I just pushed an update with a possible approach to preserving fully quoted table identifiers when `bqetl stage` does replacing.  The presumption being that each existing query is written in a way that actually works for its situation (either fully quoted table IDs or not as the case may be).

scholtzan (approved):
Ah, nice. I think this should handle the cases I've come across so far

PR #3904, opened 2023-06-05
DENG-984 Added backfill staging dataset

wwyc:
[DENG-984](https://mozilla-hub.atlassian.net/browse/DENG-984): Created a big query dataset to store backfill data.

[DENG-984]: https://mozilla-hub.atlassian.net/browse/DENG-984?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sql/moz-fx-data-shared-prod/backfills_staging_derived/dataset_metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/backfills_staging_derived/dataset_metadata.yaml
@@ -1,6 +1,6 @@
 friendly_name: Backfills Staging
 description: |-
-  Staging location for backfills
+  Temporary staging location for backfill data before it replaces production data.
 dataset_base_acl: derived
 user_facing: false
 labels: {}
--- a/sql/moz-fx-data-shared-prod/backfills_staging_derived/dataset_metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/backfills_staging_derived/dataset_metadata.yaml
@@ -4,7 +4,4 @@ description: |-
 dataset_base_acl: derived
 user_facing: false
 labels: {}
-workgroup_access:
-- role: roles/bigquery.dataViewer
-  members:
-  - workgroup:mozilla-confidential
+workgroup_access: []
--- a/sql/moz-fx-data-shared-prod/backfills_staging_derived/dataset_metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/backfills_staging_derived/dataset_metadata.yaml
@@ -1,0 +1,10 @@
+friendly_name: Backfills Staging
+description: |-
+  Staging location for backfills
+dataset_base_acl: derived
+user_facing: false
+labels: {}
+workgroup_access:
+- role: roles/bigquery.dataViewer
+  members:
+  - workgroup:mozilla-confidential


sql/moz-fx-data-shared-prod/backfills_staging_derived/dataset_metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: Backfills Staging
+description: |-
+  Staging location for backfills
```
ANich, line 3: nit:
```suggestion
  Temporary staging location for backfill data before it replaces production data.
```

ANich (approved):

sql/moz-fx-data-shared-prod/backfills_staging_derived/dataset_metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: Backfills Staging
+description: |-
+  Temporary staging location for backfill data before it replaces production data.
+dataset_base_acl: derived
+user_facing: false
+labels: {}
+workgroup_access:
+- role: roles/bigquery.dataViewer
+  members:
+  - workgroup:mozilla-confidential
```
whd, line 10: Worth noting that from a policy perspective this restricts the data that can be backfilled via this dataset to only that data which is `workgroup:mozilla-confidential`. I have https://mozilla-hub.atlassian.net/browse/DSRE-1096 for thinking about how to better support DE more autonomously backfilling workgroup-confidential data.

In addition, it might be better to remove `workgroup:confidential` access at any rate since this is an implementation detail of the platform and we're not expecting users to query tables in this dataset. One way to do that is to make this dataset `derived_restricted` and remove `workgroup_access`, rendering the dataset only accessible to infrastructure (airflow). You'd probably want to add `workgroup:platform-infra/internal` (DE) as `dataEditors` as well, but that will require a companion PR in https://github.com/mozilla-services/cloudops-infra/ since that's a nonstandard configuration.

wwyc: thanks for the recommendation!  The tables in this dataset would need to have different restrictions depending on the prod table.  The users would need to query the table and validate the data.  I see your comment in the [tmp table proposal](https://docs.google.com/document/d/1wOGXxPYQDHDHAehyd5n9VoynNvF_e6Xyw3w_WhCDFEY/edit#) about having different ACLs in `mozdata`.  Do you have any recommendations on how that could be set up?

whd: In this case the dataset will only be provisioned in `moz-fx-data-shared-prod`, since it's a `derived` dataset. I think the problem of managing workgroup-confidential dataset access for backfills is something that will require some new processes as outlined in the referenced ticket. I could see being able to delegate workgroup access management to Airflow as a method of achieving this (i.e. locking this backfill dataset down entirely, and as part of backfill provisioning airflow would be responsible for making sure appropriate workgroups have access via metadata), but it would require major tooling and process updates.

I hope that the for common case of `workgroup:mozilla-confidential` data backfills we can proceed with provisioning this dataset as-is or with the minor restrictions I've mentioned above (the intent of which is to hide as much as possible implementation details of our platform from end users, even though as a matter of policy it is permissible and perhaps easier to manage if they analysis users have direct access to the data). Which is to say: I think it would be reasonable to merge this PR as-is with the caveat that you shouldn't use the dataset for backfilling any data other than `workgroup:mozilla-confidential` data without consulting DSRE first.

ANich: IIRC there's some machinery to use a table's workgroup as defined in `metadata.yaml` files rather than the enclosing `dataset_metadata.yaml` file. Could that be an option here? The backfill process will likely use the production table's metadata. (Perhaps that's what you meant by this part:

> (i.e. locking this backfill dataset down entirely, and as part of backfill provisioning airflow would be responsible for making sure appropriate workgroups have access via metadata)

 if so ignore this)

wwyc: @whd I see that workgroup access can be empty in the [dataset_metadata.yaml ](https://github.com/mozilla/private-bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/revenue_derived/dataset_metadata.yaml) file.  Any workgroup access can then be applied in the individual backfilled tables in this staging location according to the `metadata.yaml` files from each prod table.  Do you think that would work?

whd: > @whd I see that workgroup access can be empty in the [dataset_metadata.yaml ](https://github.com/mozilla/private-bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/revenue_derived/dataset_metadata.yaml)file. Any workgroup access can then be applied in the individual backfilled tables in this staging location according to the metadata.yaml files from each prod table. Do you think that would work?

That's the general idea. This is going to require some additional design because either Airflow needs to assume the responsibility of applying ACLs for these tables (requiring a new dataset base ACL) or ops infrastructure that applies ACLs needs to have some understanding of what backfill tables exist. Furthermore the table deployment and ACL deployment processes are currently relatively decoupled (the existence of metadata for a table doesn't guarantee that the table exists until artifact deployment runs successfully). This is mitigated by DSRE having a manual process of approving ACLs for now but for backfill use cases we would want to (a) apply the ACLs as soon as possible after the tables are created and (b) not necessarily require manual review from DSRE to deploy them.

Overall I think it should be possible to iterate on design here "as-needed" when workgroup confidential backfills arise. I just want to make sure DE is aware so that we don't inadvertently repeat a situation like https://mozilla-hub.atlassian.net/browse/DSRE-1088?focusedCommentId=606457 that involved some overbroad exposure of backfill data.

PR #3906, opened 2023-06-06
[DO NOT MERGE]feat(DENG-537): Updating fxa content events v1 query to use partitioned table

kik-kik:
# feat(DENG-537): Updating fxa content events v1 query to use partitioned table

- updated `fxa_content_events_v1` query to use partitioned source table
- updated `fxa_oauth_events_v1` query with a comment explaining why it still references a sharded fxa log table.

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_content_events_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_content_events_v1/query.sql
@@ -13,16 +13,8 @@ SELECT
     ) AS jsonPayload
   )
 FROM
-  `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_content_20*`
+  `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_content`
 WHERE
-  jsonPayload.type = 'amplitudeEvent'
+  DATE(`timestamp`) = @submission_date
+  AND jsonPayload.type = 'amplitudeEvent'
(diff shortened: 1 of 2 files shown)


sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_content_events_v1/query.sql
```diff
@@ -13,16 +13,8 @@ SELECT
     ) AS jsonPayload
   )
 FROM
-  `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_content_20*`
+  `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_content`
```
akkomar, line 16: Wasn't date-sharded table used here to avoid missing data resulting from schema incompatibilities detected in BQ Cloud Logging sink? I might be completely wrong but IIRC I heard @jbuck talking about this at some point.

kik-kik: In the past there was an issue when we tried to switch over to the partitioned table. This is why for this specific table we ended up reverting back to using the sharded table. There's still very small inconsistencies between the two tables. However, the error tables for both look almost identical (in terms of schema error counts) as per my inside the ticket (DENG-537).

kik-kik:
I believe due to the migration from AWS to GCP this change is no longer required as the tables referenced in these tables are no longer being populated with data. Furthermore, these are being de-scheduled via: https://github.com/mozilla/bigquery-etl/pull/4831

PR #3907, opened 2023-06-06
Ds 2944 create external adjust table redux

Marlene-M-Hirose:
Checklist for reviewer:


sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,307 @@
...
+            print(f"Running job {job.job_id}")
+            job.result()
+
+
+def main():
+    """Input data, call functions, get stuff done."""
+    parser = ArgumentParser(description=__doc__)
+    parser.add_argument("--date", required=True)
+    parser.add_argument("--project", default="moz-fx-data-shared-prod")
+    parser.add_argument("--dataset", default="analysis")
```
wwyc, line 265: Should this be `adjust_derived` dataset?

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,307 @@
...
+                                                    ][
+                                                        4
+                                                    ],
+                                                    "sessions": device["kpi_values"][5],
+                                                    "daus": device["kpi_values"][6],
+                                                    "waus": device["kpi_values"][7],
+                                                    "maus": device["kpi_values"][8],
+                                                }
+                                                fields_list.append(field_dict)
+    # print(f'finished filling up list for {adjust_list_part["app_name"]}')
```
wwyc, line 197: remove print comment?

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,307 @@
...
+def main():
+    """Input data, call functions, get stuff done."""
+    parser = ArgumentParser(description=__doc__)
+    parser.add_argument("--date", required=True)
+    parser.add_argument("--project", default="moz-fx-data-shared-prod")
+    parser.add_argument("--dataset", default="analysis")
+
+    args = parser.parse_args()
+
+    # print("Something might be happening")
```
wwyc, line 269: remove print comment?

sql/moz-fx-data-shared-prod/adjust/adjust_derived/view.sql
```diff
@@ -0,0 +1,6 @@
+CREATE OR REPLACE VIEW
+    `moz-fx-data-shared-prod.analysis.adjust_derived_v1`
+AS
+SELECT
+    *
+FROM `moz-fx-data-shared-prod.analysis.mhirose_adjust_derived_raw`
```
wwyc, line 6: this doesn't seem to match the destination table name?

sql/moz-fx-data-shared-prod/adjust/adjust_derived/view.sql
```diff
@@ -0,0 +1,6 @@
+CREATE OR REPLACE VIEW
+    `moz-fx-data-shared-prod.analysis.adjust_derived_v1`
```
wwyc, line 2: is `analysis` the right dataset?

wwyc, line 2: is `analysis` the right dataset?  May also want to confirm the table name?

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/metadata.yaml
```diff
@@ -0,0 +1,23 @@
...
+friendly_name: Data from external source Adjust
+description: >
+  This table represents data downloaded from Adjust
+owners:
+  - mhirose@mozilla.com
+labels:
+  incremental: false
+scheduling:
+  dag_name: bqetl_adjust
+  date_partition_parameter: null
```
wwyc, line 11: is your table expected to be partitioned by date?

Marlene-M-Hirose: Yes, it will be one table partitioned by date

sql/moz-fx-data-shared-prod/adjust/adjust_derived/view.sql
Marlene-M-Hirose: yes! the dataset will be adjust_derived. I wanted to make sure my logic was right before changing to the proper name

sql/moz-fx-data-shared-prod/adjust/adjust_derived/view.sql
Marlene-M-Hirose: You're right! I hadn't checked this part. I'm not sure a view is even necessary.

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
Marlene-M-Hirose: will do!

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
Marlene-M-Hirose: I will change this

dags.yaml
```diff
@@ -1066,6 +1066,28 @@ bqetl_adjust:
...
+      - telemetry-alerts@mozilla.com
+      - mhirose@mozilla.com
+    email_on_failure: true
+    email_on_retry: true
+    end_date: null
+    owner: mhirose@mozilla.com
+    retries: 2
+    retry_delay: 30m
+    start_date: '2023-07-06'
+  description: Derived tables built on Adjust data downloaded from https://api.adjust.com/kpis/v1/<app_token>. Using mhirose's API token
```
lucia-vargas-a, line 1083: More detail on why we need to use your credentials could be useful in the future?

sql/moz-fx-data-shared-prod/adjust/adjust_derived/view.sql
```diff
@@ -0,0 +1,6 @@
+CREATE OR REPLACE VIEW
+    `moz-fx-data-shared-prod.adjust_derived.adjust_derived_v1`
```
lucia-vargas-a, line 2: Was the view bqetl-generated? Normally it wouldn't include the version, more like:

```suggestion
    `moz-fx-data-shared-prod.adjust_derived.adjust_derived`
```

dags/bqetl_adjust_derived.py
```diff
@@ -0,0 +1,56 @@
...
+    "end_date": None,
+    "email": ["telemetry-alerts@mozilla.com", "mhirose@mozilla.com"],
+    "depends_on_past": False,
+    "retry_delay": datetime.timedelta(seconds=1800),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
+}
+
+tags = ["impact/tier_2", "repo/bigquery-etl"]
```
lucia-vargas-a, line 37: Just to check, is this data not tier 1? --which would imply data and the highest level of support.

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,305 @@
...
+{"app_name":"<specific app name from Adjust dash>". "app_token":"<unique Adjust app token>}"
+Mozilla has 16 apps on Adjust.
+"""
+
+APP_TOKEN_LIST = os.environ.get("ADJUST_APP_TOKEN_LIST")
+
+"""This is Marlene's personal Adjust API Token. We don't have a special service account API Token."""
+API_TOKEN = os.environ.get("ADJUST_API_TOKEN")
+
+CSV_FIELDS = []
```
lucia-vargas-a, line 26: Is this line required?

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,305 @@
...
+        dict_writer.writeheader()
+        dict_writer.writerows(json_data)
+
+
+def download_adjust_kpi_data(date, api_token, app_token):
+    """Download data from Adjust - API token and APP tokens are called here."""
+    start_date = date
+    end_date = date
+
+    kpis = [
```
lucia-vargas-a, line 74: Would kpis and groupings fit better outside the function, perhaps after CSV_FIELDS?

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,305 @@
...
+def clean_json(adjust_response_text, adjust_list_part):
+    """JSON sometimes has missing keys, need to clean up the data."""
+    fields_list = []
+    with tempfile.NamedTemporaryFile() as tmp_json:
+        with open(tmp_json.name, "w") as f_json:
+            f_json.write(adjust_response_text)
+
+            query_export = read_json(f_json.name)
+
+            for date in query_export["result_set"]["dates"]:
```
lucia-vargas-a, line 117: You could separate the code from here below into another function e.g. get_query_fields() that would run for each date. Mainly for readability and to separate what each part of the code does. Optionally you can do that in a later PR, your call!

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,305 @@
...
+                ],
+            )
+
+            # Table names are based on the app name seen in the Adjust dashboard"
+            destination = f"{project}.{dataset}.adjust_deliverables_v1_${partition}"
+            print(destination)
+
+            job = client.load_table_from_file(f_csv, destination, job_config=job_config)
+
+            print(f"Running job {job.job_id}")
```
lucia-vargas-a, line 256: Perhaps merging line 252 and this one?
```suggestion
            print(f"Writing adjust data for all apps to {destination}. BigQuery job ID: {job.job_id}")
```

dags.yaml
```diff
@@ -1066,6 +1066,28 @@ bqetl_adjust:
...
 
+
+bqetl_adjust_derived:
+  default_args:
+    depends_on_past: false
+    email:
+      - telemetry-alerts@mozilla.com
+      - mhirose@mozilla.com
+    email_on_failure: true
+    email_on_retry: true
```
kik-kik, line 1077: I'd generally suggest not having `email_on_retry`, it created noise and it's very unlikely anyone would perform any action at that point so it's mostly just spamming your mailbox.

dags.yaml
```diff
@@ -1066,6 +1066,28 @@ bqetl_adjust:
...
+    owner: mhirose@mozilla.com
+    retries: 2
+    retry_delay: 30m
+    start_date: '2023-07-06'
+  description: Derived tables built on Adjust data downloaded from https://api.adjust.com/kpis/v1/<app_token>. Using mhirose's API token
+  repo: bigquery-etl
+  schedule_interval: 0 4 * * *
+  tags:
+    - impact/tier_2
+    - repo/bigquery-etl
```
kik-kik, line 1088:
```suggestion
```

in the bqetl repo this tag should be automatically added during the generation process so no need to include it here.

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/metadata.yaml
```diff
@@ -0,0 +1,23 @@
...
+description: >
+  This table represents data downloaded from Adjust
+owners:
+  - mhirose@mozilla.com
+labels:
+  incremental: false
+scheduling:
+  dag_name: bqetl_adjust_derived
+  date_partition_parameter: submission_date
+  parameters: ["submission_date:DATE:{{ds}}"]
```
kik-kik, line 12: I think if you specify the `date_partition_parameter` then you don't need `parameters` here.

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/metadata.yaml
```diff
@@ -0,0 +1,23 @@
...
+bigquery:
+  time_partitioning:
+    type: day
+    field: submission_date
+    require_partition_filter: false
+external_data:
+  format: csv
+  source_uris:
+    -
```
kik-kik, line 21: Is this meant to be empty? In which case can we just remove the key completely?

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,305 @@
...
+from argparse import ArgumentParser
+
+import requests
+from google.cloud import bigquery
+
+API_URI = "api.adjust.com"
+ENDPOINT = "kpis"
+API_VERSION = "v1"
+
+"""This is a list of dictionaries.
```
kik-kik, line 15: I'm not sure I understand what this comment is trying to tell me. Also, rather than saying that Mozilla has 16 apps in adjust I'd prefer you to mention here which apps this logic might be running for. If there's a link where you could find a list of the apps available and and their descriptions you could include that here.

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,305 @@
...
+Keys are the app_name, and app_token in the form:
+{"app_name":"<specific app name from Adjust dash>". "app_token":"<unique Adjust app token>}"
+Mozilla has 16 apps on Adjust.
+"""
+
+APP_TOKEN_LIST = os.environ.get("ADJUST_APP_TOKEN_LIST")
+
+"""This is Marlene's personal Adjust API Token. We don't have a special service account API Token."""
```
kik-kik, line 23:
```suggestion
```

This piece of logic is not dependent on who's credentials are in the env variable. I'd consider removing this comment as it could quickly become out of date and misleading.

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
kik-kik, line 26:
```suggestion
```

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,305 @@
...
+
+
+def read_json(filename: str) -> dict:
+    """Read JSON file."""
+    with open(filename, "r") as f:
+        data = json.loads(f.read())
+    return data
+
+
+def write_dict_to_csv(json_data, filename):
```
kik-kik, line 61: I noticed in `read_json()` definition you're using type hinting. Since this is new code it'd be great if we could use it for all function definitions in this file :)

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,305 @@
...
+    """Write a dictionary to a csv."""
+    with open(filename, "w") as out_file:
+        dict_writer = csv.DictWriter(out_file, CSV_FIELDS)
+        dict_writer.writeheader()
+        dict_writer.writerows(json_data)
+
+
+def download_adjust_kpi_data(date, api_token, app_token):
+    """Download data from Adjust - API token and APP tokens are called here."""
+    start_date = date
```
kik-kik, line 71: If these are expected to be the same, it could also be put on a single line like so:
`end_date = start_date = date`

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,305 @@
...
+    ]
+    # getting overview metrics for different kpis / Deliverables
+    url = f"https://api.adjust.com/kpis/v1/{app_token}"  # overview
+    url_params = f"start_date={start_date}&end_date={end_date}&kpis={','.join(kpis)}&grouping={','.join(groupings)}"
+    headers = {
+        "Authorization": f"Token token={api_token}",
+    }
+
+    response = requests.get(url, headers=headers, params=url_params)
+
```
kik-kik, line 104: Could add here `response.raise_for_status()` call to check that we got a success response back from the server. Otherwise, an exception would be raised and the application would exit with an error code in this case that would be reasonable since the rest of the logic depends on data being returned from the server.

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,305 @@
...
+            query_export = read_json(f_json.name)
+
+            for date in query_export["result_set"]["dates"]:
+                r_date = date["date"]
+                for app in date["apps"]:
+                    r_app = app["name"]
+                    r_app_token = app["token"]
+                    for network in app["networks"]:
+                        try:
+                            r_network = network["name"]
```
kik-kik, line 124: Looks like we could avoid the `try: except` here by using the `.get()` method. 

For example:
`r_network = network.get("name", "no network name")`

In this case if network dict has no key `name` then the second argument would be used as the default value.

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,305 @@
...
+
+    partition = f"{date}".replace("-", "")
+
+    with tempfile.NamedTemporaryFile() as tmp_csv:
+        with open(tmp_csv.name, "w+b") as f_csv:
+            write_dict_to_csv(csv_data, f_csv.name)
+
+            client = bigquery.Client(project)
+
+            print("hey, I'm about to config the job\n")
```
kik-kik, line 213: Is this print statement a useful log?

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,305 @@
...
+    parser.add_argument("--date", required=True)
+    parser.add_argument("--project", default="moz-fx-data-shared-prod")
+    parser.add_argument("--dataset", default="adjust_derived")
+
+    args = parser.parse_args()
+
+    app_list = json.loads(APP_TOKEN_LIST)
+
+    # Cycle through the apps to get the relevant kpi data
+    for num in range(len(app_list)):
```
kik-kik, line 272: A more pythonic way of iterating through the app_list would be to do:
```
for app in app_list:
    [code to handle each app]
```
This will allow you to avoid having to access the app using the index from the whole object each time you need to access an app specific value:

```
[before]
print(
    f'This is data for {app_list[num]["app_name"]} - {app_list[num]["app_token"]}'"
)
[after]
print(
    f'This is data for {app_list["app_name"]} - {app_list["app_token"]}'"
)
```

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
lucia-vargas-a: In this function you can use the [recommended practice for unexisting keys](https://docs.quantifiedcode.com/python-anti-patterns/correctness/not_using_get_to_return_a_default_value_from_a_dictionary.html#best-practice).

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,305 @@
...
+            continue
+
+        """ This section writes the tmp json data into a temp CSV file which will then
+            be put into a BigQuery table"""
+        if data:
+            upload_to_bigquery(
+                data, args.project, args.dataset, app_list[num], args.date
+            )
+        else:
+            print("no data to upload")
```
kik-kik, line 301: Perhaps we can include a bit more information like which app, etc?

sql/moz-fx-data-shared-prod/adjust/adjust_derived/view.sql
```diff
@@ -0,0 +1,6 @@
+CREATE OR REPLACE VIEW
```
kik-kik, line 1: The CI appears to fail due to the folder structure where this view is found.

The folder should be:
`sql/moz-fx-data-shared-prod/adjust/view.sql`

Instead of:
`sql/moz-fx-data-shared-prod/adjust/adjust_derived/view.sql`

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/metadata.yaml
```diff
@@ -0,0 +1,23 @@
+---
```
kik-kik, line 1: Should we consider renaming the table name here from `adjust_derived_v1` to something a bit more representative of the content like `adjust_app_kpi_v1` or `adjust_app_overview_v1` (these are just example, just something that describes the contents well)?

dags/bqetl_adjust_derived.py
Marlene-M-Hirose: The data isn't super high priority, so that is why I made it tier 2. The original table/DAG is tier 2

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
Marlene-M-Hirose: line 26? probably not. Will delete.

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
Marlene-M-Hirose: Thank you for this! Will update code

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
Marlene-M-Hirose: I was being explicit as to what the list contains - the set up is important for how the code runs. I.e. you need to have the list in the form of [{key:app_name, token:app_token}] in order for the code to work.

I've added a bit more clarity

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
Marlene-M-Hirose: George recommended that I make a note as to whose credentials are being used. If in future we have to make changes and update/take out the comment, it can be done then

Marlene-M-Hirose: I've made note of the use of my API token in Airflow. I'm just reiterating that we are using my token here as a way to remind people/future people

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
Marlene-M-Hirose: I prefer them to be separate. Helps my brain keep them separate

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
Marlene-M-Hirose: thank you! I will modify this!

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
Marlene-M-Hirose: wait, wouldn't it be:

> print(
>     f'This is data for {app["app_name"]} - {app["app_token"]}'"
> )

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
Marlene-M-Hirose: Sounds good. Will move it

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
Marlene-M-Hirose: It would be great to do that, but... the write_dict_to_csv doesn't return anything, so how would type hinting work here?

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/metadata.yaml
Marlene-M-Hirose: That is a great question. I am leaning more toward 'adjust_derived' because it's keeping with the pattern that Mozilla has established. Also, the downloaded data is pre-aggregated, so it feels derived to me. And it's called KPI based on Adjust's definition, not Mozilla's.

dags.yaml
```diff
@@ -1066,6 +1066,30 @@ bqetl_adjust:
     - impact/tier_2
     - repo/bigquery-etl
 
+
+bqetl_adjust_derived:
```
scholtzan, line 1070: There seems to be an `bqetl_adjust` DAG already. Could that be reused to schedule the newly added query? (it seems to have the same schedule interval and tier)

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/metadata.yaml
```diff
@@ -0,0 +1,23 @@
...
+friendly_name: Data from external source Adjust
+description: >
+  This table represents data downloaded from Adjust. Please note that the API token used is Marlene Hirose's personal Adjust API token
+owners:
+  - mhirose@mozilla.com
+labels:
+  incremental: false
+scheduling:
+  dag_name: bqetl_adjust_derived
```
scholtzan, line 10:
```suggestion
  dag_name: bqetl_adjust
```

unless there is a good reason for having an separate DAG

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/metadata.yaml
```diff
@@ -0,0 +1,23 @@
...
+  date_partition_parameter: submission_date
+  parameters:
```
scholtzan, line 12: CI is failing because there are some parameters missing here

dags.yaml
Marlene-M-Hirose: I'll combine the two

Marlene-M-Hirose: The other one isn't in use anymore, should I just replace the code?

scholtzan: I suggest removing `scheduling` from this query: https://github.com/mozilla/bigquery-etl/blob/e719266146fb980fdab1f7a2690a64384b523a4c/sql/moz-fx-data-marketing-prod/adjust_derived/firefox_mobile_installs_v1/metadata.yaml#L8C1-L10

scholtzan: And then keep `bqetl_adjust` in favour of `bqetl_adjust_derived`

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/metadata.yaml
Marlene-M-Hirose: Are the parameters needed?

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/metadata.yaml
```diff
@@ -0,0 +1,23 @@
...
+bigquery:
+  time_partitioning:
+    type: day
+    field: submission_date
+    require_partition_filter: false
+external_data:
+  format: csv
+  source_uris:
```
Marlene-M-Hirose, line 20: Should I add the Adjust API source URIs here? There are 11 with the APP tokens - we don't want to add those tokens to git

scholtzan: `external_data` can be removed here. It's only used in cases where you simply import a CSV file into a BigQuery table. In your case you have a script with more complex logic that handles the import

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/metadata.yaml
scholtzan: not in this case, instead you need to specify script arguments, like this: https://github.com/mozilla/bigquery-etl/blob/dc482ad8d5ee6de28c0d5a997e83d2a0a563981f/sql/moz-fx-data-marketing-prod/iprospect/adspend_v1/metadata.yaml#L15

```suggestion
  arguments: ["--date", "{{ ds }}"]
```

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,278 @@
...
+        if query_export is not None:
+            # This section writes the tmp json data into a temp CSV file which will then be put into a BigQuery table
+            data = clean_json(query_export)
+            print("finished json_read")
+            upload_to_bigquery(
+                data, args.project, args.dataset, app["app_name"], args.date
+            )
+        else:
+            print(f'no data for {app["app_name"]} today')
+            continue
```
kik-kik, line 274:
```suggestion
```

Redundant `continue`, we're already at the end of the loop.

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,278 @@
...
+        json_file = download_adjust_kpi_data(args.date, API_TOKEN, app["app_token"])
+
+        data = []
+
+        query_export = check_json(json_file.text)
+
+        if query_export is not None:
+            # This section writes the tmp json data into a temp CSV file which will then be put into a BigQuery table
+            data = clean_json(query_export)
+            print("finished json_read")
```
kik-kik, line 268: In the current form the log may not be helpful, if you think it's useful to have it try to make the message more self contained. json_read finished for? Is this good? Generally, instead of print I'd suggest using the `logging` library as it provides different log levels such as `debug`, `info`, `warning`, and `error` which helps to indicate if this log is just informational or does it suggest something is off.

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,278 @@
...
+    """Script will return an empty dictionary for apps on days when there is no data. Check for that here."""
+    with tempfile.NamedTemporaryFile() as tmp_json:
+        with open(tmp_json.name, "w") as f_json:
+            f_json.write(adjust_response_text)
+            try:
+                query_export = read_json(f_json.name)
+            except (
+                ValueError
+            ):  # ex. json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
+                return None
```
kik-kik, line 118: Should the application exit at this stop execution at this point? The error suggests like there might be a problem with the data?

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
kik-kik: You can always say that the function return value is `None`:
```
def my_fun(arg: str) -> None:
   [function_body]
```

kik-kik: And I think it's nice to include as it indicates that it does not return anything.

sql/moz-fx-data-shared-prod/adjust_derived/dataset_metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: Adjust Derived
+description: |-
+  Derived tables for Adjust data
+dataset_base_acl: derived
+user_facing: false
+labels: {}
+workgroup_access:
+- role: roles/bigquery.dataViewer
+  members:
+  - workgroup:mozilla-confidential  # TODO: @marlene please make sure this is the correct access group.
```
kik-kik, line 10: If this is the correct access group please remove the TODO comment here.

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
```diff
@@ -0,0 +1,278 @@
...
+"""The APP_TOKEN_LIST is a list of dictionaries.
+Keys are the app_name, and app_token in the form:
+{"app_name":"<specific app name from Adjust dash>". "app_token":"<unique Adjust app token>}"
+Look here to see the apps we are tracking: https://dash.adjust.com/#/.
+It is important to maintain this list in order for the script to work especially in the case of new apps
+being added to track in Adjust
+"""
+
+APP_TOKEN_LIST = os.environ.get("ADJUST_APP_TOKEN_LIST")
+
```
kik-kik, line 24:
```suggestion
```

kik-kik: nitpick

kik-kik (approved):
r+wc - please resolve any open conversations and ensure that the correct access group is specified for this data.

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/query.py
Marlene-M-Hirose: This was the solution to yesterday's issue where if there is no data, the script would get an empty dictionary. By returning None, there is an 'issue' of no data, but we don't always get data for each day.

sql/moz-fx-data-shared-prod/adjust_derived/dataset_metadata.yaml
Marlene-M-Hirose: removed. Thanks

sql/moz-fx-data-shared-prod/adjust_derived/adjust_derived_v1/metadata.yaml
Marlene-M-Hirose: it was slightly different than what we had previously. One was in marketing-prod, this one is in shared prod. Will use the old DAG

sql/moz-fx-data-shared-prod/adjust/view.sql
```diff
@@ -0,0 +1,6 @@
+CREATE OR REPLACE VIEW
+    `moz-fx-data-shared-prod.adjust_derived.adjust_derived`
+AS
+SELECT
+    *
+FROM `moz-fx-data-shared-prod.adjust_derived.adjust_derived_v1`
```
scholtzan, line 6: The `view_sql` file needs to be in a folder `adjust_derived`

PR #3920, opened 2023-06-09
Bug 1836946 - Basic ETL task for ingesting Pocket event data

badboy:
Starting the work to build the ETL task.
Still waiting for some details, e.g. if this is the right bucket.
Schema also not yet final.


sql/moz-fx-data-shared-prod/pocket_derived/events_v1/query.py
```diff
@@ -0,0 +1,112 @@
...
+  ) AS event_extra
+FROM
+  {tmp_table} AS t
+        """,
+        job_config=bigquery.QueryJobConfig(
+            destination=".".join(
+                [
+                    args.project,
+                    args.destination_dataset,
+                    f"{args.destination_table}${args.date.strftime('%Y%m%d')}",
```
kik-kik, line 100: Do we want to persist older data or do we only care about the most recent import?

sql/moz-fx-data-shared-prod/pocket_derived/events_v1/query.py
```diff
@@ -0,0 +1,112 @@
...
+
+    # The Pocket-side ETL uses dbt and includes partial days, so need some handling
+    # for that; see https://bugzilla.mozilla.org/show_bug.cgi?id=1695336#c33
+    #
+    # The Airflow DAG run for 2021-03-01 expects to process complete results up
+    # through 2021-03-01, and executes during UTC day 2021-03-02. That run
+    # will pick up a Pocket result labeled as 2021-03-02, so we need to add an
+    # offset here. Then we make sure to drop partial data for 2021-03-02
+    # in the final query.
+    publish_date = (args.date + timedelta(days=1)).isoformat()
```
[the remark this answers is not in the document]
badboy: (this PR is not yet ready and so far mostly copied from the other existing query)

No, if we use `date.today()` we can't use this to backfil for a specific day.

badboy:
We should have the Pocket data dump tomorrow, so that we can land this afterwards.

sql/moz-fx-data-shared-prod/pocket/events/view.sql
```diff
@@ -0,0 +1,7 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.pocket.events`
+AS
+SELECT
+  * EXCEPT (submission_date)
```
relud, line 5: I don't understand why the time partitioning field is being excluded from the view. i see that submission_timestamp also exists, so you might want only one of the two to be available, but then i would expect submission_timestamp to be the time partitioning field.

relud (approved):
r+ regardless of comment

scholtzan (approved):
Looks like the DAG needs to be regenerated since the ordering of the task changed. Something like `./bqetl dag generate bqetl_pocket` should suffice

badboy:
Oh I see, I manually moved it, not realizing the position was auto-generated.

badboy:
I've been running this manually using

```
python3 query.py --project mozdata --destination_dataset "tmp" --destination_table "janerik_pocket_events_v1" --date 2023-06-27
```

with the schema.yaml turned into json and

```
bq mk --project_id mozdata --table --schema schema.json --time_partitioning_field submission_timestamp --time_partitioning_type DAY --require_partition_filter=false tmp.janerik_pocket_events_v1
```

to create the temporary table.
That just works.

Pocket Data is generated at 01:00 UTC for the previous day. Running this at 12:00 along with the other pcoket tasks should be fine though.
I'm still slightly confused by the `ds` parameter though.
If this task runs today, `2023-06-29 12:00`, will `ds` be `2023-06-29` or will it be `2023-06-28`?

badboy:
I did trigger the task from this repo (through github actions). 
`validate-views` fails with some permission errors

scholtzan:
> validate-views fails with some permission errors

Yeah, those are unrelated

scholtzan:
That should all happen automatically

PR #3933, opened 2023-06-12
Simplify GLAM template for getting the latest version

Iinh:
Okay, second (and hopefully final) attempt at trying to simplify the latest_version template.

In this change we're using `buildhub2` data to make sure no bad version data gets in again. Also, `@submission_date` is added to template so the query is idempotent per @akkomar's suggestion.

Also, channel value is always `*` (channel values are switched to * in Glean pipeline because channels are processed in separate tables).

I have tested this query with some sample data and it's working as expected.

Checklist for reviewer:

### DIFF
--- a/bigquery_etl/glam/templates/latest_versions_v1.sql
+++ b/bigquery_etl/glam/templates/latest_versions_v1.sql
@@ -1,36 +1,20 @@
-{{ header }}
-WITH extracted AS (
+  {{ header }}
+WITH
+  extracted AS (
   SELECT
-    client_id,
-    channel,
-    app_version
+    build.`target`.channel AS channel,
+    MAX(mozfun.norm.extract_version(build.`target`.version,
+        'major')) AS latest_version,
   FROM
-    {{ source_table }}
+    `moz-fx-data-shared-prod.telemetry.buildhub2`
   WHERE
-    submission_date
-    BETWEEN DATE_SUB(@submission_date, INTERVAL 28 DAY)
-    AND @submission_date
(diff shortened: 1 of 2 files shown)


bigquery_etl/glam/templates/latest_versions_v1.sql
```diff
@@ -1,36 +1,20 @@
...
-  FROM
-    extracted
+    build.`source`.product = "firefox"
+    AND build.`target`.channel = {{ app_id_channel }}
+    AND DATE(build.build.date) <= @submission_date
   GROUP BY
-    channel,
-    app_version
-  HAVING
-    COUNT(DISTINCT client_id) > 5
```
akkomar, line 25: By getting rid of this condition we might allow some builds fewer than 5 users leak into the dataset.

akkomar (changes requested):
Looking at this again I realized there are two issues:
1. Lack of client count filter might leak in some builds that we currently filter out (see inline comment)
2. It seems like buildhub data is [already used in an upstream task for filtering](https://github.com/mozilla/bigquery-etl/blob/b2a06b8779f89a1ea52b3fa49a52703ac55fd221/bigquery_etl/glam/templates/clients_daily_scalar_aggregates_v1.sql#L3-L9) (which was added by me at some point but I have forgot about this since).

akkomar (approved):
LGTM. As discussed the buildhub filter in the upstream query is not equivalent to the change proposed here as it only takes into account build ids, not the version numbers.

PR #3936, opened 2023-06-13
DENG-806 Added backfill process cli command

wwyc:
[DENG-806](https://mozilla-hub.atlassian.net/browse/DENG-806):

* Added `backfill process` cli command to process backfill entries from backfill.yaml file.  All backfills are written to backfill staging location:  
`moz-fx-data-shared-prod`.`backfills_staging_derived`.`{table}_{entry_date}`
* Added `backfill schedule` cli command to get backfill entries from backfill.yaml file that require processing.
* Added tests for backfil utils

[DENG-806]: https://mozilla-hub.atlassian.net/browse/DENG-806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3941, opened 2023-06-14
DENG-984 Changed backfill staging dataset name

wwyc:
[DENG-984](https://mozilla-hub.atlassian.net/browse/DENG-984):

Dataset was created in #3904 to store backfill data.

This PR updates the naming of dataset from `backfills_staging_derived` to `backfills_staging`.

[DENG-984]: https://mozilla-hub.atlassian.net/browse/DENG-984?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


whd:
Current dataset naming conventions require that this dataset be suffixed with `_derived`. We can make an exception for this dataset (possibly by creating a new suffix `_staging`) but it will require some configuration changes in https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/tf/modules/bigquery/bqetl_tfvars.py#L25.

https://docs.google.com/document/d/1tXKq37jo13-EdPmwYOxvtS8mrbNb8rhZWKGKTiAyJBs/edit#heading=h.v9ip0n681xop contains some historical background on the `_derived` suffix.

ANich (approved):

ANich (approved):
(see my original comment here https://github.com/mozilla/bigquery-etl/pull/3936#discussion_r1229633081), the `_derived` name is slightly clunky given the purpose of this dataset, if only because it should only contain temporary backfill tables for validation. If the rename is possible then 👍 but I'll defer to @whd RE: ops requirements

whd:
> (see my original comment here [#3936 (comment)](https://github.com/mozilla/bigquery-etl/pull/3936#discussion_r1229633081)), the `_derived` name is slightly clunky given the purpose of this dataset, if only because it should only contain temporary backfill tables for validation. If the rename is possible then +1 but I'll defer to @whd RE: ops requirements

I checked with :scholtzan and we're going to continue to require `_derived` to match https://docs.telemetry.mozilla.org/cookbooks/bigquery/querying.html?highlight=derived#projects-with-bigquery-datasets and existing convention (which while somewhat obscure is consistent). For this case the hope is that tooling will mostly abstract away the dataset name eventually.

I'm closing this since the only change was going to be to the dataset name.

PR #3943, opened 2023-06-14
Process subsequent checks for a table if a prior check fails for that table

gleonard-m:
Checklist for reviewer:

**Summary:**
If a checks.sql file contains multiple checks separated by `;` each check will be executed regardless if a previous check failed.

### DIFF
--- a/bigquery_etl/cli/check.py
+++ b/bigquery_etl/cli/check.py
@@ -123,24 +123,37 @@ def _run_check(
         **{"dataset_id": dataset_id, "table_name": table},
         **parameters,
     }
+    rendered_result = render_template(
+        checks_file.name,
+        template_folder=str(checks_file.parent),
+        templates_dir="",
+        format=False,
+        **jinja_params,
+    )
+    checks = rendered_result.split(";")
+    seek_location = 0
+    check_failed = False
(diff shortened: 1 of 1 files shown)


bigquery_etl/cli/check.py
```diff
@@ -123,24 +123,37 @@ def _run_check(
...
         **parameters,
     }
+    rendered_result = render_template(
+        checks_file.name,
+        template_folder=str(checks_file.parent),
+        templates_dir="",
+        format=False,
+        **jinja_params,
+    )
+    checks = rendered_result.split(";")
```
scholtzan, line 133: I recommend using `sqlparse` here instead to determine the individual SQL statements. Similar to how we use it here: https://github.com/mozilla/bigquery-etl/blob/d82acc18562df147c1b2e418498ff2f74067d8d8/bigquery_etl/routine/parse_routine.py#L143

Otherwise we might run into issues where these tests unexpectedly break because `;` are used elsewhere, like in a comment.

gleonard-m: Oops missed that, updated.

scholtzan (approved):

PR #3949, opened 2023-06-15
Regex for matching UDF names

scholtzan:
This fixes the failing stage deploys. Stage changes UDF names to have escaped dataset and UDF names.

### DIFF
--- a/bigquery_etl/routine/parse_routine.py
+++ b/bigquery_etl/routine/parse_routine.py
@@ -86,8 +86,8 @@ class RawRoutine:
 
         if value is None or value != persistent_name:
             raise ValueError(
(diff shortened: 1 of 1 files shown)


bigquery_etl/routine/parse_routine.py
```diff
@@ -146,6 +146,7 @@ def from_file(cls, path):
         dataset = filepath.parent.parent.name
         project = filepath.parent.parent.parent
 
+        persistent_name_re = rf"`?{dataset}`?.`?{name}`?"
```
scholtzan, line 149: UDF names in stage look like ``` `fenix`.`app_info`  ```

edugfilho (approved):

PR #3962, opened 2023-06-19
DS-2947 clients_first_seen_v2

lucia-vargas-a:
This PR implements [DS-2929](https://mozilla-hub.atlassian.net/browse/DS-2929) to redefine new profiles for Firefox Desktop based on 3 pings (main, new profile and first_shutdown). Previously, the new profiles were based only on the main ping.

This are the [caveats to consider when reviewing this new version of clients_first_seen](https://docs.google.com/document/d/1qMkTUIruTOxTGRj7S60nEsyNjGHYAuVJ51-53NZxnd4/edit?usp=sharing) and the [requirements doc](https://docs.google.com/document/d/1B-zTNwschQgljij19Ez104qRV5RUC9zwjSN9tc1C8vc/edit) for further details.

[DS-2929]: https://mozilla-hub.atlassian.net/browse/DS-2929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Sample data initialized for sample_id = 0 in `moz-fx-data-shared-prod.analysis.clients_first_seen_v2`.


sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/init.sql
```diff
@@ -0,0 +1,633 @@
...
+    telemetry.first_shutdown
+  WHERE
+    DATE(submission_timestamp) >= starting_date
+  GROUP BY
+    client_id
+),
+main_ping AS (
+  SELECT
+    client_id AS client_id,
+    MIN(sample_id) AS sample_id,
```
lucia-vargas-a, line 168: Using submission_timestamp_min to allow for comparing the 3 pings, which wouldn't return an exact result using submission_date. Thius field is [calculated in the source query](https://github.com/mozilla/bigquery-etl/blob/1418e2018b45f5536b0a797ffdadfaf8502a1bbf/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql#L1166) as the the first timestamp of the day, [doesn’t return null values and matches the submission_date](https://sql.telemetry.mozilla.org/queries/92792/source)

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/init.sql
```diff
@@ -0,0 +1,634 @@
+--Query to initialize clients_first_seen_v2
+-- Earliest ping dates for each ping used in the query.
+DECLARE earliest_new_profile_date DATE DEFAULT '2017-06-26';
+
+DECLARE earliest_main_date DATE DEFAULT '2016-03-12';
+
+DECLARE earliest_shutdown_date DATE DEFAULT '2018-10-30';
+
+-- Starting date to query pings.
+DECLARE starting_date DATE DEFAULT '2020-01-01';
```
lucia-vargas-a, line 10: The current start date for clients_first_seen_v1 is '2020-01-01'. Documenting the available dates above in case older data is required.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -0,0 +1,692 @@
...
+    (client_id)
+  WHERE
+    client_id IS NOT NULL
+),
+_previous AS (
+  SELECT
+    *
+  FROM
+    `moz-fx-data-shared-prod.telemetry_derived.clients_first_seen_v2`
+)
```
gleonard-m, line 575: 1. Do we know which fields are initially null then available in a later ping and how much later that field value comes in?  
2. Do we want to consider a hard cap instead of all rows?  Only update 1 week of _previous?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -0,0 +1,692 @@
...
+  COALESCE(_previous.first_seen_date, _current.first_seen_date) AS first_seen_date,
+  COALESCE(_previous.submission_timestamp, _current.submission_timestamp) AS submission_timestamp,
+  COALESCE(_previous.architecture, _current.architecture) AS architecture,
+  COALESCE(_previous.app_build_id, _current.app_build_id) AS app_build_id,
+  COALESCE(_previous.app_display_version, _current.app_display_version) AS app_display_version,
+  COALESCE(_previous.app_name, _current.app_name) AS app_name,
+  COALESCE(_previous.vendor, _current.vendor) AS vendor,
+  COALESCE(_previous.app_version, _current.app_version) AS app_version,
+  COALESCE(_previous.xpcom_abi, _current.xpcom_abi) AS xpcom_abi,
+  COALESCE(_previous.experiments_key, _current.experiments_key) AS experiments_key,
```
gleonard-m, line 588: Should `_previous` experiment fields be updated with `_current`?  Do NULL experiment fields mean that the new profile was not part of an experiment?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
lucia-vargas-a: Yes, [this query](https://sql.telemetry.mozilla.org/queries/92681/source) is tracking the updates per column of existing records on the current_date. Or in other words how many records from X days ago were updated today. The most updated column is below 0,5% records updated.
On option we talk about with Leif is to set an "update window", where records are updated with later pings only up to one week after the `first_seen_date`. We wanted to track some data for that. Do you think one week window update is a good approach?

gleonard-m: I do agree with a window approach.

Which column is the most updated and how did you determine the 0,5%?  Shouldn't the % be a list of %?  

E.g for 2023-06-12 **(day0)**
- 0,5% records on 2023-06-11 **(day-1)** were updated
- 0,3% records on 2023-06-10 **(day-2)** were updated 
- ...

I would expect there to be a drop off and that should mark the window size?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/init.sql
lucia-vargas-a: Talking to George, we will update all available data to support acquisition analysis.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
lucia-vargas-a: That's probably the case, that the new profile was not part of an experiment.
Then the question would be, is it the requirement to retrieve the first experiment that each client was part of, or only if it had an experiment on or near the first_seen_date? cc/ @irrationalagent

gkabbz:
@SuYoungHong would be great if we could get your eyes on the schema as making changes once the table is built will have to wait for a v3.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
irrationalagent: I don't think we would ever use this particular table for determining what the order of experiment enrollments was for a particular client. That being said, I think that what we want here would be closer to the latter (not their first ever experiment, but their experiment status when they were new)

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -0,0 +1,1125 @@
...
+  WHERE
+    submission_date = @submission_date
+  GROUP BY
+    client_id
+),
+_current AS (
+  SELECT
+    client_id,
+    CAST(
+      mozfun.norm.get_earliest_value(
```
fbertsch, line 375: These `earliest_value` dimensions are getting us to a place where we may have different fields from different pings. I'm not necessarily against that, but I wonder if there's a simpler way here.

Can we limit to just fields that are common between all the pings (a small number), and take _all_ the fields from that ping alone? That would be simpler to read and reason about. Was there a requirement to include all of these fields?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
lucia-vargas-a: The query is now updating some of the columns on a 7 day window, not experiments or environment captured-columns.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
lucia-vargas-a: Sounds good, in that case the query is okay retrieving the first experiment and not updating it.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -0,0 +1,1125 @@
+-- Query for telemetry_derived.clients_first_seen_v1
+WITH new_profile_ping AS (
+  SELECT
+    client_id AS client_id,
+    MIN(sample_id) AS sample_id,
+    MIN(submission_timestamp) AS submission_timestamp,
+    ARRAY_AGG(application.architecture IGNORE NULLS ORDER BY submission_timestamp ASC)[
```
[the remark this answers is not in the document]
lucia-vargas-a: The code is cleaner.
In terms of efficiency, running the new_profile subquery for 10 days btw '2023-06-01' and '2023-06-10' results in ANY_VALUE taking 16 secs vs. ARRAY_AGG 33 secs (twice the time). 
My concern was with [ANY_VALUE behaves as if RESPECT NULLS is specified; rows for which expression is NULL are considered and may be selected](https://cloud.google.com/bigquery/docs/reference/standard-sql/aggregate_functions#any_value), however running [tests on sample data like this example](https://console.cloud.google.com/bigquery?sq=470293616514:db78929f54a9409cabba540f244d8d62) result in the first NOT NULL value returned, which would cover for this limitation.
So I think this calls for using ANY_VALUE + HAVING. Do you see any issue with using it for all the columns?
e.g.
```
SELECT
   client_id AS client_id,
   sample_id AS sample_id,
   MIN(submission_timestamp) AS submission_timestamp,
   ANY_VALUE(application.architecture HAVING MIN submission_timestamp) AS architecture,
   ANY_VALUE(application.build_id HAVING MIN submission_timestamp) AS app_build_id,
   ANY_VALUE(normalized_app_name HAVING MIN submission_timestamp) AS app_name,
   ANY_VALUE(application.platform_version HAVING MIN submission_timestamp) AS platform_version, 
``` and so on,

File: sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql [hunk 7]
```diff
@@ -0,0 +1,1125 @@
+-- Query for telemetry_derived.clients_first_seen_v1
+WITH new_profile_ping AS (
+  SELECT
+    client_id AS client_id,
+    MIN(sample_id) AS sample_id,
```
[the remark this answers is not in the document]
lucia-vargas-a: Meaning that sample_ids do need to match between pings, right?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
lucia-vargas-a: For this case, the new_profile and main ping have the most data. However if the query gets the available columns from the main ping and the rest from the new_profile ping e.g. experiments, many fields will result in NULL unless we update with later arriving pings. This happens e.g. when retrieving app_name only from the main ping, which then is NULL when the client only reports other pings:
image
I backfilled some data with this approach to `moz-fx-data-shared-prod.analysis.clients_first_seen_v2_day2`

pd/ Documenting also in the [requirements doc](https://docs.google.com/document/d/1B-zTNwschQgljij19Ez104qRV5RUC9zwjSN9tc1C8vc/edit) > "Implementation notes".

lucia-vargas-a: This is what the query was looking like by using the new_profile ping for the main attributes https://console.cloud.google.com/bigquery?sq=470293616514:54f557893eb04b09a3b73d6377385bb7
For what I see, the benefit with mixing pings data is to have a more complete dataset, independent of the source ping..

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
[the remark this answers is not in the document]
fbertsch: Yup, they should always match. They're a pipeline-generated artifact: https://github.com/mozilla/gcp-ingestion/blob/main/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java#L383

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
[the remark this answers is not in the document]
fbertsch: I think so. Just add a note about the docs being incorrect/confusing about the NULL case.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
fbertsch: I see, that makes sense. Carry on!

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -0,0 +1,843 @@
...
+  GROUP BY
+    client_id
+),
+_previous AS (
+  SELECT
+    *
+  FROM
+    `moz-fx-data-shared-prod.telemetry_derived.clients_first_seen_v2`
+  WHERE
+    first_seen_date < CURRENT_DATE()
```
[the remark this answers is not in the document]
fbertsch: On second thought, I would just drop the `WHERE` clause entirely. Rerunning historical days is probably not something we want to enable for this table.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
gkabbz: I will say that for 0.5% or even up to 1% improvement, if it changes our overall numbers and we have to wait 1 week to be final I don't think that juice is worth the squeeze. I'd prefer if we just looked as of the day we were running this and not worry about the window. When it looks like numbers are shifting around for users of the data it creates a lot more headache in my experience trying to build trust around them. But this isn't a call for me to make. What do you think @SuYoungHong  and @irrationalagent?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/schema.yaml
gkabbz, line 1: Missing descriptions for the fields. Having descriptions here would be helpful in improving our overall documentation. Data catalog will pick this up from here.

irrationalagent:
We had a discussion about this in the DS workweek yesterday. The feeling is that where a field gets a significant amount of late-arriving data (say, greater than 5% of a field receiving non-null values from pings AFTER the first ping) then we should allow a 7 day window for those **specific** fields to be updated, along with a metadata field that records the source ping for each of those fields. Outside of these fields that are prone to late arriving data, we should just keep the value from the source ping.

@lucia-vargas-a do you have a sense of which fields have > 1% of clients getting their values from pings other than the first ping? In other words, given 7 days to update, which fields get "late arriving" data for > 1% of clients? I can also help determine this when i get back. 

overall the idea is kind of a compromise - allow late arriving data for just the fields we know can be an issue (and record which ping we take the data from). otherwise, let's do the easier thing for the rest of the fields and just take data from the first ping only.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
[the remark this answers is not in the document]
lucia-vargas-a: The query had a filter initially on CURRENT_DATE and the idea was to retrieve the full dataset but exclude future dates that would potentially be invalid. That would be equivalent to removing the filter except we'd have any future date.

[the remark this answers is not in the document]
fbertsch: Curious how we might have a date future than `CURRENT_DATE` in this query. If `first_seen_date` is based on `submission_*` fields, those are all server-side calculated and are guaranteed to not have that kind of serious drift.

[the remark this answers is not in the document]
lucia-vargas-a: Indeed, if all are server side then there wouldn't be any.

lucia-vargas-a:
> @lucia-vargas-a do you have a sense of which fields have > 1% of clients getting their values from pings other than the first ping? In other words, given 7 days to update, which fields get "late arriving" data for > 1% of clients? I can also help determine this when i get back.
> 

Sure. The results of the analysis of[ how columns get updated over a 7-day window](https://drive.google.com/file/d/1rbJl8jIlrVrrcVKGuIjwD4RgR1G58l-p/view?usp=drive_link), where day 1: Jun 29, 2023:
--We could run it for some other days but this gives us an idea. In summary:
- 32 columns in total.
- Attribution columns (5) and download token (2) where 0.004 % received data on day 2 and 0,0001 received on later days.
- 13 columns updated on day 2 and following dates under 0,003%.
- 2 columns updated updated under 0,2%
- 7 columns didn’t have nulls on day 1.

lucia-vargas-a:
> > @lucia-vargas-a do you have a sense of which fields have > 1% of clients getting their values from pings other than the first ping? In other words, given 7 days to update, which fields get "late arriving" data for > 1% of clients? I can also help determine this when i get back.
> 
Additionally @irrationalagent, [the results show that some columns](https://drive.google.com/file/d/1rbJl8jIlrVrrcVKGuIjwD4RgR1G58l-p/view?usp=drive_link) have >80-90% nulls on day1 and less than 0,1% update rate within the following 5 days, so we could think nof removing them:
platform_version, xpcom_abi, document_id,  distribution_id,  the partner document columns and apple_model_id.

lucia-vargas-a:
@irrationalagent overall updating on a 7-day windows for most fields results in <1% more complete data. The query is more complex, however it is ready to do this update. Your call.

lucia-vargas-a:
@irrationalagent @gkabbz  running some QA analysis:

1. There is a difference in the `**analysis**.clients_first_seen_v2` (1 month data) vs. `telemetry.clients_first_seen`.

- [~90% of it is due to different first_seen_dates](https://sql.telemetry.mozilla.org/queries/93118/source) because the data in analysis.clients_first_seen_v2  is backfilled from '2023-06-10' but many of these clients sent a main ping earlier and therefore have different first_seen dates in `client_first_seen_v1`.
- The rest of the difference can be explained by differences in other columns as data from other pings is integrated.

2. What is [the contribution of each ping](https://sql.telemetry.mozilla.org/queries/93119/source) in the new profiles number?
Looking at the data backfilled in analysis from `2023-06-10` on average:
- 91% from main ping
- 8% from new_profile ping
- 1x% from first_shutdown ping

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
lucia-vargas-a: Removing the 7-day window based on the numbers (<1% data completeness improvement) and slack discussion with George and Leif.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -0,0 +1,643 @@
...
+    ANY_VALUE(
+      payload.processes.parent.scalars.startup_profile_selection_reason
+      HAVING
+        MIN submission_timestamp
+    ) AS startup_profile_selection_reason,
+    ANY_VALUE(
+      environment.settings.attribution.dltoken
+      HAVING
+        MIN submission_timestamp
+    ) AS download_token,
```
SuYoungHong, line 103: can we rename this to `attribution_dltoken` to fit the pattern with `attribution_source, etc.`?

They're all coming from the same place in the ping (attribution), so keeping their naming convention makes sense.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -0,0 +1,643 @@
...
+    ANY_VALUE(
+      environment.settings.attribution.dlsource
```
SuYoungHong, line 105: same as for download token, can we update this to `attribution_dlsource`?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -0,0 +1,643 @@
...
+    client_id AS client_id,
+    sample_id AS sample_id,
+    IF(
+      ANY_VALUE(submission_date HAVING MIN submission_date) >= '2018-10-30',
+      MIN(submission_timestamp_min),
+      TIMESTAMP(ANY_VALUE(submission_date HAVING MIN submission_date))
+    ) AS first_seen_date,
+    TIMESTAMP(
+      ARRAY_AGG(submission_date IGNORE NULLS ORDER BY submission_date)[SAFE_OFFSET(1)]
+    ) AS second_seen_date,
```
SuYoungHong, line 253: I'm probably missing something, but why are we using the second main ping sent day (`safe_offset(1)`) for the second seen date in this part of the query, when we're using nulls for the new profile and first shutdown ping parts? 

is this assuming that the second main ping day is always the second seen date? 

maybe I'm misunderstanding, but is second seen date supposed to be the second day we get main ping from user, by definition? Or is it supposed to be second day we get any ping from user? (definition says second day we get any which ping)

SuYoungHong (commented):
Hey, took a look and looks good for the most part. 

Have a question, though, about second seen date, and a request for naming on dl token and dl source. 

A more meta follow up question, though, is: 
* I'm understanding this PR's query to be looking over all the historical data at once (which we'll need to do to initialize and backfill), but what's going to be the daily ETL logic? 

For every day going forward, will we look at all of a new day's NPP, FSP, MP pings and then filter out everyone we've seen before by comparing to the `clients_first_seen_v2` table? Or will we be running over all the historical data raw each day?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
lucia-vargas-a: Good question. This is the result of a discussion with Ruoxi @rhythmofrain63 where we decided to exclude new_profile and first_shutdown pings to retrieve the second seen dates because the new_profile ping is being sent more than once for around ~16% of clients (the expectation is once per client_id) and this could result in both 1st and 2nd seen dates being reported by the new_profile ping.
Additionally, we think that it doesn't make sense if the new_profile ping reports the second_seen date after the main ping already reported the first (the browser is already installed and opened), what do you think?

fbertsch (commented):
I'm just a bit concerned about the backfill with the main ping

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -0,0 +1,643 @@
...
+    CAST(NULL AS STRING) AS attribution_dlsource
+  FROM
+    `moz-fx-data-shared-prod.telemetry_derived.clients_daily_v6`
+  WHERE
+    {% if is_init() %}
+      submission_date >= '2010-01-01'
+      {% if parallel_run() %}
+        AND sample_id = {sample_id}
+      {% endif %}
+    {% else %}
```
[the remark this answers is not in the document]
lucia-vargas-a: Running sample_ids (0,1) in parallel, each sample_id [runs in avg. 19 min and processes 900 GB](https://console.cloud.google.com/bigquery?sq=470293616514:4d738027e7a546488a24c89650f41f8b&project=mozdata) so I think it would be okay, I guess we need to run the 100 sample_ids for this validation.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/schema.yaml
```diff
@@ -0,0 +1,215 @@
...
+  type: STRING
+  mode: NULLABLE
+- description: 'How the profile was selected during startup. Possible reasons are:
+   unknown: Generally should not happen, set as a default in case no other reason occurred. profile-manager: The profile was selected by the profile manager.
+   profile-reset: The profile was selected for reset, normally this would mean a restart. restart: The user restarted the application,
+   the same profile as previous will be used. argument-profile: The profile was selected by the --profile command line argument.
+   argument-p: The profile was selected by the -p command line argument. firstrun-claimed-default: A first run of a dedicated profiles build chose the old
+   default profile to be the default for this install. firstrun-skipped-default: A first run of a dedicated profiles build skipped over the old default profile
+   and created a new profile. restart-claimed-default: A first run of a dedicated profiles build after a restart chose the old default.'
+  name: startup_profile_selection_reason
```
ANich, line 124: cc @SuYoungHong @irrationalagent @Marlene-M-Hirose @gkabbz revisit this, seems like this is "missing" from main ping clients

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -0,0 +1,682 @@
...
+    ) AS engine_data_submission_url,
+    CAST(NULL AS STRING) AS apple_model_id,
+    ANY_VALUE(city HAVING MIN submission_date) AS city,
+    CAST(NULL AS STRING) AS db_version,
+    ANY_VALUE(geo_subdivision1 HAVING MIN submission_date) AS subdivision1,
+    ANY_VALUE(normalized_channel HAVING MIN submission_date) AS normalized_channel,
+    ANY_VALUE(country HAVING MIN submission_date) AS country,
+    ANY_VALUE(os HAVING MIN submission_date) AS normalized_os,
+    ANY_VALUE(normalized_os_version HAVING MIN submission_date) AS normalized_os_version,
+    CAST(NULL AS STRING) AS startup_profile_selection_reason,
```
irrationalagent, line 310: cc @gkabbz 

This is the part that we potentially need backfilled into clients_daily_v6, [pulling from this probe](https://probes.telemetry.mozilla.org/?search=reason&view=detail&probeId=scalar%2Fstartup.profile_selection_reason) in the main ping. 

This is needed so long as we want to maintain the current logic of the NtM dashboard. The dashboard only counts new profiles with reason = [`firstrun-created-default`](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/telemetry_derived/desktop_funnel_new_profiles_v1/query.sql#L17). This is to help us ID new profiles that were created from an install, rather than ones that were created for some other reason. 

Without this addition, the dashboard fka NtM will ignore new profiles in the case the first ping we receive from them is a main ping.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
irrationalagent: My 2c, I think it makes sense to exclude the new profile ping from contributing the second_seen_date. Assuming the intent here is to capture their second day of use, receiving the NPP after their first main ping would not be a fair reflection of their second day (bcs its likely we SHOULD have actually received the NPP first, and we are just getting it second for uninteresting reasons). 

Technically, since the first_shutdown_ping might often be sent on the next UTC day, we might consider that their "second seen date", similarly to how we would expect a second main ping from them on their next UTC day, if they leave the browser open. So that would be an argument for allowing the FSP to define the second_seen_date. That being said, I'm not sure if that _really_ buys us all that much more in exchange for the added complexity.

On balance, I think that we should probably stick with the idea that your second_seen date is the second UTC day we receive a main ping from each client, if such a second day occurs.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -0,0 +1,682 @@
...
+  SELECT
+    *,
+    'main' AS ping_source
+  FROM
+    main_ping
+),
+-- The following 3 subqueries are required to calculate the second seen date.
+-- First, return one row per client_id with an array of first_seen TIMESTAMPS and
+-- an array of second_seen TIMESTAMPS (one date per reported ping).
+unioned_with_all_dates AS (
```
irrationalagent: If we decide that only main pings can contribute the second_seen_date, is this unioning still necessary? Could we just aggregate the main pings alone, and take the second date (if it exists) from there?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
SuYoungHong: that's a good point. The aggregation might be problematic. Taking the first of the day would be useful for **our** purposes, but it might not be true for any other/future use cases that need this field. Adding it to `clients_daily` might just not make sense. 

Other options to consider. Is pulling the data directly from main pings, and bypassing `clients_daily` an option, or is it way too cost prohibitive? I know the history of main pings doesn't go far back enough (we might only keep main pings 18 months or something), but can we initialize with some combination of both (use main pings where available, otherwise use `clients_daily`), and then just look at main pings for the daily updates?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/schema.yaml
SuYoungHong: Good call out. Just to clarify, the data does exist in main pings (`telemetry.main`), however, it's not propagated to `telemetry.clients_daily`, which is where this ETL is pulling from.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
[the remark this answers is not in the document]
SuYoungHong: @fbertsch , is that the size for the main ping table (`telemetry.main`), or client daily (`telemetry.clients_daily`)?Currently, this code is pulling the data from clients daily, not main pings.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
SuYoungHong: also, to add, `attribution.ua` is also a field necessary for Numbers that Matter, that's not available in `telemetry.clients_daily` (but is available in the raw ping). 

[line 289](https://github.com/mozilla/bigquery-etl/blob/dabd70ad9f3bd2bcbdfccbab29c738b529d99aa4/sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql#L289C5-L289C44)

SuYoungHong (commented):
Looks like locale and attribution.ua were added (+1), and the build_id source was switched (+1). 

Yesterday we decided to switch the logic from reporting the _first non null value reported from any ping_ to just reporting the values from the first ping received. However, I'm seeing this (as an example): 

```ANY_VALUE(application.architecture HAVING MIN submission_timestamp) AS architecture```

I may be wrong, but I believe `ANY_VALUE` will return the first _non-null_ value, so this query would be reporting the first non null value from the ping type we first receive (as opposed to returning the first value, regardless of nulls). 

So I think we would have to update these clauses to include a RESPECT NULLS clause. 

**However,** we may want to delay implementing any changes until we resolve the issue with missing data from `clients_daily` that @irrationalagent [raised above](https://github.com/mozilla/bigquery-etl/pull/3962#discussion_r1332186267). 

If we want to report data from the first ping we receive, and we want to get the missing data, we'd need to either source the data from main pings directly instead of clients daily, or update clients daily to include the missing data, or possibly change our decision on reporting whatever is in the first ping to an alternative (like reporting first non null from any ping we recieve with in a certain window... although... that still might not fix the issue). 

But in either case, I think we need to figure out how to address this issue first.

File: sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql (hunk 14 above)
Comment (lucia-vargas-a) [reply sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql:310]: Yes, there are 14 columns in the query that are not available in `clients_daily_v6` but are available in the ping table `main_v4` (added a comment in the query with each corresponding column).
In the query, these are retrieved from `new_profile` and `first_shutdown` ping and would be NULL if the main ping reports the first_seen_date.
Is there any other source that could be used for the dashboard while these data is available in `clients_daily_v6`?

Comment (lucia-vargas-a) [reply sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql]: Alternatively, we could retrieve startup_profile_reason and attribution.ua from another ping if it's available.

File: sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql (hunk 15 above)
Comment (lucia-vargas-a) [reply sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql]: True, this union is not required, I changed it. Just a note, we still need to look at all the pings because not always the 2nd main ping is the second_seen_date of the client. The 1st main ping reported could also be a second_seen_date if another ping was reported on a date before.

**Comment by lucia-vargas-a** [comment]
> Yesterday we decided to switch the logic from reporting the _first non null value reported from any ping_ to just reporting the values from the first ping received. However, I'm seeing this (as an example):
> 
> `ANY_VALUE(application.architecture HAVING MIN submission_timestamp) AS architecture`
> 
> I may be wrong, but I believe `ANY_VALUE` will return the first _non-null_ value, so this query would be reporting the first non null value from the ping type we first receive (as opposed to returning the first value, regardless of nulls).
> 
> So I think we would have to update these clauses to include a RESPECT NULLS clause.
> 

@SuYoungHong you're right on how it currently works.  Indeed we can be strict on retrieving the attributes from the ping that reports the earliest timestamp, even if this value is NULL and even if more pings of the same type are received on the same first_seen_date, could you confirm this?

**Review by SuYoungHong — APPROVED** [review]
Hi @lucia-vargas-a ! 

In regards to: 
> @SuYoungHong you're right on how it currently works. Indeed we can be strict on retrieving the attributes from the ping that reports the earliest timestamp, even if this value is NULL and even if more pings of the same type are received on the same first_seen_date, could you confirm this?

Yes, exactly. Let's be strict on retrieving only the data from the first reported ping (even if it's null), so the behavior of the ETL reflects "pull data from first ping" strictly. 

R.E. the missing fields in `clients_daily` issue:

In light of @gkabbz decision in our slack convo (let's finalize this PR with the missing fields from `clients_daily` as is for now, and we can backfill the missing data from `clients_daily` and in the future), I'm going to approve this PR. 

Please update the `ANY_VALUE` logic to include null values from the first ping, but other then that, I think this PR is good to go.

File: sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql [hunk 16]
```diff
@@ -0,0 +1,643 @@
...
+    unioned_with_all_dates
+  LEFT JOIN
+    UNNEST(seen_dates) AS seen_dates
+  WHERE
+    seen_dates.value IS NOT NULL
+  GROUP BY
+    client_id,
+    value
+),
+unioned_second_dates AS (
```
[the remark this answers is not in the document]
lucia-vargas-a: Good idea, I added more documentation to the query. Also this logic was updated.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
lucia-vargas-a: Resolving as per [decision](https://github.com/mozilla/bigquery-etl/pull/3962#pullrequestreview-1639351082).

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -0,0 +1,643 @@
+-- Query for telemetry_derived.clients_first_seen_v2
+{% if parallel_run() %}
```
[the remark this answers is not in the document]
lucia-vargas-a: Agree, the destination table needs to be explicit in the query. I'm currently using `INSERT INTO <destination table>`. However, this part of the code is moved [PR-4322](https://github.com/mozilla/bigquery-etl/pull/4322) so we can evaluate the best approach.

PR #3963, opened 2023-06-19
fix(): updated firefox_ios_derived models scheduling to use firefox_ios DAG

kik-kik:
fix(): updated firefox_ios_derived models scheduling to use firefox_ios DAG


fbertsch (approved):
LGTM, thank you!

PR #3966, opened 2023-06-19
DENG-984:  Updated workgroup access for backfills_staging_derived dataset

wwyc:
* The `backfills_staging_derived` dataset was added in #3904 
* This PR updates the workgroup access to `mozilla-confidential` to ensure appropriate access to this dataset.


sql/moz-fx-data-shared-prod/backfills_staging_derived/dataset_metadata.yaml
```diff
@@ -4,4 +4,7 @@ description: |-
 dataset_base_acl: derived
 user_facing: false
 labels: {}
-workgroup_access: []
+workgroup_access:
```
ANich, line 7: For my understanding - the previous, empty workgroup access array doesn't apply the now-added default workgroup?

whd: Correct. Dataset access works like this:

Take the [dataset_base_acl](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/backfills_staging_derived/dataset_metadata.yaml#L4) as the base access for the dataset. `gsutil cat gs://moz-fx-data-terraform-state-global/projects/data-shared/global/access-groups/default.tfstate | jq .outputs.dataset_base_acls.value.derived` is one simple (using non-public interfaces) way to see the value of this base ACL.

Then, add any bigquery ACLs from [workgroup_access](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/backfills_staging_derived/dataset_metadata.yaml#L7) to the values in `dataset_base_acl`.  When that value is `[]`, no additional ACLs are applied. When the value is granting `roles/bigquery.dataViewer` to e.g. `workgroup:mozilla-confidential`, take the entries from `gsutil cat gs://moz-fx-data-terraform-state-global/projects/data-shared/global/access-groups/default.tfstate | jq '.outputs.workgroups.value | .["mozilla-confidential"].bigquery_acls.default'`, and grant `READER` (`roles/bigquery.dataViewer`) to them.

whd (approved):
I'd like to be responsible for merging this as to do so in tandem with https://github.com/mozilla-services/cloudops-infra/pull/4945

PR #3970, opened 2023-06-21
changing the formatting for documentation in cookbooks

lelilia:
image

currently the list is not interpreted as such and it all just ends up in one long text blob 
numbering seems to work for other lists so changing it to numbers


docs/cookbooks/common_workflows.md
```diff
@@ -125,9 +125,9 @@ The following is an example to update a new field in `telemetry_derived.clients_
 ## Remove a field from a table schema
 
 Deleting a field from an existing table schema should be done only when is totally neccessary. If you decide to delete it:
-* [x] Validate if there is data in the column and make sure data it is either backed up or it can be reprocessed.
-* Follow [Big Query docs](https://cloud.google.com/bigquery/docs/managing-table-schemas#deleting_columns_from_a_tables_schema_definition) recommendations for deleting.
-* If the column size exceeds the allowed limit, consider setting the field as NULL. See this [search_clients_daily_v8](https://github.com/mozilla/bigquery-etl/pull/2463) PR for an example.
+1. Validate if there is data in the column and make sure data it is either backed up or it can be reprocessed.
```
lucia-vargas-a, line 128: Uh, thanks for this fix.
We normally use:
1.
1.
1.
so that there is no need to keep the number sequence manually.

lucia-vargas-a (approved):

PR #3978, opened 2023-06-22
Deng-1052 Added bigquery_usage_v2 table

wwyc:
[DENG-1052](https://mozilla-hub.atlassian.net/browse/DENG-1052):
Given access to jobs_by_organization table was granted to airflow in [DSRE-1106](https://mozilla-hub.atlassian.net/browse/DSRE-1106).  This PR adds `bigquery_usage_v2` table to test access.

### DIFF
--- a/sql/moz-fx-data-shared-prod/monitoring_derived/bigquery_usage_v2/query.py
+++ b/sql/moz-fx-data-shared-prod/monitoring_derived/bigquery_usage_v2/query.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python3
+
+"""
+Determine BigQuery usage.
+
+To read more on the source table, please visit:
+https://cloud.google.com/bigquery/docs/information-schema-jobs-by-organization
+"""
+
+from argparse import ArgumentParser
+
+from google.cloud import bigquery
+
+DEFAULT_PROJECTS = [
+    "moz-fx-data-shared-prod",
+]
+
+parser = ArgumentParser(description=__doc__)
+parser.add_argument("--date", required=True)  # expect string with format yyyy-mm-dd
+parser.add_argument("--project", default="moz-fx-data-shared-prod")
+# projects queries were run from that access table
+parser.add_argument("--source_projects", nargs="+", default=DEFAULT_PROJECTS)
+parser.add_argument("--destination_dataset", default="monitoring_derived")
+parser.add_argument("--destination_table", default="bigquery_usage_v2")
+
+# total_bytes_billed a minimum of 10MB per table total_bytes_processed
+
+
+def create_query(date, source_project):
+    """Create query for a source project."""
+    return f"""
+        SELECT
+         "{source_project}" AS source_project,
+          DATE('{date}') AS creation_date,
+          job_id,
+          job_type,
+          reservation_id,
+          cache_hit,
+          state,
+          referenced_tables.project_id AS reference_project_id,
+          dataset_id AS reference_dataset_id,
+          table_id AS reference_table_id,
+          destination_table.project_id AS  destination_project_id,
+          destination_table.dataset_id AS  destination_dataset_id,
+          destination_table.table_id AS  destination_table_id,
+          user_email,
+          REGEXP_EXTRACT(query, r"Username: (.*?),") AS username,
+          REGEXP_EXTRACT(query, r"Query ID: (\\w+), ") AS query_id,
+          end_time-start_time as task_duration,
+          ROUND(total_bytes_processed / 1024 / 1024 / 1024 / 1024, 4)
+          AS total_terabytes_processed,
+          error_result.location AS error_location,
+          error_result.reason AS error_reason,
+          error_result.message AS error_message,
+        FROM
+          `{source_project}.region-us.INFORMATION_SCHEMA.JOBS_BY_ORGANIZATION`
+        LEFT JOIN
+          UNNEST(referenced_tables) AS referenced_tables
+        WHERE
+          DATE(creation_time) = '{date}'
+    """
+
+
+def main():
+    """Run query for each source project."""
+    args = parser.parse_args()
+
+    partition = args.date.replace("-", "")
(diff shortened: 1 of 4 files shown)


sql/moz-fx-data-shared-prod/monitoring_derived/bigquery_usage_v2/query.py
```diff
@@ -0,0 +1,85 @@
...
+          cache_hit,
+          state,
+          referenced_tables.project_id AS reference_project_id,
+          dataset_id AS reference_dataset_id,
+          table_id AS reference_table_id,
+          destination_table.project_id AS  destination_project_id,
+          destination_table.dataset_id AS  destination_dataset_id,
+          destination_table.table_id AS  destination_table_id,
+          user_email,
+          REGEXP_EXTRACT(query, r"Username: (.*?),") AS username,
```
whd, line 47: There's a bit of an unfortunate complication here, which is that [JOBS_BY_ORGANIZATION](https://cloud.google.com/bigquery/docs/information-schema-jobs-by-organization) doesn't actually have the `query` column. Even though it's listed, its description says `SQL query text. Only the JOBS_BY_PROJECT view has the query column.`.

It's possible to use `JOBS_BY_ORGANIZATION` to get job ids and projects, then get associated query information from e.g. ``[PROJECT].`region-us`.INFORMATION_SCHEMA.JOBS_BY_PROJECT``, but that's going to be more complicated. Airflow has access to this metadata for all projects the same way it has access to `JOBS_BY_ORGANIZATION`.

Aside: I'm guessing the parsing of `query` here is using either Looker or STMO comment metadata. This definitely isn't going to be valid for all queries, but that's probably known for `bigquery_usage_v1`.

whd (commented):
From the decision brief you should add a `CODEOWNERS` entry for `@mozilla/dataops` on the `sql/moz-fx-data-shared-prod/monitoring_derived/bigquery_usage_v2` path.

As it stands though it looks like this isn't going to work, see inline comment :disappointed:

sql/moz-fx-data-shared-prod/monitoring_derived/bigquery_usage_v2/query.py
wwyc: Thanks for the detailed feedback!  I have added the appropriate codeowners entry.  I will leave out the parsing of the`query` field for now.

sql/moz-fx-data-shared-prod/monitoring_derived/bigquery_usage_v2/query.py
```diff
@@ -0,0 +1,84 @@
...
+          end_time-start_time as task_duration,
+          ROUND(total_bytes_processed / 1024 / 1024 / 1024 / 1024, 4)
+          AS total_terabytes_processed,
+          total_slot_ms,
+          error_result.location AS error_location,
+          error_result.reason AS error_reason,
+          error_result.message AS error_message,
+          query_info.resource_warning AS resource_warning,
+        FROM
+          `{source_project}.region-us.INFORMATION_SCHEMA.JOBS_BY_ORGANIZATION`
```
whd, line 55: The main issue here is that `source_project` in this context only specifies where this specific query is to be run. It doesn't do any kind of filtering on jobs to a specific project. So the output of this query will return metadata for jobs that have no interaction with the data platform, when our intent is to limit the metadata to only jobs interacting with the platform.

In order to limit metadata to jobs reading from or writing to destinations listed in `DEFAULT_PROJECTS`, I think you need to perform some kind of filtering in the query itself. So instead of iterating over `DEFAULT_PROJECTS` one at a time to get `JOBS_BY_PROJECT` for each, you should query `JOBS_BY_ORGANIZATION` once where `reference_project_id` and/or `destination_project_id`  are in `DEFAULT_PROJECTS`. `source_project` in this case, instead of being static, would be what is currently the value of `project_id`: the project the job ran in.

whd (commented):
The overall goal here is to see BQ Jobs run from anywhere in the org that are interacting with the data platform. For instance, metadata about jobs run by the `hubs`: https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/hubs_derived/dataset_metadata.yaml#L11 is not accessible via `bigquery_usage_v1` since those jobs originate in the Hubs project `hubs-dev-333333`.

To continue with hubs as an example: I think we would want the table to include (example with subset of relevant columns):

```
source_project,reference_project_id,destination_project_id
hubs-dev-333333,moz-fx-data-shared-prod,hubs-dev-333333
```

i.e. `source_project` is the project where hubs ran the BQ job from, `reference_project_id` is the data platform project whose datasets were referenced, and `destination_project_id` is the project where output results were written. In hubs' case, we would expect usage to look like the above, since they are querying data platform data from their own project and then writing their results to BQ in their own project. We wouldn't expect to collect a log like:
```
hubs-dev-333333,hubs-dev-333333,hubs-dev-333333
```

which would plausibly be the result of hubs then querying the results of the export of our data, but not directly interacting with the platform.

Given the above I don't think there's a need for both `project_id` and `source_project` since they seem redundant to me, but I could be missing some nuance.

sql/moz-fx-data-shared-prod/monitoring_derived/bigquery_usage_v2/query.py
wwyc: Thanks for sharing the details.  They are very helpful.  I have added the filters for `DEFAULT PROJECTS` in the query.  I also needed the jobs that run under the defaults projects.  I have been getting this data from the `JOBS_BY_PROJECT` view in the `bigquery_usage_v1` table.  I hope that works.

whd (approved):
LGTM

A good example of new data to look for in this table when this lands:

`WHERE source_project = 'moz-fx-amo-prod-9572'`

allowing us to analyze AMO's [programmatic usage](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/amo_prod/dataset_metadata.yaml#L11) of the Data Platform.

PR #3968, opened 2023-06-20
feat(DENG-353): adding missing fxa derived table schemas

kik-kik:
# feat(DENG-353): adding missing fxa derived table schemas

This PR adds schemas for:

- fxa_users_services_last_seen_v1
- fxa_oauth_events_v1
- fxa_stdout_events_v1
- fxa_content_events_v1
- fxa_auth_events_v1
- funnel_events_source_v1
- events_daily_v1
- event_types_v1

We should investigate as a follow-up if schemas for `events_daily_v1` and `event_types_v1` could be added inside the schema generator for `events_daily`.

PR #3979, opened 2023-06-22
Fix FxA ETL schemas

sean-rose:
In #3968 a `submission_date` column was erroneously added to five FxA ETLs' `schema.yaml` files, which were subsequently deployed to add the column to the associated tables (though it won't be populated by the ETLs).  This corrects the `schema.yaml` files that haven't already been corrected, and those `submission_date` columns would also ideally be dropped from these tables to avoid ongoing confusion:
- `moz-fx-data-shared-prod.firefox_accounts_derived.event_types_v1`
- `moz-fx-data-shared-prod.firefox_accounts_derived.fxa_auth_bounce_events_v1`
- `moz-fx-data-shared-prod.firefox_accounts_derived.fxa_auth_events_v1`
- `moz-fx-data-shared-prod.firefox_accounts_derived.fxa_oauth_events_v1`
- `moz-fx-data-shared-prod.firefox_accounts_derived.fxa_stdout_events_v1`

The `schema.yaml` file for the `nonprod_fxa_server_events_v1` ETL which was used to deploy the initial table also has some issues in terms of not matching the ETL query's results:
- It's missing the `logName` column.
- It contains `jsonPayload.fields.countrycode` and `jsonPayload.fields.country_code` fields which aren't present in the query's result, and could confuse people (like me ;) by implying that those are actual fields.
- The `jsonPayload.fields.t` field was deployed as a `STRING` which doesn't match the query result field's `FLOAT` type.

Since the `moz-fx-data-shared-prod.firefox_accounts_derived.nonprod_fxa_server_events_v1` doesn't contain any data yet I'd recommend deleting the table and redeploying it from the corrected schema once this PR is merged.

---
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1057)


sean-rose:
> those `submission_date` columns would also ideally be dropped from these tables to avoid ongoing confusion

I see two options for doing this:
1. Use [`ALTER TABLE ... DROP COLUMN ...` statements](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#alter_table_drop_column_statement).
   - Pros: doesn't cost anything, probably fast.
   - Cons: tables can no longer be queried with legacy SQL.
2. Overwrite the tables with `SELECT * EXCEPT (submission_date) ...`.
   - Pros: tables can continue to be queried with legacy SQL.
   - Cons: queries will cost several hundred dollars (`fxa_auth_events_v1` is 74 TB).

I've asked in [#data-platform-infra-wg](https://mozilla.slack.com/archives/C01E8GDG80N/p1687475911555699) about whether legacy SQL support is a concern, but assuming it isn't then I'd probably lean toward the cheaper/easier `ALTER TABLE ... DROP COLUMN ...` approach.

sean-rose:
> I've asked in [#data-platform-infra-wg](https://mozilla.slack.com/archives/C01E8GDG80N/p1687475911555699) about whether legacy SQL support is a concern, but assuming it isn't then I'd probably lean toward the cheaper/easier `ALTER TABLE ... DROP COLUMN ...` approach.

@relud pointed out that [Shredder uses legacy SQL to retrieve table partitions](https://github.com/mozilla/bigquery-etl/blob/a4fdf3c65eb5ffdd6ed9ec62cb5f0e6124f58a76/bigquery_etl/shredder/delete.py#L348-L358), and [these tables do get shredded](https://github.com/mozilla/bigquery-etl/blob/1bade1743672e34e990168e42e2087c6713ba6f1/bigquery_etl/shredder/config.py#L267-L272), so that's a blocker to using `ALTER TABLE ... DROP COLUMN ...` on them.

relud:
> > I've asked in [#data-platform-infra-wg](https://mozilla.slack.com/archives/C01E8GDG80N/p1687475911555699) about whether legacy SQL support is a concern, but assuming it isn't then I'd probably lean toward the cheaper/easier `ALTER TABLE ... DROP COLUMN ...` approach.
> 
> 
> 
> @relud [pointed out](https://mozilla.slack.com/archives/C01E8GDG80N/p1687543764130809?thread_ts=1687475911.555699&cid=C01E8GDG80N) that [Shredder uses legacy SQL to retrieve table partitions](https://github.com/mozilla/bigquery-etl/blob/a4fdf3c65eb5ffdd6ed9ec62cb5f0e6124f58a76/bigquery_etl/shredder/delete.py#L348-L358), and [these tables do get shredded](https://github.com/mozilla/bigquery-etl/blob/1bade1743672e34e990168e42e2087c6713ba6f1/bigquery_etl/shredder/config.py#L267-L272), so that's a blocker to using `ALTER TABLE ... DROP COLUMN ...` on them.

to be clear, shredder could use the new partition metadata available in standard sql instead, it just hasn't been updated yet.

kik-kik:
Hey @sean-rose , thanks for following up on this and sorry for causing the error.

Prior to seeing this open PR I already fixed the `nonprod_fxa_server_events_v1` schema in this PR:
https://github.com/mozilla/bigquery-etl/pull/4016

I'm not quite sure what happened there. I redeployed the table and the corresponding ETL is now completing successfully.

Perhaps as a temporary measure to avoid recreating all the tables we can add a description to the field specifying that it was added by a mistake and remove it once Shredder no longer relies on legacy_sql using the alter statement? I don't think this should be a major issue since those tables should not be accessed directly but instead via the `fxa_all_events` view where we can make sure `submission_date` is excluded so that end users don't accidentally use this field? What do you think?

---

P.S. I had to Google what `erroneously` means 😓

sean-rose:
> Prior to seeing this open PR I already fixed the `nonprod_fxa_server_events_v1` schema in this PR: #4016

That fixed the `logName` and `jsonPayload.fields.t` issues, but it still contains `jsonPayload.fields.countrycode` and `jsonPayload.fields.country_code` fields which aren't present in the query's result, and could confuse people by implying that those are actual fields.

> Perhaps as a temporary measure to avoid recreating all the tables we can add a description to the field specifying that it was added by a mistake and remove it once Shredder no longer relies on legacy_sql using the alter statement? I don't think this should be a major issue since those tables should not be accessed directly but instead via the `fxa_all_events` view where we can make sure `submission_date` is excluded so that end users don't accidentally use this field? What do you think?

That sounds reasonable to me.

kik-kik (approved):
Sorry for the delay in the response @sean-rose, I was out. Thanks for doing the changes and once again sorry for causing the issue to begin with.

PR #3992, opened 2023-06-27
Round VPN exchange rates to fit in the BigQuery decimal type (bug 1840566)

sean-rose:
OANDA's historical exchange rates API moved to a different URL, which was addressed in #3991.

However, that API now apparently has slightly different behavior where it may return exchange rates which have more than the 9 decimal places allowed by [BigQuery's decimal type](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#decimal_types), in which case the `mozilla_vpn_derived.exchange_rates_v1` ETL will fail due to a behind-the-scenes "Invalid NUMERIC value" error ([bug 1840566](https://bugzilla.mozilla.org/show_bug.cgi?id=1840566)).

---
Checklist for reviewer:

### DIFF
--- a/sql/moz-fx-data-shared-prod/mozilla_vpn_derived/exchange_rates_v1/query.py
+++ b/sql/moz-fx-data-shared-prod/mozilla_vpn_derived/exchange_rates_v1/query.py
@@ -73,7 +73,7 @@ for base_currency in args.base_currencies:
                 quote_currency=currency_pair["quoteCurrency"],
                 base_currency=currency_pair["baseCurrency"],
                 price_type=currency_pair["type"],
-                price=price,
+                price=round(Decimal(price), 9),
             )
             for currency_pair in response.json()["widget"]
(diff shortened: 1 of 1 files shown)


sql/moz-fx-data-shared-prod/mozilla_vpn_derived/exchange_rates_v1/query.py
```diff
@@ -73,7 +73,7 @@ class Quote:
                 quote_currency=currency_pair["quoteCurrency"],
                 base_currency=currency_pair["baseCurrency"],
                 price_type=currency_pair["type"],
-                price=price,
+                price=round(Decimal(price), 9),
```
sean-rose, line 76: Or we could change the `price` column type to be `BIGNUMERIC`, but this approach seemed simplest and 9 decimal places seems plenty accurate.

relud (approved):

PR #3993, opened 2023-06-27
[DENG-948] Macro support for data checks

scholtzan:
This adds support for defining data checks as macros and making them reusable.

I also implemented a few checks from https://mozilla-hub.atlassian.net/browse/DENG-945

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1075)


kik-kik (approved):
Maybe one small comment would be around having tests to make sure our checks are rendered as expected based on specific input? Otherwise, LGTM 👍

PR #4015, opened 2023-07-04
fix(bug-1841622): updated nonprod_fxa_server_events_v1 schema

kik-kik:
# fix(bug-1841622): updated nonprod_fxa_server_events_v1 schema

This is to address Airflow failures.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1103)


lelilia (approved):
Are you sure you want it as a STRING and not as a FLOAT? 
If so, this lgtm
Oh also .. what is `t`? this is not a very informative name .. and there is no description for it in the schema

kik-kik:
> Are you sure you want it as a STRING and not as a FLOAT? If so, this lgtm Oh also .. what is `t`? this is not a very informative name .. and there is no description for it in the schema

It looks like some sort of accidental field. It does not look like this month there was a single not null value in it and we do not appear to use it downstream. This is to just unblock our ETL.

PR #4021, opened 2023-07-05
bug(DENG-709): removing incorrect event filtering in the fxa_users_services_devices_last_seen_v1 query

kik-kik:
# bug(DENG-709): removing incorrect event filtering in the fxa_users_services_devices_last_seen_v1 query

Currently this specific last seen table is incorrect due to us only considering login or registration events when generating the last seen table which yields incorrect results. Once merged, we will need to recreate this table and backfill it until current date and then let the ETL take over.


akkomar (commented):
@kik-kik can you outline the plan for tackling DENG-709?

By removing filtering here we're falling back to the [condition in the upstream table](https://github.com/mozilla/bigquery-etl/blob/10680a7e81b82b2c6647e7cd996cdfc8987d8172/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_devices_daily_v1/query.sql#L33-L47) which is not entirely consistent with the [proposal from DENG-709](https://mozilla-hub.atlassian.net/browse/DENG-709?focusedCommentId=697898) which suggests to only use `Fxa_activity - access_token_checked`, `Fxa_activity - access_token_created`, `Fxa_activity - cert_signed`.

kik-kik:
Hey @akkomar ,

So I believe this table should be consistent with the upstream table. As those both tables to me reflect `user activity`. If we need to have more strict filtering then it should be happening on the `daily` level tables. And in this case we would need to updated it for the `user_services_devices_daily` and `user_services_daily` to make sure both of these are aligned. However, before doing that I want to make sure there are no other events we should be including in the daily table. Should the filter be `fxa_activity%` (are we expecting new events to be added in this category?) or specifically just for these three event types.

akkomar:
As discussed on Zoom - let's get back to the drawing board and start from defining which metrics we want to see in the tables.

Issue #3182, opened 2022-08-25
Should all SQL generators skip overwriting existing files?


sean-rose:
Originally only [`stable_views` did](https://github.com/mozilla/bigquery-etl/blob/81fbdf6c015511307f27e14b7d5b08930cdcc7a7/sql_generators/stable_views/__init__.py#L7-L9), and now [`glean_usage` does](https://github.com/mozilla/bigquery-etl/pull/3149).  Should we have all SQL generators do the same?

The main benefit is being able to manually override generated files as needed.

Some potential drawbacks are such manual overrides being overused, and checked-in files accidentally blocking SQL generation (though the status quo has the reciprocal issue of SQL generation accidentally overriding checked-in files).

ANich:
Agreed on the drawbacks. It sounds like an issue differentiating between "I checked in this file to replace what would have been generated" and "I checked this in by accident" or "this file is already checked in but I forgot to update it after updating the generator". We only have a few checked in files now so we could delete the ones we don't want to overwrite and hope they don't get checked in again but that isn't perfect.  

What do you think about specifying explicitly (in some generator config maybe) what file paths should be generated and/or ignored by a particular SQL generator?

fbertsch:
+1 to this idea, that would be my expectation when putting in PRs, and fits in with our general mantra: Prefer explicit over implicit.

Dealing with the drawbacks could be a challenge, but perhaps a good path could be visibility into what the `generated-sql` branch will look like after your change is merged (as a comment on a PR) could be a path.

sean-rose:
Copying over some comments from a [Slack thread](https://mozilla.slack.com/archives/C01E8GDG80N/p1661471288408739):
> @akkomar 
> I don't have an opinion about this apart from:
> - +1 for thinking about standardising this
> - https://github.com/mozilla/bigquery-etl/tree/main/sql_generators#readme should probably be updated with some info on the current approach
> - only somewhat related since you're asking about views and this is a query: https://github.com/mozilla/bigquery-etl/blob/main/sql_generators/search/__init__.py#L66 seems to be working in a different way than other scripts by not writing the query file and relying on the checked-in version (cc @lucia-vargas-a, @ANich)

> @lucia-vargas-a 
[...]

Issue #4068, opened 2023-07-12
Write generated artifacts to a separate directory


scholtzan:
There would be some benefits in writing artifacts generated by `sql_generators` to a separate directory instead of overwriting `sql/`. One benefit it that it would be easier to determine files for which the generated SQL should not be used (aka files that were added to `sql/` explicitly).

This is a more significant change and might have some other side-effects which should be thought through

PR #4069, opened 2023-07-12
[Bug 1821784] Be explicit about skipping generation for certain Glean artifacts

scholtzan:
This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1821784

Instead of skipping existing files during Glean generation processes, we should be explicit in what files should be skipped. This prevents us from breaking artifact deploys.

This is a less significant change than the alternative option of writing generated sql to a separate directory, which would also solve this issue: https://github.com/mozilla/bigquery-etl/issues/4068
This is something still worth considering, but this PR is a quicker fix for an issue that has been happening on a weekly basis now.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1183)

### DIFF
--- a/bqetl_project.yaml
+++ b/bqetl_project.yaml
@@ -363,6 +363,9 @@ view:
     - sql/glam-fenix-dev/glam_etl/**/view.sql
     # tests
     - sql/moz-fx-data-test-project/test/simple_view/view.sql
+  override: # Skip automatically updating the following views
+    skip:
+    - sql/moz-fx-data-shared-prod/mozilla_vpn/**/view.sql
 
 schema:
   mozilla_pipeline_schemas_uri: https://github.com/mozilla-services/mozilla-pipeline-schemas
--- a/bqetl_project.yaml
+++ b/bqetl_project.yaml
@@ -389,7 +389,9 @@ render:
   # uses {%s} which results in unknown tag exception
   - sql/mozfun/hist/string_to_json/udf.sql
 
-glean_usage:
-  override: # Skip automatically updating the following artifacts
-    skip:
-    - sql/moz-fx-data-shared-prod/mozilla_vpn/**
+generate:
+  glean_usage:
+    skip_existing: # Skip automatically updating the following artifacts
+    - sql/moz-fx-data-shared-prod/mozilla_vpn/deletion_request/**
+    - sql/moz-fx-data-shared-prod/mozilla_vpn/events/**
+    - sql/moz-fx-data-shared-prod/mozilla_vpn/main/**
--- a/bqetl_project.yaml
+++ b/bqetl_project.yaml
@@ -363,9 +363,6 @@ view:
     - sql/glam-fenix-dev/glam_etl/**/view.sql
     # tests
     - sql/moz-fx-data-test-project/test/simple_view/view.sql
-  override: # Skip automatically updating the following views
-    skip:
-    - sql/moz-fx-data-shared-prod/mozilla_vpn/**/view.sql
 
 schema:
   mozilla_pipeline_schemas_uri: https://github.com/mozilla-services/mozilla-pipeline-schemas
@@ -391,3 +388,8 @@ render:
   skip:
   # uses {%s} which results in unknown tag exception
   - sql/mozfun/hist/string_to_json/udf.sql
+
+glean_usage:
+  override: # Skip automatically updating the following artifacts
+    skip:
+    - sql/moz-fx-data-shared-prod/mozilla_vpn/**
--- a/sql_generators/glean_usage/common.py
+++ b/sql_generators/glean_usage/common.py
@@ -1,18 +1,19 @@
 """Utility functions used in generating usage queries on top of Glean."""
 
+import glob
 import logging
 import os
 import re
+from collections import namedtuple
 from pathlib import Path
 
 import requests
 from jinja2 import Environment, FileSystemLoader, TemplateNotFound
 
+from bigquery_etl.config import ConfigLoader
 from bigquery_etl.dryrun import DryRun
 from bigquery_etl.schema.stable_table_schema import get_stable_table_schemas
-from bigquery_etl.util.common import render, write_sql, get_table_dir
-from bigquery_etl.config import ConfigLoader
-import glob
+from bigquery_etl.util.common import get_table_dir, render, write_sql
 
 APP_LISTINGS_URL = "https://probeinfo.telemetry.mozilla.org/v2/glean/app-listings"
 PATH = Path(os.path.dirname(__file__))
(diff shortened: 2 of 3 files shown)


sql_generators/glean_usage/common.py
```diff
@@ -223,17 +237,31 @@ def generate_per_app_id(
...
+            artifacts = [
+                Artifact(view, "metadata.yaml", view_metadata),
+                Artifact(view, "view.sql", view_sql),
+                Artifact(table, "metadata.yaml", table_metadata),
+                Artifact(table, "query.sql", query_sql),
+            ]
 
             if not self.no_init:
-                write_sql(output_dir, table, "init.sql", init_sql, skip_existing=True)
+                artifacts.append(Artifact(table, "init.sql", init_sql))
```
sean-rose, line 250: This keeps the existing behavior of assuming an `init.sql` file, but per the logic above that sets `init_sql` the query file may use the `is_init()` macro rather than having a separate `init.sql` file.

bqetl_project.yaml
```diff
@@ -385,3 +385,8 @@ render:
   skip:
   # uses {%s} which results in unknown tag exception
   - sql/mozfun/hist/string_to_json/udf.sql
+
+glean_usage:
+  override: # Skip automatically updating the following artifacts
+    skip:
```
sean-rose, line 391: Rather than `override: skip:` I'd suggest `skip_existing:`, as I think the current structure can be misread as the skips being overrides rather than skipping overriding/overwriting (the comment does help, but IMO inherent names are the clearest documentation).

sql_generators/glean_usage/common.py
```diff
@@ -159,6 +162,17 @@ def __init__(self):
...
+    def override_skip(self):
+        """Files configured not to be overridden during generation."""
+        return [
+            file
+            for skip in ConfigLoader.get("glean_usage", "override", "skip", fallback=[])
+            for file in glob.glob(
+                skip,
+                recursive=True,
+            )
+        ]
```
sean-rose, line 174: If you agree with renaming the config `skip_existing`...
```suggestion
    def skip_existing(self):
        """Existing files configured not to be overridden during generation."""
        return [
            file
            for skip_existing in ConfigLoader.get("glean_usage", "skip_existing", fallback=[])
            for file in glob.glob(skip_existing, recursive=True)
        ]
```

sql_generators/glean_usage/common.py
```diff
@@ -223,17 +237,31 @@ def generate_per_app_id(
...
 
             if not self.no_init:
-                write_sql(output_dir, table, "init.sql", init_sql, skip_existing=True)
+                artifacts.append(Artifact(table, "init.sql", init_sql))
+
+            for artifact in artifacts:
+                destination = (
+                    get_table_dir(output_dir, artifact.table_id) / artifact.basename
+                )
+                skip_existing = destination in self.override_skip()
```
sean-rose, line 256: Calling `self.override_skip()` within a loop is less than ideal.  I'd suggest either saving the value in a variable outside the loop, or memoizing the method so it doesn't recalculate the skip list on every call.

sql_generators/glean_usage/common.py
```diff
@@ -316,10 +351,29 @@ def generate_per_app(
...
+                    ),
+                    Artifact(table, "metadata.yaml", metadata),
+                    Artifact(view, "view.sql", view_sql),
+                ]
+
+                for artifact in artifacts:
+                    destination = (
+                        get_table_dir(output_dir, artifact.table_id) / artifact.basename
+                    )
+                    skip_existing = destination in self.override_skip()
```
sean-rose, line 368: Same suggestion about calling `self.override_skip()` within a loop.

sql_generators/glean_usage/common.py
```diff
@@ -316,10 +351,29 @@ def generate_per_app(
...
-                write_sql(
-                    output_dir, table, "metadata.yaml", metadata, skip_existing=True
-                )
-                write_sql(output_dir, view, "view.sql", view_sql, skip_existing=True)
+                artifacts = [
+                    Artifact(
+                        table,
+                        "query.sql",
+                        query_sql,
+                    ),
```
sean-rose, line 359: Formatting.
```suggestion
                    Artifact(table, "query.sql", query_sql),
```

sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -162,7 +162,8 @@ def generate_per_app(
                     full_view_id,
                     "view.sql",
                     rendered_view,
-                    skip_existing=True,
+                    skip_existing=get_table_dir(output_dir, full_view_id) / "view.sql"
+                    in self.override_skip(),
```
sean-rose, line 166: - Same suggestion about calling `self.override_skip()` within a loop.
- This formatting makes it kind of confusing to read, so I'd suggest a more readable formatting (if Black allows it):
```suggestion
                    skip_existing=(
                        get_table_dir(output_dir, full_view_id) / "view.sql"
                        in self.override_skip()
                    ),
```

sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -178,7 +179,9 @@ def generate_per_app(
                         app_name=release_app["canonical_app_name"],
                         app_channels=", ".join(app_channels),
                     ),
-                    skip_existing=True,
+                    skip_existing=get_table_dir(output_dir, full_view_id)
+                    / "metadata.yaml"
+                    in self.override_skip(),
```
sean-rose, line 184: Similarly:
```suggestion
                    skip_existing=(
                        get_table_dir(output_dir, full_view_id) / "metadata.yaml"
                        in self.override_skip()
                    ),
```

sean-rose (approved):
r+wc

Will there be a similar PR for `stable_views`, which also [skips overwriting existing files](https://github.com/mozilla/bigquery-etl/blob/29dd1c31b7123a4478d17f654111400527dccba1/sql_generators/stable_views/__init__.py#L7-L9)?

And I guess this means we can close #3182.

bqetl_project.yaml
```diff
@@ -385,3 +385,8 @@ render:
   skip:
   # uses {%s} which results in unknown tag exception
   - sql/mozfun/hist/string_to_json/udf.sql
+
+glean_usage:
```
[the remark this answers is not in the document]
sean-rose: > Might be good to nest this under a top-level key like `sql_generators`

Or looking at the existing config, it seems like the general convention is the top-level keys match the associated `bqetl` commands, so in this case that'd be `generate`.

sql_generators/glean_usage/common.py
scholtzan: yeah, I think this could use some additional consideration. For now I think keeping the same behaviour and opening a follow-up ticket is the way to go

PR #4088, opened 2023-07-17
Remove buildhub2 filter in GLAM templates

Iinh:
Due to historical incidents of bad version number getting in that broke the ETL, we added a filter that only accepts build IDs that match buildhub2 builds. However, [this query](https://sql.telemetry.mozilla.org/queries/93349/source) shows many FOG builds with substantial client count are being filtered out because they do not exist in buildhub2. This affects the client count we see in GLAM. 

Considering that we already have [a guardrail to ensure only valid version numbers get in](https://github.com/mozilla/bigquery-etl/pull/3933), we should remove the buildhub2 filter. Potentially will fix https://github.com/mozilla/glam/issues/2003.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1239)

### DIFF
--- a/bigquery_etl/glam/clients_daily_histogram_aggregates.py
+++ b/bigquery_etl/glam/clients_daily_histogram_aggregates.py
@@ -111,11 +111,6 @@ def main():
         + (" --no-parameterize" if args.no_parameterize else "")
     )
 
-    # This enables build_id filtering against Buildhub data
-    # and filtering out of an erroneous version "1024" (see query template for details).
-    # Only the desktop builds are reported to Buildhub
-    filter_desktop_builds = True if args.product == "firefox_desktop" else False
-
     schema = get_schema(args.source_table)
     distributions = get_distribution_metrics(schema)
     metrics_sql = get_metrics_sql(distributions).strip()
@@ -126,7 +121,6 @@ def main():
     print(
         render_main(
             header=header,
-            filter_desktop_builds=filter_desktop_builds,
             source_table=args.source_table,
             submission_date=submission_date,
             attributes=ATTRIBUTES,
--- a/bigquery_etl/glam/templates/clients_daily_histogram_aggregates_v1.sql
+++ b/bigquery_etl/glam/templates/clients_daily_histogram_aggregates_v1.sql
@@ -1,14 +1,5 @@
 {{ header }}
-WITH
-{% if filter_desktop_builds %}
-valid_builds AS (
-  SELECT DISTINCT
-    build.build.id AS build_id
-  FROM
-    `moz-fx-data-shared-prod.telemetry.buildhub2`
-),
-{% endif %}
-extracted AS (
+WITH extracted AS (
   SELECT
     *,
     DATE(submission_timestamp) AS submission_date,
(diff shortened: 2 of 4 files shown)


bigquery_etl/glam/clients_daily_histogram_aggregates.py
```diff
@@ -111,11 +111,6 @@ def main():
         + (" --no-parameterize" if args.no_parameterize else "")
     )
 
-    # This enables build_id filtering against Buildhub data
-    # and filtering out of an erroneous version "1024" (see query template for details).
-    # Only the desktop builds are reported to Buildhub
-    filter_desktop_builds = True if args.product == "firefox_desktop" else False
```
edugfilho, line 117: Why don't we start by just setting `filter_desktop_builds` to `False`?

bigquery_etl/glam/templates/clients_daily_histogram_aggregates_v1.sql
```diff
@@ -23,20 +14,9 @@ extracted AS (
...
-  ON
-    client_info.app_build = build_id
-  {% endif %}
   WHERE
     DATE(submission_timestamp) = {{ submission_date }}
     AND client_info.client_id IS NOT NULL
-    {% if filter_desktop_builds %}
-    -- some FOG clients on Windows are sending version "1024.0.0"
-    -- see https://bugzilla.mozilla.org/show_bug.cgi?id=1768187
-    AND client_info.app_display_version != '1024.0.0'
```
edugfilho, line 38: I'm not sure if removing this check is a good idea, but can be convinced otherwise

bigquery_etl/glam/clients_daily_histogram_aggregates.py
Iinh: We can start with that to be safe, but I expect this to be fully removed at some point, as we're not using this filter anywhere, and leaving this here can be confusing in the long run.

bigquery_etl/glam/templates/clients_daily_histogram_aggregates_v1.sql
Iinh: In theory it's impossible to run into this problem again considering we now have [this guardrail that makes sure we only let in data with the correct latest version](https://github.com/mozilla/bigquery-etl/blob/main/bigquery_etl/glam/templates/latest_versions_v1.sql). We had a problem of version 999 before and it was correctly filtered out thanks to that step.

akkomar: > In theory it's impossible to run into this problem again considering we now have [this guardrail that makes sure we only let in data with the correct latest version](https://github.com/mozilla/bigquery-etl/blob/main/bigquery_etl/glam/templates/latest_versions_v1.sql). We had a problem of version 999 before and it was correctly filtered out thanks to that step.

I think this is correct.

akkomar (commented):
I think this is fine, but I'll turn to @edugfilho for a final OK as it's been a while since I worked with these queries.

bigquery_etl/glam/templates/clients_daily_histogram_aggregates_v1.sql
edugfilho: ok I've been convinced otherwise

Iinh:
Another for-argument to remove the buildhub filter (that @edugfilho pointed out):

screen_shot_2023-07-17_at_1 59 07_pm

PR #4090, opened 2023-07-17
DENG-772: Fenix population with first_session ping and min seq > 0

wwyc:
[DENG-772](https://mozilla-hub.atlassian.net/browse/DENG-772):

* As mentioned in the ticket, it was discovered some fenix clients have a `first_session_ping` but seq > 0 due to collection system error.  Changes have been made to use min `ping_info.seq` instead.

### DIFF
--- a/sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
+++ b/sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
@@ -54,6 +54,16 @@ activations AS (
     client_id
 ),
 -- Find earliest data per client from the first_session ping.
+first_session_ping_min_seq AS (
+  SELECT
+    client_info.client_id AS client_id,
+    MIN(ping_info.seq) AS seq -- Pings are sent in sequence, this guarantees that the first one is returned.
+  FROM
+    `moz-fx-data-shared-prod.fenix.first_session` AS fenix_first_session
+  WHERE
+    DATE(submission_timestamp) >= '2019-01-01'
+  GROUP BY
+    client_id ),
 first_session_ping AS (
   SELECT
     client_info.client_id AS client_id,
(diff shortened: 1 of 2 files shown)


sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
```diff
@@ -74,9 +84,12 @@ first_session_ping AS (
     ] AS adjust_creative
   FROM
     `moz-fx-data-shared-prod.fenix.first_session` AS fenix_first_session
+  INNER JOIN
```
lucia-vargas-a, line 87: The goal is to get all first_session pings and we use the seq to identify when the ping was generated on the client side rather than when it is received by the server. The INNER JOIN would exclude clients if we ever receive pings with seq = NULL and it would be a silent error if it happens. Likely a left join would solve it? Else  we may need another approach to retrieve the ping with the minimum or NULL sequence.

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/init.sql
```diff
@@ -74,9 +84,12 @@ first_session_ping AS (
...
+    first_session_ping_min_seq
+  ON
+    client_info.client_id = first_session_ping_min_seq.client_id
   WHERE
     DATE(submission_timestamp) >= '2019-01-01'
-    AND ping_info.seq = 0 -- Pings are sent in sequence, this guarantees that the first one is returned.
```
lucia-vargas-a, line 79: An option for not missing seq = NULL could be adding here:

`ping.info.seq IN (subset) OR NULL`

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
```diff
@@ -54,6 +54,16 @@ activations AS (
     client_id
 ),
 -- Find earliest data per client from the first_session ping.
+first_session_ping_min_seq AS (
+  SELECT
+    client_info.client_id AS client_id,
```
lucia-vargas-a, line 59: This would be required per sample_id to match the first_session_query?

wwyc: Can you elaborate on this?  Thanks!

lucia-vargas-a (approved):
LGTM. I guess the query gets more complex because we need to handle edge cases. Might be worth sending a message in the Fenix channel to check if there is a way to fix clients with the seq=0.

PR #4091, opened 2023-07-18
Fix processing follow up items in ParallelTopologicalSorter

scholtzan:
This fixes an issue when using `ParallelTopologicalSorter` with followup items getting added dynamically. Items can't get added to `TopologicalSorter` after `prepare()` has been called, so we need to process the followup items separately

Checklist for reviewer:


fbertsch (approved):
This looks fine, but TBH I would move the serial `followup_queue` processing to the original code in `bigquery_etl/cli/query.py`. I'd create the queue there and send it in as part of the `partial`, letting you strip out all the followup processing here (only because it's confusing to read in this file & an implementation detail of the specific function being parallelized, and it's not happening in parallel anyways).

PR #4093, opened 2023-07-18
DENG-801:  Fix firefox_android_clients activations

wwyc:
* Activations are not being captured because of the delay of 7 day in `fenix.new_profile_activation`.
* The change to a `full outer join` will fix the issue by capturing `current` activations for clients from previous `submission_dates`.


sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
```diff
@@ -262,7 +262,7 @@ _current AS (
     baseline_ping AS baseline
   USING
     (client_id)
-  LEFT JOIN
+  FULL OUTER JOIN
```
lucia-vargas-a, line 265: Looks good, we could validate that numbers don't change after a backfill or add a test --It shouldn't be the case since activations are based on baseline ping.

wwyc: It looks like there are activations that do not exist in baseline ping.  I added an inner join to the existing table.  Let me know what you think.

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
```diff
@@ -45,11 +45,13 @@ first_seen AS (
 activations AS (
```
lucia-vargas-a, line 45:
```suggestion
-- The inner join aims to exclude exclude shredded client_ids that may still exist in `fenix.new_profile_activation`.
activations AS (
```

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
```diff
@@ -45,11 +45,13 @@ first_seen AS (
...
   SELECT
     client_id,
-    ARRAY_AGG(activated ORDER BY submission_date DESC)[SAFE_OFFSET(0)] > 0 AS activated
+    ARRAY_AGG(activation.activated ORDER BY activation.submission_date DESC)[SAFE_OFFSET(0)] > 0 AS activated,
   FROM
-    fenix.new_profile_activation
+    fenix.new_profile_activation AS activation
+  INNER JOIN `moz-fx-data-shared-prod.fenix_derived.firefox_android_clients_v1` --excluding clients in activations but not in baseline ping due to shredder
```
lucia-vargas-a, line 51: Probably easier later to find the comment at the top of the subquery?

lucia-vargas-a (commented):
We're only pending to add the test for activations.

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/query.sql
ANich: Should `activation` be going through shredder as well?

lucia-vargas-a: It seems so because it has client_id. I'd suggest doing that in a separate PR and then we could remove the inner join. To my knowledge, adding it to [this script ](https://github.com/mozilla/bigquery-etl/blob/2a243c483d221d6a28ca5da7cf6975e9b159411e/bigquery_etl/shredder/config.py#L153) should be enough. What do you think Winnie?

lucia-vargas-a (approved):

PR #3873, opened 2023-05-31
Glam non norm agg

edugfilho:
Checklist for reviewer:

- [ ] Commits should reference a bug or github issue, if relevant (if a bug is referenced, the pull request should include the bug number in the title).
- [ ] If the PR comes from a fork, trigger integration CI tests by running the [Push to upstream workflow](https://github.com/mozilla/bigquery-etl/actions/workflows/push-to-upstream.yml) and provide the `<username>:<branch>` of the fork as parameter. The parameter will also show up
in the logs of the `manual-trigger-required-for-fork` CI task together with more detailed instructions.
- [ ] If adding a new field to a query, ensure that the schema and dependent downstream schemas have been updated.
- [ ] When adding a new derived dataset, ensure that data is not available already (fully or partially) and recommend extending an existing dataset in favor of creating new ones. Data can be available in the [bigquery-etl repository](https://github.com/mozilla/bigquery-etl), [looker-hub](https://github.com/mozilla/looker-hub) or in [looker-spoke-default](https://github.com/mozilla/looker-spoke-default/tree/e1315853507fc1ac6e78d252d53dc8df5f5f322b).

For modifications to schemas in restricted namespaces (see [`CODEOWNERS`](https://github.com/mozilla/bigquery-etl/blob/main/CODEOWNERS)):
- [ ] Follow the [change control procedure](https://docs.google.com/document/d/1TTJi4ht7NuzX6BPG_KTr6omaZg70cEpxe9xlpfnHj9k/edit#heading=h.ttegrcfy18ck)

PR #4121, opened 2023-07-25
Update telemetry_derived.histogram_percentiles_v1 schema

ANich:
Follow up to #3873 - looks like the checked-in schema doesn't match the deployed schema in some incompatible ways:

`[2023-07-25, 02:21:19 UTC] {pod_manager.py:235} INFO - google.api_core.exceptions.BadRequest: 400 PATCH https://bigquery.googleapis.com/bigquery/v2/projects/moz-fx-data-shared-prod/datasets/telemetry_derived/tables/histogram_percentiles_v1?prettyPrint=false: Field aggregates.KEY already exists in schema`

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/histogram_percentiles_v1/schema.yaml
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/histogram_percentiles_v1/schema.yaml
@@ -54,7 +54,7 @@ fields:
 
 - fields:
   - mode: NULLABLE
-    name: key
+    name: KEY
     type: STRING
   - mode: NULLABLE
     name: value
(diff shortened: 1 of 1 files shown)


edugfilho:
I'm willing to give it a shot, but how does capitalizing the field name change things? AFAIK column names are case-insensitive
btw thank you for that, especially for catching the missing `os`

edugfilho (approved):

ANich:
> I'm willing to give it a shot, but how does capitalizing the field name change things? AFAIK column names are case-insensitive btw thank you for that, especially for catching the missing `os`

Yeah column names are case-insensitive in queries, but I suspect changing the case isn't an allowed operation? 🤔 At least that's what the error message suggests 🤷‍♂️

PR #3951, opened 2023-06-15
DENG-1021 Added destination table param to query commands

wwyc:
[DENG-1021](https://mozilla-hub.atlassian.net/browse/DENG-1021): 

* As part of implementing the managed backfill process, backfilled data are to be first stored in a separate `destination table`.
* This PR updates the query commands to allow `deploy` and `backill` cli commands to specify a destination table.

[DENG-1021]: https://mozilla-hub.atlassian.net/browse/DENG-1021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4123, opened 2023-07-25
Fixed backfill query cli destination table name

wwyc:
A change was introduced in PR #3951 to use fully qualified destination table in query cli commands.  This PR fixes the destination table name.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1291)


bigquery_etl/cli/query.py
```diff
@@ -490,13 +490,13 @@ def _backfill_query(
...
         if destination_table is None:
-            destination_table = table
+            destination_table = f"{project}.{dataset}.{table}"
 
         if not no_partition:
             destination_table = f"{destination_table}${partition}"
 
         click.echo(
-            f"Run backfill for {project}.{dataset}.{destination_table} "
+            f"Run backfill for {destination_table} "
```
ANich, line 499: Is there a check somewhere that errors if it's not fully qualified?

ANich (approved):
👍 with one question

bigquery_etl/cli/query.py
ANich: Just seeing this sorry 😅 does this account for the `$partition` part of the destination? Given that the qualified check is after the partition check?

PR #4129, opened 2023-07-26
⚠️ do not merge - PR for planning client-regeneration clients_yearly backfills

ANich:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1299)


irrationalagent:
There also exist these tables that you could optionally use: `mozdata.analysis.regen_sim_replaced_baseline_clients_daily_with_searches_*`. 


The query that generated these tables:
```
  WITH
  base AS (
      SELECT
        c.client_id,
        c.first_seen_date AS first_seen_date,
        c.submission_date,
        c.country,
        c.device_model,
        COALESCE(a.searches, 0) AS searches,
        COALESCE(a.searches_with_ads, 0) AS searches_with_ads,
        COALESCE(a.ad_clicks, 0) AS ad_clicks,
      FROM `mozdata.fenix.baseline_clients_daily` c
      LEFT JOIN `mozdata.fenix.attributable_clients_v2` a
      USING(client_id, submission_date)
      WHERE c.submission_date >= DATE("{str(start_date)}")
  ),

  replaced AS (
      SELECT
        COALESCE(r.replacement_id, c.client_id) AS client_id,
        COALESCE(r.first_seen_date, c.first_seen_date) AS first_seen_date,
        udf.safe_sample_id(COALESCE(r.replacement_id, c.client_id)) AS sample_id,
        c.country,
        c.device_model,
        c.submission_date,
        ARRAY_AGG(c.client_id IGNORE NULLS ORDER BY regened_last_date) AS original_client_id,
        ARRAY_AGG(c.first_seen_date IGNORE NULLS ORDER BY regened_last_date) AS original_first_seen_date,
        SUM(searches) AS searches,
        SUM(searches_with_ads) AS searches_with_ads,
        SUM(ad_clicks) AS ad_clicks,
      FROM base c
      LEFT JOIN `mozdata.analysis.regen_sim_client_replacements_{str(seed)}` r
      -- we want the history starting on the regen date.
      ON (c.client_id = r.client_id) AND (c.submission_date BETWEEN r.regen_date AND r.regened_last_date)
      GROUP BY 1,2,3,4,5,6
  )

  SELECT
      *
  FROM replaced
```

The difference is that these tables contain search metrics. This is done by first left joining clients daily to firefox attributable clients. Then we do the replacement - you'll notice that in these query to dedupe the replacement ids (from cases where one ID ends a replacement and starts a new one on the same day) we use a group by instead of a window function. This sums the searches etc from the two original IDs together. We also `array_agg` the original IDs in case we need to back-correlate for some reason.

Both the tables with searches and without will return the same counts for DAU. However, if we want to use them for client-level activity metrics like searches, we should use the `with_searches` version. It would be a one-line change in the code you have here.

PR #4137, opened 2023-07-27
Added new_profile_activation_v1 to shredder

wwyc:
It was discovered in #4093 that the following table contain `client_id` but are not currently being shredded:
- `moz-fx-data-shared-prod.fenix_derived.new_profile_activation_v1`

This PR adds it to the delete target list so it will also be shredded.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1310)

### DIFF
--- a/bigquery_etl/shredder/config.py
+++ b/bigquery_etl/shredder/config.py
@@ -96,7 +96,7 @@ CONTEXTUAL_SERVICES_SRC = DeleteSource(
     table="telemetry_stable.deletion_request_v4",
     field="payload.scalars.parent.deletion_request_context_id",
 )
-FENIX_SRC = DeleteSource(table="fenix.deletion_request", field=CLIENT_ID)
+FENIX_SRC = DeleteSource(table="fenix.deletion_request", field=GLEAN_CLIENT_ID)
 FXA_HMAC_SRC = DeleteSource(
     table="firefox_accounts_derived.fxa_delete_events_v1", field="hmac_user_id"
 )


bigquery_etl/shredder/config.py
```diff
@@ -96,6 +96,7 @@ def fields(self) -> tuple[str, ...]:
     table="telemetry_stable.deletion_request_v4",
     field="payload.scalars.parent.deletion_request_context_id",
 )
+FENIX_SRC = DeleteSource(table="fenix.deletion_request", field=CLIENT_ID)
```
relud, line 99: `client_id` isn't a field in `fenix.deletion_request`
```suggestion
FENIX_SRC = DeleteSource(table="fenix.deletion_request", field=GLEAN_CLIENT_ID)
```

relud (approved):
r+wc

bigquery_etl/shredder/config.py
```diff
@@ -154,6 +155,7 @@ def fields(self) -> tuple[str, ...]:
 context_id_target = partial(DeleteTarget, field=CONTEXT_ID)
 
 DELETE_TARGETS: DeleteIndex = {
+    client_id_target(table="fenix_derived.new_profile_activation_v1"): FENIX_SRC,
```
relud, line 158: weren't there two tables here before?

lucia-vargas-a (approved):

bigquery_etl/shredder/config.py
lucia-vargas-a: Winnie, shall we add a Jira task to add firefox_android_clients_v1 after DS-2998, so we don't forget?

PR #4142, opened 2023-07-31
bug(1844385): bqetl_acoustic_contact_export contacts_raw_v1 failing due to invalid schema update error.

kik-kik:
# bug(1844385): bqetl_acoustic_contact_export contacts_raw_v1 failing due to invalid schema update error.

As we started getting unexpected values (`yes` and `no` for some fields we previously assumed to be of numeric value only):

- ensuring the `double_opt_in` field is of type `INTEGER`, this includes conversion of `Yes` to `1` and `No` to `0`.
- ensuring the `fxa_account_deleted` field is of type `INTEGER`, this includes conversion of `Yes` to `1` and `No` to `0`.
- ensuring the `has_opted_out_of_email` field is of type `INTEGER`, this includes conversion of `Yes` to `1` and `No` to `0`.

Additionally, to avoid wasted cost by attempting to re-run the query (if the query is invalid, retrying is pointless)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1318)

### DIFF
--- a/sql/moz-fx-data-marketing-prod/acoustic/contact_raw_v1/metadata.yaml
+++ b/sql/moz-fx-data-marketing-prod/acoustic/contact_raw_v1/metadata.yaml
@@ -21,6 +21,7 @@ owners:
 scheduling:
   dag_name: bqetl_acoustic_contact_export
   date_partition_parameter: submission_date
+  retries: 0
   referenced_tables:
     - [
       'moz-fx-data-bq-fivetran',
--- a/sql/moz-fx-data-marketing-prod/acoustic/contact_raw_v1/metadata.yaml
+++ b/sql/moz-fx-data-marketing-prod/acoustic/contact_raw_v1/metadata.yaml
@@ -21,7 +21,6 @@ owners:
 scheduling:
   dag_name: bqetl_acoustic_contact_export
   date_partition_parameter: submission_date
-  retries: 0
   referenced_tables:
     - [
       'moz-fx-data-bq-fivetran',
--- a/sql/moz-fx-data-marketing-prod/acoustic/contact_raw_v1/query.sql
+++ b/sql/moz-fx-data-marketing-prod/acoustic/contact_raw_v1/query.sql
@@ -1,5 +1,19 @@
+CREATE TEMP FUNCTION yes_no_to_int(val STRING) AS (
+  CASE
+    WHEN LOWER(val) = "no"
+      THEN 0
+    WHEN LOWER(val) = "yes"
+      THEN 1
+    ELSE CAST(val AS INTEGER)
+  END
+);
+
 SELECT
-  * EXCEPT (_file, _line, _fivetran_synced, _modified, last_modified_date),
+  * EXCEPT (_file, _line, _fivetran_synced, _modified, last_modified_date) REPLACE(
+    yes_no_to_int(double_opt_in) AS double_opt_in,
+    yes_no_to_int(fxa_account_deleted) AS fxa_account_deleted,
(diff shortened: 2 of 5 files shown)


lelilia:
you can't change the dag.py directly because the circleci will complain about that.
I added it to the metadata of the query but that only adds those to that task  ... not sure how to set it for the over all dag

lelilia:
why a temp function and not a udf?

lelilia:
any other non-integer val will still through an error

sql/moz-fx-data-marketing-prod/acoustic/contact_raw_v1/query.sql
```diff
@@ -1,5 +1,19 @@
+CREATE TEMP FUNCTION yes_no_to_int(val STRING) AS (
+  CASE
+    WHEN LOWER(val) = "no"
+      THEN 0
+    WHEN LOWER(val) = "yes"
+      THEN 1
+    ELSE CAST(val AS INTEGER)
```
lucia-vargas-a, line 7: I don't have visibility on the contents of these fields but since they're all type STRING, this would fail for any string different than (yes, no). Shall we add a case for safety?
```suggestion
    WHEN NULL
      CAST(val AS INTEGER)
    ELSE
      CAST(NULL AS INTEGER)
```

lelilia, line 7:
```suggestion
    ELSE SAFE_CAST(val AS INTEGER)
```

lelilia: not sure what `WHEN NULL` would do here, but probably not what we want? I suggest SAFE_CAST which will `null` everything that can't be cast

sql/moz-fx-data-marketing-prod/acoustic/contact_raw_v1/metadata.yaml
```diff
@@ -21,6 +21,7 @@ owners:
 scheduling:
   dag_name: bqetl_acoustic_contact_export
   date_partition_parameter: submission_date
+  retries: 0
```
sean-rose, line 24: > Additionally, to avoid wasted cost by attempting to re-run the query (if the query is invalid, retrying is pointless)

An invalid query isn't the only way this could fail.  I'd suggest keeping the default retries, otherwise something like a transient BigQuery service issue could cause this to sit failed until someone manually reruns it.

sql/moz-fx-data-marketing-prod/acoustic/contact_raw_v1/query.sql
kik-kik: huh? Where did that come from? 🤔

sean-rose: Given we've already had unforeseen changes to the data I'd actually recommend _not_ using `SAFE_CAST()` and leaving this logic as is, because if the data changes yet again (e.g. maybe they start sending "true" and "false" strings) we want this to fail so we can potentially update the ETL rather than nulling out anything we're not expecting and effectively losing data.

sql/moz-fx-data-marketing-prod/acoustic/contact_raw_v1/metadata.yaml
kik-kik: Honestly in this case I'm okay to then manually kick-start the task, I'll bump it up to `1` I feel like that's enough since this is not something I'd expect to see much.

lelilia (approved):
lgtm

PR #4158, opened 2023-08-03
bug(DENG-709): fxa unifying event filtering

kik-kik:
# bug(DENG-709): fxa unifying event filtering

We noticed that `fxa_users_services_devices_last_seen_v1` was using incorrect filtering rules. Currently this specific last seen table is incorrect due to us only considering login or registration events when generating the last seen table which yields incorrect results.

There has been discussion on which events should be considered when building these tables and this update reflects this across all `fxa_user_*` queries.

After this change is merged, we should review the resulting tables and ensure that we are producing the data in a form that is needed to perform our analysis.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1349)


sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_devices_daily_v1/query.sql
```diff
@@ -30,20 +30,13 @@ WITH fxa_events AS (
     AND fxa_log IN ('content', 'auth', 'oauth')
     -- re-using the filter from users_services_daily_v1 for consistency across the models
     -- at some point in the future we should re-evaluate this list
```
akkomar, line 32: This PR looks like we're re-evaluating this list, so how about we remove this comment?
We can maybe add a comment saying that three fxa_activity events are enough to detect user's activity.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_last_seen_v2/query.sql
```diff
@@ -0,0 +1,112 @@
+-- TODO: for now this is a copy of v1, we still need to adjust and make sure
+-- no other changes are required.
+
+CREATE TEMP FUNCTION contains_fxa_activity_event(events ANY TYPE) AS (
```
akkomar, line 4: This seems unused.

akkomar (changes requested):

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_services_devices_daily_v1/query.sql
kik-kik: I missed that, thanks for pointing it out.

akkomar (approved):

Issue #4135, opened 2023-07-27
Create a job to publish all metadata


Iinh:
After #3909, logic was added to include `friendly_name` and `description` in view publishing. However, since views are only published on change for performance reasons, many views were not updated, affecting downstream tools such as Data Catalog. 

Anna proposed we "schedule a job that publishes all metadata: views and otherwise."

PR #4159, opened 2023-08-03
Include metadata check in view publishing

Iinh:
Instead of writing a separate job as mentioned in #4135, we can add a check to see if metadata content has changed so views can still get updated without needing to have SQL changes.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1350)


bigquery_etl/view/__init__.py
```diff
@@ -277,9 +277,22 @@ def has_changes(self, target_project=None):
...
                 return True
 
-        if self.labels != table.labels:
-            print(f"view {target_view_id} will change: labels do not match")
-            return True
+        # check metadata
+        metadata_file = Path(self.path).parent / METADATA_FILE
+        if metadata_file.is_file():
+            view_metadata = Metadata.from_file(metadata_file)
+            if view_metadata.description != table.description:
```
ANich, line 284: nit: 
```suggestion
        if self.metadata is not None:
            if self.metadata.description != table.description
```

bigquery_etl/view/__init__.py
```diff
@@ -277,9 +277,22 @@ def has_changes(self, target_project=None):
...
+                print(f"view {target_view_id} will change: description does not match")
+                return True
+            if view_metadata.friendly_name != table.friendly_name:
```
ANich, line 287: I can't remember what we concluded about this offline 😅 I remember that view `friendly_name` isn't surfaced in UI's but it is saved.. We still wanted to populate it right?

bigquery_etl/view/__init__.py
```diff
@@ -277,9 +277,22 @@ def has_changes(self, target_project=None):
...
+                print(
+                    f"view {target_view_id} will change: friendly_name does not match"
+                )
+                return True
+            if view_metadata.labels != table.labels:
```
ANich, line 292: Is `self.labels` modified somewhere via processing outside of the metadata file? Curious about the change from `self.labels` to the metadata file labels

bigquery_etl/view/__init__.py
ANich: (and so on with `friendly_name` and potentially `labels` below)

ANich (commented):
Left a few comments, only blocking one for me is `self.labels` vs. `view_metadata.labels`

bigquery_etl/view/__init__.py
Iinh: yeah I think we agreed on still populating it just in case, but I can remove this if you don't think it's necessary

bigquery_etl/view/__init__.py
Iinh: changed it back to `self.labels`

bigquery_etl/view/__init__.py
ANich: It's fine to keep, I'd recommend monitoring this after merging to see if it's causing views to always seem changed

ANich (approved):

PR #4164, opened 2023-08-07
bug(1847233): updated checks.sql for app_store_funnel_v1

kik-kik:
# bug(1847233): updated checks.sql for app_store_funnel_v1

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1362)


lelilia (approved):
looks good to me
as far as I understand: it's not incremental so we don't need a submission date and date and country should be unique since they are grouped
do we need "\`date\`" by any chance? not sure .. try to avoid keywords

PR #4205, opened 2023-08-18
July 4th update VPN Countries

ANich:
Like most mappings, this should definitely be done via an actual API or by processing data. 

Are there any reliable data sources for this?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1422)


sean-rose (approved):

sean-rose:
> Are there any reliable data sources for this?

Not that I'm aware of.  We do track countries associated with VPN subscriptions, but for whatever reasons sometimes subscriptions can be outside of the countries our VPN is officially available in (currently I see 490 subscriptions in 47 other countries).

sean-rose:
For posterity, this reflects the VPN "wave 6" expansion ([VPN-4948](https://mozilla-hub.atlassian.net/browse/VPN-4948)).

PR #4213, opened 2023-08-28
DENG1381 - Add bqetl support for deprecation metadata

alekhyamoz:
Implements - https://mozilla-hub.atlassian.net/browse/DENG-1381

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1438)


bigquery_etl/cli/metadata.py
```diff
@@ -0,0 +1,75 @@
...
+                    dict(
+                        role="roles/bigquery.metadataViewer",
+                        members=["workgroup:deprecated"],
+                    )
+                ]
+            else:
+                if dataset_metadata.default_table_workgroup_access and (
+                    table_metadata.workgroup_access == []
+                    or table_metadata.workgroup_access is None
+                ):
```
scholtzan, line 66:
```suggestion
                # if workgroup hasn't been explicitly set for the table, set table workgroup
                # to the default workgroup that is defined in the dataset_metadata.yaml
                if dataset_metadata.default_table_workgroup_access and table_metadata.workgroup_access is None
                ):
```

If `workgroup_access: []` is set explicitly we probably don't want to overwrite it

bigquery_etl/cli/metadata.py
```diff
@@ -0,0 +1,75 @@
...
+                    table_metadata.workgroup_access = (
+                        dataset_metadata.default_table_workgroup_access
+                    )
```
scholtzan, line 69: This assumes that workgroups for restricted tables are set in the `metadata.yaml` or as the `default_table_workgroup_access` in the dataset metadata. Otherwise, the default will be applied which is `roles/bigquery.dataViewer`.

We could add a check based on whether `_restricted in dataset_metdata.dataset_base_acl` and not overwrite the `table_metadata.workgroup_access` for these cases. 
@whd do you have a preference here?

whd: I don't think we need to special case the table workgroup access metadata generation logic based on whether `dataset_base_acl` is a restricted type or not. The logic should be the same in either case, but that's assuming generation logic that looks something like:

`dataset_metadata.yaml` generation logic:
* if `dataset_metadata.yaml` doesn't exist, some default metadata are generated that include `workgroup_access`
example https://github.com/mozilla/bigquery-etl/tree/main/sql/moz-fx-data-shared-prod/burnham_derived => https://github.com/mozilla/bigquery-etl/blob/generated-sql/sql/moz-fx-data-shared-prod/burnham_derived/dataset_metadata.yaml#L7
* if `dataset_metadata.yaml` does exist, it is guaranteed to include `workgroup_access` (this might not be true but it should be, and from checking `main` is de facto true)
* if `default_table_workgroup_access` doesn't exist in `dataset_metadata.yaml`, it is set to the value of `workgroup_access` (which is guaranteed to exist) <- I haven't read the full PR but this is probably the piece that's missing

At this point all `dataset_metadata.yaml` files should have both `workgroup_access` and `default_table_workgroup_access`.

`metadata.yaml` `workgroup_access` table generation logic then looks like:
* if `workgroup_access` doesn't exist in a table's `metadata.yaml`, generate it from `default_table_workgroup_access`
* if `workgroup_access` does exist, leave the value as-is except in the the special case where a `deprecated` annotation exists that is used to override it

I do think we should have additional CI for making sure we're using restricted dataset variants correctly (https://mozilla-hub.atlassian.net/browse/DENG-1400) but for the purposes of metadata generation the logic is the same.

alekhyamoz:
@whd  I have updated the PR with your suggestions. Please take a look at your convenience.

bigquery_etl/cli/metadata.py
```diff
@@ -0,0 +1,87 @@
...
+            # to the default workgroup that is defined in the dataset_metadata.yaml
+            if (
+                dataset_metadata.default_table_workgroup_access
+                and table_metadata.workgroup_access is None
+            ):
+                print("im insider the if")
+                table_metadata.workgroup_access = (
+                    dataset_metadata.default_table_workgroup_access
+                )
+        dataset_metadata.write(dataset_metadata_path)
```
whd, line 84: This looks unnecessary. If dataset metadata was manipulated above it was already written to disk then.
```suggestion
```

tests/data/test_sql/moz-fx-data-test-project/test/dataset_metadata.yaml
```diff
@@ -0,0 +1,14 @@
+friendly_name: Test Dataset
+description: |-
+  Use for testing.
+dataset_base_acl: view
+user_facing: true
+labels: {}
+default_table_workgroup_access:
+- role: roles/bigquery.metadataViewer
```
whd, line 8: nit: you'd generally expect the roles to be reversed for `default_table_workgroup_access` and `workgroup_access`.

tests/sql/moz-fx-data-shared-prod/telemetry_derived/dataset_metadata.yaml
```diff
@@ -0,0 +1,13 @@
+friendly_name: User-Defined Functions
```
whd, line 1: nit: while these are for tests, mixing test metadata for `telemetry_derived` with what appears to have been metadata copied from `udf` is a bit confusing.

whd (commented):
From running a couple of tests locally in combination with a few updates to https://github.com/mozilla-services/cloudops-infra/pull/4971, this looks mostly ready to land to me.

The main thing I think should be changed is that when `deprecated` metadata exists, I recommend setting `workgroup_access` to:

```yaml
workgroup_access: []
```

instead of:

```yaml
workgroup_access:
- role: roles/bigquery.metadataViewer
  members:
  - workgroup:deprecated
```

This will simplify the logic in cloudops-infra. Since we have explicit `deprecated: true` metadata I think we should avoid using `workgroup:deprecated` entirely, except perhaps for datasets. Even for datasets, we might prefer to extend `deprecated: true` usage there. Basically I think `deprecated: true` is sufficiently explicit and easy to understand, and its direct translation to `workgroup_access: []` is less confusing than granting access to `workgroup:deprecated` (which then maps to `[]`) on some arbitrary role(s).

bigquery_etl/cli/metadata.py
```diff
@@ -0,0 +1,87 @@
...
+    table_metadata_files = paths_matching_name_pattern(
+        name, sql_dir, project_id=project_id, files=["metadata.yaml"]
+    )
+    dataset_metadata_path = None
+    # create and populate the dataset metadata yaml file if it does not exist
+    for table_metadata_file in table_metadata_files:
+        dataset_metadata_path = (
+            Path(table_metadata_file).parent.parent / "dataset_metadata.yaml"
+        )
+        if not dataset_metadata_path.exists():
```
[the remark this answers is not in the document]
alekhyamoz: @scholtzan please let me know your thoughts on this. I almost can't remember why we added the `if not dataset_metadata_path.exists():`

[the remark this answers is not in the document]
scholtzan: There are some datasets that don't have `dataset_metadata.yaml`: https://github.com/mozilla/bigquery-etl/tree/main/sql/moz-fx-data-shared-prod/messaging_system_derived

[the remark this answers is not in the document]
scholtzan: We should probably add them though, so that we can assume a `dataset_metadata.yaml` exists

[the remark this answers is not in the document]
whd: I meant by `i.e. it is already true that this file exists in generated branches` that the file gets generated at some point: https://github.com/mozilla/bigquery-etl/blob/generated-sql/sql/moz-fx-data-shared-prod/messaging_system_derived/dataset_metadata.yaml. So my assumption is that we already have the generation logic somewhere for this and have had it for some time.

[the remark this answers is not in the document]
whd: I did notice when double checking this that https://github.com/mozilla/bigquery-etl/tree/generated-sql/sql/moz-fx-data-shared-prod/udf_js_lib doesn't get a `dataset_metadata.yaml` file, but that appears to be because this isn't a dataset. At any rate, `dataset_metadata.yaml` is a de facto requirement for a dataset to be generated from bqetl metadata and requiring its existence de jure would be reasonable.

alekhyamoz:
@whd and @scholtzan I have updated the code based on your comments. 
Added https://github.com/mozilla/bigquery-etl/blob/bd3a9ae0b2e9c289ac93e67e426189d78749c2fd/bigquery_etl/cli/metadata.py#L51 to ignore dirs that do not have dataset metadata yaml file.

bigquery_etl/cli/metadata.py
```diff
@@ -0,0 +1,72 @@
...
+            continue
+        dataset_metadata = DatasetMetadata.from_file(dataset_metadata_path)
+        table_metadata = Metadata.from_file(table_metadata_file)
+
+        # set dataset metadata default_table_workgroup_access to table_workgroup_access if not set
+        if not dataset_metadata.default_table_workgroup_access:
+            dataset_metadata.default_table_workgroup_access = (
+                dataset_metadata.workgroup_access
+            )
```
whd, line 60: We should persist updates to `dataset_metadata.yaml`
```suggestion
            )
            dataset_metadata.write(dataset_metadata_path)
```

whd (approved):
r+wc

bigquery_etl/metadata/parse_metadata.py
```diff
@@ -405,8 +411,14 @@ class DatasetMetadata:
     dataset_base_acl: str = attr.ib()
     user_facing: bool = attr.ib(False)
     labels: Dict = attr.ib({})
+    default_table_workgroup_access: Optional[List[Dict[str, Any]]] = attr.ib(None)
     workgroup_access: list = attr.ib(DEFAULT_WORKGROUP_ACCESS)
 
+    def __attrs_post_init__(self):
+        """Set default table workgroup access to workgroup access."""
```
whd, line 418: Actually, it appears that this block makes it so that a default value is always available, which means the logic in most recent suggestion won't do anything. I suppose that explains some earlier logic that was writing this file all the time redundantly. It seems fine to do that but there are probably ways to avoid it as well.

bigquery_etl/cli/metadata.py
whd: Alternative that rewrites the file a bunch of times:
```suggestion
            )
        dataset_metadata.write(dataset_metadata_path)
```

alekhyamoz: I have removed this line of code to avoid redundancy.

whd: I don't think this line is redundant. In earlier reviews I thought it was (or at least unnecessary), but from my testing, it's required for generated `dataset_metadata.yaml`  files to have `default_table_workgroup_access`. Without this line, those values are used by bqetl logic correctly but aren't persisted to the generated branches. This line (versus the line above) causes a lot of unnecessary writes to the file, but that's realistically fine.

I want this metadata in the generated branches because it's leveraged by ops logic to determine which ACLs are safe to propagate automatically.

alekhyamoz: Thank you for the clarification!

whd:
https://github.com/mozilla/private-bigquery-etl/commit/dbc0d01b41c947833718e926ba2f1ce8ab58af31 has the generated diff from this change. From my cursory scan it looks correct with some miscellaneous formatting changes but does have some interesting effects:

* Some comments from other generation logic e.g. `# Generated by bigquery_etl.dependency`, `# Generated via ./bqetl generate stable_views` are lost due to the final stage metadata enrichment.

* Some fields e.g. `friendly_name` get explicitly set to `null` when unset. This shouldn't have any downstream effects but I'm not sure.

@scholtzan do you think it's worth trying to preserve comments in generated metadata?

scholtzan:
There shouldn't be downstream effects from this.

Afaik we can't preserve comments. Once we parse the metadata files and enrich them with additional information the comments are lost. I think we'd need to use a different library for working with yaml files, `ruamel` should be preserving comments, for example. But making those changes would be more significant, and I don't think we consume comments from metadata files anywhere.

PR #4218, opened 2023-08-29
Speed up GLAM histogram bucket counts

scholtzan:
Rewriting the `CROSS JOIN` in `telemetry_derived.clients_histogram_bucket_counts_v1` as `UNION ALL`. This reduced execution time by about 40% for some sampled data I tested it on.

The thing that is different with the `UNION`ed queries is that it can aggregate data for each of the `os` and `app_build_id` combinations in parallel on smaller datasets. Before, BigQuery ran the aggregations in a single step on a larger dataset which had all the combinations. 

One thing to note is that while the elapsed/felt time is 40% lower, this query is technically less efficient and consumes slightly more slot time and shuffles more bytes. I don't think that will be an issue, since it doesn't affect the cost and having these jobs finish quite a bit faster is desirable.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1445)


edugfilho (approved):
looking forward to seeing this in action!

PR #4225, opened 2023-08-31
Fix publishing udfs that use backticks in identifiers

scholtzan:
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1453)

### DIFF
--- a/bigquery_etl/routine/parse_routine.py
+++ b/bigquery_etl/routine/parse_routine.py
@@ -191,10 +191,12 @@ class RawRoutine:
         routines = get_routines(project)
         dependencies = []
         for udf in routines:
-            persistent_name = ".".join([f"`{name}`" for name in udf["name"].split(".")])
-            if udf["name"] in "\n".join(definitions) or persistent_name in "\n".join(
-                definitions
-            ):
+            udf_re = re.compile(
+                r"\b"
+                + r"\.".join(f"`?{name}`?" for name in udf["name"].split("."))
+                + r"\("
+            )
+            if udf_re.search("\n".join(definitions)):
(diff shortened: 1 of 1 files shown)


bigquery_etl/routine/parse_routine.py
```diff
@@ -191,7 +191,10 @@ def from_file(cls, path):
         routines = get_routines(project)
         dependencies = []
         for udf in routines:
-            if udf["name"] in "\n".join(definitions):
+            persistent_name = ".".join([f"`{name}`" for name in udf["name"].split(".")])
+            if udf["name"] in "\n".join(definitions) or persistent_name in "\n".join(
+                definitions
+            ):
```
sean-rose, line 197: Couldn't the UDF identifiers be quoted multiple different ways? (`` `foo`.`bar` ``, `` `foo.bar` ``, `` `foo`.bar ``)

Here's a possible regular expression approach that handles the quoting variations, as well as guards against accidentally matching subsections of UDF identifiers (the existing logic would have a `bar.baz` UDF match a `foobar.baz` reference, or a `foo.bar` UDF match a `foo.barbaz` reference).
```suggestion
            udf_re = re.compile(
                r"\b"
                + r"\.".join(f"`?{name}`?" for name in udf["name"].split("."))
                + r"\("
            )
            if udf_re.search("\n".join(definitions)):
```

sean-rose (approved):
r+wc

bigquery_etl/routine/parse_routine.py
scholtzan: yup, a regex seems like the best option here

PR #4226, opened 2023-08-31
initial commit for derived jobs_by_organization table for mozdata, sh…

Marlene-M-Hirose:
…ared_prod, marketing_prod, bq_data_science projects

PR adds query.py, schema.yaml and metadata.yaml files. This job pulls data from  "mozdata", "moz-fx-data-shared-prod", "moz-fx-data-marketing-prod", "moz-fx-data-bq-data-science" region-us.INFORMATION_SCHEMA.JOBS_BY_ORGANIZATION tables and puts it into one table that can be queried in the BQ console

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1454)


whd (changes requested):
There has been persistent confusion about the role of `PROJECT_ID` in [`JOBS_BY_ORGANIZATION`](https://cloud.google.com/bigquery/docs/information-schema-jobs-by-organization). I mentioned it when reviewing https://github.com/mozilla/bigquery-etl/pull/3978#discussion_r1254863914, but essentially, the docs don't do a great job of of telling you that querying all of the following views will produce the exact same result:

```
    mozdata.region-us.INFORMATION_SCHEMA.JOBS_BY_ORGANIZATION,
    moz-fx-data-shared-prod.region-us.INFORMATION_SCHEMA.JOBS_BY_ORGANIZATION,
    moz-fx-data-marketing-prod.region-us.INFORMATION_SCHEMA.JOBS_BY_ORGANIZATION,
    moz-fx-data-bq-data-science.region-us.INFORMATION_SCHEMA.JOBS_BY_ORGANIZATION
    `region-us`.INFORMATION_SCHEMA.JOBS_BY_ORGANIZATION
```

i.e. the `PROJECT_ID` field is used only to determine what project to run the query from and has no other use; it can be omitted entirely (the docs simply mention the id is optional). What you need to do in order to limit the output of these queries to only those jobs that reference tables in the data platform is to filter in the query to rows whose `project_id`,`destination_table` or at least one of the `referenced_tables` include a project id among the four listed projects. I believe the aforementioned PR ending up doing that so you can probably use that as a basis for the table here: you should only need to run one query and not one for each project.

Marlene-M-Hirose:
Right. So add this section?

        WHERE
          (project_id IN UNNEST({DEFAULT_PROJECTS})
            OR referenced_tables.project_id IN UNNEST({DEFAULT_PROJECTS})
            OR destination_table.project_id IN UNNEST({DEFAULT_PROJECTS}))

sql/moz-fx-data-shared-prod/monitoring_derived/jobs_by_organization_v1/query.py
```diff
@@ -0,0 +1,99 @@
...
+          query_info.query_hashes.normalized_literals,
+          query_info.performance_insights,
+          transferred_bytes,
+          materialized_view_statistics,
+          DATE(creation_time) as creation_date
+        FROM
+          `{project}.region-us.INFORMATION_SCHEMA.JOBS_BY_ORGANIZATION`
+        WHERE
+          DATE(creation_time) = '{date}'
+          AND (t1.project_id IN UNNEST({DEFAULT_PROJECTS})
```
whd, line 71: It appears the `bigquery_usage_v2` table uses:
```sql
        FROM
          `{project}.region-us.INFORMATION_SCHEMA.JOBS_BY_ORGANIZATION` AS t1
        LEFT JOIN
          UNNEST(referenced_tables) AS referenced_tables
```

which is why `t1` exists in that query but not here. If I were writing this I would probably write something like:
```sql
AND (project_id IN UNNEST({DEFAULT_PROJECTS})
OR EXISTS(SELECT * FROM UNNEST(referenced_tables) AS x WHERE x.project_id IN UNNEST({DEFAULT_PROJECTS}))
OR destination_table.project_id IN UNNEST({DEFAULT_PROJECTS}))
 ```
to match the upstream `JOBS_BY_ORGANIZATION` as closely as possible, but I'm not a SQL expert.

File: dags.yaml [hunk 2]
```diff
@@ -1179,3 +1179,24 @@ bqetl_default:
   tags:
     - impact/tier_3
     - triage/no_triage
+
```
whd, line 1182: Does this need to be in a separate DAG from `bqetl_monitoring`?

sql/moz-fx-data-shared-prod/monitoring_derived/jobs_by_organization_v1/query.py
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env python3
+
+"""
+
+Put data from the projects below into a table.
+
+  mozdata.region-us.INFORMATION_SCHEMA.JOBS_BY_ORGANIZATION,
```
whd, line 7: See previous comment about simplifying this since JOBS_BY_ORGANIZATION is the same no matter which project is used to query it.

whd (commented):
Disclaimer: while I was reviewing this a couple of commits were pushed so some of the things I wrote might be out of date.

dags.yaml
Marlene-M-Hirose: probably not, I will move it

Marlene-M-Hirose: DAG moved

sql/moz-fx-data-shared-prod/monitoring_derived/jobs_by_organization_v1/query.py
Marlene-M-Hirose: I've updated the code to what you have written with a change in the reference table name. Thinking if there is a clearer way to write it

sql/moz-fx-data-shared-prod/monitoring_derived/jobs_by_organization_v1/query.py
```diff
@@ -0,0 +1,93 @@
...
+            (
+              project_id IN UNNEST({DEFAULT_PROJECTS}
+              )
+            OR EXISTS
+            (
+              SELECT * FROM UNNEST(referenced_tables) AS ref_tables
+                WHERE ref_tables.project_id IN UNNEST({DEFAULT_PROJECTS})
+                )
+            OR (destination_table.project_id IN UNNEST({DEFAULT_PROJECTS}))
+            )
```
whd, line 69: I think there's one more missing `)` to balance this expression.

whd (approved):
r+wc

I ran `create_query` locally and pasted the SQL into the BQ console. Apart from needing one more `)` the query was valid.

sql/moz-fx-data-shared-prod/monitoring_derived/jobs_by_organization_v1/query.py
```diff
@@ -48,8 +48,8 @@ def create_query(date, project):
           total_slot_ms ,
           transaction_id,
           user_email,
-          query_info.resource_warning ,
-          query_info.query_hashes.normalized_literals,
+          query_info.resource_warning as query_info_resource_warning,
+          query_info.query_hashes_normalized_literals as query_info_query_hashes_normalized_literals,
```
whd, line 52: It looks like this was changed after I ran my test query and r+'d the PR. This first identifier should probably be `query_info.query_hashes.normalized_literals `

PR #4232, opened 2023-09-01
Add markers to check cli command to differentiate DQ warnings from failures

alekhyamoz:
The PR is WIP. 
With the current implementation of ETL data checks, the failure in DQ checks will stop the downstream tasks in the Airflow DAG from executing. 
This PR  enables the addition of checks as warnings, that will allow downstream tasks to run.

Pending-
Add the correct dependency between checks, warnings, and downstream tasks.

Checklist for reviewer:


bigquery_etl/query_scheduling/generate_airflow_dags.py
```diff
@@ -95,16 +95,41 @@ def get_dags(project_id, dags_config):
...
-                        tasks.append(checks_task)
-                        task_ref = TaskRef(
-                            dag_name=task.dag_name,
-                            task_id=task.task_name,
-                        )
-                        checks_task.upstream_dependencies.append(task_ref)
+                        with open(checks_file, "r") as file:
+                            file_contents = file.read()
+                        # check if file contains fail and warn and create checks task accordingly
+                        if "#fail" in file_contents:
```
kik-kik, line 101: Would it be useful to add some sort of delimiter to the checks list in `checks.sql` so that we could do a split and then bucket error checks separately from "warning" (soft fail) checks. so that we could run them as two separate tasks? This would allow us to run "warning" checks prior to running error checks?

bigquery_etl/query_scheduling/generate_airflow_dags.py
```diff
@@ -95,16 +95,41 @@ def get_dags(project_id, dags_config):
...
+                                )
+                            )
+                            tasks.append(checks_warn_task)
+                            task_ref = TaskRef(
+                                dag_name=task.dag_name,
+                                task_id=task.task_name,
+                            )
+                            checks_warn_task.upstream_dependencies.append(task_ref)
                     tasks.append(task)
+            print(tasks)
```
scholtzan, line 132:
```suggestion
```

bigquery_etl/query_scheduling/generate_airflow_dags.py
scholtzan: I guess currently we assume that if no marker is not specified for a test then `#fail` should be the default. So there could be files that have no `#fail` explicitly in them, but still a bunch of checks that should fail.

bigquery_etl/query_scheduling/generate_airflow_dags.py
```diff
@@ -95,16 +95,41 @@ def get_dags(project_id, dags_config):
...
+                                    is_check_task_fail,
+                                    dag_collection=dag_collection,
+                                )
+                            )
+                            tasks.append(checks_warn_task)
+                            task_ref = TaskRef(
+                                dag_name=task.dag_name,
+                                task_id=task.task_name,
+                            )
+                            checks_warn_task.upstream_dependencies.append(task_ref)
```
scholtzan, line 130: We could probably rewrite this a bit to something like:

```suggestion
                        is_check_task_fail = True
                        if "#warn" in file_contents:
                             is_check_task_fail = False
                             
                         checks_task = copy.deepcopy(
                                Task.of_dq_check(
                                    checks_file,
                                    is_check_task_fail,
                                    dag_collection=dag_collection,
                                )
                            )
                            tasks.append(checks_task)
                            task_ref = TaskRef(
                                dag_name=task.dag_name,
                                task_id=task.task_name,
                            )
                            checks_task.upstream_dependencies.append(task_ref)
                           
```

This will always create a task for `fail` even if there are none specified, but I think that's alright for now

PR #4223, opened 2023-08-31
Indent join conditions

sean-rose:
To improve the readability of our queries.

---
Checklist for reviewer:

- [ ] Commits should reference a bug or github issue, if relevant (if a bug is referenced, the pull request should include the bug number in the title).
- [ ] If the PR comes from a fork, trigger integration CI tests by running the [Push to upstream workflow](https://github.com/mozilla/bigquery-etl/actions/workflows/push-to-upstream.yml) and provide the `<username>:<branch>` of the fork as parameter. The parameter will also show up
in the logs of the `manual-trigger-required-for-fork` CI task together with more detailed instructions.
- [ ] If adding a new field to a query, ensure that the schema and dependent downstream schemas have been updated.
- [ ] When adding a new derived dataset, ensure that data is not available already (fully or partially) and recommend extending an existing dataset in favor of creating new ones. Data can be available in the [bigquery-etl repository](https://github.com/mozilla/bigquery-etl), [looker-hub](https://github.com/mozilla/looker-hub) or in [looker-spoke-default](https://github.com/mozilla/looker-spoke-default/tree/e1315853507fc1ac6e78d252d53dc8df5f5f322b).

For modifications to schemas in restricted namespaces (see [`CODEOWNERS`](https://github.com/mozilla/bigquery-etl/blob/main/CODEOWNERS)):
- [ ] Follow the [change control procedure](https://docs.google.com/document/d/1TTJi4ht7NuzX6BPG_KTr6omaZg70cEpxe9xlpfnHj9k/edit#heading=h.ttegrcfy18ck)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1451)

PR #4233, opened 2023-09-01
Handle references to INFORMATION_SCHEMA when deploying to stage

scholtzan:
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1462)


bigquery_etl/cli/stage.py
```diff
@@ -226,8 +229,11 @@ def _view_dependencies(artifact_files, sql_dir):
             for dependency in view.table_references:
                 dependency_components = dependency.split(".")
                 if dependency_components[2:3] == ["INFORMATION_SCHEMA"]:
-                    # INFORMATION_SCHEMA has more components that can be ignored here
-                    dependency_components = dependency_components[:3]
+                    # INFORMATION_SCHEMA has more components that will be treated as the table name
+                    # no deploys for INFORMATION_SCHEMA will happen later on
+                    dependency_components = dependency_components[:2] + [
+                        ".".join(dependency_components[2:])
+                    ]
```
scholtzan, line 236: This will ensure that any references to `INFORMATION_SCHEMA` get set to the `INFORMATION_SCHEMA` in the stage project

bigquery_etl/cli/stage.py
```diff
@@ -139,6 +139,9 @@ def deploy(
         name = artifact_file.parent.name
         test_path = TEST_DIR / project / dataset / name
 
+        if "INFORMATION_SCHEMA" in name:
+            continue
+
```
scholtzan, line 144: We don't want to create any query or schema files for `INFORMATION_SCHEMA`

bigquery_etl/cli/stage.py
sean-rose: This handles `{project}.{region}.INFORMATION_SCHEMA` and `{project}.{dataset}.INFORMATION_SCHEMA` references, but for any `{region}.INFORMATION_SCHEMA` or `{dataset}.INFORMATION_SCHEMA` references then the region/dataset will end up being considered the project and `INFORMATION_SCHEMA` will be considered the dataset.

A possible alternative approach here would be to detect such cases and fill in the project based on the view (before this `if` statement):
```py
                if dependency_components[1:2] == ["INFORMATION_SCHEMA"]:
                    dependency_components.insert(0, view.project)
```
Then `INFORMATION_SCHEMA` should always end up as part of the name, not the dataset, and my other suggested change wouldn't be necessary.  However, I'm not sure if that'd mess with how dependencies are matched.

sean-rose (approved):
r+wc

bigquery_etl/cli/stage.py
scholtzan: Yeah, I guess creating this file is necessary to have references to `some-project.region-us.INFORMATION_SCHEMA.JOBS_BY_PROJECT` updated to `bigquery-etl-integration-test.region-us.INFORMATION_SCHEMA.JOBS_BY_PROJECT`. Not copying them ensures that we don't try to deploy them to the stage environment.

sean-rose: As it is, if the `INFORMATION_SCHEMA` references are updated then it'll be to something commit-hash-specific in `bigquery-etl-integration-test`, which might work for `{project}.{dataset}.INFORMATION_SCHEMA` references (since the commit-specific dataset may get created), but wouldn't work for `{project}.{region}.INFORMATION_SCHEMA` references (because something like `region-us_{hash}` isn't a valid region).

It seems like we should either not update `INFORMATION_SCHEMA` references at all (in which case there might be CI permissions issues depending on the project/dataset component), or have an additional check so that if it's a region qualifier we don't try to append the commit hash.

The current [CI problem](https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/24247/workflows/f57818b1-52a6-4f04-8b1e-a9c1c6a48fcf/jobs/249913) for #4223 is with updating the `INFORMATION_SCHEMA` references causing invalid SQL, not with trying to deploy `INFORMATION_SCHEMA` artifacts to staging.

scholtzan: Ah right...
Not updating the references will indeed create permission problems on deploy. So we'll need to check for the region qualifier

sean-rose: Also, there might be an issue with quoting updated references, where if the `INFORMATION_SCHEMA.{view}` part is quoted separately it won't be valid.  If we don't want to special case that, we could avoid triggering that quoting style by fully quoting the original `INFORMATION_SCHEMA` references (e.g. `monitoring_derived.telemetry_missing_columns_v2` and `monitoring_derived.telemetry_missing_columns_v3` would need to be updated in that case).

scholtzan: I updated the `monitoring_derived` tables and added some logic to prevent the stage deploy from adding the suffix to anything called `INFORMATION_SCHEMA` or `region-us`

bigquery_etl/cli/stage.py
```diff
@@ -286,7 +294,11 @@ def _update_references(artifact_files, project_id, dataset_suffix, sql_dir):
         name_pattern = name.replace("*", r"\*")  # match literal *
         original_dataset = artifact_file.parent.parent.name
         deployed_dataset = original_dataset
-        if dataset_suffix:
+        if (
+            dataset_suffix
+            and "region-us" not in original_dataset
+            and "INFORMATION_SCHEMA" not in original_dataset
+        ):
```
scholtzan, line 301: only append data suffix for datasets that are not `region-us` or `INFORMATION_SCHEMA`

sean-rose: This should cover all current `INFORMATION_SCHEMA` references I see in bigquery-etl, though maybe expanding this to also cover any future EU multi-region references would be good, and since we're looking at the original dataset (not the possibly fudged table/view name) it seems like exact matches would be fine.
```suggestion
            dataset_suffix
            and original_dataset not in ("INFORMATION_SCHEMA", "region-eu", "region-us")
        ):
```

sean-rose (approved):

PR #4239, opened 2023-09-05
Add execution delta to metdata & dag

fbertsch:
Checklist for reviewer:

### DIFF
--- a/dags/bqetl_kpis_shredder.py
+++ b/dags/bqetl_kpis_shredder.py
@@ -45,15 +45,14 @@ with DAG(
 ) as dag:
     fenix_active_users_aggregates_for_deletion_requests = bigquery_etl_query(
         task_id="fenix_active_users_aggregates_for_deletion_requests",
-        destination_table='active_users_aggregates_deletion_request_v1${{ macros.ds_format(macros.ds_add(ds, -1), "%Y-%m-%d", "%Y%m%d") }}',
+        destination_table="active_users_aggregates_deletion_request_v1",
         dataset_id="fenix_derived",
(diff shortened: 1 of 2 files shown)


dags.yaml
```diff
@@ -1154,7 +1154,7 @@ bqetl_kpis_shredder:
     start_date: '2023-05-16'
   description: This DAG calculates KPIs for shredder client_ids
   repo: bigquery-etl
-  schedule_interval: 0 0 */28 * *
+  schedule_interval: 0 2 */28 * *
```
lucia-vargas-a, line 1157: Is this required? My concern with changing the time is that this process runs after shredder, when it's important that it runs before.

lucia-vargas-a (approved):

dags.yaml
fbertsch: Hmm. But if this task needs to wait for `mobile_search_clients_daily_v1` for the day this task runs, that doesn't run until `0 2 * * *`. So even if we set this earlier it will wait.

Given that requirement, we should probably stick this in the same DAG as Shredder and set Shredder downstream.

lucia-vargas-a: Indeed it needs to run later, and it is intentionally separated from shredder to avoid dependencies.
This is a prototype to evaluate the calculation of KPIs for clients' data that will be deleted, I am happy that we set it at 2 and build the rest of the prototype. If shredder deletes the data before this process runs, that will evidence that we do need the dependency. Does that make sense?

fbertsch: Yup, prototypes gonna prototype. We can merge this and see if it works.

PR #4245, opened 2023-09-06
bug(): fixing airflow_task_instance query and correcting job_id type

kik-kik:
# bug(): fixing airflow_task_instance query and correcting job_id type

It appears due to an incorrect aliasing inside the `query.sql` we ended up with an incorrect field being used to populate the `job_id` field. This PR aims to fix this.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1475)

### DIFF
--- a/sql/moz-fx-data-shared-prod/monitoring_derived/airflow_task_instance_v1/schema.yaml
+++ b/sql/moz-fx-data-shared-prod/monitoring_derived/airflow_task_instance_v1/schema.yaml
@@ -34,7 +34,7 @@ fields:
 - description:
   mode: NULLABLE
   name: job_id
-  type: TIMESTAMP
+  type: INTEGER
 - description:
   mode: NULLABLE
   name: executor_config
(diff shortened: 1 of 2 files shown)


sql/moz-fx-data-shared-prod/monitoring_derived/airflow_task_instance_v1/schema.yaml
```diff
@@ -34,7 +34,7 @@ fields:
 - description:
   mode: NULLABLE
   name: job_id
-  type: TIMESTAMP
+  type: INTEGER
```
akkomar, line 37: This change is not backwards compatible, how are we planning to handle that?

kik-kik: I have already recreated the table. This is fine since the table is overwritten by the corresponding ETL anyways.

akkomar (approved):

PR #4272, opened 2023-09-09
Add UDFs to generate regular expressions to correctly match event properties in `events_daily` tables

sean-rose:
Combined with mozilla/lookml-generator#751 this should fix mozilla/lookml-generator#281, which is causing any funnel analyses that filter on secondary properties to be wrong.

---
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1515)


sql/mozfun/event_analysis/event_property_value_to_match_string/udf.sql
```diff
@@ -0,0 +1,17 @@
+CREATE OR REPLACE FUNCTION event_analysis.event_property_value_to_match_string(
+  property_index INTEGER,
+  property_value STRING
+)
+RETURNS STRING AS (
+  CONCAT(
+    event_analysis.escape_metachars(property_value),
+    -- Event property values are stored in `events_daily.events` strings in reverse order,
+    -- so we expect the Nth property value to be followed by N-1 other property values.
+    IF(property_index > 1, CONCAT('[^,]{', (property_index - 1), '}'), '')
```
sean-rose, line 10: This is the crux of the fix for mozilla/lookml-generator#281, as the [existing logic](https://github.com/mozilla/lookml-generator/blob/10d8577c55e4cf607c43811ee4d5d214a66a0d26/generator/views/funnel_analysis_view.py#L166-L175) only matches the first property character immediately preceding the event character.

sql/mozfun/event_analysis/event_property_to_match_string/metadata.yaml
```diff
@@ -0,0 +1,6 @@
+---
+description: >
+  Given an event property index from an `event_types` table, returns a regular expression
+  that can be combined with an event index regular expression to match corresponding events
+  within an `events_daily` table's `events` string.
```
fbertsch, line 5:
```suggestion
  Given an event property index from an `event_types` table, returns a regular expression
  that can be combined with an event index regular expression to match corresponding events
  within an `events_daily` table's `events` string.
  The regex expression will match events if they are not missing the property at `property_index`.
```

fbertsch (approved):
r+wc.

It took my awhile to understand what this UDF is doing. Basically, it's giving us a match string that is true if the property at `property_index` is _not_ missing. Can we call it something different then? Maybe `event_property_index_to_existence_match_string` (or more concisely `property_index_existence_match_string`).

sql/mozfun/event_analysis/event_property_to_match_string/metadata.yaml
sean-rose: > It took my awhile to understand what this UDF is doing. Basically, it's giving us a match string that is true if the property at `property_index` is _not_ missing. Can we call it something different then? Maybe `event_property_index_to_existence_match_string` (or more concisely `property_index_existence_match_string`).

I've renamed it to `event_property_index_to_match_string` (taking a cue from the existing `event_index_to_match_string` UDF), incorporated your clarification in the description, and changed both new `event_property_*_to_match_string` UDFs to generate the entire regular expressions (rather than expecting them to be concatenated with the `event_index_to_match_string` UDF output), which I think makes things slightly less convoluted.

fbertsch: Agreed, thank you for clarifying!

PR #4298, opened 2023-09-14
bug(): Apple timestamps appear to be using PST, converting to UTC

kik-kik:
# bug(): Apple timestamps appear to be using PST, converting to UTC

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1563)


gkabbz (approved):

fbertsch (approved):
r+. Unfortunately since this data does not have hourly buckets, we can't align it with UTC days. Perhaps we have to switch the rest of the funnel to use PT-days?

PR #4300, opened 2023-09-14
Add bqetl check validate command

alekhyamoz:
Add a validate command to make sure that each check in the checks.sql file has either a #fail or #warn marker

Checklist for reviewer:


bigquery_etl/cli/check.py
```diff
@@ -294,4 +294,99 @@ def _run_check(
...
+
+    rendered_result = render_template(
+        checks_file.name,
+        template_folder=str(checks_file.parent),
+        templates_dir="",
+        format=False,
+        **jinja_params,
+    )
+
+    expected_check_types = ["#fail", "#warn"]
```
scholtzan, line 369: We could add them as constants to the top of this file something like:

```python
CHECK_TYPES = ["#fail", "#warn"]
```

Then we can also reference them in other places

bigquery_etl/cli/check.py
```diff
@@ -294,4 +294,99 @@ def _run_check(
...
+        # If no marker is found in the line, set the flag to False
+        if not contains_marker:
+            all_checks_have_markers = False
+            break
+
+    # Check and print the result
+    if all_checks_have_markers:
+        print("All checks have either #fail or #warn markers in the file.")
+    else:
+        print("Not all checks have #fail or #warn markers in the file.")
```
scholtzan, line 392:
```suggestion
        print("Not all checks have #fail or #warn markers in the file.")
        sys.exit(1)
```

It would be useful to also print out which files are affected, to make it easier for people to fix them

alekhyamoz: Correct my understanding, but the validate command is run at the table level and there is only one checks.sql.

scholtzan: We might run `./bqetl checks validate 'telemetry_derived.*'` on multiple tables at the same time. Printing out the file path to the check that failed validation would be helpful

PR #4315, opened 2023-09-18
feat(DENG-1578): docker_fxa_customs_sanitized_v2 query added and view updated

kik-kik:
feat(DENG-1578): docker_fxa_customs_sanitized_v2 query added and view updated

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1586)

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_accounts/docker_fxa_customs_sanitized/view.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts/docker_fxa_customs_sanitized/view.sql
@@ -76,7 +76,7 @@ SELECT
     labels.k8s_pod_env_code,
     labels.k8s_pod_pod_template_hash,
     labels.k8s_pod_app_kubernetes_io_name,
-    labels.k8s_pod_app_kubernetes_io_instance,
+    labels.k8s_pod_app_kubernetes_io_instance
   ) AS labels,
   operation,
   trace,
--- a/sql/moz-fx-data-shared-prod/firefox_accounts/docker_fxa_customs_sanitized/view.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts/docker_fxa_customs_sanitized/view.sql
@@ -4,42 +4,34 @@ AS
 SELECT
   `date`,
   logName,
-  (
-    SELECT AS STRUCT
-      resource.type,
-      (
-        SELECT AS STRUCT
-          resource.labels.instance_id,
-          resource.labels.zone,
-          resource.labels.project_id,
-          CAST(NULL AS STRING) AS cluster_name,
-          CAST(NULL AS STRING) AS namespace_name,
-          CAST(NULL AS STRING) AS container_name,
-          CAST(NULL AS STRING) AS pod_name,
-          CAST(NULL AS STRING) AS location
-      ) AS labels
+  STRUCT(
+    resource.type,
+    STRUCT(
+      resource.labels.instance_id,
+      resource.labels.zone,
+      resource.labels.project_id,
+      CAST(NULL AS STRING) AS cluster_name,
+      CAST(NULL AS STRING) AS namespace_name,
+      CAST(NULL AS STRING) AS container_name,
+      CAST(NULL AS STRING) AS pod_name,
+      CAST(NULL AS STRING) AS location
+    ) AS labels
   ) AS resource,
   textPayload,
-  (
-    SELECT AS STRUCT
-      jsonPayload.block,
-      jsonPayload.op,
-      jsonPayload.action
-  ) AS jsonPayload,
+  STRUCT(jsonPayload.block, jsonPayload.op, jsonPayload.action) AS jsonPayload,
   `timestamp`,
   receiveTimestamp,
   severity,
   insertId,
   httpRequest,
-  (
-    SELECT AS STRUCT
-      labels.compute_googleapis_com_resource_name,
-      CAST(NULL AS STRING) AS k8s_pod_deployment,
-      CAST(NULL AS STRING) AS k8s_pod_app_kubernetes_io_component,
-      CAST(NULL AS STRING) AS k8s_pod_env_code,
-      CAST(NULL AS STRING) AS k8s_pod_pod_template_hash,
-      CAST(NULL AS STRING) AS k8s_pod_app_kubernetes_io_name,
-      CAST(NULL AS STRING) AS k8s_pod_app_kubernetes_io_instance,
+  STRUCT(
+    labels.compute_googleapis_com_resource_name,
+    CAST(NULL AS STRING) AS k8s_pod_deployment,
+    CAST(NULL AS STRING) AS k8s_pod_app_kubernetes_io_component,
+    CAST(NULL AS STRING) AS k8s_pod_env_code,
+    CAST(NULL AS STRING) AS k8s_pod_pod_template_hash,
+    CAST(NULL AS STRING) AS k8s_pod_app_kubernetes_io_name,
+    CAST(NULL AS STRING) AS k8s_pod_app_kubernetes_io_instance
   ) AS labels,
   operation,
   trace,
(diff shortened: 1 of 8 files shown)


sql/moz-fx-data-shared-prod/firefox_accounts/docker_fxa_customs_sanitized/view.sql
```diff
@@ -2,6 +2,105 @@ CREATE OR REPLACE VIEW
...
+          CAST(NULL AS STRING) AS location
+      ) AS labels
+  ) AS resource,
+  textPayload,
+  (
+    SELECT AS STRUCT
+      jsonPayload.block,
+      jsonPayload.op,
+      jsonPayload.action
+  ) AS jsonPayload,
```
sean-rose, line 28: There are many fields in `jsonPayload`. How did you determine that these three are the only ones that should be included?

sql/moz-fx-data-shared-prod/firefox_accounts/docker_fxa_customs_sanitized/view.sql
```diff
@@ -2,6 +2,105 @@ CREATE OR REPLACE VIEW
...
+  (
+    SELECT AS STRUCT
+      labels.compute_googleapis_com_resource_name,
+      CAST(NULL AS STRING) AS k8s_pod_deployment,
+      CAST(NULL AS STRING) AS k8s_pod_app_kubernetes_io_component,
+      CAST(NULL AS STRING) AS k8s_pod_env_code,
+      CAST(NULL AS STRING) AS k8s_pod_pod_template_hash,
+      CAST(NULL AS STRING) AS k8s_pod_app_kubernetes_io_name,
+      CAST(NULL AS STRING) AS k8s_pod_app_kubernetes_io_instance,
+  ) AS labels,
```
sean-rose, line 43: I suspect they don't really matter, but the old table has some additional `labels` fields: `application`, `stack`, `env`, `type`.

sql/moz-fx-data-shared-prod/firefox_accounts/docker_fxa_customs_sanitized/view.sql
```diff
@@ -2,6 +2,105 @@ CREATE OR REPLACE VIEW
...
+  trace,
+  spanId,
+  traceSampled,
+  sourceLocation,
+  split,
 FROM
   `moz-fx-data-shared-prod.firefox_accounts_derived.docker_fxa_customs_sanitized_v1`
+  -- this filter indicates up until when the source table is expected to contains events
+  -- due to AWS to GCP migration, after this date we expect the v2 to contain all the events from GCP.
+  WHERE `date` < "2023-09-15"
```
sean-rose, line 54: Looks like the v1 ETL is still pulling in some data beyond 2023-09-14.  Should that data be included as well?

And should the v1 ETL be descheduled?

sql/moz-fx-data-shared-prod/firefox_accounts/docker_fxa_customs_sanitized/view.sql
```diff
@@ -2,6 +2,105 @@ CREATE OR REPLACE VIEW
...
+      labels.k8s_pod_app_kubernetes_io_instance,
+  ) AS labels,
+  operation,
+  trace,
+  spanId,
+  traceSampled,
+  sourceLocation,
+  split,
+FROM
+  `moz-fx-data-shared-prod.firefox_accounts_derived.docker_fxa_customs_sanitized_v1`
```
sean-rose, line 103:
```suggestion
  `moz-fx-data-shared-prod.firefox_accounts_derived.docker_fxa_customs_sanitized_v2`
```

sql/moz-fx-data-shared-prod/firefox_accounts_derived/docker_fxa_customs_sanitized_v2/metadata.yaml
```diff
@@ -0,0 +1,25 @@
+friendly_name: Docker Fxa Customs Sanitized
+description: |-
+  Sanitized version of customs events from GCP based FxA service stderr table.
+  PII is hashed.
```
sean-rose, line 4: I'd suggest mentioning the date range when the GCP migration started and ended (though given the v1 ETL is still pulling data maybe the end is still a bit questionable).

And I'd also suggest updating the description for the v1 table to clarify it only contains logs from when it was running in AWS, and the date range when the GCP migration started and ended.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/docker_fxa_customs_sanitized_v2/query.sql
```diff
@@ -0,0 +1,20 @@
...
+    (
+      SELECT AS STRUCT
+        jsonPayload.* REPLACE (SHA256(jsonPayload.email) AS email, SHA256(jsonPayload.ip) AS ip)
+    ) AS jsonPayload
+  )
+FROM
+  `moz-fx-fxa-prod.gke_fxa_prod_log.stderr`
+WHERE
+  DATE(_PARTITIONTIME) BETWEEN
+    DATE_SUB(@submission_date, INTERVAL 1 DAY) AND @submission_date
```
sean-rose, line 14: - I'd recommend also including the day after the submission date (if anything, that's more important than the day before, as slightly late arriving data is much more likely than clock skew in the opposite direction).
- I'd suggest putting parentheses around the `(... BETWEEN ... AND ...)` expression so `bqetl format` doesn't format it confusingly to look like the `AND ...` is a normal boolean condition.

sql/moz-fx-data-shared-prod/firefox_accounts_derived/docker_fxa_customs_sanitized_v2/metadata.yaml
```diff
@@ -0,0 +1,25 @@
...
+owners:
+- kik@mozilla.com
+labels:
+  application: fxa
+  incremental: true
+  schedule: daily
+  dag: bqetl_fxa_events
+  owner1: kik
+scheduling:
+  dag_name: bqetl_fxa_events
```
sean-rose, line 14: DAG should be regenerated and checked in.

sql/moz-fx-data-shared-prod/firefox_accounts/docker_fxa_customs_sanitized/view.sql
```diff
@@ -2,6 +2,105 @@ CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.firefox_accounts.docker_fxa_customs_sanitized`
 AS
 SELECT
-  *
+  `date`,
+  logName,
+  (
+    SELECT AS STRUCT
```
sean-rose, line 8: `SELECT AS STRUCT` is very handy when you want to use `* EXCEPT (...)` or `* REPLACE (...)`, but in cases like these where all the fields are being listed explicitly I'd recommend using the `STRUCT(...)` constructor instead, which should be simpler and more efficient.

sql/moz-fx-data-shared-prod/firefox_accounts/docker_fxa_customs_sanitized/view.sql
kik-kik: I asked Wil on Slack. 
image

kik-kik: Also, since this union is at the view level we can always easily add more fields to the view in case more are required.

sql/moz-fx-data-shared-prod/firefox_accounts/docker_fxa_customs_sanitized/view.sql
kik-kik: yes, the ETL should be descheduled.

sean-rose: What about the data that's still showing up in the v1 ETL?  If the migration is actually complete, why is that data showing up?

sql/moz-fx-data-shared-prod/firefox_accounts_derived/docker_fxa_customs_sanitized_v2/query.sql
sean-rose: IMO this is a must-fix before approval.

sql/moz-fx-data-shared-prod/firefox_accounts/docker_fxa_customs_sanitized/view.sql
sean-rose: - You updated some but not all, so now there's an odd mix of the two approaches.
- Unfortunately you can't have a trailing comma after the last `STRUCT()` argument like you can in a `SELECT AS STRUCT` subquery.
- There shouldn't be a space before the opening parenthesis in the `STRUCT()` constructor calls.
- Indention is a bit off within some `STRUCT()` calls.

(re-running `bqetl format` on this file should fix those last two)

sql/moz-fx-data-shared-prod/firefox_accounts_derived/docker_fxa_customs_sanitized_v1/metadata.yaml
```diff
@@ -8,9 +16,11 @@ labels:
   application: fxa
   incremental: true
   schedule: daily
-scheduling:
-  dag_name: bqetl_fxa_events
-  arguments: ['--schema_update_option=ALLOW_FIELD_ADDITION']
+# descheduled as this table was populated using data arriving from the AWS deployment,
+# since the migration there is no need data arriving that needs to be copied.
```
sean-rose, line 20: This wording doesn't quite make sense.  Did you perhaps mean this:
```suggestion
# since the migration there is no new data arriving that needs to be copied.
```

sql/moz-fx-data-shared-prod/firefox_accounts_derived/docker_fxa_customs_sanitized_v1/metadata.yaml
```diff
@@ -1,5 +1,13 @@
...
+  This table contains FxA customs (sanitized) events up until the migration from AWS to GCP was concluded.
+  There is a time window between 2023-09-07 and 2023-09-15 when we were receiving events from both deployments.
+  As the migration as concluded, new data will land and be available via the v2 of this table.
+
+  ---
+
+  Original description
+
   Sanitized version of docker_fxa_customs from FxA.
   PII is hashed.
```
sean-rose, line 12: I'd suggest keeping this concise summary first, and putting the additional details as a secondary paragraph.

sql/moz-fx-data-shared-prod/firefox_accounts/docker_fxa_customs_sanitized/view.sql
kik-kik: hmm, I see that too. I will try and confirm with jbuck and in the interest of time I added back the v1 to the ETL and will create an action item to deschedule it once we're confident no new data is arriving from the AWS based deployment.

sql/moz-fx-data-shared-prod/firefox_accounts/docker_fxa_customs_sanitized/view.sql
```diff
@@ -2,6 +2,91 @@ CREATE OR REPLACE VIEW
...
+  spanId,
+  traceSampled,
+  sourceLocation,
+  split,
 FROM
   `moz-fx-data-shared-prod.firefox_accounts_derived.docker_fxa_customs_sanitized_v1`
+  -- this filter indicates up until when the source table is expected to contains events
+  -- due to AWS to GCP migration, after this date we expect the v2 to contain all the events from GCP.
+WHERE
+  `date` < "2023-09-15"
```
sean-rose, line 47: Maybe this filter should be removed until we know when the actual cutoff is?

sean-rose (approved):
r+wc

To get the `deploy-changes-to-stage` CI job to succeed you'll need to create a `schema.yaml` file for `docker_fxa_customs_sanitized_v1` (CI can't dryrun it to determine its schema due to permissions).

PR #4317, opened 2023-09-18
bug(DENG-1558): firefox_ios_derived.app_store_funnel_v1 updated the source of new profiles

kik-kik:
# bug(DENG-1558): firefox_ios_derived.app_store_funnel_v1 updated the source of new profiles

Previously, the query has been using an incorrect source for getting new_profile numbers. This changes updated the query to use the new more recent table for getting this information.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1592)

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_funnel_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_funnel_v1/query.sql
@@ -54,9 +54,8 @@ _new_profiles AS (
   FROM
     firefox_ios.firefox_ios_clients
   WHERE
-    DATE(submission_timestamp) >= '2022-01-01'
-    AND first_seen_date >= '2022-01-01'
-    -- TODO: do we need to filter here for "release" channel only?
+    first_seen_date >= '2022-01-01'
+    AND channel = "release"
   GROUP BY
     `date`,
     country
--- a/sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_funnel_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_funnel_v1/query.sql
@@ -48,13 +48,13 @@ store_stats AS (
 ),
 _new_profiles AS (
   SELECT
-    first_seen_date AS `date`,
-    first_reported_country AS country,
-    COUNT(*) AS new_profiles,
+    submission_date AS `date`,
+    country,
+    SUM(new_profiles) AS new_profiles,
   FROM
-    firefox_ios.firefox_ios_clients
+    firefox_ios.active_users_aggregates
   WHERE
-    first_seen_date >= '2022-01-01'
+    submission_date >= '2022-01-01'
     AND channel = "release"
   GROUP BY
     `date`,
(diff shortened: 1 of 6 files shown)


sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_funnel_v1/query.sql
```diff
@@ -46,20 +46,17 @@ store_stats AS (
...
-    firefox_ios.new_profile_activation
+    firefox_ios.firefox_ios_clients
   WHERE
-    submission_date >= '2022-01-01'
+    DATE(submission_timestamp) >= '2022-01-01'
     AND first_seen_date >= '2022-01-01'
-     -- below filter required due to untrusted devices anomaly,
-     -- more information can be found in this bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1846554
-    AND NOT (app_display_version = '107.2' AND submission_date >= '2023-02-01')
+    -- TODO: do we need to filter here for "release" channel only?
```
kik-kik, line 59: @bani should we be filtering here for `release` channel only?

bani: yes, the funnel should be release channel only

sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_funnel_v1/query.sql
```diff
@@ -46,20 +46,17 @@ store_stats AS (
...
   SELECT
     first_seen_date AS `date`,
-    country,
-    SUM(new_profile) AS new_profiles,
-    SUM(activated) AS activations,
+    first_reported_country AS country,
+    COUNT(*) AS new_profiles,
   FROM
-    firefox_ios.new_profile_activation
+    firefox_ios.firefox_ios_clients
```
bani, line 55: I think we wanted to use active user aggregates so that we wouldn't be impacted by shredder, righ?

kik-kik: The ticket was specifying that this is the table we should be using and it should give us very close to AUA numbers 🤔 I did check `firefox_ios_clients` against `AUA` and the variance indeed was very small between the two. 30 or less difference between the two tables. Maybe we should have that discussion, right now we have three tables to calculate new_profiles and it's unclear to me why.

soGaussian: This etl seems to be a full table rebuilt rebuild on every run and this will expose us to impacts of shredder. I think we can keep the source but make each run an incremental load of data partition by date. @kik-kik is this on the road map or the etl is going to stay full table rebuild going forward?

kik-kik: yes, it's a small query so it's okay to just rebuild it every time. The potential problem I can see in converting this to an incremental query we run each day is that getting the data from apple is quite slow and would need to further investigate over time time to make sure we would feel confident enough that all the data for each day would be there by the time we run our incremental ETL. Currently, just looking at the results of the `store_stats` CTE it appears we potentially have a 4 day lag. Which might even be good make that more explicit, but the problem is we cannot even be certain that this is a fixed delay and that some days it might be 5 or 3 days delay before the data is available to us. We can look into this more, but I also think for now this is okay. 

Or at least if we use AUA as the source of the new_profile number then we should see pretty stable results.

soGaussian (approved):
Since we are now using active user aggregate table, this code looks good to me.

PR #4322, opened 2023-09-19
DS-3054. Support running an initialization query in parallel

lucia-vargas-a:
Resolves [DS-3054](https://mozilla-hub.atlassian.net/browse/DS-3054) and enables initializing clients_first_seen_v2, see [DS-2948](https://mozilla-hub.atlassian.net/browse/DS-2948).

Checklist for reviewer:

[DS-3054]: https://mozilla-hub.atlassian.net/browse/DS-3054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DS-2948]: https://mozilla-hub.atlassian.net/browse/DS-2948?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1601)


bigquery_etl/cli/query.py
```diff
@@ -1243,6 +1245,73 @@ def validate(
...
+    client,
+    query_file,
+    job_config,
+    project_id,
+    dataset,
+    table,
+    addl_templates,
+    sample_id,
+):
+    """Insert data for o given sample_id into the provided table."""
```
fbertsch, line 1258:
```suggestion
    """Insert data for a given sample_id into the provided table."""
```

bigquery_etl/cli/query.py
```diff
@@ -1243,6 +1245,73 @@ def validate(
...
+                client,
+                query_file,
+                job_config,
+                project_id,
+                dataset_id,
+                table,
+                addl_templates,
+            ),
+            list(range(0, 100)),
+        )
```
fbertsch, line 1311: I believe that we can replace the `_run_initialization_query_for_sample_id` function with `_run_query` here. If we do that, we can use this function for both single-thread and parallelized init. Something like:
```suggestion
        if(parallel_sample_id_run): # Pass this in as a param to `_run_initialization_query_in_parallel` now
          mapped_values = ['-p={"sample_id":"{sample_id}"}' for sample_id in range(0, 100)]
        else:
          mapped_values = ['']
        pool.map(
            partial(
                _run_query,
                query_files=[query_file],
                project_id=project_id,
                public_project_id=None,
                destination_table=None,
                dataset_id=dataset_id,
                addl_templates=addl_templates
            ),
            mapped_values
        )
```

However I would actually prefer to have the parallel init options defined in the `metadata.yaml` rather than in conditional code blocks here...

bigquery_etl/cli/query.py
```diff
@@ -976,8 +981,25 @@ def _run_query(
...
+                # Run the query as shell command so that passed parameters can be used as is.
+                subprocess.check_call(["bq"] + query_arguments, stdin=query_stream)
+            else:
+                query_content = query_stream.read()
+                query = query_content.format(
+                    project_id=project_id,
+                    dataset_id=dataset_id,
+                    table_id=destination_table,
+                    mapped_values=mapped_values,
+                )
```
sean-rose, line 994: IMO it would be better to avoid using Python string formatting syntax in queries.
* Instead of ``INSERT INTO `{project_id}.{dataset_id}.{table_id}` `` you could add `--append-table=true` and `--replace=false` parameters to the `bq query` command when parallelizing the table init. And even if `INSERT INTO {table}` is necessary, that could be added dynamically here as needed rather than requiring that template-ish code be added to the queries themselves.
* Instead of `{mapped_values}` you could configure the query parameter(s) to use in metadata, refer to those in the conditional init SQL like `@sample_id`, and pass them in to the `bq query` command.

sql/moz-fx-data-shared-prod/stripe_external/subscriptions_changelog_v1/query.sql
```diff
@@ -1,3 +1,7 @@
+{% if is_init() %}
+INSERT INTO
+ `{project_id}.{dataset_id}.{table_id}`
+{% endif %}
```
sean-rose, line 4: Per my other comment I don't think this `INSERT INTO` approach is necessary, but if we do go with this approach then IMO the `INSERT INTO {table}` part should either be added dynamically by the tooling at runtime when needed or only added to queries that are actually using parallelized init. Not be added to every query that just happens to use the `is_init()` approach to init.

It's also worth noting this would change the current init behavior of overwriting the table contents and instead always append to the table.

bigquery_etl/cli/query.py
sean-rose: If you do end up still using Python string formatting syntax in queries, then `bqetl format` would need to be changed to not reformat the Python string parameters in the SQL, as it currently reformats parameters like `{foo}` to `{foo }` (with an extra space).

lucia-vargas-a: Thank you. We need the insert but I removed the need for `{project_id}.{dataset_id}.{table_id}` 👍

sql/moz-fx-data-shared-prod/stripe_external/subscriptions_changelog_v1/query.sql
lucia-vargas-a: Indeed, it was work in progress to validate the best approach. I removed the need to modify existing queries and how they're currently updated. For running in parallel the sample_id range can be added to the metadata.

bigquery_etl/cli/query.py
```diff
@@ -1282,36 +1337,64 @@ def initialize(name, sql_dir, project_id, dry_run):
...
         if "is_init()" in sql_content:
             project = query_file.parent.parent.parent.name
             dataset = query_file.parent.parent.name
             destination_table = query_file.parent.name
-            Schema.from_schema_file(query_file.parent / SCHEMA_FILE).deploy(
-                f"{project}.{dataset}.{destination_table}"
-            )
+            full_table_id = f"{project}.{dataset}.{destination_table}"
+
+            try:
```
lucia-vargas-a, line 1350: Prevent initializing existing tables accidentally.

bigquery_etl/cli/query.py
```diff
@@ -1282,36 +1337,64 @@ def initialize(name, sql_dir, project_id, dry_run):
...
+                client.get_table(full_table_id)
+                raise PreconditionFailed(
+                    f"Table {full_table_id} already exists. The initialization process is terminated."
+                )
+            except NotFound:
+                ctx.invoke(deploy, name=full_table_id, force=True)
+
+            query_file_path = Path(query_file)
+            metadata = Metadata.of_query_file(query_file_path)
+            if metadata and metadata.initialization:
```
lucia-vargas-a, line 1360: Retrieve initialization parameters from the table metadata. Currently supported for parallel initializaiton by sample_id.

bigquery_etl/cli/query.py
```diff
@@ -1260,8 +1310,13 @@ def validate(
     "--dry-run/--no-dry-run",
     help="Dry run the initialization",
 )
-def initialize(name, sql_dir, project_id, dry_run):
+@click.pass_context
+def initialize(ctx, name, sql_dir, project_id, dry_run):
     """Create the destination table for the provided query."""
+    client = bigquery.Client()
+    parallelized = False
+    mapped_values = list(range(0, 1))
```
lucia-vargas-a, line 1318: Default value for one iteration.

bigquery_etl/cli/query.py
```diff
@@ -976,8 +981,22 @@ def _run_query(
...
-            # run the query as shell command so that passed parameters can be used as is
-            subprocess.check_call(["bq"] + query_arguments, stdin=query_stream)
+            if mapped_values is None or parallelized is False:
+                # Run the query as shell command so that passed parameters can be used as is.
+                subprocess.check_call(["bq"] + query_arguments, stdin=query_stream)
+            else:
+                query_content = query_stream.read()
+                query = query_content.format(
+                    mapped_values=mapped_values,
+                )
```
fbertsch, line 991: I am hesitant about making any changes to the `_run_query` function; it shouldn't have to know if this is a parallel execution. The mapped values can either be transmitted into the function via query parameters, or jinja templating (but the former will probably be simpler). If we do that we can remove this section and use the subprocess call for all uses of `_run_query`.

fbertsch (commented):
We are almost there :)

bigquery_etl/cli/query.py
lucia-vargas-a: Updated.

bigquery_etl/cli/query.py
lucia-vargas-a: Updated.

bigquery_etl/cli/query.py
lucia-vargas-a: Makes sense. The change in run_query() is limited to determine if the mapped_values should be added to the query arguments.

bigquery_etl/cli/query.py
```diff
@@ -954,6 +956,11 @@ def _run_query(
 
             query_arguments.append("--destination_table={}".format(destination_table))
 
+        if mapped_values is not None and mapped_values != "":
+            query_arguments = query_arguments + [
+                f"--parameter=sample_id:INT64:{mapped_values}"
+            ]
```
fbertsch, line 962: I hate to do this to you, but `query_arguments` is passed to this function. We can do this in the calling function and leave `_run_query` as-is

bigquery_etl/cli/query.py
```diff
@@ -1282,36 +1320,52 @@ def initialize(name, sql_dir, project_id, dry_run):
...
+                raise PreconditionFailed(
+                    f"Table {full_table_id} already exists. The initialization process is terminated."
+                )
+            except NotFound:
+                ctx.invoke(deploy, name=full_table_id, force=True)
+
+            if "@sample_id" in sql_content:
+                mapped_values = list(range(0, 100))
+            if "INSERT INTO" in sql_content:
+                destination_table = None
```
fbertsch, line 1344: Not sure we should formalize this pattern. Is this to deal with scripts, rather than pure queries?

bigquery_etl/cli/query.py
lucia-vargas-a: Interesting, how would that be?
The [mapped_values is the iterable](https://github.com/mozilla/bigquery-etl/blob/50f99cf47b4a390af7cc32d0b721626803b52bcc/bigquery_etl/cli/query.py#L1276), so not known in advance in the calling function. Even calling it as
```
pool.map(
  partial
  (
      _run_query,
      arguments + [f"--parameter=sample_id:INT64:{mapped_values}"] if mapped_values != "" else arguments,```
  )
```
would result in the complete list passed.

bigquery_etl/cli/query.py
lucia-vargas-a: This is for scripots, which return an error when the destination table is passed.

bigquery_etl/cli/query.py
fbertsch: Something like this:
```
pool.map(
            partial(
                _run_query,
                [query_file],
                project,
                None,
                table,
                dataset,
                addl_templates,
            ),
            [f"--parameter=sample_id:INT64:{v}" for v in mapped_values],
        )
```

sean-rose: - The `query_arguments` argument comes before the `addl_templates` argument, so I think `addl_templates` needs to be specified as a keyword argument for this to work.
- The `query_arguments` argument should be a list.
```python
        pool.map(
            partial(
                _run_query,
                [query_file],
                project,
                None,
                table,
                dataset,
                addl_templates=addl_templates,
            ),
            [[f"--parameter=sample_id:INT64:{v}"] for v in mapped_values],
        )
```

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -1,7 +1,7 @@
 -- Query for telemetry_derived.clients_first_seen_v2
-{% if is_init() and parallel_run() %}
+{% if is_init() %}
 INSERT INTO
-    `moz-fx-data-shared-prod.telemetry_derived.clients_first_seen_v2`
+ `moz-fx-data-shared-prod.telemetry_derived.clients_first_seen_v2`
 {% endif %}
```
sean-rose, line 5: I still think you could avoid the need for this `INSERT INTO` kludge by passing `--append_table` and `--noreplace` arguments to the `bq query` command ([some](https://github.com/mozilla/bigquery-etl/blob/79fa5487c3e5470fb80d12b4a30109fdd3b6030a/script/glam/generate_and_run_desktop_sql#L55) [examples](https://github.com/mozilla/bigquery-etl/blob/a5bcebc30878f0f433653fc9f668ac69f88d01a8/sql/moz-fx-data-shared-prod/stripe_external/customers_changelog_v1/metadata.yaml#L19-L21)).

bigquery_etl/cli/query.py
lucia-vargas-a: Removed as it's not required when --append_table is used.

bigquery_etl/cli/query.py
```diff
@@ -1310,29 +1341,43 @@ def initialize(name, sql_dir, project_id, dry_run):
...
-            Schema.from_schema_file(query_file.parent / SCHEMA_FILE).deploy(
-                f"{project}.{dataset}.{destination_table}"
-            )
+            full_table_id = f"{project}.{dataset}.{destination_table}"
+
+            try:
+                client.get_table(full_table_id)
+                raise PreconditionFailed(
+                    f"Table {full_table_id} already exists. The initialization process is terminated."
+                )
```
sean-rose, line 1358: Requiring that the table not exist at all is overly strict IMO (e.g. the table may have already been created via the `bqetl_artifact_deployment.publish_new_tables` task).  I'd suggest allowing an empty table and/or having a `--force` option to override this.
```suggestion
                table = client.get_table(full_table_id)
                if table.num_rows > 0:
                    raise PreconditionFailed(
                        f"Table {full_table_id} already exists and contains data. The initialization process is terminated."
                    )
```
Also, using the BigQuery client's own `PreconditionFailed` exception for this purpose seems a bit odd to me.

bigquery_etl/cli/query.py
```diff
@@ -1310,29 +1341,43 @@ def initialize(name, sql_dir, project_id, dry_run):
...
+            except NotFound:
+                ctx.invoke(deploy, name=full_table_id, force=True)
+
+            if "@sample_id" in sql_content:
+                sample_ids = list(range(1, 100))
```
sean-rose, line 1363: Sample IDs range from 0 to 99.
```suggestion
                sample_ids = list(range(0, 100))
```

bigquery_etl/cli/query.py
```diff
@@ -1310,29 +1341,43 @@ def initialize(name, sql_dir, project_id, dry_run):
...
+                arguments += ["--dry_run"]
+
+            _initialize_in_parallel(
+                project=project,
+                table=destination_table,
+                dataset=dataset,
+                query_file=query_file,
+                arguments=arguments,
+                parallelism=DEFAULT_PARALLELISM,
+                sample_ids=sample_ids,
```
sean-rose, line 1380: This is going to fail for non-parallelized cases because an empty value for an integer parameter results in an "Unparseable query parameter" error.  You could change the default sample ID value from `""` to `"NULL"`, but IMO it'd be better to simply call `_run_query()` as before for the non-parallelized case.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/metadata.yaml
```diff
@@ -31,6 +34,7 @@ scheduling:
   date_partition_parameter: null
   parameters:
   - submission_date:DATE:{{ds}}
+  - sample_id:INT64:{{sample_id}}
```
sean-rose, line 37: `sample_id` won't be a parameter during normal runs, and so shouldn't be specified here.

bigquery_etl/cli/query.py
lucia-vargas-a: Decided to separate the parallelized an non-parallelized cases for reduced complexity.

sean-rose (approved):

PR #4327, opened 2023-09-19
FXA-6721 Setup import of accounts table from FxA stage CloudSQL

akkomar:
This adds a federated query to import `accounts` table from FxA database from stage environment.

This imports the data to `accounts_backend_external` and assumes that derived tables will be created in `accounts_backend_derived`, similar to the approach taken with `mozilla_vpn_*`.
Important detail is that FxA backend telemetry will land in `accounts_backend_stable`, so eventually in `accounts_backend_derived` we'll have tables derived from both server telemetry and database dumps. I think this is fine since both come from the same application but we should get an agreement on that, if not now then at least before the production imports start.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1607)


sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/query.sql
```diff
@@ -0,0 +1,22 @@
+SELECT
+  *
+FROM
+  EXTERNAL_QUERY(
+    "moz-fx-fxa-nonprod.us.fxa-rds-nonprod-stage-fxa",
+    """SELECT
```
akkomar, line 6: This query cannot be made incremental because this table does not have a "last modified" column.
I'm going to evaluate Datastream for incremental imports in parallel to this work in case full dumps turn out to be problematic.

sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/metadata.yaml
kik-kik, line 1: Just a more general question, are we agreeing here that the convention here is that we should be using `_external` dataset/namespace for these tables to keep some kind of logical separation between the other derived tables? I remember this was a topic in the past and it seemed like we wanted to avoid that?

sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/query.sql
akkomar: Snapshot is fine for this table, we can track history in a derived table.

sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/metadata.yaml
akkomar: Per the [recommendation](https://docs.google.com/document/d/1tXKq37jo13-EdPmwYOxvtS8mrbNb8rhZWKGKTiAyJBs/edit):
> reserve _external usage for importing data from bq-external workgroup-confidential systems

I expect that we might need to restrict some tables imported here later so this seems like a right place to me. @whd wdyt?

kik-kik (approved):
In general looks good and has no impact on any downstream processes. So we're good to try it out. We should just agree on a convention on how we want to handle and where to store data coming from OLTP systems.

sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/metadata.yaml
whd: Belatedly, I wrote up https://docs.google.com/document/d/1tXKq37jo13-EdPmwYOxvtS8mrbNb8rhZWKGKTiAyJBs/edit#heading=h.v9ip0n681xop last time we discussed this and I think `_external` is appropriate in this case. However, the ACL is nonstandard so that requires additional configuration in cloudops-infra to deploy.

sql/moz-fx-data-shared-prod/accounts_backend_external/dataset_metadata.yaml
```diff
@@ -0,0 +1,11 @@
+friendly_name: Firefox Accounts Backend External
+description: |-
+  Data extracted from the FxA backend services databases.
+  See https://mozilla.github.io/ecosystem-platform/reference/database-structure#database-fxa for more information.
+dataset_base_acl: restricted
+user_facing: false
+labels: {}
+workgroup_access:
```
whd, line 8: This dataset configuration is [nonstandard](Data+Access+Workgroups#DataAccessWorkgroups-Exampledatasetwithuniformspecificworkgroupaccess(creatingadatasetthatisaccessibletoaspecificworkgroup)) in that it's granting broad DE access to a dataset that is `restricted` (workgroup-confidential). There's some discussion of this in https://mozilla-hub.atlassian.net/browse/DSRE-1023, and an example of how we've approached that elsewhere in https://mozilla-hub.atlassian.net/browse/DSRE-1023?focusedCommentId=771018.

At any rate, because this configuration is nonstandard it requires matching configuration in https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/tf/prod/envs/prod/bigquery-new/namespaces.auto.tfvars.json. If you're sure it's acceptable for DE to have this access (looks like that's how we have some other `_external` datasets set up) I can update ops logic to include an entry to allow this to deploy.

sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/metadata.yaml
```diff
@@ -0,0 +1,17 @@
+---
+friendly_name: FxA Accounts nonprod
+description: >
+  A mirror of the `accounts` table from the staging FxA CloudSQL database,
+  updated daily to match the current state of the table.
+owners:
+  - akomar@mozilla.com
+labels:
+  application: accounts_backend
+  schedule: daily
```
sean-rose, line 10: I've sometimes specified `incremental: false` under `labels:` for non-incremental ETLs. Labels with `false` values are ignored when deploying the table, so it isn't visible in BigQuery itself, but I figure maybe it's a helpful indication for us developers. 🤷

sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/query.sql
sean-rose: It's my understanding that FxA hard deletes an account's database records when the user clicks "Delete account" in their profile settings, which will be something to keep in mind for downstream uses of this data or if we adopt any sort of incremental approach.

Currently we're taking Firefox Account deletion events from the logs as the cue to shred related FxA data in BigQuery.  Since FxA hard deletes records and we're overwriting this table every time it's effectively self-shredding, but if we change it to be incremental then we'd need to configure it for shredding.

sean-rose (commented):
Apologies for the belated post-merge review.

PR #4344, opened 2023-09-25
feat(DENG-1233): Queries to generate firefox_ios retention metrics

kik-kik:
# feat(DENG-1233): Queries to generate firefox_ios retention metrics

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1640)


sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_retention_week_4_v1/query.sql
```diff
@@ -0,0 +1,38 @@
...
+)
+SELECT
+  first_seen.first_seen_date,
+  clients_retention.client_id,
+  clients_retention.sample_id,
+  mozfun.bits28.active_in_range(days_seen_bits, -6, 7) AS retained_week_4,
+  LENGTH(REPLACE(mozfun.bits28.to_string(days_seen_bits), "0", "")) AS days_seen_in_first_28_days,
+  -- TODO: Should we be using the UDF here?
+  -- mozfun.bits28.retention(days_seen_bits, @submission_date).day_27.active_in_week_3 AS retention_week_4,
+  -- mozfun.bits28.retention(days_seen_bits, @submission_date) AS retention, -- retention week 2 query should return the same result as day_13.active_in_week_1? Maybe a good test to make sure we're aligned with the UDF?
```
bani, line 32: yes, please test `day_13.active_in_week_1` versus our retention_week_2 calculation

sql/moz-fx-data-shared-prod/firefox_ios/app_store_retention/view.sql
```diff
@@ -0,0 +1,16 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.firefox_ios.app_store_retention`
+AS
+SELECT
+  first_seen_date,
+  client_id,
+  sample_id,
+  retained_week_2,
+  retained_week_4,
+  days_seen_in_first_28_days,
```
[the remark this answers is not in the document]
kik-kik: repeat user meaning more than 1 day within the 28 days @soGaussian?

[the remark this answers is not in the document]
kik-kik: Added the flag field to the week 4 retention query.

sql/moz-fx-data-shared-prod/firefox_ios/app_store_retention/view.sql
gkabbz, line 1: @soGaussian and @kik-kik Not sure about the naming of this view and underlying tables. It implies that this is retention from the app store. But this retention is being calculated from telemetry data. Also... I'm assuming this will be the weekly retention view, would it make sense to include weekly_retention in the name?

kik-kik: yeah, I asked for some ideas for what this dataset should be named on Slack. I couldn't think of anything clever myself 😬

soGaussian: I propose iOS_funnel_retention

soGaussian (approved):

sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_retention_week_2_v1/query.sql
```diff
@@ -0,0 +1,48 @@
...
+    submission_date = @submission_date
+    AND normalized_channel = "release"
+),
+first_seen AS (
+  SELECT
+    client_id,
+    sample_id,
+    first_seen_date,
+  FROM
+    firefox_ios.baseline_clients_first_seen
```
gkabbz, line 19: @kik-kik and @soGaussian we spent a bunch of time in the last sprints working on the `firefox_ios_clients` and `firefox_android_clients` tables and figuring out how we can calculate new profiles from it. We aligned on the new profile calculation and did work to validate if it matches up to `active_users_aggregates` as the way we calculate new profiles. From this it looks like we are choosing not to use the `ios_clients` table. Is there a reason why it won't fit our needs to calculate the cohort? Think through some of the effects this might end up having e.g. what if the cohort calc here starts to deviate from the new profiles calc? ios clients has attribution - do we want to implement attribution in two different places?

kik-kik: So the main thing I'd be careful about when using `ios_clients` table as the base table is this:

https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/firefox_ios_derived/firefox_ios_clients_v1/query.sql#L167-L194

The problem is that all those attributes we set for a client in the clients table is done over time giving us a 2 week window before this query would run against a user for the first time. I would assume this should not be a problem at this time frame, but lets say we'd want to add another attribution metric for week 1 or even day 1 it would potentially mean that at the time when this query would run some of the user attributes could be missing.

It's also not the end of the world since we could assume that the later funnel calculations are more up-to-date and complete (lets say the week_2 retention table is not expected to see this behaviour). This is something we could resolve in the view combining all the data.

sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_retention_week_2_v1/query.sql
```diff
@@ -0,0 +1,48 @@
...
+  -- Two weeks need to elapse before calculating the week 2 retention
+  WHERE
+    submission_date = DATE_SUB(@submission_date, INTERVAL 13 DAY)
```
fbertsch, line 22: Strong opinion incoming: `first_seen` tables should not be filtered on `submission_date`. The grain of this table is `client_id`, and filtering on `submission_date` introduces possible error without much benefit (the table is very small, and joinable even at the scale of Firefox desktop).

The `submission_date` fields in any first_seen tables make them much more confusing to understand and use.

sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_retention_week_2_v1/query.sql
```diff
@@ -0,0 +1,48 @@
...
+SELECT
+  * EXCEPT (retention) REPLACE(
+  -- metric date should align with first_seen_date, if that is not the case then the query will fail.
+    IF(
+      retention.day_13.metric_date <> first_seen_date,
+      ERROR("Metric date misaligned with first_seen_date"),
+      first_seen_date
+    ) AS first_seen_date
+  ),
+  retention.day_13.active_in_week_1 AS retained_week_2,
```
fbertsch, line 46: This is correct. This is saying: "For users who started 13 days ago, which were active in days 7-13 (0-indexed)?"

sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_retention_week_4_v1/query.sql
```diff
@@ -0,0 +1,52 @@
...
+    -- metric date should align with first_seen_date, if that is not the case then the query will fail.
+    IF(
+      retention.day_27.metric_date <> first_seen_date,
+      ERROR("Metric date misaligned with first_seen_date"),
+      first_seen_date
+    ) AS first_seen_date
+  ),
+  days_seen_in_first_28_days > 1 AS repeat_first_month_user,
+  -- added for testing.
+  retention.day_27.active_in_week_1 AS retained_week_2,
```
fbertsch, line 49: This one is not correct. What we need here is: "For users who started 27 days ago, which were active in days 7-27 (0-indexed)"? Otherwise, we run the risk of a non-consecutive funnel: week-4 retained users who were not week-2 retained.

Note that with this change, the value will be different than the week 2 retention in the previous table.

sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_retention_week_4_v1/query.sql
```diff
@@ -0,0 +1,52 @@
...
+  retention.day_27.active_in_week_3 AS retained_week_4,
```
fbertsch, line 50: This is correct. This is saying "For users who started 27 days ago, which were active on days 21-27 (0-indexed)?"

fbertsch (commented):
Aside from the 2-week retention metric change, my recommendation is to _not_ use the `bits28.retention` fields, mainly because those are 0-indexed and we are now 1-indexing (which causes confusion). Using the direct bit fields from the retention UDF (e.g. `bits28.active_in_range(bits, -6, 7) AS retained_week_4`) may be more obvious to readers of this query.

sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_retention_week_4_v1/query.sql
fbertsch: cc @bani @soGaussian if we want to include other funnel metrics, we need to outline them in the decision brief

sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_retention_week_2_v1/query.sql
kik-kik: so in general I agree here, unfortunately, due to how the underlying tables are set up this filter is required. Otherwise, we see the following error (when removing this filter):
```
Cannot query over table 'moz-fx-data-shared-prod.org_mozilla_ios_firefox_derived.baseline_clients_daily_v1' without a filter over column(s) 'submission_date' that can be used for partition elimination 
```

sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_retention_week_4_v1/query.sql
soGaussian: @fbertsch not sure I follow your first comment, are you referring to repeat first month user metric or retained week 2 metric. If it is the repeat first month user metric, then the logic is correct, the way we defined it in the brief was any new profile that returned again with the first 27 days after their first day. The retained week 2 metric is not part of the funnel, it is a metric @kik-kik  add because the UDF calculated that along with the other retention metrics.

PR #4350, opened 2023-09-27
tweak(): submission_date hardcoded to avoid having to specify submission_date when using the first_seen_view

kik-kik:
tweak(): submission_date hardcoded to avoid having to specify submission_date when using the first_seen_view

This simplifies the downstream usage of the per product first_seen view which requires submission_date filter making it more tricky to filter for a specific first_seen_date value.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1651)


scholtzan (approved):

kik-kik:
There seem to be some disagreement on whether or not this should be added. For now closing this, if needed we can always open a new PR for this change.

PR #4354, opened 2023-09-27
Prioritize ping type and first record for ping types with the same TIMESTAMP

lucia-vargas-a:
[DS-3186](https://mozilla-hub.atlassian.net/browse/DS-3186) After initializing `clients_first_seen_v2` for a subset of sample_ids in parallel, some client_ids return more than one record due to the same timestamp among pings.

This PR prioritizes ping types and only first record when this happens.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1658)

[DS-3186]: https://mozilla-hub.atlassian.net/browse/DS-3186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
@@ -3,7 +3,6 @@
 INSERT INTO
     `moz-fx-data-shared-prod.telemetry_derived.clients_first_seen_v2`
 {% endif %}
--- Query for telemetry_derived.clients_first_seen_v2
 -- Each ping type subquery retrieves all attributes as reported on the first
 -- ping received and respecting NULLS.
 -- Once the first_seen_date is identified after comparing all pings, attributes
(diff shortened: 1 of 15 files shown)


sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -273,32 +278,71 @@ reported_pings AS (
     client_id
 ),
 _current AS (
+  -- Get first value when the same ping type returns more than one record with the exact same TIMESTAMP
```
lucia-vargas-a, line 281: At this point the value is expected to be unique, however is safer to consider first in case the same ping type was sent more than once. Thoughts?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -307,14 +351,16 @@ _previous AS (
...
 SELECT
-  {% if is_init() %}
-    IF(_previous.client_id IS NULL, _current, _previous).*
-  {% else %}
-    -- For the daily update:
-    -- The reported ping status in the metadata is updated when it's NULL.
-    -- The second_seen_date is updated when it's NULL and only if there is a
-    -- main ping reported on the submission_date.
-    -- Every other attribute remains as reported on the first_seen_date.
+{% if is_init() %}
```
lucia-vargas-a, line 354: For initialization there is no need to compare with the existing records. Is it possible to have a client_id with different sample_id that would create a duplicate due to inserting directly to the table and in parallel?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
fbertsch: I'm of two minds here. I prefer our ETLs to be completely idempotent. But this adds a good bit of complexity for a scenario that is quite possible to never occur. If we leave this part out, what happens when we receive two first pings of the same type from the same timestamp for a client?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
fbertsch: No, `(client_id, sample_id)` pairs should be stable

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
lucia-vargas-a: In that case the subquery of that ping type returns one record with the values depending on the order in which data is returned by that subquery. So two records for the same (client_id1, sample_id1, timestamp1) and different country would result in (client_id1, sample_id1, timestamp1, first_country_depending_on_the_order).

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
lucia-vargas-a: Thanks. Probably worth a data check, for when we implement them for this table.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
fbertsch: I think that is acceptable in this case, given how rarely this should occur. Let's verify this with the team during the infra meeting tomorrow (thesis being: we should _almost_ never receive the same ping with the same timestamp from a client, though I suppose we can check this with a query).

fbertsch (approved):

PR #4357, opened 2023-09-28
chore(): reformetted all sql which gets picked up by the bqetl format command

kik-kik:
chore(): reformetted all sql which gets picked up by the bqetl format command

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1669)


sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -14,48 +14,134 @@ WITH new_profile_ping AS (
...
+        submission_timestamp
+    )[SAFE_OFFSET(0)] AS attribution_dlsource,
   FROM
     `moz-fx-data-shared-prod.telemetry.new_profile`
   WHERE
     {% if is_init() %}
       DATE(submission_timestamp) >= '2010-01-01'
       {% if parallel_run() %}
-        AND sample_id = {sample_id}
+        AND sample_id = {sample_id }
```
sean-rose, line 144: This is weird.  Should this be `{{sample_id}}` or `@sample_id` instead?  Though I don't see how any of `{sample_id}`, `{{sample_id}}`, or `@sample_id` would work as currently written, and I also don't see any `parallel_run()` macro available in this context.

@lucia-vargas-a it looks you added this recently in #3962.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -70,48 +156,134 @@ shutdown_ping AS (
...
+        submission_timestamp
+    )[SAFE_OFFSET(0)] AS attribution_dlsource,
   FROM
     `moz-fx-data-shared-prod.telemetry.first_shutdown`
   WHERE
     {% if is_init() %}
       DATE(submission_timestamp) >= '2010-01-01'
       {% if parallel_run() %}
-        AND sample_id = {sample_id}
+        AND sample_id = {sample_id }
```
sean-rose, line 286: ditto

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -135,48 +307,102 @@ main_ping AS (
...
+      NULL AS STRING
+    ) AS attribution_dlsource -- main_v4:environment.settings.attribution.dlsource
   FROM
     `moz-fx-data-shared-prod.telemetry_derived.clients_daily_v6`
   WHERE
     {% if is_init() %}
       submission_date >= '2010-01-01'
       {% if parallel_run() %}
-        AND sample_id = {sample_id}
+        AND sample_id = {sample_id }
```
sean-rose, line 405: ditto

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
lucia-vargas-a: Hey, the formatting is correct! This code runs as part of improvements to the initialization command currently in review in [PR-4322](https://github.com/mozilla/bigquery-etl/pull/4322).

lucia-vargas-a, line 144:
```suggestion
        AND sample_id = {sample_id}
```

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
lucia-vargas-a, line 405:
```suggestion
        AND sample_id = {sample_id}
```

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -1,6 +1,6 @@
 -- Query for telemetry_derived.clients_first_seen_v2
 {% if is_init() and parallel_run() %}
-INSERT INTO
+  INSERT INTO
```
lucia-vargas-a, line 3:
```suggestion
INSERT INTO
```

lucia-vargas-a (changes requested):
Suggested changes required for the jinja templates.

lucia-vargas-a (commented):
It requires updating with the main branch because some jinja templates have changed.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
sean-rose: If we keep this Python string formatting parameter syntax then `bqetl format` would need to be changed to not break it (it's currently automatically adding the extra space.

However, IMO it would be better to avoid this syntax:
* `{sample_id}` (or `{mapped_values}` as #4322 currently has it) would be better as a query parameter like `@sample_id` or `@mapped_values`.  This is safer, and avoids having to manually SQL-quote string values going in.
* Instead of ``INSERT INTO `{project_id}.{dataset_id}.{table_id}` `` you could add `--append-table=true` and `--replace=false` parameters to the `bq query` command when parallelizing the table init, which will have the same effect without requiring that kludge.

(I'll also add PR comments to that effect in #4322)

PR #4362, opened 2023-09-29
feat(): firefox_ios_clients_v1 table partitoning and clustering settings update + filtering out "suspicious ios" users

kik-kik:
# feat(): firefox_ios_clients_v1 table partitoning and clustering settings update + filtering out "suspicious ios" users 

filtering out "suspicious ios" users due to bug-1846554

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1674)

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_ios_derived/firefox_ios_clients_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_ios_derived/firefox_ios_clients_v1/query.sql
@@ -11,14 +11,12 @@ WITH first_seen AS (
     device_manufacturer,
     device_model,
     normalized_os_version AS os_version,
-    app_display_version AS app_version
+    app_display_version AS app_version,
   FROM
     firefox_ios.baseline_clients_first_seen
   WHERE
     submission_date = @submission_date
     AND client_id IS NOT NULL
-    -- filtering out suspicious devices on iOS, for more info see: bug-1846554
-    AND NOT (app_display_version = '107.2' AND submission_date >= '2023-02-01')
 ),
 -- Find the most recent activation record per client_id.
 activations AS (
@@ -143,7 +141,9 @@ _current AS (
           THEN "metrics"
         ELSE NULL
       END AS adjust_info__source_ping
-    ) AS metadata
+    ) AS metadata,
+    -- field to help us identify suspicious devices on iOS, for more info see: bug-1846554
+    (app_version = '107.2' AND submission_date >= '2023-02-01') AS is_suspicious_device_client,
   FROM
     first_seen
   FULL OUTER JOIN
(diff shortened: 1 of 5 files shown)


sql/moz-fx-data-shared-prod/firefox_ios_derived/firefox_ios_clients_v1/query.sql
```diff
@@ -17,6 +17,8 @@ WITH first_seen AS (
   WHERE
     submission_date = @submission_date
     AND client_id IS NOT NULL
+    -- filtering out suspicious devices on iOS, for more info see: bug-1846554
+    AND NOT (app_display_version = '107.2' AND submission_date >= '2023-02-01')
```
lucia-vargas-a, line 21: - Shall we filter this in the view so that it's still available for debugging?
- Also, since the traffic went back to normal on May 5, does it make sense to limit the dates instead of excluding them indefinitely?

soGaussian: This was my question in the channel but @bani's Answer made sense to me. Adding her answer in the thread below.
"it's random people in the world doing weird stuff that is not going through the app store in a large volume (see this [thread](https://mozilla.slack.com/archives/C041PKJRA7N/p1693338793787679?thread_ts=1680284882.811289&cid=C041PKJRA7N)).
At some point in August when we were working on the funnel I considered time bounding because the original incident was over, and then I noticed a second spike with the same characteristics.
Since this is an app store funnel, a better filter would be anything that isn't the current version (+/- 1 because the release is gradual), but since we may not have high quality data, this filter at least cleans up some garbage that was preventing the iOS funnel from looking like a funnel when we were doing top -> bottom funnel."

kik-kik: Ok, in the end doing something similar to what you suggested @lucia-vargas-a with a small variation. Please let me know if you think this is a reasonable approach.

lucia-vargas-a: Adding a flag to identify them makes sense, data is still available and we can add other scenarios if needed 👍

soGaussian (approved):

lucia-vargas-a (approved):
LGTM, only one update to the new column recommended.

sql/moz-fx-data-shared-prod/firefox_ios_derived/firefox_ios_clients_v1/query.sql
```diff
@@ -208,6 +210,7 @@ SELECT
       _current.metadata.adjust_info__source_ping
     ) AS adjust_info__source_ping
   ) AS metadata,
+  COALESCE(_previous.is_suspicious_device_client, _current.is_suspicious_device_client) AS is_suspicious_device_client,
```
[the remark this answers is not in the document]
kik-kik: This should never happen, once the value is decided it should never change and every existing entry should be either True or False.

PR #4367, opened 2023-09-29
Put assert UDFs in `mozfun` project

sean-rose:
So the assert UDFs can be used wherever people want, not just in the automated UDF tests.

---
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1680)


relud (approved):

sean-rose:
@relud it looks like SQLGlot [can't handle](https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/25203/workflows/d95f27aa-e52b-45f2-8765-9ce57bef1b50/jobs/262356) the syntax in `assert.all_fields_null()` where [`.*` appears after an array subscript operation](](https://github.com/mozilla/bigquery-etl/blob/b7024c6812a1720d5653c014cfb83104938bc409/sql/mozfun/assert/all_fields_null/udf.sql#L3)):
```sql
    TO_JSON_STRING((SELECT AS STRUCT [actual][SAFE_OFFSET(-1)].*)) = TO_JSON_STRING(actual),
```

I know you worked on the bqetl SQLGlot integration.  Any advice on how to address this situation?

relud:
> @relud it looks like SQLGlot [can't handle](https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/25203/workflows/d95f27aa-e52b-45f2-8765-9ce57bef1b50/jobs/262356) the syntax in `assert.all_fields_null()` where [`.*` appears after an array subscript operation](https://github.com/mozilla/bigquery-etl/blob/b7024c6812a1720d5653c014cfb83104938bc409/sql/mozfun/assert/all_fields_null/udf.sql#L3):
> 
> ```sql
>     TO_JSON_STRING((SELECT AS STRUCT [actual][SAFE_OFFSET(-1)].*)) = TO_JSON_STRING(actual),
> ```
> 
> I know you worked on the bqetl SQLGlot integration. Any advice on how to address this situation?

file an upstream bug, and remove the `.*` with a regex, kind of like we do here: https://github.com/mozilla/bigquery-etl/blob/a1092cd0b574e3d484902d2e2b5091a0c67913d6/bigquery_etl/dependency.py#L36-L43

the functionality of the code sqlglot is parsing doesn't need to be preserved, so long as it doesn't impact identifying what tables a query references.

sean-rose:
Turns out the SQLGlot parsing issue affecting `assert.all_fields_null()` was fixed in a newer version.

sean-rose:
> @relud 
> file an upstream bug, and remove the `.*` with a regex, kind of like we do here

SQLGlot currently has [no open issues](https://github.com/tobymao/sqlglot/issues?q=is%3Aopen+is%3Aissue), so assuming they correctly fixed everything you reported, we could theoretically remove all the existing workarounds for SQLGlot issues.

bigquery_etl/routine/parse_routine.py
sean-rose, line 1: @scholtzan I'm requesting your review because I ended up refactoring this `parse_routine` module pretty heavily, and from the history it looks like you've been the person most involved with its development recently.

bigquery_etl/cli/stage.py
```diff
@@ -203,15 +203,12 @@ def _udf_dependencies(artifact_files):
     for udf_file in udf_files:
         # all referenced UDFs need to be deployed in the same stage project due to access restrictions
         raw_routine = RawRoutine.from_file(udf_file)
-        udfs_to_publish = accumulate_dependencies([], raw_routines, raw_routine.name)
-
-        for dependency in udfs_to_publish:
-            dataset, name = dependency.split(".")
-            file_path = (
-                raw_routine.filepath.parent.parent.parent / dataset / name / UDF_FILE
-            )
```
sean-rose, line 212: > @scholtzan any ideas on [this `deploy-changes-to-stage` error](https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/25207/workflows/441baa1c-236f-42f8-a139-45ed34754d2e/jobs/262393)?
> 
> Seems like it's expecting some UDFs from `mozfun` to be in `moz-fx-data-shared-prod` instead for some reason.

I believe this was the cause of the `deploy-changes-to-stage` error (assuming the dependent UDFs were in the same project directory as the UDF calling them).

In hindsight I think I could have simply replaced this with logic using `raw_routines[dependency].filepath` without refactoring the `parse_routine` module so heavily to use fully qualified routine IDs, so if you think that refactoring is a bridge too far then I can try reverting it and just making this change instead.

bigquery_etl/cli/stage.py
```diff
@@ -203,15 +203,12 @@ def _udf_dependencies(artifact_files):
...
-            if file_path not in artifact_files:
-                artifact_files.add(file_path)
```
sean-rose, line 214: This seems like it was a bug, modifying the passed in `artifact_files` set rather than adding to the `udf_dependencies` set that gets returned.

bigquery_etl/routine/parse_routine.py
```diff
@@ -257,68 +240,57 @@ def parse_routines(project_dir):
...
+    raw_routine = raw_routines[routine_id]
     for dep in raw_routine.dependencies:
         deps = accumulate_dependencies(deps, raw_routines, dep)
-    if udf_name in deps:
-        return deps
-    else:
-        return deps + [udf_name]
+        if dep not in deps:
+            deps = deps + [dep]
+    return deps
```
sean-rose, line 266: This change means that the initially passed in routine will no longer be included in the list of its own dependencies (which was a bug IMO).

bigquery_etl/routine/parse_routine.py
```diff
@@ -257,68 +240,57 @@ def parse_routines(project_dir):
...
+        routine_usage_re = re.compile(
+            r"(?<!\.)\b"
+            rf"(`?{routine['project']}`?\.){'?' if routine['project'] == project else ''}"
+            rf"`?{routine['dataset']}`?\.`?{routine['name']}`?\("
+        )
+        if routine_usage_re.search(sql):
+            routine_usages.append(routine["id"])
 
-    return sorted(set(udf_usages))
+    routine_usages.extend(TEMP_UDF_USAGE_RE.findall(sql))
```
sean-rose, line 286: I'm unclear why including temp UDFs is necessary/useful here, as it seems like everywhere `routine_usages_in_text()` is being used we only care about the persistent UDFs/procedures and have to take extra steps to filter out the temp UDFs (e.g. to prevent invalid key lookups in raw routines dictionaries).

bigquery_etl/routine/parse_routine.py
```diff
@@ -327,28 +299,52 @@ def sub_local_routines(test, project, raw_routines=None, stored_procedure_test=F
...
+
+    for routine in routines:
+        for definition in routine.definitions:
+            routine_match = PERSISTENT_UDF_RE.match(definition)
+            routine_project = routine_match.group(1) or routine.project
+            routine_dataset = routine_match.group(2)
+            routine_name = routine_match.group(3)
+            replace_routine_name = re.sub(
+                r"\W", "_", f"{routine_project}_{routine_dataset}_{routine_name}"
+            )
```
sean-rose, line 333: I was hoping this regex logic could be avoided and simply use `replace_routine_name = re.sub(r"\W", "_", routine.id)`, but it turns out there's one UDF file that doesn't follow the one-UDF-per-file convention and defines multiple UDFs: [sql/moz-fx-data-shared-prod/udf/safe_crc32_uuid/udf.sql](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/udf/safe_crc32_uuid/udf.sql)

bigquery_etl/cli/stage.py
sean-rose: Never mind, I've gone ahead and reverted the refactoring of `parse_routine`.

A sticking point turned out to be that when ETLs are deployed to stage they're moved into `sql/bigquery-etl-integration-test/` and we lose any notion of which project they were in originally, so for any routine calls in the ETL that aren't fully qualified we can't tell for certain what their fully qualified routine IDs would be.

One possible solution for that would be to also deploy ETLs' UDF dependencies to stage like is done for views, then the partially qualified UDF calls would be resolvable to the staged UDFs, but that's really only necessary for ETLs with tests and I felt trying to do that was going to be too much for this PR, which was already overreaching its scope quite a bit.

I've reverted the refactoring I attempted rather than rebasing/erasing them just in case it might be useful for reference in some other future refactoring attempt.

scholtzan (approved):

PR #4370, opened 2023-10-02
fix(): tweaks to partitioning settings for firefox_ios_derived.funnel_retention tables

kik-kik:
fix(): tweaks to partitioning settings for firefox_ios_derived.funnel_retention tables

This is to correctly work with the bqetl mechanism when inserting data into a specific data partition of a table:
https://github.com/mozilla/telemetry-airflow/blob/2048f039c6c68bf99254cb9721bc35923232cd82/dags/utils/gcp.py#L197

Otherwise, when running the query we get an error trying to insert the data into a "wrong partition"


scholtzan (approved):
FYI this will require redeploying the tables with the new partitioning and copying over the existing data

kik-kik:
Already done. Thanks!

PR #4372, opened 2023-10-02
DENG-476 - Create derived table for HCM clients dashboards

relud:
so that they no longer need to depend on main summary

[DENG-476](https://mozilla-hub.atlassian.net/browse/DENG-476)

Checklist for reviewer:

[DENG-476]: https://mozilla-hub.atlassian.net/browse/DENG-476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1685)

### DIFF
--- a/dags/bqetl_main_summary.py
+++ b/dags/bqetl_main_summary.py
@@ -430,6 +430,21 @@ with DAG(
         task_concurrency=1,
     )
 
+    telemetry_derived__hcm_clients__v1 = bigquery_etl_query(
+        task_id="telemetry_derived__hcm_clients__v1",
+        destination_table="hcm_clients_v1",
+        dataset_id="telemetry_derived",
+        project_id="moz-fx-data-shared-prod",
+        owner="dthorn@mozilla.com",
+        email=[
+            "dthorn@mozilla.com",
+            "jklukas@mozilla.com",
+            "telemetry-alerts@mozilla.com",
+        ],
+        date_partition_parameter="submission_date",
+        depends_on_past=False,
+    )
+
     telemetry_derived__main_1pct__v1 = bigquery_etl_query(
         task_id="telemetry_derived__main_1pct__v1",
         destination_table="main_1pct_v1",
@@ -704,6 +719,8 @@ with DAG(
     telemetry_derived__firefox_desktop_usage__v1.set_upstream(
         firefox_desktop_exact_mau28_by_dimensions_v2
     )
+
+    telemetry_derived__hcm_clients__v1.set_upstream(wait_for_copy_deduplicate_main_ping)
 
     telemetry_derived__main_1pct__v1.set_upstream(wait_for_copy_deduplicate_main_ping)
 
--- a/dags/bqetl_main_summary.py
+++ b/dags/bqetl_main_summary.py
@@ -28,11 +28,7 @@ default_args = {
     "owner": "dthorn@mozilla.com",
     "start_date": datetime.datetime(2018, 11, 27, 0, 0),
     "end_date": None,
-    "email": [
-        "telemetry-alerts@mozilla.com",
-        "dthorn@mozilla.com",
-        "jklukas@mozilla.com",
-    ],
+    "email": ["telemetry-alerts@mozilla.com", "dthorn@mozilla.com"],
     "depends_on_past": False,
     "retry_delay": datetime.timedelta(seconds=1800),
     "email_on_failure": True,
@@ -57,7 +53,6 @@ with DAG(
         owner="wlachance@mozilla.com",
         email=[
             "dthorn@mozilla.com",
-            "jklukas@mozilla.com",
             "telemetry-alerts@mozilla.com",
             "wlachance@mozilla.com",
         ],
@@ -75,7 +70,6 @@ with DAG(
         email=[
             "dthorn@mozilla.com",
             "frank@mozilla.com",
-            "jklukas@mozilla.com",
             "telemetry-alerts@mozilla.com",
         ],
         date_partition_parameter="submission_date",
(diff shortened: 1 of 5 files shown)


sql/moz-fx-data-shared-prod/telemetry_derived/hcm_clients_v1/query.sql
```diff
@@ -0,0 +1,23 @@
...
+CROSS JOIN
+  UNNEST(payload.processes.parent.keyed_scalars.a11y_theme)
+WHERE
+  {% if is_init() %}
+    DATE(submission_timestamp) >= '2021-01-01'
+  {% else %}
+    DATE(submission_timestamp) = @submission_date
+  {% endif %}
+  AND payload.processes.parent.keyed_scalars.a11y_theme IS NOT NULL
+  AND ARRAY_LENGTH(payload.processes.parent.keyed_scalars.a11y_theme) > 0
```
fbertsch, line 17: Just for optimization? The cross join will also filter these out

sql/moz-fx-data-shared-prod/telemetry_derived/hcm_clients_v1/query.sql
```diff
@@ -0,0 +1,23 @@
...
+  AND value IS TRUE
+  AND key != "always"
```
fbertsch, line 19: Some description of this business logic, either here or in the table description, could be useful

dags/bqetl_main_summary.py
```diff
@@ -430,6 +430,21 @@
...
 
+    telemetry_derived__hcm_clients__v1 = bigquery_etl_query(
+        task_id="telemetry_derived__hcm_clients__v1",
+        destination_table="hcm_clients_v1",
+        dataset_id="telemetry_derived",
+        project_id="moz-fx-data-shared-prod",
+        owner="dthorn@mozilla.com",
+        email=[
+            "dthorn@mozilla.com",
+            "jklukas@mozilla.com",
```
fbertsch, line 441:
```suggestion
```

fbertsch (approved):
r+wc

dags/bqetl_main_summary.py
relud: will do, but it will apply to the entire dag

sql/moz-fx-data-shared-prod/telemetry_derived/hcm_clients_v1/query.sql
relud: it was in the original, and i left it in because maybe optimization, but imma remove it

sql/moz-fx-data-shared-prod/telemetry_derived/hcm_clients_v1/query.sql
relud: @MReschenberg could you provide a description for this logic

MReschenberg: "value" is derived from a LookAndFeel boolean which is TRUE when an OS high contrast mode (HCM) is present and FALSE otherwise. OS HCM is defined as HCM on windows, Increase contrast on macOS, and high contrast theme on linux. 

"key" is derived from the config pref `browser.display.document_color_use" which is has the following values / meanings: 
- "always" = always use document colors, regardless of OS HCM or colors chosen in the colors dialog (about:preferences > manage colors)
- "never" = never use document colors (this is equivalent to "always use HCM colors")
- "default" = use document colors when OS HCM is off, and use OS HCM colors when OS HCM is on.

MReschenberg: In the context of the entire query: we're looking for users who have OS HCM on and who have _not_ forced document colors everywhere (ie. the OS HCM colors should propagate to web content).

relud: thank you!

sql/moz-fx-data-shared-prod/telemetry_derived/hcm_clients_v1/query.sql
```diff
@@ -0,0 +1,34 @@
...
+  * "always" = always use document colors, regardless of OS HCM or colors chosen in the colors
+    dialog (about:preferences > manage colors)
+  * "never" = never use document colors (this is equivalent to "always use HCM colors")
+  * "default" = use document colors when OS HCM is off, and use OS HCM colors when OS HCM is on.
+
+  This condition selects users who have OS HCM on and who have not forced document colors everywhere
+  (ie. the OS HCM colors should propagate to web content).
+  */
+  AND a11y_theme.value IS TRUE
+  AND a11y_theme.key != "always"
```
relud, line 30: @MReschenberg I just noticed of the 5 dashboards, only the generic HCM clients one uses the above logic, while the other 4 use this logic:

```sql
(n.key = "never" OR (n.key != "always" AND n.value = TRUE))
```

can you advise on which logic is preferred?

MReschenberg: Ah sorry this is kinda confusing: they're measuring different things.

This logic: `(n.key = "never" OR (n.key != "always" AND n.value = TRUE))` tracks clients who have either:
 - HCM colors forced everywhere OR
 - OS HCM enabled AND document colors _not_ forced (ie. OS HCM is never being overridden by the use_document_colors preference)

The logic: `a11y_theme.value IS TRUE AND a11y_theme.key != "always"` tracks clients who both:
- have OS HCM enabled AND
- have _not_ forced document colors
(ie. the last part of the first query)

The former allows us to count clients who are overriding document colors without using their OS HCM theme -- they're doing it in Firefox directly. The latter requires clients have an OS HCM enabled to be counted. 

These are two different kinds of users, and we want to be able to measure them each. The generic dashboard should probably be renamed to "OS HCM clients" instead of "HCM clients" if it isn't already named that way 😁

Thanks for your patience, lemme know if you have any other questions.

PR #2529, opened 2021-11-25
Add a11y_theme fields to clients daily

scholtzan:
For migrating accessibility dashboards to Looker: https://mozilla-hub.atlassian.net/browse/DENG-52

Making the fields `a11y_theme`, `a11y_hcm_foreground` and `a11y_hcm_background` available in clients_daily

PR #4375, opened 2023-10-03
Remove hcm_clients in favor of clients_daily

relud:


sql/moz-fx-data-shared-prod/telemetry_derived/hcm_clients_v1/query.sql
```diff
@@ -1,34 +1,37 @@
+WITH clients AS (
+  SELECT
+    DATE(main.submission_timestamp) AS submission_date,
+    main.client_id,
+    mozfun.stats.mode_last(ARRAY_AGG(main.normalized_channel)) AS channel,
+    mozfun.stats.mode_last(ARRAY_AGG(main.environment.system.os.name)) AS os,
+    mozfun.json.mode_last(ARRAY_AGG(a11y_theme)) AS a11y_theme,
+    main.environment.system.os.name AS os,
```
fbertsch, line 8: From a maintenance POV, it's looking like we should make sure these are in `clients_daily` and have them query that table directly. We could expose a subset of fields as a new Looker explore.

relud: turns out it's already in clients_daily, i just missed it. it was added in https://github.com/mozilla/bigquery-etl/pull/2529, and now appears to already have history at least as far back as 2021-01-01.

fbertsch (approved):

PR #4379, opened 2023-10-04
feat(): adding firefox ios retention week 4 aggregates query

kik-kik:
feat(): adding firefox ios retention week 4 aggregates query

This means that unless we decide to "backfill" this table the numbers reported by it should not be affected by Shredder.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1695)

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_funnel_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_funnel_v1/query.sql
@@ -46,6 +46,8 @@ store_stats AS (
   ON
     country_names.name = views_data.country_name
 ),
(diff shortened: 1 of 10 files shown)


sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_funnel_v1/query.sql
```diff
@@ -46,6 +46,8 @@ store_stats AS (
   ON
     country_names.name = views_data.country_name
 ),
+-- TODO: we may need to consider updating the source for new profiles to be firefox_ios.firefox_ios_clients
```
kik-kik, line 49: @soGaussian what are your thoughts here? I think it would make sense to use use an upstream table instead to make sure that at least this subset of tables is aligned.

soGaussian (approved):

PR #4396, opened 2023-10-09
DS-2948 Update second seen date

lucia-vargas-a:
Improved calculation of second seen date and removing two aggregations, which will also reduce the execution times during initialization.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1723)


sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -242,23 +243,21 @@ first_seen_date AS (
...
+--  main ping after first_seen_date. Dates reported by other pings are excluded.
 second_seen_date AS (
   SELECT
     client_id,
-    IF(
-      ARRAY_LENGTH(ARRAY_AGG(seen_dates)) > 1,
-      ARRAY_AGG(seen_dates ORDER BY value_date ASC)[SAFE_OFFSET(1)],
-      NULL
-    ) AS second_seen_date
+    ARRAY_AGG(seen_dates ORDER BY value_date ASC)[SAFE_OFFSET(0)] AS second_seen_date
```
kik-kik, line 251: If we say second date can only come from the main ping, wouldn't it be easier to use the `main_ping` CTE here as the source instead which already contains sorted dates field? This way we could avoid unnesting and aggregating again?

lucia-vargas-a: Yes, I think you're right, now that we have condition `> first_seen_date` allows us to use the main ping aggregation. Since this data is already tested, are you happy that I open a Jira ticket as an improvement to implement together with the data checks?

kik-kik (approved):
I'd consider removing the doc link and simplifying the second date CTE by using main_ping CTE instead.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
lucia-vargas-a: Indeed, two aggregations reduced, updated, thanks!

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -242,25 +221,19 @@ first_seen_date AS (
...
 second_seen_date AS (
-  SELECT
+   SELECT
     client_id,
-    IF(
-      ARRAY_LENGTH(ARRAY_AGG(seen_dates)) > 1,
-      ARRAY_AGG(seen_dates ORDER BY value_date ASC)[SAFE_OFFSET(1)],
-      NULL
-    ) AS second_seen_date
+    all_dates[SAFE_OFFSET(0)] AS second_seen_date
```
kik-kik, line 229: And what if the first date is the same as the first date of the main ping? Shouldn't we be capturing offset 1 here?

lucia-vargas-a: That would happen is there was another ping type before, so it's okay to retrieve the first main ping reported as long as it's after the `first_seen_date`.

PR #4404, opened 2023-10-10
add code for cohort_daily_statistics using clients_first_seen_v2 with…

Marlene-M-Hirose:
… new columns from clients_first_seen_v2

Beginning of updating a current retention model - telemetry_derived.cohort_daily_statistics.

Name of table will change as well as location of files.  

Some columns are not added as they are specifically in the table for mobile (ex. device_model) or unreliable (also ex. device_model)

Some new columns include apple_model_id, which is probably for mobile, and can be removed.

Added columns for num_clients_active_once_in_last_28_days as well as num_clients_active_once_in_last_7_days.

Activity_date of "2023-10-01" chosen at random to build a small table for testing and analysis purposes

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1740)


sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
```diff
@@ -0,0 +1,167 @@
...
+  COUNTIF((active_client_id IS NOT NULL) AND (active_clients_days_since_seen = 0)) AS num_clients_active_on_day,
+  COUNTIF((active_client_id IS NOT NULL) AND
+    (COALESCE(
+    BIT_COUNT(mozfun.bits28.from_string('0000000000000000000001111111') & active_client_days_seen_bits) > 0,
+    FALSE
+    )) ) AS num_clients_active_once_in_last_28_days,
+  COUNTIF((active_client_id IS NOT NULL) AND
+    ( COALESCE(
+    BIT_COUNT(mozfun.bits28.from_string('0111111111111111111111111111') & active_client_days_seen_bits) > 0,
+    FALSE) )) AS num_clients_active_once_in_last_7_days,
```
ANich, line 133: I think these are swapped, should be 
```suggestion
  COUNTIF((active_client_id IS NOT NULL) AND
    (COALESCE(
    BIT_COUNT(mozfun.bits28.from_string('0000000000000000000001111111') & active_client_days_seen_bits) > 0,
    FALSE
    )) ) AS num_clients_active_once_in_last_7_days,
  COUNTIF((active_client_id IS NOT NULL) AND
    ( COALESCE(
    BIT_COUNT(mozfun.bits28.from_string('0111111111111111111111111111') & active_client_days_seen_bits) > 0,
    FALSE) )) AS num_clients_active_once_in_last_28_days,
```

ANich (commented):
At a glance, this lgtm. Worth updating to say the initial version will only contain desktop numbers given the change from `unified_metrics` -> `clients_last/first_seen`

Marlene-M-Hirose:
I haven't added any filters that would make this table Desktop only yet. I believe the original cohort_daily_statistics covered all, desktop and mobile. Is this right?

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
```diff
@@ -0,0 +1,167 @@
...
+    -- (2) Note this is a pretty big scan... Look here for problems
+),
+activity_cohort_match AS (
+  SELECT
+    cohorts_in_range.client_id AS cohort_client_id,
+    submission_date_activity.client_id AS active_client_id,
+    submission_date_activity.days_seen_bits as active_client_days_seen_bits,
+    submission_date_activity.days_visited_1_uri_bits,
+    submission_date_activity.days_interacted_bits,
+    mozfun.bits28.days_since_seen(submission_date_activity.days_seen_bits) as active_clients_days_since_seen,
```
irrationalagent, line 82: we may want parallel fields here from `days_interacted_bits` and `days_visited_1_uri_bits`

irrationalagent:
It occurs to me that one safety check we should do here is to ensure that there aren't (too many) any weird entries in CLS that occur before the FSD in CFS_v2. That shouldn't happen, but to my knowledge we haven't checked that yet.

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
```diff
@@ -0,0 +1,167 @@
...
+  normalized_os_version,
+  os_version_major,
+  os_version_minor,
+  COUNT(cohort_client_id) AS num_clients_in_cohort,
+  COUNTIF((active_client_id IS NOT NULL) AND (active_clients_days_since_seen = 0)) AS num_clients_active_on_day,
+  COUNTIF((active_client_id IS NOT NULL) AND
+    (COALESCE(
+    BIT_COUNT(mozfun.bits28.from_string('0000000000000000000001111111') & active_client_days_seen_bits) > 0,
+    FALSE
+    )) ) AS num_clients_active_once_in_last_7_days,
```
irrationalagent, line 129: nit: should this be "at least once" or something similar?

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
```diff
@@ -0,0 +1,167 @@
...
+    ) AS os_version_minor,
+  FROM
+    -- telemetry_derived.rolling_cohorts_v1
+    telemetry_derived.clients_first_seen_v2
+  WHERE
+    -- cohort_date
+    first_seen_date
+    BETWEEN DATE_SUB(DATE("2023-10-01"), INTERVAL 180 DAY)
+    AND DATE_SUB(DATE("2023-10-01"), INTERVAL 1 DAY)
+    -- (1) No need to get activity for the cohort created on activity_date - everyone will be retained
```
irrationalagent, line 72: although, I don't think this is true anymore given how we are counting clients as retained based on the main ping. retention won't be 100% on the first day because some clients in CFS never send a main ping. so, there is a question whether we do actually want to calculate "retention" on first_seen_date = activity_date.

if we do, then people will be confused. if we don't, then we are technically dropping some information that MIGHT someday be useful.

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
```diff
@@ -0,0 +1,167 @@
...
+    -- (2) Note this is a pretty big scan... Look here for problems
+),
+activity_cohort_match AS (
+  SELECT
+    cohorts_in_range.client_id AS cohort_client_id,
```
gkabbz, line 77: I don't think we should rename the field to cohort_client_id. Should leave it as client_id.

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
```diff
@@ -0,0 +1,167 @@
...
+    days_seen_bits,
+    days_visited_1_uri_bits,
+    days_interacted_bits
+),
+-- Get all the cohorts that are still in range of the current day of activity (180 days)
+cohorts_in_range AS (
+  SELECT
+    client_id,
+    sample_id,
+    first_seen_date AS cohort_date,
```
gkabbz, line 25: Should we be renaming this? I think it's useful to have consistency between tables. My thinking is this should be kept as `first_seen_date`.

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
```diff
@@ -0,0 +1,167 @@
...
+  distribution_id,
+  -- is_default_browser,
+  locale,
+  normalized_app_name,
+  normalized_channel,
+  normalized_os,
+  normalized_os_version,
+  os_version_major,
+  os_version_minor,
+  COUNT(cohort_client_id) AS num_clients_in_cohort,
```
gkabbz, line 123: What's the rationale for materializing this column? Wouldn't not having it and doing a count(*) or count(client_id) at analysis work the same?

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
```diff
@@ -0,0 +1,167 @@
+-- Get all clients who had activity in the last 28 days and their bit patterns.
+WITH submission_date_activity AS (
```
gkabbz, line 2: We should change the naming of this cte to show that it's pulling active clients. Something like `active_clients` or `client_activity`. Don't think the submission_date is giving me any additional info about what's being pulled in the CTE.

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
```diff
@@ -0,0 +1,167 @@
...
+  SELECT
+    client_id,
+    days_seen_bits,
+    days_visited_1_uri_bits,
+    days_interacted_bits,
```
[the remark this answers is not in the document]
gkabbz: I could see them being used to distinguish between overall retention and retention of clients who count towards DAU. But now don't seem to be getting used.

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
Marlene-M-Hirose: makes for an awfully long column name, unless we want to create a new acronym of "alo". Happy to change though

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
Marlene-M-Hirose: Are we using first_seen_date = activity_date elsewhere?

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
Marlene-M-Hirose: Will users understand that first_seen_date is the cohort_date if we don't rename it?

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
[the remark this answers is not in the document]
Marlene-M-Hirose: I thought it was used in the bitcount section - it is in Alex's table for a different calculation - which I was patterning off of. Taking these out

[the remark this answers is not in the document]
irrationalagent: I think it would be helpful to have these DAU-based retention measures. but happy to discuss.

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
irrationalagent: or we could even be more literal `client_main_ping_activity` since as an org on desktop we now refer to being "Active" as having > 0 active hours and URIs.

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
Marlene-M-Hirose: This is supposed to be an aggregated table at the end. Should not have selected client_id and sample_id as columns for the table. My mistake

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
```diff
@@ -0,0 +1,167 @@
...
+    submission_date AS activity_date
+  FROM
+    telemetry.clients_last_seen_v1 -- this might cause an issue because definition of first_seen_date in this table is different from clients_first_seen_v2
+  WHERE
+    submission_date = DATE("2023-10-01")
```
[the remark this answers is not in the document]
Marlene-M-Hirose: This is my test table for data from 2023-10-01. The hard coded dates will go away once we settle on a model and get the DAG running

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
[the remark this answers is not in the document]
Marlene-M-Hirose: taken out

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
Marlene-M-Hirose: replaced submission_date_activity with client_activity

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
gkabbz, line 1: File name has a spelling mistake.

Marlene-M-Hirose: That was on purpose. I removed the vowel to make it slightly shorter. Should have changed "seen" to "sn"

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
```diff
@@ -0,0 +1,150 @@
...
+  COUNT(client_id) AS num_clients_in_cohort,
+  COUNTIF((active_client_id IS NOT NULL) AND (active_clients_days_since_seen = 0)) AS num_clients_active_on_day,
+  COUNTIF((active_client_id IS NOT NULL) AND
+    (COALESCE(
+    BIT_COUNT(mozfun.bits28.from_string('0000000000000000000001111111') & active_client_days_seen_bits) > 0,
+    FALSE
+    )) ) AS num_clients_active_atleastonce_in_last_7_days,
+  COUNTIF((active_client_id IS NOT NULL) AND
+    ( COALESCE(
+    BIT_COUNT(mozfun.bits28.from_string('0111111111111111111111111111') & active_client_days_seen_bits) > 0,
```
irrationalagent, line 118: This is a slightly tricky bit. I think we want this column to count all clients active in the past 28 days, but the `mozfun.bits28.from_string('0111111111111111111111111111') &` part will ignore the first day in that 28 day period (so its only looking at 27 days actually). 

The original intention of this code from `clients_first_seen_28_days_later` is to calculate so-called first month repeat users, which ignores the client's first day, and only counts them if they show up for an _additional day in their first 28 days_. 

I think we would want to capture both first month repeat users AND 28-day retention in this table. Probably the best way to do that would be to change this column to remove the bitmask, i.e.  

```sql
COUNTIF((active_client_id IS NOT NULL) AND
    ( COALESCE(
    BIT_COUNT(active_client_days_seen_bits) > 0,
    FALSE) )) AS num_clients_active_atleastonce_in_last_28_days
```
Then, add a new column, like

```sql
COUNTIF((active_client_id IS NOT NULL) AND
    DATE_DIFF(activity_date, cohort_date, DAY) <= 27 AND
    ( COALESCE(
    BIT_COUNT(mozfun.bits28.from_string('0111111111111111111111111111') & active_client_days_seen_bits) > 0,
    FALSE) )) AS num_clients_repeat_first_month_users
```
This new condition should only be TRUE in cases where clients are active at least 2 days anytime in their first 28 days.

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
```diff
@@ -0,0 +1,150 @@
...
+    ) AS os_version_major,
+    COALESCE(
+      SAFE_CAST(NULLIF(SPLIT(normalized_os_version, ".")[SAFE_OFFSET(1)], "") AS INTEGER),
+      0
+    ) AS os_version_minor,
+  FROM
+    telemetry_derived.clients_first_seen_v2
+  WHERE
+    first_seen_date
+    BETWEEN DATE_SUB(@activity_date, INTERVAL 180 DAY)
```
irrationalagent, line 61: One awkward thing about 180 days is that its not equally divisible by 28, which is one of our main retention intervals. 196 would give us 7 complete months, and would tell us if any clients make it into their second half their first year.

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
```diff
@@ -0,0 +1,150 @@
...
+  -- is_default_browser,
+  locale,
+  normalized_app_name,
+  normalized_channel,
+  normalized_os,
+  normalized_os_version,
+  os_version_major,
+  os_version_minor,
+  COUNT(client_id) AS num_clients_in_cohort,
+  COUNTIF((active_client_id IS NOT NULL) AND (active_clients_days_since_seen = 0)) AS num_clients_active_on_day,
```
irrationalagent, line 110: We should probably document how people should use this table to calculate the 28 day and 7 day retention measures, since the queries aren't too bad but neither are they that obvious. For 28 day retention, I think it would be something like this, for individual cohorts:

```sql
SELECT
    cohort_date,
    SAFE_DIVIDE(DATE_DIFF(activity_date, cohort_date, DAY) + 1, 28) AS month,
    SAFE_DIVIDE(num_clients_active_atleastonce_in_last_28_days, num_clients_in_cohort) AS retention_rate
FROM cohort_daily_statistics
WHERE MOD(DATE_DIFF(activity_date, cohort_date, DAY) + 1, 28) = 0
ORDER BY cohort_date, activity_date
```
(I haven't tested that code exactly, but that should be roughly it.)

The idea is to measure at equally spaced 28 day intervals.

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
```diff
@@ -0,0 +1,150 @@
...
+  apple_model_id, -- from clients_first_seen_v2
+  distribution_id,
+  -- is_default_browser,
+  locale,
+  normalized_app_name,
+  normalized_channel,
+  normalized_os,
+  normalized_os_version,
+  os_version_major,
+  os_version_minor,
```
ANich, line 108: Could this be a `activity_cohort_match.* EXCEPT`?

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/metadata.yaml
```diff
@@ -0,0 +1,30 @@
+friendly_name: Cohort Daily Statistics v2
```
irrationalagent, line 1: Are we treating this PR as blocked by https://github.com/mozilla/bigquery-etl/pull/4482/ ? IMO we probably should, and then add a `startup_profile_selection_reason` column from `clients_first_seen_v2`

irrationalagent (approved):

irrationalagent (dismissed):
meant to mark this as r+wc

ANich (approved):
I'm unclear on whether the use-case for retention on specific days still exists but Technically this LGTM. Left one comment

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/metadata.yaml
Marlene-M-Hirose: I can add that

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
Marlene-M-Hirose: Sounds reasonable to me. Will make this change

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
Marlene-M-Hirose: Will test your code and give this a think. Update tomorrow

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
```diff
@@ -0,0 +1,150 @@
+-- Get all clients who had activity in the last 28 days and their bit patterns.
+WITH submission_date_activity AS (
+  SELECT
+    client_id,
+    days_seen_bits,
```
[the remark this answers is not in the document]
Marlene-M-Hirose: Got it. Will modify

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
```diff
@@ -0,0 +1,150 @@
...
+    submission_date,
+    days_seen_bits
+),
+-- Get all the cohorts that are still in range of the current day of activity (180 days)
+cohorts_in_range AS (
+  SELECT
+    client_id,
+    first_seen_date AS cohort_date,
+    DATE(@activity_date) AS activity_date,
+    -- activity_segment, -- for desktop: from clients_last_seen, for mobile: calculated field from mobile_with_searches in unified_metrics
```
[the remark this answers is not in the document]
Marlene-M-Hirose: The intention is to match the original cohort_daily_statistics and add appropriate columns. I'll update the code

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
Marlene-M-Hirose: this makes sense to me. Will update. Any objections @ANich ?

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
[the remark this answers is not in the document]
irrationalagent: We decided to leave these out for now to avoid confusion about when they are measured.

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
irrationalagent: We should put this in the lookml stage.

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
Marlene-M-Hirose: Going to leave this as is right now.

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_stats_clients_frst_seen_v2/query.sql
Marlene-M-Hirose: Changed direction with this code, comment no longer applies

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
```diff
@@ -0,0 +1,200 @@
...
+    days_seen_bits,
+    (days_visited_1_uri_bits & days_interacted_bits)
+),
+-- Get all the cohorts that are still in range of the current day of activity (196 days)
+cohorts_in_range AS (
+  SELECT
+    client_id,
+    first_seen_date AS cohort_date,
+    DATE(@activity_date) AS activity_date,
+    app_build_id,
```
gkabbz, line 24: Remove this field as well. Pretty big in terms of unique values. See: https://docs.google.com/spreadsheets/d/1s_rCwApytiask9a7rcIF-Mj5_4p7GVjW2dqH0dQkd70/edit#gid=0

Marlene-M-Hirose: I've taken out document_id, dl_token, app_build_id columns, rename activity_date to submission_date, rename cohort_date to first_seen_date to match clients_first_seen_28_days_later as per conversation with George

Marlene-M-Hirose:
> meant to mark this as r+wc

updated

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
Marlene-M-Hirose: I've also changed the retention date to look back 106 days instead of 196

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
Marlene-M-Hirose: sounds good, will put the info in at the lookml stage

gkabbz (approved):

PR #4413, opened 2023-10-12
FXA-6721 Add user-facing view for stage FxA accounts table

akkomar:
This adds missing fields to nonprod_accounts_v1, restricts its access to a workgroup and introduces mozilla-confidential user-facing view with unrestricted columns (currently all except email).

This will require a corresponding PR to cloudops-infra.

Checklist for reviewer:


sql/moz-fx-data-shared-prod/accounts_backend_external/dataset_metadata.yaml
```diff
@@ -2,10 +2,13 @@ friendly_name: Firefox Accounts Backend External
...
+  `mozilla-confidential` user-facing views are available in the `accounts_backend` dataset.
 dataset_base_acl: restricted
 user_facing: false
 labels: {}
 workgroup_access:
-- role: roles/bigquery.dataEditor
+- role: roles/bigquery.dataViewer
   members:
-  - workgroup:platform-infra/internal
+  - workgroup:accounts-confidential
```
akkomar, line 14: This restricts access to the whole `accounts_backend_external` dataset to a group created in https://mozilla-hub.atlassian.net/browse/DSRE-1408.
non-confidential user-facing views are going to be deployed in `accounts_backend` dataset.

sql/moz-fx-data-shared-prod/accounts_backend/dataset_metadata.yaml
kik-kik, line 1: Rather than pre-prending `nonprod` at the start of each non prod dataset name should we just have a nonprod or staging namespace?
For example: `accounts_backend_nonprod` or `accounts_backend_staging`?

sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/metadata.yaml
kik-kik, line 1: Should we also include label `incremental: false` here?

sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/query.sql
kik-kik, line 1: Should we also have a corresponding `schema.yaml` for this query?

sql/moz-fx-data-shared-prod/accounts_backend/dataset_metadata.yaml
akkomar: Either is fine but for consistency with frontend I'd keep it as is. In Glean.js both prod and stage data goes to the same dataset.

sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/metadata.yaml
akkomar: It would be no-op, I didn't add it since the following comment makes it clear that this is not an incremental query:
```
  # destination is the whole table, not a single partition,
  # so don't use date_partition_parameter
```

sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/query.sql
akkomar: There is one but I forgot to update it. It is done now.

sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/schema.yaml
```diff
@@ -26,6 +29,9 @@ fields:
 - name: profileChangedAt
   type: NUMERIC
   mode: NULLABLE
+- name: keysChangedAt
+  type: NUMERIC
```
sean-rose, line 33: I'd strongly recommend converting the timestamp and boolean columns into the appropriate native BigQuery type for usability.

Looks like the timestamp column values are milliseconds-since-Unix-epoch, so that'd be `SAFE.TIMESTAMP_MILLIS(SAFE_CAST(... AS INT))`, and the boolean columns would be `SAFE_CAST(... AS BOOL)`.

sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/schema.yaml
sean-rose, line 1: Column descriptions could be helpful (although there's currently an issue with ETLs that overwrite the entire table where the column descriptions get removed when the ETL runs).

sql/moz-fx-data-shared-prod/accounts_backend_external/dataset_metadata.yaml
sean-rose: IMO it'd be good for DEs who are working on these ETLs to still have Data Editor access here (e.g. we could have a custom subgroup like `accounts-confidential/writers`).

@whd thoughts?

whd: I mentioned this to akkomar as an option but IIRC we were in agreement that no direct write access would be preferable if possible and that we'd add it via a writers subgroup if it wound up being necessary. A general solution or pattern is the subject of https://mozilla-hub.atlassian.net/browse/DSRE-1023.

akkomar: I was expecting there will be simple 1-1 copy-data queries here, without any complex logic so write access didn't seem to be required to me.

sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/schema.yaml
akkomar: I'm wondering if it's better to keep them as is in the table to:
1. Preserve original format to make it less surprising for FxA developers looking at this table
2. Make it easier to switch to some other CDC tool (e.g. Dataform) in case we decide it's preferable instead of federated queries.

For other consumers we could do the conversions you proposed in the view.

What do you think?

sql/moz-fx-data-shared-prod/accounts_backend/nonprod_accounts/view.sql
```diff
@@ -0,0 +1,17 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.accounts_backend.nonprod_accounts`
+AS
+SELECT
+  uid,
+  verifierVersion,
+  verifierSetAt,
```
sean-rose, line 7: I'd also include `emailVerified`, which is a key signal for whether the account is legit.

sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/schema.yaml
sean-rose: That approach would be OK, just less convenient IMO.

Personally I wouldn't worry too much about FxA developers being surprised.  If anything, I'd think they'd be _pleasantly_ surprised they didn't have to figure out that `TIMESTAMP_MILLIS()` is the BigQuery function to convert milliseconds into a usable timestamp value, and then deal with the confusion of not being able to pass the `NUMERIC` millisecond values directly to that function because it expects an `INT64`.

And I wouldn't be worried about switching to some other CDC tool, because in that case the data would probably be landing in some other project and we'd have rote copying ETLs here where we could apply similar data type conversions.

sql/moz-fx-data-shared-prod/accounts_backend/nonprod_accounts/view.sql
```diff
@@ -0,0 +1,17 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.accounts_backend.nonprod_accounts`
```
sean-rose, line 2: Since this is an abridged/redacted version of the accounts data, should we give the view a more distinctive name that indicates that?

Or what is the plan for exposing the sensitive data columns as necessary with appropriate restrictions?

akkomar: Restricted columns will be accessed directly via the table in `_external` dataset that will be accessible by members of `accounts-confidential` group. These columns won't be exposed in Looker.

Having said that the name sounds fine to me but I'm open to suggestions.

sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/schema.yaml
akkomar: Makes sense, I'll convert these in the table.

sql/moz-fx-data-shared-prod/accounts_backend/nonprod_accounts/metadata.yaml
```diff
@@ -0,0 +1,17 @@
+---
+friendly_name: Accounts table from FxA database in stage environment
+description: |-
+  An authorized view on top of the `accounts_backend_external.nonprod_accounts_v1` table
+  that only includes non-sensitive fields.
+  Some fields in this table are converted to a more user-friendly, BigQuery-native format:
+    - `uid` is converted from bytes to a hex string
+    - `emailVerified` is converted from INTEGER to BOOLEAN
```
sean-rose, line 8: Suggest making this more generic, as future boolean columns would presumably also be converted.
```suggestion
    - boolean integer columns are converted to BOOL
```

sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/metadata.yaml
```diff
@@ -1,8 +1,14 @@
 ---
-friendly_name: FxA Accounts nonprod
+friendly_name: Accounts table from FxA database in stage environment
 description: >
   A mirror of the `accounts` table from the staging FxA CloudSQL database,
   updated daily to match the current state of the table.
+  Some fields in this table are converted to a more user-friendly, BigQuery-native format:
+    - `uid` is converted from bytes to a hex string
+    - `emailVerified` is converted from INTEGER to BOOLEAN
```
sean-rose, line 8:
```suggestion
    - boolean integer columns are converted to BOOL
```

sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/schema.yaml
```diff
@@ -1,37 +1,47 @@
 fields:
-- name: uid
-  type: BYTES
+- description: |-
+    Account ID in hexadecimal format.
+    FxA stores this as bytes, for purposes of logging or integration
+    with other systems we convert it to a hex string.
+  name: uid
+  type: STRING
   mode: NULLABLE
```
sean-rose, line 8: IMO it'd be more readable to put the description last rather than first.

sean-rose (approved):
r+wc

sql/moz-fx-data-shared-prod/accounts_backend_external/nonprod_accounts_v1/schema.yaml
akkomar: We should have a formatter for schema.yaml.

PR #4424, opened 2023-10-13
feat(DENG-1774): adding fenix derived firefox android clients v2

kik-kik:
# feat(DENG-1774): adding fenix derived firefox android clients v2

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1775)


sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v2/query.sql
```diff
@@ -0,0 +1,224 @@
...
+      CASE
+        WHEN first_session.adjust_info IS NOT NULL
+          THEN "first_session"
+        WHEN metrics.adjust_info IS NOT NULL
+          THEN "metrics"
+        ELSE NULL
+      END AS adjust_info__source_ping
+    ) AS metadata,
+    -- field to help us identify suspicious devices on iOS, for more info see: bug-1846554
+    (app_version = '107.2' AND submission_date >= '2023-02-01') AS is_suspicious_device_client,
```
bani, line 149: this shouldn't be here, the incident is iOS only

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v2/query.sql
```diff
@@ -0,0 +1,224 @@
...
+      _previous.adjust_creative,
+      _previous.adjust_network
+    )
+  ).*,
+  STRUCT(
+    COALESCE(
+      _previous.metadata.is_reported_first_session_ping
+      OR _current.metadata.is_reported_first_session_ping,
+      FALSE
+    ) AS is_reported_first_session_ping,
```
bani, line 205: current name on Android is just `reported_first_session_ping`. We can change it for consistency, but this field is used a lot so it'll break a bunch of queries once we update the view.

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v2/query.sql
kik-kik: oh yes, fair point. I thought I deleted this.

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v2/query.sql
kik-kik: ok, changed the naming for android.

bani (approved):

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v2/init.sql
```diff
@@ -0,0 +1,183 @@
+-- Query first observations for Firefox Android Clients.
```
fbertsch, line 1: Let's remove the `init.sql` file in favor of `is_init()` in the `query.sql` file

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v2/query.sql
```diff
@@ -0,0 +1,222 @@
...
+  COALESCE(
+    _previous.first_reported_country,
+    _current.first_reported_country
+  ) AS first_reported_country,
+  COALESCE(_previous.first_reported_isp, _current.first_reported_isp) AS first_reported_isp,
+  COALESCE(_previous.channel, _current.channel) AS channel,
+  COALESCE(_previous.device_manufacturer, _current.device_manufacturer) AS device_manufacturer,
+  COALESCE(_previous.device_model, _current.device_model) AS device_model,
+  COALESCE(_previous.os_version, _current.os_version) AS os_version,
+  COALESCE(_previous.app_version, _current.app_version) AS app_version,
```
fbertsch, line 180: Can you confirm this isn't missing any fields from `firefox_android_clients`? (sans `last_reported` fields, we can drop those). I'm seeing e.g. locale missing.

fbertsch (changes requested):
Before we implement this change, I think we need to have a discussion with Adjust. When we were working on client-id regeneration, they indicated to me that attribution _can_ change. IIRC they said it may go from missing to _something_. If that "missing" is NULLs or empty strings, this change will work as we expect. However it they fill in even a single value then we won't get the updated attribution later.

fbertsch:
@kik-kik I sent them an email and cc'ed you.

kik-kik:
Hey @fbertsch , sorry I've been out on PTO. What is the conclusion, apart from using `is_init()` and adding `locale` are we good?

kik-kik:
@fbertsch `locale` is now in the query and the query has been modified to support `is_init()`. Could you please re-review the PR? 🙇

sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v2/query.sql
kik-kik: As far as I can see they are all there, locale should be there now too.

fbertsch (approved):
LGTM. Adjust has indicated that we will only get a different attribution if a user is [reattributed](https://help.adjust.com/en/article/reattribution), which would happen if they become dormant. That's not a situation we optimize for in marketing.

kik-kik:
> LGTM. Adjust has indicated that we will only get a different attribution if a user is [reattributed](https://help.adjust.com/en/article/reattribution), which would happen if they become dormant. That's not a situation we optimize for in marketing.

Thanks for double checking @fbertsch 🙇.

Will merge the change now, and deploy.

To infinity and beyond! 🚀

PR #4426, opened 2023-10-14
DENG1546 - Fix bqetl serp DAG failure due to missing parameter

alekhyamoz:
The bqetl_serp DAG is failing due to the below error 
Error in query string: Error processing job 'moz-fx-data-shared-
```Query parameter - 'submission_date' not found at [13:48]```

Error logs - https://workflow.telemetry.mozilla.org/dags/bqetl_serp/grid?dag_run_id=scheduled__2023-10-13T00%3A00%3A00%2B00%3A00&task_id=firefox_desktop_serp_events__v1&tab=logs

This should fix the issue

### DIFF
--- a/dags/bqetl_serp.py
+++ b/dags/bqetl_serp.py
@@ -57,6 +57,9 @@ with DAG(
         ],
         date_partition_parameter=None,
         depends_on_past=False,
+        parameters=[
+            'submission_date:DATE:{{ (execution_date - macros.timedelta(hours=24)).strftime("%Y%m%d") }}'
+        ],
         sql_file_path="sql/moz-fx-data-shared-prod/firefox_desktop_derived/serp_events_v1/query.sql",
     )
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: dags/bqetl_serp.py [hunk 1]
```diff
@@ -57,6 +57,9 @@
         ],
         date_partition_parameter=None,
         depends_on_past=False,
+        parameters=[
+            'submission_date:DATE:{{ (execution_date - macros.timedelta(hours=24)).strftime("%Y%m%d") }}'
+        ],
```
scholtzan, line 62: So if this query runs for 2023-10-05, the query itself subtracts one day. Should the results be stored in the partition for 2023-10-05 or 2023-10-04?

The current change will have it in 2023-10-04.

alekhyamoz: The results should be stored in 10-05-2023

scholtzan (approved):

PR #4437, opened 2023-10-16
DENG-476 - Remove unnecessary reference to main_v4 from ssl ratios

relud:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1791)


chutten (approved):
I don't rightly know what the impact of this is. I think it means that the DAG no longer depends on main_v4 _and_ dry runs to determine deps are fast in the new one? r+ if so.

relud:
> I don't rightly know what the impact of this is. I think it means that the DAG no longer depends on main_v4 _and_ dry runs to determine deps are fast in the new one? r+ if so.

The only impact here is that the dag generation will rely on auto detection of dependencies, and we know it works because the generate-dags CI detected no missing dag changes.

we no longer dry run to determine referenced tables. instead we parse the sql locally, which isn't impacted by the described issue, and is fast.

PR #4445, opened 2023-10-17
Glean usage checks

fbertsch:
A few API changes:
1. Check files now need to have the marker anywhere in the query, not just at the beginning
2. Backfill commands now run the check on each partition

Checklist for reviewer:


bigquery_etl/cli/generate.py
```diff
@@ -19,6 +19,9 @@ def generate_group():
     generator_path = ROOT / SQL_GENERATORS_DIR
 
     for path in generator_path.iterdir():
+        if "__pycache__" in path.parts:
```
fbertsch, line 22: Because we now have an init file in `sql_generators`, we _also_ get a `__pycache__` dir which has to be ignored

bigquery_etl/cli/check.py
```diff
@@ -53,7 +55,7 @@ def _render_result_split_by_marker(marker, rendered_result):
 
     for sql_statement in rendered_result:
         sql_statement = sql_statement.strip()
-        if re.search(f"^#{marker}", sql_statement, re.IGNORECASE):
+        if re.search(f"#{marker}", sql_statement, re.IGNORECASE):
```
fbertsch, line 58: We could remove this change. But then we'd have to move the `#fail` marker to above the `{{ header }}`, which feels strange for a header.

In reality I don't see why the marker needs to be at the beginning, but happy to change this back if the team thinks otherwise.

sql_generators/glean_usage/templates/baseline_clients_daily.checks.sql
scholtzan, line 1: This file can be removed since the check won't be run for the view (as those aren't scheduled)

sql_generators/glean_usage/templates/baseline_clients_first_seen.checks.sql
scholtzan, line 1: This file can also go

scholtzan (approved):

sql_generators/glean_usage/templates/baseline_clients_daily.checks.sql
fbertsch: I forgot I had made these before figuring out a view check wouldn't work!

PR #4446, opened 2023-10-17
Fix Glean usage generation; don't skip query generation when tables don't exist

scholtzan:
This should fix the artifact deploys.

`glean_usage` is generating various queries and corresponding views for each Glean application. The logic seems to skip the generation if a table that is referenced by a view does not exist. However, this is initially the case when a new app gets added. This PR fixes this behaviour.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1804)

### DIFF
--- a/sql_generators/glean_usage/common.py
+++ b/sql_generators/glean_usage/common.py
@@ -234,22 +234,23 @@ class GleanTable:
                     **render_kwargs,
                 )
 
+        # generated files to update
+        Artifact = namedtuple("Artifact", "table_id basename sql")
+        artifacts = [
+            Artifact(view, "metadata.yaml", view_metadata),
+            Artifact(table, "metadata.yaml", table_metadata),
+            Artifact(table, "query.sql", query_sql),
+        ]
+
         if not (referenced_table_exists(view_sql)):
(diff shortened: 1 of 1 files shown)


sql_generators/glean_usage/common.py
```diff
@@ -234,22 +234,23 @@ def generate_per_app_id(
...
+        Artifact = namedtuple("Artifact", "table_id basename sql")
+        artifacts = [
+            Artifact(view, "metadata.yaml", view_metadata),
+            Artifact(table, "metadata.yaml", table_metadata),
+            Artifact(table, "query.sql", query_sql),
+        ]
+
         if not (referenced_table_exists(view_sql)):
             logging.info("Skipping view for table which doesn't exist:" f" {table}")
-            return
```
scholtzan, line 239: This also skips generating the query of the table that is referenced in the view in the first place

relud (approved):

sql_generators/glean_usage/common.py
fbertsch: This must have worked at some point? But it hasn't for the past two years? curious...

PR #4447, opened 2023-10-17
DS3244 - Add derived datasets for review checker data

alekhyamoz:
Related to https://mozilla-hub.atlassian.net/browse/DS-3244

Checklist for reviewer:


betling (approved):
Compared to dashboard data and looks good from my end.  Only tiny differences due to the shredder.

sql_generators/review_checker/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+    default="moz-fx-data-shared-prod",
+)
+@use_cloud_function_option
+def generate(target_project, output_dir, use_cloud_function):
+    """Generate per-app queries, views and metadata for review checker aggregates.
+
+    The parent folders will be created if not existing and existing files will be overwritten.
+    """
+    with open(THIS_PATH / "templates/templating.yaml", "r") as f:
+        template_config = yaml.safe_load(f) or {}
```
ANich, line 43: nit: Wouldn't you want to error if the path doesn't exist?
```suggestion
        template_config = yaml.safe_load(f)
```

sql_generators/review_checker/__init__.py
ANich, line 1: Overall comment - since there's only one application, why make this a `sql_generator`? Seems like this could be directly checked in queries

sql_generators/review_checker/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+                )
+            )
+
+            write_sql(
+                output_dir=output_dir,
+                full_table_id=f"{target_project}.{browser.name}_derived.{query}_v1",
+                basename="query.sql",
+                sql=query_sql,
+                skip_existing=False,
+            )
```
ANich, line 68: I think you could call something from the schema module here to write the `schema.yaml` files we discussed offline

sql_generators/review_checker/templates/review_checker_microsurvey/query.sql
```diff
@@ -0,0 +1,103 @@
+WITH new_and_existing_users AS (
+  SELECT
+    submission_date,
+    client_id,
+    IF(DATE_DIFF(submission_date, first_seen_date, DAY) <= 27, 1, 0) AS new_user,
+    IF(DATE_DIFF(submission_date, first_seen_date, DAY) > 27, 1, 0) AS existing_user
+  FROM
+    `moz-fx-data-shared-prod.telemetry.clients_last_seen`
+  WHERE
```
ANich, line 9: FYI that this doesn't contain all new clients - see `clients_first_seen_v2`. Might be fine depending on your use-case

ANich (approved):
Left a few brief comments. Main question is whether this needs to be a sql generator why not just check in SQL? LGTM overall

sql_generators/review_checker/__init__.py
alekhyamoz: We will be adding mobile in the next iteration and the query be very similar to this.

sql_generators/review_checker/templates/review_checker_microsurvey/query.sql
alekhyamoz: The table is `clients_last_seen` and I do not see a `v2` of this. My rationale behind using a view is that even if underlying table change the table will remain the same.

PR #4469, opened 2023-10-24
Added a filter to only include playstore data

soGaussian:
In keeping the bottom of the funnel consistent with the upper funnel, we have to only include installs from play store in the bottom of the funnel metrics

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1849)

### DIFF
--- a/sql/moz-fx-data-shared-prod/fenix_derived/funnel_retention_clients_week_4_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/fenix_derived/funnel_retention_clients_week_4_v1/query.sql
@@ -28,6 +28,7 @@ clients_first_seen AS (
     -- 28 days need to elapse before calculating the week 4 and day 28 retention metrics
     first_seen_date = DATE_SUB(@submission_date, INTERVAL 27 DAY)
     AND channel = "release"
+    AND install_source = 'com.android.vending'
 ),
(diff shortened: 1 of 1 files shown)


kik-kik:
@soGaussian if this filter is needed, should we also include it in [funnel_retention_clients_week_2_v1](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/fenix_derived/funnel_retention_clients_week_2_v1/query.sql)?

soGaussian:
So the week 2 table is not used in the funnel and so I cannot say. but we should expose the filter there as well. I believe the consensus was not to filter it out but to expose the column in the final table. Is this accurate @gkabbz

kik-kik:
> So the week 2 table is not used in the funnel and so I cannot say. but we should expose the filter there as well. I believe the consensus was not to filter it out but to expose the column in the final table. Is this accurate @gkabbz

Actually, it is by the downstream view @soGaussian. We should keep both of those queries in sync. Eventually, we can create some sort of "generator" to auto-generate these queries for each time frame we want.

kik-kik:
Superseded by https://github.com/mozilla/bigquery-etl/pull/4561

PR #4481, opened 2023-10-25
 Bug 1860814 -  fix amo_prod__desktop_addons_by_client

lelilia:
[Bug 1860814](https://bugzilla.mozilla.org/show_bug.cgi?id=1860814)

The problem is that we have one very active customer_id with lot's of entries in main 
with the previous code all the copies for the addons that come with every row needed to be aggregated.
By adding a step in the ETL where I only get the distinct addons entries for a customer_id the aggregation in the next step is not so big that it exceeds the query limits any longer 

This is not the most elegant solution, so I'm very open to feedback on how to achieve the same result in a better way.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1864)

### DIFF
--- a/sql/moz-fx-data-shared-prod/amo_prod/desktop_addons_by_client_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/amo_prod/desktop_addons_by_client_v1/query.sql
@@ -15,26 +15,12 @@ try {
 }
 """;
 
-WITH per_client AS (
+WITH cleaned_main AS (
   SELECT
     DATE(submission_timestamp) AS submission_date,
     client_id,
     sample_id,
-    ARRAY_CONCAT_AGG(
-      -- As of 2020-07-17, we parse addon info from the payload.info.addons string
-      -- because it provides more complete information compared to environment.addons.active_addons
-      -- See https://bugzilla.mozilla.org/show_bug.cgi?id=1653499
-      -- Previous logic in https://github.com/mozilla/bigquery-etl/blob/55a429924beee3c31aca4fee0063d655f1d527f2/sql/amo_prod/desktop_addons_by_client_v1/query.sql
-      ARRAY(
-        SELECT AS STRUCT
-          decode_uri_component(REGEXP_EXTRACT(TRIM(addon), "(.+):")) AS id,
-          REGEXP_EXTRACT(TRIM(addon), ":(.+)") AS version
-        FROM
-          UNNEST(SPLIT(payload.info.addons, ",")) AS addon
-      )
-      ORDER BY
-        submission_timestamp
-    ) AS addons,
+    payload.info.addons,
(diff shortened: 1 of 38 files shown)


Iinh (commented):
Considering that the DAG failed due to resource constraints, rewriting the query this way to fix the failure makes sense to me. However, because this is a tier-1 DAG and I'm not familiar with it, I'd appreciate it if another engineer with more context could also review these changes before we ship.

sql/moz-fx-data-shared-prod/amo_prod/desktop_addons_by_client_v1/query.sql
```diff
@@ -57,7 +43,41 @@ WITH per_client AS (
   GROUP BY
     submission_date,
     sample_id,
-    client_id
+    client_id,
+    addons
+),
+per_client AS (
```
fbertsch, line 49: Unfortunately this is not per_client, it is per_client/app-version/country/locale/app_os. If a user has pings with different values for e.g. country _and_ different `payload.info.addons`, we'll end up with multiple rows for that client in this CTE.

Is there any way we can instead pull this information from `clients_daily_v6`? That table has addon information, and probably already has this other data aggregated.

fbertsch (changes requested):

sql/moz-fx-data-shared-prod/amo_prod/desktop_addons_by_client_v1/query.sql
lelilia: oh you are right ... how about moving that step out and joining it later? 
The problem with `clients_daily_v6` would be that they get the addons data from environment.addons while this query gets it from `payload.info.addons` (which is deprecated as of now) but they values are not the same and so switching would change the outcome ... which we might want to do at some point but maybe not for a triage fix

sql/moz-fx-data-shared-prod/amo_prod/desktop_addons_by_client_v1/query.sql
```diff
@@ -58,6 +43,65 @@ WITH per_client AS (
...
+  SELECT
+    DATE(submission_timestamp) AS submission_date,
+    client_id,
+    sample_id,
+    payload.info.addons
+  FROM
+    telemetry.main
+  WHERE
+    DATE(submission_timestamp) = @submission_date
+    AND client_id IS NOT NULL
```
fbertsch, line 57: One nit: I'd move this `FROM` and `WHERE` clause to a new CTE that is queried by both `per_clients_just_addons_base` and `per_clients_without_addons`. That way we only define the filter/source once.

fbertsch (approved):
This LGTM, thank you!

PR #4473, opened 2023-10-24
DENG-1705 - Add startup_profile_selection_reason_first to clients_daily_v6

ANich:
See https://github.com/mozilla/bigquery-etl/pull/3962#discussion_r1332186267 - get the first value for a client's `startup_profile_selection_reason`

PR #4482, opened 2023-10-25
DENG-1705 - Add startup_profile_selection_reason from first ping to clients_daily, clients_first_seen_v2 and downstream

ANich:
Follow up to https://github.com/mozilla/bigquery-etl/pull/4473. Small refactor to cfs_28_days_later

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1865)

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
@@ -1453,6 +1453,9 @@ aggregates AS (
     ARRAY_AGG(startup_profile_selection_reason ORDER BY submission_timestamp ASC)[
       OFFSET(0)
     ] AS startup_profile_selection_reason_first,
+    mozfun.stats.mode_last(
+        ARRAY_AGG(IF(subsession_counter = 1, startup_profile_selection_reason, NULL) ORDER BY submission_timestamp)
+    ) as startup_profile_selection_first_ping_only,
     SUM(
       scalar_parent_browser_ui_interaction_textrecognition_error
     ) AS scalar_parent_browser_ui_interaction_textrecognition_error_sum,
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
@@ -1454,7 +1454,7 @@ aggregates AS (
       OFFSET(0)
     ] AS startup_profile_selection_reason_first,
     mozfun.stats.mode_last(
-        ARRAY_AGG(IF(subsession_counter = 1, startup_profile_selection_reason, NULL) ORDER BY submission_timestamp)
+        ARRAY_AGG(IF(subsession_counter = 1, startup_profile_selection_reason, NULL) ORDER BY submission_timestamp ASC)
     ) as startup_profile_selection_first_ping_only,
     SUM(
       scalar_parent_browser_ui_interaction_textrecognition_error
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_28_days_later_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_28_days_later_v1/query.sql
@@ -16,7 +16,7 @@ WITH clients_first_seen_28_days_ago AS (
     attribution_experiment,
     attribution_dltoken,
     attribution_dlsource,
-    -- startup_profile_selection_reason, when startup_profile_selection_reason is available
+    startup_profile_selection_reason,
     first_seen_date,
   FROM
     telemetry_derived.clients_first_seen_v2
@@ -38,24 +38,20 @@ clients_first_seen_28_days_ago_with_days_seen AS (
     AND cls.submission_date = @submission_date
 )
 SELECT
-  client_id,
-  sample_id,
-  first_seen_date,
-  @submission_date AS submission_date,
-  country_code,
-  channel,
-  build_id,
-  os,
-  os_version,
-  distribution_id,
-  attribution_source,
-  attribution_ua,
-  attribution_medium,
-  attribution_campaign,
-  attribution_content,
-  attribution_experiment,
-  attribution_dltoken,
-  attribution_dlsource,
+  * REPLACE (
+    COALESCE(
+      days_seen_bits,
+      mozfun.bits28.from_string('0000000000000000000000000000')
+    ) AS days_seen_bits,
+    COALESCE(
+      days_visited_1_uri_bits,
+      mozfun.bits28.from_string('0000000000000000000000000000')
+    ) AS days_visited_1_uri_bits,
+    COALESCE(
+      days_interacted_bits,
+      mozfun.bits28.from_string('0000000000000000000000000000')
+    ) AS days_interacted_bits
+  ),
   COALESCE(
     BIT_COUNT(mozfun.bits28.from_string('1111111000000000000000000000') & days_seen_bits) >= 5,
     FALSE
@@ -84,17 +80,6 @@ SELECT
     ) > 0,
     FALSE
   ) AS qualified_week4,
-  COALESCE(
-    days_seen_bits,
-    mozfun.bits28.from_string('0000000000000000000000000000')
-  ) AS days_seen_bits,
-  COALESCE(
-    days_visited_1_uri_bits,
-    mozfun.bits28.from_string('0000000000000000000000000000')
-  ) AS days_visited_1_uri_bits,
-  COALESCE(
-    days_interacted_bits,
-    mozfun.bits28.from_string('0000000000000000000000000000')
-  ) AS days_interacted_bits,
+  @submission_date AS submission_date,
 FROM
   clients_first_seen_28_days_ago_with_days_seen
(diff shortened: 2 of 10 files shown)


tests/sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/moz-fx-data-shared-prod.telemetry_derived.clients_daily_v6.schema.json
```diff
@@ -112,5 +112,9 @@
   {
     "type": "STRING",
     "name": "normalized_os_version"
+  },
```
irrationalagent, line 115: are you going to be adding the actual colum to clients_daily in this PR too? if so, how we aggregate still probably needs to be discussed. we previously said to take the value from the first main ping, but based on @SuYoungHong 's recent analysis, we might want to consider just checking whether any ping has ` startup_profile_selection_reason_first = 'firstrun-created-default'`. Because the first arriving ping might not be the first _generated_ ping. TBD

irrationalagent: see discussion in this [gdoc comment](https://docs.google.com/document/d/1mG43ZLMcGcLBnJEPw7mxPIQHGvp_S8T56pnT6lvSXXo/edit?disco=AAAA96heEEM)

ANich: `clients_daily` was updated in https://github.com/mozilla/bigquery-etl/pull/4473, I'll follow up in the doc

SuYoungHong (changes requested):
Hey, @ANich I just reviewed the changes, and there's a few things I wanted to flag. 

1) [Blocking Issue] so this PR is adding `startup_profile_selection_reason` for new profiles sourced from Clients Daily, but looking at the code, I realize now that we'll also need to do the same (add and backfill) the `environment.settings.attribution.ua` field. 

Looks like the code is currently filling nulls for clients sourced from Clients Daily (presumably because Clients Daily doesn't have it). [Link to code](https://github.com/mozilla/bigquery-etl/blob/97882f729cd9b319fafd2b90d8d2789037fbdc4b/sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql#L156)

My apologies, I should have made that requirements more clear somewhere. 
But the Numbers that Matter / moz.org Funnel V1.2 refresh **needs** that field. 

2) [Non-Blocking Issue] Similar to above, it looks like the current code is filling nulls in a number of places for clients sourced from Clients Daily. 
```
architecture
platform_version
xpcom_abi
document_id
partner_distribution_version
partner_distributor
partner_distributor_channel
partner_id
apple_model_id
db_version
attribution_dlsource
```
Lines [136](https://github.com/mozilla/bigquery-etl/blob/97882f729cd9b319fafd2b90d8d2789037fbdc4b/sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql#L136) to [171](https://github.com/mozilla/bigquery-etl/blob/97882f729cd9b319fafd2b90d8d2789037fbdc4b/sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql#L171)

I don't think any of these fields are that important (with the exception of `attribution_dlsource`, I anticipate that field will be important for future analysis going forward), but if we're going to do a backfill, it seems sensible to get the rest of these as well. 

That is, unless we have some plans to switch to a design where we're bypassing Clients Daily, and reading straight from Main Ping on the roadmap. 

But something to think about. 

3) [Non-Blocking Issue] So there's one last thing I noticed... and I'm not sure this PR is the right place to address it, but I wanted to get it on your radar. 

The `normalized_os` in Clients First Seen V2 is inconsistent. What I mean by that is that, for clients sourced from the New Profile Pings or First Shutdown Pings, the values in that field are slightly different from those sourced from Clients Daily. In the code. 

For example, windows are reported as "Windows_NT" in Clients Daily, while NPP and FSP reports it as "Windows". See: [Example: Inconsistent normalized_os in Clients First Seen V2](https://sql.telemetry.mozilla.org/queries/95569/source)

I think it's because in NPP and FSP, we're pulling from `normalized_os` while in Clients Daily, looks like we're pulling from just `os`. Might be a simple switch. 

In any case, it's not a blocker for the Numbers that Matter / moz.org Funnel V1.2 refresh because I can work around it in the query, but it's probably worth addressing if we have the bandwidth. 

Again, this seems like an issue that arises out of looking directly at pings in some cases, and at a processed dowstream dataset like Clients Daily in other cases. If we have any plans to switch to read directly from the raw pings for Main Ping cases, that would make a lot of these problems go away I think. 

Food for thought.

irrationalagent:
> I think it's because in NPP and FSP, we're pulling from normalized_os while in Clients Daily, looks like we're pulling from just os. Might be a simple switch.

Annoyingly, clients_daily does not have normalized_os but we can pass it through [this udf](https://github.com/mozilla/bigquery-etl/blob/main/sql/mozfun/norm/os/udf.sql) so that the results match the new profile and fsd pings. which I think is a good idea

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_28_days_later_v1/query.sql
```diff
@@ -38,24 +38,20 @@ clients_first_seen_28_days_ago_with_days_seen AS (
...
-  attribution_medium,
-  attribution_campaign,
-  attribution_content,
-  attribution_experiment,
-  attribution_dltoken,
-  attribution_dlsource,
+  * REPLACE (
+    COALESCE(
+      days_seen_bits,
+      mozfun.bits28.from_string('0000000000000000000000000000')
```
fbertsch, line 44: Do you do it this way to make it clear that this is a zero-activity-days bit pattern? Over e.g. `0`

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -374,9 +374,9 @@ main_ping AS (
     ARRAY_AGG(normalized_os_version RESPECT NULLS ORDER BY submission_date)[
       SAFE_OFFSET(0)
     ] AS normalized_os_version,
-    CAST(
-      NULL AS STRING
-    ) AS startup_profile_selection_reason, -- main_v5:payload.processes.parent.scalars.startup_profile_selection_reason
+    ARRAY_AGG(startup_profile_selection_reason_first RESPECT NULLS ORDER BY submission_date)[
+      SAFE_OFFSET(0)
+    ] AS startup_profile_selection_reason,
```
fbertsch, line 379: Given the discussion in the doc, should we call this
```suggestion
    ] AS first_reported_startup_profile_selection_reason,
```

fbertsch (approved):
This LGTM for this field, but does it cover what was reported in the doc? Should we add the boolean field you also recommended?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
ANich: I think that applies to all columns in this table no? They should all be the first reported value

fbertsch: hah, I guess so. I would have said to name that all that way if I had realized

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_28_days_later_v1/schema.yaml
```diff
@@ -85,3 +85,7 @@ fields:
 - name: days_interacted_bits
   type: INTEGER
   mode: NULLABLE
+- name: startup_profile_selection_reason
+  type: STRING
+  mode: NULLABLE
+  description: Client's first startup_profile_selection reason.
```
lucia-vargas-a, line 91: This column is fully described in main_v5, however since this is higher level table, the users could benefit from a more explanatory description displayed in the data catalog, e.g.
```suggestion
  description: How the profile was selected during startup, as reported by the first ping received.
```

lucia-vargas-a (approved):

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
```diff
@@ -1453,6 +1453,9 @@ aggregates AS (
     ARRAY_AGG(startup_profile_selection_reason ORDER BY submission_timestamp ASC)[
       OFFSET(0)
     ] AS startup_profile_selection_reason_first,
+    mozfun.stats.mode_last(
```
irrationalagent, line 1456: 👍  There should be very very few cases where we receive MULTIPLE pings with subsession_counter = 1 from a single client_id. But in those cases, if maybe one of them has the startup_profile_selection_reason we are looking for (while maybe 2+ others don't), we will miss it. 

These would be weird clients that we possibly would want to ignore for other reasons, so I'm not advocating for any change here, just noting what the behavior will be for the record.

irrationalagent (approved):

PR #4503, opened 2023-11-01
Update scheduler of aggregates to run after upstreams.

lucia-vargas-a:
The DAG bqetl_firefox_ios is scheduled after its downstream dependency bqetl_analytics_aggregations which results in the aggregates not being updated and set in status 'wait for retry'. This PR updates the scheduler for the aggregates to run after the upstream instead of running the upstream earlier. It also updates the search_dashboard to align.

image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1895)


kik-kik:
It appears the downstream DAG is correctly waiting for the firefox_ios task to complete before running?
image


We can change the schedule, but I'm not sure it actually changes anything?

kik-kik (approved):
Approving, but I'm not sure this is really necessary as the dependency between tasks appears to be correctly defined inside Airflow. The aggregates task waits for the firefox_ios task complete before running.

lucia-vargas-a:
> Approving, but I'm not sure this is really necessary as the dependency between tasks appears to be correctly defined inside Airflow. The aggregates task waits for the firefox_ios task complete before running.

It is meant to solve the aggregates not being updated:

image

PR #4505, opened 2023-11-01
DENG-1705 - Add missing client attribution columns to clients daily/first-seen

ANich:
Followup to https://github.com/mozilla/bigquery-etl/pull/4482#pullrequestreview-1707538782 

Adds missing columns to clients daily and first-seen. Sample data from beta here: `moz-fx-data-shared-prod.tmp.anich_clients_daily_1705_test`, `moz-fx-data-shared-prod.tmp.anich_clients_first_seen_v2_test_1705`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1897)

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
@@ -177,6 +177,7 @@ clients_summary AS (
     submission_timestamp,
     client_id,
     sample_id,
+    document_id,
     metadata.uri.app_update_channel AS channel,
     normalized_channel,
     normalized_os_version,
@@ -184,6 +185,7 @@ clients_summary AS (
     metadata.geo.city,
     metadata.geo.subdivision1 AS geo_subdivision1,
     metadata.geo.subdivision2 AS geo_subdivision2,
+    metadata.geo.db_version AS geo_db_version,
     metadata.isp.name AS isp_name,
     metadata.isp.organization AS isp_organization,
     environment.system.os.name AS os,
@@ -194,6 +196,7 @@ clients_summary AS (
     SAFE_CAST(environment.system.os.windows_ubr AS INT64) AS windows_ubr,
     SAFE_CAST(environment.system.os.install_year AS INT64) AS install_year,
     environment.system.is_wow64,
+    environment.system.apple_model_id,
     SAFE_CAST(environment.system.memory_mb AS INT64) AS memory_mb,
     environment.system.cpu.count AS cpu_count,
     environment.system.cpu.cores AS cpu_cores,
@@ -214,6 +217,10 @@ clients_summary AS (
     payload.info.subsession_counter,
     payload.info.subsession_length,
     environment.partner.distribution_id,
+    environment.partner.partner_id,
+    environment.partner.distribution_version,
+    environment.partner.distributor,
+    environment.partner.distributor_channel,
     IFNULL(
       environment.services.account_enabled,
       udf.boolean_histogram_to_boolean(payload.histograms.fxa_configured)
@@ -235,6 +242,8 @@ clients_summary AS (
     environment.build.build_id AS env_build_id,
     environment.build.version AS env_build_version,
     environment.build.architecture AS env_build_arch,
+    environment.build.platform_version AS env_build_platform_version,
+    environment.build.xpcom_abi AS env_build_xpcom_abi,
     environment.settings.e10s_enabled,
     environment.settings.locale,
     environment.settings.update.channel AS update_channel,
@@ -251,7 +260,8 @@ clients_summary AS (
         environment.settings.attribution.experiment,
         environment.settings.attribution.variation,
         environment.settings.attribution.dltoken,
-        environment.settings.attribution.dlsource
+        environment.settings.attribution.dlsource,
+        environment.settings.attribution.ua
       ),
       NULL
     ) AS attribution,
@@ -686,6 +696,7 @@ aggregates AS (
   SELECT
     DATE(submission_timestamp) AS submission_date,
     client_id,
+    ARRAY_AGG(document_id ORDER BY submission_timestamp)[OFFSET(0)] AS first_document_id,
     SUM(aborts_content) AS aborts_content_sum,
     SUM(aborts_gmplugin) AS aborts_gmplugin_sum,
     SUM(aborts_plugin) AS aborts_plugin_sum,
@@ -772,10 +783,20 @@ aggregates AS (
     mozfun.stats.mode_last(
       ARRAY_AGG(distribution_id ORDER BY submission_timestamp)
     ) AS distribution_id,
+    mozfun.stats.mode_last(ARRAY_AGG(partner_id ORDER BY submission_timestamp)) AS partner_id,
+    mozfun.stats.mode_last(ARRAY_AGG(distribution_version ORDER BY submission_timestamp)) AS distribution_version,
+    mozfun.stats.mode_last(ARRAY_AGG(distributor ORDER BY submission_timestamp)) AS distributor,
+    mozfun.stats.mode_last(ARRAY_AGG(distributor_channel ORDER BY submission_timestamp)) AS distributor_channel,
     mozfun.stats.mode_last(ARRAY_AGG(e10s_enabled ORDER BY submission_timestamp)) AS e10s_enabled,
     mozfun.stats.mode_last(
       ARRAY_AGG(env_build_arch ORDER BY submission_timestamp)
     ) AS env_build_arch,
+    mozfun.stats.mode_last(
+      ARRAY_AGG(env_build_platform_version ORDER BY submission_timestamp)
+    ) AS env_build_platform_version,
+    mozfun.stats.mode_last(
+      ARRAY_AGG(env_build_xpcom_abi ORDER BY submission_timestamp)
+    ) AS env_build_xpcom_abi,
     mozfun.stats.mode_last(ARRAY_AGG(env_build_id ORDER BY submission_timestamp)) AS env_build_id,
     mozfun.stats.mode_last(
       ARRAY_AGG(env_build_version ORDER BY submission_timestamp)
@@ -858,6 +879,7 @@ aggregates AS (
           submission_timestamp
       )
     ).*,
+    mozfun.stats.mode_last(ARRAY_AGG(geo_db_version ORDER BY submission_timestamp)) AS geo_db_version,
     mozfun.json.mode_last(
       ARRAY_AGG(
         IF(
@@ -978,6 +1000,7 @@ aggregates AS (
       ARRAY_AGG(is_default_browser ORDER BY submission_timestamp)
     ) AS is_default_browser,
     mozfun.stats.mode_last(ARRAY_AGG(is_wow64 ORDER BY submission_timestamp)) AS is_wow64,
+    mozfun.stats.mode_last(ARRAY_AGG(apple_model_id ORDER BY submission_timestamp)) AS apple_model_id,
     mozfun.stats.mode_last(ARRAY_AGG(locale ORDER BY submission_timestamp)) AS locale,
     mozfun.stats.mode_last(ARRAY_AGG(memory_mb ORDER BY submission_timestamp)) AS memory_mb,
     mozfun.stats.mode_last(
@@ -1135,6 +1158,8 @@ aggregates AS (
     udf.aggregate_search_counts(ARRAY_CONCAT_AGG(search_counts ORDER BY submission_timestamp)).*,
     AVG(session_restored) AS session_restored_mean,
     COUNTIF(subsession_counter = 1) AS sessions_started_on_this_day,
+    MAX(subsession_counter)AS max_subsession_counter,
+    MIN(subsession_counter)AS min_subsession_counter,
     SUM(shutdown_kill) AS shutdown_kill_sum,
     SUM(subsession_length / NUMERIC '3600') AS subsession_hours_sum,
     SUM(ssl_handshake_result_failure) AS ssl_handshake_result_failure_sum,
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/schema.yaml
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/schema.yaml
@@ -154,6 +154,9 @@ fields:
   - name: dlsource
     type: STRING
     mode: NULLABLE
+  - name: ua
+    type: STRING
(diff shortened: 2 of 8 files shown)


sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
```diff
@@ -772,10 +783,20 @@ aggregates AS (
     mozfun.stats.mode_last(
       ARRAY_AGG(distribution_id ORDER BY submission_timestamp)
     ) AS distribution_id,
+    mozfun.stats.mode_last(ARRAY_AGG(partner_id ORDER BY submission_timestamp)) AS partner_id,
+    mozfun.stats.mode_last(ARRAY_AGG(distribution_version ORDER BY submission_timestamp)) AS distribution_version,
+    mozfun.stats.mode_last(ARRAY_AGG(distributor ORDER BY submission_timestamp)) AS distributor,
+    mozfun.stats.mode_last(ARRAY_AGG(distributor_channel ORDER BY submission_timestamp)) AS distributor_channel,
```
ANich, line 789: Used mode here similar to existing, similar columns.

irrationalagent:
@SuYoungHong would it all make sense to add any or all the following columns in addition (at least to clients_daily, it might not make sense to add to clients_first_seen since the other pings dont have these fields)

* `min_subsession_counter` 
* `max_subsession_counter`

Thinking about whether these would help at all to flag any potential client_id regens or forks. For example, if the first field is >> 1 on first_seen_date, it might be convenient to have in clients_daily. Wasn't sure though.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/schema.yaml
```diff
@@ -2303,3 +2306,30 @@ fields:
...
+- name: distribution_version
+  type: STRING
+  mode: NULLABLE
+- name: distributor
+  type: STRING
+  mode: NULLABLE
+- name: distributor_channel
+  type: STRING
+  mode: NULLABLE
+- name: env_build_platform_version
```
lucia-vargas-a, line 2324: I notice a different naming convention e.g. `environment.partner.partner_id` is named `partner_id`, while `environment.build.xpcom_abi` is named `env_build_xpcom_abi` which might cause confusion to users and also differs from the naming in v2. It makes sense to me to keep the naming without the suffix for consistency e.g. `xpcom_abi` instead of `env_build_xpcom_abi`, wdyt?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
```diff
@@ -235,6 +242,8 @@ clients_summary AS (
     environment.build.build_id AS env_build_id,
     environment.build.version AS env_build_version,
     environment.build.architecture AS env_build_arch,
+    environment.build.platform_version AS env_build_platform_version,
+    environment.build.xpcom_abi AS env_build_xpcom_abi,
```
lucia-vargas-a, line 246: For most fields, the decision was to use `environment` instead of `application` as the source. However I  notice that the source in `new_profile` and `first_shutdown` pings for these fields is `application.xpcom_abi`. Worth a double check that this doesn't result would result in inconsistent data?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
```diff
@@ -235,6 +242,8 @@ clients_summary AS (
     environment.build.build_id AS env_build_id,
     environment.build.version AS env_build_version,
     environment.build.architecture AS env_build_arch,
+    environment.build.platform_version AS env_build_platform_version,
```
lucia-vargas-a, line 245: Same case of xpcom_abi, just a double check that data from environment.build (main ping) is consistent with data from application (new_profile and first_shutdown pings)

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/schema.yaml
```diff
@@ -2303,3 +2306,30 @@ fields:
...
+  type: STRING
+  mode: NULLABLE
+- name: env_build_xpcom_abi
```
lucia-vargas-a, line 2327: Same case of env_build_platform_version about naming convention.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
```diff
@@ -196,14 +196,14 @@ unioned AS (
   SELECT
     *,
     'shutdown' AS source_ping,
-    3 AS source_ping_priority
+    2 AS source_ping_priority
```
lucia-vargas-a, line 199: Since this priority changes and differs from what's in the requirements & caveats docs, could we update the [inline comment](https://github.com/mozilla/bigquery-etl/blob/dffbb3afef5a3b0b5adc396fa4b874bc0f617b6b/sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql#L187) to explain the reasoning for future reference?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v1/schema.yaml
```diff
@@ -2305,3 +2308,30 @@ fields:
...
+- name: distribution_version
+  type: STRING
+  mode: NULLABLE
+- name: distributor
+  type: STRING
+  mode: NULLABLE
+- name: distributor_channel
+  type: STRING
+  mode: NULLABLE
+- name: env_build_platform_version
```
lucia-vargas-a, line 2326: Same comment about naming, which also differs from v2. It might be confusing for users.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v1/schema.yaml
```diff
@@ -2305,3 +2308,30 @@ fields:
...
+  type: STRING
+  mode: NULLABLE
+- name: env_build_xpcom_abi
```
lucia-vargas-a, line 2329: Same comment about naming as above.

lucia-vargas-a (commented):
Thank you for integrating of missing fields. Shall we also update in this PR the schema for [`telemetry_derived.clients_last_seen_joined_v2`](https://github.com/mozilla/bigquery-etl/blob/7a371d1ce3639877ecd744b7bbcc9e65863b92c9/sql/moz-fx-data-shared-prod/telemetry_derived/clients_last_seen_joined_v1/query.sql#L4)?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
ANich: I ran some queries looking for instances where they're different and didn't get any results at least for recent 2023 data

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
ANich: Will update 👍

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/schema.yaml
lucia-vargas-a: Yeah, some prefix and some don't e.g. partner_id, is_wow64

ANich: Those are from `environment.partner` and `environment.system` not `environment.build` so doesn't seem like there was a previous convention

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
ANich: Update, following up on this offline this change might not be needed

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/schema.yaml
```diff
@@ -2303,3 +2306,30 @@ fields:
 - name: startup_profile_selection_reason_first
   type: STRING
   mode: NULLABLE
+- name: first_document_id
+  type: STRING
+  mode: NULLABLE
+- name: partner_id
+  type: STRING
+  mode: NULLABLE
+- name: distribution_version
```
fbertsch, line 2315: It would be lovely to have descriptions for these

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql
```diff
@@ -858,6 +879,7 @@ aggregates AS (
           submission_timestamp
       )
     ).*,
+    mozfun.stats.mode_last(ARRAY_AGG(geo_db_version ORDER BY submission_timestamp)) AS geo_db_version,
```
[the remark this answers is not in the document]
ANich: I can't say, this is used downstream in `clients_first_seen_v2` cc @lucia-vargas-a

SuYoungHong:
hey, one question I had about this: 

I noticed that in [bigquery-etl/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql ](https://github.com/mozilla/bigquery-etl/blob/b7f199fe643d89bee05ae006e65d5b10e90b0c6a/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql), when we aggregate the `attribution` info and the `startup_profile_selection_reason` field, we're taking the first value by `submission_timestamp` (which is fine). 

But do these respect nulls (if the first ping had null for these values, do we get null back, or does it return first non-null)?

Code for reference: 
```
ARRAY_AGG(startup_profile_selection_reason ORDER BY submission_timestamp ASC)[
      OFFSET(0)
    ] AS startup_profile_selection_reason_first,
```
and 
```
mozfun.json.mode_last(ARRAY_AGG(attribution ORDER BY submission_timestamp)) AS attribution,
```

SuYoungHong:
@irrationalagent , yes, I think those fields would absolutely be helpful
* `min_subsession_counter`
* `max_subsession_counter`

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/schema.yaml
lucia-vargas-a: Unfortunately this leads to an unnecessary complication of needing to update downstream queries when cascading the changes due to different naming in v1 and v2. It'd be so much better if we align the naming between them in this PR rather than updating the schema or expanding the queries later, see e.g. [platfform_version](https://github.com/mozilla/bigquery-etl/blob/b7f199fe643d89bee05ae006e65d5b10e90b0c6a/sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/schema.yaml#L34C25-L34C25)

ANich:
> hey, one question I had about this:
> 
> I noticed that in [bigquery-etl/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql ](https://github.com/mozilla/bigquery-etl/blob/b7f199fe643d89bee05ae006e65d5b10e90b0c6a/sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/query.sql), when we aggregate the `attribution` info and the `startup_profile_selection_reason` field, we're taking the first value by `submission_timestamp` (which is fine).
> 
> But do these respect nulls (if the first ping had null for these values, do we get null back, or does it return first non-null)?
> 
> Code for reference:
> 
> ```
> ARRAY_AGG(startup_profile_selection_reason ORDER BY submission_timestamp ASC)[
>       OFFSET(0)
>     ] AS startup_profile_selection_reason_first,
> ```
> 
> and
> 
> ```
> mozfun.json.mode_last(ARRAY_AGG(attribution ORDER BY submission_timestamp)) AS attribution,
> ```

`ARRAY_AGG(...)[OFFSET(0)]` respects nulls by default so yeah return null if the first ping had null. 

`mode_last` only returns null if there's all nulls. It doesn't actually return the first value by `submission_timestamp`, it gets the most frequent value and latest value in case of ties. If only the "real" first ping (by subsession/submission_timestamp) has `environment.settings.attribution` data then we're in a better spot than usual since that would be the only non-null.

ANich:
@irrationalagent @SuYoungHong there's currently this field: `COUNTIF(subsession_counter = 1) AS sessions_started_on_this_day` - is this sufficient? If a client has 0 sessions started that day using this definition, then it's a potential regen?

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/schema.yaml
ANich: I don't follow. The schemas for `clients_first_seen_v2` and `clients_first_seen_v1` are already very different. Keeping the convention makes sense for the same reason we wouldn't change the upstream `env_build_arch`, `default_search_engine_data_load_path`, or `geo_subdivision1` to match the downstream column names.

irrationalagent:
> @irrationalagent @SuYoungHong there's currently this field: `COUNTIF(subsession_counter = 1) AS sessions_started_on_this_day` - is this sufficient? If a client has 0 sessions started that day using this definition, then it's a potential regen?

I think it would be true that it would be a _potential_ regen, but I think it could also be caused by the ping ordering issue (where a main-ping-first client's first main ping is from a subsequent, usually second, subsession). If we had the precise number then we could be more confident its a regen specifically.

sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
ANich: I'll revert and fix this in a separate PR according to the discussion [here](https://docs.google.com/document/d/1LmfuThLp856-W0HDxyH9Lco4piVKW_6ok4mfnBnXOl4/edit?disco=AAAA_MAPsZw).

irrationalagent (approved):
alright, LGTM

sql/moz-fx-data-shared-prod/telemetry_derived/clients_daily_v6/schema.yaml
ANich: I agree as you know, but in the case of `clients_daily` which currently has.. zero descriptions I'm slightly more 🤷‍♂️

PR #4533, opened 2023-11-07
DENG-1935 Change data ordering from pings in clients-first-seen-v2

ANich:
Fixes the ping ordering issue described [here](https://docs.google.com/document/d/1LmfuThLp856-W0HDxyH9Lco4piVKW_6ok4mfnBnXOl4/edit). 

Using the heuristic from the doc:
```
On the first day that a client_id appears for the first time: 
- Use the New Profile Ping as the data source if present
- If not, use the first First Shutdown Ping if present
- If not, use the first (based profile_subsession_counter) Main Ping
```  

Except for the last point - the source table is currently `clients_daily` which doesn't differentiate between the main ping based on `profile_subsession_counter`. Though fields for debugging this are being added in https://github.com/mozilla/bigquery-etl/pull/4505

Checklist for reviewer:


tests/sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/test_source_pings/moz-fx-data-shared-prod.telemetry.first_shutdown.yaml
```diff
@@ -1,4 +1,4 @@
 ---
 - client_id: client-3
   sample_id: 30
-  submission_timestamp: '2023-06-01 11:03:00'
+  submission_timestamp: '2023-06-01 10:03:00'
```
ANich, line 4: Changed these so the `new_profile` ping arrives second but is still chosen over the `first_shutdown` ping

fbertsch (approved):
LGTM, but it's a good idea to add a regression test for this change - a client who reports both a `main` ping and a `first_shutdown`, and takes the `first_shutdown` ping value. Could stick it alongside the `test_seen_dates` test.

PR #4547, opened 2023-11-10
# bug(1859232): removed check for firefox_ios_clients_v1 which used different filtering settings causing result mismatch

kik-kik:
# bug(1859232): removed check for firefox_ios_clients_v1 which used different filtering settings causing result mismatch

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1955)

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_ios_derived/firefox_ios_clients_v1/checks.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_ios_derived/firefox_ios_clients_v1/checks.sql
@@ -1,9 +1,9 @@
 #fail
 {{ is_unique(columns=["client_id"]) }}
 #fail
-{{ not_null(columns=["client_id"]) }}
+{{ not_null(columns=["client_id"], where="first_seen_date = @submission_date") }}
 #fail
-{{ min_row_count(1, "first_seen_date = @submission_date") }}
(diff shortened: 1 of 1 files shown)


sql/moz-fx-data-shared-prod/firefox_ios_derived/firefox_ios_clients_v1/checks.sql
```diff
@@ -1,9 +1,9 @@
 #fail
 {{ is_unique(columns=["client_id"]) }}
 #fail
-{{ not_null(columns=["client_id"]) }}
+{{ not_null(columns=["client_id"], where="first_seen_date = @submission_date") }}
 #fail
-{{ min_row_count(1, "first_seen_date = @submission_date") }}
+{{ min_row_count(1, where="first_seen_date = @submission_date") }}
```
akkomar, line 6: Was the SQL generated before this change correct?

akkomar (approved):

sql/moz-fx-data-shared-prod/firefox_ios_derived/firefox_ios_clients_v1/checks.sql
kik-kik: yes, I noticed in the next check jinja error would arise when omitting the `where` so decided to add it to this one explicitly too.

PR #4554, opened 2023-11-13
Add ga_sessions_v1 table & view

fbertsch:
This table aggregates session-level data from GA.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1963)

### DIFF
--- a/sql/moz-fx-data-shared-prod/mozilla_org/ga_sessions/view.sql
+++ b/sql/moz-fx-data-shared-prod/mozilla_org/ga_sessions/view.sql
@@ -3,11 +3,11 @@ CREATE OR REPLACE VIEW
 AS
 SELECT
   * REPLACE (
-    mozdata.analysis.ga_nullify_string(campaign) AS campaign,
-    mozdata.analysis.ga_nullify_string(source) AS source,
-    mozdata.analysis.ga_nullify_string(medium) AS medium,
-    mozdata.analysis.ga_nullify_string(term) AS term,
-    mozdata.analysis.ga_nullify_string(content) AS content
+    mozfun.ga.nullify_string(campaign) AS campaign,
+    mozfun.ga.nullify_string(source) AS source,
+    mozfun.ga.nullify_string(medium) AS medium,
+    mozfun.ga.nullify_string(term) AS term,
+    mozfun.ga.nullify_string(content) AS content
   )
 FROM
   `moz-fx-data-shared-prod.mozilla_org_derived.ga_sessions_v1`
--- a/sql/moz-fx-data-shared-prod/mozilla_org/ga_sessions/view.sql
+++ b/sql/moz-fx-data-shared-prod/mozilla_org/ga_sessions/view.sql
@@ -1,0 +1,13 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.mozilla_org.ga_sessions`
+AS
+SELECT
+  * REPLACE (
+    mozdata.analysis.ga_nullify_string(campaign) AS campaign,
+    mozdata.analysis.ga_nullify_string(source) AS source,
+    mozdata.analysis.ga_nullify_string(medium) AS medium,
+    mozdata.analysis.ga_nullify_string(term) AS term,
+    mozdata.analysis.ga_nullify_string(content) AS content
+  )
+FROM
+  `moz-fx-data-shared-prod.mozilla_org_derived.ga_sessions_v1`
--- a/sql/moz-fx-data-shared-prod/mozilla_org_derived/dataset_metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/mozilla_org_derived/dataset_metadata.yaml
@@ -1,0 +1,14 @@
+friendly_name: Mozilla.org Derived
+description: |-
+  Mozilla.org data, usually derived from Google Analytics
+dataset_base_acl: derived
+user_facing: false
+labels: {}
+default_table_workgroup_access:
+- role: roles/bigquery.dataViewer
+  members:
+  - workgroup:mozilla-confidential
+workgroup_access:
+- role: roles/bigquery.dataViewer
+  members:
+  - workgroup:mozilla-confidential
--- a/sql/moz-fx-data-shared-prod/mozilla_org_derived/ga_sessions_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/mozilla_org_derived/ga_sessions_v1/metadata.yaml
@@ -17,5 +17,15 @@ bigquery:
     expiration_days: null
   clustering:
     fields: ["country"]
+  depends_on:
+  - task_id: mozilla_org_derived__ga_sessions__v1__backfill_-1
+    dag_name: ga_sessions_backfill
+    execution_delta: 1h
+  - task_id: mozilla_org_derived__ga_sessions__v1__backfill_-2
+    dag_name: ga_sessions_backfill
+    execution_delta: 1h
+  - task_id: mozilla_org_derived__ga_sessions__v1__backfill_-3
+    dag_name: ga_sessions_backfill
+    execution_delta: 1h
 references: {}
 deprecated: false
--- a/sql/moz-fx-data-shared-prod/mozilla_org_derived/ga_sessions_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/mozilla_org_derived/ga_sessions_v1/metadata.yaml
@@ -9,6 +9,16 @@ labels:
 scheduling:
   dag_name: bqetl_mozilla_org_derived
   date_partition_parameter: session_date
+  depends_on:
+  - task_id: "mozilla_org_derived__ga_sessions__v1__backfill_-1"
+    dag_name: ga_sessions_backfill
+    execution_delta: 1h
+  - task_id: "mozilla_org_derived__ga_sessions__v1__backfill_-2"
+    dag_name: ga_sessions_backfill
+    execution_delta: 1h
+  - task_id: "mozilla_org_derived__ga_sessions__v1__backfill_-3"
+    dag_name: ga_sessions_backfill
+    execution_delta: 1h
 bigquery:
   time_partitioning:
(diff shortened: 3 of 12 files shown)


fbertsch:
Depends on https://github.com/mozilla/bigquery-etl/pull/4556/files

sql/moz-fx-data-shared-prod/mozilla_org_derived/dataset_metadata.yaml
```diff
@@ -0,0 +1,14 @@
...
+description: |-
+  Mozilla.org data, usually derived from Google Analytics
+dataset_base_acl: derived
+user_facing: false
+labels: {}
+default_table_workgroup_access:
+- role: roles/bigquery.dataViewer
+  members:
+  - workgroup:mozilla-confidential
+workgroup_access:
```
whd, line 11: Table ACLs are additive to Dataset ACLs. Therefore, in order to restrict access to https://github.com/mozilla/bigquery-etl/pull/4557/files#diff-7d19dc15354cd1c33edcdb1e504d4617804c730316620813fcb43e800ccd3d4aR22 within this dataset the dataset ACL here has to be something more restrictive than `workgroup:mozilla-confidential` (like `[]`), and all other tables in the dataset that are `workgroup:mozilla-confidential` would need to be annotated specifically (or potentially automatically via `default_table_workgroup_access`).

Additionally the dataset base acl here being `derived` grants implicit read/write access to DE. That needs to be `derived_restricted` if it's not considered permissible for all of DE to have read/write access to any workgroup-confidential data in the dataset. In that configuration if anyone apart from Airflow needs direct write access to the dataset that would need to be granted here as well (https://mozilla-hub.atlassian.net/browse/DSRE-1023).

fbertsch: Many thanks, indeed we'll lock down the stub attribution logs in this follow-up PR: https://github.com/mozilla/bigquery-etl/pull/4557

sql/moz-fx-data-shared-prod/mozilla_org_derived/ga_sessions_v1/query.sql
```diff
@@ -0,0 +1,154 @@
...
+    MIN_BY(device.browserVersion, visitStartTime) AS browser_version,
+  FROM
+    `moz-fx-data-marketing-prod.65789850.ga_sessions_*`
+  WHERE
+    -- This table is partitioned, so we only process the data from session_date
+    -- To handle late-arriving data, we process 3 days of data each day (re-processing the past 2)
+    -- as separate Airflow tasks (or via bqetl backfill, I haven't decided yet)
+    --
+    -- Here, we need to take data from yesterday, just in case some of our sessions from today
+    -- actually started yesterday. If they did, they'll be filtered out in the HAVING clause
```
ANich, line 82: Would it be acceptable to process data once it's "settled" 3 days later?

sql/moz-fx-data-shared-prod/mozilla_org_derived/ga_sessions_v1/query.sql
```diff
@@ -0,0 +1,154 @@
...
+    _TABLE_SUFFIX
+    BETWEEN FORMAT_DATE('%Y%m%d', DATE_SUB(@session_date, INTERVAL 1 DAY))
+    -- However, we have data for today that will arrive _tomorrow_! Some inter-day sessions
+    -- will be present in two days, with the same ids. A session should never span more
+    -- than two days though, see https://sql.telemetry.mozilla.org/queries/95882/source
+    -- If one does, our uniqueness check will alert us
+    AND FORMAT_DATE('%Y%m%d', DATE_ADD(@session_date, INTERVAL 1 DAY))
```
ANich, line 89: This part broke my brain a little, how will `@session_date` be set? Isn't this getting data from the future? See above question on getting data once it's settled.

ANich (commented):
Main question is about handling `@session_date` - is the case for updating past sessions a future concern?

sql/moz-fx-data-shared-prod/mozilla_org/ga_sessions/view.sql
```diff
@@ -0,0 +1,13 @@
...
+  `moz-fx-data-shared-prod.mozilla_org.ga_sessions`
+AS
+SELECT
+  * REPLACE (
+    mozfun.ga.nullify_string(campaign) AS campaign,
+    mozfun.ga.nullify_string(source) AS source,
+    mozfun.ga.nullify_string(medium) AS medium,
+    mozfun.ga.nullify_string(term) AS term,
+    mozfun.ga.nullify_string(content) AS content
+  )
```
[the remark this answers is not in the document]
fbertsch: Honestly, just because we don't have to. Because it's a 1:1 scalar transform I generally prefer to keep it as superficial as possible (just in case we want to investigate the pre-transformed data in some way).

sql/moz-fx-data-shared-prod/mozilla_org_derived/ga_sessions_v1/metadata.yaml
```diff
@@ -0,0 +1,20 @@
+friendly_name: Ga Sessions
+description: |-
+  One row for each GA session.
+owners:
+- frank@mozilla.com
+labels:
+  incremental: true
+  owner1: frank@mozilla.com
+scheduling:
+  dag_name: bqetl_mozilla_org_derived
```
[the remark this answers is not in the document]
fbertsch: You're right, I realized the same when I was looking at the generated DAG earlier - thank you!

sql/moz-fx-data-shared-prod/mozilla_org_derived/ga_sessions_v1/query.sql
```diff
@@ -0,0 +1,154 @@
+-- First note: This table is meant to be forwards-compatible
+--   with the GA4 schema: https://support.google.com/analytics/answer/7029846
```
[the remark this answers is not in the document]
fbertsch: Compatible as in - we are able to get the same data. Usually it means "this field is available in some form in GA4", but the only one where that is not the case is `mobileDeviceInfo` (in GA4, that field is split into its components, like manufacturer and model). My thought was we could recreate the field in the view for backwards compatibility.

sql/moz-fx-data-shared-prod/mozilla_org_derived/ga_sessions_v1/query.sql
```diff
@@ -0,0 +1,154 @@
...
+    MIN_BY(device.deviceCategory, visitStartTime) AS device_category,
+    MIN_BY(device.mobileDeviceModel, visitStartTime) AS mobile_device_model,
+    MIN_BY(device.mobileDeviceInfo, visitStartTime) AS mobile_device_string,
+    MIN_BY(device.operatingSystem, visitStartTime) AS os,
+    MIN_BY(device.operatingSystemVersion, visitStartTime) AS os_version,
+    MIN_BY(device.language, visitStartTime) AS language,
+    MIN_BY(device.browser, visitStartTime) AS browser,
+    MIN_BY(device.browserVersion, visitStartTime) AS browser_version,
+  FROM
+    `moz-fx-data-marketing-prod.65789850.ga_sessions_*`
```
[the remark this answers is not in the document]
fbertsch: Definitely. This is probably worth doing alongside the GA4 migration

sql/moz-fx-data-shared-prod/mozilla_org_derived/ga_sessions_v1/query.sql
fbertsch: That's probably waiting too long. I would prefer to have this updated daily (and it's usually a large majority of the data that arrives same-day).

sql/moz-fx-data-shared-prod/mozilla_org_derived/ga_sessions_v1/query.sql
fbertsch: I'm writing an Airflow job that will process (today-1, today-2, and today-3). That will incrementally re-process the past 3 days, in which case they will look into the next day to ensure they have the full session information.

ANich: This seems pretty complex. Won't any tables downstream of this that don't have late-arriving data have to do the same then? Or risk having incorrect data

ANich (approved):
The reprocessing part seems a little complex. Waiting 3-days for complete data rather than dealing with incomplete data downstream and in dashboards seems worth the tradeoff - ✅ to unblock

sql/moz-fx-data-shared-prod/mozilla_org_derived/ga_sessions_v1/query.sql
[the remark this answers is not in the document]
gkabbz: I don't know if we have much option for the ga_sessions_*. GA basically drops this data in this form in our warehouse. We don't do much to it. Looks like they are still doing it even for the GA4 datasets

sql/moz-fx-data-shared-prod/mozilla_org_derived/ga_sessions_v1/query.sql
gkabbz: I can see the utility of being able to see the campaign immediately but I'd be worried about changing numbers and how peeps might feel if reports shift on them.

fbertsch:
@ANich @gkabbz many thanks for the review and thoughts here.

In terms of the complication of handling late-arriving data vs. waiting, I strongly believe that we will decrease trust in the platform if we are not able to handle this in a manner similar to other tools. In the GA dashboards, for example, they don't wait for data to settle before displaying. For some ad-hoc & fast-turnaround analyses, this is an acceptable tradeoff.

For other datasets and metrics (dare I say, silver-level?), we may want to wait to report them until the data has settled. I would like to leave that to downstream datasets/metrics, however, than force it on them here.

PR #4562, opened 2023-11-14
feat(): updated fenix.firefox_android_clients view to use v2 of the corresponding table

kik-kik:
# feat(): updated fenix.firefox_android_clients view to use v2 of the corresponding table

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1975)

PR #4563, opened 2023-11-14
feat(): updated fenix_derived.funnel_retention_clients_* to use clients view instead of table directly

kik-kik:
# feat(): updated fenix_derived.funnel_retention_clients_* to use clients view instead of table directly

This can be merged once PR: https://github.com/mozilla/bigquery-etl/pull/4562 has been merged.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1976)


soGaussian (commented):
What is the rational behind this change

soGaussian (approved):

kik-kik:
> What is the rational behind this change

The view does some transformation (so this is for consistency across consumers of this table) and generally acts as an abstraction so that if we ever update the underlying table to v3 this query will not require any modifications.

PR #4565, opened 2023-11-14
feat(): updated fenix_derived.firefox_android_clients_v2 to include reported_baseline_ping field

kik-kik:
# feat(): updated fenix_derived.firefox_android_clients_v2 to include reported_baseline_ping field

Looks like we forgot to include this field when originally creating the v2. It appears to be used downstream and hence why the query needs to be updated to include it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1978)

### DIFF
--- a/sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v2/query.sql
+++ b/sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v2/query.sql
@@ -153,7 +153,7 @@ _current AS (
     device_model,
     os_version,
     app_version,
-    baseline_first_seen.locale,
+   locale,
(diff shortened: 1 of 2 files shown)


sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v2/query.sql
```diff
@@ -153,12 +153,13 @@ _current AS (
     device_model,
     os_version,
     app_version,
-    first_seen.locale,
+    baseline_first_seen.locale,
```
lucia-vargas-a, line 156:
```suggestion
   locale,
```

PR #4576, opened 2023-11-16
Define `event_monitoring_live_v1` views in `view.sql` files

sean-rose:
So they get automatically deployed by the `bqetl_artifact_deployment.publish_views` Airflow task and we don't get failures like [bug 1864961](https://bugzilla.mozilla.org/show_bug.cgi?id=1864961).

---
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1989)


sql_generators/glean_usage/event_monitoring_live.py
```diff
@@ -69,8 +69,8 @@ def generate_per_app_id(
         artifacts = []
 
         if not self.no_init:
```
sean-rose, line 71: I'm not sure if these `self.no_init` checks are still applicable.

scholtzan: I guess we can change this to `True`: https://github.com/mozilla/bigquery-etl/blob/5c2682a9cea4a8b5c984bfc9c71c771e5ec7f18c/sql_generators/glean_usage/event_error_monitoring.py#L25 and remove the `if not self.no_init`

scholtzan (approved):

bigquery_etl/view/__init__.py
```diff
@@ -192,10 +192,17 @@ def _valid_view_naming(self):
...
+                tokens[1].normalized == "VIEW"
+                or (
+                    tokens[1].normalized.upper() == "MATERIALIZED"
+                    and tokens[2].normalized == "VIEW"
+                )
+            )
         )
         if is_view_statement:
-            target_view = str(tokens[2]).strip().split()[0]
+            view_id_token = tokens[2] if tokens[1].normalized == "VIEW" else tokens[3]
```
sean-rose, line 204: sqlparse doesn't currently recognize `MATERIALIZED` as a keyword, so `normalized` doesn't automatically uppercase it.

sql_generators/glean_usage/templates/event_monitoring_live_v1.view.sql
```diff
@@ -1,9 +1,11 @@
-CREATE MATERIALIZED VIEW
-IF
-  NOT EXISTS `{{ project_id }}.{{ derived_dataset }}.event_monitoring_live_v1`
-  OPTIONS
-    (enable_refresh = TRUE, refresh_interval_minutes = 60) AS
-    {% if dataset_id not in ["telemetry", "accounts_frontend", "accounts_backend"] %}
+CREATE OR REPLACE MATERIALIZED VIEW
+  `{{ project_id }}.{{ derived_dataset }}.event_monitoring_live_v1`
```
sean-rose, line 2: IMO we should overwrite the materialized views every time so that any changes to their logic will get automatically deployed.  That does have the downside of losing the cached data when they're deployed, but since they're refreshing every hour anyway that doesn't seem like a dealbreaker to me, unless we decide to deploy views much more frequently.

sql_generators/glean_usage/templates/event_monitoring_live_v1.view.sql
```diff
@@ -1,9 +1,11 @@
...
+OPTIONS(
+  enable_refresh = TRUE,
+  refresh_interval_minutes = 60
+)
+AS
+  {% if dataset_id not in ["telemetry", "accounts_frontend", "accounts_backend"] %}
```
sean-rose, line 8: I just noticed that if the `dataset_id` is "telemetry" then this will end up generating invalid SQL, because "telemetry" isn't included in the `elif` and there's no `else`.

sql_generators/glean_usage/templates/event_monitoring_live_v1.view.sql
scholtzan: When we redeploy materialized views it will backfill the data every time. We do have a filter for `date(submission_timestamp)` for that reason as part of the query for when the materialized view gets initially deployed.
Also, materialized views cannot be replaced. They need to be deleted first: https://cloud.google.com/bigquery/docs/materialized-views-intro#limitations

bigquery_etl/view/__init__.py
```diff
@@ -190,14 +190,13 @@ def _valid_view_naming(self):
...
-                )
+        is_view_statement = " ".join(
+            tokens[0].normalized.split()
+        ) == "CREATE OR REPLACE" and (
+            tokens[1].normalized == "VIEW"
+            or (
+                tokens[1].normalized.upper() == "MATERIALIZED"
+                and tokens[2].normalized == "VIEW"
             )
         )
```
sean-rose, line 201: Generally I'm fine with the formatting Black requires, but then occasionally it does something like this, which I find so much less readable. 😞

sql_generators/glean_usage/templates/event_monitoring_live_v1.view.sql
scholtzan: Yes, we intentionally don't want to monitor `telemetry`. This is for Glean apps only. Telemetry gets filtered out by default, since these views only get created for Glean apps, so we could just remove it from this list

sql_generators/glean_usage/templates/event_monitoring_live_v1.view.sql
sean-rose: OK, I'll revert this part then.

bigquery_etl/view/__init__.py
scholtzan: Might be worth opening a bug for this. We've opened a couple in the past (like [this](https://github.com/andialbrecht/sqlparse/issues/583) one). Usually, the response/fix time is super fast

sql_generators/glean_usage/templates/event_monitoring_live_v1.view.sql
sean-rose: Since this PR is no longer making any other changes to that file's contents and it's essentially a non-issue, I'm going to leave it as is in this PR.

bigquery_etl/view/__init__.py
sean-rose: Good idea: https://github.com/andialbrecht/sqlparse/issues/752

bigquery_etl/view/__init__.py
```diff
@@ -190,16 +190,28 @@ def _valid_view_naming(self):
...
+                    t.normalized for t in tokens[view_kw_index + 1 : view_kw_index + 4]
                 )
+                == "IF NOT EXISTS"
+            ):
+                view_id_token = tokens[view_kw_index + 4]
+            else:
+                view_id_token = tokens[view_kw_index + 1]
+            target_view = str(view_id_token).replace("`", "").strip().split()[0]
+            try:
+                [project_id, dataset_id, view_id] = target_view.split(".")
```
scholtzan, line 214: This took me a minute to parse. Not sure if some comments might make sense here?

scholtzan (approved):
LGTM. Optional suggestion to maybe add a comment

bigquery_etl/view/__init__.py
sean-rose: Yeah, I was annoyed how verbose this logic ended up being.  I was hoping we could just grab the first identifier token, but unfortunately sqlparse currently misclassifies the `MATERIALIZED` keyword as an identifier.

Actually, since this isn't a pressing issue, I think I'll wait for them to fix the bug I submitted, then this can be simplified to just look for the first identifier token.

Or an alternate approach would be using sqlparse just to remove comments from the SQL, then using regular expressions to extract the view ID.

sean-rose: > Or an alternate approach would be using sqlparse just to remove comments from the SQL, then using regular expressions to extract the view ID.

I've implemented this alternate approach.

scholtzan (approved):

scholtzan:
Ah, turns out this is breaking view deploys: https://github.com/mozilla/bigquery-etl/blob/6e2e8f6f688a91f292774995f119ecdd0b63cd8f/bigquery_etl/view/__init__.py#L258

Materialized views don't have a `view_query`. Reverting these changes for now

PR #4610, opened 2023-11-28
feat(DENG-1774 / cancelled): deleting fenix_derived/firefox_android_clients_v2, v1 will remains the active model

kik-kik:
# feat(DENG-1774 / cancelled): deleting fenix_derived/firefox_android_clients_v2, v1 will remains the active model

Deploying the v2 version of the table proved problematic and in the interest of time we decided for now to continue using v1 as it already contains all the data we need.

There is no impact to this change as the table has not been productionised and there are no current references to the table inside the bqetl repo.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2044)


lucia-vargas-a (commented):
Hi Kik, for reference and understanding, could you please add in the description why the decision to remove the table? or a link to the Jira task?

lucia-vargas-a (approved):

lucia-vargas-a:
Ideally in the description we could specify what problems arised that required removing the table, and if this work will continue in a separate PR.

PR #4619, opened 2023-11-29
Speed up glean_usage generation by caching the table getter

badboy:
`get_tables` is deterministic under the assumption that the tables don't change in between invocations. Which I hope holds here. We therefore can just cache that value so that subsequent runs quickly return without needing a roundtrip to BigQuery again.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2085)

### DIFF
--- a/sql_generators/glean_usage/__init__.py
+++ b/sql_generators/glean_usage/__init__.py
@@ -1,5 +1,5 @@
 """GLEAN Usage."""
-from functools import partial
+from functools import cache, partial
 from pathlib import Path
 
 import click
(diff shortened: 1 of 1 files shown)


sql_generators/glean_usage/__init__.py
```diff
@@ -99,6 +99,7 @@ def generate(
     elif exclude:
         table_filter = partial(table_matches_patterns, exclude, True)
 
+    @cache
```
ANich, line 102: Could it be more beneficial to cache `sql_generators.glean_usage.common.list_tables` instead? This fn mostly delegates there then does a filter (unless the filter is what takes a lot of time here)

ANich (approved):
LGTM left one question

sql_generators/glean_usage/__init__.py
badboy: It wouldn't make much of a difference. The only call to that `list_tables` is this one here in `get_tables`.

Issue #4532, opened 2023-11-07
`bqetl query backfill` does not support non-partitioned tables


fbertsch:
The table `fenix_derived.client_adclicks_history_v1` is not partitioned, but I would like to backfill it by-day:
```
bqetl query backfill fenix_derived.client_adclicks_history_v1 --sql_dir sql --project_id moz-fx-data-shared-prod --start_date 2020-08-01 --end_date 2023-11-07 --max_rows 4 --parallelism 1 --checks
```

However the above fails with:
```
  File "/Users/fbertsch/repos/bigquery-etl/bigquery_etl/cli/query.py", line 725, in backfill
    partitioning_type = metadata.bigquery.time_partitioning.type
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'type'
```

Looking through the code, we hardcode the partitioning spec in several places (it requires daily or monthly in both `backfill` and `_backfill_query`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-1934)

ANich:
I've been looking into this on and off as a part of the managed backfills work. There's also a couple other things that aren't by default supported:
- `date_partition_offset` (technically possible iff there's also a `date_partition_parameter`) 
- Queries with additional parameters

I'm working on a few suggestions/PRs. I'll link this issue where relevant

PR #4627, opened 2023-11-29
Support offset backfills, require metadata 

ANich:
Add support for `dataset_partition_offset` for `bqetl query backfill`. Also require a `metadata.yaml` file for backfilling a table

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2098)

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -507,8 +507,9 @@ def _backfill_query(
     run_checks,
 ):
     """Run a query backfill for a specific date."""
-    if backfill_date in exclude:
-        click.echo(f"Skip {query_file_path} backfill for run date {backfill_date}")
+    backfill_date_str = backfill_date.strftime("%Y-%m-%d")
+    if backfill_date_str in exclude:
+        click.echo(f"Skip {query_file_path} backfill for run date {backfill_date_str}")
         return True
 
     project, dataset, table = extract_from_query_path(query_file_path)
@@ -520,12 +521,10 @@ def _backfill_query(
         case PartitionType.MONTH:
             if date_partition_parameter != 0:
                 # TODO: Support offsets here e.g. desktop_mobile_search_clients_monthly_v1
-                raise ValueError("TODO: Can't set offset for month partitions.")
+                raise ValueError("Unsupported offset for month partitions.")
             partition = backfill_date.strftime("%Y%m")
         case _:
             raise ValueError(f"Unsupported partitioning type: {partitioning_type}")
-
-    backfill_date_str = backfill_date.strftime("%Y-%m-%d")
 
     if destination_table is None:
         destination_table = f"{project}.{dataset}.{table}"
@@ -764,7 +763,7 @@ def backfill(
             case _:
                 raise ValueError(f"Unsupported partitioning type: {partitioning_type}")
 
-        if depends_on_past and exclude != []:
+        if depends_on_past and exclude:
             click.echo(
                 f"Warning: depends_on_past = True for {query_file_path} but the"
                 f"following dates will be excluded from the backfill: {exclude}"
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -509,7 +509,7 @@ def _backfill_query(
     """Run a query backfill for a specific date."""
     backfill_date_str = backfill_date.strftime("%Y-%m-%d")
     if backfill_date_str in exclude:
-        click.echo(f"Skip {query_file_path} backfill for run date {backfill_date_str}")
+        click.echo(f"Skipping {query_file_path} backfill for run date {backfill_date_str}")
         return True
 
     project, dataset, table = extract_from_query_path(query_file_path)
@@ -519,9 +519,9 @@ def _backfill_query(
             partition_date = backfill_date + timedelta(days=date_partition_offset)
             partition = partition_date.strftime("%Y%m%d")
         case PartitionType.MONTH:
-            if date_partition_parameter != 0:
+            if date_partition_offset != 0:
                 # TODO: Support offsets here e.g. desktop_mobile_search_clients_monthly_v1
-                raise ValueError("Unsupported offset for month partitions.")
+                raise ValueError("Offsets are unsupported for non-daily partitions (found date_partition_offset={date_partition_offset}).")
             partition = backfill_date.strftime("%Y%m")
         case _:
             raise ValueError(f"Unsupported partitioning type: {partitioning_type}")
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -702,59 +702,49 @@ def backfill(
         )
         query_files = paths_matching_name_pattern(name, ctx.obj["TMP_DIR"], project_id)
 
-    dates = [start_date + timedelta(i) for i in range((end_date - start_date).days + 1)]
-
     for query_file in query_files:
         query_file_path = Path(query_file)
 
-        depends_on_past = False
-        date_partition_parameter = "submission_date"
-
         try:
             metadata = Metadata.of_query_file(str(query_file_path))
-            depends_on_past = metadata.scheduling.get(
-                "depends_on_past", depends_on_past
-            )
-            date_partition_parameter = metadata.scheduling.get(
-                "date_partition_parameter", date_partition_parameter
-            )
-
-            # For backwards compatibility assume partitioning type is day
-            # in case metadata is missing
-            if metadata.bigquery:
-                partitioning_type = metadata.bigquery.time_partitioning.type
-            else:
-                partitioning_type = PartitionType.DAY
-                click.echo(
-                    "Bigquery partitioning type not set. Using PartitionType.DAY"
-                )
-
-            match partitioning_type:
-                case PartitionType.DAY:
-                    dates = [
-                        start_date + timedelta(i)
-                        for i in range((end_date - start_date).days + 1)
-                    ]
-                case PartitionType.MONTH:
-                    dates = list(
-                        rrule(
-                            freq=MONTHLY,
-                            dtstart=start_date.replace(day=1),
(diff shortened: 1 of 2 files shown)


bigquery_etl/cli/query.py
```diff
@@ -702,61 +717,54 @@ def backfill(
...
+                    for day in exclude
+                ]
+            case _:
+                raise ValueError(f"Unsupported partitioning type: {partitioning_type}")
 
-        except FileNotFoundError:
-            click.echo(f"No metadata defined for {query_file_path}")
-
-        if depends_on_past and exclude != []:
+        if depends_on_past and exclude:
```
ANich, line 766: The default is actually an empty `tuple` so this was being printed even if there's nothing in `exclude`

bigquery_etl/cli/query.py
```diff
@@ -702,61 +717,54 @@ def backfill(
...
-                case _:
-                    raise ValueError(
-                        f"Unsupported partitioning type: {partitioning_type}"
+        # For backwards compatibility assume partitioning type is day
+        # in case metadata is missing
+        if metadata.bigquery:
+            partitioning_type = metadata.bigquery.time_partitioning.type
+        else:
+            partitioning_type = PartitionType.DAY
+            click.echo("Bigquery partitioning type not set. Using PartitionType.DAY")
```
ANich, line 742: Will change this assumption in upcoming PR(s), for example backfilling unpartitioned tables. See - https://github.com/mozilla/bigquery-etl/issues/4532 (though I think this is somewhat supported via `--no-partition`)

bigquery_etl/cli/query.py
```diff
@@ -506,76 +507,89 @@ def _backfill_query(
...
-            f"Run backfill for {destination_table} "
-            f"with @{date_partition_parameter}={backfill_date}"
-        )
+    scheduling_params = []
+    if date_partition_parameter is not None:
+        # TODO: Support non date_partition_parameter scheduling.parameters e.g. clients_first_seen_28_days_later_v1
+        offset_param = backfill_date + timedelta(days=date_partition_offset)
+        scheduling_params += [
+            f"--parameter={date_partition_parameter}:DATE:{offset_param.strftime('%Y-%m-%d')}"
+        ]
```
ANich, line 545: Other scheduling parameters will be added here, likely (re-)using some of the Airflow functionality: https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html#macros

bigquery_etl/cli/query.py
```diff
@@ -784,7 +792,7 @@ def backfill(
             run_checks=checks,
         )
 
-        if not depends_on_past:
+        if not depends_on_past and parallelism > 0:
```
ANich, line 795: `pytest` doesn't like Multiprocessing apparently. I get an indefinite hang when I try to run a test with parallelism on. This is just a way to force sequential runs in tests and otherwise.

bigquery_etl/cli/query.py
```diff
@@ -506,76 +507,89 @@ def _backfill_query(
...
+
     project, dataset, table = extract_from_query_path(query_file_path)
 
     match partitioning_type:
         case PartitionType.DAY:
-            partition = backfill_date.strftime("%Y%m%d")
+            partition_date = backfill_date + timedelta(days=date_partition_offset)
+            partition = partition_date.strftime("%Y%m%d")
         case PartitionType.MONTH:
+            if date_partition_parameter != 0:
```
fbertsch, line 522:
```suggestion
            if date_partition_offset != 0:
```
I would suggest now a unit test for this 😄

bigquery_etl/cli/query.py
```diff
@@ -506,76 +507,89 @@ def _backfill_query(
     run_checks,
 ):
     """Run a query backfill for a specific date."""
+    backfill_date_str = backfill_date.strftime("%Y-%m-%d")
+    if backfill_date_str in exclude:
+        click.echo(f"Skip {query_file_path} backfill for run date {backfill_date_str}")
```
fbertsch, line 512:
```suggestion
        click.echo(f"Skipping {query_file_path} backfill for run date {backfill_date_str}")
```

fbertsch: There's something off here. Like this function should be simple: run a backfill on a date. The filtering of dates should happen outside of this function, e.g.:
```
dates.filter(date -> date not in excludes).map(date -> _backfill_query_closure(date))
```

fbertsch: Would also let us remove the early return 😉  (but not blocking!)

bigquery_etl/cli/query.py
```diff
@@ -506,76 +507,89 @@ def _backfill_query(
...
-    if backfill_date not in exclude:
-        if destination_table is None:
-            destination_table = f"{project}.{dataset}.{table}"
+    if destination_table is None:
+        destination_table = f"{project}.{dataset}.{table}"
 
-        if not no_partition:
-            destination_table = f"{destination_table}${partition}"
+    if not no_partition:
+        destination_table = f"{destination_table}${partition}"
```
fbertsch, line 533: I don't even understand this option anymore. We can only backfill partitioned tables, in what situation would we want a single date for the backfill to overwrite the entire table? Are they even backfilling other dates, or do they all overwrite the entire table?? Is it a backfill to init the table with a single date??? Then why not just write to that date's partition?????

bigquery_etl/cli/query.py
```diff
@@ -506,76 +507,89 @@ def _backfill_query(
...
+        click.echo("Destination table must be named like: <project>.<dataset>.<table>")
+        sys.exit(1)
 
-        click.echo(
-            f"Run backfill for {destination_table} "
-            f"with @{date_partition_parameter}={backfill_date}"
-        )
+    scheduling_params = []
+    if date_partition_parameter is not None:
+        # TODO: Support non date_partition_parameter scheduling.parameters e.g. clients_first_seen_28_days_later_v1
```
fbertsch, line 541: Is this possible? These tables would have failed above with no partitioning type, right?

fbertsch (commented):
r+, but what is the purpose here? Why not put the dates you want?

bigquery_etl/cli/query.py
ANich: Dates could be filtered earlier here yeah, I assume the original motivation was to allow logging which I think is still useful. I think in `_backfill_query`, `exclude` could probably be changed to a boolean. Something like `skip=False`

bigquery_etl/cli/query.py
ANich: `:facepalm:` this logic should be going away pretty soon, hence the TODO below

bigquery_etl/cli/query.py
ANich: One use-case is rolling tables e.g. `firefox_android_clients_v1`. The output table might be partitioned or not but it's specifically not partitioned by the query parameter. e.g. a table partitioned by `first_seen_date` but that uses `submission_date` as a parameter. Generally it shouldn't be necessary, if you're running a backfill the metadata is probably enough to tell us how to run the query (it's enough to define the task in Airflow so)

bigquery_etl/cli/query.py
ANich: This matches the Airflow logic, `date_partition_offset` even with a null `date_partition_parameter` adds a partition template. `date_partition_parameter` can be null (but the table still partitioned) if you want to run your query with something other than the partition field as a "daily" query parameter. In the TODO example the table is partitioned by `first_seen_date` but the query is run with today's `submission_date`.

This is kinda confusing but `date_partition_parameter` exists as a convenient default for tables that are partitioned by the same field they're meant to be updated daily by

ANich:
> r+, but what is the purpose here? Why not put the dates you want?

See replies - the dates are parameter dependent, sometimes that's not the same as the execution date based on the metadata

bigquery_etl/cli/query.py
fbertsch: The problem is this doesn't generally support those incremental tables, just the partitioned ones. So it almost seems like it accidentally works for those tables

bigquery_etl/cli/query.py
fbertsch: I think what is confusing me _is_ the TODO. This currently does not support that situation? But we want to support it eventually? Then the comment "These tables would have failed above with no partitioning type, right?" is still correct, but we will eventually change that behavior?

bigquery_etl/cli/query.py
ANich: Yes I agree, it works but in a roundabout way. The current assumption defaults to daily partition and `no_partition` will override that when deciding where to write. I'm actually in favour of removing `no_partition` and replacing it with just using metadata but that was out of scope for this PR

bigquery_etl/cli/query.py
ANich: Yes this isn't currently supported. Some queries set additional scheduling `parameters` using Airflow syntax (e.g. `submission_date:DATE:{{ds}}`) in metadata, those still aren't supported. This PR only adds support for `date_partition_offset` The goal here is that you should be able to backfill any table with a query that's run in Airflow

ANich:
> @ANich can you add some examples of how you use this and on which tables (or probably just 1 in particular you're working on)? That may help me grok what you're saying about use cases.

`ga_derived.downloads_with_attributions_v2` (or any table with a `date_partition_offset`) is an example. In Airflow it runs on execution day `n` but the partition that gets written to is `n - 1`. This mirrors that so you run a backfill with an execution date and let the backfill (like Airflow) handle the actual parameter setting. Another example is `telemetry_derived.clients_first_seen_28_days_later_v1` which has a 28 day offset but this currently doesn't supported because other scheduling parameters generally aren't supported (which will be in an upcoming PR)

fbertsch (approved):
I see. I didn't realize this was an option in scheduling as well. Thank you for adding this

PR #4631, opened 2023-11-30
feat(DENG-2083): added firefox_ios_derived.clients_activation_v1 and corresponding view

kik-kik:
# feat(DENG-2083): added firefox_ios_derived.clients_activation_v1 and corresponding view

The purpose of this change is to avoid the activations table diverging away from the clients table. This is currently possible, as both the clients and activations table are built independently using different sources. Naturally, this resulted us observing unexpected differences in the client counts between the two tables. This should not be the case and is something that is addressed by this change.

This change aims to improve the two following aspects of this data model:

1. Make the data lineage simpler and representative of the expected stage in the data lifecycle (there should be no activation for a client, unless they are recorded as a client in the clients table).
2. Remove the self (previous state) dependency from this model. This simplifies the logic and makes each partition idempotent making it much easier to rebuild specific partitions if necessary.

----

depends on: https://github.com/mozilla/bigquery-etl/pull/4638

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2103)

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_ios_derived/clients_activation_v1/checks.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_ios_derived/clients_activation_v1/checks.sql
@@ -7,7 +7,6 @@ WITH upstream_clients_count AS (
   SELECT COUNT(*)
   FROM `{{ project_id }}.firefox_ios.firefox_ios_clients`
   WHERE first_seen_date = DATE_SUB(@submission_date, INTERVAL 6 DAY)
-  AND NOT is_suspicious_device_client
 ),
 activations_clients_count AS (
   SELECT COUNT(*)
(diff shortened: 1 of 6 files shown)


sql/moz-fx-data-shared-prod/firefox_ios_derived/clients_activation_v1/checks.sql
```diff
@@ -0,0 +1,31 @@
+#fail
+{{ is_unique(["client_id"]) }}
+#fail
+{{ min_row_count(1, "`submission_date` = @submission_date") }}
+#fail
+WITH upstream_clients_count AS (
+  SELECT COUNT(*)
+  FROM `{{ project_id }}.firefox_ios.firefox_ios_clients`
+  WHERE first_seen_date = DATE_SUB(@submission_date, INTERVAL 6 DAY)
+  AND NOT is_suspicious_device_client
```
bani, line 10: the suspicious client filter was a funnel decision. Are we going to start applying it more broadly? Activation isn't even a funnel metric.

kik-kik: Filter has been removed. Thanks!

bani (approved):

PR #4649, opened 2023-12-05
feat(DENG-2049): Migrated DIM checks over to ETL checks for telemetry_derived.unified_metrics_v1

kik-kik:
# feat(DENG-2049): Migrated DIM checks over to ETL checks for telemetry_derived.unified_metrics_v1

DIM checks from here have been moved over to BQETL data checks:

- https://github.com/mozilla/dim/blob/main/dim_checks/moz-fx-data-shared-prod/telemetry_derived/unified_metrics_v1/dim_checks.yaml
- https://github.com/mozilla/dim/blob/main/dim_checks/moz-fx-data-shared-prod/telemetry/unified_metrics/dim_checks.yaml

---

*In the future*, we should turn some of these checks into macros:

- membership test (validate that a value in a column is expected)
- value length check
- value check to make sure it matches a regex expression.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2127)

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/unified_metrics_v1/checks.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/unified_metrics_v1/checks.sql
@@ -1,0 +1,108 @@
+#fail
+{{ min_row_count(1000, where="submission_date = @submission_date") }}
+
+#fail
+{{ not_null(columns=["submission_date", "client_id", "first_seen_date"], where="submission_date = @submission_date") }}
+
+#fail
+{{ row_count_within_past_partitions_avg(number_of_days=7, threshold_percentage=5) }}
+
+#fail
+SELECT IF(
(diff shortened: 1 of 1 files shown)


lelilia:
how big is the liklyhood of going back and doing the macros?
I think we should do it now while at it? 

[dbt example for accepted_values](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/include/global_project/macros/generic_test_sql/accepted_values.sql)

lelilia:
you could ... write the macro in a PR and then use the new macros in this PR though?

lelilia:
https://github.com/mozilla/bigquery-etl/pull/4650
here is the first one :)

kik-kik:
Yes, but I'd still like to merge this first so that I can get rid off DIM 🙃

kik-kik:
Thanks!

PR #4655, opened 2023-12-06
Add a generator for events stream tables

badboy:
Our plan is:

A (persisted) events stream table per product, with one row per event.
Over the past few days I learned more about bigquery-etl and how it works.
And thus ended up using sql_generators/glean_usage to do that for Glean products.
This replaces #4463

cc @quiiver  

This is a draft and not yet ready to land, there's some open questions:

* How does init work?
  * Is this manually triggered? How do we backfill to a certain date?
* The schema is defined in the SQL query. How does this behave on changes in the future?
* Configuration: Right now it's inline in Python. Where should we put this?

Probably good for someone like @scholtzan to give this a look?


sql_generators/glean_usage/templates/events_stream_v1.query.sql
```diff
@@ -0,0 +1,64 @@
+{{ header }}
+
+{% if init %}
+CREATE TABLE IF NOT EXISTS
+  `{{ events_stream_table }}`
+PARTITION BY
+  submission_date
+CLUSTER BY
+  sample_id,
+  submission_date
```
quiiver, line 10: Given that we already partition on `submission_date`, what value do we get from clustering on `submission_date`? Thinking about the query patterns most common, what do you think about clustering on event name and or client id?

sql_generators/glean_usage/events_stream.py
```diff
@@ -0,0 +1,62 @@
...
+PREFIX = "events_stream"
+# Only generate it for Fenix for now
+APP_ID_ALLOW = {
+    "org_mozilla_firefox",
+    "org_mozilla_fenix",
+    "org_mozilla_fenix_beta",
+}
+DATASET_ALLOW = {
+    "fenix",
+}
```
scholtzan, line 17: It would be preferable to have this kind of configuration live in here: https://github.com/mozilla/bigquery-etl/blob/be60f5aa56a3ae53123149a4812205a32ebc9462/bqetl_project.yaml#L431

sql_generators/glean_usage/templates/events_stream_v1.query.sql
```diff
@@ -0,0 +1,64 @@
+{{ header }}
```
scholtzan, line 1: This looks like it's essentially doing the same as events_unnested but writing it to a table: https://github.com/mozilla/bigquery-etl/blob/main/sql_generators/glean_usage/templates/cross_channel_events_unnested.view.sql

badboy:
> Wasn't the goal of the event stream tables to work on the live tables and provide unnested events as soon as they come in?

Not a requirement from me and I would have defer to you what is better.
(I often hear that people want data _faster_, but rarely live data is really needed for anything they do).

scholtzan:
I'd be curious if we have some specific use cases that would use these "event stream" tables.

I think there is a case to be made about both approaches:
* we had some users complaining that using the `events_unnested` views is getting slower as the amount of events we store increase: https://bugzilla.mozilla.org/show_bug.cgi?id=1869433 We could just materialize and cluster/partition `events_unnested`. Essentially what is done in this PR, although I would call these tables `events_unnested_v1` since `stream` does imply some real-time component, but it's all just a daily batch
* there were a couple of possible use cases being discussed to do real-time event analysis for mozsocial, so a materialized view would be potentially helpful.

The most cost-effective solution, which we are also using for event monitoring and experiments, is to use a materialized table for historic data and a materialized view to provide real-time data for the current day and `UNION` both of these in a view that users can access. Which I guess would cover all the use cases here...

scholtzan:
I just remembered we had a proposal for this: https://docs.google.com/document/d/1ndSuTpEP2EMa3PRkqNVHUuqTb01YadS6SsE8ypDUnUI/edit
I'll give it another read to see what the considerations were

scholtzan:
Alright, sorry for the confusion. After reading the proposal again, it seems there are some minor but significant differences between `events_unnested` and the `event_stream` tables. So it does make sense to create a new table like done here.
I guess a scheduled table that gets populated once a day is sufficient for now. If we do have some "real-time" use cases we can relatively easily combine the table with a materialized view to serve those needs.

sql_generators/glean_usage/templates/events_stream_v1.query.sql
scholtzan: I guess clustering on event name could make sense. One drawback with it could be that we have a relatively large number of different events. The more distinct values there are for clustering, the larger the metadata that needs to be scanned when querying which could potentially slow things down as well. Usually we clustered on event category which was a good option.

sql_generators/glean_usage/events_stream.py
scholtzan: This PR looks good, just moving the configuration here in the config file is the biggest todo to make this move ahead

sql_generators/glean_usage/templates/events_stream_v1.query.sql
```diff
@@ -0,0 +1,64 @@
...
+    ),
+    DATE(submission_timestamp) AS submission_date,
+    client_info.client_id AS client_id,
+    ping_info.reason AS reason,
+    `mozfun.json.from_map`(ping_info.experiments) AS experiments,
+    SAFE.TIMESTAMP_ADD(
+      ping_info.parsed_start_time,
+      INTERVAL event.timestamp MILLISECOND
+    ) AS event_timestamp,
+    (event.category || '.' || event.name) AS event,
```
scholtzan, line 47: I wonder if it makes sense to still have separate `category` and `event name` fields here. One for clustering purposes which can significantly help with cost and performance, and also filtering for multiple events or anything that would need to involve wildcard-matching might have some performance implication.

sql_generators/glean_usage/templates/cross_channel_events_stream.query.sql
```diff
@@ -0,0 +1,20 @@
...
+UNION ALL
+{% endif -%}
+SELECT
+  "{{ dataset }}" AS normalized_app_id,
+  e.*
+  REPLACE(
+    {% if app_name == "fenix" -%}
+    mozfun.norm.fenix_app_info("{{ dataset }}", client_info.app_build).channel AS normalized_channel,
+    {% elif datasets|length > 1 -%}
+    "{{ channel }}" AS normalized_channel,
```
scholtzan, line 16: to fix the failing CI task
```suggestion
    "{{ channel }}" AS normalized_channel
```

sql_generators/glean_usage/templates/events_stream_v1.query.sql
```diff
@@ -0,0 +1,56 @@
...
+    event.category as event_category,
+    event.name as event_name,
+    ARRAY_TO_STRING([event.category, event.name], ',') AS event, -- handles NULL values better
+    `mozfun.json.from_map`(event.extra) AS event_extra,
+  FROM
+    `{{ events_view }}` AS e
+  CROSS JOIN
+    UNNEST(events) AS event
+  WHERE
+    {% if is_init() %}
```
scholtzan, line 46:
```suggestion
    {% raw %}
    {% if is_init() %}
```

sql_generators/glean_usage/templates/events_stream_v1.query.sql
```diff
@@ -0,0 +1,56 @@
...
+      DATE(submission_timestamp) >= '2023-11-01'
+    {% else %}
+      DATE(submission_timestamp) = @submission_date
+    {% endif %}
```
scholtzan, line 50:
```suggestion
    {% endif %}
    {% endraw %}
```

sql_generators/glean_usage/common.py
```diff
@@ -120,6 +120,9 @@ def table_names_from_baseline(baseline_table, include_project_id=True):
         last_seen_view=f"{prefix}.baseline_clients_last_seen",
         first_seen_view=f"{prefix}.baseline_clients_first_seen",
         event_monitoring=f"{prefix}_derived.event_monitoring_live_v1",
+        events_view=f"{prefix}.events",
+        events_stream_table=f"{prefix}_derived.events_stream_v1",
+        events_stream_view=f"{prefix}_derived.events_stream",
```
scholtzan, line 125:
```suggestion
       events_stream_view=f"{prefix}.events_stream",
```

sql_generators/glean_usage/events_stream.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def __init__(self):
+        """Initialize events_stream table."""
+        GleanTable.__init__(self)
+        self.target_table_id = TARGET_TABLE_ID
+        self.prefix = PREFIX
+        self.no_init = True
+        self.per_app_enabled = True
+        self.per_app_id_enabled = True
+        self.across_apps_enabled = False
```
scholtzan, line 23:
```suggestion
        self.across_apps_enabled = True
```

sql_generators/glean_usage/templates/events_stream_v1.query.sql
```diff
@@ -0,0 +1,58 @@
...
+    client_info.client_id AS client_id,
+    ping_info.reason AS reason,
+    `mozfun.json.from_map`(ping_info.experiments) AS experiments,
+    SAFE.TIMESTAMP_ADD(
+      ping_info.parsed_start_time,
+      INTERVAL event.timestamp MILLISECOND
+    ) AS event_timestamp,
+    event.category as event_category,
+    event.name as event_name,
+    ARRAY_TO_STRING([event.category, event.name], ',') AS event, -- handles NULL values better
```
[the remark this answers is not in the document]
scholtzan: ah yes... let me push a fix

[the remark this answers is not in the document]
scholtzan: https://github.com/mozilla/bigquery-etl/pull/4838

PR #4670, opened 2023-12-08
bug(1868848): Skipping copying checks for baseline tables for apps marked as not receiving the baseline ping

kik-kik:
# bug(1868848): Skipping copying checks for baseline tables for apps marked as not receiving the baseline ping

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2154)


fbertsch (approved):
I have a few nits with this, but we can merge this and fix them later:
1. I believe that `does not send baseline ping` should be indicated in a Glean app's `repositories.yaml` entry. This would be useful information elsewhere as well (e.g. the Glean dictionary would not index the baseline ping)
2. It would be nice to have a way to verify that these are not sending that data

sql_generators/glean_usage/common.py
```diff
@@ -18,6 +18,24 @@
...
+    "burnham",
+    "firefox_reality_pc",
+    "lockwise_android",
+    "mach",
+    "monitor_cirrus",
+    "moso_mastodon_backend",
+    "mozphab",
+    "mozregression",
+)
+
```
scholtzan, line 38: I think we should move this into the config file here: https://github.com/mozilla/bigquery-etl/blob/6e2e8f6f688a91f292774995f119ecdd0b63cd8f/bqetl_project.yaml#L433

kik-kik: @scholtzan  That sounds like a reasonable suggestion, where would be the correct place to file a ticket for this so we do not forget about it?

scholtzan: There is a PR here: https://github.com/mozilla/bigquery-etl/pull/4676

PR #4673, opened 2023-12-08
feat/DENG-2051 migrate active users aggregates

lelilia:
Moving Data tests from dim repo into bigquery_etl

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2158)


sql_generators/active_users/__init__.py
```diff
@@ -49,14 +49,24 @@ def generate(target_project, output_dir, use_cloud_function):
...
+    # checks templates
+    desktop_checks_template = env.get_template("desktop_checks.sql")
+    focus_android_checks_template = env.get_template("focus_android_checks.sql")
+    fenix_checks_template = env.get_template("fenix_checks.sql")
+    focus_ios_checks_template = env.get_template("focus_ios_checks.sql")
+    firefox_ios_checks_template = env.get_template("firefox_ios_checks.sql")
+    klar_ios_checks_template = env.get_template("klar_ios_checks.sql")
 
     for browser in Browsers:
         if browser.name == "firefox_desktop":
```
kik-kik, line 72: Could we avoid repeating `if browser.name == "firefox_desktop":` if we just places the check render logic inside the same block as the query render call?

lelilia: I had that before but the when translating the different checks I saw that the checks are very specific for each `browser.name`

and yes for firefox_desktop I could move it up but for the rest not so for me it's simpler to read to do the checks in one place all together and not put some here and some there

or am I missreading your point?

sql_generators/active_users/templates/desktop_checks.sql
```diff
@@ -0,0 +1,152 @@
...
+    FULL OUTER JOIN
+        overactive
+    USING
+        (client_id)
+    WHERE
+        overactive.client_id IS NULL
+)
+SELECT
+    IF(
+        (SELECT * FROM dau_sum) <> (SELECT * FROM distinct_client_count),
```
kik-kik, line 147: Could we tweak this to allow delta of 1?

sql_generators/active_users/templates/focus_android_checks.sql
```diff
@@ -0,0 +1,60 @@
+{#
+   We use raw here b/c the first pass is rendered to create the checks.sql
+   files, and the second pass is rendering of the checks themselves.
+   For example, the header above is rendered for every checks file
```
fbertsch, line 4: nit: No header above these docstrings!

lelilia: good catch 🙏

sql_generators/active_users/templates/desktop_checks.sql
lelilia: for all of them or this one specific?

kik-kik: this specific one.

lucia-vargas-a (commented):
The checks LGTM.
But the idea of the sql_generators is to use a single template to generate similar code. Is there a reason not to generate a template for Mobile where the code can be re-utilized? The distinct_client_count could use a e.g. a dictionary of upstream tables to UNION as a template parameter.

lelilia:
@lucia-vargas-a ok I combined some of them, but fenix is just to weird to be included 
hope 3 template are ok

lucia-vargas-a (commented):
Question, are the tests set up to _not_ affect the update of the tables?
It's important that the checks don't block the update of the tables as they provide KPIs.

lucia-vargas-a (approved):
LGTM! Only one question about the checks not affecting the update.

kik-kik:
> LGTM! Only one question about the checks not affecting the update.

Thanks for taking a look @lucia-vargas-a . So as you can see in the templates, all checks are marked as `#warn`, this means that they will run immediately after the corresponding query completes and will not prevent downstream tasks  or DAGs from running. In this case, we will see errors for the warning checks but everything else will run as normal.

sql_generators/active_users/__init__.py
```diff
@@ -83,6 +90,81 @@ def generate(target_project, output_dir, use_cloud_function):
...
+                app_value=browser.value,
+                app_name=browser.name,
+            )
+        elif browser.name == "fenix":
+            checks_sql = fenix_checks_template.render(
+                project_id=target_project,
+                app_value=browser.value,
+                app_name=browser.name,
+            )
+        elif browser.name == "firefox_ios":
```
lucia-vargas-a, line 106: I think there is a way to not require these multiple "if" and making it easy to add/remove sources in the future. We could declare a dictionary at the beginning of the file similar to:
{
"firefox_ios": ['moz-fx-data-shared-prod.org_mozilla_ios_firefox_live.baseline_v1', 'moz-fx-data-shared-prod.org_mozilla_ios_firefoxbeta_live.baseline_v1', 'moz-fx-data-shared-prod.org_mozilla_ios_fennec_live.baseline_v1'], "focus_ios": ['moz-fx-data-shared-prod.org_mozilla_ios_focus_live.baseline_v1'],
"klar_ios": ['moz-fx-data-shared-prod.org_mozilla_ios_klar_live.baseline_v1'], "focus_android": ['moz-fx-data-shared-prod.org_mozilla_focus_live.baseline_v1', 'moz-fx-data-shared-prod.org_mozilla_focus_beta_live.baseline_v1', 'moz-fx-data-shared-prod.org_mozilla_focus_nightly_live.baseline_v1']
}
and use it to compose the "channels". The mobile template template would be rendered only once.
It can also be part of a later PR! Your call.

lucia-vargas-a (approved):

sql_generators/active_users/__init__.py
lelilia: just when the CI is nearly done with checking .. but no you are right good point let me quickly do that

PR #4697, opened 2023-12-13
Copy updated shredder progress view and remove duplicate

relud:
monitoring_derived.shredder_progress was updated for main_v5 changes, but monitoring.shredder_progress was not. solution: overwrite monitoring.shredder_progress with the changes from monitoring_derived.shredder_progress, and then delete monitoring_derived.shredder_progress because it shouldn't exist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2191)


relud:
marking this as a draft until after the change freeze

akkomar (approved):

relud:
actually, this is low risk, bug fix, and is effectively for deng and dsre only, so i'll merge now to avoid any complications of it now being left unfixed.

PR #4707, opened 2023-12-14
Bug 1868850 - Remove failing warn checks from UM

fbertsch:
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2201)

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/unified_metrics_v1/checks.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/unified_metrics_v1/checks.sql
@@ -11,8 +11,6 @@
 {{ value_length(column="client_id", expected_length=36, where="submission_date = @submission_date") }}
 
 #warn
--- The following field is sometimes NULL
--- "normalized_os_version"
 {{ not_null(columns=[
   "activity_segment",
   "normalized_app_name",
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/unified_metrics_v1/checks.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/unified_metrics_v1/checks.sql
@@ -10,6 +10,11 @@
 #fail
 {{ value_length(column="client_id", expected_length=36, where="submission_date = @submission_date") }}
 
+-- Commented out due to upstream duplication issue inside Fenix data
+-- which will cause this check to fail, see: bug(1803609).
+-- Once the duplication issue has been resolved, this check can be uncommented.
+-- #fail
+-- {{ is_unique(columns=["client_id"], where="submission_date = @submission_date") }}
 #warn
 {{ not_null(columns=[
   "activity_segment",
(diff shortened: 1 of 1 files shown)


sql/moz-fx-data-shared-prod/telemetry_derived/unified_metrics_v1/checks.sql
```diff
@@ -11,9 +11,8 @@
 {{ value_length(column="client_id", expected_length=36, where="submission_date = @submission_date") }}
 
 #warn
-{{ is_unique(columns=["client_id"], where="submission_date = @submission_date") }}
```
kik-kik, line 14: This table should have a grain of 1 entry per client per partition. If this checks fails then there's something going wrong, by just deleting the check we're ignoring the source issue.

sql/moz-fx-data-shared-prod/telemetry_derived/unified_metrics_v1/checks.sql
```diff
@@ -11,9 +11,8 @@
...
-
-#warn
+-- The following field is sometimes NULL
```
kik-kik, line 14:
```suggestion
```

sql/moz-fx-data-shared-prod/telemetry_derived/unified_metrics_v1/checks.sql
```diff
@@ -11,9 +11,8 @@
...
+-- "normalized_os_version"
```
kik-kik, line 15:
```suggestion
```

sql/moz-fx-data-shared-prod/telemetry_derived/unified_metrics_v1/checks.sql
fbertsch: If that's the case, this should be a `fail` warning and someone should investigate. I do not think we should go into the freeze with a check that's known to fail (and was known to fail historically)

fbertsch:
@ANich context here: https://bugzilla.mozilla.org/show_bug.cgi?id=1775137

tl;dr is we see duplicates in the Fenix tables due to a single client running multiple channels. Until this is fixed upstream this may continue to fail in UM. Until we fix upstream, I believe we should remove this check

kik-kik (approved):

sql/moz-fx-data-shared-prod/telemetry_derived/unified_metrics_v1/checks.sql
kik-kik: I think the reason we decided to go with `warn` now is to avoid breaking jobs and progressively switch over to `fail` mode for some checks once we get a feel for how many things fail and do not adhere to our assumptions about the data.

But I agree, this should become a fail check. However, before we do that switch here we should address the duplication inside of the Fenix tables. I'll push a commit to this branch commenting this check out along with a reference to Fenix duplication bug explaining why for now we chose to omit it.

fbertsch: That's perfect, it's all merged. Thanks for rounding this out!

PR #4714, opened 2023-12-15
tweak(): making the active_users_aggregates checks more lenient

kik-kik:
# tweak(): making the active_users_aggregates checks more lenient

I don't think in normal circumstances we should be seeing a variance of more than 10 between these tables.


kik-kik:
@lucia-vargas-a Do you think a valid case when the variance could be more than `10` between these tables?

sql_generators/active_users/templates/desktop_checks.sql
```diff
@@ -94,8 +94,17 @@ live_table_qdau_count AS (
 )
 SELECT
   IF(
-    (SELECT * FROM live_table_qdau_count) <> (SELECT * FROM qdau_sum),
-    ERROR("QDAU mismatch between aggregates table and live table"),
+    ABS((SELECT * FROM qdau_sum) - (SELECT * FROM live_table_qdau_count)) > 10,
+    ERROR(
+      CONCAT(
+        "QDAU mismatch between aggregates table and live table is greated than 10.",
+        " (live: ",
```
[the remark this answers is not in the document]
kik-kik: total here instead of live might be confusing, total of what? I did follow one of your previous suggestions and and try to indicate the source tables. Unfortunately, since multiple tables are used the best I can do is to try and provide the app_name and the table name. Now the message looks like this:
```"DAU mismatch between the firefox_ios `baseline_v1` live (all channels) and active_users_aggregates (`firefox_ios_derived.active_users_aggregates_v2`) tables is greated than 10."```

File: sql_generators/active_users/templates/desktop_checks.sql [hunk 2]
```diff
@@ -94,8 +94,18 @@ live_table_qdau_count AS (
...
+    ABS((SELECT * FROM qdau_sum) - (SELECT * FROM live_table_qdau_count)) > 10,
+    ERROR(
+      CONCAT(
+        "QDAU mismatch between the live (`telemetry_live.main_v5`) and active_users_aggregates (`{{ dataset_id }}.{{ table_name }}`) tables is greated than 10." ,
+        " Live table count: ",
+        (SELECT * FROM live_table_qdau_count),
+        " | active_users_aggregates (QDAU): ",
+        (SELECT * FROM qdau_sum),
+        " | Delta detected: ",
+        ABS((SELECT * FROM qdau_sum) - (SELECT * FROM live_table_qdau_count))
```
lucia-vargas-a, line 106: The delta is useful, thanks for adding it. So are the totals, I woudn't remove them

kik-kik: The totals are still there. I made a small tweak and added all channel tables to the message.

lucia-vargas-a (approved):

PR #4728, opened 2023-12-19
SVCSE-1595 Convert binary fields to hex strings in FxA database queries

akkomar:
We already convert binary `uid` fields to hex strings in these queries, this does the same with the rest to make querying easier.

Because schemas are changed, after merging this I will manually trigger the queries in Airflow so the next schema deployment doesn't fail. Tables will be overwritten because they are not day-partitioned and queries are ran with `"writeDisposition": "WRITE_TRUNCATE"`.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2238)


sql/moz-fx-data-shared-prod/accounts_db_external/fxa_oauth_clients_v1/query.sql
```diff
@@ -1,5 +1,5 @@
 SELECT
-  id,
+  TO_HEX(id) AS id,
```
sean-rose, line 2: The `accounts_db.fxa_oauth_clients` view is currently calling `TO_HEX()` on this table's `id` column, so this PR needs to update that view not to do that or the view query will fail with an error.

sean-rose (approved):
r+wc

PR #4731, opened 2023-12-19
Fb 1866834 v2 dynamic use counters

kwindau:
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2241)


sql/moz-fx-data-shared-prod/telemetry/fenix_and_firefox_use_counters/view.sql
```diff
@@ -19,4 +19,4 @@ SELECT
   metric,
   rate
 FROM
-  `moz-fx-data-shared-prod.fenix_derived.fenix_use_counters_v1`
```
ANich, line 22: This was recently added. Are these `_v1`'s going to stick around? If the output for the generated files look mostly the same. I don't think there's a reason to bump the version number

sql_generators/fenix_use_counters/__init__.py
```diff
@@ -0,0 +1,109 @@
...
+    """Copy metadata.yaml file to destination directory."""
+    shutil.copyfile(
+        FILE_PATH / "templates" / "metadata.yaml",
+        write_dir / project / dataset / destination_table / "metadata.yaml",
+    )
+
+def generate_schema(project, dataset, destination_table, write_dir):
+    """Generate the table schema."""
+    # get schema
+    table_schema = Schema.for_table(project, dataset, destination_table)
```
ANich, line 64: I'm not following the use of the production schema here? Won't this `sql_generator` be the only thing changing it?

sql_generators/firefox_desktop_use_counters/__init__.py
ANich, line 1: Most of the code between these two generators are the same. Maybe it's worth collapsing these into one generator?

sql_generators/firefox_desktop_use_counters/templates/templating.yaml
ANich, line 1: nit: this could have a more descriptive name: `config.yaml` or `use_counter_names` or something

sql_generators/firefox_desktop_use_counters/templates/query.sql
```diff
@@ -0,0 +1,130 @@
...
+                REGEXP_REPLACE(
+                  REGEXP_REPLACE(metric, 'use_counter_css_doc_', 'use.counter.css.doc.'),
+                  'use_counter_css_page_',
+                  'use.counter.css.page.'
+                ),
+                'use_counter_worker_dedicated',
+                'use.counter.worker.dedicated.'
+              ),
+              'use_counter_worker_shared_',
+              'use.counter.worker.shared.'
```
ANich, line 110: This seems like it could be a single regular expression? It's generally replacing `_` with `.` also looking for some specific strings.

sql_generators/firefox_desktop_use_counters/templates/query.sql
```diff
@@ -0,0 +1,130 @@
...
+    {% endfor %}
+  FROM
+    `moz-fx-data-shared-prod.firefox_desktop.use_counters`
+  WHERE
+    DATE(submission_timestamp) = @submission_date
+  GROUP BY
+    1,
+    2,
+    3,
+    4
```
ANich, line 24:
```suggestion
    submission_date,
    version_major,
    country,
    platform
```

sql_generators/fenix_use_counters/__init__.py
```diff
@@ -0,0 +1,109 @@
...
+
+    write_sql(
+        write_dir / project,
+        f"{project}.{dataset}.{destination_table}",
+        "query.sql",
+        reformat(template.render(**render_kwargs)),
+    )
+
+
+def generate_view(project, dataset, destination_table, write_dir):
```
ANich, line 34: This could also be a template `view.sql` file, so it could be reused for firefox desktop.

ANich (commented):
Overall I think you could collapse these two into a single generator and rely on configuration/hard-coded values to generate them. Also left a comment on whether a version bump is necessary.

sql_generators/firefox_desktop_use_counters/__init__.py
kwindau: Only issue is the columns could potentially differ between the 2, so they each would need separate template yaml files and separate SQL statements, otherwise would totally agree.

sql/moz-fx-data-shared-prod/telemetry/fenix_and_firefox_use_counters/view.sql
kwindau: Totally agree - I am planning to remove the V1 tables in a future release

sql_generators/firefox_desktop_use_counters/templates/query.sql
kwindau: It can't be, because it needs to leave the dots in the latter half to match the glean dictionary names (otherwise I most definitely would've done that, that would have been sooo much easier)  :)

sql_generators/fenix_use_counters/__init__.py
kwindau: Not sure I'm following - what do you mean?

sql_generators/fenix_use_counters/__init__.py
kwindau: So if you don't do this step, it won't actually make the table and it needs to make the table first  - the table actually stays the same format even as more columns get added to the template YAML file because the table is a pivoted format (where all the bajillion wide columns get pivoted into a single metric column)

sql_generators/firefox_desktop_use_counters/templates/templating.yaml
kwindau: The init.py file that is commonly reused across these "sql generated" repos has it built for it to be looking for the "templating YAML" file, so I just went with common practice so I wouldn't have to confuse future people - most people re-use this bit of init.py code and it is built to search for the "templating YAML" file

sql_generators/firefox_desktop_use_counters/templates/query.sql
kwindau: Pushed the fix

kwindau:
Closing to refactor into 1 single folder per Alex's request

PR #4745, opened 2023-12-22
DENG-2263 - Adding try catch to udf_js.decodeURIComponent

kwindau:
- https://mozilla-hub.atlassian.net/browse/DENG-2263

- There are occasionally rows where the URI decode function fails, and this makes the view unusable (throws an error if you try to query that row) - so we added a try / catch to return null if the decoding fails.

- Once this is in place, will be easier to look at examples of where the decoding is failing, so we can make any changes to that as needed in the future

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2265)

### DIFF
--- a/sql/moz-fx-data-shared-prod/udf_js/decode_uri_attribution/udf.sql
+++ b/sql/moz-fx-data-shared-prod/udf_js/decode_uri_attribution/udf.sql
@@ -30,6 +30,7 @@ const columns = [
   'ua',
   'variation',
 ];
+try { 
 return decodeURIComponent(decodeURIComponent(attribution))
   .split('&')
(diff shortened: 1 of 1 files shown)


ANich (approved):
Thanks for this fix, I meant to take a look but completely forgot about it (If it's EOD for you I'd hold off on merging til after the holidays) cc @SuYoungHong @irrationalagent who also use this

sql/moz-fx-data-shared-prod/udf_js/decode_uri_attribution/udf.sql
ANich, line 1: Could also use a test

PR #4746, opened 2023-12-27
DENG-2228 Add table to monitor total numbers of Mozilla Accounts

akkomar:
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2266)


sql/moz-fx-data-shared-prod/accounts_db_derived/accounts_aggregates_v1/query.sql
```diff
@@ -0,0 +1,65 @@
+-- Source tables of this query aren't partitioned and are overwritten daily.
+-- To make this query idempotent we need to make sure that the data for the
+-- current submission_date is not overwritten. If data for the current
```
ANich, line 3: What does `submission_date` mean in this context since it doesn't look like any pings are involved?

sql/moz-fx-data-shared-prod/accounts_db_derived/accounts_aggregates_v1/query.sql
```diff
@@ -0,0 +1,65 @@
...
+    submission_date,
+    active,
+    two_factor_auth_enabled,
+    recovery_keys_enabled,
+    total_accounts,
+  FROM
+    `moz-fx-data-shared-prod.accounts_db_derived.accounts_aggregates_v1`
+  WHERE
+    submission_date = @submission_date
+),
```
ANich, line 18: Why is the current data necessary here if only to UNION since the table is already partitioned by `submission_date`? Is the query meant to run multiple times per-day?

sql/moz-fx-data-shared-prod/accounts_db_derived/accounts_aggregates_v1/query.sql
akkomar: Yeah, no telemetry is involved here. To be precise, `submission_date` is "date when the database snapshot was made and aggregation was ran". I thought of calling this `day` but went with `submission_date` for consistency with other tables.

sql/moz-fx-data-shared-prod/accounts_db_derived/accounts_aggregates_v1/query.sql
akkomar: Query is meant to be run once a day. I wanted to make sure it won't overwrite past partitions if the task gets cleared in Airflow. Since source tables are not partitioned, re-runs could overwrite history with today's aggregate.
To ensure this doesn't happen I need to check if the partition we're trying to write to already has any data - hence the current data CTE.

sql/moz-fx-data-shared-prod/accounts_db_derived/accounts_aggregates_v1/query.sql
ANich: nit: in that case I'd probably call it something like `last_updated` or something

sql/moz-fx-data-shared-prod/accounts_db_derived/accounts_aggregates_v1/query.sql
ANich: Hmm it seems like the table doesn't really need to be partitioned then? With a `last_updated` column and a check to see if it ever changes by more than a day or two past `CURRENT_DATE` (in which case we could use BQ time travel)?

ANich (approved):
The partition part seems weird unless you're intentionally keeping all of history but looks fine otherwise. I'll leave the actual query logic review to @ksiegler1

ksiegler1:
I'm not sure I fully understand the discussion about partitioning vs. not, but I reviewed the query logic previously for constructing the table and it looks good to me.

ksiegler1 (approved):

akkomar:
> The partition part seems weird unless you're intentionally keeping all of history but looks fine otherwise. I'll leave the actual query logic review to @ksiegler1

Yes, my intent was to keep the history.

sql/moz-fx-data-shared-prod/accounts_db_derived/accounts_aggregates_v1/init.sql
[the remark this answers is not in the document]
akkomar: Yeah, you're right. Init.sql was useful during development but is not required here.

PR #4769, opened 2024-01-02
Support backfilling unpartitioned tables and non-date query parameters

ANich:
Closes https://github.com/mozilla/bigquery-etl/issues/4532

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -495,7 +495,8 @@ def info(ctx, name, sql_dir, project_id, cost, last_updated):
 
 def _parse_parameter(parameter: str, param_date: str) -> str:
     # TODO: Parse more complex parameters such as macro_ds_add
-    if "{{ds}}" not in parameter:
+    param_name, param_type, param_value = parameter.split(":")
+    if param_type == "DATE" and param_value != "{{ds}}":
         raise ValueError(f"Unable to parse parameter {parameter}")
 
     return f"--parameter={parameter.replace('{{ds}}', param_date)}"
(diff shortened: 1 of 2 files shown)


bigquery_etl/cli/query.py
```diff
@@ -493,6 +493,14 @@ def info(ctx, name, sql_dir, project_id, cost, last_updated):
         click.echo("")
 
 
+def _parse_parameter(parameter: str, param_date: str) -> str:
+    # TODO: Parse more complex parameters such as macro_ds_add
+    if "{{ds}}" not in parameter:
```
ANich, line 498: This should ideally use the same parsing logic as airflow does, as an import or something

bigquery_etl/cli/query.py
```diff
@@ -740,17 +769,14 @@ def backfill(
...
+        partitioning_type = None
+        if metadata.bigquery and metadata.bigquery.time_partitioning:
             partitioning_type = metadata.bigquery.time_partitioning.type
-        else:
-            partitioning_type = PartitionType.DAY
-            click.echo("Bigquery partitioning type not set. Using PartitionType.DAY")
 
         match partitioning_type:
-            case PartitionType.DAY:
+            case None | PartitionType.DAY:
```
ANich, line 779: Even if the table isn't partitioned, most queries still run daily, so format the dates in the same way

bigquery_etl/cli/query.py
```diff
@@ -493,6 +493,15 @@ def info(ctx, name, sql_dir, project_id, cost, last_updated):
         click.echo("")
 
 
+def _parse_parameter(parameter: str, param_date: str) -> str:
+    # TODO: Parse more complex parameters such as macro_ds_add
+    param_name, param_type, param_value = parameter.split(":")
+    if param_type == "DATE" and param_value != "{{ds}}":
+        raise ValueError(f"Unable to parse parameter {parameter}")
```
ANich, line 500: This should ideally use the same parsing logic as airflow does, as an import or something

scholtzan (approved):

fbertsch (approved):
LGTM. I'm a little concerned about the `_parse_parameter` function being a difficult-to-debug place, but since this is only used for ad-hoc purposes it should be fine to merge now and continue improving on.

Could you also update the `bqetl_backfill` DAG to support the new parameter?

bigquery_etl/cli/query.py
```diff
@@ -493,6 +493,15 @@ def info(ctx, name, sql_dir, project_id, cost, last_updated):
         click.echo("")
 
 
+def _parse_parameter(parameter: str, param_date: str) -> str:
+    # TODO: Parse more complex parameters such as macro_ds_add
+    param_name, param_type, param_value = parameter.split(":")
+    if param_type == "DATE" and param_value != "{{ds}}":
```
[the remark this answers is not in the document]
ANich: RE: this and top level comment, yeah this function should eventually be replaced by an Airflow util or a copy-paste of whatever Airflow's parameter parsing logic is

PR #4774, opened 2024-01-03
Remove dryrun from view validation in CI

ANich:
There's no need to dryrun views during view validation. Views can be dryrun directly from the dryrun CLI, and are already in CI part of the `dry-run-sql` CI job.

The view validation skips under "access_denied" in `bqetl_project.yaml` are really dryrun skips.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2300)

### DIFF
--- a/bigquery_etl/cli/view.py
+++ b/bigquery_etl/cli/view.py
@@ -11,13 +11,11 @@ from traceback import print_exc
 import rich_click as click
 
 from ..cli.utils import (
-    no_dryrun_option,
     parallelism_option,
     paths_matching_name_pattern,
     project_id_option,
     respect_dryrun_skip_option,
     sql_dir_option,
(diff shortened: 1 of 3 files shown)


bigquery_etl/cli/view.py
```diff
@@ -127,24 +108,11 @@ def validate(
...
-    else:
-        click.echo("Dry run skipped for view files.")
-
     click.echo("All views are valid.")
 
 
-def _view_is_valid(view):
-    return view.is_valid()
+def _view_is_valid(v: View) -> bool:
+    return v.is_valid()
```
ANich, line 115: `view` shadows the name from the outer scope

scholtzan (approved):

PR #4777, opened 2024-01-04
DAG docs - fix broken links and add tags to docs

lelilia:
At the moment the link in the Airflow DAG docs leads to the dag-file in `main` that no longer exists.
So switching the link to point to the file in the `generated-sql` branch should fix this.

And while I was here I thought it might be nice to add the tags to the docs, because inside the DAG in airflow one can't see what the impact is, which is annoying during triage.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2303)


kwindau (approved):
lgtm

lelilia:
ups need to update the tests as well ... on it

PR #4792, opened 2024-01-08
Add skip-existing option to ./bqetl query initialize

scholtzan:
Add a `--skip-existing` option to the `./bqetl query initialize` command. 
Explicitly setting `--skip-existing` ensures that existing materialized views won't get re-deployed and `init.sql` that miss the `IF NOT EXISTS` clause won't error out when running initialization.

Required for https://github.com/mozilla/telemetry-airflow/pull/1893

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2343)


bigquery_etl/cli/query.py
```diff
@@ -1412,12 +1419,19 @@ def initialize(ctx, name, sql_dir, project_id, dry_run, parallelism):
...
                 client = bigquery.Client(project=project)
 
+                if skip_existing:
+                    try:
+                        client.get_table(full_table_id)
+                        # table exists; skip initialization
+                        continue
+                    except NotFound:
+                        # continue with creating the table
+                        pass
```
sean-rose, line 1431: It'd be nice if both init approaches were consistent with table existence and emptiness checks, so I'd suggest having a single set of checks before the init approach `if` statement (with a variable to store the table existence result for the `is_init()` approach to determine whether to deploy the table schema), rather than having the logic duplicated slightly differently like this.

bigquery_etl/cli/query.py
```diff
@@ -1438,6 +1452,15 @@ def initialize(ctx, name, sql_dir, project_id, dry_run, parallelism):
                     if not dry_run:
                         job.result()
 
+    with ThreadPool(parallelism) as pool:
+        failed_initializations = [r for r in pool.map(_initialize, query_files) if r]
+
+    if len(failed_initializations) > 0:
+        click.echo("The following tables could not be deployed:")
+        for failed_deploy in failed_initializations:
+            click.echo(failed_deploy)
```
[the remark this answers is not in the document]
scholtzan: Yeah, I added some exception handling to `_initialize` to return the file paths that cause issues. We don't want for `Threadpool` to raise an exception and then stop _all_ initializations from happening

bigquery_etl/cli/query.py
scholtzan: I moved the `skip_existing` logic to the beginning of the method. We do need to check if the provided files can be initialized, so if they are either `init.sql` files or have a `is_init()` statement. Only then we check via `client.get_table()` if the table exists, otherwise if we pass a wildcard `*` to `./bqetl initialize` we'd end up calling `client.get_table()` for all tables we have which takes forever to run.

bigquery_etl/cli/query.py
```diff
@@ -1349,94 +1355,113 @@ def initialize(ctx, name, sql_dir, project_id, dry_run, parallelism):
         )
         sys.exit(1)
 
-    for query_file in query_files:
+    def _initialize(query_file):
+        client = bigquery.Client()
+        project, dataset, destination_table = extract_from_query_path(query_file)
```
sean-rose, line 1360: For the `init.sql` approach I think it was important for the client to be initialized with target project, because the table/view names in the `CREATE TABLE` and `CREATE MATERIALIZED VIEW` statements might not be fully qualified.
```suggestion
        project, dataset, destination_table = extract_from_query_path(query_file)
        client = bigquery.Client(project=project)
```
(this was previously done by instantiating a new client, but your latest revision removes that code)

bigquery_etl/cli/query.py
```diff
@@ -1349,94 +1355,113 @@ def initialize(ctx, name, sql_dir, project_id, dry_run, parallelism):
...
+        # check if the provided file can be initialized and whether existing ones should be skipped
+        if "is_init()" in sql_content or len(init_files) > 0:
+            if skip_existing:
+                try:
+                    table = client.get_table(full_table_id)
+                    # table exists; skip initialization
+                    return
+                except NotFound:
+                    # continue with creating the table
+                    pass
```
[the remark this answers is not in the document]
sean-rose: To allow that specifically for materialized views we could check if the table's `table_type` is "MATERIALIZED VIEW", but I still think it's worth applying the empty check for standard tables to avoid potential accidental data loss if the table already exists with data.  IMO if we want to allow overwriting existing populated tables we should have a `--force` option for that.

sean-rose (approved):
r+wc

bigquery_etl/cli/query.py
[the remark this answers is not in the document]
scholtzan: I guess that changes slightly how initialization has been working. Most `init.sql` have `CREATE OR REPLACE TABLE` to explicitly overwrite existing tables. If we'd check for data then in most cases `CREATE OR REPLACE TABLE`  would effectively become `CREATE TABLE` (unless we are overwriting an empty table). 

I opened a follow-up issue: https://github.com/mozilla/bigquery-etl/issues/4798. Might be worth discussing in the workgroup meeting to see if people have a preference.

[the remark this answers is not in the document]
scholtzan: Ah well, thinking about it having `--force` to overwrite existing tables does seem like a safer option. I'll add it to this PR

bigquery_etl/cli/query.py
```diff
@@ -1349,94 +1362,118 @@ def initialize(ctx, name, sql_dir, project_id, dry_run, parallelism):
...
 
+        # check if the provided file can be initialized and whether existing ones should be skipped
+        if "is_init()" in sql_content or len(init_files) > 0:
             try:
                 table = client.get_table(full_table_id)
-                if table.num_rows > 0:
+                if skip_existing:
+                    # table exists; skip initialization
+                    return
+                if not force and table.num_rows > 0:
```
scholtzan, line 1381: Added a `--force` option. I think it's also fine to use this to overwrite materialized views.

bigquery_etl/cli/query.py
```diff
@@ -1349,94 +1362,118 @@ def initialize(ctx, name, sql_dir, project_id, dry_run, parallelism):
...
-                        click.echo(f"Create destination table for {init_file}")
 
-                    job = client.query(init_sql, job_config=job_config)
+                        if "CREATE MATERIALIZED VIEW" in init_sql:
+                            click.echo(f"Create materialized view for {init_file}")
+                            # existing materialized view have to be deleted before re-creation
+                            view_name = query_file.parent.name
+                            client.delete_table(
+                                f"{project}.{dataset}.{view_name}", not_found_ok=True
+                            )
```
sean-rose, line 1457:
```suggestion
                            client.delete_table(full_table_id, not_found_ok=True)
```

bigquery_etl/cli/query.py
```diff
@@ -1321,11 +1321,24 @@ def _initialize_in_parallel(
...
+    "--skip-existing",
+    "--skip_existing",
+    help="Skip initialization for existing artifacts. "
+    + "This ensures that artifacts, like materialized views only get initialized if they don't already exist.",
+    default=False,
+    is_flag=True,
+)
+@click.option(
+    "--force/--noforce",
+    help="Run the initialization event if the destination table contains data.",
```
sean-rose, line 1334:
```suggestion
    help="Run the initialization even if the destination table contains data.",
```

PR #4744, opened 2023-12-22
Event flows monitoring

scholtzan:
Checklist for reviewer:

- [ ] Commits should reference a bug or github issue, if relevant (if a bug is referenced, the pull request should include the bug number in the title).
- [ ] If the PR comes from a fork, trigger integration CI tests by running the [Push to upstream workflow](https://github.com/mozilla/bigquery-etl/actions/workflows/push-to-upstream.yml) and provide the `<username>:<branch>` of the fork as parameter. The parameter will also show up
in the logs of the `manual-trigger-required-for-fork` CI task together with more detailed instructions.
- [ ] If adding a new field to a query, ensure that the schema and dependent downstream schemas have been updated.
- [ ] When adding a new derived dataset, ensure that data is not available already (fully or partially) and recommend extending an existing dataset in favor of creating new ones. Data can be available in the [bigquery-etl repository](https://github.com/mozilla/bigquery-etl), [looker-hub](https://github.com/mozilla/looker-hub) or in [looker-spoke-default](https://github.com/mozilla/looker-spoke-default/tree/e1315853507fc1ac6e78d252d53dc8df5f5f322b).

For modifications to schemas in restricted namespaces (see [`CODEOWNERS`](https://github.com/mozilla/bigquery-etl/blob/main/CODEOWNERS)):
- [ ] Follow the [change control procedure](https://docs.google.com/document/d/1TTJi4ht7NuzX6BPG_KTr6omaZg70cEpxe9xlpfnHj9k/edit#heading=h.ttegrcfy18ck)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2264)

PR #4795, opened 2024-01-09
GLAM historical aggregates tables

edugfilho:
ETL to refresh GLAM historical aggregates tables. Part of effort to make GLAM serve data from BQ.
@scholtzan I'd like your input here on using BQ DML instead of SQL. I tested and it runs pretty quickly (in 20 seconds at most) but I'm wondering if `bqetl` will handle it well (for instance, it doesn't have a target table as part of the job config, it's in the statement instead)

Checklist for reviewer:


scholtzan:
I wonder if we should rename `live_` to something else. There might be some confusion since we are usually referring to live tables when we're talking about the tables the ingestion pipeline dumps "realtime" data.

sql/glam-fenix-dev/glam_etl/firefox_desktop_glam_beta__live_v1/query.sql
```diff
@@ -0,0 +1,57 @@
+MERGE INTO
```
scholtzan, line 1: Instead of doing this in a `query.sql` file we'll need a `script.sql` file here. `query.sql` expect some `SELECT` result to be returned, which is not the case with MERGE INTO

scholtzan:
This is quite similar to the PR I'm currently working on related to events in case you want another example: https://github.com/mozilla/bigquery-etl/pull/4744/files#r1443511575

edugfilho:
> I wonder if we should rename `live_` to something else. There might be some confusion since we are usually referring to live tables when we're talking about the tables the ingestion pipeline dumps "realtime" data.

would `extract` be a good choice?

scholtzan:
Something like `firefox_desktop_glam_release__extract_probe_counts_serving_v1` since it's basically a copy of `firefox_desktop_glam_release__extract_probe_counts_v1` but retaining all historical data too.

edugfilho:
> Something like `firefox_desktop_glam_release__extract_probe_counts_serving_v1` since it's basically a copy of `firefox_desktop_glam_release__extract_probe_counts_v1` but retaining all historical data too.

I see. I'm not the biggest fan of `probe_counts` because, despite being the names of the ETL tables, it doesn't make a lot of sense to me. I don't know why they were named that way, so I'd like to give this something more meaningful. 
On GLAM postgres the tables are simply called `glam_[fog | fenix | desktop]_{channel}_aggregation`, for reference, which I find way better.  What do you think?

scholtzan:
yup something like `glam_[fog | fenix | desktop]_{channel}_aggregates_v<n>` sounds good to me

scholtzan (approved):

PR #4848, opened 2024-01-17
Restrict derived view schema generation to views with upstream schema files and directly copy reference schemas for simple views.

ANich:
View schema files are only used for adding descriptions, this PR adds that assumption to the generator. 

Previously it was attempting to generate schema files for all views, including where they wouldn't/didn't have descriptions and so wouldn't be useful. This added a lot of extra dryruns to CI. 

Also copies the reference schema directly for "simple"/default views.  

A few schema files will be removed from this PR (see diff comment), but these aren't currently used for anything.

### DIFF
--- a/sql_generators/derived_view_schemas/__init__.py
+++ b/sql_generators/derived_view_schemas/__init__.py
@@ -9,22 +9,14 @@ from pathos.multiprocessing import ProcessingPool
 
 from bigquery_etl.cli.utils import use_cloud_function_option
 
-NON_USER_FACING_DATASET_SUBSTRINGS = (
-    "_derived",
-    "_external",
-    "_bi",
-    "_restricted",
-    "udf",
-)
-
 
 def _generate_view_schema(sql_dir: Path, view_directory: Path) -> None:
     import logging
 
-    from view import View
-
+    from bigquery_etl.format_sql.formatter import reformat
     from bigquery_etl.metadata.parse_metadata import Metadata
     from bigquery_etl.schema import Schema
+    from bigquery_etl.view import View
 
     logging.basicConfig(format="%(levelname)s (%(filename)s:%(lineno)d) - %(message)s")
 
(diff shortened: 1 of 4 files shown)


sql_generators/derived_view_schemas/__init__.py
```diff
@@ -2,136 +2,102 @@
...
-        # Fully qualify the reference:
-        elif len(parts) == 2:
-            reference_project_id = target_project
-            reference_dataset_id, reference_table_id = parts
-        elif len(parts) == 1:
-            reference_project_id = target_project
-            reference_dataset_id = target_dataset
-            reference_table_id = parts[0]
-        else:
-            return
```
ANich, line 64: This logic is all available more concisely in the `View` class used below

sql_generators/derived_view_schemas/__init__.py
```diff
@@ -2,136 +2,102 @@
...
 
         bigquery_metadata = reference_metadata.bigquery
-        if bigquery_metadata is None:
-            logging.warning(
-                f"No bigquery metadata at {ref_path}, unable to get partition key."
-            )
-            return
-
-        partition_metadata = bigquery_metadata.time_partitioning
-        if partition_metadata is None:
```
ANich, line 89: Same error as the one below no need to duplicate logging

scholtzan (approved):

sean-rose:
> View schema files are only used for adding descriptions, this PR adds that assumption to the generator.
> 
> Previously it was attempting to generate schema files for all views, including where they wouldn't/didn't have descriptions and so wouldn't be useful.

@ANich it turns out view schema files are also used to [determine whether the view schema has changed](https://github.com/mozilla/bigquery-etl/blob/09bc346dec8302600465fbee9311b0f0d45f36bb/bigquery_etl/view/__init__.py#L270-L278), and with view publishing logic defaulting to only publishing views that have changed this means views without `schema.yaml` files are no longer getting published when their schema changes (e.g. when upstream table schemas change).

I'd suggest reverting the part of this PR about restricting view schema generation, or finding some other suitable workaround so that views are published consistently.

Issue #4873, opened 2024-01-23
dataset_metadata.yaml not correctly updated when deprecated: true


scholtzan:
`workgroup_access` in the `dataset_metadata.yaml` files doesn't get updated correctly to `[]` nor does `default_table_workgroup_access` get set to `workgroup:mozilla-confidential` when `deprecated: true` is set in one of the query `metadata.yaml` files.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2469)

whd:
I think in the PR this isn't an issue, but the intent should be for `default_table_workgroup_access` to get set to whatever `workgroup_access` was as opposed to `workgroup:mozilla-confidential` specifically. For instance, a deprecated table in https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/contextual_services/dataset_metadata.yaml should wind up with the generated dataset metadata looking like:
```yaml
workgroup_access: []
default_table_workgroup_access:
- role: roles/bigquery.dataViewer
  members:
  - workgroup:contextual-services
```

scholtzan:
Right, `default_table_workgroup_access` has been getting set correctly to whatever `workgroup_access` was. The missing piece was setting  `workgroup_access: []` for datasets with deprecated tables.

PR #4874, opened 2024-01-23
Fix dataset deprecation metadata

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/4873
Sets `workgroup_access` for `dataset_metadata.yaml` files to `[]` if it contains deprecated tables.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2470)

### DIFF
--- a/bigquery_etl/cli/metadata.py
+++ b/bigquery_etl/cli/metadata.py
@@ -58,16 +58,19 @@ def update(name: str, sql_dir: Optional[str], project_id: Optional[str]) -> None
             dataset_metadata.default_table_workgroup_access = (
                 dataset_metadata.workgroup_access
             )
-        dataset_metadata.write(dataset_metadata_path)
+        
         if table_metadata.deprecated:
             # set workgroup: [] if table has been tagged as deprecated
             # this overwrites existing workgroups
             table_metadata.workgroup_access = []
+            dataset_metadata.workgroup_access = []
         else:
             if table_metadata.workgroup_access is None:
                 table_metadata.workgroup_access = (
                     dataset_metadata.default_table_workgroup_access
                 )
+
+        dataset_metadata.write(dataset_metadata_path)
         table_metadata.write(table_metadata_file)
         click.echo(f"Updated {table_metadata_file}")
     return None
--- a/bigquery_etl/cli/metadata.py
+++ b/bigquery_etl/cli/metadata.py
@@ -58,7 +58,7 @@ def update(name: str, sql_dir: Optional[str], project_id: Optional[str]) -> None
             dataset_metadata.default_table_workgroup_access = (
                 dataset_metadata.workgroup_access
             )
-        
+
         if table_metadata.deprecated:
             # set workgroup: [] if table has been tagged as deprecated
             # this overwrites existing workgroups
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/active_profiles_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/active_profiles_v1/metadata.yaml
@@ -1,0 +1,2 @@
+friendly_name: Active Profiles
+deprecated: true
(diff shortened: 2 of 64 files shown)


whd:
I want to reiterate from https://github.com/mozilla/bigquery-etl/pull/4869 that we should decide how to handle partial metadata in `telemetry_derived` before we land a fix that affects that dataset.

scholtzan:
Blocked until missing `metadata.yaml` files have been added as described in https://github.com/mozilla/bigquery-etl/pull/4869#issuecomment-1907031465

scholtzan:
Alright I looked through the tables that exist in `telemetry_derived` but don't have a `metadata.yaml` file and added ones for those that are still in active use

whd:
> Alright I looked through the tables that exist in `telemetry_derived` but don't have a `metadata.yaml` file and added ones for those that are still in active use

Would it make sense at this time to also add `metadata.yaml` files with `deprecated: true` for the ones that aren't in active use so that they can go through the new deprecation process?

scholtzan:
Yeah, that makes sense. Added a metadata file for those too

whd:
> Check for non-deprecated non-bqetl-managed tables in the dataset. Should this be done by comparing the tables that exist under the dataset and the bqetl repo?

Basically, yes (specifically, the private generated branch as it has metadata that is meant to be canonical).

> Add metadata.yaml files for all these tables. 
> How do we know where these tables come from? 

We might not (in general). For a table that isn't managed in bqetl/docker-etl, with metadata like creation/modification times that are very old: these are strong signals that a table should be deprecated. For instance, the table `telemetry_derived.eng_workflow_build_parquet_v1` I referenced above was manually copied into place and I happened to have the bugzilla ticket where that was created, but doing archaeology for all the things could be more effort than it's worth.

> Is there a programmatic way to do this step if there are many tables?

You could definitely write a script or automation to do this. In fact, I would say that one of the motivations behind the deprecation work and bq usage tables is to allow us to programatically deprecate tables based on specific criteria. Taking `telemetry_derived.eng_workflow_build_parquet_v1` again as an example, we should be able to tell (a) it was created over 3 years ago and has not been updated since (b) doesn't exist in bqetl (c) from bq usage tables hasn't been queried in the last year and from this information be confident that this table should be deprecated and eventually deleted.

>  How did you determine which tables were not in active use and should add metadata.yaml files with deprecated: true?

I'm assuming Anna more or less did something like I described above. A rule of thumb might be if we know a table is being actively updated by e.g. bqetl or docker-etl it probably shouldn't be deprecated. Certainly I'd expect the bq usage tables to be helpful in determining whether these tables are actively used even if they aren't updated.

FWIW, I'd expect `telemetry_derived` to be one of the higher effort datasets to audit. In most newer datasets e.g. everything glean I wouldn't expect there to be nearly as much "unmanaged" stuff that we don't have clear provenance information for. In theory automation could find some tables that were manually created by a DE that never got checked into bqetl or stuff like that but I'd imagine the number of such tables to be very small.

whd:
Now that we have more or less complete metadata for tables and views, the deploy ordering issue I mentioned yesterday is actually just the result of an [optimization](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/tf/modules/derived-acls/main.tf#L54) in the ACL deployment pipeline to only deploy table ACLs when they are not redundant. I'm investigating removing the optimization, which will create 1000s of ACLs but will enable seamless deprecation without deployment order changes. This is probably going to require changes to the way we use the inventory API to reduce the number of individual calls but is worth trying to do before investigating more complicated alternatives.

It seems the ability to update the branch from the UI is gone since we no longer require branches to be up to date. I'll try to get to rebasing and deploying this tomorrow, using the manual ACL deployment if the inventory API changes are too complicated to land in a day.

PR #4891, opened 2024-01-26
DO NOT MERGE: feat(): Update fxa_users_ views to use v2 of the corresponding tables

kik-kik:
# feat(): Update fxa_users_ views to use v2 of the corresponding tables

depends on: https://github.com/mozilla/bigquery-etl/pull/4893

Prior to merging this PR we need to update the scheduling configuration for the corresponding v2 table:
https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v2/metadata.yaml#L14-L22

Also, we need to make sure nothing will break in Looker. Currently, it appears both `fxa_users_first_seen` and `fxa_users_last_seen` are used inside Looker:
https://github.com/mozilla/lookml-generator/blob/main/custom-namespaces.yaml#L516-L525C1

=> https://github.com/mozilla/looker-hub/blob/main/firefox_accounts/views/fxa_first_seen_table.view.lkml#L7-L11
This is where the first seen table is referenced, but it looks like the field we want to remove is actually hidden in this Looker table view.

---

Last seen appear to be utilized here:
https://github.com/mozilla/looker-hub/blob/main/firefox_accounts/views/growth_accounting.view.lkml

And the v2 version of this table is missing fields which are used:
- days_since_seen_no_monitor (want to deprecate / remove)
- language
- app_version
- country
- os_name
- os_version

We should consider if these fields should instead be retrieved from the daily table instead.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2496)

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_accounts/fxa_users_daily/view.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts/fxa_users_daily/view.sql
@@ -4,4 +4,4 @@ AS
 SELECT
   *
 FROM
-  `moz-fx-data-shared-prod.firefox_accounts_derived.fxa_users_daily_v1`
+  `moz-fx-data-shared-prod.firefox_accounts_derived.fxa_users_daily_v2`
--- a/sql/moz-fx-data-shared-prod/firefox_accounts/fxa_users_first_seen/view.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_accounts/fxa_users_first_seen/view.sql
@@ -2,6 +2,8 @@ CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.firefox_accounts.fxa_users_first_seen`
 AS
 SELECT
-  *
+  *,
+  -- TODO: remove the `services_used` field once we confirm no downstream usage.
+  [CAST(NULL AS STRING)] AS services_used,
 FROM
-  `moz-fx-data-shared-prod.firefox_accounts_derived.fxa_users_first_seen_v1`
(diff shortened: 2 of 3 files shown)


ksiegler1 (commented):
I'm wondering if the scheduling configurations for the v2 tables will need to be updated, since we are in the process of deriving these tables (`users_[daily, first_seen, last_seen`) from the new Glean activity events.

 I think in the near future we are planning to point the Looker views towards those derived tables e.g the Growth Accounting Explore linked in the description would use `moz-fx-data-shared-prod.accounts_backend_derived.users_services_last_seen_v1`. Does that sound correct @akkomar?

kik-kik:
> I'm wondering if the scheduling configurations for the v2 tables will need to be updated, since we are in the process of deriving these tables (`users_[daily, first_seen, last_seen`) from the new Glean activity events.
> 
> I think in the near future we are planning to point the Looker views towards those derived tables e.g the Growth Accounting Explore linked in the description would use `moz-fx-data-shared-prod.accounts_backend_derived.users_services_last_seen_v1`. Does that sound correct @akkomar?

So if we follow a similar approach where we derive this data from the server logs then we will need to merge the sources together to ensure there's some kind of continuation of the historical data and the new data. But I do not think the scheduling would be affected per say and the query will continue to run (at least for now). Unless you mean something different by scheduling here @ksiegler1?

akkomar:
Yeah, eventually we want to move all dashboards to Glean telemetry. I guess here the question is if we need data continuity for (some of) them? I was assuming that we do, so putting these old tables in order would make sense.

kik-kik:
Outdated and no longer valid change.

PR #4917, opened 2024-01-30
Package bqetl and publish to PyPI

scholtzan:
This is replacing `setup.py` with the recommended `pyproject.toml` for packaging and publishing the `bigquery-etl` CLI tooling to PyPI. CI will be set up to push a new package version whenever a new calver tag/a new tagged release is created. 

Packages will be published to https://pypi.org/project/mozilla-bigquery-etl/

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2540)


kwindau (approved):

whd:
I just accepted some requests for the `mozdataops` pypi machine user to be added as owner to the pypi package which I assume is related to this work.

> CI will be set up to push a new package version whenever a new calver tag/a new tagged release is created.

OIDC / trusted publishing is now the preferred way of publishing packages to PyPi, but is currently only integrated with GHA: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/ . In the event we want to keep this is CircleCI we could potentially set up OIDC+GSM to store PyPi API tokens securely (similar to what we're planning to do to allow triggering Airflow DAGs from CI) but it's also probably fine to store the token in CI if we want to.

I haven't really wrapped my head around the implications of splitting the tooling and sql but I know it's been planned for a while and I expect the pypi package won't affect docker-based systems which comprise the majority of ops logic and production usage.

scholtzan:
> I just accepted some requests for the mozdataops pypi machine user to be added as owner to the pypi package which I assume is related to this work.

Yes, I added mozdataops to some projects I was the sole owner of.

> In the event we want to keep this is CircleCI we could potentially set up OIDC+GSM to store PyPi API tokens securely (similar to what we're planning to do to allow triggering Airflow DAGs from CI) but it's also probably fine to store the token in CI if we want to.

Having all the CI actions in one place is slightly preferable. So I'd keep this in CircleCI. Setting up OIDC+GSM to store PyPi API tokens would make sense here. I guess until we have this set up we can keep the token stored in CircleCI.

> I haven't really wrapped my head around the implications of splitting the tooling and sql but I know it's been planned for a while and I expect the pypi package won't affect docker-based systems which comprise the majority of ops logic and production usage.

Having this published as a package doesn't have any immediate impact on our docker-based setup. It's mainly for user convenience, especially when working in private-bigquery-etl.

PR #4921, opened 2024-01-30
Speed up generate-sql

scholtzan:
This is to speed up the `generate-sql` task. With the changes it finishes in about ca 7 min vs the usual 14 min

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2550)

### DIFF
--- a/bigquery_etl/cli/metadata.py
+++ b/bigquery_etl/cli/metadata.py
@@ -53,23 +53,35 @@ def update(name: str, sql_dir: Optional[str], project_id: Optional[str]) -> None
         dataset_metadata = DatasetMetadata.from_file(dataset_metadata_path)
         table_metadata = Metadata.from_file(table_metadata_file)
 
+        dataset_metadata_updated = False
+        table_metadata_updated = False
+
         # set dataset metadata default_table_workgroup_access to table_workgroup_access if not set
         if not dataset_metadata.default_table_workgroup_access:
             dataset_metadata.default_table_workgroup_access = (
                 dataset_metadata.workgroup_access
             )
+            dataset_metadata_updated = True
 
         if table_metadata.deprecated:
             # set workgroup: [] if table has been tagged as deprecated
             # this overwrites existing workgroups
             table_metadata.workgroup_access = []
+            table_metadata_updated = True
             dataset_metadata.workgroup_access = []
+            dataset_metadata_updated = True
         else:
             if table_metadata.workgroup_access is None:
                 table_metadata.workgroup_access = (
                     dataset_metadata.default_table_workgroup_access
                 )
+                table_metadata_updated = True
 
-        dataset_metadata.write(dataset_metadata_path)
-        table_metadata.write(table_metadata_file)
+        if dataset_metadata_updated:
+            dataset_metadata.write(dataset_metadata_path)
+            click.echo(f"Updated {dataset_metadata_path}")
+        if table_metadata_updated:
+            table_metadata.write(table_metadata_file)
+            click.echo(f"Updated {table_metadata_file}")
+
     return None
--- a/bigquery_etl/cli/metadata.py
+++ b/bigquery_etl/cli/metadata.py
@@ -42,7 +42,7 @@ def update(name: str, sql_dir: Optional[str], project_id: Optional[str]) -> None
     table_metadata_files = paths_matching_name_pattern(
         name, sql_dir, project_id=project_id, files=["metadata.yaml"]
     )
-    dataset_metadata_path = None
+
     # create and populate the dataset metadata yaml file if it does not exist
     for table_metadata_file in table_metadata_files:
         dataset_metadata_path = (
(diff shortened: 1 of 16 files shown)


bigquery_etl/cli/metadata.py
```diff
@@ -72,5 +72,4 @@ def update(name: str, sql_dir: Optional[str], project_id: Optional[str]) -> None
 
         dataset_metadata.write(dataset_metadata_path)
         table_metadata.write(table_metadata_file)
-        click.echo(f"Updated {table_metadata_file}")
```
sean-rose, line 75: I'm guessing you're removing this because it's saying basically every one is being updated?

Instead of removing the logging IMO it'd be better to improve this logic to keep track of whether the metadata was actually updated, and then only write the files and log when they actually were (which should also be a bit faster since fewer files would be written).
```python
        dataset_metadata_updated = False
        table_metadata_updated = False

        # set dataset metadata default_table_workgroup_access to table_workgroup_access if not set
        if not dataset_metadata.default_table_workgroup_access:
            dataset_metadata.default_table_workgroup_access = (
                dataset_metadata.workgroup_access
            )
            dataset_metadata_updated = True

        if table_metadata.deprecated:
            # set workgroup: [] if table has been tagged as deprecated
            # this overwrites existing workgroups
            if table_metadata.workgroup_access:
                table_metadata.workgroup_access = []
                table_metadata_updated = True
            if dataset_metadata.workgroup_access:
                dataset_metadata.workgroup_access = []
                dataset_metadata_updated = True
        elif table_metadata.workgroup_access is None:
            table_metadata.workgroup_access = (
                dataset_metadata.default_table_workgroup_access
            )
            table_metadata_updated = True

        if dataset_metadata_updated:
            dataset_metadata.write(dataset_metadata_path)
            click.echo(f"Updated {dataset_metadata_path}")
        if table_metadata_updated:
            table_metadata.write(table_metadata_file)
            click.echo(f"Updated {table_metadata_file}")
```

bigquery_etl/cli/query.py
```diff
@@ -1530,49 +1530,53 @@ def _initialize(query_file):
...
+    dataset_id = query_file.parent.parent.name
+    project_id = query_file.parent.parent.parent.name
+
+    jinja_params = {
+        **{
+            "project_id": project_id,
+            "dataset_id": dataset_id,
+            "table_name": table_name,
+        },
+    }
```
sean-rose, line 1557: This could be simplified.
```suggestion
    jinja_params = {
        "project_id": project_id,
        "dataset_id": dataset_id,
        "table_name": table_name,
    }
```

sql_generators/glean_usage/common.py
```diff
@@ -185,6 +185,7 @@ def __init__(self):
         self.across_apps_enabled = True
         self.cross_channel_template = "cross_channel.view.sql"
         self.base_table_name = "baseline_v1"
+        self.parallelism = 8
```
sean-rose, line 188: `bqetl generate glean_usage` has a `--parallelism` parameter, so IMO it'd be better to use that rather than hard-coding the parallelism.

sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -13,6 +13,7 @@
 
 from jinja2 import Environment, FileSystemLoader
 from mozilla_schema_generator.glean_ping import GleanPing
+from pathos.multiprocessing import ThreadingPool
```
sean-rose, line 16: The other SQL generators that do parallel processing use `ProcessingPool` from `pathos.multiprocessing`, not `ThreadingPool`.  Is there a benefit to using `ThreadingPool`?

Also, why do the SQL generators use the third-party `pathos` package for parallel processing, while all other bigquery-etl code seems to use the built-in `multiprocessing` module?

sean-rose (approved):
r+wc

Twice as fast is an impressive speed-up!

sql_generators/glean_usage/glean_app_ping_views.py
scholtzan: this was the reason we ended up using pathos: https://github.com/mozilla/bigquery-etl/pull/2567/files#r768227839

_"The standard multiprocessing module has problems with executing functions in parallel that are not part of a package (resulting in pickling errors)."_

I tried using `ProcessingPool` here, but ran into two issues: one there were some issues with pickling the `GleanAppPingViews` class due to the `self` references, and second `ProcessingPool`s cannot be nested. `glean_usage` is using `ProcessingPool` to kick off generating the different tables per app.

sean-rose: Interesting.  I think that'd be worthy of an explanatory comment where `ThreadingPool` is being used in `GleanAppPingViews` (at least about why `ProcessingPool` couldn't be used).

PR #4926, opened 2024-01-31
events stream: Expose additional metrics as a new JSON column

badboy:
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2556)

### DIFF
--- a/sql_generators/glean_usage/templates/events_stream_v1.query.sql
+++ b/sql_generators/glean_usage/templates/events_stream_v1.query.sql
@@ -37,7 +37,11 @@ WITH base AS (
     event.category as event_category,
     event.name as event_name,
     ARRAY_TO_STRING([event.category, event.name], '.') AS event, -- handles NULL values better
-    `mozfun.json.from_map`(event.extra) AS event_extra
+    `mozfun.json.from_map`(event.extra) AS event_extra,
+    JSON_STRIP_NULLS(
+      TO_JSON(metrics),  -- expose as easy to access JSON column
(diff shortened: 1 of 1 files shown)


sql_generators/glean_usage/templates/events_stream_v1.query.sql
```diff
@@ -37,7 +37,11 @@ WITH base AS (
     event.category as event_category,
     event.name as event_name,
     ARRAY_TO_STRING([event.category, event.name], '.') AS event, -- handles NULL values better
-    `mozfun.json.from_map`(event.extra) AS event_extra
+    `mozfun.json.from_map`(event.extra) AS event_extra,
+    JSON_STRIP_NULLS(
+      TO_JSON(metrics),  -- expose as easy to access JSON column
+      remove_empty => true -- no reason to keep empty objects around
```
badboy, line 43: There's also an option `include_arrays` that defaults to `TRUE`, which removes `null` from arrays.
So `[1, null, 2]` will turn into `[1, 2]`. 
We shouldn't have that data I think, but would it be safer to `include_arrays => FALSE` still?

quiiver: I would be incredibly surprised if we needed that but it also doesn't hurt to include the parameter.

quiiver (approved):
I'll leave it up to your decision if you want to add the `include_arrays => FALSE` parameter. otherwise, lgtm!

PR #4939, opened 2024-01-31
Path Filtering in CI

scholtzan:
This PR introduces path filtering in the CI. To use path filtering the CI config needs to be split up into two files:
* `.circleci/config.yml` which checks which paths were changed in the pull-request and sets pipeline parameters that function as flags
* `.circleci/workflows.yml` which contains the CI tasks definitions and the workflows that should be run depending on which flag is set

Currently, there are 3 different flags:
* `validate-bqetl` which triggers the workflow to run integration and unit tests on the `bqetl` tooling if a tooling related file changes
* `validate-sql` which runs SQL checks, stage deploys etc when a SQL file or generator changes
* `validate-routine` which runs UDF checks in addition to SQL checks

Workflows can be triggered manually using the pipeline parameters

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2597)


.circleci/config.yml
```diff
@@ -1,832 +1,40 @@
...
+      and:
+        - not:
+            equal: [main, << pipeline.git.branch >>]
+    jobs:
+      - path-filtering/filter:
+          base-revision: main
+          config-path: .circleci/ci-config.yml
+          mapping: |
+            bigquery_etl/.* validate-bqetl true
+            tests/.* validate-bqetl true
```
ANich, line 33: I suspect there might be some changes here that it would be good to run through the sql jobs for - changes to how we handle schemas, metadata, views, etc.

.circleci/ci-config.yml
```diff
@@ -0,0 +1,934 @@
...
+      - deploy-changes-to-stage:
+          requires:
+            - generate-sql
+      - dry-run-sql:
+          requires:
+            - deploy-changes-to-stage
+      - test-sql:
+          context: data-eng-circleci-tests
+          requires:
+            - deploy-changes-to-stage
```
ANich, line 741: IIRC these take the majority of CI time. A short version of the proposal I was drafting:

It should be possible to run the sql jobs only on the affected artifacts. For e.g. if I change `table_a` which has upstream or downstream `table_b`, `udf_1`, `sql_generator_1`, and `view_a`, then only run the necessary generators and stage the necessary artifacts similar to the way `deploy-to-stage` works now. This would require some changes to sql generators since sometimes the dependencies aren't known until the generator gets run

kwindau (approved):

PR #4981, opened 2024-02-07
Bug 1879039 - events stream: Convert nested maps in metrics

badboy:
I'm not entirely sure that this works. Can we add a new UDF and use it in the same PR?

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2659)

### DIFF
--- a/sql/mozfun/json/from_nested_map/udf.sql
+++ b/sql/mozfun/json/from_nested_map/udf.sql
@@ -1,0 +1,34 @@
+-- Definition for json.from_nested_map
+CREATE OR REPLACE function json.from_nested_map(input JSON)
+RETURNS json
+LANGUAGE js
+AS
+  """
+  if (input && Object.keys(input).length) {
+    for (const k in input) {
+      if (input[k] && input[k].length) {
+        input[k] = input[k].reduce((acc, {key, value}) => {
+          acc[key] = value;
+          return acc;
+        }, {});
+      }
+    }
+  }
+  return input;
+""";
+
+-- Tests
+SELECT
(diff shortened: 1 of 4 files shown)


sql/mozfun/json/from_nested_map/udf.sql
```diff
@@ -0,0 +1,28 @@
...
+    }
+    return input;
+  }
+  return null;
+""";
+
+-- Tests
+SELECT
+  assert.null(json.from_nested_map(NULL)),
+  assert.json_equal(JSON '{ "metric": {} }', json.from_nested_map(JSON '{ "metric": [] }')),
```
quiiver, line 24: I would suggest adding a few more test cases.

For example what would be the expected out come for the following: 
```json
{"metric": [1, 2, 3]}
```
```json
{"metric": "value"}
```
```json
{"metric": [{"key": "value"}]}
```

sql/mozfun/json/from_nested_map/udf.sql
```diff
@@ -0,0 +1,28 @@
+-- Definition for json.from_nested_map
+CREATE OR REPLACE function json.from_nested_map(input JSON)
+RETURNS json
+LANGUAGE js
+AS
+  """
+  if (input && Object.keys(input).length) {
+    for (const k in input) {
+      if (input[k] && input[k].length) {
```
quiiver, line 9: if `input` is an empty object this will return `null` correct? probably best to just return `input` at the bottom and never return `null`

quiiver (approved):
few comments which are not blockers. lgtm!

PR #4997, opened 2024-02-08
Don't overwrite existing `gcp.json` files unless that's actually necessary

sean-rose:
#4924 didn't eliminate the "_File /tmp/gcp.json is not a valid json file_" errors we keep getting sporadically in CI jobs, and I think the reason is that the individual parallelized tests that call the entrypoint script weren't getting their own unique `gcp.json` files and instead were using (and overwriting) the same "unique" `gcp.json` file created by the initial entrypoint script run that kicked off the overall test run, since the `GOOGLE_APPLICATION_CREDENTIALS` environment variable was set by that initial entrypoint script.

This PR takes a different tack and goes back to using one `gcp.json` file, but avoids overwriting that `gcp.json` file unless it doesn't contain the expected credentials.

---
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2691)

### DIFF
--- a/script/entrypoint
+++ b/script/entrypoint
@@ -8,12 +8,12 @@ if [ -n "$GCLOUD_SERVICE_KEY" ]; then
     # Google's client libraries will check for GOOGLE_APPLICATION_CREDENTIALS
     # and use a file in that location for credentials if present;
     # See https://cloud.google.com/docs/authentication/production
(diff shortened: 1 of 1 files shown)


whd (approved):
Mikael mentioning `base64` makes me thing it might be preferable to compare file hashes e.g. `md5sum` instead of `cat`ing the full credentials for comparison.

PR #5034, opened 2024-02-13
Add BigQuery schema conversion util

ANich:
Adds utils for converting from BigQuery schema to local `Schema` and vice-versa.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2757)

### DIFF
--- a/bigquery_etl/schema/__init__.py
+++ b/bigquery_etl/schema/__init__.py
@@ -10,6 +10,7 @@ import attr
 import yaml
 from google.api_core.exceptions import NotFound
 from google.cloud import bigquery
+from google.cloud.bigquery import SchemaField
 
 from .. import dryrun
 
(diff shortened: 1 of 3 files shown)


kwindau (approved):
lgtm

bigquery_etl/schema/__init__.py
```diff
@@ -282,3 +283,14 @@ def to_json_file(self, json_path: Path):
     def to_json(self):
         """Return the schema data as JSON."""
         return json.dumps(self.schema)
+
+    def to_bigquery_schema(self) -> List[SchemaField]:
+        """Get the BigQuery representation of the schema."""
+        with NamedTemporaryFile(suffix=".json") as f:
+            self.to_json_file(Path(f.name))
+            return [SchemaField.from_api_repr(field) for field in json.load(f)]
```
sean-rose, line 291: I see this is just moving/de-duping the code formerly in `query.py`, but it seems like the temp file isn't necessary.
```suggestion
        return [SchemaField.from_api_repr(field) for field in self.schema["fields"]]
```

sean-rose (approved):
r+wc

bigquery_etl/schema/__init__.py
ANich: This works yeah

PR #5039, opened 2024-02-14
Support waiting for tables/partitions to exist before running ETLs

sean-rose:
This feature is configured via new `depends_on_tables_existing` and `depends_on_table_partitions_existing` ETL scheduling metadata sections, which result in Airflow sensor tasks to wait for the table/partition being added to the DAG and set as upstream dependencies of the ETL task.

This is an alternative to the "empty check" ETL approach for waiting on externally synced data needed by the ETL to become available, at least when the external data tables are partitioned or date-sharded.

As an initial test this PR replaces the empty-check ETLs for Google Search Console with the new approach.  Assuming things go smoothly once deployed, then we could look into replacing other empty-check ETLs.

---
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2763)


bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -304,6 +316,69 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
...
+        deferrable=True,
+        {% if table_sensor_task.poke_interval != None -%}
+        poke_interval={{ table_sensor_task.poke_interval | format_timedelta | format_repr }},
+        {% else -%}
+        poke_interval=datetime.timedelta(minutes=5),
+        {% endif -%}
+        {% if table_sensor_task.timeout != None -%}
+        timeout={{ table_sensor_task.timeout | format_timedelta | format_repr }},
+        {% else -%}
+        timeout=datetime.timedelta(hours=8),
```
sean-rose, line 336: I chose 8 hours as the default timeout so that when combined with the common DAG-level defaults of 2 retries it would result in the sensor waiting for ~24 hours total before giving up and failing entirely.

bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -304,6 +316,69 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
...
 
+    {% for table_sensor_task in task.depends_on_tables_existing -%}
+    {% set project_id, dataset_name, table_name = table_sensor_task.table_id.split('.') -%}
+    {{ table_sensor_task.task_id }} = BigQueryTableExistenceSensor(
+        task_id={{ table_sensor_task.task_id | format_repr }},
+        project_id={{ project_id | format_repr }},
+        dataset_id={{ dataset_name | format_repr }},
+        table_id={{ table_name | format_repr }},
+        gcp_conn_id='google_cloud_shared_prod',
+        deferrable=True,
```
sean-rose, line 327: I opted to use the newer [deferrable operator](https://airflow.apache.org/docs/apache-airflow/2.7.3/authoring-and-scheduling/deferring.html) approach (rather than `mode="reschedule"`) since the BigQuery table sensors support it.  This might be a bit of a risk, since I think this could be the very first use of deferrable operators in our Airflow setup, but it seemed worth trying.

tests/data/dags/test_dag_with_bigquery_table_sensors
```diff
@@ -0,0 +1,93 @@
...
+        date_partition_parameter="submission_date",
+        depends_on_past=True,
+        arguments=["--append_table"],
+    )
+
+    wait_for_foo_bar_baz = BigQueryTableExistenceSensor(
+        task_id="wait_for_foo_bar_baz",
+        project_id="foo",
+        dataset_id="bar",
+        table_id="baz_{{ ds_nodash }}",
```
ANich, line 68: No strong opinions here but if this sensor can be used for sharded tables, can it be used for partitioned tables? So we only need `BigQueryTableExistenceSensor` - something like `baz${{ ds_nodash }}`

bigquery_etl/query_scheduling/task.py
```diff
@@ -125,6 +122,66 @@ def get_execution_delta(self, schedule_interval):
         return None
 
 
+@attr.s(auto_attribs=True, frozen=True)
+class TableSensorTask:
+    """Representation of a sensor task to wait for a table to exist."""
+
+    task_id: str = attr.ib()
```
scholtzan, line 129: For other task types we are generating the `task_id` based on dataset identifiers (like `wait_for`). Should we do the same here and keep this option to overwrite the default?

scholtzan (approved):
Just one question I left inline

sql/moz-fx-data-marketing-prod/google_search_console_derived/search_impressions_by_page_v2/metadata.yaml
```diff
@@ -21,9 +21,22 @@ scheduling:
   date_partition_parameter: date
   # Google Search Console exports typically happen two days after the data date in UTC.
```
ANich, line 22: Would be nice to have this added directly as a comment in the generated DAG

ANich (approved):
LGTM, left some thoughts

sql/moz-fx-data-marketing-prod/google_search_console_derived/search_impressions_by_page_v2/metadata.yaml
sean-rose: An equivalent comment is [in the DAG description](https://github.com/mozilla/bigquery-etl/blob/cfd16aaef47dacdbc802683f2b81d9a9d66ec1d1/dags.yaml#L1480), which ends up in the DAG code as a docstring and is visible in the Airflow UI.

tests/data/dags/test_dag_with_bigquery_table_sensors
sean-rose: A clever idea, but I just tried it, and the way the underlying sensor code is written that doesn't work as intended.

sql/moz-fx-data-marketing-prod/google_search_console_derived/search_impressions_by_site_v2/metadata.yaml
```diff
@@ -21,9 +21,22 @@ scheduling:
   date_partition_parameter: date
   # Google Search Console exports typically happen two days after the data date in UTC.
   date_partition_offset: -1
-  depends_on:
-  - task_id: google_search_console_derived__site_impression_exports_empty_check__v1
-    dag_name: bqetl_google_search_console
+  depends_on_table_partitions_existing:
```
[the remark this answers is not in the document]
sean-rose: It could be, but personally I don't see a big benefit to doing so, and it seems like it'd require some advanced `cattrs` code for polymorphism. (or maybe you meant to have the sub-items as keys rather than a list? but even then I'm not sure it'd be worth it)

bigquery_etl/query_scheduling/task.py
sean-rose: That was my initial thought as well, but ran into some complicating factors:
- There will almost certainly be Jinja in the underlying identifiers.  Some ways I considered handling that:
  - Keep a sanitized version of the Jinja in the task ID (but then we'd have awkward task ID suffixes like `_macros_ds_add_ds_1_ds_nodash`, and the task ID could get altered whenever the Jinja is tweaked).
  - Remove the Jinja and append a hash of the overall underlying value to guarantee uniqueness (but then when viewing the tasks in Airflow you couldn't easily reason about which one was which if there were multiple for the same table, and again the task ID could get altered whenever the Jinja is tweaked).
  - Remove the Jinja and append a numerical suffix like `_1`, `_2`, etc. in case there are multiple for the same table (but then the meaning of those numbered tasks could get mixed up depending on how we're ordering them and when tasks get added/removed).
- If we're auto-generating the task IDs we take on the responsibility of handling task ID conflicts (e.g. if multiple ETLs depend on the same tables/partitions). Some ways I considered handling that:
  - Only output one sensor task in cases where multiple ETLs depend on the same tables/partitions (but then that could be finicky, like what if the Jinja isn't _exactly_ the same but is effectively the same).
    - IMO this would be the main reason to go to the effort of implementing auto-generated task IDs for these, but I don't think it'd be a common edge case and it can be manually worked around (e.g. one ETL uses `depends_on_tables_existing` and another ETL uses `depends_on` to also wait for the first ETL's `depends_on_tables_existing` task, or they can just pick unique task IDs as the sensors are so low-impact having duplicates shouldn't really matter).
  - Include the ETL table ID in the generated task ID (but then we'd get super long task IDs that are harder to read and could easily hit Airflow's 250 character limit for task IDs).

So ultimately I decided it was going to be too complex/fragile to try to implement auto-generated task IDs for this.

However, if you have ideas for implementing this functionality in a more straightforward way than I've made it out to be then I'd be happy to hear them!

bigquery_etl/query_scheduling/templates/airflow_dag.j2
sean-rose: It turns out I was wrong about sensors retrying when they hit their timeout, as that behavior was [changed way back in Airflow 2.2.0](https://airflow.apache.org/docs/apache-airflow/stable/release_notes.html#if-a-sensor-times-out-it-will-not-retry) so that sensors don't retry when they hit their timeout.  I'm addressing this in #7117.

PR #5055, opened 2024-02-15
Add queueing to avoid dirty writes on main

ANich:
Queue sync jobs to avoid previous commits on main clobbering syncs from newer ones.

Uses `eddiewebb/queue` orb: [docs](https://circleci.com/developer/orbs/orb/eddiewebb/queue#usage-single_concurrency_job)

Open question: if the previous job in the queue fails, does the current get sent to the front of the line?  

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2782)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,7 @@ orbs:
   gcp-gcr: circleci/gcp-gcr@0.13.0
   docker: circleci/docker@2.2
   python: circleci/python@2.1.1
+  queue: eddiewebb/queue@2.2.1
 
 parameters:
   python-version:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -292,6 +293,8 @@ jobs:
   docs:
     docker: *docker
     steps:
+      - queue/until_front_of_line:
+          only-on-branch: main
```
ANich, line 297: By default this waits 10m before giving up and I assume failing the job, should we consider extending this or having it wait indefinitely? I think we can monitor here and adjust as needed

ANich:
Converted back to draft, this isn't a [certified orb](https://circleci.com/docs/orb-intro/#orb-designation) so I'll open an RRA.

ANich:
Won't do see https://bugzilla.mozilla.org/show_bug.cgi?id=1881337

PR #5057, opened 2024-02-15
Bug 1868244 Publish views if dry run schema differs from deployed

BenWu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1868244

This adds a dry run step to the function that checks if a view has changes to be published.  The dry run result includes the schema of the resulting view that can be compared to the active schema.

Did a quick performance test locally with this:
```python
for view in _collect_views(None, "sql", "moz-fx-data-shared-prod", False, False):
    view.has_changes()
```
The dry run adds on average ~0.4s to every view that reaches that code path.  The total time to check all the views after running `./script/bqetl generate stable_views` was 526 seconds without dry run and 688 seconds with dry run.  This seems acceptable to me but I don't have all the context around how this is run other than in the airflow task.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2785)

### DIFF
--- a/bigquery_etl/view/__init__.py
+++ b/bigquery_etl/view/__init__.py
@@ -9,7 +9,7 @@ from pathlib import Path
 
 import attr
 import sqlparse
-from google.api_core.exceptions import BadRequest, NotFound
+from google.api_core.exceptions import BadRequest, NotFound, Forbidden
 from google.cloud import bigquery
 
 from bigquery_etl.config import ConfigLoader
@@ -266,17 +266,6 @@ class View:
             print(f"view {target_view_id} will change: query does not match")
             return True
 
-        # check schema
-        schema_file = Path(self.path).parent / "schema.yaml"
-        if schema_file.is_file():
-            view_schema = Schema.from_schema_file(schema_file)
-            table_schema = Schema.from_json(
-                {"fields": [f.to_api_repr() for f in table.schema]}
-            )
-            if not view_schema.equal(table_schema):
-                print(f"view {target_view_id} will change: schema does not match")
-                return True
-
         # check metadata
         if self.metadata is not None:
             if self.metadata.description != table.description:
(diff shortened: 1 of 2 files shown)


bigquery_etl/view/__init__.py
```diff
@@ -291,6 +280,30 @@ def has_changes(self, target_project=None):
...
+                    job_config=bigquery.QueryJobConfig(
+                        dry_run=True, use_legacy_sql=False
+                    ),
+                )
+                view_schema = Schema.from_bigquery_schema(query_job.schema)
+            except Forbidden:
+                print(
+                    f"Missing permission to dry run view {target_view_id} to get schema"
+                )
+                view_schema = None
```
scholtzan, line 301: I guess the downside here is that we no longer update views that have any access restrictions. Maybe we should always update them instead?

scholtzan (approved):

bigquery_etl/view/__init__.py
```diff
@@ -291,6 +280,30 @@ def has_changes(self, target_project=None):
                 print(f"view {target_view_id} will change: labels do not match")
                 return True
 
+        schema_file = Path(self.path).parent / "schema.yaml"
+        table_schema = Schema.from_bigquery_schema(table.schema)
+        # check schema based on schema file
+        if schema_file.is_file():
+            view_schema = Schema.from_schema_file(schema_file)
```
ANich, line 287: nit: could be a property
```
@property/cached_property
def local_schema(self) -> Optional[Schema]:
    (some of this section)
    return None
```
Then the next section becomes `if self.local_schema is None`... etc.

bigquery_etl/view/__init__.py
```diff
@@ -291,6 +280,30 @@ def has_changes(self, target_project=None):
...
+                    ),
+                )
+                view_schema = Schema.from_bigquery_schema(query_job.schema)
+            except Forbidden:
+                print(
+                    f"Missing permission to dry run view {target_view_id} to get schema"
+                )
+                view_schema = None
+
+        if view_schema is not None and not view_schema.equal(table_schema):
```
ANich, line 303: How do descriptions factor in here? If the view is deployed with descriptions, will comparing schemas w/ the dryrun schema return False?

ANich (approved):
LGTM left a nit + question

bigquery_etl/view/__init__.py
BenWu: Do you mean view-level or column descriptions?  For view descriptions, they're compared in the `if self.metadata is not None:` section above so schemas are only compared if view descriptions match.  
For column descriptions, my understanding is the `.equal` method here ignores them so this will only return true if there are different fields.  I actually didn't consider that when writing this since my intention was to keep the behaviour as similar to before as possible.

bigquery_etl/view/__init__.py
BenWu: Good point.  I don't think this will run multiple times per object very often but probably doesn't hurt to add.

bigquery_etl/view/__init__.py
ANich: I meant schema descriptions here. Might be worth a test to double-check, otherwise we might get schemas that always "have changes" because the dryrun schema won't contain those descriptions

BenWu: I can confirm descriptions in the deployed schema don't cause the view to be published.  
For example, running `print(f"{view.name}: {view.has_changes()}")` with the `airflow_task_fail` view outputs:
```
Warning: descriptions for root.dag_id differ
Warning: descriptions for root.task_id differ
airflow_task_fail: False
```

PR #5075, opened 2024-02-20
CI path filtering; with run-all workflow on main

scholtzan:
This reverts mozilla/bigquery-etl#5053 and brings back path filtering. This time with a `run-all` workflow on `main` that runs the deploys

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2806)


.circleci/workflows.yml
```diff
@@ -0,0 +1,1094 @@
...
+  ### Validate bqetl tooling
+  validate-bqetl:
+    when:
+      or:
+        - << pipeline.parameters.validate-bqetl >>
+    jobs:
+      - test-bqetl
+      - integration
+
+  run-all:
```
BenWu, line 962: What's the reason for going with `run-all` rather than using "gcr deploy" parameter and a workflow that runs only those jobs?  In that case builds on main can just run with all the params set to true.  My initial impression is the `run-all` workflow might be confusing or error-prone because of the jobs duplicated in the workflows

scholtzan: The deployment tasks depend on the `generate-sql` and `private-generate-sql` tasks. I guess we could add the deployment tasks to the `validate-sql` workflow, but then it might be confusing as well that it runs validation and does deploys when run on main.
The other option would be to run the `generate-sql` tasks twice, when validating SQL and before doing the deploys, but that just feels like unnecessarily increasing CircleCI cost and run time.

There doesn't seem to be a way to wait for another workflow in CircleCI, only for other jobs

BenWu: Ok I see the issue there.  Some jobs are already set with filters to run only on main so I don't think it's too much of a stretch to put the deploys in the other workflow and only run on main.  Just an opinion though and I'm not completely against this setup.  My current concern is that we always need to remember to add/remove jobs in both workflows and it wouldn't immediately be clear if we make a mistake and the workflows diverge.

.circleci/workflows.yml
```diff
@@ -0,0 +1,1094 @@
...
+      - sync-dags-repo:
+          name: 🔃 Synchronize bigquery-etl submodule
+          repo-to-sync: ${CIRCLE_PROJECT_REPONAME}
+          target-branch: generated-sql
+          requires:
+            - push-generated-sql
+          filters:
+            branches:
+              only:
+                - main
```
BenWu, line 941: I think this and the `ignore: main` above can be removed since this workflow doesn't run on main, right?

BenWu (commented):
Just one more comment.  For my other comment, I'm open to other opinions so I don't think it should be a blocker.

.circleci/workflows.yml
scholtzan: Yeah, I guess either way it's not ideal. It's possible that someone might add a task in a different workflow that deployment tasks depend on, which would not work then.
Maybe being more explicit and having different workflows is okay for now,

BenWu: Got it, I think this is good enough then.  Thanks

BenWu (approved):

akkomar (commented):
Could we instead have a single workflow and execute job steps conditionally?
For example:
```
  generate-sql:
    docker: *docker
    resource_class: large
    steps:
      - when:
          condition:
            or: [ << pipeline.parameters.validate-sql >>, << pipeline.parameters.validate-routines >> ]
          steps:
            - checkout
            - *restore_venv_cache
            (...)
```

scholtzan:
conditions don't exist for steps, only for workflows

akkomar:
> conditions don't exist for steps, only for workflows

Ah, I took it from [this post](https://discuss.circleci.com/t/conditional-steps-in-jobs-called-in-workflows/37965/2), but can't find it mentioned anywhere else in the docs.

scholtzan:
The other alternative is to add conditions on a task level via bash scripting:
```
if [ "<< pipeline.parameters.validate-sql >>" = "true" ]; then
```

The annoyance here is that it needs to be added to every step that should or should not be executed. This might add more complexity than having multiple workflows. I guess the advantage is that we can still mark certain tasks as required.

PR #5093, opened 2024-02-22
DENG-2831 create UDF to extract UTM parameters from URL

lelilia:
[ticket](https://mozilla-hub.atlassian.net/browse/DENG-2831)

Adding a UDF to parse UTM parameters from URL

Checklist for reviewer:

### DIFF
--- a/sql/mozfun/utils/extract_utm_from_url/metadata.yaml
+++ b/sql/mozfun/utils/extract_utm_from_url/metadata.yaml
@@ -1,0 +1,5 @@
+description: >
+  Extract UTM parameters from URL.
+  Returns a STRUCT<utm_source STRING, utm_medium STRING, utm_campaign STRING, utm_content STRING, utm_term STRING>
+  UTM (Urchin Tracking Module) parameters are URL parameters used by marketing to track the effectiveness of online marketing campaigns.
+friendly_name: Extract UTM From Url
--- a/sql/mozfun/utils/extract_utm_from_url/udf.sql
+++ b/sql/mozfun/utils/extract_utm_from_url/udf.sql
@@ -1,0 +1,59 @@
+
+                -- Definition for utils.extract_utm_from_url
+                -- For more information on writing UDFs see:
+                -- https://docs.telemetry.mozilla.org/cookbooks/bigquery/querying.html
+CREATE OR REPLACE FUNCTION utils.extract_utm_from_url(url STRING)
+RETURNS STRUCT<
+  utm_source STRING,
+  utm_medium STRING,
+  utm_campaign STRING,
+  utm_content STRING,
+  utm_term STRING
+> AS (
+  STRUCT(
+    REGEXP_EXTRACT(url, r"utm_source=([^&\'\"\]]+)"),   -- utm_source
+    REGEXP_EXTRACT(url, r"utm_medium=([^&\'\"\]]+)"),   -- utm_medium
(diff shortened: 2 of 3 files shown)


kwindau (approved):
LGTM

sql/mozfun/utils/extract_utm_from_url/udf.sql
```diff
@@ -0,0 +1,59 @@
...
+  utm_campaign STRING,
+  utm_content STRING,
+  utm_term STRING
+> AS (
+  STRUCT(
+    REGEXP_EXTRACT(url, r"utm_source=([^&\'\"\]]+)"),   -- utm_source
+    REGEXP_EXTRACT(url, r"utm_medium=([^&\'\"\]]+)"),   -- utm_medium
+    REGEXP_EXTRACT(url, r"utm_campaign=([^&\'\"\]]+)"), -- utm_campaign
+    REGEXP_EXTRACT(url, r"utm_content=([^&\'\"\]]+)"),  -- utm_content
+    REGEXP_EXTRACT(url, r"utm_term=([^&\'\"\]]+)")      -- utm_term
```
sean-rose, line 18: - Probably also want to stop at `#` characters that start the fragment part of URLs.
- I'm curious why this stops at quote characters or right square bracket characters, as it seems like those could techincally be in UTM param values, even though that'd be unusual.

sean-rose: Also, if it'd be useful to differentiate between UTM parameters being in the URL with no value (e.g. `...&utm_campaign=&...`) versus not in the URL at all then you could match the values with `*` rather than `+`, which would give you an empty string in the former case and a null value in the latter case.

sql/mozfun/utils/extract_utm_from_url/udf.sql
```diff
@@ -0,0 +1,59 @@
+
+                -- Definition for utils.extract_utm_from_url
+                -- For more information on writing UDFs see:
+                -- https://docs.telemetry.mozilla.org/cookbooks/bigquery/querying.html
```
sean-rose, line 4: Boilerplate comment could be removed.

sql/mozfun/utils/extract_utm_from_url/metadata.yaml
```diff
@@ -0,0 +1,5 @@
+description: >
+  Extract UTM parameters from URL.
+  Returns a STRUCT<utm_source STRING, utm_medium STRING, utm_campaign STRING, utm_content STRING, utm_term STRING>
+  UTM (Urchin Tracking Module) parameters are URL parameters used by marketing to track the effectiveness of online marketing campaigns.
+friendly_name: Extract UTM From Url
```
sean-rose, line 5:
```suggestion
friendly_name: Extract UTM From URL
```

sql/mozfun/utils/extract_utm_from_url/udf.sql
lelilia: Thank you very much! 

* anchoring the beginning ✅ 
* add # to the stop ✅ 
* removed the quotes from the stop ✅ (I thought that it never can be part of a URL but .. turns out .. thank you good catch
* added the AS instead of the comment  ✅  even though I'm not 100% sure on that ... 

```
create temp function leli_test() 
returns struct<utm_source string, utm_con string> as (
  struct( 
    "bu" as utm_source,
    "foo" as utm_configuration)
);

select leli_test() as test 
```

image

So it ignores the names I give it and maps it according to the output order. With naming it that fact might be obscured ... but no too strong feels on it so following your advice

PR #5095, opened 2024-02-22
Conditional steps in CI

scholtzan:
Alternative to https://github.com/mozilla/bigquery-etl/pull/5075

Use conditionals in steps like:

```yaml
  reset-stage-env:
    docker: *docker
    steps:
      - when:
          condition:
            or: 
              - << pipeline.parameters.validate-sql >>
              - << pipeline.parameters.validate-routines >>
          steps:
            - checkout
            - *skip_forked_pr
            - *build
            - run:
                name: "Delete stage datasets"
                command: |
                  export GOOGLE_APPLICATION_CREDENTIALS="/tmp/gcp.json"
                  echo "$GCLOUD_SERVICE_KEY" > "$GOOGLE_APPLICATION_CREDENTIALS"

                  PATH="venv/bin:$PATH" script/bqetl stage clean --dataset-suffix=$CIRCLE_SHA1 --delete-expired
      - run: exit 0
```

This PR is also removing the `nightly` job since it has been failing to run due to permission issues.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2836)


.circleci/workflows.yml
```diff
@@ -0,0 +1,1151 @@
...
+          steps:
+            - checkout
+            - *restore_venv_cache
+            - *build
+            - run:
+                name: Verify that SQL is correctly formatted
+                command: |
+                  PATH="venv/bin:$PATH" script/bqetl format --check \
+                  $(git ls-tree -d HEAD --name-only)
+      - run: exit 0
```
scholtzan, line 84: CircleCI needs to have at least one `run` step, so adding these as dummy steps

kwindau (approved):

BenWu (approved):
It is a bit clunky needing to add the condition to each job but I prefer this method since it avoids the duplication.  It would be nice if we could mark a job as skipped rather than successful if it doesn't match the condition so it's clearer in the checks.  I think it's possible in gh actions but I don't see a way to do that here. 
Thanks for revisiting this

.circleci/workflows.yml
sean-rose: Maybe instead of having an always-run no-op step this could instead be an `unless` conditional, using the same condition as the `when` (e.g. via YAML reference to keep it DRY), with a single step named something meaningful like "skipped" (and again could use YAML references to avoid repeating the same skipped/no-op step code everywhere)?

.circleci/workflows.yml
```diff
@@ -0,0 +1,1151 @@
...
+                # yamllint enable rule:line-length
+      - run: exit 0
+  validate-metadata:
+    docker: *docker
+    steps:
+      - when:
+          condition:
+            or:
+              - << pipeline.parameters.validate-sql >>
+              - << pipeline.parameters.validate-routines >>
```
sean-rose, line 209: I wouldn't think that routines changing would make any difference for validating query metadata, except the `bqetl query validate` command being run is currently also dry-running the queries (even though that's kind of duplicative of the `dry-run-sql` job) which routines could affect.

.circleci/workflows.yml
```diff
@@ -0,0 +1,1151 @@
...
+                no_output_timeout: 30m
+      - run: exit 0
+  generate-dags:
+    docker: *docker
+    steps:
+      - when:
+          condition:
+            or:
+              - << pipeline.parameters.validate-sql >>
+              - << pipeline.parameters.validate-routines >>
```
sean-rose, line 258: - I wouldn't think that routines changing would make a difference for generating DAGs.
- The `dags.yaml` file changing should set the `validate-sql` parameter (or some other DAG-specific parameter could be used).

.circleci/workflows.yml
```diff
@@ -0,0 +1,1151 @@
...
+  validate-dags:
+    executor:
+      name: python/default
+      tag: 3.10.12
+    steps:
+      - when:
+          condition:
+            or:
+              - << pipeline.parameters.validate-sql >>
+              - << pipeline.parameters.validate-routines >>
```
sean-rose, line 292: ditto

.circleci/workflows.yml
```diff
@@ -0,0 +1,1151 @@
...
+                  - staged-generated-sql
+      - run: exit 0
+  push-generated-sql:
+    docker: *docker
+    steps:
+      - when:
+          condition:
+            or:
+              - << pipeline.parameters.validate-sql >>
+              - << pipeline.parameters.validate-routines >>
```
sean-rose, line 573: Since `push-generated-sql` should only happen on deployments, it seems like this should just be checking the `deploy` parameter (assuming you want to stick with that approach).

.circleci/config.yml
```diff
@@ -1,977 +1,67 @@
...
-  nightly:
-    # Run after schema-generator to ensure we are up-to-date
-    triggers:
-      - schedule:
-          cron: "0 5 * * *"
-          filters:
-            branches:
-              only:
-                - main
-    jobs: *build_jobs
```
sean-rose, line 977: This is removing the `nightly` workflow without a replacement.  I'm guessing that's intentional since that workflow hasn't really been working as intended, but it'd be good to note (e.g. in the PR description).

sean-rose (commented):
Thanks for going to the effort of trying this approach, and so quickly!

And I also appreciate you splitting the YAML linting and PyTest steps out from the `build` job into their own separate jobs.

.circleci/config.yml
```diff
@@ -1,977 +1,67 @@
...
 
+setup: true
+
 orbs:
-  gcp-gcr: circleci/gcp-gcr@0.13.0
-  docker: circleci/docker@2.2
-  python: circleci/python@2.1.1
+  path-filtering: circleci/path-filtering@0.1.3
 
 parameters:
```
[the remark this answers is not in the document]
scholtzan: Some of them are used in the path-filter mapping. But I'll remove the ones not being used.

.circleci/workflows.yml
```diff
@@ -0,0 +1,1151 @@
...
+    default: false
+  deploy:
+    type: boolean
+    default: false
+  trigger-sql-generation:
+    type: boolean
+    default: false
+  skip-stage-deploys:
+    type: boolean
+    default: false
```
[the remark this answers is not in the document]
scholtzan: Yes, that is only for manual use

.circleci/config.yml
```diff
@@ -1,977 +1,67 @@
...
+            bigquery_etl/.* validate-bqetl true
+            tests/.* validate-bqetl true
+            bigquery_etl/query_scheduling/.* validate-sql true
+            sql/.* validate-sql true
+            sql_generators/.* validate-sql true
+            tests/sql/.* validate-sql true
+            biquery_etl/routine/.* validate-routines true
+            sql/mozfun/.* validate-routines true
+            sql/moz-fx-data-shared-prod/udf/.* validate-routines true
+            sql/moz-fx-data-shared-prod/udf_js/.* validate-routines true
```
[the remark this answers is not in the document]
scholtzan: I added `.circleci` to that list as well

.circleci/config.yml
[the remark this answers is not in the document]
sean-rose: From what I could tell looking at the path-filtering orb code it doesn't seem to do anything with parameters in the config it's used in, it's only defining parameters to pass to a downstream config.  So the parameters are definitely needed in `workflows.yml`, but I don't think they're needed here.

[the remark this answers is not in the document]
scholtzan: Ah right, they are not needed for path-filtering. However they need to be listed to allow manual triggering of specific workflows. Otherwise there is an unexpected argument error: https://app.circleci.com/pipelines/gh/mozilla/bigquery-etl/30607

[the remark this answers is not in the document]
sean-rose: OK.  Does there need to be extra logic to pass through the parameter values to `workflows.yml` if it's triggered manually?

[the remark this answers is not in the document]
scholtzan: No extra logic needed. It just passes them through.

.circleci/workflows.yml
```diff
@@ -0,0 +1,1198 @@
...
+          condition: *validate-sql
+          steps:
+            - *skip
+  test-routines:
+    docker: *docker
+    steps:
+      - when:
+          condition: &validate-routines
+            or:
+              - << pipeline.parameters.validate-routines >>
```
sean-rose, line 378: Should this also check the `deploy` parameter like the other validate conditions?
```suggestion
          condition: &validate-routines
            or:
              - << pipeline.parameters.validate-routines >>
              - << pipeline.parameters.deploy >>
```

scholtzan: yes https://github.com/mozilla/bigquery-etl/pull/5115

PR #5103, opened 2024-02-23
Remove submission_date from events_stream tables

scholtzan:
Wil pointed out that the `events_stream` tables have both `submission_date` and `submission_timestamp` fields. Currently, the tables are partitioned on `submission_date` and are requiring a partition date filter. This is causing some issues when using the explores in Looker, but is also potentially quite inconvenient for users that want to filter down to a specific timestamp range and now need to filter on both `submission_date` and `submission_timestamp`. Looker is representing timestamps and dates as dimension groups. In this case with both `submission_date` and `submission_timestamp` being present there is a naming "collision" with `submission_timestamp` taking precedence resulting in the date filter not being applied to `submission_date`.

I think it would be more convenient to remove `submission_date` from the tables and partition on `submission_timestamp` instead. We'd need to redeploy the existing tables with the new partitioning.
I added `submission_date` to the user-facing views for backwards compatibility.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2849)

### DIFF
--- a/sql_generators/glean_usage/templates/events_stream_v1.metadata.yaml
+++ b/sql_generators/glean_usage/templates/events_stream_v1.metadata.yaml
@@ -17,7 +17,7 @@ scheduling:
 bigquery:
   time_partitioning:
     type: day
-    field: submission_date
+    field: submission_timestamp
     require_partition_filter: true
   clustering:
     fields:
--- a/sql_generators/glean_usage/templates/events_stream_v1.query.sql
+++ b/sql_generators/glean_usage/templates/events_stream_v1.query.sql
@@ -26,7 +26,6 @@ WITH base AS (
         ping_info.ping_type
       ) AS ping_info
     ),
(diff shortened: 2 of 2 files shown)


sql_generators/glean_usage/templates/events_stream.view.sql
```diff
@@ -4,6 +4,7 @@ CREATE OR REPLACE VIEW
   `{{ project_id }}.{{ events_stream_view }}`
 AS
 SELECT
+  DATE(submission_timestamp) AS submission_date,
```
quiiver, line 7: I don't think we have any downstream consumers yet so probably better to just take this out now

quiiver (approved):
Thank you for doing this!

badboy (approved):
Yeah, I noticed this too. I didn't go to address that because of the implications of removing fields. But I guess better now before we have more of that. Thanks!

PR #5124, opened 2024-02-27
DENG-2095 Fix accounts_backend.users_services_daily_v1

akkomar:
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2895)

### DIFF
--- a/sql/moz-fx-data-shared-prod/accounts_backend_derived/users_services_daily_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/accounts_backend_derived/users_services_daily_v1/query.sql
@@ -56,13 +56,10 @@ windowed AS (
     udf.mode_last(ARRAY_AGG(country) OVER w1) AS country,
     udf_contains_tier1_country(ARRAY_AGG(country) OVER w1) AS seen_in_tier1_country,
     LOGICAL_OR(event_name = 'reg_complete') OVER w1 AS registered,
-    ARRAY_AGG(event_name) OVER w1 AS service_events,
   FROM
     fxa_events
   WHERE
     DATE(submission_timestamp) = @submission_date
-    AND user_id_sha256 != ''
-    AND service != ''
   QUALIFY
     ROW_NUMBER() OVER (
       PARTITION BY
@@ -88,16 +85,9 @@ windowed AS (
 SELECT
   DATE(@submission_date) AS submission_date,
   windowed.user_id_sha256,
-  oa.name AS service,
+  windowed.service,
   windowed.country,
   windowed.seen_in_tier1_country,
   windowed.registered,
-  windowed.service_events,
 FROM
   windowed
-JOIN
-  `accounts_db.fxa_oauth_clients` AS oa
-  ON windowed.service = oa.id
-WHERE
-  user_id_sha256 IS NOT NULL
-  AND service IS NOT NULL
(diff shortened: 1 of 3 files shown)


sql/moz-fx-data-shared-prod/accounts_backend_derived/users_services_daily_v1/query.sql
```diff
@@ -56,13 +56,10 @@ windowed AS (
     udf.mode_last(ARRAY_AGG(country) OVER w1) AS country,
     udf_contains_tier1_country(ARRAY_AGG(country) OVER w1) AS seen_in_tier1_country,
     LOGICAL_OR(event_name = 'reg_complete') OVER w1 AS registered,
-    ARRAY_AGG(event_name) OVER w1 AS service_events,
   FROM
     fxa_events
   WHERE
     DATE(submission_timestamp) = @submission_date
-    AND user_id_sha256 != ''
-    AND service != ''
```
akkomar, line 65: Removing this results in some new rows with empty `service` field. I'm not sure if we're not undercounting with this condition.

sql/moz-fx-data-shared-prod/accounts_backend_derived/users_services_daily_v1/query.sql
```diff
@@ -88,16 +85,9 @@ windowed AS (
...
   windowed.registered,
-  windowed.service_events,
 FROM
   windowed
-JOIN
-  `accounts_db.fxa_oauth_clients` AS oa
-  ON windowed.service = oa.id
-WHERE
-  user_id_sha256 IS NOT NULL
-  AND service IS NOT NULL
```
akkomar, line 103: There are service names that map to multiple `id`s in `accounts_db.fxa_oauth_clients`. Because we use `service` field for joining in `users_services_last_seen`, this was resulting in a cross product there.
I see three options to fix this problem:
1. Remove the join and service name from here, add in the view
2. Add a new `service_name` column here
3. Keep the join here, but move it to an earlier stage and window over service name. This will yield less rows than current implementation and above options. The question is if here we need to distinguish between services that map to the same name in `accounts_db.fxa_oauth_clients`.

sql/moz-fx-data-shared-prod/accounts_backend_derived/users_services_last_seen_v1/query.sql
```diff
@@ -27,7 +15,6 @@ WITH _current AS (
     accounts_backend_derived.users_services_daily_v1
   WHERE
     submission_date = @submission_date
-    AND contains_qualified_fxa_activity_event(service_events)
```
akkomar, line 30: This was copied from the legacy table but we don't need this check here.

sql/moz-fx-data-shared-prod/accounts_backend_derived/users_services_daily_v1/query.sql
ksiegler1: I'm leaning towards option 1. Do you have a preference?

sql/moz-fx-data-shared-prod/accounts_backend_derived/users_services_daily_v1/query.sql
ksiegler1: We can remove it. I would like to understand how many additional rows this condition had removed, but I can do the comparison afterwards.

ksiegler1 (approved):

PR #5144, opened 2024-02-29
Change Fivetran Airflow operators to run synchronously so the task concurrency limit is respected

sean-rose:
#5083 tried to limit Fivetran task concurrency so it would only run one sync per connector at a time, but that didn't actually work because task concurrency limits apparently don't apply when tasks get deferred (maybe an Airflow bug?).  This will address that.

While having the operator be deferred while waiting for the Fivetran sync to complete would be ideal from an Airflow resource usage perspective, I believe having the sync behavior work as intended is more important, and having the Fivetran operator take up a worker slot while it's waiting shouldn't be a dealbreaker since that was the behavior until very recently when we switched to Astronomer's Fivetran Airflow provider package with its deferrable operator implementation (https://github.com/mozilla/telemetry-airflow/pull/1922, https://github.com/mozilla/bigquery-etl/pull/5080).

Related PR: https://github.com/mozilla/telemetry-airflow/pull/1932

---
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2925)

### DIFF
--- a/bigquery_etl/query_scheduling/templates/airflow_dag.j2
+++ b/bigquery_etl/query_scheduling/templates/airflow_dag.j2
@@ -159,6 +159,8 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
     {{ fivetran_task.task_id }}_sync_start = FivetranOperator(
         connector_id='{% raw %}{{{% endraw %} var.value.{{ fivetran_task.task_id }}_connector_id {% raw %}}}{% endraw %}',
         task_id='{{ fivetran_task.task_id }}_task',
+        {# Have the operator run synchronously so the task concurrency limit is respected. -#}
+        deferrable=False,
         task_concurrency=1,
     )
 
--- a/tests/data/dags/simple_test_dag
+++ b/tests/data/dags/simple_test_dag
@@ -51,12 +51,14 @@ with DAG(
     fivetran_import_1_sync_start = FivetranOperator(
         connector_id="{{ var.value.fivetran_import_1_connector_id }}",
         task_id="fivetran_import_1_task",
+        deferrable=False,
         task_concurrency=1,
     )
 
     fivetran_import_2_sync_start = FivetranOperator(
         connector_id="{{ var.value.fivetran_import_2_connector_id }}",
         task_id="fivetran_import_2_task",
+        deferrable=False,
         task_concurrency=1,
     )
 


BenWu (commented):
Out of curiosity, I took a look at how the `deferrable` arg is implemented and it look like it just runs the sleep function instead of doing a typical sensor operator poke since it's not a sensor operator. https://github.com/astronomer/airflow-provider-fivetran-async/blob/main/fivetran_provider_async/operators.py#L133
I think this means it will actually take up a slot during the entire wait.  Do you know if this is actually the case?  And how long do these typically wait for?

BenWu:
This might be a non-issue, especially if the syncs don't take that long.  It's just that I was under the impression that it would behave like a regular sensor and taking up slots waiting might stall other tasks.  Another option is to use the FivetranSensor which I believe we were previously using.

sean-rose:
> Out of curiosity, I took a look at how the `deferrable` arg is implemented and it look like it just runs the sleep function instead of doing a typical sensor operator poke since it's not a sensor operator. https://github.com/astronomer/airflow-provider-fivetran-async/blob/main/fivetran_provider_async/operators.py#L133 I think this means it will actually take up a slot during the entire wait. Do you know if this is actually the case? 

Yep, in non-deferrable mode it'll take up a worker slot the whole time it's running, but so do sensors in the default `poke` mode (as opposed to `reschedule`), which is how the Fivetran sensors were running prior to mozilla/telemetry-airflow#1922 + #5080.  So it's definitely not ideal, but is basically a return to the previous status quo regarding Fivetran tasks taking up worker slots (ideal would be having some way to control the concurrency of deferred task instances for the same task so we could run these in deferrable mode).

> And how long do these typically wait for?

It probably varies widely depending on the data source.  The Fivetran connector I'm most familiar with is Stripe, which typically takes less than 30 minutes, but could easily take much longer if a historical resync is happening or the Stripe API is being slow.

sean-rose:
> Another option is to use the FivetranSensor which I believe we were previously using.

I wouldn't want to return to using a separate sensor, because then we wouldn't be able to control the concurrency of when it tries to start Fivetran syncs since the start task completes immediately.

BenWu (approved):
Oh right, reschedule is the word I was thinking of.  I thought we were doing that with fivetran before but if it worked before then I have no problem returning to that.  I don't think this is a big issue, just want to make sure we understand the behaviour.

And not using the sensor makes sense to me as well

PR #5165, opened 2024-03-05
[DENG-2854] Search glean per-app derived datasets in shredder

BenWu:
https://mozilla-hub.atlassian.net/browse/DENG-2854

This adds tables in the per glean app derived datasets to shredder (e.g. `fenix_derived`, `firefox_ios_derived`).  There are new 45 tables that will be added with this change.  Tables are listed in a comment in the ticket.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2958)


BenWu:
Code is ready for review but I want to verify the list of added tables

bigquery_etl/shredder/config.py
```diff
@@ -486,35 +466,63 @@ def find_glean_targets(
...
             sources[table.dataset_id] += (source,)
             sources[derived_dataset] += (source,)
-    glean_derived_tables = list(
-        pool.map(
+
+            if app_name is not None and app_name != channel_name:
+                app_names.add(app_name)
+                sources[app_name + "_derived"] += (source,)
+
+    # use deletion request view containing all channels if found
```
akkomar, line 493: Why do these views exist only for some apps? How were they created?

akkomar (commented):
LGTM, thank you!
Can you dry-run this before and after the change and diff the output to confirm that we're only adding tables to shred? (I'm not sure how you got the list in the Jira ticket)

bigquery_etl/shredder/config.py
BenWu: This for apps that are ignored by glean_usage sql generation listed in: https://github.com/mozilla/bigquery-etl/blob/0ec741c910ebffed6e3f96807bf2d55c7b70deca/bqetl_project.yaml#L428-L434

BenWu:
I got the list of tables by running this script with and without the changes:
```py
with ThreadPool(10) as pool:
    glean_targets = config.find_glean_targets(pool, client)

targets_filtered = {  # to remove the manually added glean derived tables
    target: source
    for target, source in glean_targets.items()
    if target.table not in SEARCH_IGNORE_TABLES
}

for target_table, source_table in targets_filtered.items():
    print(target_table.table, target_table.field, source_table)
```
and then copying the two outputs to another script to load them into sets to get the diff.  Looking at the code in delete.py, the list of sources is used as-is so this should match the tables in the real run.

bigquery_etl/shredder/config.py
```diff
@@ -462,6 +441,12 @@ def fields(self) -> tuple[str, ...]:
     ("telemetry_stable.prio_v4", ID),
 }
 
+# names of tables in glean *_derived datasets to ignore
+GLEAN_DERIVED_IGNORE_TABLES = {
+    # to be added in https://mozilla-hub.atlassian.net/browse/DENG-2089
+    "clients_last_seen_joined_v1",
```
lucia-vargas-a, line 447: @BenWu The `clients_last_seen_joined` tables are also meant to be shredded, that's the intention of DENG-2089. I updated the doc, excuses for the confusion, I thought it had been updated already.

BenWu: Oh I thought that ticket was meant to be completed at a later time, after this was completed.  That means there's no need for an ignore list right now.
Unified metrics is only `moz-fx-data-shared-prod.telemetry_derived.unified_metrics_v1` right?

akkomar (approved):
r+wc

https://github.com/mozilla/bigquery-etl/pull/5165#discussion_r1540016902 should be resolved before landing this.

bigquery_etl/shredder/config.py
lucia-vargas-a: Yes, correct.

PR #5177, opened 2024-03-06
DENG-2979 - Update the source for GA client ID from visit_id to client_id_ga4

kwindau:
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2981)


scholtzan (approved):

fbertsch (commented):
If we merge this as-is right now, that will affect the GCLID conversions pipeline. We need to either version this table (and use the new client_id values), or add a new field for ga4 client id

kwindau:
@fbertsch  My initial thought was that, the new GA4stub attribution process went live on Tuesday 3/4/24 around 4pm Central time, so if we switch the table to use the new GA4 client ID going forward, then the code would just use GA4 going forward starting with the next Airflow run. (Most of the times the GA4 client ID matches the GA3 client ID but there are some instances where they are different).  

If you want, I can definitely switch this PR to be a separate versioned if you prefer that route, but then we would need to also create a view combining the old table and the new table and we'd also have to update everything using this table to use the new view instead, so I just initially thought switching the column might be simpler. But open to either option, we can talk more tomorrow / figure out next steps. 👍

PR #5200, opened 2024-03-12
Update query.sql for users_services_daily_v1

ksiegler1:
Semi complicated logic to handle various situations seen in the oauth_client_id and service fields

1. Occurrences of oauth_client_id populating the service field while oauth_client_id field is empty string. This is primarily happening in non-activity events.

2. Occurrences of relying_party_service field being = 'sync' and manually switching this this to be the oauth_client_id

3. Transforming empty strings to nulls for proper handling in `ARRAY_AGG` function

@akkomar I created this PR with changes for how I managed this, but if you have a different way of handling LMK.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3017)


ksiegler1:
Based on advice from @akkomar to use CTEs to simplify the logic for easier understanding, I switched the order of operations and used CTEs to simplify.

1. First, transform empty strings into nulls separately for both `relyng_party_service` and `relying_party_oauth_client_id` fields, then
2. In the `transform_service_field` CTE, handle the `'sync'` string case by checking if the coalescing of `relying_party_oauth_client_id` and `relying_party_service` = `'sync'`
3. If that coalesce doesn't equal `'sync'`, just return the coalesce of the two columns (prioritizing what's in `oauth_client_id` field

akkomar:

PR #5203, opened 2024-03-12
DRAFT implementation of active user at client level using Table vs View.

lucia-vargas-a:
Option 1
/active_users/view
/udf/<active_user_bits>

Option 2
/active_users/query
/active_users/query_view


lucia-vargas-a:
Evaluation of options completed.

PR #5208, opened 2024-03-13
Create derived dataset files for accounts_backend.users_first_seen_v1

ksiegler1:
@akkomar I'm not sure how to initialize the dataset with historical legacy telemetry from fxa_users_first_seen, but I set up the query.sql file for the version using the Glean data (built similarly off the [previous version](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_users_first_seen_v2/query.sql))

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3041)


kik-kik (commented):
We should also add a `schema.yaml` file to `users_first_seen_v1`. This file contains information about which columns exist in this table and is used to initialize this table as an empty table. This is also needed to make you query pass the CI checks, I suspect right now the query dry-run CI check will be failing due to `users_first_seen_v1` not existing.

sql/moz-fx-data-shared-prod/accounts_backend_derived/users_first_seen_v1/query.sql
```diff
@@ -0,0 +1,41 @@
+WITH users_services_daily_new_entries AS (
+  SELECT DISTINCT
+    user_id_sha256,
+    country,
+    seen_in_tier1_country,
+    registered
+  FROM
+    accounts_backend.users_services_daily
```
kik-kik, line 8: Wouldn't it be more efficient to use users_services_first_seen instead for this table and reprocessing all the clients inside `users_services_daily`?

sql/moz-fx-data-shared-prod/accounts_backend_derived/users_first_seen_v1/query.sql
```diff
@@ -0,0 +1,41 @@
...
+  WHERE
+{% if is_init() %}
+    FALSE
+  {% else %}
+    DATE(submission_date) < @submission_date
+  {% endif %}
+)
+SELECT
+  new_entries.*,
+  DATE(@submission_date) AS first_seen_date,
```
kik-kik, line 30: This seems to be inconsistent in relation to `users_services_first_seen_v1` which does not seem to have `first_seen_date`, but instead has `submission_date`:
https://github.com/mozilla/bigquery-etl/pull/5210/files#diff-ff00b83b200545f893fc543ef891ac6cf2e9613f36067c0badf461bc87fd60c6R30

I think we should follow the same naming as the only real difference between the two is the grain of the table.

sql/moz-fx-data-shared-prod/accounts_backend_derived/users_first_seen_v1/query.sql
```diff
@@ -0,0 +1,41 @@
...
+FROM
+  users_services_daily_new_entries AS new_entries
+FULL OUTER JOIN
+  existing_entries
+  USING (user_id_sha256)
+WHERE
+  existing_entries.user_id_sha256 IS NULL
+{% if is_init() %}
```
kik-kik, line 38: Unfortunately, it seems when we try to initialize this table we end up with a query failure due to the query taking too long to complete and eventually timing out. We may have to deploy the schema and then run a daily backfill (process a single date partition at a time).

akkomar:

PR #5210, opened 2024-03-14
Create derived dataset files for accounts_backend.users_services_first_seen_v1

ksiegler1:
Similar to https://github.com/mozilla/bigquery-etl/pull/5208 in needing assistance to initializing with legacy data and backfilling cc @akkomar

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3045)


sql/moz-fx-data-shared-prod/accounts_backend_derived/users_services_first_seen_v1/init.sql
kik-kik, line 1: We made the decision to longer support `init.sql` scripts. Instead we decided to rely on the `metadata.yaml` + `schema.yaml` configuration if we seek to initialize an empty table. In cases where we want to initialize a new table and populate with historical data we have `is_init()` mechanism which can be used inside your `query.sql` file.

Here's an example of this: https://github.com/mozilla/bigquery-etl/blob/8323733162134c27b6207080059a8f7ea49263e8/sql/moz-fx-data-shared-prod/subscription_platform_derived/subplat_flow_events_v1/query.sql#L38-L42

Here's related docs: https://mozilla.github.io/bigquery-etl/bqetl/#initialize

sql/moz-fx-data-shared-prod/accounts_backend_derived/users_services_first_seen_v1/metadata.yaml
```diff
@@ -0,0 +1,19 @@
...
+scheduling:
+  dag_name: bqetl_accounts_derived
+bigquery:
+  time_partitioning:
+    type: day
+    field: submission_date
+    require_partition_filter: true
+    expiration_days: null
+  clustering:
+    fields: [country, service]
```
kik-kik, line 18: Should we order the clustering by service and then country?

kik-kik (commented):
We should add a `schema.yaml` to `users_services_first_seen_v1`. Here's an example: 
https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-marketing-prod/ga_derived/blogs_daily_summary_v2/schema.yaml#L37

I provide a bit more context on why we should do this in my comment regarding the `init.sql` file and the other related PR adding `users_first_seen_v1`.

sql/moz-fx-data-shared-prod/accounts_backend_derived/users_services_first_seen_v1/schema.yaml
```diff
@@ -0,0 +1,16 @@
+fields:
+- name: user_id_sha256
+  type: STRING
+  mode: NULLABLE
+- name: service
+  type: STRING
+  mode: NULLABLE
```
kik-kik, line 7: Should we add `description` key here to provide information about this field. For example, that the value is actually corresponding to the service's id rather than the name. Could also include information that the name mapping is done via the view. Here's an example of adding a description to a field via the schema.yaml:
https://github.com/mozilla/bigquery-etl/blob/459c22e625aadf17c6de90af4fc5255dd41cf91e/sql/moz-fx-data-shared-prod/accounts_db_external/fxa_signin_codes_v1/schema.yaml#L5

akkomar:

PR #5217, opened 2024-03-14
DENG-802 Changed backfill cli commands

wwyc:
* changed `process` cli command to `initiate`
* changed `complete` cli command to use `get_scheduled_backfills`

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3092)

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -41,7 +41,7 @@ stripe==6.4.0
 symbolic==12.8.0
 siggen==2.0.20231009
 tomli==2.0.1 # for backwards compatibility with python < 3.11
-types-python-dateutil==2.8.19.20240311
+types-python-dateutil==2.9.0.20240316
 types-pytz==2024.1.0.20240203
 types-PyYAML==6.0.12.20240311
 types-requests==2.31.0.20240311
--- a/requirements.txt
+++ b/requirements.txt
@@ -1621,9 +1621,9 @@ tomli==2.0.1 \
     --hash=sha256:939de3e7a6161af0c887ef91b7d41a53e7c5a1ca976325f429cb46ea9bc30ecc \
     --hash=sha256:de526c12914f0c550d15924c62d72abc48d6fe7364aa87328337a31007fe8a4f
     # via -r requirements.in
-types-python-dateutil==2.8.19.20240311 \
-    --hash=sha256:51178227bbd4cbec35dc9adffbf59d832f20e09842d7dcb8c73b169b8780b7cb \
-    --hash=sha256:ef813da0809aca76472ca88807addbeea98b19339aebe56159ae2f4b4f70857a
+types-python-dateutil==2.9.0.20240316 \
+    --hash=sha256:5d2f2e240b86905e40944dd787db6da9263f0deabef1076ddaed797351ec0202 \
+    --hash=sha256:6b8cb66d960771ce5ff974e9dd45e38facb81718cc1e208b10b1baccbfdbee3b
     # via -r requirements.in
 types-pytz==2024.1.0.20240203 \
     --hash=sha256:9679eef0365db3af91ef7722c199dbb75ee5c1b67e3c4dd7bfbeb1b8a71c21a3 \
--- a/sql_generators/glean_usage/templates/events_stream_v1.query.sql
+++ b/sql_generators/glean_usage/templates/events_stream_v1.query.sql
@@ -15,7 +15,9 @@ WITH base AS (
         client_info.os AS os,
         client_info.os_version AS os_version,
         client_info.telemetry_sdk_build AS telemetry_sdk_build,
-        client_info.build_date AS build_date
+        client_info.build_date AS build_date,
+        client_info.session_id AS session_id,
+        client_info.session_count AS session_count
       ) AS client_info,
       STRUCT(
         ping_info.seq,


bigquery_etl/cli/backfill.py
```diff
@@ -343,35 +343,43 @@ def scheduled(ctx, qualified_table_name, sql_dir, project_id, status, json_path=
...
     if backfills_to_process_dict:
-        entry_to_process = backfills_to_process_dict[qualified_table_name]
+        entry_to_initiate = backfills_to_process_dict[qualified_table_name]
 
-        click.echo(f"\nValidating backfill for {qualified_table_name} via dry run:")
-        _process_backfill(ctx, qualified_table_name, entry_to_process, dry_run=True)
+        click.echo(
+            f"\nInitiating backfill(s) for {qualified_table_name} with entry date {entry_to_initiate.entry_date} via dry run:"
+        )
+        _initiate_backfill(ctx, qualified_table_name, entry_to_initiate, dry_run=True)
```
ANich, line 360: A potential issue here: Some of these dry runs could take a while, potentially longer than an hour. The dry run also happens before the actual processing starts, which creates the staging table. If the next run starts before the dry run ends, `get_scheduled_backfills` won't see the staging table and return the backfill twice so we'll have two backfills for the same table being initiated concurrently. 

Some options I can think of:
1. Create the staging table immediately and if the dry run fails, delete it. 
2. Choose some other method of validating the backfill besides dry run
3. Limit the initiate DAG run to one max active run

Wdyt?

bigquery_etl/cli/backfill.py
```diff
@@ -343,35 +343,43 @@ def scheduled(ctx, qualified_table_name, sql_dir, project_id, status, json_path=
...
+            f"\nInitiating backfill(s) for {qualified_table_name} with entry date {entry_to_initiate.entry_date}:"
+        )
+        _initiate_backfill(ctx, qualified_table_name, entry_to_initiate)
 
-        click.echo(f"Backfill processing completed for {qualified_table_name}.")
+        click.echo(
+            f"Processed backfill for {qualified_table_name} with entry date {entry_to_initiate.entry_date}"
+        )
+    else:
+        click.echo(f"No backfill processed for {qualified_table_name}")
```
ANich, line 371: nit: This might be easier to read if we reverse the if condition, we won't need an else. Something like:
```
if not backfills_to_process_dict:
    click.echo(f"No backfill processed for {qualified_table_name}")
    return

{remaining code}
```

bigquery_etl/cli/backfill.py
```diff
@@ -343,35 +343,43 @@ def scheduled(ctx, qualified_table_name, sql_dir, project_id, status, json_path=
...
 
 
-def _process_backfill(ctx, qualified_table_name, entry_to_process, dry_run=None):
+def _initiate_backfill(ctx, qualified_table_name, entry_to_initiate, dry_run=None):
```
ANich, line 374: Some nits: since the fn is called `_initiate_backfill` we know the backfill arg is the one to be initiated so we can just call it `entry`. Also type hints!

```suggestion
def _initiate_backfill(ctx, qualified_table_name: str, entry: Backfill, dry_run: bool = None):
```

bigquery_etl/cli/backfill.py
```diff
@@ -425,71 +433,60 @@ def complete(ctx, qualified_table_name, sql_dir, project_id):
...
+        # replace partitions in production table that have been backfilled
+        for backfill_date in dates:
+            if backfill_date in entry_to_complete.excluded_dates:
+                click.echo(f"Skipping excluded date: {backfill_date}")
+                continue
+
+            partition = backfill_date.strftime("%Y%m%d")
+            production_table = f"{qualified_table_name}${partition}"
+            backfill_table = f"{backfill_staging_qualified_table_name}${partition}"
+            _copy_table(backfill_table, production_table, client)
```
ANich, line 477: A todo for the (near) future, not all tables being backfilled are partitioned, even though the backfill requires multiple queries. This should be one copy operation if the table is unpartitioned. No need to fix in this PR, but we should follow up on this fairly soon

bigquery_etl/cli/backfill.py
```diff
@@ -425,71 +433,60 @@ def complete(ctx, qualified_table_name, sql_dir, project_id):
...
-    )
-
-    click.echo(
-        f"Completed backfill for {qualified_table_name} with entry date {entry_to_complete.entry_date}"
-    )
+        click.echo(
+            f"Processed backfill for {qualified_table_name} with entry date {entry_to_complete.entry_date}"
+        )
+    else:
+        click.echo(f"No backfill processed for {qualified_table_name}")
```
ANich, line 489: nit: similar comment to above about switching the condition

ANich (approved):
Left a few nits and follow-up questions, nothing blocking for this PR but would like your thoughts before merging

bigquery_etl/cli/backfill.py
```diff
@@ -425,40 +434,26 @@ def complete(ctx, qualified_table_name, sql_dir, project_id):
...
+    if not backfills_to_process_dict:
+        click.echo(f"No backfill processed for {qualified_table_name}")
+        return
+
+    entry_to_complete = backfills_to_process_dict[qualified_table_name]
 
-    entry_to_complete = entries[0]
     click.echo(
-        f"Completing backfill for {qualified_table_name} with entry date {entry_to_complete.entry_date}:"
+        f"Completing backfill(s) for {qualified_table_name} with entry date {entry_to_complete.entry_date}:"
```
ANich, line 450: It's just one backfill no?

bigquery_etl/cli/backfill.py
ANich: Sounds good

bigquery_etl/cli/backfill.py
ANich: Awesome, thanks

PR #5219, opened 2024-03-15
DENG-2823: Added deprecate cli command

wwyc:
New cli command is added to support deprecation process.
`Deletion date` field is added to help keep track when tables are due to deletion.  This will allow us to capture deletion dates in labels and query them in `bigquery_tables_inventory`.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3098)

### DIFF
--- a/script/glam/generate_glean_sql
+++ b/script/glam/generate_glean_sql
@@ -66,7 +66,7 @@ function write_clients_daily_aggregates {
     tables=$(
         bq ls "$qualified" \
         | grep TABLE \
-        | grep 'glean_ping_\d\+' \
+        | grep -E 'schema_id:glean_ping_\d+' \
         | grep -v use_counters \
         | awk '{print $1}'
     )


kwindau (approved):

Marlene-M-Hirose (approved):

scholtzan:
It would be cool if the command could spit out a warning if it finds queries that are still referencing the table to be deprecated

bigquery_etl/cli/metadata.py
```diff
@@ -126,3 +128,51 @@ def publish(name: str, sql_dir: Optional[str], project_id: Optional[str]) -> Non
...
+    "-d",
+    help="Date when table is scheduled for deletion. Date format: yyyy-mm-dd",
+    type=click.DateTime(formats=["%Y-%m-%d"]),
+    default=datetime.today() + relativedelta(months=+3),
+)
+def deprecate(
+    name: str,
+    sql_dir: Optional[str],
+    project_id: Optional[str],
+    deletion_date,
```
sean-rose, line 159: Suggest type hint here for consistency.

bigquery_etl/cli/metadata.py
```diff
@@ -126,3 +128,51 @@ def publish(name: str, sql_dir: Optional[str], project_id: Optional[str]) -> Non
...
+)
+@click.argument("name")
+@project_id_option(
+    ConfigLoader.get("default", "project", fallback="moz-fx-data-shared-prod")
+)
+@sql_dir_option
+@click.option(
+    "--deletion_date",
+    "--deletion-date",
+    "-d",
```
sean-rose, line 150: For cases like this where it isn't a common parameter IMO it'd be preferable not to have a short option, which are more cryptic.

bigquery_etl/cli/metadata.py
```diff
@@ -126,3 +128,51 @@ def publish(name: str, sql_dir: Optional[str], project_id: Optional[str]) -> Non
...
+        metadata.deprecated = True
+        metadata.deletion_date = deletion_date.date()
+
+        metadata.write(metadata_file)
+        click.echo(f"Updated {metadata_file} with deprecation.")
+
+    if not table_metadata_files:
+        raise FileNotFoundError(f"No metadata file(s) were found for: {name}")
+
+    return None
```
sean-rose, line 178: Seems like the `return None` could be left off.  Or is Black requiring that for some reason?

bigquery_etl/metadata/parse_metadata.py
```diff
@@ -153,6 +154,7 @@ class Metadata:
     references: Dict = attr.ib({})
     external_data: Optional[ExternalDataMetadata] = attr.ib(None)
     deprecated: bool = attr.ib(False)
+    deletion_date: date = attr.ib(None)
```
sean-rose, line 157:
```suggestion
    deletion_date: Optional[date] = attr.ib(None)
```

bigquery_etl/cli/metadata.py
BenWu: I'll also add that `sql_dir` and `project_id` are optional as command line args but they're not expected to be null in the function.  I think `paths_matching_name_pattern` would actually error if sql_dir was null.

This might cause false warnings in code inspectors but maybe not in this case since `paths_matching_name_pattern` doesn't use type hints.

bigquery_etl/cli/metadata.py
```diff
@@ -125,4 +127,49 @@ def publish(name: str, sql_dir: Optional[str], project_id: Optional[str]) -> Non
...
+        metadata.deprecated = True
+        metadata.deletion_date = deletion_date.date()
+
+        metadata.write(metadata_file)
+        click.echo(f"Updated {metadata_file} with deprecation.")
+
+    if not table_metadata_files:
+        raise FileNotFoundError(f"No metadata file(s) were found for: {name}")
+
     return None
```
sean-rose, line 175: I just noticed this other `return None` that could be removed, and that actually this is currently resulting in `return None` being removed from the `publish()` function, which is unrelated to this PR.

So I'd suggest removing `return None` here and putting it back at the end of the `publish()` function to avoid having an unrelated change.

tests/cli/test_cli_metadata.py
```diff
@@ -311,3 +315,73 @@ def test_metadata_publish_with_no_metadata_file(
...
+            distutils.dir_util.copy_tree(str(TEST_DIR), str(tmpdirname))
+
+            qualified_table_name = (
+                "moz-fx-data-shared-prod.telemetry_derived.clients_daily_v6"
+            )
+            result = runner.invoke(
+                deprecate,
+                [
+                    qualified_table_name,
+                    "--deletion_date=2024-03-02",
```
kik-kik, line 354: Should we also have a test case for when the date format passed is incorrect?

kik-kik (approved):

PR #5122, opened 2024-02-27
feat(): Acoustic configuration updated, init.sql files removed and user facing views defined

kik-kik:
feat(): Acoustic configuration updated, init.sql files removed and user facing views defined

Follow-up: https://github.com/mozilla/lookml-generator/pull/894

---

Checklist for reviewer:

- [ ] Commits should reference a bug or github issue, if relevant (if a bug is referenced, the pull request should include the bug number in the title).
- [ ] If the PR comes from a fork, trigger integration CI tests by running the [Push to upstream workflow](https://github.com/mozilla/bigquery-etl/actions/workflows/push-to-upstream.yml) and provide the `<username>:<branch>` of the fork as parameter. The parameter will also show up
in the logs of the `manual-trigger-required-for-fork` CI task together with more detailed instructions.
- [ ] If adding a new field to a query, ensure that the schema and dependent downstream schemas have been updated.
- [ ] When adding a new derived dataset, ensure that data is not available already (fully or partially) and recommend extending an existing dataset in favor of creating new ones. Data can be available in the [bigquery-etl repository](https://github.com/mozilla/bigquery-etl), [looker-hub](https://github.com/mozilla/looker-hub) or in [looker-spoke-default](https://github.com/mozilla/looker-spoke-default/tree/e1315853507fc1ac6e78d252d53dc8df5f5f322b).

For modifications to schemas in restricted namespaces (see [`CODEOWNERS`](https://github.com/mozilla/bigquery-etl/blob/main/CODEOWNERS)):
- [ ] Follow the [change control procedure](https://docs.google.com/document/d/1TTJi4ht7NuzX6BPG_KTr6omaZg70cEpxe9xlpfnHj9k/edit#heading=h.ttegrcfy18ck)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-2892)

PR #5220, opened 2024-03-15
Remove workgroup:mozilla-confidential access from user-facing acoustic views

whd:
This is one potential resolution of the discussion in https://github.com/mozilla/bigquery-etl/pull/5122#pullrequestreview-1939418017. It keeps `moz-fx-data-shared-prod.acoustic` and `mozdata.acoustic` accessible to `workgroup:mozilla-confidential` but every view in it that references `moz-fx-data-marketing-prod.acoustic` will only be query-able by members of `workgroup:braze`. You'd want to keep the `authorized: true` metadata only for those views without PII that you want authorized so that `workgroup:mozilla-confidential` can access them.

The perhaps more straightforward thing to do  is have the same `workgroup_access` applied to `moz-fx-data-marketing-prod.acoustic` as `moz-fx-data-shared-prod.acoustic`. We should do this  if we we're not trying to expose some of the non-PII data in the user-facing datasets to members outside of `workgroup:braze` ( `workgroup:mozilla-confidential`).

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3099)


kik-kik:
Hey @whd,

I'm not sure I fully understand this change. Let me explain what we are trying to achieve. So the data inside `moz-fx-data-marketing-prod.acoustic` should only be accessible by users that are part of the `braze` workgroup. This is mainly due to the following two datasets:
- [contact_raw_v1](https://github.com/mozilla/bigquery-etl/tree/main/sql/moz-fx-data-marketing-prod/acoustic/contact_raw_v1)
- [raw_recipient_raw_v1](https://github.com/mozilla/bigquery-etl/tree/main/sql/moz-fx-data-marketing-prod/acoustic/raw_recipient_raw_v1)

Which contain PII data, the remaining 3 datasets should not as far as we understand:
- [raw_recipient_v1](https://github.com/mozilla/bigquery-etl/tree/main/sql/moz-fx-data-marketing-prod/acoustic/raw_recipient_v1)
- [contact_v1](https://github.com/mozilla/bigquery-etl/tree/main/sql/moz-fx-data-marketing-prod/acoustic/contact_v1)
- [contact_current_snapshot_v1](https://github.com/mozilla/bigquery-etl/tree/main/sql/moz-fx-data-marketing-prod/acoustic/contact_current_snapshot_v1) 

Could you please confirm if any of the following fields would be considered as such?

```
- moz-fx-data-marketing-prod.acoustic.contact_v1
email_id
recipient_id
basket_token
sfdc_id
fxa_id

---

- moz-fx-data-marketing-prod.acoustic.contact_current_snapshot_v1
email_id
recipient_id
basket_token
sfdc_id
fxa_id

---

- moz-fx-data-marketing-prod.acoustic.raw_recipient_v1
email
recipient_id
report_id
mailing_id
campaign_id
content_id
```

All of those acoustic tables live in the same BQ dataset primarily to keep them together in one place rather than fractioning them across multiple projects. Also, this was set up prior to us utilizing the `_external` dataset pattern for restricted data and then `_derived` for stripped down, more accessible data. We could always change this structure and move `contact_v1`, `contact_current_snapshot_v1`, and `raw_recipient_v1` inside `moz-fx-data-shared-prod.acoustic_derived` dataset, but I'm also not sure it is worth the effort since those tables will cease to be populated in the near future. This is why it seems like making those specific tables accessible via authorized_views seems like the most straight forward path to take.

Please let us know your thoughts.

whd:
@kik-kik thanks for explaining the intent here: the configuration make a lot more sense to me now :sweat_smile:. The combination of these factors led me to review this with more scrutiny and come to incorrect conclusions:

1. all the views defined were authorized as  `select *` from a dataset classified as workgroup-confidential into a dataset classified as mozilla-confidential
2. there are two source tables: `raw_recipient_raw_v1`, `raw_recipient_v1`, both with `raw` in the name
  a. `moz-fx-data-marketing-prod.acoustic.raw_recipient_v1` does NOT have `email` in it, it has `email_id` (the description above I believe get this wrong)

(1) above triggers a review from DSRE before ACLs authorizations are applied. This is partially because authorizations depending on the dataset being authorized may need to be deployed manually e.g. in cases where the [source datasets are administered by other teams](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/activity_stream/tile_id_types/view.sql). I implicitly assumed `raw` in the table name meant there was some PII in it but that's not the case here. The configuration is fine as-is but the thing I would say would make it more obvious that the workgroup-access change is permissible and expected would be to make leverage the authorizations as views on top of `raw_recipient_raw_v1` i.e.
1. write data to fivetran
2. etl to copy it into `moz-fx-data-marketing-prod.acoustic.raw_recipient_raw_v1`
3. make `moz-fx-data-marketing-prod.acoustic.raw_recipient_v1` and `moz-fx-data-shared-prod.acoustic.raw_recipient_v1`  (authorized) `SELECT * EXCEPT()` (plus the other formatting changes) instead of having a denormalized copy of the data

In general any kind of annotation or metadata about the sensitivity of data in tables is useful for people like me reviewing without much context, so e.g. noting in descriptions which tables have PII and which don't would be useful. Authorized views in particular are subject of higher scrutiny. If the PR had `SELECT * EXCEPT` for at least one authorized view I probably would have assumed the PR was correct.

kik-kik:
@whd thanks a lot for taking the time. I agree generally this is not the most optimal set up and could be improved. However, as I mentioned, this should be going away soon so it's probably not worth investing time now.

And I think you raise a very good point around clearer descriptions clearly indicating the content and the intent of the data inside. It would probably be worth raising this as a more general topic and coming up with some sort of standard for this. As far as I am aware currently there isn't one.

Thanks again! 🙇

PR #5233, opened 2024-03-19
Filter tables using the `glean-min` schema

quiiver:
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3138)


edugfilho (approved):

script/glam/generate_glean_sql
```diff
@@ -60,11 +60,13 @@ function write_clients_daily_aggregates {
...
     local tables;
-    # we're excluding use_counters because they are not supported in GLAM
-    # and generated query is too big to run
+    # GLAM only supports tables with glean_ping_* schema.
+    # Also excluding use_counters because they are not supported
+    # and their generated query is too big to run
     tables=$(
         bq ls "$qualified" \
         | grep TABLE \
+        | grep 'glean_ping_\d\+' \
```
sean-rose, line 69: IMO this is somewhat risky because `grep` is line based and `bq ls` lists each table label on a separate line, so only the very first label will be considered due to `grep TABLE` filtering to just the initial line for each table.

It looks like Glean tables currently have consistent `schema_id` and `schemas_build_id` labels, and since `bq ls` appears to always list labels in alphabetical order and `schema_id` < `schemas_build_id` this works for now.  However, if Glean tables ever get labels that are alphabetically before `schema_id` this will silently stop working. I'd suggest adding an error check for no matching tables being found so we at least get alerted if that ends up happening (though that might not be feasible if there are currently legit cases where no tables would be found).

Also, I think the label pattern should be made a bit more precise, as future ping schemas might have `glean_ping_#` as a suffix (e.g. the min schema name could've just as easily been `min-glean_ping_1`):
```suggestion
        | grep 'schema_id:glean_ping_\d\+' \
```

PR #5241, opened 2024-03-20
Bug 1879826 - events stream: Default enable for all apps with an exclusion list

badboy:
Inverting the check, excluding tiktok just because it was excluded for the other part too and to showcase it.

What apps do we need to exclude here to begin with?

@scholtzan :
With respect to `datasets` / `skip_datasets`: I assumed we would get a merged view for `fenix` over the app IDs, but I guess that never actually happened. Will this happen with the inversion now? Is there some manual step we need to make it happen?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3154)

### DIFF
--- a/sql_generators/glean_usage/templates/cross_channel_events_stream.query.sql
+++ b/sql_generators/glean_usage/templates/cross_channel_events_stream.query.sql
@@ -16,5 +16,5 @@ SELECT
     "{{ channel }}" AS normalized_channel
     {% endif -%}
   ),
-FROM `{{ project_id }}.{{ dataset }}_derived.events_stream` AS e
+FROM `{{ project_id }}.{{ dataset }}.events_stream` AS e
 {% endfor %}
--- a/sql_generators/glean_usage/templates/cross_channel_events_stream.query.sql
+++ b/sql_generators/glean_usage/templates/cross_channel_events_stream.query.sql
@@ -9,12 +9,10 @@ UNION ALL
 SELECT
   "{{ dataset }}" AS normalized_app_id,
   e.*
-  REPLACE(
-    {% if app_name == "fenix" -%}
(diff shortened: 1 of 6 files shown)


sql_generators/glean_usage/templates/cross_channel_events_stream.query.sql
```diff
@@ -16,5 +16,5 @@ SELECT
     "{{ channel }}" AS normalized_channel
     {% endif -%}
   ),
-FROM `{{ project_id }}.{{ dataset }}_derived.events_stream` AS e
+FROM `{{ project_id }}.{{ dataset }}.events_stream` AS e
```
scholtzan, line 19: this should get us `fenix` views

scholtzan (approved):

sql_generators/glean_usage/templates/cross_channel_events_stream.query.sql
quiiver: does this mean the output datasets will change locations?

quiiver (approved):

sql_generators/glean_usage/templates/cross_channel_events_stream.query.sql
badboy: @scholtzan what wil asks. And how can we verify this?

scholtzan: Nothing is changing locations from this change. It's just that we either need to reference the views (like `org_mozilla_fenix.events_stream`) or we could also reference the tables which are in the `_derived` datasets and add the table version (`org_mozilla_fenix_derived.events_stream_v1`)

PR #5269, opened 2024-03-22
Avoid table stage deploy when there are no queries

BenWu:
Follow up to https://github.com/mozilla/bigquery-etl/pull/5262

This fixes table deploys by allowing tuples in `paths_matching_name_pattern` since click nargs are passed in as tuples instead of lists.  [airflow log](https://workflow.telemetry.mozilla.org/dags/bqetl_artifact_deployment/grid?root=&dag_run_id=manual__2024-03-22T00%3A47%3A18.169752%2B00%3A00&task_id=publish_new_tables&tab=logs)

Also fixes stage deploys when there are no queries ([CI log](https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/31671/workflows/efa8d915-53d0-4d79-9f30-ac9a85ec739c/jobs/344716)) by not attempting to deploy.  This was the previous behaviour.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3201)

### DIFF
--- a/bigquery_etl/cli/stage.py
+++ b/bigquery_etl/cli/stage.py
@@ -363,29 +363,30 @@ def _deploy_artifacts(ctx, artifact_files, project_id, dataset_suffix, sql_dir):
         and "*" not in file.parent.name and file.parent.name != "INFORMATION_SCHEMA"
     ]
 
-    # checking and creating datasets needs to happen sequentially
-    for query_file in query_files:
-        dataset = query_file.parent.parent.name
-        create_dataset_if_not_exists(
-            project_id=project_id, dataset=dataset, suffix=dataset_suffix
-        )
+    if len(query_files) > 0:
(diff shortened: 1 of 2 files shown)


BenWu:
Error for stage deploy is
```
File "/root/project/bigquery_etl/cli/query.py", line 2068, in deploy
    output_dir=ctx.obj["TMP_DIR"],
               ~~~~~~~^^^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable
```
My understanding is that when doing a stage deploy, this ctx is of the stage click group and `obj["TMP_DIR"]` is only set for query commands so it fails.  This wasn't an issue before because stage deploys never passed in empty query lists so it never reached this code path

scholtzan (approved):

PR #5282, opened 2024-03-26
fix(managed-backfills): fix date formats from CLI and from backfill initiate

ANich:
There were some mismatched types between how we invoke CLI backfills causing excludes to be ignored 

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3231)

### DIFF
--- a/tests/cli/test_cli_query.py
+++ b/tests/cli/test_cli_query.py
@@ -557,8 +557,8 @@ class TestQuery:
                     "--project_id=moz-fx-data-shared-prod",
                     "--start_date=2021-01-05",
                     "--end_date=2021-01-09",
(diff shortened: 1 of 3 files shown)


tests/cli/test_cli_query.py
```diff
@@ -557,19 +557,19 @@ def test_query_backfill_with_offset(self, runner):
                     "--project_id=moz-fx-data-shared-prod",
                     "--start_date=2021-01-05",
                     "--end_date=2021-01-09",
+                    "--exclude=2021-01-06",
                     "--parallelism=0",
-                    "--no-partition",
```
ANich, line 561: Was removed much earlier but forgot to update tests

PR #5285, opened 2024-03-27
GROWTH-147 Add new LTV UDF called "get_state_ios_v2"

kwindau:
Create a new UDF for getting the iOS LTV state

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3239)

### DIFF
--- a/sql/mozfun/ltv/get_state_ios_v2/udf.sql
+++ b/sql/mozfun/ltv/get_state_ios_v2/udf.sql
@@ -48,3 +48,9 @@ RETURNS STRING AS (
     active
   )
 );
+
+--Tests
+SELECT
+mozfun.assert.equals('2_dow4_2_3', ltv.get_state_ios_v2(10, 2,'2024-03-27', 300, 10, 3, 3)),
+mozfun.assert.equals('54_dow4_00_0', ltv.get_state_ios_v2(500, 436, '2024-03-27', 300, 0, 0, 54)),
+mozfun.assert.equals('4_dow4_1_22', ltv.get_state_ios_v2(28,1, '2024-03-27', 300, 3, 22, 4))
--- a/sql/mozfun/ltv/get_state_ios_v2/udf.sql
+++ b/sql/mozfun/ltv/get_state_ios_v2/udf.sql
@@ -1,0 +1,35 @@
+CREATE OR REPLACE FUNCTION ltv.get_state_ios_v2 (
+   adjust_network STRING,
+   days_since_first_seen INT64,
+   days_since_seen INT64,
+   submission_date DATE,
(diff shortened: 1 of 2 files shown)


sql/mozfun/ltv/get_state_ios_v2/udf.sql
```diff
@@ -0,0 +1,50 @@
...
+        AND 21
+        THEN '3'
+      WHEN pattern > 21
+        THEN '4'
+      ELSE CAST(pattern AS string)
+    END,
+    '_',
+    active
+  )
+);
```
kik-kik, line 50: Assert / test statement at the end would be nice 🙃 Here's an example:
https://github.com/mozilla/bigquery-etl/blob/main/sql/mozfun/norm/extract_version/udf.sql#L16-L32

kwindau: Added - thanks for the suggestion!

sql/mozfun/ltv/get_state_ios_v2/udf.sql
```diff
@@ -0,0 +1,56 @@
...
+        BETWEEN 8
+        AND 14
+        THEN '2'
+      WHEN pattern
+        BETWEEN 15
+        AND 21
+        THEN '3'
+      WHEN pattern > 21
+        THEN '4'
+      ELSE CAST(pattern AS string)
```
kik-kik, line 45: Should we have a default value in case of nulls? Trying this in the console it appears if one of the values inside the concat is null the whole return values becomes null:
```
SELECT CONCAT("test", CAST(NULL AS STRING))
UNION ALL
SELECT CONCAT("test", "_", "test2", CAST(NULL AS STRING))
```
returns:
```
null
null
```

kwindau: @irrationalagent As the UDF stands, if any input value is null, the whole result is null - is this intended or do you want any logic added for null handling?  Or are all of these values always not null?

irrationalagent: The way we have been handling this is to manage the input so that it is never null (i.e. by using COALESCE, or just filtering out nulls in the input dataset). Let's continue to do that - these functions aren't really meant for use outside of the datasets that we tailor for the input.

kwindau: Sounds good, thanks for confirming!

Marlene-M-Hirose (approved):

irrationalagent (commented):
LGTM! thank you

PR #5296, opened 2024-03-29
Make updates to serp events table

dzeber:
Update the serp_events table design to fix some issues and make it easier to use. [This query](https://console.cloud.google.com/bigquery?sq=687003905036:011b738339b7402b9917757949d881c6) gives a sample of how the result of querying the new view would look.

Summary of changes:
- 1 row per impression ID with ad impression & engagement info grouped into arrays (previous version has multiple rows per impression ID)
- Count all engagement actions rather than specific ones. New engagement actions added to the events later on will automatically get included in the table.
- Mapping logic moved to view layer
- Change to the logic in counting visible ads following updated discussions with Eng.
    - previously, we used `num_ads_showing` to count visible ads, depending on the ad blocker inference logic.
    - we now use reported `ads_visible` as the count of visible ads, and `num_ads_showing` column is removed
    - `_reported` suffix dropped from ad impression counts
    - change naming of `_hidden` to `_blocked`
- Change naming in table columns from 'ad impressions' to 'component/element impressions' to reflect broader usage of `ad_impression` event
- Add summary columns for impression-level ad impressions & engagements

Jira ticket: [RS-993](https://mozilla-hub.atlassian.net/browse/RS-993)

Checklist for reviewer:

[RS-993]: https://mozilla-hub.atlassian.net/browse/RS-993?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3269)

### DIFF
--- a/sql_generators/serp_events/templates/desktop_query.sql
+++ b/sql_generators/serp_events/templates/desktop_query.sql
@@ -1,3 +1,18 @@
+-- CREATE TEMP FUNCTION label_ad_components(component_info ARRAY) AS (
+--   ARRAY(
+--     SELECT AS STRUCT
+--       c.*,
+--       COALESCE(
+--         c.component IN ('ad_carousel', 'ad_image_row', 'ad_link', 'ad_sidebar', 'ad_sitelink'),
+--         FALSE
+--       ) AS is_ad_component
+--     FROM UNNEST(component_info) AS c
+--   )
+-- );
+-- TODO:
+-- ad blocker inferred
+--   COALESCE(is_engaged, FALSE) AS is_engaged,
+--   COALESCE(has_ads_loaded, FALSE) AS has_ads_loaded,
 WITH raw_serp_events AS (
   SELECT
     *,
@@ -58,11 +73,15 @@ impressions AS (
     mozfun.norm.browser_version_info(client_info.app_display_version) AS browser_version_info,
     sample_id,
     ping_info.experiments,
-    -- SERP session characteristics
-    CAST(mozfun.map.get_key(event.extra, 'is_shopping_page') AS bool) AS is_shopping_page,
+    -- SERP impression features
+    COALESCE(
+      SAFE_CAST(mozfun.map.get_key(event.extra, 'is_shopping_page') AS bool),
+      FALSE
+    ) AS is_shopping_page,
+    COALESCE(SAFE_CAST(mozfun.map.get_key(event.extra, 'is_private') AS bool), FALSE) AS is_private,
     mozfun.map.get_key(event.extra, 'provider') AS search_engine,
     mozfun.map.get_key(event.extra, 'source') AS sap_source,
-    CAST(mozfun.map.get_key(event.extra, 'tagged') AS bool) AS is_tagged
+    COALESCE(SAFE_CAST(mozfun.map.get_key(event.extra, 'tagged') AS bool), FALSE) AS is_tagged
   FROM
     serp_events
   WHERE
@@ -83,12 +102,11 @@ engagement_counts AS (
   SELECT
     impression_id,
     component,
-    COUNTIF(action = 'clicked') AS num_clicks,
-    COUNTIF(action = 'expanded') AS num_expands,
-    COUNTIF(action = 'submitted') AS num_submits,
+    action,
+    COUNT(*) AS num_engagements
   FROM
     (
-    -- 1 row per engagement event
+      -- 1 row per engagement event
       SELECT
         impression_id,
         mozfun.map.get_key(event.extra, 'action') AS action,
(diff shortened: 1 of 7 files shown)


jaredsnyder:
Could you provide more detail on what "replacing" `num_ads_showing` with `ads_visible`?  Are we essentially choosing to not consider a special case where some but not all ads are hidden?

sql_generators/serp_events/templates/desktop_query.sql
```diff
@@ -193,64 +188,25 @@ SELECT
   sample_id,
   experiments,
   is_shopping_page,
+  is_private,
```
jaredsnyder, line 191: Does this mean we have info about PBM now?

jaredsnyder (commented):
I don't feel qualified to "Approve" but as a consumer it all looks good.  My one concern is that the DTMO docs, especially the example queries, get updated once this is merged but I don't know what that process looks like.

sql_generators/serp_events/templates/desktop_query.sql
dzeber: yes, this was [added](https://bugzilla.mozilla.org/show_bug.cgi?id=1851495) in 120.

jaredsnyder: Nice!

dzeber:
> Could you provide more detail on what "replacing" `num_ads_showing` with `ads_visible`? Are we essentially choosing to not consider a special case where some but not all ads are hidden?

This is per the latest guidance from Eng. There are a couple reasons why we would could get some ads visible and some hidden, which we can't currently distinguish between, and neither of which we expect to be that common:
1. it could be an ad blocker selectively allowing some ads through
2. it could be an issue with the detection counting an ad link as hidden either erroneously or when it's actually associated with a different, visible ad on the page.

In case (1), `ads_visible` is the correct count. In case (2), when it's associated with a visible ad, then `ads_visible` would be the correct count and adding `ads_hidden` would actually double-count.
So we figure that relying on `ads_visible` at face value is the best approach, absent more specific info on which situation is occurring when.

> My one concern is that the DTMO docs, especially the example queries, get updated once this is merged but I don't know what that process looks like.

There's no automated process to enforce this, it's more of a case where I have a related ticket for updating docs that I plan to address after :).

sql_generators/serp_events/templates/desktop_query.sql
pissac17: @dzeber Are there any metrics that we should not be tracking in PBM, or are we tracking everything just like in non-PBM mode?

pissac17 (approved):

sql_generators/serp_events/templates/desktop_query.sql
dzeber: Not sure, I haven't followed the discussion on this update or dug into the data. I saw this flag was added to the event so I wanted to capture it in this table.

sql_generators/serp_events/templates/desktop_query.sql
```diff
@@ -145,39 +152,27 @@ ad_impression_counts AS (
...
+      STRUCT(
+        component,
+        ads_loaded AS num_elements_loaded,
+        ads_visible AS num_elements_visible,
+        -- elements explicitly hidden using JS/CSS, as by an ad blocker
+        ads_hidden AS num_elements_blocked,
+        -- elements that are not visible but not explicitly blocked, eg. below the fold
+        ads_loaded - ads_visible - ads_hidden AS num_elements_notshowing
+      )
+    ) AS component_impressions
```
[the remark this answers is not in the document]
dzeber: I'm fine with making this change. My only concern would be confusion with top-level "SERP impressions" and the impression_id, but maybe this wouldn't be a major issue.

sql_generators/serp_events/templates/view.sql
```diff
@@ -1,8 +1,85 @@
---- User-facing view. Generated via sql_generators.active_users.
+CREATE TEMP FUNCTION is_ad_component(component STRING) AS (
```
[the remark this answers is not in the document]
dzeber: Sounds good, I'll update that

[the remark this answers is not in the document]
dzeber: @alekhyamoz I moved both of these functions to mozfun

alekhyamoz (approved):
LGTM

dzeber:
These changes have been split off across #5417 and #5427

PR #5300, opened 2024-04-01
feat(): convert all funnel_retention mobile checks to warn instead of fail

kik-kik:
# feat(): convert all funnel_retention mobile checks to warn instead of fail

This is to unblock downsteam jobs from failing to run as the result of the fail checks failing (minor duplication issue that seems to come up).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3282)


gkabbz (approved):
Let's keep an eye on the warnings so they don't end up being large variances.

Marlene-M-Hirose (approved):

PR #5303, opened 2024-04-02
feat: convert KPI source query checks from fail to warn

kik-kik:
# feat: convert KPI source query checks from fail to warn

This is to prevent blocking KPI metric generation but still maintain visibility into potential data issues.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3285)

### DIFF
--- a/sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/checks.sql
+++ b/sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/checks.sql
@@ -1,10 +1,10 @@
-#fail
+#warn
 {{ is_unique(columns=["client_id"]) }}
 
-#fail
+#warn
 {{ not_null(columns=["client_id", "sample_id"], where="submission_date = @submission_date") }}
 
-#fail
+#warn
 {{ min_row_count(1, "submission_date = @submission_date") }}
 
 #warn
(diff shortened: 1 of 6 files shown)


sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/checks.sql
```diff
@@ -1,10 +1,10 @@
-#fail
+#warn
 {{ is_unique(columns=["client_id"]) }}
 
-#fail
+#warn
 {{ not_null(columns=["client_id", "sample_id"], where="submission_date = @submission_date") }}
 
-#fail
+#warn
```
lelilia, line 7: This sounds like it really shoud fail the downstream?
I see how non unique client_ids or something especially with regards to this being a problem might not warrant to stop downstream, but if there is no data ... there is no data to run?

kik-kik: Let's just call it "over-compensating" the other way. We will still see failures and can easily re-run the pipeline from this point.

lelilia: But I thought the whole idea of the fails was that we don't want to rerun expensive queries all the time?

kik-kik: ok, row count checks are back to being a fail. Others we can assume they're okay to not pass and instead should only create a warning not stop ETL execution.

lelilia (approved):
lgtm

PR #5321, opened 2024-04-03
[AD-278] Added new column to detail country market tier as defined by Mozilla Ads

sergiosonline:
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3310)


m-d-bowerman (approved):
lgtm. Only comment is some apprehension about using the case-sensitive 'Tier x' strings in the field. It'll work here -- and even be more convenient to work with -- when it's being used in Looker. It might, though, be a good idea to settle on a format for that field in case it's also used in places, as that field might trip people up who interact with it in BigQuery. I'm thinking about the Rest Of World problem, where there were scattered instances using 'row', 'RoW', or 'ROW' before we settled on 'ROW'.

sergiosonline:
@m-d-bowerman : That's a great point. I'm happy to make it lower case, e.g., `tier 1`, and add an explanation for this to `schema.json`.

sql/moz-fx-data-shared-prod/static/country_codes_v1/data.csv
```diff
@@ -1,253 +1,253 @@
...
-Wallis and Futuna,WF,WLF,Oceania,Polynesia,Unspecified,FALSE,FALSE,FALSE
-Western Sahara,EH,ESH,Africa,Northern Africa,Unspecified,FALSE,FALSE,FALSE
-Yemen,YE,YEM,Asia,Western Asia,Unspecified,FALSE,FALSE,FALSE
-Zambia,ZM,ZMB,Africa,Sub-Saharan Africa,Eastern Africa,FALSE,FALSE,FALSE
-Zimbabwe,ZW,ZWE,Africa,Sub-Saharan Africa,Eastern Africa,FALSE,FALSE,FALSE
-Kosovo,XK,XKK,Unspecified,Unspecified,Unspecified,FALSE,FALSE,FALSE
-Rest of World,ROW,ROW,Unspecified,Unspecified,Unspecified,FALSE,FALSE,FALSE
-Unknown Country Name,??,???,Unknown Region,Unknown Sub-region,Unknown Intermediate Region,FALSE,FALSE,FALSE
+name,code,code_3,region_name,subregion_name,intermediate_region_name,pocket_available_on_newtab,mozilla_vpn_available,sponsored_tiles_available_on_newtab,ads_value_tier
+Afghanistan,AF,AFG,Asia,Southern Asia,Unspecified,FALSE,FALSE,FALSE,null
```
sean-rose, line 2: @sergiosonline this is resulting in the literal string "null" ending up in the `ads_value_tier` column.  If you want actual nulls in that column then its mode would need to be changed from `REQUIRED` to `NULLABLE` and the "null" values in this file should be removed (but not the commas, basically leaving empty values for the column).

PR #5359, opened 2024-04-10
DENG-2621: Add pageload derived tables

wwyc:
https://mozilla-hub.atlassian.net/browse/DENG-2621

Checklist for reviewer:


sql/moz-fx-data-shared-prod/firefox_desktop_derived/pageload_1pct_v1/metadata.yaml
```diff
@@ -0,0 +1,23 @@
...
+  Contains 1pct sample from pageload_v1.  Document_id was used to sample data since most pings did not have client_id.
+owners:
+- wichan@mozilla.com
+- acreskey@mozilla.com
+- dpalmeiro@mozilla.com
+labels:
+  incremental: true
+scheduling:
+  dag_name: bqetl_pageload_v1
+  task_name: firefox_desktop_pageload_1pct__v1
```
BenWu, line 12: You might need to set `--schema_update_option=ALLOW_FIELD_ADDITION'` since the schema is based on the ping instead of schema.yaml
```suggestion
scheduling:
  dag_name: bqetl_pageload_v1
  task_name: firefox_desktop_pageload_1pct__v1
  arguments: ['--schema_update_option=ALLOW_FIELD_ADDITION']
```

sql/moz-fx-data-shared-prod/firefox_desktop_derived/pageload_experiments_v1/query.sql
```diff
@@ -0,0 +1,7 @@
+SELECT
+  *
```
BenWu, line 2: It might be useful to give the option of using a sample id
```suggestion
  * REPLACE (MOD(ABS(FARM_FINGERPRINT(document_id)), 100) AS sample_id)
```
Clustering info in metadata.yaml would need to be updated and the other two tables should have it as well for consistency

sql/moz-fx-data-shared-prod/firefox_desktop_derived/pageload_1pct_v1/schema.yaml
BenWu, line 1: I don't think schemas need to be checked in since it's the ping schema and will outdated  if the ping is changed

dags.yaml
```diff
@@ -1570,3 +1570,23 @@ bqetl_braze_currents:
...
+    email:
+      - telemetry-alerts@mozilla.com
+      - wichan@mozilla.com
+    email_on_failure: true
+    email_on_retry: true
+    end_date: null
+    owner: wichan@mozilla.com
+    retries: 2
+    retry_delay: 30m
+    start_date: '2023-10-01'
```
kik-kik, line 1586: I read this as intention to have the DAG automatically create tasks since this date to "backfill" the table. Is this correct? 

If `catchup` setting for the DAG is not set to TRUE will this correctly create tasks for each of the schedule dates since the start_date until today?

Alternatively, I'd say this to a more recent date and manually backfill the table until the most recent date (could probably use managed backfills here too?) and change the start_date to current date.

sql/moz-fx-data-shared-prod/firefox_desktop_derived/pageload_experiments_v1/query.sql
kik-kik: I'd be cautions about calling this `sample_id`. If I recall correctly client_id is normally used to determine the sample_id value and I'm a bit worried that someone might use this and assume that specific sample_id in this table may correspond to the results they would get another table.

sql/moz-fx-data-shared-prod/firefox_desktop_derived/pageload_nightly_v1/metadata.yaml
```diff
@@ -0,0 +1,23 @@
...
+  Contains pings from nightly channel only from pageload_v1
+owners:
+- wichan@mozilla.com
+- acreskey@mozilla.com
+- dpalmeiro@mozilla.com
+labels:
+  incremental: true
+scheduling:
+  dag_name: bqetl_pageload_v1
+  task_name: firefox_desktop_pageload_nightly__v1
```
kik-kik, line 12: Is there a specific reason why we create separate tables in the `_derived` dataset vs having a table unioning all of them and clustering on channel?

sql/moz-fx-data-shared-prod/firefox_desktop_derived/pageload_experiments_v1/query.sql
BenWu: We do use client_id to calculate sample_id but client/sample ids are null in this table so I'm not worried about trying to connect sampled results from this to another table.  On second thought though this is probably an unnecessary optimization so I think we should leave it as-is and not add complexity

sql/moz-fx-data-shared-prod/firefox_desktop_derived/pageload_1pct_v1/schema.yaml
BenWu: I think the schema is always updated based on the query during the stage deploy https://github.com/mozilla/bigquery-etl/blob/eac0ac80c2e5aa63c23ee69272ebbfcfa97d5e7c/bigquery_etl/cli/stage.py#L440.
And as a result, if the schema.yaml becomes outdated CI shouldn't break.  My thinking is more what's the point of the schema.yaml if it isn't the source of truth, rather than needing to keep it up to date.  This is a nit though and including the schema shouldn't really matter

sql/moz-fx-data-shared-prod/firefox_desktop_derived/pageload_nightly_v1/metadata.yaml
BenWu: I believe it's because of the bytes scanned limit we have in redash (20 TB?) which uses the dry run estimate which usually doesn't take clustering into account

sql/moz-fx-data-shared-prod/firefox_desktop_derived/pageload_1pct_v1/schema.yaml
BenWu: This includes columns from the common ping schema so it wouldn't never change but certainly not often.  But again, this shouldn't matter

dags.yaml
kik-kik: So catchup flag on the DAG means that the scheduler will look at the start_date and the schedule_interval and calculate all execution dates when the DAG should have run and create DAG instances for each of them and run the tasks (think @submission_date is the same as the execution_date in most of our cases).

If the setting is turned off (which if I recall is the case with the autogenerated DAGs), it would only figure out what is the most recent execution_date that should have run for the DAG and only run that one date and continue scheduling future dates as normal.

kik-kik (approved):

PR #5363, opened 2024-04-11
feat(DENG-656): Add client_id deduplication to fenix_derived.clients_last_seen_joined query

kik-kik:
# feat(DENG-656): Add client_id deduplication to fenix_derived.clients_last_seen_joined query

In some rare cases we end up with same client_id in multiple channels in Fenix. In those cases, this union can result in client_id duplicates. This logic ensures that the resulting table only includes the client from the channel we've seen first.

---

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3399)

### DIFF
--- a/sql_generators/glean_usage/templates/clients_last_seen_joined.metadata.yaml
+++ b/sql_generators/glean_usage/templates/clients_last_seen_joined.metadata.yaml
@@ -7,7 +7,7 @@ description: |-
   such that a given client will not appear in the result if
   it is only represented in the metrics-based table.
 
-  NOTE FENIX ONLY:
+  NOTE:
   In rare cases the union can result in the same client_id entry from different channels.
   In such cases, the resulting data will only include the entry for the client_id
   with the oldest first_seen_date value.
--- a/sql_generators/glean_usage/templates/clients_last_seen_joined.metadata.yaml
+++ b/sql_generators/glean_usage/templates/clients_last_seen_joined.metadata.yaml
@@ -6,6 +6,11 @@ description: |-
   Join on baseline and metrics views, but uses a LEFT JOIN
   such that a given client will not appear in the result if
   it is only represented in the metrics-based table.
+
+  NOTE FENIX ONLY:
+  In rare cases the union can result in the same client_id entry from different channels.
+  In such cases, the resulting data will only include the entry for the client_id
+  with the oldest first_seen_date value.
 owners:
   - ascholtz@mozilla.com
 labels:
--- a/sql_generators/glean_usage/templates/clients_last_seen_joined.query.sql
+++ b/sql_generators/glean_usage/templates/clients_last_seen_joined.query.sql
@@ -1,29 +1,29 @@
 WITH baseline AS (
-  SELECT 
+  SELECT
     *
-  FROM 
-    `{{ project_id }}.{{ app_name }}.baseline_clients_last_seen` 
-  WHERE 
+  FROM
+    `{{ project_id }}.{{ app_name }}.baseline_clients_last_seen`
+  WHERE
     submission_date = @submission_date
 ),
 metrics AS (
-  SELECT 
-    * 
-  FROM 
+  SELECT
+    *
+  FROM
     `{{ project_id }}.{{ app_name }}.metrics_clients_last_seen`
   WHERE
     -- The join between baseline and metrics pings is based on submission_date with a 1 day delay,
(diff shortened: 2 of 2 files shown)


sql_generators/glean_usage/templates/clients_last_seen_joined.metadata.yaml
```diff
@@ -6,6 +6,11 @@ description: |-
   Join on baseline and metrics views, but uses a LEFT JOIN
   such that a given client will not appear in the result if
   it is only represented in the metrics-based table.
+
+  NOTE FENIX ONLY:
```
scholtzan, line 10:
```suggestion
```

I believe this could happen for any app. For now we have seen it only on fenix (though I thought at some point this also happened on iOS)

sql_generators/glean_usage/templates/clients_last_seen_joined.query.sql
```diff
@@ -1,33 +1,46 @@
...
-  baseline 
+  baseline
 LEFT JOIN metrics
 ON baseline.client_id = metrics.client_id AND
   baseline.sample_id = metrics.sample_id AND
   (
     baseline.normalized_channel = metrics.normalized_channel OR
     (baseline.normalized_channel IS NULL AND metrics.normalized_channel IS NULL)
   )
+{% if app_name == "fenix" -%}
```
scholtzan, line 34:
```suggestion
```

sql_generators/glean_usage/templates/clients_last_seen_joined.query.sql
```diff
@@ -1,33 +1,46 @@
...
+-- This logic ensures that the resulting table only includes the client from the channel
+-- we've seen first.
+QUALIFY
+  ROW_NUMBER() OVER (
+    PARTITION BY
+      client_id
+    ORDER BY
+      first_seen_date ASC
+) = 1
+{% endif -%}
```
scholtzan, line 46:
```suggestion
```

sql_generators/glean_usage/templates/clients_last_seen_joined.query.sql
```diff
@@ -1,33 +1,46 @@
...
+  baseline
 LEFT JOIN metrics
 ON baseline.client_id = metrics.client_id AND
   baseline.sample_id = metrics.sample_id AND
   (
     baseline.normalized_channel = metrics.normalized_channel OR
     (baseline.normalized_channel IS NULL AND metrics.normalized_channel IS NULL)
   )
+{% if app_name == "fenix" -%}
+-- In some rare cases we end up with same client_id in multiple channels in Fenix
```
scholtzan, line 35:
```suggestion
-- In some rare cases we end up with the same client_id in multiple channels
```

sql_generators/glean_usage/templates/clients_last_seen_joined.query.sql
```diff
@@ -1,33 +1,46 @@
...
+-- In those cases, this union can result in client_id duplicates.
+-- This logic ensures that the resulting table only includes the client from the channel
+-- we've seen first.
+QUALIFY
+  ROW_NUMBER() OVER (
+    PARTITION BY
+      client_id
+    ORDER BY
+      first_seen_date ASC
```
scholtzan, line 44: I wonder if we should use the client we have seen most recently instead? It seems likely that it would "replace" the client we saw before that.

kik-kik: So I'm not sure, because we could end up with rows for the same client_id being mixed over time without a real indicator as to why other than the first_seen_date changing which also could have an effect downstream resulting in duplication when processing clients in batches based on their first_seen_date?

Marlene-M-Hirose (approved):

sql_generators/glean_usage/templates/clients_last_seen_joined.query.sql
BenWu: I think using the first client seen is better than the last since results per client would be stable over time but it's a rare edge case so I'm not also not sure

BenWu (approved):
I'm guessing the duplicates are due to users copying their app profiles or something rather than uuid collision.  Just curious, do you know how to reproduce this on your own client?

kik-kik:
@scholtzan any objections to me merging this?

scholtzan (approved):

PR #5397, opened 2024-04-19
initial add of query.sql WIP for ga downloads_with_attr_v3

Marlene-M-Hirose:
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3470)


kwindau (approved):
would simplify the group by and exclude empty strings/ (not set) but besides that LGTM

SuYoungHong:
I reviewed, left some notes. Supplementary materials here: [Downloads with Attribution V3 Review Doc](https://docs.google.com/document/d/1t0M4cOuZRW43FOUO5cpoo_ThptoPJmd_BHGXCN1Bifo/edit) 

Actually, I would step back and think about what you want this table to be. I think we discussed it a bit last time. To me, there’s two choices: 
	1) Replication of downloads_with_attribution_v2
		- Will need to be all records from the DL token logs (so 1 record per day per dl token) with the following conditions
			- Matched via stub_session_id and GA client id
			- Matched via GA client id only with only one session found
			- Matched via GA client id only with multiple sessions found (one arbitrarily chosen for match) 
			- Has GA client id and/or stubsession_id but not match in GA data
			- No GA info found so cannot match
	2) A table of GA sessions per day 
		(use case, for marketing to use) 
		append the following metrics
		- Has DL (from the GA side) 
		- Has DL token logs match
			- Is there one, is there multiple? 
		- Has new profile 
			- Matched via DL token
			- In case of multiple DL tokens or profiles, log that info
		- Has returning user profile
			- Ditto as above
		Has wk4 retained
			- Ditto as above

PR #5439, opened 2024-04-25
Use % formatting in logging instead of braces

BenWu:
These logs will result in `TypeError: not all arguments converted during string formatting`.  This can be seen by renaming a metadata file and running the query through bqetl, e.g. `bqetl query run firefox_ios_derived.firefox_ios_clients_v1 --dry_run --nouse_legacy_sql --project_id=moz-fx-data-shared-prod`
It doesn't actually crash because the logger will catch it.  It does crash for unit tests though but I didn't look too deep into why.

The logger defaults to the `%` style formatting and not the `{}` https://docs.python.org/3/howto/logging-cookbook.html#use-of-alternative-formatting-styles
This can be changed by configuring a formatter and handler but it didn't seem worth it to me.

Checklist for reviewer:

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -935,7 +935,7 @@ def _run_query(
             logging.error(e)
             sys.exit(1)
         except FileNotFoundError:
-            logging.warning("No metadata.yaml found for {}", query_file)
+            logging.warning("No metadata.yaml found for %s", query_file)
 
         if not use_public_table and destination_table is not None:
             # destination table was parsed by argparse, however if it wasn't modified to
--- a/tests/test_entrypoint.py
+++ b/tests/test_entrypoint.py
@@ -31,12 +31,12 @@ class TestEntrypoint:
                 b"+---+-----+\n| a |  b  |\n+---+-----+\n| 1 | abc |\n+---+-----+\n"
                 in result
             )
-            assert b"No metadata.yaml found for {}" in result
+            assert b"No metadata.yaml found for " in result
         except subprocess.CalledProcessError as e:
             # running bq in CircleCI will fail since it's not installed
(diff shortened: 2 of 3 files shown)


tests/test_entrypoint.py
```diff
@@ -31,12 +31,12 @@ def test_run_query(self, tmp_path, project_id):
                 b"+---+-----+\n| a |  b  |\n+---+-----+\n| 1 | abc |\n+---+-----+\n"
                 in result
             )
-            assert b"No metadata.yaml found for {}" in result
```
BenWu, line 34: It now logs the file path instead of `{}`

bigquery_etl/cli/query.py
```diff
@@ -935,7 +935,7 @@ def _run_query(
             logging.error(e)
             sys.exit(1)
         except FileNotFoundError:
-            logging.warning("No metadata.yaml found for {}", query_file)
+            logging.warning("No metadata.yaml found for %s", query_file)
```
fbertsch, line 938: Any reason not to use an f-string?

kik-kik (approved):

bigquery_etl/cli/query.py
kik-kik: My understanding that `%` is the preferred approach in logging as it is lazily evaluated.

fbertsch: Personally I'd prefer the obviousness of f-strings even at a tiny performance loss, but this color looks fine for a bikeshed as well

BenWu: In this case not really other than it feels like the proper way to do it since logging has this built in.  More generally doing this means the string formatting is only done if it's actually logged based on the logging config which is better for performance.  Not relevant here but kind of interesting to know

PR #5442, opened 2024-04-25
Add billing project option for bqetl queries and backfills

BenWu:
Related tickets:
Closes https://github.com/mozilla/bigquery-etl/issues/3515
https://mozilla-hub.atlassian.net/browse/DENG-845
https://bugzilla.mozilla.org/show_bug.cgi?id=1892284

This adds a `--billing-project` option to bqetl `backfill initiate`, `query backfill`, `query initialize`, `query run`.  This allows queries to run in a different project while referencing the tables in the original query.

This uses the `@@dataset_project_id` and `@@dataset_id` within a session if billing project is set.  The goal here is to not make any changes to the way queries are run if billing project is not set, to limit the complexity/things that can go wrong. 

Some testing notes:
- I wrote a script to dry run every `query.sql` in the repo from a test project as a brute force test and I only got a few permission errors so this should be reasonably robust
- I tested the backfill `create`, `initiate`, `complete` workflow and it seemed to work and run the query in the billing project except for the final copy but that doesn't use slots so I think it's fine

Checklist for reviewer:


BenWu:
cc @wwyc for changes to backfill commands

whd (commented):
I don't have time to review this in depth but I can say that given the current backfill in https://mozilla-hub.atlassian.net/browse/DSRE-1575 is already moved to the new backfill reservation, I'm not concerned that this change is going to have any inadvertent cost effects. We just need to make sure we only switch queries to the on-demand project when it benefits us cost-wise to do so.

akkomar (commented):
Do we need this new argument when there already is `--project_id` in `query run` that allows to run queries in a custom project (for example as in https://mozilla-hub.atlassian.net/browse/DSRE-1575)?
FWIW I haven't looked deep into this, maybe backfill command is different.

BenWu:
AFAIK a new second argument is needed to decouple the project it runs in with the sql directory.  And then you also need to know which project to run in and where the tables are since qualifiers need to be added if they're not the same.
e.g. `bqetl query run firefox_ios_derived.app_store_funnel_v1 --parameter='submission_date:DATE:2024-04-22' --dry_run --project_id=moz-fx-data-backfill-1` will look for the query in `sql/moz-fx-data-backfill-1/`

We could add some logic to parse project from the argument like `moz-fx-data-shared-prod.firefox_ios_derived.app_store_funnel_v1` but to me that's less explicit in usage and code changes would be more complex.

Looking at the DSRE-1575, the query uses a new path in `backfills_staging_derived/` and had the project qualifier was added.  I'm guessing the qualifiers were manually added and the query was moved to `sql/moz-fx-data-backfill-1/backfills_staging_derived/` when Lucia switched the query to run in that project last week.  I asked her to confirm.

BenWu:
Turns out the backfill in https://mozilla-hub.atlassian.net/browse/DSRE-1575 isn't using bqetl cli at all.  This is the script https://github.com/mozilla/bigquery-backfill/blob/main/backfill/2024-03-23-clients_last_seen/telemetry_derived_clients_last_seen_v2_20240322/backfill_clients_last_seen_v2_in_BQproject_backfill_1.py

whd:
> Turns out the backfill in https://mozilla-hub.atlassian.net/browse/DSRE-1575 isn't using bqetl cli at all. This is the script https://github.com/mozilla/bigquery-backfill/blob/main/backfill/2024-03-23-clients_last_seen/telemetry_derived_clients_last_seen_v2_20240322/backfill_clients_last_seen_v2_in_BQproject_backfill_1.py

Thanks for linking this: it likely explains why the slots for that backfill got diverted automatically when I consolidated reservations without this PR landing.

akkomar:
> AFAIK a new second argument is needed to decouple the project it runs in with the sql directory. And then you also need to know which project to run in and where the tables are since qualifiers need to be added if they're not the same. e.g. `bqetl query run firefox_ios_derived.app_store_funnel_v1 --parameter='submission_date:DATE:2024-04-22' --dry_run --project_id=moz-fx-data-backfill-1` will look for the query in `sql/moz-fx-data-backfill-1/`
> 
> We could add some logic to parse project from the argument like `moz-fx-data-shared-prod.firefox_ios_derived.app_store_funnel_v1` but to me that's less explicit in usage and code changes would be more complex.

I guess as a user I would slightly prefer:
1. To have a single `--project_id` argument that works the same as in `bq` cli. BTW it is now documented as `--project_id                  TEXT       GCP project ID` which would need to be updated with this PR.
2. Require queries to fully qualify table names so we don't need to rewrite them.

This probably would require a bit more work so I'm not going to block on this if you think the tradeoff is not worth it. I'd love to also hear @sean-rose 's opinion.

BenWu:
> 1. To have a single `--project_id` argument that works the same as in `bq` cli. BTW it is now documented as `--project_id                  TEXT       GCP project ID` which would need to be updated with this PR.
> 2. Require queries to fully qualify table names so we don't need to rewrite them.

Using a single argument wouldn't allow us to specify that a table and query exists in one project but we want to run it in another unless we parse it from the arg.  But that's still a separate arg, just unnamed.  My expectation is that only a few queries will manually specify a billing project and most people won't need to care about a billing project.  For those cases, the queries are scheduled in airflow and having a second argument seems like the easiest way to work with the dag generation we have.  

Not needing to rewrite queries at run time would be the ideal situation but I think it would be too disruptive to change the expectations around how we write queries in order to support this functionality which should be an implementation detail.

sean-rose:
> bq doesn't support an option like this other than `dataset_id` which only works if one dataset is referenced so the query needs to be rewritten with table references qualified with a project and dataset if they don't have one. A mix of sqlglot parsing and regex is used which I found to be the best way to cover edge cases without too much complexity.

I think it might work to prepend `SET @@dataset_project_id = '...';` to the query to avoid having to rewrite non-fully-qualified table/view/UDF references ([docs](https://cloud.google.com/bigquery/docs/reference/system-variables)).

sean-rose:
> > 1. To have a single `--project_id` argument that works the same as in `bq` cli. BTW it is now documented as `--project_id                  TEXT       GCP project ID` which would need to be updated with this PR.
> 
> Using a single argument wouldn't allow us to specify that a table and query exists in one project but we want to run it in another unless we parse it from the arg. But that's still a separate arg, just unnamed. My expectation is that only a few queries will manually specify a billing project and most people won't need to care about a billing project. For those cases, the queries are scheduled in airflow and having a second argument seems like the easiest way to work with the dag generation we have.

A couple of issues come to mind with having the `--project-id` argument work like it does for the `bq` CLI:
- I'm not sure it'd be advisable to have `bqetl` potentially running queries in all the projects bigquery-etl defines SQL artifacts for (currently there are 10 such projects), in terms of things like managing project permissions, reservations, query monitoring, etc.
- It isn't uncommon for `--project-id` to default to `moz-fx-data-shared-prod`, which would be problematic for non-DE folks who don't have permission to run queries in `moz-fx-data-shared-prod`.

My biggest issue with the existing `--project-id` arguments is their inconsistent behavior between various `bqetl` commands: sometimes they default to a project like `moz-fx-data-shared-prod` or `mozdata`, and sometimes not; sometimes they affect which project queries/commands are run in, and sometimes not (falling back to the default project set in `gcloud` config).

So at the moment I'd lean toward having a new `--billing-project` argument, but if so then I think it'd be good if `--project-id` no longer had any effect on which project queries run in (i.e. always fall back to using the default project set in `gcloud` config if there's no explicit `--billing-project` argument).

I think @scholtzan will be back tomorrow, and I'd be interested to hear her thoughts on this as well.

BenWu:
> I think it might work to prepend SET @@dataset_project_id = '...'; to the query to avoid having to rewrite non-fully-qualified table/view/UDF references ([docs](https://cloud.google.com/bigquery/docs/reference/system-variables)).

That's a good find and I think that would be the better option if it can work.  From a quick initial test I think it can work.  Some limitations are

- can't set a destination table
  - I'll try to put something together that uses DML in a transaction
- dry run on a script only does syntax checking and doesn't catch things like incorrect column names
  - this should be fine since we can assume most queries in the repo pass dry run

I agree with inconsistencies with `--project-id` and that it's worth fixing up but there are a lot of implications with changing how all queries are run which probably warrants a proposal.  I would prefer to start with a `--billing-project` arg that doesn't change any existing behaviour and can be introduced transparently, then decide to what to do with `--project-id` later.  Not because I don't want to do the work, but because this particular change is blocking some immediate significant cost/compute savings with moving to on-demand

sean-rose:
> * can't set a destination table    
>   * I'll try to put something together that uses DML in a transaction
> * dry run on a script only does syntax checking and doesn't catch things like incorrect column names    
>   * this should be fine since we can assume most queries in the repo pass dry run

Good points. And I guess that would also prevent us from using dry runs to get the result schema, which is a key use case.

However, I think those could be worked around using [sessions](https://cloud.google.com/bigquery/docs/sessions-intro), running `SET @@dataset_project_id = '...'` in one query to start the session, and then running the actual query separately in the same session.

BenWu:
I finally had time to get back to this and the session + system variables is working well (from my testing).  This allows me to get rid of the potentially flaky regex so I rewrote this using that method.

I also thought a bit more about the project arguments and it is convoluted at the moment with three different project args (not including `public_project_id`): `project_id`, `billing_project`, and project in `destination_table`.  I think it could make sense to put `billing_project` logic into `project_id`, but at this time I want to limit the number of side effects by not changing existing behaviour in this PR.  I'm not too worried about potentially removing the `billing_project` argument in the future because it shouldn't be widely used.

bigquery_etl/cli/query.py
```diff
@@ -918,24 +941,68 @@ def _run_query(
...
+        query_parts.append(f"SET @@dataset_project_id = '{project_id}';")
+    if dataset_id is not None:
+        query_parts.append(f"SET @@dataset_id = '{dataset_id}';")
+
+    if len(query_parts) == 0:  # need to run a non-empty query
+        session_query = "SELECT 1"
+    else:
+        session_query = "\n".join(query_parts)
+
+    client = bigquery.Client()
```
sean-rose, line 992: This will create the session using the default project, which will cause a "_Session project (...) doesn't match with the project for query job (...)_" error when trying to run `bq query` with `--project_id` set to something different than the session project.

bigquery_etl/cli/query.py
```diff
@@ -918,24 +941,68 @@ def _run_query(
...
+    query_parts = []
+    if project_id is not None:
+        query_parts.append(f"SET @@dataset_project_id = '{project_id}';")
+    if dataset_id is not None:
+        query_parts.append(f"SET @@dataset_id = '{dataset_id}';")
+
+    if len(query_parts) == 0:  # need to run a non-empty query
+        session_query = "SELECT 1"
+    else:
+        session_query = "\n".join(query_parts)
```
sean-rose, line 990: IMO it would make more sense to join the query parts with `;\n` and not have the semicolons as part of the individual statements.

bigquery_etl/cli/query.py
```diff
@@ -918,24 +941,68 @@ def _run_query(
...
+    client = bigquery.Client()
+
+    job_config = bigquery.QueryJobConfig(
+        create_session=True,
+        use_legacy_sql=False,
+    )
+    result = client.query(session_query, job_config)
+
+    if result.session_info is None:
+        raise RuntimeError(f"Failed to get session id with {query}")
```
sean-rose, line 1001: I don't see a `query` variable.
```suggestion
        raise RuntimeError(f"Failed to get session id with {session_query}")
```

bigquery_etl/cli/query.py
```diff
@@ -918,24 +941,68 @@ def _run_query(
...
+    else:
+        session_query = "\n".join(query_parts)
+
+    client = bigquery.Client()
+
+    job_config = bigquery.QueryJobConfig(
+        create_session=True,
+        use_legacy_sql=False,
+    )
+    result = client.query(session_query, job_config)
```
sean-rose, line 998: Should we call `result()` on the job to wait for the query to complete?
```suggestion
    query_job = client.query(session_query, job_config)
    query_job.result()
```

BenWu: Yes that would catch query errors as well. I forgot the actual query was async

bigquery_etl/cli/query.py
BenWu: Right, so this function needs another project id

bigquery_etl/cli/query.py
```diff
@@ -918,24 +941,78 @@ def _run_query(
...
 
+        query_text = render_template(
+            query_file.name,
+            template_folder=str(query_file.parent),
+            templates_dir="",
+            format=False,
+            **addl_templates,
+        )
+
+        # create a session, setting default project and dataset
```
ANich, line 952: Flyby comment: there's ample discussion in this PR, but a comment explaining why the query is run in a session in this case could be useful in future

bigquery_etl/cli/query.py
```diff
@@ -847,14 +861,23 @@ def _run_query(
...
+    project_id is the default project to use with table/view/udf references in the query that
+    do not have a project id qualifier.
+    billing_project is the project to run the query in for the purposes of billing and
+    slot reservation selection.  This is project_id if billing_project is not set
+    """
     if dataset_id is not None:
         # dataset ID was parsed by argparse but needs to be passed as parameter
         # when running the query
-        query_arguments.append("--dataset_id={}".format(dataset_id))
+        query_arguments.append(f"--dataset_id={dataset_id}")
```
sean-rose, line 876: I think `--dataset_id` probably shouldn't be set if `billing_project` is specified.

bigquery_etl/cli/query.py
```diff
@@ -918,24 +941,78 @@ def _run_query(
...
+            session_id = create_query_session(
+                session_project=billing_project,
+                default_project=project_id or default_project,
+                default_dataset=dataset_id or default_dataset,
+            )
+            query_arguments.append(f"--session_id={session_id}")
+
         # write rendered query to a temporary file;
         # query string cannot be passed directly to bq as SQL comments will be interpreted as CLI arguments
         with tempfile.NamedTemporaryFile(mode="w+") as query_stream:
```
sean-rose, line 965: I think you could avoid the need for writing to disk (and the `write()` and `seek()` calls) by using [`io.StringIO`](https://docs.python.org/3.11/library/io.html#io.StringIO).
```suggestion
        with io.StringIO(query_text) as query_stream:
```

sean-rose (approved):
r+wc

bigquery_etl/cli/query.py
BenWu: `StringIO` doesn't seem to work with the subprocess call because a file descriptor is required: `io.UnsupportedOperation: fileno`

One way that would work is
```py
with subprocess.Popen(["bq"] + query_arguments, stdin=subprocess.PIPE) as proc:
    proc.communicate(query_text.encode(encoding="utf-8"))
```
Another way would be removing comments from the sql with some simple regex but I don't see these as strictly better in terms of readability or performance so I'll opt to leave it since it's unrelated to this pr.

bigquery_etl/cli/query.py
BenWu: Yes, `SET @@dataset_id` takes precedence so this shouldn't be set if it doesn't do anything

sean-rose: Oh wait, I forgot I also suggested we only do the session logic if `billing_project` was different than `project_id`, so if they're the same project then the `--dataset_id` arg should still be added.

BenWu: Oh right, I'll fix this in another pr when I add `--billing_project` to something

PR #5446, opened 2024-04-26
Add `event_id` column to `events_unnested` views

sean-rose:
To allow us to uniquely identify each unnested event.

---
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3551)


sql_generators/glean_usage/templates/cross_channel_events_unnested.view.sql
```diff
@@ -62,10 +62,11 @@ SELECT
       ping_info.parsed_end_time
     ) AS ping_info
   ),
+  event_offset + 1 AS event_number,
```
sean-rose, line 65: Now I'm second-guessing the `event_number` column name, since people might misinterpret that as being globally monotonic, when it's only monotonic per document, not overall.

On the other hand, this is kind of following the lead of `event_timestamp`, which also document/ping-specific (number of milliseconds since ping start).  Though not all precedents are good precedents to follow.

Maybe something like `document_event_number` would be a better name?

BenWu: Since this is more metadata than for analysis, I would say it's better to be verbose.  `document_event_number` sounds fine to me.  Or `document_event_sequence`

quiiver: Do you think it makes sense to concat the offset and document id here in the view and call it `event_id`? That should imply a globally unique value for the event.

sean-rose: Adding `event_id` seems reasonable, and would enable similar count-distinct-events logic outside of Looker. Are we guaranteed to always have a document ID though?  (I'd think so, but I noticed some code in [your lookml-generator PR](https://github.com/mozilla/lookml-generator/pull/960/files#diff-803c8f7ebfe44430b73b80daabcc0935d36ebf9ccccfedb5e4609bbb11b9a8e9) was guarding against the document ID possibly not being there)

sean-rose: > Are we guaranteed to always have a document ID though?

I checked all existing `events_unnested` views in `shared-prod` and they all have `document_id` columns, so I'm going to proceed with the `event_id` idea.

quiiver: I think the existing guards in the generator code are a result of some copy paste action. all glean pings should have a document id

PR #5452, opened 2024-04-27
Update all affected test files when deploying to stage

sean-rose:
Otherwise if tables/views that don't have explicit tests but are dependencies for other tests get deployed to stage, then those other tests will fail because the query got rewritten to target the staging versions of the tables/views but the associated test data files didn't get renamed to match ([example such test failures](https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/33104/workflows/e848735e-7a52-4ee4-9a7f-22334b2ce023/jobs/365287) from #5446).

---
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3560)


bigquery_etl/cli/stage.py
```diff
@@ -169,50 +176,48 @@ def deploy(
...
-                            test_dataset = f"{test_dataset}_{dataset_suffix}"
-
-                        name = f"{project_id}.{test_dataset}.{test_name}"
-                        if test_file_path.name.endswith(f".schema{file_suffix}"):
-                            name += ".schema"
-                        name += file_suffix
+                shutil.rmtree(artifact_test_path)
+
+        # rename test files
+        for test_file_path in map(Path, glob(f"{TEST_DIR}/**/*", recursive=True)):
```
sean-rose, line 198: Re-scanning all test files for every deployed artifact isn't ideal, so I'm going to refactor this to process the test files in a single pass after the main artifact processing loop.

One minor downside is the staging dataset suffix logic will be duplicated, but that was already the case before with the existing logic, so I don't see that as a dealbreaker.

quiiver (approved):
Thanks, that makes sense. Would be great if we had some tests around this critical piece of deployment logic but certainly not a blocker for this pr. Approving with the caveat that I'm not super familiar with this code.

bigquery_etl/cli/stage.py
```diff
@@ -169,50 +175,59 @@ def deploy(
...
 
-                        if dataset_suffix:
-                            test_dataset = f"{test_dataset}_{dataset_suffix}"
-
-                        name = f"{project_id}.{test_dataset}.{test_name}"
-                        if test_file_path.name.endswith(f".schema{file_suffix}"):
-                            name += ".schema"
-                        name += file_suffix
+    # rename test files
+    for test_file_path in map(Path, glob(f"{TEST_DIR}/**/*", recursive=True)):
```
BenWu, line 197: Don't you only need to look at the files that were moved in the last step?  I know this wasn't changed in this pr but I think it's worth considering if I'm understanding correctly
```suggestion
    for test_file_path in map(Path, glob(f"{TEST_DIR}/{project_id}/**/*", recursive=True)):
```

sean-rose: No, we need to look at all test files.  Only modified artifacts and upstream dependencies get deployed to stage, but any downstream dependencies of those have their SQL changed to refer to what got deployed to stage, and if those downstream dependencies have tests with data files meant to override the tables/views that did get deployed to stage then those need to get renamed to match.

For example, that's what's happening with #5446 to cause [three tests to fail](https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/33104/workflows/e848735e-7a52-4ee4-9a7f-22334b2ce023/jobs/365287): the `events` views are getting deployed to stage, and `sql/moz-fx-data-shared-prod/fenix_derived/event_types_history_v1/query.sql` and `sql/moz-fx-data-shared-prod/fenix_derived/events_daily_v1/query.sql` are getting rewritten to refer to the staging versions of the `events` views, but the test data files in `tests/sql/moz-fx-data-shared-prod/fenix_derived/event_types_history_v1` and `tests/sql/moz-fx-data-shared-prod/fenix_derived/events_daily_v1` aren't getting renamed to match the staging versions of the `events` views.

PR #5455, opened 2024-04-29
Revert "Store artifacts when `test-sql` CI job fails for troubleshooting"

sean-rose:
Reverts mozilla/bigquery-etl#5451

It turns out the `when:` conditional parameter is only valid for `run` steps, so after #5451 it's actually always storing the artifacts, and I'd rather not increase the CI runtimes across the board (even by the ~1 minute this currently appears to take) just to help troubleshoot occasional test failures.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3569)

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -178,12 +178,6 @@ jobs:
                 name: Run SQL tests
                 command: |
                     PATH="venv/bin:$PATH" script/entrypoint -m sql -n 8
-            - store_artifacts:
-                path: sql
-                when: on_fail
-            - store_artifacts:
-                path: tests
-                when: on_fail
       - unless:
           condition: *validate-sql-or-routines
           steps:


quiiver (approved):

BenWu:
If the extra time is just from the uploading of all the files, could you use a step that moves everything to another directory with `when: on_fail`?  Like `cp -r sql/ debug_artifacts/sql/`

BenWu:
I'll add that I think it's worth a bit of effort to try make this work since this would be very useful to have and would have helped debug a few things in the recent past.  Especially for stage deploys since those can be hard to reason about

sean-rose:
> If the extra time is just from the uploading of all the files, could you use a step that moves everything to another directory with `when: on_fail`? Like `cp -r sql/ debug_artifacts/sql/` and `store_artifacts` uses that path

That's a good idea.  CircleCI's docs even have an [example of doing exactly that](https://circleci.com/docs/artifacts/#uploading-core-files).  I'll give that a shot in #5454, and if it works as expected I'll close this and open another PR to use the same approach in this case.

sean-rose:
Closing in favor of #5457.

PR #5473, opened 2024-05-02
feat: add a debug log statement in case TypeError is raised while formating changed view

kik-kik:
# feat: add a debug log statement in case TypeError is raised while formating changed view

This is to help us identifying the source of the following bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1894701

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3593)

### DIFF
--- a/bigquery_etl/view/__init__.py
+++ b/bigquery_etl/view/__init__.py
@@ -282,12 +282,18 @@ class View:
             print(f"view {target_view_id} will change: does not exist in BigQuery")
             return True
 
-        expected_view_query = CREATE_VIEW_PATTERN.sub(
-            "", sqlparse.format(self.content, strip_comments=True), count=1
-        ).strip(";" + string.whitespace)
-        actual_view_query = sqlparse.format(
-            table.view_query, strip_comments=True
-        ).strip(";" + string.whitespace)
+        try:
+            expected_view_query = CREATE_VIEW_PATTERN.sub(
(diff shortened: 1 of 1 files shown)


bigquery_etl/view/__init__.py
```diff
@@ -282,12 +282,18 @@ def has_changes(self, target_project=None):
...
+            expected_view_query = CREATE_VIEW_PATTERN.sub(
+                "", sqlparse.format(self.content, strip_comments=True), count=1
+            ).strip(";" + string.whitespace)
+
+            actual_view_query = sqlparse.format(
+                table.view_query, strip_comments=True
+            ).strip(";" + string.whitespace)
+        except TypeError as err:
+            print(f"ERROR: There has been an issue formating: {target_view_id}")
+            raise err
```
BenWu, line 295: (In case you didn't know) you can just `raise` to do a reraise.  Doesn't matter though
```suggestion
            raise
```

BenWu (approved):

bigquery_etl/view/__init__.py
kik-kik: ah ok, so it will just automatically use the most recent exception?

BenWu: Yes whatever it's currently "excepting"

PR #5497, opened 2024-05-03
Deploy referenced views to stage

scholtzan:
Related to https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/33384/workflows/b9a2207f-5d42-45fb-b2ac-2616967916bc/jobs/369291

The stage deploys should deploy views that are referenced in queries to stage. We are already doing this for views referenced in other views. Currently, we are missing cases where a query is referencing a view, which might not have changed but is referencing another view with changes. The query still references the original deployed in production in those cases. 
The changes made in this PR should fix this.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3625)


bigquery_etl/cli/stage.py
```diff
@@ -260,13 +261,35 @@ def _udf_dependencies(artifact_files):
...
+            # check query files for view dependencies
+            # views need to be deployed to stage as they might reference other changed artifacts upstream
+            # without getting changed themselves
+            dependencies = extract_table_references(
+                render(dep_file.name, dep_file.parent)
+            )
+            for dependency in dependencies:
+                dependency_components = dependency.split(".")
+                if len(dependency_components) == 3:
+                    # views are always fully qualified, so they'll consist of 3 parts
```
sean-rose, line 282: I don't think view references are always fully qualified ([example](https://github.com/mozilla/bigquery-etl/blob/dd824d66ca2f8a1ca3f546cab758f041d7794155/sql/moz-fx-data-shared-prod/fenix_derived/events_daily_v1/query.sql#L32)).

kik-kik: That is correct, I believe only inside `view.sql` we are required to use fully qualified.

bigquery_etl/cli/stage.py
```diff
@@ -260,13 +261,51 @@ def _udf_dependencies(artifact_files):
...
+                    [
+                        depdendency_project_id,
+                        dependency_dataset_id,
+                        dependency_table_id,
+                    ] = dependency_components
+                # Fully qualify the reference:
+                elif len(dependency_components) == 2:
+                    depdendency_project_id = dep_file.parent.parent.parent.name
+                    [dependency_dataset_id, dependency_table_id] = dependency_components
+                elif len(dependency_components) == 1:
```
kik-kik, line 292: If there is only one component wouldn't it be unlikely the reference contained a `.` character?

bigquery_etl/cli/stage.py
```diff
@@ -260,13 +261,51 @@ def _udf_dependencies(artifact_files):
...
+
+                file_path = (
+                    Path(sql_dir)
+                    / depdendency_project_id
+                    / dependency_dataset_id
+                    / dependency_table_id
+                    / VIEW_FILE
+                )
+                if file_path.exists():
+                    view_dependency_files.append(file_path)
```
kik-kik, line 307: So is the idea here that we recurse over query files?

bigquery_etl/cli/stage.py
```diff
@@ -260,13 +261,51 @@ def _udf_dependencies(artifact_files):
...
+            # check query files for view dependencies
+            # views need to be deployed to stage as they might reference other changed artifacts upstream
+            # without getting changed themselves
+            dependencies = extract_table_references(
+                render(dep_file.name, dep_file.parent)
+            )
+            for dependency in dependencies:
+                dependency_components = dependency.split(".")
+
+                if len(dependency_components) == 3:
```
kik-kik, line 282: Could this be a good place to use `match`?
https://docs.python.org/3/tutorial/controlflow.html#match-statements

bigquery_etl/cli/stage.py
```diff
@@ -260,13 +261,51 @@ def _udf_dependencies(artifact_files):
...
+                    [
+                        depdendency_project_id,
+                        dependency_dataset_id,
+                        dependency_table_id,
+                    ] = dependency_components
+                # Fully qualify the reference:
+                elif len(dependency_components) == 2:
+                    depdendency_project_id = dep_file.parent.parent.parent.name
+                    [dependency_dataset_id, dependency_table_id] = dependency_components
```
kik-kik, line 291: I don't think we need the `[]` here to unpack the values?

kik-kik (approved):
r+wc

bigquery_etl/cli/stage.py
scholtzan: yup

bigquery_etl/cli/stage.py
```diff
@@ -260,13 +261,57 @@ def _udf_dependencies(artifact_files):
...
+                    None,
+                    None,
+                    None,
+                )
+
+                match len(dependency_components):
+                    case 3:
+                        dependency_project_id,
+                        dependency_dataset_id,
+                        dependency_table_id = dependency_components
```
sean-rose, line 291: I was surprised this is valid Python syntax.  I know @kik-kik suggested not using list/tuple syntax for one of the unpacking assignment statements, but I think that was intended just for that case where it all fit on one line.  With Black auto-formatting this so poorly (IMO) I think using list/tuple syntax in this case would be more readable.

bigquery_etl/cli/stage.py
```diff
@@ -260,13 +261,57 @@ def _udf_dependencies(artifact_files):
...
+                        dependency_project_id = dep_file.parent.parent.parent.name
+                        dependency_dataset_id, dependency_table_id = (
+                            dependency_components
+                        )
+                    case 1:
+                        dependency_project_id = dep_file.parent.parent.parent.name
+                        dependency_dataset_id = dep_file.parent.parent.name
+                        dependency_table_id = dependency_components[0]
+                    case _:
+                        return
```
sean-rose, line 303: I don't think we'd want to stop and return nothing at all, just skip this particular dependency.
```suggestion
                        continue
```

PR #5501, opened 2024-05-06
GLAM fix Glean All Os client count for release metrics

edugfilho:
This PR addresses part of https://github.com/mozilla/glam/issues/2768.
Since a 10% sampling was added for Glean Windows/Release, the client count for `All OS` has been wrongly taking only 10% of Windows for Release metrics. This fixes it

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3631)


bigquery_etl/glam/templates/probe_counts_v1.sql
```diff
@@ -1,98 +1,130 @@
...
+  FROM
+    probe_counts
+  WHERE
+    os = "Windows"
+)
+SELECT
+  pc.* EXCEPT (total_users),
+  IF(
+    pc.os = "*",
+    CAST(COALESCE((pc.total_users + (wpc.total_users * 0.9)), pc.total_users) AS INT64),
```
BenWu, line 113: Where does the `* 0.9` come from?  What's the purpose of the join with `windows_probe_counts`?

bigquery_etl/glam/templates/probe_counts_v1.sql
```diff
@@ -1,98 +1,130 @@
...
+              )
+          WHEN metric_type IN ({{ boolean_metric_types }})
+            THEN ['always', 'never', 'sometimes']
+        END,
+        SUM(count)
+      ) AS aggregates
     {% else %}
-        agg_type AS client_agg_type,
-        'histogram' as agg_type,
-        {% if channel == "release" %}
```
BenWu, line 71: Is this not needed anymore because the sample value were multiplied by 10 in `histogram_bucket_counts`?

BenWu (commented):
I don't have all the context but I tried.  There are a few things I don't understand

bigquery_etl/glam/templates/histogram_bucket_counts_v1.sql
```diff
@@ -23,11 +25,12 @@ build_ids AS (
       COUNT(DISTINCT client_id) > {{ minimum_client_count }}),
 normalized_histograms AS (
   SELECT
+    sampled,
     {{ attributes }},
     ARRAY(
       SELECT AS STRUCT
         {{metric_attributes}},
-        mozfun.glam.histogram_normalized_sum(value, 1.0) AS aggregates
+        mozfun.glam.histogram_normalized_sum(value, IF(sampled, 10.0, 1.0)) AS aggregates
```
sean-rose, line 33: It seems like for histogram bucket counts the `os` and `channel` columns are included in the attributes list, and therefore they'd be available to use here, so could you just do something like this rather than using the SQL injection approach?
```suggestion
        mozfun.glam.histogram_normalized_sum(
          value,
          IF(all_combos.os = 'Windows' AND all_combos.channel = 'release', 10.0, 1.0)
        ) AS aggregates
```

bigquery_etl/glam/templates/macros.sql
```diff
@@ -31,6 +31,9 @@ static_combos AS (
         table.* EXCEPT(
             {{ cubed_attributes | join(",") }}
         ),
+        {% if sampled %}
+            {{ sampled_exp }},
+        {% endif %}
```
sean-rose, line 36: - It seems like you could just use `sampled_exp`.
- Having the calling code set the column alias seems a little sketchy.
```suggestion
        {% if sampled_exp %}
            {{ sampled_exp }} AS sampled,
        {% endif %}
```

bigquery_etl/glam/templates/probe_counts_v1.sql
```diff
@@ -1,98 +1,130 @@
...
+    ping_type,
+    app_version,
+    app_build_id,
+    channel,
+    metric,
+    metric_type,
+    KEY,
+    client_agg_type,
+    agg_type
+  )
```
sean-rose, line 130: - Are you absolutely certain there will only be one `windows_probe_counts` row per unique combination of these columns, so this won't cause any fan-outs?
- Should the list of join columns perhaps be dynamically determined using Jinja variables, like the CTE column lists are?

bigquery_etl/glam/templates/macros.sql
```diff
@@ -1,4 +1,4 @@
-{% macro enumerate_table_combinations(source_table, output_table, cubed_attributes, attribute_combinations) %}
+{% macro enumerate_table_combinations(source_table, output_table, cubed_attributes, attribute_combinations, sampled, sampled_exp) %}
```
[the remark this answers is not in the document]
edugfilho: That's correct, they default to `None`. I also didn't know but confirmed after testing.

bigquery_etl/glam/templates/probe_counts_v1.sql
edugfilho: @sean-rose 's answer is the reason. I'm adding comments and will accept the suggested change in a new commit.

bigquery_etl/glam/templates/probe_counts_v1.sql
edugfilho: That's right. I was able to correct the histogram counts at the `bucket_counts` level, but because scalars use `HLL` I had to fix it at `probe_counts`, which is a step later

bigquery_etl/glam/templates/macros.sql
edugfilho: Yes to all!

bigquery_etl/glam/generate.py
```diff
@@ -271,6 +271,7 @@ def main():
                 source_table=f"glam_etl.{args.prefix}__clients_histogram_aggregates_v1",
                 **config[args.prefix],
             ),
+            channel=channel_prefixes[args.prefix],
```
edugfilho, line 274: I added this after realizing the property `channel` in the data is always `*` for these tables (because the table names already contain the channel). The sampling logic must be applied to `Release` only

bigquery_etl/glam/models.py
```diff
@@ -130,6 +130,7 @@ def probe_counts(**kwargs):
 
     return dict(
         attributes=",".join(attributes),
+        attributes_no_os=",".join([attr for attr in attributes if attr != "os"]),
```
edugfilho, line 133: To be used in `probe_counts_v1.sql`

bigquery_etl/glam/templates/histogram_bucket_counts_v1.sql
```diff
@@ -7,7 +7,8 @@ WITH
         source_table,
         "all_combos",
         cubed_attributes,
-        attribute_combinations
+        attribute_combinations,
+        add_windows_release_sample = channel == "release"
```
edugfilho, line 11: Changed the logic here, the macro now only needs a boolean to indicate if the sampled column should be included

bigquery_etl/glam/templates/histogram_bucket_counts_v1.sql
```diff
@@ -7,7 +7,9 @@ WITH
         source_table,
         "all_combos",
         cubed_attributes,
-        attribute_combinations
+        attribute_combinations,
+        true,
+        "table.os = 'Windows' AND channel = 'release' AS sampled"
```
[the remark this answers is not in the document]
edugfilho: The initial reason was because the code usually only processes the last 3 versions of data and the current versions - 3 are all over 119. That said, I'll keep it in case the code gets used for back-filling.

bigquery_etl/glam/models.py
BenWu: Might be a little cleaner to pass in the array and put this logic in the template but not worth changing.

BenWu (approved):
This makes more sense to me now.  I think adding more comments to glam as you go would be useful to make it easier more approachable

edugfilho:
I'm waiting until GLAM Glean ETL catches up (ETA is later today) to land this

edugfilho:
Also, I really really really appreciate both of your thorough reviews. I know this code is messy and you two helped me a lot with the suggestions!

PR #5518, opened 2024-05-07
Deng 3591 braze current syncs

lelilia:
DENG-3591

adding all the different braze currents to bigquery_etl


chelseybeck (approved):
Great work! 🤩

Issue #5364, opened 2024-04-11
Adding the ability to override scheduling parameters on backfill


braunreyes:
Currently the `query backfill` command only supports `{{ds}}` as a `submission_date:DATE:` scheduling parameter.  However, it is possible for folks to add more complex logic for the normal bqetl runs.

For example: https://github.com/mozilla/private-bigquery-etl/blob/56190e888f051c7190241767d0b7519665bc70c9/sql/moz-fx-data-shared-prod/ads_derived/tiles_addressable_inventory_hourly_v1/metadata.yaml#L22

We need to backfill this table for a fill range of days, which means we do not need this custom logic for submission date.  Given the way it is currently written, it seems as though providing the ability to completely override the list of parameters at backfill runtime is the best solution.

If the new flag is used, then scheduling parameters will be completely overridden.  This will not be a 'merge' of parameters.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3400)

PR #5540, opened 2024-05-09
feat(query-backfill): adding a more flexible approach to overriding scheduling attributes

braunreyes:
This PR further expands on allowing for overrides to scheduling attributes when running the `bqetl query backfill` command.  bqetl gives developers the ability to set `date_partition_parameter` to null and manage partitioning manually via the destination table.  This make is impossible to use the backfill dag `bqetl_backfill` because if this parameter is null, it will blow away the whole table.  

For this query: https://github.com/mozilla/private-bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/ads_derived/tiles_addressable_inventory_hourly_v1/metadata.yaml, we need to be able to re-process a whole days worth of data and override the partition or add a new one.

This change would allow us to run a backfill on a query like this with attributes that activate proper daily backfill processing, yet still have the custom logic for the ongoing task runs.

This work will eventually close: https://github.com/mozilla/bigquery-etl/issues/5364 

Checklist for reviewer:

- [X] Commits should reference a bug or github issue, if relevant (if a bug is referenced, the pull request should include the bug number in the title).
- [N/A] If the PR comes from a fork, trigger integration CI tests by running the [Push to upstream workflow](https://github.com/mozilla/bigquery-etl/actions/workflows/push-to-upstream.yml) and provide the `<username>:<branch>` of the fork as parameter. The parameter will also show up
- [N/A] If adding a new field to a query, ensure that the schema and dependent downstream schemas have been updated.
- [N/A] When adding a new derived dataset, ensure that data is not available already (fully or partially) and recommend extending an existing dataset in favor of creating new ones. Data can be available in the [bigquery-etl repository](https://github.com/mozilla/bigquery-etl), [looker-hub](https://github.com/mozilla/looker-hub) or in [looker-spoke-default](https://github.com/mozilla/looker-spoke-default/tree/e1315853507fc1ac6e78d252d53dc8df5f5f322b).

- [N/A] Follow the [change control procedure](https://docs.google.com/document/d/1TTJi4ht7NuzX6BPG_KTr6omaZg70cEpxe9xlpfnHj9k/edit#heading=h.ttegrcfy18ck)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3698)


bigquery_etl/cli/query.py
```diff
@@ -685,14 +686,16 @@ def backfill(
         depends_on_past = metadata.scheduling.get("depends_on_past", False)
         # If date_partition_parameter isn't set it's assumed to be submission_date:
         # https://github.com/mozilla/telemetry-airflow/blob/dbc2782fa23a34ae8268e7788f9621089ac71def/utils/gcp.py#L194C48-L194C48
-        date_partition_parameter = metadata.scheduling.get(
-            "date_partition_parameter", "submission_date"
+        scheduling_overrides_dict = json.loads(scheduling_overrides)
+        date_partition_parameter = scheduling_overrides_dict.get(
+            "date_partition_parameter",
+            metadata.scheduling.get("date_partition_parameter", "submission_date"),
         )
```
ANich, line 693: The below nested `get`'s are fairly hard to read at first glance. I'd suggest something like the following:
```suggestion
        scheduling_metadata = dict(metadata.scheduling) 
        scheduling_metadata.update(json.loads(scheduling_overrides))
        date_partition_parameter = scheduling_metadata.get("date_partition_parameter", "submission_date")
```
and so on for `scheduling_parameters`, `date_partition_offset`, etc.

sean-rose: I agree.  However, assuming you're doing `dict(metadata.scheduling)` to copy the original scheduling metadata dictionary to avoid modifying the original, I'd recommend using `metadata.scheduling.copy()` instead to make that more explicit.

bigquery_etl/cli/query.py
```diff
@@ -626,14 +627,14 @@ def _backfill_query(
     "--checks/--no-checks", help="Whether to run checks during backfill", default=False
 )
 @click.option(
-    "--scheduling_parameters_override",
-    "--scheduling-parameters-override",
-    "-spo",
+    "--scheduling_overrides",
+    "--scheduling-overrides",
+    "-so",
```
sean-rose, line 632: I'd suggest removing the shortened `-so` form as it's kind of non-standard for Unixy CLIs (usually single-dashed options are one character, and such options that are flags can be specified with a single dash followed by multiple characters, like `-so` instead of `-s -o`).

bigquery_etl/cli/query.py
braunreyes: added changes requested...thank you for pointing that out

bigquery_etl/cli/query.py
braunreyes: removed

bigquery_etl/cli/query.py
```diff
@@ -626,14 +627,14 @@ def _backfill_query(
...
     required=False,
-    multiple=True,
-    default=[],
+    type=str,
+    default="{}",
     help=(
-        "Pass a list of parameters to override query's existing scheduling parameters. "
+        "Pass overrides for scheduling sections: parameters and/or date_partition_parameter as needed. "
```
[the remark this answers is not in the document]
braunreyes: updated

ANich (approved):

PR #5396, opened 2024-04-19
Changes to active_users_aggregates query for Mobile

lucia-vargas-a:
This PR implements [DENG-2989](https://mozilla-hub.atlassian.net/browse/DENG-2989) & [DENG-2529](https://mozilla-hub.atlassian.net/browse/DENG-2529) to update the query for active_users_aggregates:

- Use baseline_clients_last_seen tables for Mobile in order to retrieve growth metrics and combine with search for is_default_browser and uri_count.
- Change language_name which is currently not used in queries for locale which has been requested.

Checklist for reviewer:

- [x] Commits should reference a bug or github issue, if relevant (if a bug is referenced, the pull request should include the bug number in the title).
- [ ] If the PR comes from a fork, trigger integration CI tests by running the [Push to upstream workflow](https://github.com/mozilla/bigquery-etl/actions/workflows/push-to-upstream.yml) and provide the `<username>:<branch>` of the fork as parameter. The parameter will also show up
in the logs of the `manual-trigger-required-for-fork` CI task together with more detailed instructions.
- [ ] If adding a new field to a query, ensure that the schema and dependent downstream schemas have been updated.
- [ ] When adding a new derived dataset, ensure that data is not available already (fully or partially) and recommend extending an existing dataset in favor of creating new ones. Data can be available in the [bigquery-etl repository](https://github.com/mozilla/bigquery-etl), [looker-hub](https://github.com/mozilla/looker-hub) or in [looker-spoke-default](https://github.com/mozilla/looker-spoke-default/tree/e1315853507fc1ac6e78d252d53dc8df5f5f322b).

For modifications to schemas in restricted namespaces (see [`CODEOWNERS`](https://github.com/mozilla/bigquery-etl/blob/main/CODEOWNERS)):
- [ ] Follow the [change control procedure](https://docs.google.com/document/d/1TTJi4ht7NuzX6BPG_KTr6omaZg70cEpxe9xlpfnHj9k/edit#heading=h.ttegrcfy18ck)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3469)

[DENG-2989]: https://mozilla-hub.atlassian.net/browse/DENG-2989?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[...]

PR #5548, opened 2024-05-10
Add checks for DAU increased by more than 50% versus previous date.

lucia-vargas-a:
Add checks for DAU increased by more than 50% vs. previous date which could indicate duplicates.
Address the review comment in [PR-5396](https://github.com/mozilla/bigquery-etl/pull/5396/files#r1577016626) related to the join of baseline or main ping and metrics ping in active_users_aggregates.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3709)


kwindau (approved):

sql_generators/active_users_aggregates_v3/templates/mobile_checks.sql
```diff
@@ -57,3 +57,33 @@ SELECT
     ),
     NULL
   );
+
+{% raw -%}
```
bochocki, line 61: Is this `raw` in the correct place? I'm not familiar with it's use,  it it seems different from the other files.

bochocki (approved):
Just one comment, otherwise lgtm!

sql_generators/active_users_aggregates_v3/templates/mobile_checks.sql
kik-kik: In this case `raw` is used to indicate that the text between this and the corresponding end block should not be modified by jinja when generating the output of this template. We have to do this in some cases, for example in this case we generate another jinja template so we want to preserve some of the template features like:

```
`{{ project_id }}.{{ dataset_id }}.{{ table_name }}`
```

kik-kik:
@lucia-vargas-a you may have not see this change. A change has been added to `clients_last_seen_joined` query to remove duplication:

https://github.com/mozilla/bigquery-etl/pull/5363

My question, if we're trying to check for duplication should we use the `is_unique()` macro instead for this purpose?

lucia-vargas-a:
> @lucia-vargas-a you may have not see this change. A change has been added to `clients_last_seen_joined` query to remove duplication:
> 
> #5363
> 
> My question, if we're trying to check for duplication should we use the `is_unique()` macro instead for this purpose?

Thanks for the link. This checks are more oriented towards avoiding overcounting when joining the baseline and metrics ping. I think we could use `is_unique` but it would require uniqueness by all the columns since we don't have a client_id, and that seems more difficult to maintain when adding or removing columns. Wdyt?

sql_generators/active_users_aggregates_v3/templates/desktop_checks.sql
```diff
@@ -165,4 +165,29 @@ SELECT
...
+  SELECT
+    SUM(dau) AS dau
+  FROM
+    `{{ project_id }}.{{ dataset_id }}.{{ table_name }}`
+  WHERE
+    submission_date = DATE_SUB(@submission_date, INTERVAL 1 DAY)
+)
+SELECT
+  IF(
+    ABS((SELECT SUM(dau) FROM dau_current) / (SELECT SUM(dau) FROM dau_previous)) > 1.5,
```
lelilia, line 186: Not sure why you would need the `ABS` here? 
But if it's to account for the unlikely case of negative DAU (?) what about securing against dividing by zero?

PR #5576, opened 2024-05-14
[DENG-3417] Get event pings from probe info for monitoring events

scholtzan:
https://mozilla-hub.atlassian.net/browse/DENG-3417

This PR makes some changes to the `event_monitoring` datasets. Instead of fetching events just from the `event_v1` tables, the generation logic fetches the pings that are sending events from https://probeinfo.telemetry.mozilla.org/glean/ for each app.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3748)

### DIFF
--- a/bqetl_project.yaml
+++ b/bqetl_project.yaml
@@ -356,10 +356,10 @@ generate:
       - topsites-impression # access denied
       - serp-categorization # access denied
       - background-update # table doesn't exist
-      events_tables:
-        accounts_frontend:
+      events_tables: # overwrite event tables
+        accounts_frontend: # event pings land in accounts_events_v1 tables
         - accounts_events_v1
         accounts_backend:
         - accounts_events_v1
-        pine:
+        pine: # Probe info returns pings that don't have tables, only use events_v1
         - events_v1
--- a/bqetl_project.yaml
+++ b/bqetl_project.yaml
@@ -353,9 +353,13 @@ generate:
       - org_mozilla_tiktokreporter
     events_monitoring:
       skip_pings:
-      - topsites-impression
+      - topsites-impression # access denied
+      - serp-categorization # access denied
+      - background-update # table doesn't exist
       events_tables:
         accounts_frontend:
         - accounts_events_v1
         accounts_backend:
         - accounts_events_v1
+        pine:
+        - events_v1
--- a/bqetl_project.yaml
+++ b/bqetl_project.yaml
@@ -352,8 +352,10 @@ generate:
       - org_mozilla_ios_tiktok_reporter_tiktok_reportershare
       - org_mozilla_tiktokreporter
     events_monitoring:
-      default_event_table: events_v1
-      event_table:
-        mozillavpn: main_v1
-        accounts_frontend: accounts_events_v1
-        accounts_backend: accounts_events_v1
+      skip_pings:
+      - topsites-impression
+      events_tables:
+        accounts_frontend:
+        - accounts_events_v1
+        accounts_backend:
+        - accounts_events_v1
--- a/sql_generators/glean_usage/event_monitoring_live.py
+++ b/sql_generators/glean_usage/event_monitoring_live.py
@@ -174,16 +174,44 @@ class EventMonitoringLive(GleanTable):
         aggregate_table = "event_monitoring_aggregates_v1"
         target_view_name = "_".join(self.target_table_id.split("_")[:-1])
 
-        default_events_table = ConfigLoader.get(
-            "generate",
-            "glean_usage",
-            "events_monitoring",
-            "default_event_table",
-            fallback="events_v1",
-        )
         events_table_overwrites = ConfigLoader.get(
-            "generate", "glean_usage", "events_monitoring", "event_table", fallback={}
+            "generate", "glean_usage", "events_monitoring", "events_tables", fallback={}
         )
+
+        event_tables_per_dataset = {}
+
+        for app in apps:
+            for app_dataset in app:
+                dataset = app_dataset["bq_dataset_family"]
+                app_name = [
+                    app_dataset["app_name"]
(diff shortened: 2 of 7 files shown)


sql_generators/glean_usage/event_monitoring_live.py
```diff
@@ -35,6 +39,21 @@ def __init__(self) -> None:
         self.custom_render_kwargs = {}
         self.base_table_name = "events_v1"
 
+    def _get_tables_with_events(self, v1_name: str) -> Set[str]:
+        pings = set()
+        resp = requests.get(METRICS_INFO_URL.format(app_name=v1_name))
+        if resp.status_code == 404:
```
BenWu, line 45: Are there 404s when you run this?  I think an empty json should still exist for apps that don't have metrics

sql_generators/glean_usage/event_monitoring_live.py
```diff
@@ -35,6 +39,21 @@ def __init__(self) -> None:
...
+            return pings
+        resp.raise_for_status()
+        metrics_json = resp.json()
+
+        for _, metric in metrics_json.items():
+            if metric.get("type", None) == "event":
+                latest_history = metric.get("history", [])[-1]
+                pings.update(latest_history.get("send_in_pings", []))
```
BenWu, line 53: This might be an edge not worth considering but only looking at the pings in the latest revision could miss events from older versions of the app if a ping doesn't have any events in the latest version.  It might be safe to ignore those though, just want to make sure that's considered

sql_generators/glean_usage/templates/event_monitoring_aggregates_v1.query.sql
```diff
@@ -2,7 +2,8 @@
 {% for app in apps %}
 {% set outer_loop = loop -%}
 {% for dataset in app -%}
-{% if dataset['bq_dataset_family'] not in ["telemetry", "accounts_frontend", "accounts_backend"] %}
+{% if dataset['bq_dataset_family'] not in ["telemetry", "accounts_frontend", "accounts_backend"]
```
BenWu, line 5: Would it be straightforward to wrap the if around only the different parts for `accounts_frontend` and `accounts_backend`, or are there too many small differences?

scholtzan: Yeah, there are a couple of differences. Having two separate queries seemed more readable in this case

sql_generators/glean_usage/event_monitoring_live.py
scholtzan: There were some cases where pings got seemingly removed. Older history entries were referring to pings that (no longer?) had a table in BigQuery. Using the latest history entry should account for deprecated pings.

Example: There are some events for the [firefox_desktop_background_tasks](https://probeinfo.telemetry.mozilla.org/glean/firefox-desktop-background-tasks/metrics) that got sent as part of `remove-directory` and later as part of `background-tasks` pings. There is no `remove_directory` table.

sql_generators/glean_usage/event_monitoring_live.py
scholtzan: I though I did during testing, but trying it out now all seems good. Removed that part

bqetl_project.yaml
```diff
@@ -352,8 +352,14 @@ generate:
...
-      default_event_table: events_v1
-      event_table:
-        mozillavpn: main_v1
-        accounts_frontend: accounts_events_v1
-        accounts_backend: accounts_events_v1
+      skip_pings:
+      - topsites-impression # access denied
+      - serp-categorization # access denied
+      - background-update # table doesn't exist
+      events_tables:
```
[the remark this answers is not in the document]
scholtzan: Yeah, that's to overwrite the tables events should be queried from. For accounts, probe info lists `events` as ping, but the source tables are called `accounts_events_v1`. I left some comments in the config file

sql_generators/glean_usage/event_monitoring_live.py
BenWu: Maybe it was never added to pings.yaml.  I'm not sure if that would cause errors somewhere.  Ok, makes sense to use the latest

bqetl_project.yaml
```diff
@@ -352,8 +352,14 @@ generate:
       - org_mozilla_ios_tiktok_reporter_tiktok_reportershare
       - org_mozilla_tiktokreporter
     events_monitoring:
-      default_event_table: events_v1
-      event_table:
-        mozillavpn: main_v1
-        accounts_frontend: accounts_events_v1
-        accounts_backend: accounts_events_v1
+      skip_pings:
```
BenWu, line 355: Should these pings be scoped per app?  Maybe with an all apps category.  `firefox_desktop_background_update_stable.background_update_v1` does exist but no events are being sent in it https://probeinfo.telemetry.mozilla.org/glean/firefox-desktop-background-update/metrics

I just looked at the table and I see there are some `nimbus_events` in the events that aren't defined in glean.  Do those need to be considered for these tables? i.e. should those be filtered out or should we also check for pings that have those events?

akkomar (approved):
One comment on accounts situation: we're now working on migrating them from `accounts_events` to `events`. When that's done we'll be able to get rid of some of the complexity here.

bqetl_project.yaml
scholtzan: Yeah, I figured to keep the configuration simple, `background-updates` can generally be ignored. The `nimbus_events` aren't super interesting, I don't think we need to explicitly filter them out, but it's unlikely that anyone will come looking for them as there are better ways to monitor nimbus.

BenWu (approved):

PR #5592, opened 2024-05-16
Fix bugs in `derived_view_schemas` SQL generator

sean-rose:
The `derived_view_schemas` SQL generator was running against all types of directories (not just view directories), and it was reading the view schemas from the currently deployed views (not based on the latest view code).

Changes:
* Limit `derived_view_schemas` SQL generator to actual view directories.

* Fix the `derived_view_schemas` SQL generator to get the view schemas by dry-running their latest SQL and/or from their latest `schema.yaml` file.  Getting the schema from the currently deployed view wasn't appropriate because it wouldn't reflect the latest view code.

* Rename `View.view_schema` to `View.schema`.

* Change `View` so its schema dry-runs use the cloud function (CI doesn't have permission to run dry-run queries directly).

* Apply partition column filters in view dry-run queries when possible for speed/efficiency.

* Don't allow missing fields to prevent view schema enrichment.

* Only copy column descriptions during view schema enrichment.

* Only try enriching view schemas from their reference table `schema.yaml` files if those files actually exist.

* Change `main_1pct` view to select directly from the `main_remainder_1pct_v1` table, so the `derived_view_schemas` SQL generator can detect the partition column to use and successfully dry-run the view to determine its schema.

* Formalize the order `bqetl generate all` runs the SQL generators in.

* Have `bqetl generate all` run `derived_view_schemas` last, in case other SQL generators create derived views.

* Fix `Schema._traverse()` to only recurse if both fields are records.

---
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3767)


sql_generators/derived_view_schemas/__init__.py
```diff
@@ -178,7 +178,13 @@ def generate(target_project, output_dir, parallelism, use_cloud_function):
     ]
 
     for dataset_path in dataset_paths:
-        view_directories = [path for path in dataset_path.iterdir() if path.is_dir()]
+        view_directories = [
+            path
+            for path in dataset_path.iterdir()
+            if path.is_dir()
+            and (path / VIEW_FILE).exists()
+            and not (path / SCHEMA_FILE).exists()
```
BenWu, line 186: I don't think view deploys update the schema from dry run if a schema.yaml exists, unlike table deploys.  So this might cause view deployment to publish outdated schemas and fail.  IMO we can update the schema if it already exists, but if not then any schemas that need to be updated should be included in this PR

sean-rose: I think if there's an explicit `schema.yaml` file for a view it's reasonable to expect that to be kept up to date and be accurate (e.g. [your code for `View.view_schema`](https://github.com/mozilla/bigquery-etl/pull/5057/files#diff-3b6a7e291017a4af9ec3835836f0aac50418cd91772c9c649ddbbc69ab77ba2cR182-R183) seems to make that assumption).

BenWu: It's a safe assumption because those schemas get updated by this generator but that wouldn't be the case with this change.  Running this generator locally, at least `funnel_retention_clients` and `funnel_retention_week_4` have columns added so those would fail to publish here https://github.com/mozilla/bigquery-etl/blob/23fa414aa5f475f013ac5eecb6ab2839519f9f96/bigquery_etl/view/__init__.py#L371-L376
Those should be updated in this PR.  I would still lean towards updating existing schemas since we can't always rely on schemas being up to date which is the case for tables

sean-rose: OK:
- I've changed the `View.view_schema` logic to always try dry-running first, then augment with the `schema.yaml` if one exists.
- I've changed the `derived_view_schemas` SQL generator to use `View.view_schema` to get schemas, rather than getting the deployed view schema with `Schema.for_table()` (which I suspect was probably preventing the logic you implemented in #5057 from working properly).
- I've changed the `derived_view_schemas` SQL generator to not skip views that already have `schema.yaml` files.

bigquery_etl/view/__init__.py
```diff
@@ -176,27 +176,27 @@ def udf_references(self):
 
     @cached_property
     def view_schema(self):
-        """Derive view schema from a schema file or a dry run result."""
-        schema_file = Path(self.path).parent / "schema.yaml"
+        """Derive view schema from a dry run result and/or a schema file."""
+        schema = None
+
+        # check schema based on dry run results
```
BenWu, line 182: I think dry running everything makes sense to make sure everything is updated but it will probably add a few minutes to the publish views task.  Views should usually be up to date since the generator already ran.  I don't have any issues with this but worth checking later on to see how much time this adds.

BenWu (approved):

bigquery_etl/view/__init__.py
```diff
@@ -176,27 +177,45 @@ def udf_references(self):
...
+                query=dedent(
+                    f"""
+                    WITH view_query AS (
+                        {CREATE_VIEW_PATTERN.sub("", self.content)}
+                    )
+                    SELECT *
+                    FROM view_query
+                    WHERE FALSE
+                    """
+                ),
```
sean-rose, line 199: I was pleasantly surprised that dry-running the full `CREATE VIEW ...` query returns the view's schema, nicely side-stepping the issue of potentially having to add query filters on any partition columns from the underlying tables.  However, it turns out that only works if you actually have permission to create the view in the target dataset, which we DEs generally do but CI does not, so the previous dry-run-to-get-view-schema logic would have always failed in a CI context.

bigquery_etl/cli/generate.py
```diff
@@ -99,7 +99,22 @@ def generate_all(ctx, output_dir, target_project, ignore, use_cloud_function):
...
-    for _, cmd in reversed(generate.commands.items()):
+
+    def generator_command_sort_key(command):
+        match command.name:
+            # Run `stable_views` after `glean_usage` because both update `dataset_metadata.yaml` files
+            # and we want the `stable_views` updates to take precedence.
+            case "glean_usage":
+                return (1, command.name)
+            case "stable_views":
+                return (2, command.name)
```
sean-rose, line 110: This is based on [comments in #2594](https://github.com/mozilla/bigquery-etl/pull/2594#discussion_r779871871).

bigquery_etl/view/__init__.py
sean-rose: I did manage to get the dry-runs on some of the views that select from large tables to run quicker by adding partition column filters, but I'm still going to leave this with the original behavior of using the schema from `schema.yaml` by default, with the `derived_view_schemas` SQL generator using custom logic to always try to dry-run the views.

bigquery_etl/cli/generate.py
```diff
@@ -99,7 +99,22 @@ def generate_all(ctx, output_dir, target_project, ignore, use_cloud_function):
...
+            # Run `derived_view_schemas` last in case other SQL generators create derived views.
+            case "derived_view_schemas":
+                return (4, command.name)
+            case _:
+                return (3, command.name)
+
+    for cmd in sorted(generate.commands.values(), key=generator_command_sort_key):
```
sean-rose, line 117: I'm hoping running the SQL generators in a deterministic order will also reduce the seemingly random unrelated diffs that sometimes get reported in `sql.diff`.  Currently the SQL generators are run in the reverse order that `Path.iterdir()` returns the `sql_generators/*` subdirectories, and according to the [docs](https://docs.python.org/3/library/pathlib.html#pathlib.Path.iterdir) that order is "arbitrary".

bigquery_etl/schema/__init__.py
```diff
@@ -247,7 +247,7 @@ def _traverse(
                                     f"for {prefix}.{field_path} are incompatible"
                                 )
 
-                if dtype == "RECORD":
+                if dtype == "RECORD" and nodes[node_name]["type"] == "RECORD":
```
BenWu, line 250: Did an issue with this pop up while you were testing?  I'm `ignore_incompatible_fields` would need to be true

BenWu (approved):
The changes make sense to me but definitely something to keep an eye on to make sure there aren't unintended side effects

bigquery_etl/schema/__init__.py
sean-rose: > Did an issue with this pop up while you were testing?

Yes, it was causing the `derived_view_schemas` generator to fail for a couple views that transform an underlying struct column into something else:
- `telemetry.clients_daily_joined`
- `telemetry.clients_daily_v6`

> I'm `ignore_incompatible_fields` would need to be true

The `derived_view_schemas` generator calls `Schema.merge()` with `attributes=["description"]`, which kind of a has a similar effect since the field incompatibility logic is applied at the attribute level.

bigquery_etl/cli/generate.py
scholtzan: I believe `for _, cmd in reversed(generate.commands.items()):` causes `stable_views` to be run before `glean_usage`. I'm not sure whether it makes a difference here though? I believe `glean_app_ping_views` might depended on the stable views at some point.

scholtzan (approved):

bigquery_etl/cli/generate.py
sean-rose: OK, I've reversed the order in this PR so that `stable_views` runs before `glean_usage`, and now no `dataset_metadata.yaml` changes appear in the `sql.diff`.

PR #5594, opened 2024-05-16
feat: add kpi_support_metrics SQL generator

kik-kik:
Checklist for reviewer:


kwindau (approved):

kwindau (approved):

sql_generators/kpi_support_metrics/templates/active_users.view.sql
```diff
@@ -0,0 +1,53 @@
...
+CREATE OR REPLACE VIEW
+  `{{ project_id }}.{{ dataset }}.{{ name }}`
+AS
+SELECT
+  *,
+  CASE
+    WHEN LOWER(isp) = 'browserstack'
+      THEN CONCAT("{{ friendly_name }}", ' ', isp)
+    ELSE "{{ friendly_name }}"
+  END AS app_name,
```
bochocki, line 11: This needs to account for the Mozilla Online case, too. See, for example, [the desktop query](https://github.com/mozilla/bigquery-etl/blob/4ebf436f057ffb5d4a9d73c4e56a217c19469b6a/sql/moz-fx-data-shared-prod/telemetry/clients_last_seen_v2/view.sql#L30).

```suggestion
  CASE
    WHEN LOWER(isp) = 'browserstack'
      THEN CONCAT("{{ friendly_name }}", ' ', isp)
    WHEN LOWER(distribution_id) = 'mozillaonline'
      THEN CONCAT("{{ friendly_name }}", ' ', distribution_id)
    ELSE "{{ friendly_name }}"
  END AS app_name,
```

sql_generators/kpi_support_metrics/templates/active_users.view.sql
```diff
@@ -0,0 +1,53 @@
...
+  -- Metrics based on pings sent
+  IFNULL(mozfun.bits28.days_since_seen(days_seen_bits) = 0, FALSE) AS is_daily_user,
+  IFNULL(mozfun.bits28.days_since_seen(days_seen_bits) < 7, FALSE) AS is_weekly_user,
+  IFNULL(mozfun.bits28.days_since_seen(days_seen_bits) < 28, FALSE) AS is_monthly_user,
+  {% if is_mobile_kpi %}
+    (
+      LOWER(IFNULL(isp, "")) <> "browserstack"
+      {% if has_mozilla_online %}
+        AND LOWER(IFNULL(distribution_id, "")) <> "mozillaonline"
+      {% endif %}
```
bochocki, line 42: Not blocking, but just wondering if we still need this conditional check, since all the mobile products will have a `distribution_id` column now -- even if it's all NULL for most products.

bochocki (changes requested):
One small issue with the `app_name`, otherwise looks good. Thanks, @kik-kik!

sql_generators/kpi_support_metrics/templates/active_users.view.sql
kik-kik: ah right, I must have copied it over from the firefox_ios version. Added now.

sql_generators/kpi_support_metrics/templates/active_users.view.sql
kik-kik: That is a good point, this should no longer be required. Removed this additional piece of logic.

sql_generators/kpi_support_metrics/templates/active_users.view.sql
```diff
@@ -0,0 +1,53 @@
+{{ header }}
+CREATE OR REPLACE VIEW
+  `{{ project_id }}.{{ dataset }}.{{ name }}`
+AS
+SELECT
+  *,
```
bochocki, line 6: Sorry, I didn't catch the `SELECT *` on my first review.

We don't actually _need_ everything from `baseline_clients_last_seen` -- `submission_date`, `client_id`, `sample_id`, `days_seen_bits`, and `days_active_bits` is sufficient. So this view could be more lightweight. But if DE prefers to have all of the columns available for other tasks (e.g. writing `active_users_aggregates`) then keeping the `SELECT *` works for me.

bochocki (approved):
🙏 Thank you, Kik!

Just one nonblocking comment.

PR #5614, opened 2024-05-18
fix: baseline_clients_daily to only grab distribution_id for fenix app

kik-kik:
fix: baseline_clients_daily to only grab distribution_id for fenix app

This is to address an incorrect assumption that every `baseline` ping would have the same structure. It turns out the field containing distribution_id only exists for Fenix.

---

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3829)

### DIFF
--- a/sql_generators/glean_usage/templates/baseline_clients_daily_v1.query.sql
+++ b/sql_generators/glean_usage/templates/baseline_clients_daily_v1.query.sql
@@ -24,7 +24,11 @@ WITH base AS (
     normalized_channel,
     normalized_os,
     normalized_os_version,
+    {% if has_distribution_id %}
+    metrics.string.metrics_distribution_id AS distribution_id,
+    {% else %}
     CAST(NULL AS STRING) AS distribution_id,
+    {% endif %}
     metadata.geo.subdivision1 AS geo_subdivision,
   FROM
     `{{ baseline_table }}`
--- a/sql_generators/glean_usage/common.py
+++ b/sql_generators/glean_usage/common.py
@@ -36,6 +36,8 @@ NO_BASELINE_PING_APPS = (
     "mozregression",
 )
 
+APPS_WITH_DISTRIBUTION_ID = ("fenix",)
+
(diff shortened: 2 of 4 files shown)


sql_generators/glean_usage/templates/baseline_clients_daily_v1.query.sql
```diff
@@ -24,7 +24,11 @@ WITH base AS (
     normalized_channel,
     normalized_os,
     normalized_os_version,
+    {% if has_distribution_id %}
+    CAST(metrics.string.metrics_distribution_id AS STRING) AS distribution_id,
```
sean-rose, line 28: I don't think casting the actual column is necessary.
```suggestion
    metrics.string.metrics_distribution_id AS distribution_id,
```

sean-rose (commented):
Should the `metrics_distribution_id` field you added to the `org_mozilla_ios_firefox_stable.baseline_v1.schema.json` test files in #5611 be removed, since the field is only present in Fenix?

sql_generators/glean_usage/templates/baseline_clients_daily_v1.query.sql
kik-kik: I wonder if we should keep it to make sure when the schema file is generated it is generated with correct type for all apps, WDYT @sean-rose ?

sean-rose: All string metric fields should be of type `STRING`, so casting such fields as `STRING` should never be necessary.

`CAST(NULL AS STRING)` in the `else` clause is necessary because BigQuery defaults to treating raw `NULL`s as the `INTEGER` type.

kik-kik: Got it, thanks for taking a moment to explain this 🙇 I made the change as requested.

PR #5619, opened 2024-05-20
DSRE-1626

the-chronomancer:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3837)


sql/moz-fx-data-shared-prod/firefox_desktop/newtab/metadata.yaml
```diff
@@ -0,0 +1,15 @@
+---
```
whd, line 1: You should reduce the explicit metadata to just `workgroup_access` e.g. https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/telemetry_derived/events_live/metadata.yaml

sql/moz-fx-data-shared-prod/firefox_desktop/newtab/metadata.yaml
```diff
@@ -0,0 +1,15 @@
...
+description: |-
+  A view of pings sent for the
+  `firefox-desktop/newtab`
+  document type.
+
+  Clustering fields: `submission_timestamp`
+workgroup_access:
+- role: roles/bigquery.dataViewer
+  members:
+  - workgroup:pocket/prefect
```
whd, line 13: There's currently a "bug" where table-level metadata should be redundant with dataset-level data but isn't, so you should add `workgroup:mozilla-confidential` explicitly to this list and the list in `newtab_live` (which should be undone later when we fix that bug). I will file a bqetl issue for this if one doesn't exist.

sql/moz-fx-data-shared-prod/firefox_desktop/newtab_live/metadata.yaml
```diff
@@ -12,3 +12,4 @@ workgroup_access:
   members:
   - workgroup:pocket/prefect
   - workgroup:mozsocial/prefect
+  - workgroup:mozsocial/service
```
whd, line 15:
```suggestion
  - workgroup:mozsoc-ml/service
```

whd (approved):

PR #5668, opened 2024-05-28
[DENG-3905] Support queries with temp udfs when billing project is set

BenWu:
fixes DENG-3905

temp udfs turn a query into a script when running in a session (i.e. `--billing-project` is set) and scripts can't have a destination table.  This creates the temp udfs in a separate query.

Also reverts the workaround in #5666 for readability.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3913)


sql_generators/glean_usage/templates/events_stream_v1.query.sql
BenWu, line 1: Changes in this file are a revert of https://github.com/mozilla/bigquery-etl/pull/5666

sean-rose:
> temp udfs turn a query into a script when running in a session

Is that a BigQuery bug?

bigquery_etl/cli/query.py
```diff
@@ -1043,6 +1051,37 @@ def create_query_session(
...
+
+    client = bigquery.Client(project=project_id)
+
+    job_config = bigquery.QueryJobConfig(
+        use_legacy_sql=False,
+        connection_properties=[bigquery.ConnectionProperty("session_id", session_id)],
+    )
+    client.query_and_wait(udf_def_statement, job_config=job_config)
+
+    return updated_query
```
sean-rose, line 1082: When a session is involved, would it perhaps be simpler to just split non-script queries into separate statements (if any), and run each of those statements separately, with only the final statement query having the destination table set?

BenWu:
> > temp udfs turn a query into a script when running in a session
> 
> Is that a BigQuery bug?

I thought so at first but I think it's intended because temp udfs are session-scoped https://cloud.google.com/bigquery/docs/sessions-write-queries#use_temporary_functions_in_sessions

bigquery_etl/cli/query.py
BenWu: Yes that's a good point.  I was initially accounting for non-udf statements that might be in the query especially since it can't be dry run but I'm not sure there are any other possible statement types

BenWu: I ran sqlparse on the output of `[f for f in paths_matching_name_pattern(None, "sql", None) if f.name != "script.sql"]` and confirmed the only other statements are udfs so it's safe to take out the temp udf specific code.  I'm thinking of still running those as one query though since it's easier to look through in the job logs.

sean-rose (approved):

quiiver:
a little late on this but is there a reason to keep these udfs as temporary vs adding them to our udf deployment?

BenWu:
Do you mean generally or specifically for the events stream udfs?  Generally, temp udfs do something specific to a single query so it doesn't make sense to share them.  I think that applies to the events stream udfs because I wrote them with the data types in those columns i mind.  I couldn't come up with one that works generally like `JSON.parse` which is why I wrote two.  I think it would be confusing to put them in the udf datasets because the error out if not being used in these specific cases

PR #5643, opened 2024-05-23
DENG-3869 Create schema before deploying backfill staging table

wwyc:
https://mozilla-hub.atlassian.net/browse/DENG-3869

When deploying the staging table, it will fail due to missing `schema.yaml` file.  This PR creates a schema file before deployment.

Checklist for reviewer:

- [ ] Commits should reference a bug or github issue, if relevant (if a bug is referenced, the pull request should include the bug number in the title).
- [ ] If the PR comes from a fork, trigger integration CI tests by running the [Push to upstream workflow](https://github.com/mozilla/bigquery-etl/actions/workflows/push-to-upstream.yml) and provide the `<username>:<branch>` of the fork as parameter. The parameter will also show up
in the logs of the `manual-trigger-required-for-fork` CI task together with more detailed instructions.
- [ ] If adding a new field to a query, ensure that the schema and dependent downstream schemas have been updated.
- [ ] When adding a new derived dataset, ensure that data is not available already (fully or partially) and recommend extending an existing dataset in favor of creating new ones. Data can be available in the [bigquery-etl repository](https://github.com/mozilla/bigquery-etl), [looker-hub](https://github.com/mozilla/looker-hub) or in [looker-spoke-default](https://github.com/mozilla/looker-spoke-default/tree/e1315853507fc1ac6e78d252d53dc8df5f5f322b).

For modifications to schemas in restricted namespaces (see [`CODEOWNERS`](https://github.com/mozilla/bigquery-etl/blob/main/CODEOWNERS)):
- [ ] Follow the [change control procedure](https://docs.google.com/document/d/1TTJi4ht7NuzX6BPG_KTr6omaZg70cEpxe9xlpfnHj9k/edit#heading=h.ttegrcfy18ck)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3870)

PR #5674, opened 2024-05-29
chore: refactor schema deploys, add and use deploy utils

ANich:
Mainly introduces a `deploy` module for table deployment logic. Additionally does the following:

- Adds specific exceptions for skipped and failed deploys. 
-  Updates multiprocessing code to `concurrent.futures` to collect individual exceptions as exception handling instead of using return values. 
- Moves metadata attaching logic to metadata module. 
- Update backfill initiate tests to mock deployment. These tests currently pass because skipped deploys currently "fail" silently in the click deploy command. Running a backfill without a schema in production will skip the deploy and fail when running the backfill queries. An example of this is the conversation linked in https://github.com/mozilla/bigquery-etl/pull/5646. (Missing schemas should be fixed by https://github.com/mozilla/bigquery-etl/pull/5643).

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -2197,9 +2197,7 @@ def deploy(
         click.echo("The following tables could not be deployed:")
         for failed_deploy in failed_deploys:
             click.echo(failed_deploy)
-        sys.exit(
-            1
-        )  # TODO: It would be nice to raise an ExceptionGroup here with the FailedDeployExceptions
+        sys.exit(1)
 
 
 def _deploy_external_data(
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -1,5 +1,6 @@
 """bigquery-etl CLI query command."""
 
+import concurrent.futures
 import copy
 import datetime
 import json
@@ -45,6 +46,7 @@ from ..cli.utils import (
 )
 from ..config import ConfigLoader
 from ..dependency import get_dependency_graph
+from ..deploy import FailedDeployException, SkippedDeployException, deploy_table
 from ..dryrun import DryRun
 from ..format_sql.format import skip_format
 from ..format_sql.formatter import reformat
@@ -59,6 +61,7 @@ from ..metadata.parse_metadata import (
     PartitionMetadata,
     PartitionType,
 )
+from ..metadata.publish_metadata import attach_metadata
 from ..query_scheduling.dag_collection import DagCollection
 from ..query_scheduling.generate_airflow_dags import get_dags
 from ..schema import SCHEMA_FILE, Schema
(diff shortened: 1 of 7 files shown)


bigquery_etl/cli/query.py
```diff
@@ -2118,148 +2120,52 @@ def deploy(
...
+    if skipped_deploys:
+        click.echo("The following deploys were skipped:")
+        for skipped_deploy in skipped_deploys:
+            click.echo(skipped_deploy)
+
+    if failed_deploys:
         click.echo("The following tables could not be deployed:")
         for failed_deploy in failed_deploys:
             click.echo(failed_deploy)
         sys.exit(1)
```
ANich, line 2167: It'd be nice if this could collect and raise the various exceptions as an `ExceptionGroup`.

bigquery_etl/deploy.py
```diff
@@ -0,0 +1,120 @@
...
+    dataset_name = query_file.parent.parent.name
+    project_name = query_file.parent.parent.parent.name
+
+    if destination_table is None:
+        destination_table = f"{project_name}.{dataset_name}.{table_name}"
+
+    existing_schema_path = query_file.parent / SCHEMA_FILE
+    try:
+        existing_schema = Schema.from_schema_file(existing_schema_path)
+    except Exception as e:  # TODO: Raise/catch more specific exception
```
ANich, line 64: `Schema.from_schema_file` currently raises an `Exception`, it probably should be a `ValueError` or something similar I just didn't want to change that module much in this PR. Could do this in a follow-up.

bigquery_etl/deploy.py
```diff
@@ -0,0 +1,120 @@
...
+
+def _create_or_update(
+    client: bigquery.Client,
+    table: bigquery.Table,
+    destination_table: str,
+    skip_existing: bool = False,
+) -> None:
+    if table.created:
+        if skip_existing:
+            raise SkippedDeployException(f"{destination_table} already exists.")
```
ANich, line 104: I wanted to use `table.full_table_id` here, but this apparently doesn't get populated when you create a `google.cloud.bigquery.Table` object 🤔 Assigning it (similar to the schema/metadata) seems strange given we already provide a table reference str in the constructor. Curious for thoughts here if any.

kwindau (approved):

bigquery_etl/deploy.py
scholtzan: I believe `table.table_id` and `table.dataset_id` should be available.

The [docs](https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.table.Table#google_cloud_bigquery_table_Table_full_table_id) mention: _None until set from the server_

scholtzan (approved):

bigquery_etl/deploy.py
BenWu: `str(table)` actually gives the full id as well https://github.com/googleapis/python-bigquery/blob/f19d39801cb5828929f1117ab93492aefdada5fd/google/cloud/bigquery/table.py#L1092-L1093

ANich: Ahh good find! I'll use the string representation

quiiver:
thank you for the tests 🥳

PR #5700, opened 2024-05-31
fix(deploy): skip (instead of fail) deploys with explicitly null destination_table

ANich:
Previous logic was a skip rather than a fail. This isn't mis-configuration, these are used for empty check queries, e.g. `site_metrics_empty_check_v1`. Medium-long term these could probably be migrated to a data check.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-3951)


BenWu (approved):

BenWu:
I didn't see that this was what it was already doing before

PR #5750, opened 2024-06-05
Have most production Stripe ETLs run hourly (DENG-3736)

sean-rose:
## [DENG-3736](https://mozilla-hub.atlassian.net/browse/DENG-3736): Increase frequency for product/subplat data updates

Fresher subscriptions data is needed for Braze (cc @chelseybeck).

---
Checklist for reviewer:

[DENG-3736]: https://mozilla-hub.atlassian.net/browse/DENG-3736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4023)

### DIFF
--- a/dags.yaml
+++ b/dags.yaml
@@ -168,6 +168,19 @@ bqetl_subplat:
   tags:
     - impact/tier_1
 
+bqetl_subplat_hourly:
+  schedule_interval: 0 * * * *
+  description: |
+    Hourly imports for Subscription Platform data from Stripe as well as derived tables based on that data.
+  default_args:
+    owner: srose@mozilla.com
+    start_date: "2022-03-24"
+    email: ["telemetry-alerts@mozilla.com", "srose@mozilla.com"]
+    retries: 2
+    retry_delay: 5m
+  tags:
+    - impact/tier_3
+
 bqetl_mozilla_vpn_site_metrics:
   schedule_interval: 0 15 * * *
   description: |
@@ -770,19 +783,6 @@ bqetl_regrets_reporter_summary:
   schedule_interval: 0 4 * * *
   tags:
     - impact/tier_1
-
-bqetl_cjms_nonprod:
-  schedule_interval: 0 * * * *
-  description: |
-    Hourly ETL for cjms nonprod.
-  default_args:
-    owner: srose@mozilla.com
-    start_date: "2022-03-24"
-    email: ["telemetry-alerts@mozilla.com", "srose@mozilla.com"]
-    retries: 2
-    retry_delay: 5m
-  tags:
-    - impact/tier_3
 
 bqetl_acoustic_contact_export:
   schedule_interval: 0 9 * * *
--- a/dags.yaml
+++ b/dags.yaml
@@ -169,9 +169,13 @@ bqetl_subplat:
     - impact/tier_1
 
 bqetl_subplat_hourly:
-  schedule_interval: 0 * * * *
+  schedule_interval: 30 * * * *
   description: |
     Hourly imports for Subscription Platform data from Stripe as well as derived tables based on that data.
+
+    This DAG is scheduled with consideration for its ETLs that write to date-partitioned tables,
+    so the last run targeting a particular date partition will happen at 00:30 the following day,
+    allowing time for all of the data for that date to be ready from external systems like Stripe.
   default_args:
     owner: srose@mozilla.com
     start_date: "2022-03-24"
@@ -179,7 +183,7 @@ bqetl_subplat_hourly:
     retries: 2
     retry_delay: 5m
   tags:
-    - impact/tier_3
+    - impact/tier_1
 
 bqetl_mozilla_vpn_site_metrics:
   schedule_interval: 0 15 * * *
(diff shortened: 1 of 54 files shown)


sql/moz-fx-cjms-prod-f3c7/cjms_bigquery/refunds_v1/metadata.yaml
```diff
@@ -11,3 +11,15 @@ scheduling:
...
+  depends_on:
+    - dag_name: bqetl_subplat_hourly
+      task_id: stripe_external__charge__v1
+      execution_delta: -21h45m
+    - dag_name: bqetl_subplat_hourly
+      task_id: stripe_external__invoice__v1
+      execution_delta: -21h45m
+    - dag_name: bqetl_subplat_hourly
+      task_id: stripe_external__refund__v1
+      execution_delta: -21h45m
```
sean-rose, line 25: It's annoying this is necessary. The execution deltas generated by default (1h15m) are sensible from the perspective of the run date (i.e. data interval end), but it turns out Airflow actually applies the execution deltas to the execution date (i.e. data interval start).

I think a better ultimate solution would be leveraging `ExternalTaskSensor`'s `execution_date_fn` parameter to dynamically calculate the dependency's execution date to wait for based on actual cron logic and configurable criteria (e.g. most recent run, first/last run in same date), but I didn't want to fall down that rabbit hole in this PR.

dags.yaml
```diff
@@ -168,6 +168,23 @@ bqetl_subplat:
   tags:
     - impact/tier_1
 
+bqetl_subplat_hourly:
+  schedule_interval: 30 * * * *
```
BenWu, line 172: Is it possible to make this 45 so that it's aligned with `bqetl_subplat`?  Just a nit but it would make the negative execution delta easier to reason about

sql/moz-fx-cjms-prod-f3c7/cjms_bigquery/refunds_v1/metadata.yaml
BenWu: It might be possible to use timetables to mess with the interval https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/timetable.html#crontriggertimetable

e.g. run hourly but with a 24 hour data interval.  Execution delta works with the `logical_date` property though and I'm not sure how that's affected relative to the interval.  It feels a bit hacky though and any other dependent dags would need to account for it.

Current way is fine though and I don't want to over-engineer it

BenWu (approved):

dags.yaml
sean-rose: It's possible, but here's my reasoning for picking the schedule I did:
- With some of the hourly ETLs writing to date-partitioned tables, in between the first and second run of each day the data lag for such tables will be greater than one hour.  With the current schedule that lag should be at most ~1.5 hours (half an hour more than usual), while if we used `45 * * * *` that max lag would be longer (~1.75 hours, or 45 minutes more than usual).
  - @chelseybeck how much would the max data lag matter for Braze?
- By starting the hourly ETLs 15 minutes before the daily ETLs that select from some of those hourly ETL tables, we allow the hourly ETLs time to potentially complete before the downstream daily ETLs start running.  With the current execution deltas the daily ETLs aren't depending on the most recent hourly DAG run, so they won't block on waiting for the possibly-still-running hourly ETLs as they're modifying the tables some daily ETLs are selecting from.  This isn't an absolute dealbreaker, but I think ideally we'd want the hourly ETLs to normally finish before the daily ETLs run to minimize confusion about what state those hourly tables were in when the daily ETLs were selecting from them.
  - Following this logic, if the hourly DAG runs end up taking appreciably longer than 15 minutes we may want to change this schedule to something like `15 * * * *` to give those more time to complete before the daily ETLs run.
  - Or alternatively, I could change the execution deltas for cases where a daily ETL depends on an hourly ETL that _doesn't_ write to a specific date-partition so it depends on the most recent run. However, that makes calculating/managing the time deltas a bit more complicated.  What do you think @BenWu?

sean-rose: And another angle to this is the execution deltas that will need to be calculated/managed for the `bqetl_braze` DAG tasks that will depend on `bqetl_subplat_hourly` tasks.  The [`bqetl_braze` DAG's schedule](https://github.com/mozilla/bigquery-etl/blob/80de43a57f198f58a5605dcafce5445c2dd442f0/dags.yaml#L1544) is currently `30 4,12,20 * * *`.

@chelseybeck what's your plan for how the `bqetl_braze` DAG will ultimately be scheduled?  Will it end up running hourly?

sql/moz-fx-cjms-prod-f3c7/cjms_bigquery/refunds_v1/metadata.yaml
sean-rose: Using timetables is an interesting idea. 🤔 

> I think a better ultimate solution would be leveraging `ExternalTaskSensor`'s `execution_date_fn` parameter to dynamically calculate the dependency's execution date

I've now realized this wouldn't be as great of a solution since `ExternalTaskMarker`s are a factor here as well (though IMO it seems like Airflow shouldn't need them), and they don't appear to have such a way to dynamically calculate the execution date/delta.

dags.yaml
chelseybeck: yes, our plan is to run it hourly as soon as we can get ctms data that frest :)

sean-rose:
This has been superseded by #7582.

Issue #3803, opened 2023-05-18
Schemas and yaml filenames in tests have to match dependencies exactly 


ANich:
For example in this [CI run](https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/20906/workflows/b09c0604-ca00-48b7-88d6-09b5417b1205/jobs/205028) the tests fail due to not being able to find and query the dependency tables. 

One dependency of the table is `moz-fx-data-shared-prod.firefox_desktop_stable.newtab_v1`. The test file must be exactly named `moz-fx-data-shared-prod.firefox_desktop_stable.newtab_v1.schema.json/.yaml` in order for the tests to detect them. 

Some ideas to solve this are either we make sure the names match in the testing infrastructure or we could only allow fully qualified names in queries (and tests I suppose)

SimardeepSingh1450:
@ANich can you please suggest some file-path/directory which i can look upto for this issue ?

ANich:
Hi Simardeep, most changes will likely be in the repository pytest plugin `bigquery_etl/pytest_plugin` and the sql test directory `tests/sql` generally. 

Fair warning though the tests and CI for this repository are pretty closely tied to internal gcp resources so it might be difficult to fork and test changes.

PR #5764, opened 2024-06-06
Fully qualified identifiers in SQL queries

scholtzan:
This implements https://docs.google.com/document/d/1S4e42e4tbhjJuRsKlhTsxzPAYk7QNefKC7Hq1OJdDxA/edit

`bqetl format` will now also fully-qualify table and UDF references and check for them in the CI.

Next steps:
- remove datasets from `skip_format` that should be formatted
- apply this to private-bigquery-etl 

`dry-run` is failing due to `metrics_clients_last_seen_v1` referencing themselves and having to explicit schema. So stage cannot deploy the table which is referenced in the dry run. This should be fixed in a follow-up PR

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4039)


.circleci/workflows.yml
```diff
@@ -564,6 +564,7 @@ jobs:
                   PATH="venv/bin:$PATH" script/bqetl metadata update \
                     --sql-dir /tmp/workspace/generated-sql/sql/ \
                     /tmp/workspace/generated-sql/sql/
+                  PATH="venv/bin:$PATH" script/bqetl format /tmp/workspace/generated-sql/sql/
```
scholtzan, line 567: Running the formatting as part of the SQL generation to ensure that identifiers are fully qualified when the content gets pushed to `generated-sql`.

scholtzan:
dry-run is failing due to `metrics_clients_last_seen_v1` referencing themselves and having no explicit schema. So stage cannot deploy the table which is referenced in the dry run. This should be fixed in a follow-up PR

bigquery_etl/format_sql/format.py
```diff
@@ -52,7 +54,13 @@ def format(paths, check=False):
         for path in sql_files:
             with open(path) as fp:
                 query = fp.read()
```
ANich, line 56: nit: since using a path object downstream anyway
```suggestion
            query_path = Path(path):
            query = query_path.read_text()
```
or `query_path.read_text()` on line 61

.circleci/workflows.yml
ANich: IIUC we could still have unqualified queries on `main`? Though they'll be qualified in `generated-sql`. Would be nice to have a validation check so we don't end up in that situation, maybe in a pre-commit?

bigquery_etl/util/common.py
```diff
@@ -162,6 +163,120 @@ def write_sql(output_dir, full_table_id, basename, sql, skip_existing=False):
...
+        **DEFAULT_QUERY_TEMPLATE_VARS,
+    )
+    init_query = render(
+        path.name,
+        template_folder=path.parent,
+        format=False,
+        **{
+            "is_init": lambda: True,
+            "metrics": MetricHub(),
+        },
```
ANich, line 203: nit: why not
```suggestion
       is_init=lambda: True,
       metrics=MetricHub(),
```

bigquery_etl/util/common.py
```diff
@@ -162,6 +163,120 @@ def write_sql(output_dir, full_table_id, basename, sql, skip_existing=False):
         f.write("\n")
 
 
+def qualify_table_references_in_path(path: Path) -> str:
```
ANich, line 166: Might I suggest a test for this, it's pretty long 😂

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_amplitude_export_v1/query.sql
```diff
@@ -114,22 +117,27 @@ _previous AS (
...
     ) AS days_seen_bits,
-    udf.combine_days_seen_maps(
+    `moz-fx-data-shared-prod.udf.combine_days_seen_maps`(
       _previous.os_used_month,
-      ARRAY(SELECT STRUCT(key, CAST(TRUE AS INT64) AS value) FROM _current.os_used_month AS key)
+      ARRAY(
+        SELECT
+          STRUCT(key, CAST(TRUE AS INT64) AS value)
+        FROM
+          `moz-fx-data-shared-prod._current.os_used_month` AS key
```
ANich, line 139: Not sure what's happening here. Looks like `_current` is an alias for a CTE but it's being qualified

sql/mozfun/bits28/examples/extract_week_specific_bits.sql
```diff
@@ -3,6 +3,6 @@ SELECT
   mozfun.bits28.range(days_seen_bits, -13 + 0, 7) AS week_0_bits,
   mozfun.bits28.range(days_seen_bits, -13 + 7, 7) AS week_1_bits
 FROM
-  telemetry.clients_last_seen
+  `mozfun.telemetry.clients_last_seen`
```
ANich, line 6: Looks like things in `mozfun/*/examples` should probably be manually qualified? I suspect this was meant to be 
```suggestion
  `mozdata.telemetry.clients_last_seen`
```
See far above comment on validation for manually written queries

ANich (commented):
Left a few comments, a couple manual fixes, one suggestion. The comment on `_current` aliased CTE might need a manual fix but also might be a bug in how that query is being parsed

bigquery_etl/util/common.py
BenWu: counterpoint: why not
```suggestion
        *list((path.name,)),
        **{
            "template_folder": path.parent,
            "format": False,
            "is_init": lambda: True,
            "metrics": MetricHub(),
        },
```

.circleci/workflows.yml
BenWu: My understanding is that we need to run this after sql generation because some of the generated sql doesn't have fully qualified identifiers.  Everything in main should be caught by `verify-format-sql`

bigquery_etl/util/common.py
ANich: Very nit, but passing in a dict literal with `**{...}` seems strange. Above `**DEFAULT_QUERY_TEMPLATE_VARS` makes sense to me because it's a constant (a variable would also make sense), but for a literal, I can't see how it's better than just named args

bigquery_etl/format_sql/format.py
```diff
@@ -40,7 +42,7 @@ def format(paths, check=False):
                     # skip tests/**/input.sql
                     and not (path.startswith("tests") and filename == "input.sql")
                     for filepath in [os.path.join(dirpath, filename)]
-                    if filepath not in skip_format()
+                    if not any([filepath.endswith(s) for s in skip_format()])
```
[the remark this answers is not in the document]
scholtzan: this is just a general fix. when running formatting in CI we usually write files to some temporary paths that don't exactly match the ones configured in the skip list

sql/mozfun/bits28/examples/extract_week_specific_bits.sql
scholtzan: added UDF examples to the format skip list

sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_amplitude_export_v1/query.sql
scholtzan: I found a way to extract table aliases and ignore references related to these aliases, similar to what was done with CTEs already

bigquery_etl/util/common.py
```diff
@@ -162,6 +163,130 @@ def write_sql(output_dir, full_table_id, basename, sql, skip_existing=False):
...
+            }
+
+            for table_expr in statement.find_all(sqlglot.exp.Table):
+                # existing table ref including backticks without alias
+                table_expr.set("alias", "")
+                reference_string = table_expr.sql(dialect="bigquery")
+
+                matched_cte = [
+                    re.match(
+                        rf"^{cte}(?![a-zA-Z0-9_])",
```
BenWu, line 235: Does a check for an optional `.` cover the same cases?  This would be a little clearer about what it's doing.  Also somewhat more robust since table names can have hyphens and accents even though we don't use them and the other regex in this function doesn't account for them
```suggestion
                        rf"^{cte}\.?",
```

bqetl_project.yaml
```diff
@@ -193,12 +193,9 @@ dry_run:
...
   - bigquery_etl/glam/templates/*.sql
-  - sql_generators/events_daily/templates/**/*.sql
-  - sql_generators/glean_usage/templates/*.sql
-  - sql_generators/search/templates/*.sql
-  - sql_generators/experiment_monitoring/templates/**/*.sql
-  - sql_generators/feature_usage/templates/*.sql
-  - sql_generators/funnels/templates/*.sql
+  - sql_generators/**/*.sql
+  - tests/**/*.sql
+  - sql/mozfun/**/examples/*.sql
```
BenWu, line 198: I think there's still some value in formatting some of these.  Could we have a instead have a separate category like `skip_qualifying_references`that gets used in `format.py`?

sql/mozfun/norm/examples/fenix_app_info.sql
```diff
@@ -12,12 +12,12 @@ WITH base AS (
   SELECT
     extract_fields('org_mozilla_firefox_beta', m).*
   FROM
-    org_mozilla_firefox_beta.metrics AS m
+    `mozfun.org_mozilla_firefox_beta.metrics` AS m
```
BenWu, line 15:
```suggestion
    `mozdata.org_mozilla_firefox_beta.metrics` AS m
```

sql/mozfun/norm/examples/fenix_app_info.sql
```diff
@@ -12,12 +12,12 @@ WITH base AS (
...
   UNION ALL
   SELECT
     extract_fields('org_mozilla_fenix', m).*
   FROM
-    org_mozilla_fenix.metrics AS m
+    `mozfun.org_mozilla_fenix.metrics` AS m
```
BenWu, line 20:
```suggestion
    `mozdata.org_mozilla_fenix.metrics` AS m
```

BenWu (commented):
Some small things but LGTM

bigquery_etl/util/common.py
scholtzan: With the optional `.` there might be cases where the CTE is a substring of some table name, which would then match incorrectly

BenWu: Oh right it would need to be something like `^{cte}(\.?|$)`.  It's fine as-is though

bqetl_project.yaml
scholtzan: I added a config option for this and added the udf examples folders

bigquery_etl/util/common.py
scholtzan: I think I'll leave it as is for now.

sql/moz-fx-data-shared-prod/search_terms_derived/adm_daily_aggregates_v1/query.sql
```diff
@@ -5,7 +5,7 @@ SELECT
   COUNT(*) AS impressions,
   COUNTIF(is_clicked) AS clicks,
 FROM
-  search_terms_derived.suggest_impression_sanitized_v2
+  `moz-fx-data-shared-prod.search_terms_derived.suggest_impression_sanitized_v2`
```
scholtzan, line 8: @whd there are a few `search_terms` queries which are getting fully-qualified in this PR that need your approval. (Or we can add them to a skip list if for some reason they shouldn't be fully-qualified)

BenWu (approved):

sql/moz-fx-data-shared-prod/contextual_services_derived/suggest_revenue_levers_daily_v1/query.sql
```diff
@@ -27,7 +27,7 @@ WITH client_shares AS (
       END
     ) AS eligible_clients
   FROM
-    telemetry.unified_metrics
+    `moz-fx-data-shared-prod.telemetry.unified_metrics`
```
scholtzan, line 30: @alekhyamoz  There are some `contextual_services` revenue datasets which got fully qualified in this PR and need your approval. (Or we can put them on a skip list if references should for some reason not be fully-qualified)

alekhyamoz (approved):

whd (approved):
Changes to restricted namespaces in this PR are not subject to change control

ANich (approved):
LGTM, one question, one thought:
1. I doubt it but do any tests need to be updated? RE https://github.com/mozilla/bigquery-etl/issues/3803
2. I would probably suggest something similar to dbt's `sources` as a medium-long term approach

scholtzan:
I updated test files as part of this PR

PR #5771, opened 2024-06-07
Print list of failed queries at end of cli dry run

BenWu:
Prints sql files that failed.  Printing the error messages would be better but needs more work

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4050)

### DIFF
--- a/bigquery_etl/cli/dryrun.py
+++ b/bigquery_etl/cli/dryrun.py
@@ -141,5 +141,11 @@ def _sql_file_valid(
         respect_skip=respect_skip,
     )
     if validate_schemas:
-        return result.validate_schema(), sqlfile
+        try:
+            success = result.validate_schema()
+        except Exception as e:  # validate_schema raises base exception
+            click.echo(e, err=True)
+            success = False
+        return success, sqlfile
+
     return result.is_valid(), sqlfile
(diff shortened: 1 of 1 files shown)


sean-rose (approved):

BenWu:
I just noticed the `validate_schema` path only catches some exceptions so the script exits when the pool map returns.  Added a catch and print for exception there

bigquery_etl/cli/dryrun.py
```diff
@@ -134,5 +141,11 @@ def _sql_file_valid(
         respect_skip=respect_skip,
     )
     if validate_schemas:
-        return result.validate_schema()
-    return result.is_valid()
+        try:
+            success = result.validate_schema()
+        except Exception as e:  # validate_schema raises base exception
+            click.echo(e, err=True)
```
sean-rose, line 147: I'd suggest adding an error message (including the file in question) with the exception at the end.  Otherwise it might be hard to tell what the exception is for in the deluge of parallel-processed dry-runs.

BenWu: The exceptions here always have the file name like `sql/moz-fx-data-shared-prod/search_derived/acer_cohort_v1/query.sql ERROR` so I didn't add it but I guess that can change

sean-rose: OK, never mind then. 👍

PR #5798, opened 2024-06-17
Skip apps with no events in monitoring.event_monitoring_live

BenWu:
Should fix view deploys by not including `gleanjs_docs` in the `event_monitoring_live` view but `gleanjs_docs` shouldn't actually be skipped and the issue described in https://github.com/mozilla/bigquery-etl/issues/5797 still exists.

This change is to match the app filtering that's used in generating the `event_monitoring_live_v1` materialized view.  Currently, `monitoring.event_monitoring_live` unions views that weren't generated.

Checklist for reviewer:

- [ ] Commits should reference a bug or github issue, if relevant (if a bug is referenced, the pull request should include the bug number in the title).
- [ ] If the PR comes from a fork, trigger integration CI tests by running the [Push to upstream workflow](https://github.com/mozilla/bigquery-etl/actions/workflows/push-to-upstream.yml) and provide the `<username>:<branch>` of the fork as parameter. The parameter will also show up
in the logs of the `manual-trigger-required-for-fork` CI task together with more detailed instructions.
- [ ] If adding a new field to a query, ensure that the schema and dependent downstream schemas have been updated.
- [ ] When adding a new derived dataset, ensure that data is not available already (fully or partially) and recommend extending an existing dataset in favor of creating new ones. Data can be available in the [bigquery-etl repository](https://github.com/mozilla/bigquery-etl), [looker-hub](https://github.com/mozilla/looker-hub) or in [looker-spoke-default](https://github.com/mozilla/looker-spoke-default/tree/e1315853507fc1ac6e78d252d53dc8df5f5f322b).

For modifications to schemas in restricted namespaces (see [`CODEOWNERS`](https://github.com/mozilla/bigquery-etl/blob/main/CODEOWNERS)):
- [ ] Follow the [change control procedure](https://docs.google.com/document/d/1TTJi4ht7NuzX6BPG_KTr6omaZg70cEpxe9xlpfnHj9k/edit#heading=h.ttegrcfy18ck)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4096)

PR #5799, opened 2024-06-18
Generate event monitoring view for all apps with events ping

### LINKED ISSUE #5797: event_monitoring_live generation is skipping apps that only have events defined in dependencies
Because of https://github.com/mozilla/bigquery-etl/pull/5576, `event_monitoring_live` generation looks at metrics definitions to find apps that have events defined.  However, the probe-scraper output doesn't include events from dependencies, so this is causing events monitoring to skip apps that only have events in dependencies. 

https://github.com/mozilla/bigquery-etl/blob/979fb253cd802108efb4fc9e99450408048b7db4/sql_generators/glean_usage/event_monitoring_live.py#L42-L53

gleanjs-docs was recently added which has events defined in [glean-js metrics](https://probeinfo.telemetry.mozilla.org/glean/glean-js/metrics) but not in [its own metrics](https://probeinfo.telemetry.mozilla.org/glean/gleanjs-docs/metrics).  This is causing `event_monitoring_live` to not generate the `gleanjs_docs_derived.event_monitoring_live_v1` but the `monitoring.event_monitoring_live` view does include it.  This is breaking view deploys.

View deploys can be fixed by making `generate_across_apps` match the filtering in `generate_per_app_id` in https://github.com/mozilla/bigquery-etl/blob/979fb253cd802108efb4fc9e99450408048b7db4/sql_generators/glean_usage/event_monitoring_live.py but the `_get_tables_with_events` still needs to get fixed so `gleanjs_docs` and future apps can have `event_monitoring_live_v1`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4095)

BenWu:
fixes #5797

This recreates the materialized views that were removed from the generated sql in https://github.com/mozilla/bigquery-etl/pull/5576.  Looking at the generated sql diff, `monitoring/event_monitoring_live/view.sql` wasn't changed so I think that means the nothing new gets added or removed as expected (other than `gleanjs_docs_derived.event_monitoring_live_v1`).

@scholtzan One thing I'm not sure about is that this adds `events_v1` to the union in some of the materialized views.  Did we intentionally want to exclude those from the views for apps not listed in the bqetl config?

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4098)


sql_generators/glean_usage/event_monitoring_live.py
```diff
@@ -50,6 +60,9 @@ def _get_tables_with_events(self, v1_name: str) -> Set[str]:
                 latest_history = metric.get("history", [])[-1]
                 pings.update(latest_history.get("send_in_pings", []))
 
+        if bq_dataset_name in self._get_prod_datasets_with_event():
+            pings.add("events")
+
```
scholtzan, line 65:
```suggestion

```

The materialized views only need to monitor tables that have event-type probes as defined in probe-info. Some apps don't seem to have such probes in the `events` tables, so we shouldn't add them here since not monitoring them reduces cost a bit.

sql_generators/glean_usage/event_monitoring_live.py
```diff
@@ -39,7 +39,17 @@ def __init__(self) -> None:
...
 
-    def _get_tables_with_events(self, v1_name: str) -> Set[str]:
+    def _get_prod_datasets_with_event(self) -> List[str]:
+        """Get glean datasets with an events table in generated schemas."""
+        return [
+            s.bq_dataset_family
+            for s in get_stable_table_schemas()
+            if s.schema_id == "moz://mozilla.org/schemas/glean/ping/1"
+            and s.bq_table == "events_v1"
+        ]
```
scholtzan, line 49: This was a bit of a hacky way to get all relevant datasets, since (almost?) all of them do have `event` pings.

sql_generators/glean_usage/event_monitoring_live.py
BenWu: Ok so do we only care about custom events defined in the app and not in the libraries?  In that case, https://github.com/mozilla/bigquery-etl/pull/5798 would be at least partially correct and a bunch of materialized views would need to be deleted to get what's deployed consistent with the generated sql.  I didn't think it made sense to suddenly delete them.

If we do care about events in libraries then I think they all will get at least `glean.restarted` in the events ping because of glean-core.  If it's more complicated than that then I can leave this for now since nothing is broken at the moment.

scholtzan: Yeah, good point. I guess adding `events_v1` to the monitoring shouldn't cause any issues of stuff breaking.

scholtzan (approved):

BenWu:
I filed https://github.com/mozilla/bigquery-etl/issues/5804 to track deployed materialized views not getting updated.  If I'm looking at it correctly, merging this shouldn't actually change any of the deployed views until that's fixed.

I also just sorted the events lists that get passed into the templates so the diffs should be more stable

PR #5800, opened 2024-06-18
Try to fully specify `metrics` fields to select in Glean union views

sean-rose:
To try to avoid problems when the underlying table/view schemas change due to new metrics being added.

---
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4100)


scholtzan (approved):

sean-rose:
I was worried doing this might cause BigQuery to always scan all `metrics` columns since they're being restructured in the view query, but in a minimal test dry-running queries to count the number of rows with a particular boolean metric set to true BigQuery reported the same number of bytes scanned both with and without this change.  So it appears BigQuery is successfully able to optimize through the restructuring logic and only scan the data that's needed (at least with simple `STRUCT()` calls like in this case).

sean-rose:
One thing we'll need to watch out for is this can make the view SQL significantly longer.  For example, with this change [sql/moz-fx-data-shared-prod/fenix/metrics/view.sql](https://github.com/mozilla/bigquery-etl/blob/generated-sql/sql/moz-fx-data-shared-prod/fenix/metrics/view.sql) would increase from ~99k characters to ~225k characters, which would be getting close to BigQuery's [256k character limit for view SQL](https://cloud.google.com/bigquery/quotas#view_limits) (I'm going to ask Google whether that limit can be increased).

BenWu:
You may have already figured this out but `use_counters` is even bigger than the metrics view (783k characters).  whd pointed out that we should be able to get jenkins to trigger artifact deployment soon so this might no longer be an issue soon, although it would still be a bit of ETL quirkiness https://bugzilla.mozilla.org/show_bug.cgi?id=1883727#c3

sean-rose:
> You may have already figured this out but `use_counters` is even bigger than the metrics view (783k characters). whd pointed out that we should be able to get jenkins to trigger artifact deployment soon so this might no longer be an issue soon, although it would still be a bit of ETL quirkiness https://bugzilla.mozilla.org/show_bug.cgi?id=1883727#c3

Yeah, with this change `fenix.use_counters` ends up being 783k characters and `focus_android.use_counters` ends up being 469k characters. I missed those initially because I was only running the `glean_usage` generator locally without running the `stable_views` generator first.

I'm going to add a workaround for that, then ping you and @scholtzan for re-review, and y'all can weigh in on whether it's worth merging this, or if it would be better to just wait for the Jenkins-triggered artifact deployment.

BenWu (commented):
This makes sense but I would lean towards fixing the schema deploy timing issue rather than adding more code here, especially since this is just a best effort solution which doesn't work for every view.  My opinion is that effort is better spent pushing [bug 1883727](https://bugzilla.mozilla.org/show_bug.cgi?id=1883727) forward rather than patching these views up but I'm not opposed to these changes

scholtzan (approved):

PR #5857, opened 2024-06-27
Exclude `monitor_backend` Glean app from baseline ping checks

sean-rose:
The `monitor_backend` app won't have any baseline pings.

---
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4180)

### DIFF
--- a/sql_generators/glean_usage/common.py
+++ b/sql_generators/glean_usage/common.py
@@ -31,6 +31,7 @@ NO_BASELINE_PING_APPS = (
     "firefox_reality_pc",
     "lockwise_android",
     "mach",
+    "monitor_backend",
     "monitor_cirrus",
     "moso_mastodon_backend",
     "mozphab",


BenWu (approved):

BenWu:
If `monitor_backend` is only dependent on `glean-server` then it won't have a baseline table created anyway and this isn't strictly necessary, right?  Not that I don't think this change should be made, just want to make sure I understand the server apps

sean-rose:
> If `monitor_backend` is only dependent on `glean-server` then it won't have a baseline table created anyway and this isn't strictly necessary, right? Not that I don't think this change should be made, just want to make sure I understand the server apps

Oh, you're right!  I did this based on having noticed `accounts_backend` configured this way when I was adding a Cirrus app to `NO_BASELINE_PING_APPS`, but it looks like only the earliest backend apps like Accounts and MozSocial got stuck with unused baseline ping tables, and newer backend apps don't have such baseline ping tables.

So this change was unnecessary and could be reverted, but I'll probably just leave it as is since it isn't hurting anything.

PR #5873, opened 2024-07-01
DENG-4047 Don't wrap metrics in JSON in accounts events stream tables

akkomar:
Accounts telemetry makes heavy use of ping-level metrics with events. We would like to have an explore in Looker that would allow to use these metrics.
Events stream is a nice model to explore events because of clustering per event category, but it would be easier to work with in both Looker and BQ console if metrics weren't wrapped as JSON.
This PR makes metrics a nested record in `events_stream` for `accounts_frontend` and `accounts_backend`. This will require manual deletion of these two tables before next or of these queries.

As next steps I propose to:
1. After I confirm that Looker explores affected by this change work correctly, switch `events_stream` to store metrics as struct by default for newly created tables.
2. Switch existing tables to the new approach, taking into account backfills and cross-channel views.

I'd like to do this in stages since step 2. will require some caution and a) I am open to comments if this should be applied by default to all tables (IMO it makes sense), b) we're under a little time pressure with some Accounts dashboarding work that this PR unblocks.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4222)


sql_generators/glean_usage/templates/events_stream_v1.query.sql
```diff
@@ -91,8 +91,11 @@ WITH base AS (
         ping_info.end_time,
         ping_info.parsed_end_time,
         ping_info.ping_type
-      ) AS ping_info,
+      ) AS ping_info
+      {% if not metrics_as_struct %}
+      ,
```
quiiver, line 96: bigquery sql is fine with trailing commas https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#trailing_commas

akkomar: Yeah, but in the SELECT statement. This is inside of REPLACE (...) and is not allowed (and painful to read).

quiiver (approved):
i don't love special casing specific events stream tables. if this works well for your explores lets consider moving all events stream tables to structured metrics

quiiver:
oh i just re-read the pr description and saw this
>    Switch existing tables to the new approach, taking into account backfills and cross-channel views.

PR #5892, opened 2024-07-08
fix(dag_gen): depends on external task group

edugfilho:
According to the [documentation](https://airflow.apache.org/docs/apache-airflow/2.8.2/howto/operator/external_task_sensor.html#externaltasksensor-with-task-group-dependency), for tasks belonging to a `task_group` we need to use `external_task_group_id` instead.
I checked the repo and found no dags using this feature so far.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4266)

### DIFF
--- a/bigquery_etl/query_scheduling/templates/airflow_dag.j2
+++ b/bigquery_etl/query_scheduling/templates/airflow_dag.j2
@@ -75,7 +75,7 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
         task_id='wait_for_{{ dependency.task_id }}',
         external_dag_id='{{ dependency.dag_name }}',
         {% if dependency.task_group -%}
-        external_task_group_id='{{ dependency.task_group }}'
+        external_task_group_id='{{ dependency.task_group }}',
         {% endif -%}
         external_task_id='{{ dependency.task_id }}',
         {% if dependency.get_execution_delta(schedule_interval) -%}
--- a/bigquery_etl/query_scheduling/templates/airflow_dag.j2
+++ b/bigquery_etl/query_scheduling/templates/airflow_dag.j2
@@ -75,10 +75,9 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
         task_id='wait_for_{{ dependency.task_id }}',
         external_dag_id='{{ dependency.dag_name }}',
         {% if dependency.task_group -%}
-        external_task_id='{{dependency.task_group}}.{{ dependency.task_id }}',
-        {% else -%}
-        external_task_id='{{ dependency.task_id }}',
+        external_task_group_id='{{ dependency.task_group }}'
         {% endif -%}
+        external_task_id='{{ dependency.task_id }}',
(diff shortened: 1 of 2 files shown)


bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -75,10 +75,9 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
         task_id='wait_for_{{ dependency.task_id }}',
         external_dag_id='{{ dependency.dag_name }}',
         {% if dependency.task_group -%}
-        external_task_id='{{dependency.task_group}}.{{ dependency.task_id }}',
-        {% else -%}
-        external_task_id='{{ dependency.task_id }}',
+        external_task_group_id='{{ dependency.task_group }}'
```
scholtzan, line 78:
```suggestion
        external_task_group_id='{{ dependency.task_group }}',
```

bigquery_etl/query_scheduling/templates/public_data_json_airflow_dag.j2
```diff
@@ -74,10 +74,9 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
         task_id='wait_for_{{ dependency.task_id }}',
         external_dag_id='{{ dependency.dag_name }}',
         {% if dependency.task_group -%}
-        external_task_id='{{dependency.task_group}}.{{ dependency.task_id }}',
-        {% else -%}
-        external_task_id='{{ dependency.task_id }}',
+        external_task_group_id='{{ dependency.task_group }}'
```
scholtzan, line 77:
```suggestion
        external_task_group_id='{{ dependency.task_group }}',
```

scholtzan (commented):
mh, there are some tasks which depend on a task within a task group. This change would make then depend on the task group instead, which is not desirable...

Do you have a PR somewhere with a task that depends on a task group?

scholtzan (commented):
ah wait, this should be fine. it still has the `external_task_id` defined. I guess I'll wait to see the generated DAGs to see what this will change

edugfilho:
> mh, there are some tasks which depend on a task within a task group. This change would make then depend on the task group instead, which is not desirable...
> 
> Do you have a PR somewhere with a task that depends on a task group?

Yea, my understanding is that they would depend on the task inside the group. But let me check more on the docs. Maybe I don't need this PR at all, because the task names are not conflicting with anything: https://github.com/mozilla/bigquery-etl/pull/5893

edugfilho:
Looks like we can't set `external_task_group_id` and `external_task_id` at the same time:
https://airflow.apache.org/docs/apache-airflow/stable/_modules/airflow/sensors/external_task.html
Search for ```Values for `external_task_group_id` and `external_task_id` can't be set at the same time.```

edugfilho:
Sorry, I assumed both values could be set at once. I guess the current `dot` works for what I want then, if I do https://github.com/mozilla/bigquery-etl/pull/5893.
I'll wait for confirmation to close this.

PR #5898, opened 2024-07-09
DSRE-1683: Remove use of Circle CI context, add GCP OIDC config

corbanCloud:
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4274)


.circleci/workflows.yml
```diff
@@ -732,7 +732,7 @@ jobs:
   artifact-deployment:
     executor: gcp-cli/google
     steps:
-      - run:
+      - run: &oidc-setup
```
whd, line 735: While this anchor makes sense, the final step https://github.com/mozilla/bigquery-etl/blob/main/.circleci/workflows.yml#L743 is going to have a different value for each distinct SA we're assuming. So we probably want to parameterize this in some fashion because https://app.circleci.com/settings/project/github/mozilla/bigquery-etl/environment-variables?return-to=https%3A%2F%2Fapp.circleci.com%2Fpipelines%2Fgithub%2Fmozilla%2Fbigquery-etl currently only has `GCP_SERVICE_ACCOUNT_EMAIL` set to the airflow DAG triggering SA. This workload uses the SA defined in https://github.com/mozilla-services/cloudops-infra/blob/master/projects/airflow/tf/prod/envs/prod/resources/main.tf#L151-L173

.circleci/workflows.yml
```diff
@@ -853,8 +853,12 @@ jobs:
           steps:
             - *skip
   deploy-to-private-gcr:
-    executor: ubuntu-machine-executor
+
+    executor: gcp-cli/google
     steps:
+      - run: *oidc-setup
+      - gcp-cli/setup:
```
whd, line 860: I don't think we need to add a `gcp-cli/setup` here, since https://circleci.com/developer/orbs/orb/circleci/gcp-gcr#commands-gcr-auth and https://circleci.com/developer/orbs/orb/circleci/gcp-gcr both support `use_oidc` directly. It's not immediately obvious to me which of the two should be updated (possibly both).

https://github.com/mozilla/bigquery-etl/blob/main/.circleci/workflows.yml#L868
https://github.com/mozilla/bigquery-etl/blob/main/.circleci/workflows.yml#L872

Side note: these orbs [might need updates](https://github.com/mozilla/bigquery-etl/blob/main/.circleci/workflows.yml#L5) to get the OIDC flags.

whd (commented):
There are a couple of ways to test this without just attempting to land it, but they will require adjusting settings in this branch and GCP-side settings. For example:

1. Change `main`  to `DSRE-1683` in https://github.com/mozilla-services/cloudops-infra/blob/master/projects/airflow/tf/prod/envs/prod/resources/main.tf#L152 and reapply terraform to allow OIDC-based access to the SA from a test branch
1. Switch the image name from `bigquery-etl` to e.g. `bigquery-etl-test` in https://github.com/mozilla/bigquery-etl/blob/main/.circleci/workflows.yml#L870 so the uploaded test image doesn't overwrite the actual prod image
1. Switch the branch [here](https://github.com/mozilla/bigquery-etl/blob/main/.circleci/workflows.yml#L1193) to `DSRE-1683` to allow the test image upload from this branch

.circleci/workflows.yml
corbanCloud: Ok so what I'd normally do here is just add a step to overwrite the ennvar afterwards
Because the yaml node here is just a set of strings, theres not a way to use the override `<<` operator either since we'd have to redefine the entire `command` node :|

corbanCloud: So I'd just handle it like this:
https://github.com/mozilla/bigquery-etl/pull/5898/commits/e3ee174e482c619c53df49f2529c3a098f36c48a#diff-7bf091450b804ea25dc3d11ac28c50202d19178f7b53f54a02ad59c946afa6a7R860

within the confines of yaml and bash I think just setting the new value over the top is clean enough? Unless we want to make a custom orb for this that accepts the SA as a parameter. I'm not opposed to that if theres other OIDC needs out here we can solve with it but the effort-to-outcome ratio seems a bit steep

.circleci/workflows.yml
corbanCloud: updated GCP CLI, GCR, and Docker (since its closely related to the GCR one) here https://github.com/mozilla/bigquery-etl/pull/5898/commits/46d48cbc7b17d6670a946928fa37a1b08a043958

.circleci/workflows.yml
corbanCloud: resolved in DMs

.circleci/workflows.yml
whd: My thought is that `gcp-cli/setup`  isn't necessary since you can just add `use_oidc` to one or both of the lines I linked above. I'm not super confident in that assertion but I think Mikael had used `gcp-cli/setup`  since we were executing arbitrary commands via `gcloud` whereas image building is done via standard upstream orbs.

PR #5902, opened 2024-07-10
Fix udf_js.gunzip to allow very long inputs

BenWu:
This function currently fails for long payloads.  This can be reproduced with:
```sql
SELECT `moz-fx-data-shared-prod.udf_js.gunzip`(CAST(REPEAT("a", 500000) AS BYTES))
```
which gives `RangeError: Maximum call stack size exceeded at UDF$1(BYTES) line 13, columns 35-36`

or more a more realistic example:
```sql
SELECT
  `moz-fx-data-shared-prod.udf_js.gunzip`(payload)
FROM
  `moz-fx-data-shared-prod.payload_bytes_error.telemetry`
WHERE
  TIMESTAMP_TRUNC(submission_timestamp, DAY) = TIMESTAMP("2024-07-08")
  AND document_type = 'crash'
  AND LENGTH(payload) > 40000
```

Performance seems to about the same in my limited testing

### DIFF
--- a/sql/moz-fx-data-shared-prod/udf_js/gunzip/udf.sql
+++ b/sql/moz-fx-data-shared-prod/udf_js/gunzip/udf.sql
@@ -22,7 +22,10 @@ AS
 
     function binary2String(byteArray) {
         // converts a UTF-16 byte array to a string
-        return String.fromCharCode.apply(String, byteArray);
+        return byteArray.reduce(
+          (accumulator, char_code) => accumulator + String.fromCharCode(char_code),
+          "",
+        )
     }
 
     // BYTES are base64 encoded by BQ, so this needs to be decoded
@@ -70,6 +73,7 @@ unzipped AS (
 )
   --
 SELECT
-  mozfun.assert.equals(expected, result)
+  mozfun.assert.equals(expected, result),
+  mozfun.assert.equals(udf_js.gunzip(CAST(REPEAT("a", 500000) AS BYTES)), REPEAT("a", 500000)),
 FROM
   unzipped


scholtzan (approved):

whd:
Hey I remember mentioning this in https://mozilla-hub.atlassian.net/browse/DSRE-1177?focusedCommentId=639948, good to see there's an easy fix

BenWu:
I was trying to use this with a different date range  and I ran into a `UDF out of memory` error while the existing udf works.  I'll hold off on merging this for now

BenWu:
Switched back to the `String.fromCharCode.apply` way but in chunks of up to 50000 chars at a time and then concatenating at the end.  That seems to work in all cases

sql/moz-fx-data-shared-prod/udf_js/gunzip/udf.sql
```diff
@@ -70,6 +75,7 @@ unzipped AS (
 )
   --
 SELECT
-  mozfun.assert.equals(expected, result)
+  mozfun.assert.equals(expected, result),
+  mozfun.assert.equals(udf_js.gunzip(CAST(REPEAT("abc", 200000) AS BYTES)), REPEAT("abc", 200000)),
```
akkomar, line 79: Can this test case be written the same way as the two other above?

akkomar (approved):
r+wc

sql/moz-fx-data-shared-prod/udf_js/gunzip/udf.sql
BenWu: Thanks for pointing that out.  I wasn't paying attention to the other tests

PR #5940, opened 2024-07-18
GROWTH-90 Fix Grain of New Table - can have multiple rows per client per search engine

kwindau:
Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4347)

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_desktop_derived/adclick_history_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_desktop_derived/adclick_history_v1/query.sql
@@ -2,21 +2,15 @@ WITH new_data AS (
   SELECT
     client_id,
     sample_id,
-    ad_click AS ad_clicks
+    sum(ad_click) AS ad_clicks
   FROM
     `moz-fx-data-shared-prod.search_derived.search_clients_daily_v8`
   WHERE
     submission_date = @submission_date
     AND ad_click > 0
-),
-existing_data AS (
-  SELECT
-    client_id,
-    sample_id,
-    ad_click_history
-  FROM
-    `moz-fx-data-shared-prod.firefox_desktop_derived.adclick_history_v1`
+    GROUP BY 1,2
 )
+    
 SELECT
   client_id,
   sample_id,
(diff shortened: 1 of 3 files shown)


sql/moz-fx-data-shared-prod/firefox_desktop_derived/adclick_history_v1/query.sql
```diff
@@ -2,20 +2,15 @@ WITH new_data AS (
...
-),
-existing_data AS (
-  SELECT
+  GROUP BY
     client_id,
-    sample_id,
-    ad_click_history
-  FROM
-    `moz-fx-data-shared-prod.firefox_desktop_derived.adclick_history_v1`
+    sample_id
```
lucia-vargas-a, line 13: Each client_id is related to only one sample_id, is it still required?

kwindau: I just wanted to include it so that people could query by sample ID if they wanted to limit results - but to your point, if I took it out of the schema/query completely, the query would probably run faster?

kwindau: OK I ended up taking it out since really it just adds work to the query and I don't really need it in this table, trying to make this run as efficiently as possible - thanks for the suggestion!

lucia-vargas-a (approved):

sql/moz-fx-data-shared-prod/firefox_desktop_derived/adclick_history_v1/query.sql
```diff
@@ -1,28 +1,20 @@
 WITH new_data AS (
   SELECT
     client_id,
-    sample_id,
```
irrationalagent, line 4: for my sake, any reason we're removing this?

kwindau: It makes it more efficient, trying to basically take out anything that isn't absolutely necessary to make the backfill run faster

kwindau: Basically, to clarify, if we leave it in, either I include it in the join with client ID (which slows stuff down since now it has to join on 2 things), or I do a max() or min() to only group by client ID, but still, that slows it down/adds extra work, when really we don't really need sample ID in this table, if that makes sense

PR #5962, opened 2024-07-23
internet_outage add new udf for geo with no null values

lelilia:
to be able to join data better on country, city, geo_subdivision1 and geo_subdivision2 it's actually nice to have a "??" value as a placeholder (that later can be changed back to null by the already existing `udf.geo_struct` udf

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4382)


Dexterp37 (approved):
This looks good to me! I'd wait for kik for a final r+

sql/moz-fx-data-shared-prod/udf/geo_struct_unknown/udf.sql
```diff
@@ -0,0 +1,41 @@
+/*
```
kik-kik, line 1: Instead of creating another UDF that looks very similar to what we already have, would it make sense to perhaps update:
https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/udf/geo_struct/udf.sql

to not return nulls but `??` instead if this is the preferred behaviour for derived tables?

lelilia: no because that udf is used other places and also i want to use that udf at the end to null it again

lelilia: also I had a look if I could do a flag but in BigQuery udfs you can't have parameters with default values ...

kik-kik: ok, my only comment then is. Could we come up with a different, more descriptive name? Maybe something like `geo_struct_normalize` or `geo_struct_set_defaults`? If you have other ideas, I'm open to it. I'm just not sure if geo_struct`_unknown` is a good name here.

kik-kik (approved):

PR #5964, opened 2024-07-24
feat(): add attribution_clients template to mobile_kpi_support_metrics sql generator

kik-kik:
# feat(): add attribution_clients template to mobile_kpi_support_metrics sql generator

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4385)


sql_generators/mobile_kpi_support_metrics/__init__.py
```diff
@@ -19,13 +19,15 @@
...
-    "engagement_clients.view.sql",
-    "engagement.query.sql",
-    "engagement.view.sql",
+    # "active_users.view.sql",
+    # "retention_clients.view.sql",
+    # "retention.query.sql",
+    # "retention.view.sql",
+    # "engagement_clients.view.sql",
+    # "engagement.query.sql",
+    # "engagement.view.sql",
```
bani, line 28: are you going to uncomment before merging?

sql_generators/mobile_kpi_support_metrics/templates/attribution_clients.schema.yaml
```diff
@@ -0,0 +1,29 @@
+{{ header }}
```
bani, line 1: I don't think the schema is supposed to have a query?

sql_generators/mobile_kpi_support_metrics/templates/attribution_clients.view.sql
```diff
@@ -0,0 +1,29 @@
+{{ header }}
+CREATE OR REPLACE VIEW
```
bani, line 2: Should we have submission_date on the view? In particular given it's the partition for the table

sql_generators/mobile_kpi_support_metrics/templates/attribution_clients.query.sql
```diff
@@ -0,0 +1,225 @@
+{{ header }}
+WITH new_profiles AS (
+  SELECT
+    submission_date,
+    client_id,
+    sample_id,
+  FROM `{{ project_id }}.{{ dataset }}.active_users`
```
bani, line 7: is there a reason why we are reading from active_users?
While the data is there, it doesn't seem we are leveraging anything specific to active_users in this query, and it's really hard to trace back the new profile logic from here.
Also, this is a table that has all clients active in the last 28 days, so it's unnecessarily large for this query

bani:
I think this is from the previous PR, but there was a UNIONed view for attribution_clients and I noticed two things in it:
1. it was accounting for `is_suspicious_device_client` but I don't see that in the source tables you created now
2. Query looked a bit weird first with SELECT * EXCEPT and then placing the fields back
3. The fields changed to STRUCT but you are still querying with the old format

sql_generators/mobile_kpi_support_metrics/__init__.py
kik-kik: yes, that's my bad. Thanks

sql_generators/mobile_kpi_support_metrics/templates/attribution_clients.schema.yaml
kik-kik: sorry, copy and paste mistake. Updated.

sql_generators/mobile_kpi_support_metrics/templates/attribution_clients.view.sql
kik-kik: I don't think it's strictly necessary here and partitioning would still be used correctly when getting the data from the underlying table if someone applied the filter on the view. For now I added it to reduce the confusion.

kik-kik:
> I think this is from the previous PR, but there was a UNIONed view for attribution_clients and I noticed two things in it:
> 
>     1. it was accounting for `is_suspicious_device_client` but I don't see that in the source tables you created now
> 
>     2. Query looked a bit weird first with SELECT * EXCEPT and then placing the fields back
> 
>     3. The fields changed to STRUCT but you are still querying with the old format

So regarding 1. this indeed appears to be missing in the generated attribution table. I will take a look tomorrow and get this resolved.

2. This is necessary in the union as different products have different fields which means these fields are also present in their schemas in a different order. I need to do this EXCEPT and add them to the SELECT in order to ensure in the view all of the product specific selects get the fields in the same order. Alternative approach here would be to not hav ethe SELECT * EXCEPT and have each field explicitly listed. @bani would that be a preferred approach?
3. 
4. The attribution fields are a struct in the new attribution table, but the view flattens these out. Would you prefer if we already did that in the query?

sql_generators/mobile_kpi_support_metrics/templates/attribution_clients.query.sql
kik-kik: In this case, I think we could also change this to the baseline_clients_first_seen view to act as the source.

bani:
I see. No problem with the current EXCEPT/SELECT and with the STRUCT flattened out.

sql_generators/mobile_kpi_support_metrics/templates/attribution_clients.schema.yaml
```diff
@@ -0,0 +1,22 @@
...
+- mode: NULLABLE
+  name: submission_date
+  type: DATE
+  description:
+
+- mode: NULLABLE
+  name: client_id
+  type: STRING
+  description:
+
```
bani, line 11: missing sample_id

sql_generators/mobile_kpi_support_metrics/templates/attribution_clients.query.sql
```diff
@@ -0,0 +1,240 @@
+{{ header }}
+WITH new_profiles AS (
+  SELECT
+    submission_date,
+    client_id,
+    sample_id,
+    {% if 'is_suspicious_device_client' in product_attribution_group_names %}
```
bani, line 7: I see it on the table but not on the view

kik-kik: You are correct. I added it now.

sql_generators/mobile_kpi_support_metrics/templates/attribution_clients.schema.yaml
kik-kik: added.

bani (approved):

PR #5975, opened 2024-07-25
mobile internet outages - rough draft

lelilia:
hack week first draft for mobile

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4397)


sql/moz-fx-data-shared-prod/internet_outages/mobile_global_outages_v1/query.sql
```diff
@@ -0,0 +1,180 @@
+-- combine baseline pings from different products
+
+with
+-- Firefox for Android
+fenix_baseline as (
+    select
+        metadata.geo.country,
```
kik-kik, line 7: Should we just use your udf here and have `??` if values are missing. This way we know that downstream we only need to deal with `??`

lelilia: thought about it, but since we are not joining we don't really need it though?

PR #5982, opened 2024-07-26
Remove glam intermediate tables from shredder

BenWu:
These tables have seven days of retention so shredder doesn't actually do anything when it tries to delete.  I'm looking to clean up the config and be more explicit about what doesn't need shredding.

@edugfilho Does removing these tables from shredder make sense to you?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4407)


edugfilho (approved):
Makes total sense. I didn't know they were there. Thank you!

PR #5984, opened 2024-07-29
Remove percentiles columns from client_probe_counts schema

BenWu:
percentiles columns were removed in https://github.com/mozilla/bigquery-etl/pull/5966 so this is failing dry run which wasn't caught in the original pr because these queries weren't changed.  Also updated the upstream `glam_client_probe_counts_extract_v1` schema for consistency but I don't think this uses bqetl tooling.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4410)


edugfilho (approved):
Many thanks for this.

PR #6002, opened 2024-08-01
Auxiliary functions for shredder mitigation

lucia-vargas-a:
Auxiliary functions required to generate the query for a backfill with shredder mitigation.
These functions will be used to extract information from the queries and don't validate the correctness of these queries.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4443)


bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -0,0 +1,193 @@
...
+    """Different status of a column during shredder mitigation."""
+
+    COMMON = "COMMON"
+    ADDED = "ADDED"
+    REMOVED = "REMOVED"
+    UNDETERMINED = "None"
+
+
+class DataTypeGroup(Enum):
+    """Data types in BigQuery. Not including ARRAY and STRUCT as these are not expected in aggregate tables."""
```
lucia-vargas-a, line 30: This grouping is mostly for debugging purposes. If a backfill with shredder mitigation fails, the output / logs will return information about the columns used, failed, data type group, if it's a new or existing column and if it's a metric or a dimension. That's why a general group works for starters.
It's also possible to remove the grouping and return the exact type in case a message like `Checks are failing due to a new column X of a **numeric** type used as a dimension in the query.` doesn't give enough details. Open to suggestions.

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -0,0 +1,193 @@
...
+        "BIGNUMERIC",
+        "DECIMAL",
+        "INT64",
+        "INT",
+        "SMALLINT",
+        "BIGINT",
+        "TINYINT",
+        "BYTEINT",
+    )
+    FLOAT = ("FLOAT",)
```
lucia-vargas-a, line 46: The separation of FLOAT from other numeric values is to facilitate rounding columns with this dtype in order to avoid exponential numbers when computing very small numbers.

bigquery_etl/cli/utils.py
```diff
@@ -251,3 +251,42 @@ def temp_dataset_option(
         help="Dataset where intermediate query results will be temporarily stored, "
         "formatted as PROJECT_ID.DATASET_ID",
     )
+
+
+def extract_last_group_by_from_query(query_file=None, query=None):
+    """Return the list of columns in the latest group by of a query."""
+    if query_file:
```
lelilia, line 258: This is not tested at all ... also what format do you expect the query_file parameter to be in?

bigquery_etl/cli/utils.py
```diff
@@ -251,3 +251,42 @@ def temp_dataset_option(
         help="Dataset where intermediate query results will be temporarily stored, "
         "formatted as PROJECT_ID.DATASET_ID",
     )
+
+
+def extract_last_group_by_from_query(query_file=None, query=None):
+    """Return the list of columns in the latest group by of a query."""
```
lelilia, line 257: Could you expand a little bit what this is used for? 
I'm not sure I understand why ;)

tests/cli/test_cli_utils.py
```diff
@@ -64,3 +65,54 @@ def test_table_matches_patterns(self):
             pattern="telemetry_live.event_v4",
             invert=True,
         )
+
+    def test_extract_last_group_by_from_query(self):
+        assert ["ALL"] == extract_last_group_by_from_query(
+            query="SELECT column_1 FROM test_table GROUP BY ALL"
+        )
+        assert ["1"] == extract_last_group_by_from_query(
+            query="SELECT column_1, SUM(metric_1) AS metric_1 FROM test_table GROUP BY 1;"
```
lelilia, line 74: the tests don't pass if I lower case `group by` ... which should never happen according to our style guide .. but since you are testing on `GROUP BY 1` which should also never happen, I wanted to flag it 🎏

bigquery_etl/cli/utils.py
lucia-vargas-a: It expects a string. This will be called from the backfill command where it's validated in advance. Do you want me to add test at this point?

tests/cli/test_cli_utils.py
lucia-vargas-a: TY! I updated the function to case insensitive.

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -0,0 +1,196 @@
...
+                    key,
+                    get_bigquery_type(value),
+                    ColumnType.DIMENSION,
+                    ColumnStatus.ADDED,
+                )
+            )
+        elif (
+            key not in existing_columns
+            and key not in new_columns
+            and get_bigquery_type(value) is DataTypeGroup.NUMERIC
```
lelilia, line 160: I think you need to add () around the `or`

lelilia, line 160: you could use `:=` to save the result for `get_bigquery_type(value)` so you don't have to call the function again

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -0,0 +1,196 @@
...
+def classify_columns(
+    new_row: dict, existing_columns: list, new_columns: list
+) -> tuple[list[Column], list[Column], list[Column], list[Column], list[Column]]:
+    """Compare the new row with the existing columns and return the list of common, added and removed columns by type."""
+    common_dimensions = []
+    added_dimensions = []
+    removed_dimensions = []
+    metrics = []
+    undefined = []
+
```
lelilia, line 126: what is the expected behavior if either `existing_columns` or `new_columns` is an empty list? 
I assume in that case we can't do anything and might want to stop the process?

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -0,0 +1,196 @@
...
+    for key in existing_columns:
+        if key in existing_columns and key not in new_columns:
```
lelilia, line 128: in line 127 you have `existing_columns` and in line 128 as well ... one of those should probably be the `new_row` instead?

lelilia:
```suggestion
        if key not in new_columns:
```

bigquery_etl/backfill/shredder_mitigation.py
lelilia: another thing that might need to be handled: if a dimension is in the `new_columns` but not in `new_row` this is not catched at all currently

lucia-vargas-a: Thanks! I added a validation and tests for both cases.

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -0,0 +1,207 @@
...
+                    get_bigquery_type(value),
+                    ColumnType.METRIC,
+                    ColumnStatus.COMMON,
+                )
+            )
+        else:
+            undefined.append(
+                Column(
+                    key,
+                    get_bigquery_type(None),
```
lelilia, line 187: not sure I totally get why you default to UNDEFINED here ... 
wouldn't it be helpful for debugging to see that it's a string or a numeric value?

lelilia (approved):
left one small comment but otherwise this looks good and I can't think of more tests at the moment :)

bigquery_etl/backfill/shredder_mitigation.py
lucia-vargas-a: The default to UNDETERMINED is for metric/dimension which I couldn't be identified. The key and value would show, it's all the info available.
-- I did change that I was calling the function instead of `value_type`.

PR #6060, opened 2024-08-14
Generate query with shredder mitigation

lucia-vargas-a:
Auto-generate the query to use in a backfill with mitigation.

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4525)


bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -50,30 +67,187 @@ class DataTypeGroup(Enum):
...
     """Representation of a column in a query, with relevant details for shredder mitigation."""
 
-    def __init__(self, name, data_type, column_type, status):
-        """Initialize class with required attributes."""
-        self.name = name
-        self.data_type = data_type
-        self.column_type = column_type
-        self.status = status
-
-    def __eq__(self, other):
```
lelilia, line 63: why did you get rid of the __eq__? that sounds useful

lucia-vargas-a: Cause I'm be changing it for an autogenerated `@attr.define(eq=True)`

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -50,30 +67,187 @@ class DataTypeGroup(Enum):
...
+    project_id: str = attr.field(default=DEFAULT_PROJECT_ID)
+    expiration_days: Optional[float] = attr.field(default=None)
+
+    @destination_table.validator
+    def validate_table_version(self, attribute, value):
+        """Raise an exception if the version table name doesn't end with an integer."""
+        match = re.search(r"v(\d+)$", self.destination_table)
+        try:
+            version = int(match.group(1))
+            if not isinstance(version, int) or version <= 0:
```
lelilia, line 116: how would you even get a negative version number with that regex? 
not saying that it's not worth testing for but `table_v-10` would not be caught by the regex as is. 
If you want to match it and then filter it out here (not sure why you would want to though) you could do `r"v(-?\d+)$"` instead

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -50,30 +67,187 @@ class DataTypeGroup(Enum):
...
+                )
+        except AttributeError:
+            f"Invalid table version for {value}"
+
+    @property
+    def expiration_ms(self) -> Optional[float]:
+        """Convert partition expiration from days to milliseconds."""
+        if self.expiration_days is None:
+            return None
+        return int(self.expiration_days * 86400000)
```
lelilia, line 128:
```suggestion
        return int(self.expiration_days * 86_400_000)
```
tiny nit .. but makes it easier for me to see large numbers with splits ... can be ignored if you don't agree :)

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -50,30 +67,187 @@ class DataTypeGroup(Enum):
...
+
+    @property
+    def version(self):
+        match = re.search(r"v(\d+)$", self.destination_table)
+        return int(match.group(1))
+
+    @property
+    def name_previous_version(self):
+        """Return the table name for the previous version of the table."""
+        return f"{self.destination_table[:-1]}{int(self.version)-1}"
```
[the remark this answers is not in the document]
lelilia: oh and also ... if there is a table `_v10` only removing the last digit would return `v19` 
maybe use the parsed version to remove the right amount of strings 
```suggestion
        return f"{self.destination_table.removesuffix(str(self.version)}{self.version-1}"
```
also .. you don't need to `int()` the version as it is returned as an integer

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -50,30 +67,187 @@ class DataTypeGroup(Enum):
...
+    def validate_status(self, attribute, value):
+        """Check that the type of parameter column_status is as expected."""
+        if not isinstance(value, ColumnStatus):
+            raise ValueError(f"Invalid data type for: {value}.")
+
+
+@attr.define
+class Subset:
+    client: bigquery.Client()
+    destination_table: str = attr.field()
```
lelilia, line 103: I'm slightly confused by the parameters

is the `destination_table` with or without the `_v3`?

From the version function I assume with ... but what is the suffix then? in the `name()` function you add the suffix to it and I'm not sure what that would be... is it related to partitioning?

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -50,30 +67,187 @@ class DataTypeGroup(Enum):
...
+        else:
+            return self.destination_table
+
+    @property
+    def version(self):
+        match = re.search(r"v(\d+)$", self.destination_table)
+        return int(match.group(1))
+
+    @property
+    def name_previous_version(self):
```
lelilia, line 143: could this function be named `previous_destination_table` or something like this? because the `name()` function includes a suffix (even though I don't really know what that would be yet) and this name-function does not ...

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -50,30 +67,187 @@ class DataTypeGroup(Enum):
...
+        return f"{self.destination_table[:-1]}{int(self.version)-1}"
+
+    @property
+    def full_table_id(self):
+        return f"{self.project_id}.{self.dataset}.{self.name}"
+
+    @property
+    def partitioning(self) -> str:
+        metadata = Metadata.from_file(
+            Path("sql") / self.project_id / self.dataset / self.name / METADATA_FILE
```
lelilia, line 154: sure this would be `self.name` and not `self.destination_table`? because whatever the suffix might be .. it's probably not in the file path???

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -50,30 +67,187 @@ class DataTypeGroup(Enum):
...
+        from_clause=None,
+        where_clause=None,
+        group_by_clause=None,
+        order_by_clause=None,
+    ):
+        """Build query to populate the table."""
+
+        query = f"SELECT {', '.join(select_list)}"
+        query += f" {from_clause}" if from_clause is not None else ""
+        query += f" {where_clause}" if where_clause is not None else ""
```
lelilia, line 175: So you would need to set the where_clause as  "WHERE x = y" instead of just "x+y"?

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -50,30 +67,187 @@ class DataTypeGroup(Enum):
...
+        query += f" {group_by_clause}" if group_by_clause is not None else ""
+        query += f" {order_by_clause};" if order_by_clause is not None else ""
+        return query
+
+    def get_query_path(self):
+        """Return the path of the query file for this subset."""
+        query_files = paths_matching_name_pattern(
+            f"{self.dataset}.{self.name}", "sql", self.project_id
```
lelilia, line 183: again self.name might include some suffix?

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -50,30 +67,187 @@ class DataTypeGroup(Enum):
...
+        )
+        if query_files == []:
+            raise click.ClickException(
+                f"Query file not found for `{self.full_table_id}`."
+            )
+        return query_files[0]
+
+    def get_query_path_results(
+        self, backfill_date: date = PREVIOUS_DATE, rows: int = None, **kwargs
```
lelilia, line 192: Is the previous date always a day off? couldn't it be a query that runs once a week or something?

lelilia, line 192: could we rename `rows` to `row_limit` as it seems to be the row limit

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -212,3 +390,188 @@ def classify_columns(
...
+    previous_group_by = extract_last_group_by_from_query(previous.get_query_path())
+    new_group_by = extract_last_group_by_from_query(new.get_query_path())
+    if (
+        "ALL" in previous_group_by
+        or "ALL" in new_group_by
+        or not all(isinstance(e, str) for e in previous_group_by)
+        or not all(isinstance(e, str) for e in new_group_by)
+        or not previous_group_by
+        or not new_group_by
+    ):
```
lelilia, line 421: should you also avoid `GROUP BY 1, 2` here?

bigquery_etl/backfill/shredder_mitigation.py
lucia-vargas-a: Makes sense!

bigquery_etl/backfill/shredder_mitigation.py
lucia-vargas-a: Done!

bigquery_etl/backfill/shredder_mitigation.py
lucia-vargas-a: Actually moved the validation to the property itself.

bigquery_etl/backfill/shredder_mitigation.py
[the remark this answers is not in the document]
lucia-vargas-a: I've added a validation in the function `if new.version < 2:`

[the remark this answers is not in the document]
lucia-vargas-a: Thanks for catching the problem removing only the last digit! I've moved this outside the class because it's not common to all subsets and I'm using the version for this. It's now called `destination_table_previous_version`.

bigquery_etl/backfill/shredder_mitigation.py
lucia-vargas-a: Indeed! The suffix is generating confusion and it's only available in case it's required to save the subsets in temp tables. Since it's not the case the suffix is removed and this now uses `destination_table`.

bigquery_etl/backfill/shredder_mitigation.py
lucia-vargas-a: Updated to use destination_table, thanks!

bigquery_etl/backfill/shredder_mitigation.py
lucia-vargas-a: Definitely.

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -50,30 +67,187 @@ class DataTypeGroup(Enum):
...
+
+        with open(self.get_query_path(), "r") as file:
+            sql_text = file.read().strip()
+
+        if sql_text.endswith(";"):
+            sql_text = sql_text[:-1]
+        if having_clause:
+            sql_text = f"{sql_text} {having_clause}"
+        if rows:
+            sql_text = f"{sql_text} LIMIT {rows};"
```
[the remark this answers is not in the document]
lucia-vargas-a: That's me wanting the queries to be tidy and end with ; in case there is another query after. In this case it's not required so instead of moving it outside LIMIT, it's simpler to not add it. Thanks,

bigquery_etl/backfill/shredder_mitigation.py
lucia-vargas-a: Yes, the destination_table includes the version.
The intention of the suffix was to enable saving the result of the subset results to temp tables which would have e.g. timestamp as suffix. However the subsets are calculated in a single query which makes suffixes an unnecessary complication, thus removed. Please do unresolve if needed!

bigquery_etl/backfill/shredder_mitigation.py
lucia-vargas-a: Definitely, I added a validation, thanks!

bigquery_etl/backfill/shredder_mitigation.py
lucia-vargas-a: Yes, only SELECT takes a list, for all the other statements the expectation is to pass them complete. That's  because they can include more than just a list, are harder to predict.

lucia-vargas-a: Ended up changing it so it works the same in all cases. Thanks!

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -29,11 +52,11 @@ class ColumnStatus(Enum):
 
 
 class DataTypeGroup(Enum):
-    """Data types in BigQuery. Not including ARRAY and STRUCT as these are not expected in aggregate tables."""
+    """Data types in BigQuery. Not supported: TIMESTAMP, ARRAY and STRUCT, which are dtypes currently not expected in aggregates."""
 
     STRING = ("STRING", "BYTES")
     BOOLEAN = "BOOLEAN"
-    NUMERIC = (
+    INTEGER = (
```
lelilia, line 59: Numeric is not an integer though? It's closer to a float?

lucia-vargas-a: Indeed numeric covers most of the types, including integer. I'd propose:
 INTEGER / BIGINT
 NUMERIC
 FLOAT
and let the other types be added incrementally as aggreggates require them, since most of these types are in client-level tables and not in aggregates. What do you think?

lelilia: not sure I understand your suggestion?

Do you mean
NUMERIC = (integer, numeric, float)
or 
INTEGER = (integer, int64, bigint,..)
NUMERIC = (numeric, bignumeric,...)
FLOAT = (float, double, ..)
?
fine either way

lucia-vargas-a: Not quite. I mean
INTEGER = INTEGER, INT64, INT
NUMERIC = NUMERIC
FLOAT = FLOAT

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -29,83 +52,247 @@ class ColumnStatus(Enum):
...
+        if not select_list or not from_clause:
+            raise click.ClickException(
+                f"Missing required clause to generate query.\nActuals: SELECT: {select_list}, FROM: {self.full_table_id}"
+            )
+        query = f"SELECT {', '.join(map(str, select_list))}"
+        query += f" FROM {from_clause}" if from_clause is not None else ""
+        query += f" WHERE {where_clause}" if where_clause is not None else ""
+        query += f" GROUP BY {group_by_clause}" if group_by_clause is not None else ""
+        query += f" ORDER BY {order_by_clause}" if order_by_clause is not None else ""
+        query += (
```
lelilia, line 179: `HAVING` needs to be before `ORDER BY`

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -29,83 +52,247 @@ class ColumnStatus(Enum):
...
+        return query
+
+    def get_query_path(self):
+        """Return the path of the query.sql file for this subset."""
+        query_files = paths_matching_name_pattern(
+            Path("sql")
+            / self.project_id
+            / self.dataset
+            / self.destination_table
+            / "query.sql",
```
lelilia, line 193: is there a possibility that this might be a query.py or script.sql file?

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -29,97 +47,260 @@ class ColumnStatus(Enum):
...
+                    query_parameters=parameters,
+                    use_legacy_sql=False,
+                    dry_run=False,
+                    use_query_cache=False,
+                ),
+            )
+            query_results = query_job.result()
+        except NotFound as e:
+            raise click.ClickException(
+                f"Unable to query current data, table {self.full_table_id} doesn't exist."
```
lelilia, line 247: here you might want to have the full name + the partition? (in case that the suffix)

lucia-vargas-a: Yes, the partition is useful, I'll add it. The suffix (that I was using before) is removed.

bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -29,97 +47,259 @@ class ColumnStatus(Enum):
...
+            return DataTypeGroup.TIMESTAMP
+    except (ValueError, TypeError, AttributeError):
+        pass
     if isinstance(value, time):
-        return DataTypeGroup.DATE
+        return DataTypeGroup.TIME
     if isinstance(value, date):
         return DataTypeGroup.DATE
-    elif isinstance(value, bool):
+    if isinstance(value, bool):
```
lelilia, line 286: why did you move away from `elif` to `if`? 
don't think it makes a difference since you return

lelilia (approved):
awesome

PR #6084, opened 2024-08-22
Include tables with no expirations in table_partition_expirations

BenWu:
Adds stable tables that don't have partition expiration to `monitoring_derived.table_partition_expirations_v1`.  This is only a few tables in `telemetry_stable` right now but it might be useful to be able to check for glean tables that don't have retention set but should

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4585)

### DIFF
--- a/sql_generators/table_partition_expirations/templates/table_partition_expirations.sql
+++ b/sql_generators/table_partition_expirations/templates/table_partition_expirations.sql
@@ -10,7 +10,7 @@ WITH
       FROM
         `moz-fx-data-shared-prod.{{ dataset_id }}.INFORMATION_SCHEMA.PARTITIONS`
       WHERE
-        partition_id != '__NULL__'
+        partition_id != "__NULL__"
       QUALIFY
         ROW_NUMBER() OVER (PARTITION BY table_name ORDER BY partition_id) = 1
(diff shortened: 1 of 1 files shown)


kwindau (approved):

sql_generators/table_partition_expirations/templates/table_partition_expirations.sql
```diff
@@ -55,6 +55,14 @@ current_partitions AS (
     {% endif %}
   {% endfor %}
 ),
+partition_expirations AS (
+  SELECT
+    *
+  FROM
+    `moz-fx-data-shared-prod.region-us.INFORMATION_SCHEMA.TABLE_OPTIONS`
```
kwindau, line 62: Is it always in region-us or should this be {{ dataset_id }} also?

BenWu: This is outside of the for loop so `region-us` works here because we can just get options for every table in the project and left join them which is faster.  `{{ dataset_id }}` is used for `INFORMATION_SCHEMA.PARTITIONS` since it only supports that

PR #6094, opened 2024-08-23
fix(dryrun): table_metadata used before reference

ANich:
The following is showing up in a few places:
```
INFO - [base]  cannot access local variable 'table_metadata' where it is not associated with a value
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4597)

### DIFF
--- a/bigquery_etl/dryrun.py
+++ b/bigquery_etl/dryrun.py
@@ -276,23 +276,7 @@ class DryRun:
                     else:
                         raise e
 
-                if (
-                    self.project is not None
-                    and self.table is not None
-                    and self.dataset is not None
-                ):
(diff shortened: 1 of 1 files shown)


bigquery_etl/dryrun.py
ANich, line 1: Related to this snippet:
```
                if (
                    self.project is not None
                    and self.table is not None
                    and self.dataset is not None
                ):
```
there are also local `project`, and `dataset` variables defined above that might be used here? Haven't looked too deeply into the significance of using the properties

kwindau (approved):
LGTM

PR #6138, opened 2024-08-30
Bug 1915724 - Update accounts events sql check 

akkomar:
This updates sql check by filtering out non-production/stage data that appeared recently.

Also removes some no-longer necessary workarounds for bugs that are now fixed in production.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4679)


kwindau (approved):
LGTM

PR #6183, opened 2024-09-12
fixup: Allow skipping apps in event monitoring generator

akkomar:
We need to also skip the app in the view that combines all apps.

### DIFF
--- a/sql_generators/glean_usage/event_monitoring_live.py
+++ b/sql_generators/glean_usage/event_monitoring_live.py
@@ -1,9 +1,7 @@
 """Generate Materialized Views and aggregate queries for event monitoring."""
 
 import os
-import re
-
-from collections import namedtuple, OrderedDict
+from collections import OrderedDict, namedtuple
 from datetime import datetime
 from pathlib import Path
 from typing import List, Set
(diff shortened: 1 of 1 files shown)


sql_generators/glean_usage/event_monitoring_live.py
```diff
@@ -107,6 +94,12 @@ def generate_per_app_id(
             if dataset == app_dataset["bq_dataset_family"]
         ][0]
 
+        # Skip any not-allowed app.
+        if app_name in ConfigLoader.get(
```
akkomar, line 98: We can use app_name instead of app_id here and avoid a regex above.

BenWu (approved):

sql_generators/glean_usage/event_monitoring_live.py
```diff
@@ -212,7 +209,7 @@ def generate_across_apps(
                     event_tables_per_dataset[dataset] = events_table_overwrites[
                         app_name
                     ]
-                else:
+                elif app_name not in skipped_apps:
```
BenWu, line 212: CI failure is because the app is still in the view that unions all the apps so the app needs to be removed from the list or `render_kwargs` needs to be changed

BenWu (changes requested):

sql_generators/glean_usage/event_monitoring_live.py
BenWu: Actually the error is for `accounts_backend` so something else is happening here but I don't know what

scholtzan: mh, the stage deploy pipeline doesn't seem to be able to deploy the materialized views

BenWu: Oh right, I was talking to Arkadiusz about this as well.  I think it's because initialize only runs for queries with `is_init()` https://github.com/mozilla/bigquery-etl/blob/9fc513ba9f10286accca16ebe69e2cf18c64b33c/bigquery_etl/cli/query.py#L1490

akkomar:
Superseded by https://github.com/mozilla/bigquery-etl/pull/6191

Issue #6185, opened 2024-09-12
Engineers want a PR template that encourages better description of what's being changed.


edugfilho:
During recent conversations, some engineers signaled an interest in having Pull Requests that are more descriptive. This issue captures the desire for a template in this repository that encourages that.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4792)

PR #6186, opened 2024-09-12
chore: update PR template

edugfilho:
## Description

This PR updates the Pull Request template of this repo with the intention to make users write more descriptive PRs.

* https://github.com/mozilla/bigquery-etl/issues/6185

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -15,4 +15,4 @@ configured to automatically insert hyperlinks for DSRE and DENG tickets.
 See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-autolinks-to-reference-external-resources
 -->
 
-Reviewer, please follow [this checklist](https://github.com/mozilla/bigquery-etl/.github/reviewer_checklist.md)
+**Reviewer, please follow [this checklist](https://github.com/mozilla/bigquery-etl/.github/reviewer_checklist.md)**
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,10 +1,18 @@
-Checklist for reviewer:
+## Description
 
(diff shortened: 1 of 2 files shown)


kwindau (approved):

.github/pull_request_template.md
```diff
@@ -1,10 +1,18 @@
...
+* DENG-XXXX
+* DSRE-XXXX
+
+<!--
+Please reference related Jira tickets, GitHub issues or Bugzilla. This repo has been
+configured to automatically insert hyperlinks for DSRE and DENG tickets.
+See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-autolinks-to-reference-external-resources
+-->
+
+Reviewer, please follow [this checklist](https://github.com/mozilla/bigquery-etl/.github/reviewer_checklist.md)
```
BenWu, line 18: I expect that putting this in a separate file will make it even more ignored so maybe we're better off not having it to not set expectations we won't meet.  I'm ok with leaving this line here since it doesn't take up much space but I think it's worth considering keeping the checklist and making it a lot shorter.

We could also have the checklist in a separate automated comment like https://github.com/marketplace/actions/github-script#comment-on-an-issue so the preview is cleaner but I don't know if that's better or worse.

scholtzan (approved):

.github/pull_request_template.md
edugfilho: I agree that there's even less of a chance reviewers are going to look at the checklist now. 
At the same time I think there's value in that list, and I couldn't think of anything there that I would remove without feeling like I removed something sort of important (except for the first item, which is covered in the new template), especially in the case of new reviewers. 
Couple that with the intention of adding the least amount of noise, I figured that one line will work as documentation: very few will actually read it, but we can point folks to it and it gives new hires some direction. Btw I think I will make it **bold** to be purposefully a little more annoying.

PR #6189, opened 2024-09-12
RS-1278 Desktop tiles forecast inputs fixes

curtismorales:
A couple things here:

* Remove the `change_controlled` flag on the table; we don't need it
* Change the destination partition and the `@submission_date` (renamed to `@submission_month`) parameter to point to the first of the current month (this is currently broken)
* Update the SQL accordingly

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4799)


kwindau (approved):

jaredsnyder (commented):
lgtm!

PR #6194, opened 2024-09-12
Generate Bigeye monitoring configs in CI

scholtzan:
This triggers `bqetl monitoring update` in the CI to generate the Bigconfig files for the tables that should be monitored

Checklist for reviewer:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4806)

### DIFF
--- a/bigquery_etl/cli/monitoring.py
+++ b/bigquery_etl/cli/monitoring.py
@@ -5,7 +5,7 @@ from pathlib import Path
 from typing import Optional
 
 import click
-from bigeye_sdk.authentication.api_authentication import ApiAuth
+from bigeye_sdk.authentication.api_authentication import APIKeyAuth
 from bigeye_sdk.client.datawatch_client import datawatch_client_factory
 from bigeye_sdk.controller.metric_suite_controller import MetricSuiteController
 from bigeye_sdk.exceptions.exceptions import FileLoadException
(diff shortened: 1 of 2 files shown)


bigquery_etl/cli/monitoring.py
```diff
@@ -175,9 +203,15 @@ def validate(name: str, sql_dir: Optional[str], project_id: Optional[str]) -> No
 
     invalid = False
 
-    for bigconfig_file in bigconfig_files:
+    for bigconfig_file in list(set(bigconfig_files)):
         try:
             BigConfig.load(bigconfig_file)
+        except FileLoadException as e:
+            if "Duplicate" in e.message:
+                pass
```
scholtzan, line 211: Bigeye doesn't like processing the same file twice. It seems to keep a cache of files that it processed, so if this command is called twice on the same file this exception is thrown. We don't want everything to stop, so just ignore duplicate files

bigquery_etl/cli/monitoring.py
```diff
@@ -48,31 +49,58 @@ def monitoring(ctx):
...
 @click.option(
     "--workspace",
-    default="DEFAULT",
+    default=463,
     help="Bigeye workspace to use when authenticating to API.",
 )
+@click.option(
+    "--api-key",
+    "--api_key",
+    required=True,
```
kik-kik, line 58: I wonder if this should be an option, it sounds like at some point we could potentially end up with plain text API key being passed into the command at some point and getting stored in some logs? Maybe we could just have a check inside the function to make sure the env variable is defined?

kik-kik (approved):

bigquery_etl/cli/monitoring.py
scholtzan: I was thinking it would be slightly more convenient when running the tooling locally. This way users get an error automatically if the env variable is not set and at the same time see that an API key is required when calling `bqetl monitoring deploy --help`.

I'll remove the option for now just to be safe. I guess we can add it back later if it turns out to be more convenient

PR #6217, opened 2024-09-18
CI validation of tables with the shredder_mitigation label

lucia-vargas-a:
## Description
Implements [DENG-4659](https://mozilla-hub.atlassian.net/browse/DENG-4659) for CI validation of tables with the shredder_mitigation label.

[DENG-4659]: https://mozilla-hub.atlassian.net/browse/DENG-4659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4872)

### DIFF
--- a/bigquery_etl/metadata/id_level_columns.yaml
+++ b/bigquery_etl/metadata/id_level_columns.yaml
@@ -1,0 +1,3 @@
+id_level_columns:
+  - client_id
+  - device_id
--- a/bigquery_etl/metadata/id_level_columns.yaml
+++ b/bigquery_etl/metadata/id_level_columns.yaml
@@ -1,3 +1,5 @@
 id_level_columns:
   - client_id
   - device_id
+  - profile_id
+  - profile_group_id
--- a/bigquery_etl/metadata/validate_metadata.py
+++ b/bigquery_etl/metadata/validate_metadata.py
@@ -4,13 +4,16 @@ import logging
 import os
 import sys
 from argparse import ArgumentParser
+from pathlib import Path
 
 import click
+import yaml
 
 from bigquery_etl.config import ConfigLoader
+from bigquery_etl.schema import SCHEMA_FILE, Schema
 
 from ..util import standard_args
-from ..util.common import project_dirs
+from ..util.common import project_dirs, extract_last_group_by_from_query
 from .parse_metadata import DatasetMetadata, Metadata
 
 parser = ArgumentParser(description=__doc__)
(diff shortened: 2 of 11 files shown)


bigquery_etl/metadata/id_level_columns.yaml
```diff
@@ -0,0 +1,3 @@
+id_level_columns:
```
lucia-vargas-a, line 1: Using a YAML to make it easier to extend this list in the future.

kwindau (approved):

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -130,6 +192,13 @@ def validate(target):
                     ):
                         failed = True
 
+                    if not validate_shredder_mitigation(
+                        query_file_path=root,
```
BenWu, line 196: `query_file_path` is a somewhat misleading name because it's the query dir instead and gets `query.sql` appended to it.  I would recommend renaming it to something like `query_dir`

tests/cli/test_cli_metadata.py
BenWu, line 1: It would be a useful regression test to test with `"labels": {}` and make sure the checks don't fail non-shredder mitigation queries.

BenWu (commented):
some minor comments

tests/cli/test_cli_metadata.py
lucia-vargas-a: Added!

tests/cli/test_cli_metadata.py
```diff
@@ -414,3 +419,125 @@ def test_metadata_deprecate_no_metadata(self, runner):
...
+        metadata = {
+            "friendly_name": "Test",
+            "labels": {},
+        }
+        schema = """
+            fields:
+              - mode: NULLABLE
+                name: column_1
+                type: DATE
+                description: description 1
```
BenWu, line 530: I think this should be a case that would fail if the label was present so that it verifies the code path isn't run at all.
```suggestion
```

BenWu (approved):
One suggestion but LGTM

tests/cli/test_cli_metadata.py
lucia-vargas-a: Makes sense

PR #6227, opened 2024-09-19
Fix desktop_tiles_forecast_inputs submission_month parameter and move query to dryrun skip list

curtismorales:
## Description

This PR fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1918461 (I think). I had changed `execution_date` to `ds` not realizing that `ds` is a string representation of the Pendulum object `execution_date`, and this caused `strftime` to error. To change this back, I also need to add this query to the dry-run skip list because the dryrun code only supports parameters that use `ds`

* [Bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1918461)

### DIFF
--- a/sql/moz-fx-data-shared-prod/ads_derived/desktop_tiles_forecast_inputs_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/ads_derived/desktop_tiles_forecast_inputs_v1/metadata.yaml
@@ -14,10 +14,10 @@ labels:
 scheduling:
   dag_name: bqetl_ads
   destination_table: >-
-    desktop_tiles_forecast_inputs_v1${{ds.strftime("%Y%m01")}}
+    desktop_tiles_forecast_inputs_v1${{execution_date.strftime("%Y%m01")}}
   parameters:
     - >-
-      submission_month:DATE:{{ds.strftime('%Y-%m-01')}}
+      submission_month:DATE:{{execution_date.strftime('%Y-%m-01')}}
   query_file_path:
     # explicit query file path is necessary because the destination table
     # includes a partition identifier that is not in the path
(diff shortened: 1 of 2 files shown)


sql/moz-fx-data-shared-prod/ads_derived/desktop_tiles_forecast_inputs_v1/metadata.yaml
```diff
@@ -14,10 +14,10 @@ labels:
 scheduling:
   dag_name: bqetl_ads
   destination_table: >-
-    desktop_tiles_forecast_inputs_v1${{ds.strftime("%Y%m01")}}
+    desktop_tiles_forecast_inputs_v1${{execution_date.strftime("%Y%m01")}}
```
BenWu, line 17: `execution_date` is deprecated but `logical_date` should be equivalent
https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html#deprecated-variables
```suggestion
    desktop_tiles_forecast_inputs_v1${{logical_date.strftime("%Y%m01")}}
```

sql/moz-fx-data-shared-prod/ads_derived/desktop_tiles_forecast_inputs_v1/metadata.yaml
```diff
@@ -14,10 +14,10 @@ labels:
...
   parameters:
     - >-
-      submission_month:DATE:{{ds.strftime('%Y-%m-01')}}
+      submission_month:DATE:{{execution_date.strftime('%Y-%m-01')}}
```
BenWu, line 20:
```suggestion
      submission_month:DATE:{{logical_date.strftime('%Y-%m-01')}}
```

BenWu (approved):
r+wc

PR #6251, opened 2024-09-24
feat(): add distribution id to attribution clients template

kik-kik:
## feat(): add distribution id to attribution clients template

We need to add this field to the attribution field to be able to identify which partnership a specific client came from.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4935)


lucia-vargas-a (approved):
LGTM

PR #6256, opened 2024-09-24
Allow specifying a collection for monitored tables

scholtzan:
## Description

This extends the `monitoring` metadata to allow for specifying a collection that the metric should be added to.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4943)

### DIFF
--- a/bigquery_etl/cli/monitoring.py
+++ b/bigquery_etl/cli/monitoring.py
@@ -157,13 +157,13 @@ def update(name: str, sql_dir: Optional[str], project_id: Optional[str]) -> None
                                     metric.metric_type.predefined_metric
                                 )
 
-                        if (
-                            metadata.monitoring.collection
-                            and collection.collection is None
-                        ):
-                            collection.collection = SimpleCollection(
-                                name=metadata.monitoring.collection
-                            )
+                    if (
+                        metadata.monitoring.collection
+                        and collection.collection is None
+                    ):
+                        collection.collection = SimpleCollection(
+                            name=metadata.monitoring.collection
+                        )
 
                 if len(default_metrics) > 0:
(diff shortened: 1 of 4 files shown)


bigquery_etl/cli/monitoring.py
```diff
@@ -156,6 +157,14 @@ def update(name: str, sql_dir: Optional[str], project_id: Optional[str]) -> None
                                     metric.metric_type.predefined_metric
                                 )
 
+                        if (
+                            metadata.monitoring.collection
+                            and collection.collection is None
+                        ):
+                            collection.collection = SimpleCollection(
+                                name=metadata.monitoring.collection
+                            )
```
sean-rose, line 166: This doesn't seem to use the `deployment` values, so I don't think it needs to be done in that loop, but could be done in the parent `collection` loop.
```suggestion
                    if (
                        metadata.monitoring.collection
                        and collection.collection is None
                    ):
                        collection.collection = SimpleCollection(
                            name=metadata.monitoring.collection
                        )
```

bigquery_etl/metadata/parse_metadata.py
```diff
@@ -157,6 +157,7 @@ class MonitoringMetadata:
     """Metadata for specifying observability and monitoring configuration."""
 
     enabled: bool = attr.ib(True)
+    collection: Optional[str] = attr.ib(None)
```
sean-rose, line 160: In the Bigeye UI a metric can be added to multiple collections, and it's unfortunate that the `bigconfig` setup doesn't seem to allow for that.

sean-rose (approved):
r+wc

PR #6260, opened 2024-09-25
feat(DENG-4577): add monitoring bigeye_usage view to allow us to get more insight into bq usage and costs from bigeye

kik-kik:
# feat(DENG-4577): add monitoring bigeye_usage view to allow us to get more insight into bq usage and costs from bigeye

## Description

This aims to provide a data structure to enable us to build a dashboard in Looker to allow us to monitor BigEye related spending and usage in BigQuery.


sql/moz-fx-data-shared-prod/monitoring_derived/bigquery_usage_v2/query.sql
```diff
@@ -48,6 +48,7 @@ jobs_by_project AS (
       user_email,
       REGEXP_EXTRACT(query, r'Username: (.*?),') AS username,
       REGEXP_EXTRACT(query, r'Query ID: (\w+), ') AS query_id,
+      REGEXP_EXTRACT(query, r'\'metric_id:(\d+)\'') AS bigeye_metric_id,
```
kik-kik, line 51: @Marlene-M-Hirose @wwyc was there a specific reason why we didn't just pass the query through here? There is another bit of information I could extract here for BigEye, but wasn't sure if I should be polluting this query with BigEye specific details? If I had query available here I could just do it downstream in my bigeye monitoring view.

Marlene-M-Hirose: @kik-kik , I'm not sure about adding this specific information here. This table is supposed to be focused on BigQuery usage, not for monitoring bigeye. I would prefer to keep bigeye separate from this table

kik-kik: @wwyc I made a tweak to let the query through but only for the bigeye user. Do you think this would be acceptable?

kik-kik: ok, for now I will remove the query from here and move forward with a more basic version of the model for bigeye monitoring without including the bigeye metadata like metric_id. We can look into extending it in the future or see if BigEye could make this information available via labels which I assume would be fine to store @whd ?

Marlene-M-Hirose (approved):

sql/moz-fx-data-shared-prod/monitoring_derived/bigquery_usage_v2/query.sql
whd: Yes derived data like labels would be fine, assuming that's similar to user names and query ids from redash metadata.

PR #6277, opened 2024-09-30
Add distribution_id to feature_usage_events_v1

m-d-bowerman:
## Description

This PR adds the client distribution_id. This is necessary to count app push notification events and set to default events to support mobile distribution deals.

Adds engagement with the sponsored tiles feature.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-5013)

### DIFF
--- a/sql/moz-fx-data-shared-prod/fenix_derived/feature_usage_events_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/fenix_derived/feature_usage_events_v1/query.sql
@@ -339,7 +339,7 @@ event_ping_clients_feature_usage AS (
     COUNTIF(
       event_category = 'top_sites'
       AND event_name = 'contile_impression'
-    ) AS top_sites_contile_impression
+    ) AS top_sites_contile_impression,
   FROM
     `moz-fx-data-shared-prod.fenix.events_unnested`
   WHERE
--- a/sql/moz-fx-data-shared-prod/fenix_derived/feature_usage_events_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/fenix_derived/feature_usage_events_v1/query.sql
@@ -362,6 +362,7 @@ SELECT
   country,
   adjust_network,
   is_default_browser,
+  distribution_id,
 /*Logins*/
 --autofill_prompt_shown
   SUM(autofill_password_detected_logins) AS autofill_password_detected_logins,
@@ -948,7 +949,6 @@ SELECT
         THEN client_id
     END
   ) AS home_page_customize_home_clicked_users,
-  distribution_id,
 /*Sponsored Tiles*/
 --top_sites_contile_click
   SUM(top_sites_contile_click) AS top_sites_contile_click,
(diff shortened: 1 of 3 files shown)


sql/moz-fx-data-shared-prod/fenix_derived/feature_usage_events_v1/query.sql
```diff
@@ -938,7 +947,27 @@ SELECT
       WHEN home_page_customize_home_clicked > 0
         THEN client_id
     END
-  ) AS home_page_customize_home_clicked_users
+  ) AS home_page_customize_home_clicked_users,
+  distribution_id,
```
kik-kik, line 951: this `distribution_id` is a bit hidden right now among all the aggregations. Would it make sense to make it live just under the `is_default_browser`? We could keep all the grouping keys together this way and the query easier to read.

sql/moz-fx-data-shared-prod/fenix_derived/feature_usage_events_v1/query.sql
```diff
@@ -330,7 +331,15 @@ event_ping_clients_feature_usage AS (
...
-    ) AS home_page_customize_home_clicked
+    ) AS home_page_customize_home_clicked,
+    COUNTIF(
+      event_category = 'top_sites'
+      AND event_name = 'contile_click'
+    ) AS top_sites_contile_click,
+    COUNTIF(
+      event_category = 'top_sites'
+      AND event_name = 'contile_impression'
+    ) AS top_sites_contile_impression
```
kik-kik, line 342:
```suggestion
    ) AS top_sites_contile_impression,
```

sql/moz-fx-data-shared-prod/fenix_derived/feature_usage_events_v1/query.sql
```diff
@@ -30,6 +30,7 @@ client_attribution AS (
     client_id,
     channel,
     adjust_network,
+    distribution_id
```
kik-kik, line 33:
```suggestion
    distribution_id,
```

kik-kik (approved):
r+wc

Just some small comments. Please read through before merging.

kik-kik:
One thing we should think about, since we're adding a new grouping value should we recompute partitions that were generated prior to when we merge this change? Or alternatively, should we have a note in the metadata.yaml inside the description indicating that this data was only added after certain date to minimise the risk of confusion when trying to use this data?

m-d-bowerman:
> One thing we should think about, since we're adding a new grouping value should we recompute partitions that were generated prior to when we merge this change? Or alternatively, should we have a note in the metadata.yaml inside the description indicating that this data was only added after certain date to minimise the risk of confusion when trying to use this data?

Yes, I'll do some validation before merging this and update the metadata.

kik-kik:
@m-d-bowerman Are we good to merge this?

kik-kik (changes requested):
Is this the correct events that we want to track?

kik-kik (approved):

PR #6285, opened 2024-10-02
feat(DENG-4602): add BigEye RunMetricOperator to DAG generation

kik-kik:
# feat(DENG-4602): add BigEye RunMetricOperator to DAG generation

## Description

Changes to add bigeye tasks to trigger BigEye Metric runs for a specific dataset if a corresponding `bigconfig.yml` exists in the datasets directory.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-5031)

### DIFF
--- a/bigquery_etl/query_scheduling/formatters.py
+++ b/bigquery_etl/query_scheduling/formatters.py
@@ -19,10 +19,16 @@ def format_schedule_interval(interval):
 def format_attr(d, attribute, formatter_name):
     """Apply a formatter to a dict attribute."""
     for name in dir(query_scheduling.formatters):
+        if name != formatter_name:
+            continue
+
         formatter_func = getattr(query_scheduling.formatters, name)
-        if callable(formatter_func) and name == formatter_name:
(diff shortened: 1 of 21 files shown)


bigquery_etl/query_scheduling/formatters.py
```diff
@@ -19,10 +19,16 @@ def format_schedule_interval(interval):
 def format_attr(d, attribute, formatter_name):
     """Apply a formatter to a dict attribute."""
     for name in dir(query_scheduling.formatters):
+        if name != formatter_name:
```
kik-kik, line 22: Had to make this tweak here because for some reason the same date attribute was being put through the formatted function twice which resulted in a TypeError inside `format_date` function because the second around the start_date value was already a datetime object. Alternatively, we could adjust the `format_date` function to accept both string and datetime and just return the value if already of datetime type.

scholtzan:
Depends on https://github.com/mozilla/telemetry-airflow/pull/2107

PR #6288, opened 2024-10-02
Give reader role to dry run account for datasets in stage

BenWu:
## Description

This attempts to add the dry run function service account as a reader to datasets created in stage.  Currently, dry run accounts have the reader role at the project level which doesn't match how it's set in shared-prod, resulting in cases where dry runs succeed in stage but fail on main/prod (e.g. querying certain information schema views).  Granting at the dataset level should make it similar to prod so only the job user role is set in the project.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-5064)

### DIFF
--- a/bigquery_etl/cli/stage.py
+++ b/bigquery_etl/cli/stage.py
@@ -11,12 +11,13 @@ import rich_click as click
 from google.cloud import bigquery
 from google.cloud.bigquery.enums import EntityTypes
 
+from .. import ConfigLoader
 from ..cli.query import deploy as deploy_query_schema
 from ..cli.query import update as update_query_schema
 from ..cli.routine import publish as publish_routine
 from ..cli.utils import paths_matching_name_pattern, sql_dir_option
 from ..cli.view import publish as publish_view
-from ..dryrun import DryRun, get_dry_run_service_accounts, get_id_token
+from ..dryrun import DryRun, get_id_token
 from ..routine.parse_routine import (
     ROUTINE_FILES,
     UDF_FILE,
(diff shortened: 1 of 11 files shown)


bigquery_etl/cli/stage.py
```diff
@@ -104,8 +105,8 @@ def deploy(
         # copy SQL to a temporary directory
         tmp_dir = Path(tempfile.mkdtemp())
         tmp_dir.mkdir(parents=True, exist_ok=True)
-        shutil.copytree(sql_dir, tmp_dir)
-        sql_dir = tmp_dir / sql_dir.name
+        shutil.copytree(sql_dir, tmp_dir, dirs_exist_ok=True)
+        sql_dir = tmp_dir / Path(sql_dir).name
```
BenWu, line 109: The `--copy-sql-to-tmp-dir` wasn't working when I was using it locally

bqetl_project.yaml
```diff
@@ -20,6 +20,9 @@ default:
 
 dry_run:
   function: https://us-central1-moz-fx-data-shared-prod.cloudfunctions.net/bigquery-etl-dryrun
+  function_accounts:
+  - bigquery-etl-dryrun@moz-fx-data-shared-prod.iam.gserviceaccount.com
+  - bigquery-etl-dryrun@moz-fx-data-shar-nonprod-efed.iam.gserviceaccount.com
```
BenWu, line 25: It's possible to programatically get the service accounts from the url which is what I did first (in https://github.com/mozilla/bigquery-etl/pull/6288/commits/16fba8a72538e024b971bf4b362620f6d0d3a78a) but it doesn't seem worth the code complexity

BenWu (commented):
I made some trivial changes to sql to verify that deploys are still working https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/39780/workflows/21cd9111-f84b-4457-b534-494660ccf020/jobs/454112

scholtzan (approved):
I guess we'll have to remove project-level permissions for the dry run accounts once this is merged?

BenWu:
Yes the roles can be changed to `bigquery.jobUser` and `bigquery.readSessionUser` to match shred-prod.  I'll wait a bit before changing them to make sure this change by itself doesn't break anything unexpected

PR #6302, opened 2024-10-07
Add bqetl CLI command for setting partition columns

scholtzan:
## Description
This adds a new bqetl CLI command for setting partition columns in Bigeye. We'll need to explicitly set partition columns for views that should be monitored, otherwise the Bigeye monitors will fail to run.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-5203)


alekhyamoz (approved):
LGTM

PR #6372, opened 2024-10-18
Add shell completion script for bqetl

BenWu:
## Description

I often forget the exact command names while writing them so I think this would be useful. Based on these [click docs](https://click.palletsprojects.com/en/8.1.x/shell-completion/) combining the bash and zsh scripts into one.  This works for me on macos with bash and zsh but it's kind of slow (like `bqetl --help`) possibly due to all the imports in the cli files

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-5622)

### DIFF
--- a/docs/bqetl.md
+++ b/docs/bqetl.md
@@ -50,3 +50,29 @@ $ ./bqetl [command] --help
 ```
 
 <!--- Documentation for comments is generated via ./bqetl docs generate -->
+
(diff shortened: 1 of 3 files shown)


scholtzan (approved):
That's useful! Can confirm that it's a bit slow, it seems to take a second or two to auto-complete for me.

PR #6377, opened 2024-10-21
DENG-4768 Add 3 new columns to stable tables views (app_version_major, app_version_minor, app_version_patch)

kwindau:
## Description

This PR adds 3 new columns to the generated stable views:

- app_version_major
- app_version_minor
- app_version_patch

* [DENG-4768](https://mozilla-hub.atlassian.net/browse/DENG-4768)

[DENG-4768]: https://mozilla-hub.atlassian.net/browse/DENG-4768?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-5633)


sql_generators/stable_views/__init__.py
```diff
@@ -28,7 +28,10 @@
 AS
 SELECT
   * REPLACE(
-    {replacements})
+    {replacements}),
+  mozfun.norm.extract_version(client_info.app_display_version, 'major') as app_version_major,
```
ncalexan, line 32: I mentally remembered that these were `STRING`, but it looks like it's `NUMERIC` (per https://github.com/mozilla/bigquery-etl/blob/generated-sql/sql/mozfun/norm/extract_version/udf.sql#L2).  Can you confirm that we can compare with `>=` and friends?

There's a `beta` field as well -- is there any reason to not expose it at the same time?

ncalexan (approved):
This makes sense to me but a stamp from somebody in the ETL space would be better than my review.

ncalexan:
@kwindau -- when thinking about an email to announce this functionality, I realized that it's not obvious how to add this functionality for additional version string fields.  That is, this does what I ask for `client_info.app_display_version` -- but doesn't have any abstraction or affordance for other fields that people might want to explode.  Is there an expression of this idea that's more general, so that we can do it in other places?  Or should we take the win now and deal with generality when the need arises?

sql_generators/stable_views/__init__.py
```diff
@@ -49,7 +65,10 @@
         distribution_id
     ) AS funnel_derived,
   `moz-fx-data-shared-prod`.udf.distribution_model_installs(distribution_id) AS distribution_model,
-  `moz-fx-data-shared-prod`.udf.partner_org_installs(distribution_id) AS partner_org
+  `moz-fx-data-shared-prod`.udf.partner_org_installs(distribution_id) AS partner_org,
+  mozfun.norm.extract_version(client_info.app_display_version, 'major') as app_version_major,
+  mozfun.norm.extract_version(client_info.app_display_version, 'minor') as app_version_minor,
+  mozfun.norm.extract_version(client_info.app_display_version, 'patch') as app_version_patch,
```
BenWu, line 71: `firefox_installer_stable.install_v1` doesn't have `client_info` or a `app_display_version` column so this should probably be `installer_version`

sql_generators/stable_views/__init__.py
```diff
@@ -281,6 +301,23 @@ def write_view_if_not_exists(target_project: str, sql_dir: Path, id_token=None,
             ),
             trailing_newline=True,
         )
+    #For views without app_display_version, use the template without app version parsing
```
BenWu, line 304: `client_info` is a glean field so this would break all other views.  You can make use of `schema.schema_id == "moz://mozilla.org/schemas/glean/ping/1"` like on line 186 above so that `client_info.app_display_version` is only used on those tables.

For non-glean, `application.display_version` is used in some places but not everywhere so you would need to have a lot of special cases.  I would consider just scoping this change to glean views

BenWu (changes requested):
This would only work for glean so I would recommend `VIEW_QUERY_TEMPLATE_NO_CLIENT_INFO` being the default and `VIEW_QUERY_TEMPLATE` only being used for glean

sql_generators/stable_views/__init__.py
kwindau: Thanks @BenWu - will do!! :)

kwindau:
This got messy, going to refactor and open a new PR - thanks for all the help!

PR #6386, opened 2024-10-22
chore(bugzilla_metrics): switch to single dataset for syndication

whd:
/CC @dklawren 

Per https://github.com/mozilla-it/webservices-infra/pull/3058#issuecomment-2428014346 we're changing syndication here to be stage/prod instead of prod/prod.


dklawren (approved):

mikaeld (approved):
:medal_sports:

whd:
CI was so slow on this I got bumped out of the merge queue. Since this needs be timed with a dataset delete operation I'm going to try to land this tomorrow instead.

PR #6401, opened 2024-10-25
Speed up view deploys by using processing pool

scholtzan:
## Description

This speeds up view deploys by around 3x (in some of my local tests). Looking at the `cProfile` profile using `ThreadPool` here comes with some significant overhead.

### DIFF
--- a/bigquery_etl/cli/view.py
+++ b/bigquery_etl/cli/view.py
@@ -5,6 +5,7 @@ import re
 import string
 import sys
 from fnmatch import fnmatchcase
+from functools import partial
 from graphlib import TopologicalSorter
 from multiprocessing.pool import Pool, ThreadPool
 from traceback import print_exc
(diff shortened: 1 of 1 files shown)


scholtzan:
We shouldn't merge this before next week, just in case

scholtzan:
I'll open a follow-up PR to do something similar for `./bqetl view clean`, which also relies on `ThreadPool`

BenWu (approved):
😮

bigquery_etl/cli/view.py
```diff
@@ -207,9 +208,11 @@ def publish(
         for view in views:
             view.labels["managed"] = ""
     if not force:
+        has_changes = partial(_view_has_changes, target_project)
+
         # only views with changes
-        with ThreadPool(parallelism) as p:
-            changes = p.map(lambda v: v.has_changes(target_project), views, chunksize=1)
+        with Pool(parallelism) as p:
+            changes = p.map(has_changes, views)
```
sean-rose, line 215: I wonder if removing `chunksize=1` may have also contributed to the speed-up, as it will now preemptively divide up the list between the workers rather than giving each worker an item at a time.

If so, we may want to look for other cases where `chunksize=1` is being specified but the order the work is done in doesn't really matter, and try removing `chunksize=1` there as well.

scholtzan: It seems to make a small difference. It was a little bit (a few seconds) faster when testing it out locally.

PR #6418, opened 2024-10-29
Speed up schema updates

scholtzan:
## Description

This makes some optimizations to speed up table deploys:
- build the table dependency graph in parallel 
- reuse GCP client credentials
- use the GCP API to get table schemas when not using the cloud function, instead of doing a dry run 

These improvements shave off a few minutes for `./bqetl schema update` runs (around 10-15% faster in local runs). The main bottleneck for schema updates are the query dry runs though.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-6011)

### DIFF
--- a/bigquery_etl/schema/__init__.py
+++ b/bigquery_etl/schema/__init__.py
@@ -60,23 +60,37 @@ class Schema:
     @classmethod
     def for_table(cls, project, dataset, table, partitioned_by=None, *args, **kwargs):
         """Get the schema for a BigQuery table."""
-        query = f"SELECT * FROM `{project}.{dataset}.{table}`"
-
-        if partitioned_by:
-            query += f" WHERE DATE(`{partitioned_by}`) = DATE('2020-01-01')"
-
         try:
-            return cls(
-                dryrun.DryRun(
-                    os.path.join(project, dataset, table, "query.sql"),
-                    query,
-                    project=project,
-                    dataset=dataset,
-                    table=table,
-                    *args,
-                    **kwargs,
-                ).get_schema()
-            )
+            print(kwargs["use_cloud_function"])
+            if (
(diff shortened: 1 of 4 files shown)


kwindau (approved):

bigquery_etl/schema/__init__.py
```diff
@@ -60,30 +60,47 @@ def from_json(cls, json_schema):
     @classmethod
     def for_table(cls, project, dataset, table, partitioned_by=None, *args, **kwargs):
         """Get the schema for a BigQuery table."""
-        query = f"SELECT * FROM `{project}.{dataset}.{table}`"
+        try:
+            print(kwargs["use_cloud_function"])
```
sean-rose, line 64: Print statement left over from debugging?

bigquery_etl/schema/__init__.py
```diff
@@ -60,30 +60,47 @@ def from_json(cls, json_schema):
...
+            ):
+                if "credentials" in kwargs:
+                    client = bigquery.Client(credentials=kwargs["credentials"])
+                else:
+                    client = bigquery.Client()
 
-        if partitioned_by:
-            query += f" WHERE DATE(`{partitioned_by}`) = DATE('2020-01-01')"
+                table = client.get_table(f"{project}.{dataset}.{table}")
+                return cls({"fields": [field.to_api_repr() for field in table.schema]})
```
sean-rose, line 75: Would this end up ignoring schema updates when an ETL is being updated in a way that would/should change the schema?

scholtzan: This should still detect these changes. Instead of doing a `SELECT * FROM table` to get the table schema, it is slightly faster to just use the API to get the table schema

Issue #3665, opened 2023-03-15
deploy-changes-to-stage CI step cannot handle materialized views


scholtzan:
If any files that reference materialized views get changed, then the `deploy-changes-to-stage` CI step currently fails as it cannot deploy the view as a table, or at all. An example of where this happened: https://github.com/mozilla/bigquery-etl/pull/3623

https://circleci.com/gh/mozilla/bigquery-etl/189313

BenWu:
Should be fixed by https://github.com/mozilla/bigquery-etl/pull/6421

PR #6421, opened 2024-10-30
Fix materialized views failing stage deploy

BenWu:
## Description

Stage deploys fail when there is a view that depends on a materialized view ([example](https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/40746/workflows/86591783-5539-48ec-add6-354c1769809c/jobs/466462)) because the view dependency code doesn't check for materialized views so it creates an empty query.py as a placeholder.  Deploy then fails because it tries to deploy the query.py and the materialized view.  Fix is to just add materialized views to the list of files to check for.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-6024)


bigquery_etl/cli/stage.py
```diff
@@ -106,8 +106,13 @@ def deploy(
...
         tmp_dir.mkdir(parents=True, exist_ok=True)
-        shutil.copytree(sql_dir, tmp_dir, dirs_exist_ok=True)
-        sql_dir = tmp_dir / Path(sql_dir).name
+        new_sql_dir = tmp_dir / Path(sql_dir).name
+        shutil.copytree(sql_dir, new_sql_dir, dirs_exist_ok=True)
+
+        # rename paths to tmp_dir
+        paths = [path.replace(sql_dir, f"{new_sql_dir}/", 1) for path in paths]
+
+        sql_dir = new_sql_dir
```
BenWu, line 115: This is a fix for local usage.  The materialized view renaming below was happening on the ones in `sql/` instead of the tmp dir

scholtzan (approved):
This fixes https://github.com/mozilla/bigquery-etl/issues/3665

PR #6424, opened 2024-10-30
Use INFORMATION_SCHEMA to get existing views for cleaning

scholtzan:
## Description

For `./bqetl view clean` a significant amount of time is needed to list all available views. This can be sped up by querying against `INFORMATION_SCHEMA` to get existing views.

As an example, for `moz-fx-data-experiments` this process currently takes around 20minutes (as there are a ton of tables). This change brings it down to a few seconds

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-6067)

### DIFF
--- a/bigquery_etl/cli/view.py
+++ b/bigquery_etl/cli/view.py
@@ -361,14 +361,16 @@ def clean(
 
     with ThreadPool(parallelism) as p:
         managed_view_ids = {
-            sql_table_id(view)
+            view
             for views in p.starmap(
                 client_q.with_client,
-                ((_list_managed_views, dataset, name) for dataset in datasets),
+                (
+                    (_list_managed_views, dataset, name, skip_authorized)
+                    for dataset in datasets
+                ),
                 chunksize=1,
             )
             for view in views
-            if not skip_authorized or "authorized" not in view.labels
         }
 
         remove_view_ids = sorted(managed_view_ids - expected_view_ids)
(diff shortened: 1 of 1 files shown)


bigquery_etl/cli/view.py
```diff
@@ -379,12 +380,29 @@ def clean(
...
+    # running a query against information schema instead of using the API to list tables is much faster
+    job = client.query(query)
+    result = list(job.result())
     return [
-        table
-        for table in client.list_tables(dataset)
-        if table.table_type == "VIEW"
-        and "managed" in table.labels
-        and (pattern is None or fnmatchcase(sql_table_id(table), f"*{pattern}"))
+        row.table_id
```
BenWu, line 403: Doesn't this change the return type from a table object(?) to a string?  Does the caller need to be updated?

scholtzan: mh, right

bigquery_etl/cli/view.py
```diff
@@ -378,13 +381,32 @@ def clean(
...
+        for row in result
+        if (pattern is None or fnmatchcase(sql_table_id(row.table_id), f"*{pattern}"))
+        and (not skip_authorized or not row.is_authorized)
```
scholtzan, line 409: moving `skip_authorized` into this function, so that we can just return the view IDs

BenWu (approved):

PR #6425, opened 2024-10-30
Switch to baseline beginning Aug 01st 2024

alekhyamoz:
With the recent KPI DAU changes and the need to align the search DAU with KPI.
We will be changing the SQL to pull the search related metrics from baseline tables instead of metrics.
The planned cut-off is Aug 1st 2024.

As part of this effort, we need to -
Before Aug 1st 2024 (Implemented as part of - https://github.com/mozilla/bigquery-etl/pull/5977)
**[Done]**create a v2 table to pull from the baseline table
[Done] monitor and validate v1 and v2 tables for ~1 week
[Done] create a prehistory table mobile_clients_daily_historical_pre202408 with all historical data from metrics table
[Done]Create the view to pull from pre2024 and V2

This PR handles
Changemobile_aggregates to pull from both history and v2
Change other downstream tables to pull from the new view mobile_search_clients_daily
cut off populating v1

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-6092)


bigquery_etl/shredder/config.py
```diff
@@ -176,6 +176,9 @@ def fields(self) -> tuple[str, ...]:
     client_id_target(
         table="search_derived.mobile_search_clients_daily_v1"
     ): DESKTOP_SRC,
+    client_id_target(
+        table="search_derived.mobile_search_clients_daily_v2"
+    ): DESKTOP_SRC,
```
BenWu, line 181: This should be using mobile sources instead of desktop.  I have an example in this draft https://github.com/mozilla/bigquery-etl/pull/5914/files.

I can fix this after this gets merged if that works for you

PR #6463, opened 2024-11-08
Initiate backfill of `subscription_platform_derived.monthly_active_logical_subscriptions_v1`

sean-rose:
## Description
To propagate the bug fix from #6458 to historical data.

This backfill shouldn't be initiated until the `daily_active_logical_subscriptions_v1` backfill started in #6462 has been completed.

## Related Tickets & Documents
* https://github.com/mozilla/bigquery-etl/pull/6458
* https://github.com/mozilla/bigquery-etl/pull/6462
* DENG-974: Stripe subscriptions ETL v2

**Reviewer, please follow [this checklist](https://github.com/mozilla/bigquery-etl/blob/main/.github/reviewer_checklist.md)**

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-6332)

PR #6489, opened 2024-11-13
Fix the `monthly_active_logical_subscriptions_v1` ETL to correctly calculate `was_active_at_month_end` when a subscription ends exactly at midnight

sean-rose:
## Description
The `monthly_active_logical_subscriptions_v1` ETL has been relying on the latest `daily_active_logical_subscriptions_v1.was_active_at_day_end` field value for each subscription in the month to calculate its own `was_active_at_month_end` field, but that wouldn't work correctly if the subscription ended at midnight on a day that isn't the first of the month.  That's because when a subscription ends exactly at midnight there will be no `daily_active_logical_subscriptions_v1` record for it that day (since it wasn't active at any point during that day), and its `daily_active_logical_subscriptions_v1` record for the preceding day will have `was_active_at_day_end = TRUE` because it hadn't yet ended by the last instant of that day.

This PR fixes that problem by changing the `monthly_active_logical_subscriptions_v1` ETL to use similar logic as the [`daily_active_logical_subscriptions_v1` ETL](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/subscription_platform_derived/daily_active_logical_subscriptions_v1/query.sql) and select directly from the underlying `logical_subscriptions_history_v1` table rather than from `daily_active_logical_subscriptions_v1` (which will have an added benefit of making it easier to backfill `monthly_active_logical_subscriptions_v1` since we won't need to wait for a preceding backfill of `daily_active_logical_subscriptions_v1` to complete first).

* DENG-974: Stripe subscriptions ETL v2

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-6385)

### DIFF
--- a/sql/moz-fx-data-shared-prod/subscription_platform_derived/monthly_active_logical_subscriptions_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/subscription_platform_derived/monthly_active_logical_subscriptions_v1/query.sql
@@ -2,7 +2,8 @@ WITH months AS (
   {% if is_init() %}
     SELECT
       month_start_date,
-      LAST_DAY(month_start_date, MONTH) AS month_end_date
+      LAST_DAY(month_start_date, MONTH) AS month_end_date,
+      (LAST_DAY(month_start_date, MONTH) + 1) AS next_month_start_date
     FROM
       UNNEST(
         GENERATE_DATE_ARRAY(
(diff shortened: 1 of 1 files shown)


scholtzan (approved):

sean-rose:
Luckily this bug has only adversely affected a single `monthly_active_logical_subscriptions_v1` record in March 2021.

I'm going to use the `monthly_active_logical_subscriptions_v1` backfill being triggered in #6463 to fix that record as well.

PR #6509, opened 2024-11-15
Set retention to 25 months on firefox_ios_derived.active_users_aggreg…

alekhyamoz:
This is a test PR for setting the retention policy to 25 months. 
The table `firefox_ios_derived.active_users_aggregatess_v1` is not in use and so is a good test case.

The mininum submission_date on this table now is `2021-01-01`


kwindau (approved):
I noticed that the table is partitioned by submission date in BQ but it doesn't say that in metadata.yaml? Not sure if that would be an issue or not

whd:
@alekhyamoz pretty sure per https://github.com/mozilla/bigquery-etl/pull/6522/files that the key here is wrong and that it needs to be nested under `expiration_days`

PR #6521, opened 2024-11-18
Support results with no rows in matches_pattern check

BenWu:
## Description

The check divides by zero if there are no rows matching the where clause which I don't think is the right behaviour when we have the `min_row_count` check that should be used for that case.  This changes it so the the fail rate is 0 when there are no rows

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-6427)

### DIFF
--- a/tests/checks/matches_pattern.jinja
+++ b/tests/checks/matches_pattern.jinja
@@ -4,7 +4,10 @@
 
   SELECT
     IF(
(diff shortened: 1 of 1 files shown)


scholtzan (approved):

BenWu:
There was an extra ) so that was causing sql generation to fail in the merge queue.  Rerunning CI now with `trigger-sql-generation: true` to test

PR #6529, opened 2024-11-19
Support for deploying queries consisting of only metadata.yaml and schema.yaml

scholtzan:
## Description

This adds support for deploying queries that do not have a query file. We have more instances where tables get populated outside of bigquery-etl (e.g. through outerbounds), but still get deployed through bigquery-etl. The current workaround is to add an empty `query.py` file. This PR removes the requirement for that.

One such example where this happened is: https://github.com/mozilla/private-bigquery-etl/pull/487

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-6438)

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -79,6 +79,8 @@ DESTINATION_TABLE_RE = re.compile(r"^[a-zA-Z0-9_$]{0,1024}$")
 DEFAULT_DAG_NAME = "bqetl_default"
 DEFAULT_INIT_PARALLELISM = 10
 DEFAULT_CHECKS_FILE_NAME = "checks.sql"
+VIEW_FILE = "view.sql"
+MATERIALIZED_VIEW = "materialized_view.sql"
 
 
 @click.group(help="Commands for managing queries.")
@@ -2217,6 +2219,8 @@ def deploy(
         metadata_file
         for metadata_file in metadata_files
         if metadata_file.parent not in query_file_paths
+        and not (metadata_file.parent / VIEW_FILE).exists()
+        and not (metadata_file.parent / MATERIALIZED_VIEW).exists()
     ]
 
     _deploy = partial(
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -2189,7 +2189,11 @@ def deploy(
         name, sql_dir, project_id, ["query.*", "script.sql"]
     )
 
-    if not query_files:
+    metadata_files = paths_matching_name_pattern(
+        name, sql_dir, project_id, ["metadata.yaml"]
+    )
+
+    if not query_files and not metadata_files:
         # run SQL generators if no matching query has been found
         ctx.invoke(
             generate_all,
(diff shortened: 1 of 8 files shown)


kwindau (approved):

bigquery_etl/cli/query.py
```diff
@@ -2220,23 +2234,23 @@ def deploy(
     failed_deploys, skipped_deploys = [], []
     with concurrent.futures.ThreadPoolExecutor(max_workers=parallelism) as executor:
         future_to_query = {
-            executor.submit(_deploy, query_file): query_file
-            for query_file in query_files
-            if str(query_file)
+            executor.submit(_deploy, artifact_file): artifact_file
```
BenWu, line 2237: It would make sense to also rename the argument in the `deploy_table` function

bigquery_etl/cli/query.py
```diff
@@ -2199,12 +2203,22 @@ def deploy(
...
 
     credentials = get_credentials()
     id_token = get_id_token(credentials=credentials)
 
+    query_file_paths = [query_file.parent for query_file in query_files]
+    metadata_files_without_query_file = [
+        metadata_file
+        for metadata_file in metadata_files
+        if metadata_file.parent not in query_file_paths
+    ]
```
BenWu, line 2220: Wouldn't this also pick up the metadata files of `view.sql` and then deploy them as tables if they have schemas?

scholtzan: mh yeah, I guess for newly added views that don't already exist that would happen...

scholtzan: Added a filter to filter out metadata files for which views and materialized views exist

bigquery_etl/cli/query.py
```diff
@@ -2199,12 +2205,24 @@ def deploy(
         query_files = paths_matching_name_pattern(
             name, ctx.obj["TMP_DIR"], project_id, ["query.*", "script.sql"]
         )
-        if not query_files:
+        metadata_files = paths_matching_name_pattern(
+            name, ctx.obj["TMP_DIR"], project_id, ["metadata.yaml"]
+        )
+        if not query_files and not metadata_files:
```
lucia-vargas-a, line 2211: I like that metadata is the minimum required, so that only `schema.yaml` would make the check fail.

bigquery_etl/cli/query.py
```diff
@@ -2199,12 +2205,24 @@ def deploy(
...
     credentials = get_credentials()
     id_token = get_id_token(credentials=credentials)
 
+    query_file_paths = [query_file.parent for query_file in query_files]
+    metadata_files_without_query_file = [
+        metadata_file
+        for metadata_file in metadata_files
+        if metadata_file.parent not in query_file_paths
+        and not (metadata_file.parent / VIEW_FILE).exists()
+        and not (metadata_file.parent / MATERIALIZED_VIEW).exists()
```
BenWu, line 2223: Might need to handle udfs as well otherwise they'll raise a lot of `SkippedDeployException`

BenWu: Also `stored_procedure.sql`.  Maybe it's better to explicitly look for all metadata with no sql or query.py

BenWu (approved):

PR #6552, opened 2024-11-22
feat: remove the condition resulting in exception if no metric_type value is defined

kik-kik:
# feat: remove the condition resulting in exception if no metric_type value is defined

## Description

This condition made a wrong assumption that the `metric_type` must always be a not null value after parsing the config. This is incorrect, for example when `saved_metric_id` is referenced in the config the object created will have `metric_type` value of `None`.


alekhyamoz (approved):
Tested locally and seemed to have worked. Thanks Kik for putting in the fix.

PR #6558, opened 2024-11-25
Handle external_data table deploys mostly the same as for regular tables

curtismorales:
## Description

Fixes a bug introduced in #6529 -- this PR added support for tables that are defined with a `metadata.yaml` but no `query.sql`, but didn't check for tables that are built on external data (Google Sheets or CSV)

This adds logic to the `deploy_table` function to handle external tables differently. In the process I also did some refactoring of how we deploy external tables. I moved the metadata update logic that was in `_deploy_external_data` into an `attach_external_data_config` function in `publish_metadata.py` and removed the duplicated code.

Is this reasonable? This also cuts one of the passes we're doing through all the sql and metadata files during a deploy, so I think it's a performance benefit. Unless there was a reason that we weren't deploying external tables in the same threads as the rest of the table deploys?

Also, I should add some tests; this bug is kind of my fault for not adding tests in the first place. If it's okay with you, I'd like to do that in a separate PR.

* [AD-501](https://mozilla-hub.atlassian.net/browse/AD-501) (where this bug came up)

[AD-501]: https://mozilla-hub.atlassian.net/browse/AD-501?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-6564)


bigquery_etl/deploy.py
```diff
@@ -40,12 +45,24 @@ def deploy_table(
...
 
+    metadata = None
     try:
         if artifact_file.name == METADATA_FILE:
             metadata = Metadata.from_file(artifact_file)
         else:
             metadata = Metadata.of_query_file(artifact_file)
 
+        if metadata.external_data:
+            if str(artifact_file).endswith("query.sql"):
```
scholtzan, line 56: Should we check for other types of query files, like `query.py` here as well. Maybe similar to https://github.com/mozilla/bigquery-etl/blob/5d216fa2dcbb5674f1e42ad4ebeaf7a5369edefa/bigquery_etl/cli/query.py#L2222-L2226

scholtzan (approved):

scholtzan:
Having some more unit tests for this would be nice, but it's fine to do that in a follow-up PR.

PR #6566, opened 2024-11-26
Initial commit, work in progress

kwindau:
## Description

This PR adds a new boolean option to the backfill CLI command, which is to prevent accidental deletion of data when running backfills starting more than a set number of days back.

* DENG-XXXX
* DSRE-XXXX

### DIFF
--- a/bigquery_etl/cli/backfill.py
+++ b/bigquery_etl/cli/backfill.py
@@ -122,6 +122,13 @@ def backfill(ctx):
     "--shredder_mitigation/--no_shredder_mitigation",
     help="Wether to run a backfill using an auto-generated query that mitigates shredder effect.",
 )
+@click.option(
+    "--override-retention-range-limit",
+    required=True,
+    type=bool,
+    help="True to allow running a backfill outside the retention policy limit.",
+    default=False,
+)
 # If not specified, the billing project will be set to the default billing project when the backfill is initiated.
 @billing_project_option()
 @click.pass_context
@@ -135,6 +142,7 @@ def create(
     watcher,
     custom_query_path,
     shredder_mitigation,
+    override_retention_range_limit,
     billing_project,
 ):
     """CLI command for creating a new backfill entry in backfill.yaml file.
@@ -163,6 +171,7 @@ def create(
         status=BackfillStatus.INITIATE,
         custom_query_path=custom_query_path,
         shredder_mitigation=shredder_mitigation,
+        override_retention_limit=override_retention_range_limit,
         billing_project=billing_project,
     )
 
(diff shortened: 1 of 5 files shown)


bigquery_etl/cli/query.py
```diff
@@ -81,6 +81,7 @@
 DEFAULT_CHECKS_FILE_NAME = "checks.sql"
 VIEW_FILE = "view.sql"
 MATERIALIZED_VIEW = "materialized_view.sql"
+NBR_DAYS_RETAINED = 775
```
scholtzan, line 84: I guess this is a good start. Though I wonder if we can use the `expiration_days` metadata to determine how far back the backfill should go. The retention can be different for different datasets, like here: https://github.com/mozilla/bigquery-etl/blob/7cdeef94d497dbcde97bccf0078f4ab0cba9da00/sql/moz-fx-data-shared-prod/braze_external/changed_products_sync_v1/metadata.yaml#L18

Ideally, the smallest retention range based on all upstream tables should be used here.

bigquery_etl/cli/backfill.py
```diff
@@ -122,6 +122,13 @@ def backfill(ctx):
     "--shredder_mitigation/--no_shredder_mitigation",
     help="Wether to run a backfill using an auto-generated query that mitigates shredder effect.",
 )
+@click.option(
+    "--override-retention-range-limit",
```
scholtzan, line 126:
```suggestion
    "--override-retention-range-limit",
    "--override_retention_range_limit",
```

bigquery_etl/cli/backfill.py
```diff
@@ -122,6 +122,13 @@ def backfill(ctx):
...
+    required=True,
+    type=bool,
+    help="True to allow running a backfill outside the retention policy limit.",
+    default=False,
+)
```
scholtzan, line 131: I guess this could also be a flag. `is_flag=True`

Then just `--override-retention-range-limit` would suffice to override the range.

bigquery_etl/cli/query.py
```diff
@@ -658,6 +659,13 @@ def _backfill_query(
         "parameters and/or date_partition_parameter as needed."
     ),
 )
+@click.option(
+    "--override-retention-range-limit",
+    required=True,
+    type=bool,
+    help="True to allow running a backfill outside the retention policy limit.",
+    default=False,
+)
```
scholtzan, line 668: same suggestions as above

bigquery_etl/cli/backfill.py
kwindau: Just pushed this - thanks for the suggestion!

bigquery_etl/cli/query.py
kwindau: I agree this would be a really good iteration idea - would be much more accurate and prevent more issues.

scholtzan (approved):

lucia-vargas-a (commented):
LGTM

PR #6569, opened 2024-11-27
chore: update the report_title to use the full report name 

kik-kik:
# chore: update the report_title to use the full report name 

This is to ensure that we do not accidentally match any future choice screen related reports for newer ios versions


alekhyamoz (approved):
lgtm

PR #6571, opened 2024-11-27
feat: include bigconfig.yml from the project directory containing generic monitor definitions

kik-kik:
# feat: include bigconfig.yml from the project directory containing generic monitor definitions

This is to allow usage of saved_metric_ids defined inside `sql/moz-fx-data-share-prod/bigconfig.yml`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-6615)


bigquery_etl/cli/monitoring.py
```diff
@@ -181,7 +181,8 @@ def deploy(
             input_path=[
                 metadata_file.parent / BIGCONFIG_FILE
                 for metadata_file in list(set(metadata_files))
-            ],
+            ]
+            + [Path(f"{sql_dir}/{project_id}/{BIGCONFIG_FILE}")],
```
scholtzan, line 185: We should probably check if this file exists, since not all projects might have it.

kik-kik: moved into sql_dir, should always exist there

scholtzan (approved):

PR #6572, opened 2024-11-27
[DENG-6141] Add syndicate configuration and public views for Service Desk Jira data

scholtzan:
## Description

https://mozilla-hub.atlassian.net/browse/DENG-6141

This adds a configuration for syndicating the jira service desk data that is imported via Fivetran. At this moment the data is coming from a Jira sandbox project. Once it has been validated, we'll switch over to the production instance and put some workgroups in place to restrict access.


kwindau (approved):
LGTM

whd:
Looks like this just landed. Reviewing the config here it seems like it would be simpler to syndicate to a user-facing dataset directly, since the set of user-facing and syndicated views are the same. This config makes sense if the  intent is for `jira_service_desk` to include other non-syndicated views eventually  since mixed management isn't supported yet (see https://mozilla-hub.atlassian.net/browse/DSRE-1662). Otherwise I'd recommend simplifying this.

scholtzan:
The service desk team wants to join some of these tables. It would probably make sense to just do that in a separate view. Or alternatively in Looker directly, in which case this config could be simplified. I'll check with them to see what's preferred

PR #6582, opened 2024-11-29
CI check to make sure that tables that are exempted from retention are not set to expire

alekhyamoz:
As part of the data retention project, certain tables were identified to be exempt from retention. 
This PR 
1. Add a section `retention_exclusion_list` to  the config file `bigconfig.yaml`. This is list of files that should be excluded from setting the retention policy.
2. adds a function `validate_exclusion_list_expiration_days` to fail the metadata validation if the table's metadata contains `expiration_days` set

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-6697)


kwindau (approved):
Looks good to me, thanks for doing this!

scholtzan (commented):
lgtm

PR #6586, opened 2024-12-02
Reschedule bqetl_search_dashboard DAG

akkomar:
bqetl_glean_usage which is a dependency now runs longer than before. This will make search_dashboard start after glean_usage usually finishes.

### DIFF
--- a/dags.yaml
+++ b/dags.yaml
@@ -497,7 +497,7 @@ bqetl_search_dashboard:
     retries: 2
     retry_delay: 30m
     start_date: "2020-12-14"
-  schedule_interval: 30 4 * * *
+  schedule_interval: 30 5 * * *
   tags:
     - impact/tier_2
 


dags.yaml
```diff
@@ -497,7 +497,7 @@ bqetl_search_dashboard:
     retries: 2
     retry_delay: 30m
     start_date: "2020-12-14"
-  schedule_interval: 30 4 * * *
+  schedule_interval: 30 5 * * *
```
kik-kik, line 500: Just to double check, the execution_delta in the generated version of this DAG will get automatically updated to match the upstream DAGs execution date?

kik-kik (approved):
Just one question. Otherwise, LGTM.

dags.yaml
akkomar: Yes

PR #6588, opened 2024-12-02
DENG-877 enforce backfills with mitigation

lucia-vargas-a:
## Description

This PR:
- Updates classes to use attrs instead of attr.
- Stops a managed backfill if the shredder_mitigation option is not used for  tables with shredder mitigation label in the metadata.

* DENG-877

### DIFF
--- a/bigquery_etl/backfill/shredder_mitigation.py
+++ b/bigquery_etl/backfill/shredder_mitigation.py
@@ -11,7 +11,7 @@ from pathlib import Path
 from types import NoneType
 from typing import Any, Optional, Tuple
 
-import attr
+import attrs
 import click
 from dateutil import parser
 from gcloud.exceptions import NotFound  # type: ignore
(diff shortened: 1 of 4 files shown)


bigquery_etl/backfill/shredder_mitigation.py
```diff
@@ -83,44 +83,35 @@ def from_schema_type(schema_type):
...
 class Subset:
     """Representation of a subset/CTEs in the query and the actions related to this subset."""
 
     client: bigquery.Client
-    destination_table: str = attr.field(default="")
-    query_cte: str = attr.field(default="")
-    dataset: str = attr.field(default=TEMP_DATASET)
-    project_id: str = attr.field(default=DEFAULT_PROJECT_ID)
-    expiration_days: Optional[float] = attr.field(default=None)
+    destination_table: str = attrs.field(default="")
```
kik-kik, line 110: Should default be specified here? Seems weird to set it to an empty string rather than throwing an error if it is required to execute the logic?

kwindau (approved):

bigquery_etl/backfill/shredder_mitigation.py
lucia-vargas-a: Good call! Added a validator and test.

alekhyamoz (approved):
Thanks for working on this Lucia 🥇

PR #6612, opened 2024-12-05
fix: Temporarily remove bigeye checks on android_app_campaign_stats_v1 & ios_app_campaign_stats_v1 until we can update with 28 day lookback

kwindau:
This PR removes the Bigeye checks for the table `android_app_campaign_stats_v1` & `ios_app_campaign_stats_v1` since they don't work out of the box because this table has a 28 day lag, so we need to write a new bigeye config YAML file to make this check work properly (otherwise it is failing the DAG in airflow). For now removing until we have the time to write this new YAML file - made a ticket so we don't lose track of adding it back - https://mozilla-hub.atlassian.net/browse/DENG-6886

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-6885)

### DIFF
--- a/sql/moz-fx-data-shared-prod/google_ads_derived/android_app_campaign_stats_v1/checks.sql
+++ b/sql/moz-fx-data-shared-prod/google_ads_derived/android_app_campaign_stats_v1/checks.sql
@@ -1,2 +1,2 @@
 #fail
-{{ is_unique(["date", "campaign", "ad_group"], "date = DATE_SUB(@submission_date, INTERVAL 27 DAY)") }}
+{{ is_unique(["date", "campaign", "ad_group"]) }}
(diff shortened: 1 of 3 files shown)


Marlene-M-Hirose (approved):

alekhyamoz (approved):

sql/moz-fx-data-shared-prod/google_ads_derived/android_app_campaign_stats_v1/checks.sql
```diff
@@ -1,2 +1,2 @@
 #fail
-{{ is_unique(["date", "campaign", "ad_group"], "date = DATE_SUB(@submission_date, INTERVAL 27 DAY)") }}
+{{ is_unique(["date", "campaign", "ad_group"]) }}
```
kik-kik, line 2: @kwindau considering the query is running with a 27 day lag, this specific filter that was removed here was actually required. Right now this will most likely target partitions that do not exist.

kik-kik, line 2: Apologies, I just thought about it again and this is fine. It just means that the check will be executed against the whole table and not just a single partition. In this case I believe the table is quite small so it doesn't really matter.

PR #6620, opened 2024-12-06
Create new dataset ltv_state_values_derived

kwindau:
## Description

This PR is intended to create a new dataset in shared-prod where the data science team has write access.
In discussion with Leif, learned that they may be soon replacing these models with new models, so not worth the effort to productionalize their existing models, but rather move them to a safe place in shared-prod until the new ones are ready to replace the old ones.

We will be moving 3 tables from analysis into this dataset since they are used downstream in other production ETL for the Fenix, Firefox Desktop, and iOS LTV models.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-6915)

### DIFF
--- a/sql/moz-fx-data-shared-prod/ltv_state_values_derived/dataset_metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/ltv_state_values_derived/dataset_metadata.yaml
@@ -1,0 +1,14 @@
+friendly_name: LTV State Values Derived
+description: |
+  Dataset where the data science team has write access to write LTV state values from their models
+dataset_base_acl: derived_restricted
+user_facing: false
+labels: {}
+workgroup_access:
+  - role: roles/bigquery.dataViewer
+    members:
+      - workgroup:datascience/developers
+      - workgroup:mozilla-confidential
+  - role: roles/bigquery.dataEditor
+    members:
+      - workgroup:datascience/developers
--- a/sql/moz-fx-data-shared-prod/ltv_state_values_derived/dataset_metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/ltv_state_values_derived/dataset_metadata.yaml
@@ -7,7 +7,6 @@ labels: {}
 workgroup_access:
   - role: roles/bigquery.dataViewer
     members:
-      - workgroup:datascience/developers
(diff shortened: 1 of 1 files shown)


sql/moz-fx-data-shared-prod/ltv_state_values_derived/dataset_metadata.yaml
```diff
@@ -0,0 +1,14 @@
+friendly_name: LTV State Values Derived
+description: |
+  Dataset where the data science team has write access to write LTV state values from their models
+dataset_base_acl: derived_restricted
```
whd, line 4: The configuration you've written would create a dataset that DE doesn't have write access to: only DS would. Assuming we trust DE not to accidentally delete the data, and since the data is presumably not sensitive (it was in the `analysis` dataset which is `workgroup:mozilla-confidential`) the standard archetype + DE write access would be implemented by:
```suggestion
dataset_base_acl: derived
```

sql/moz-fx-data-shared-prod/ltv_state_values_derived/dataset_metadata.yaml
```diff
@@ -0,0 +1,14 @@
...
+user_facing: false
+labels: {}
+workgroup_access:
+  - role: roles/bigquery.dataViewer
+    members:
+      - workgroup:datascience/developers
```
whd, line 10: This is redundant with the write access granted and the read access granted to `workgroup:mozilla-confidential` which is a superset of `workgroup:datascience/developers`.
```suggestion
```

whd (commented):
My braze example wasn't the best because it's a workgroup-confidential dataset. We probably don't have any production datasets with that are both non-sensitive and that some specific workgroup outside DE has write access to. Anywhile your PR should work, but here's the difference the between your config and the suggested config:

Your version: restricted write access to DS, open read access
Recommended version: DE+DS write access, open read access

I do notice some ltv datasets are workgroup-confidential: on the off chance this data is sensitive and should not be `workgroup:mozilla-confidential`, you should use restricted variants and remove grants to `workgroup:mozilla-confidential` to implement that.

whd (approved):
LGTM

PR #6625, opened 2024-12-06
feat: Add minimum row count check to google ads table google_ads_derived.daily_campaign_stats_v1

kwindau:
## Description

This table is a full delete and rebuild each day from Google Ads.  When the connection breaks, the DAG doesn't throw an error, it still marks the run as successful.  This logic change should ensure that if no data is loaded for the current submission date, an error will be thrown in the future.

### DIFF
--- a/sql/moz-fx-data-shared-prod/google_ads_derived/daily_campaign_stats_v1/checks.sql
+++ b/sql/moz-fx-data-shared-prod/google_ads_derived/daily_campaign_stats_v1/checks.sql
@@ -1,2 +1,2 @@
 #fail
-{{ min_row_count(1, "date = @submission_date") }}
+{{ min_row_count(1, "`date` = @submission_date") }}
--- a/sql/moz-fx-data-shared-prod/google_ads_derived/daily_campaign_stats_v1/checks.sql
+++ b/sql/moz-fx-data-shared-prod/google_ads_derived/daily_campaign_stats_v1/checks.sql
@@ -1,2 +1,2 @@
-# fail
+#fail
 {{ min_row_count(1, "last_updated_date = @submission_date") }}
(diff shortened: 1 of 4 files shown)


scholtzan (approved):

sql/moz-fx-data-shared-prod/google_ads_derived/daily_campaign_stats_v1/query.sql
```diff
@@ -1,5 +1,6 @@
 SELECT
   date_day AS `date`,
-  * EXCEPT (date_day)
+  * EXCEPT (date_day),
+  @submission_date AS last_updated_date
```
sean-rose, line 4: Since the ETL is overwriting the entire table every time I don't think this `last_updated_date` column is going to work as intended (the `min_row_count` check will always pass).

I think you can just use the existing `date` column, as we'd be expecting to have gotten data for that date (and checking the `moz-fx-data-shared-prod.google_ads_derived.daily_campaign_stats_v1` table I'm even already seeing data for today, i.e. the day after `@submission_date`).

sql/moz-fx-data-shared-prod/google_ads_derived/daily_campaign_stats_v1/checks.sql
```diff
@@ -0,0 +1,2 @@
+# fail
+{{ min_row_count(1, "last_updated_date = @submission_date") }}
```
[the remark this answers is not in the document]
sean-rose: Never mind, you already caught this while I was reviewing.

sql/moz-fx-data-shared-prod/google_ads_derived/daily_campaign_stats_v1/checks.sql
```diff
@@ -0,0 +1,2 @@
+# fail
```
sean-rose, line 1: I think this needs to be `#fail` with no space (or at least all existing fail checks are written that way).
```suggestion
#fail
```

sql/moz-fx-data-shared-prod/google_ads_derived/daily_campaign_stats_v1/query.sql
kwindau: Ahh very good point, will update that - thanks Sean!

PR #6631, opened 2024-12-09
feat: add bq job labels to bigquery_usage and bigeye_usage views

kik-kik:
# feat: add bq job labels to bigquery_usage and bigeye_usage views


alekhyamoz (approved):
lgtm

PR #6633, opened 2024-12-09
DENG-6958 Apply shredder_mitigation to accounts derived tables

akkomar:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-6962)


akkomar:
On another look these are client-level tables, so shredder_mitigation is not applicable to them.

PR #6645, opened 2024-12-10
Remove materialized view append from referenced tables in jobs

BenWu:
## Description

This reverts https://github.com/mozilla/bigquery-etl/pull/4786.  The materialized view already shows up in the referenced tables so the `ARRAY_CONCAT` is adding a duplicate value which gives incorrect results if we do a cross join without deduping. [example](https://sql.telemetry.mozilla.org/queries/104377/source)

I'm guessing that this was needed because the materialized view used to not be included in the array.  @scholtzan do you remember if that was the case?


scholtzan (approved):
Yeah, the materialized views definitely did not use to show in there. I guess something changed.

whd (approved):
Nothing subject to change control here

PR #6652, opened 2024-12-11
Wait 5 minutes between task sensor retries

akkomar:
Default `poke_interval` is 60s. [Airflow docs](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/sensors/base/index.html#airflow.sensors.base.BaseSensorOperator) recommend it to be more than that in `reschedule` mode to decrease load on scheduler. In our case worker pods take up 30s to start so it seems wasteful to restart them every 60s. We also have [some issues with git-sync init on worker pods](https://mozilla-hub.atlassian.net/browse/DENG-6838) potentially related to too many pods running at the same time which this bump might help with.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7002)

### DIFF
--- a/bigquery_etl/query_scheduling/templates/airflow_dag.j2
+++ b/bigquery_etl/query_scheduling/templates/airflow_dag.j2
@@ -106,6 +106,7 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
         {% endif -%}
         check_existence=True,
         mode='reschedule',
+        poke_interval=datetime.timedelta(minutes=5),
         allowed_states=ALLOWED_STATES,
         failed_states=FAILED_STATES,
         pool='DATA_ENG_EXTERNALTASKSENSOR',
--- a/bigquery_etl/query_scheduling/templates/public_data_json_airflow_dag.j2
+++ b/bigquery_etl/query_scheduling/templates/public_data_json_airflow_dag.j2
@@ -83,6 +83,7 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
         {% endif -%}
         check_existence=True,
         mode='reschedule',
+        poke_interval=datetime.timedelta(minutes=5),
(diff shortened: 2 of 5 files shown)


scholtzan (approved):

BenWu (approved):
There's also a template for `BigQueryTablePartitionExistenceSensor` and a sensor in `public_data_json_airflow_dag.j2`.  Should those also be changed?

akkomar:
> There's also a template for `BigQueryTablePartitionExistenceSensor` and a sensor in `public_data_json_airflow_dag.j2`. Should those also be changed?

`BigQueryTablePartitionExistenceSensor` already has a default of `poke_interval=datetime.timedelta(minutes=15)`.
I'll update `public_data_json_airflow_dag.j2`.

bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -106,6 +106,7 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
         {% endif -%}
         check_existence=True,
         mode='reschedule',
+        poke_interval=datetime.timedelta(minutes=5),
```
sean-rose, line 109: This should be `poll_interval` ([docs](https://airflow.apache.org/docs/apache-airflow/2.9.3/_api/airflow/sensors/external_task/index.html#airflow.sensors.external_task.ExternalTaskSensor)):
```suggestion
        poll_interval=datetime.timedelta(minutes=5),
```
(ditto for the other `ExternalTaskSensor`s)

sean-rose: Never mind. It looks like `poll_interval` only applies when the sensors are run in deferrable mode ([code](https://github.com/apache/airflow/blob/c92929b99fc15c42ccc0799cc696417a5af76161/airflow/sensors/external_task.py#L357)).

PR #6684, opened 2024-12-17
feat(DENG-4975): add bigconfig to baseline clients daily and first seen

kik-kik:
# feat(DENG-4975): add bigconfig to baseline clients daily and first seen

### DIFF
--- a/sql/bigconfig.yml
+++ b/sql/bigconfig.yml
@@ -8,9 +8,9 @@ saved_metric_definitions:
       threshold:
         type: CONSTANT
         upper_bound: 0
-      schedule_frequency:
-        interval_type: MINUTES
-        interval_value: 0
+      metric_schedule:
+        named_schedule:
+          name: default
       lookback:
         lookback_type: DATA_TIME
         lookback_window:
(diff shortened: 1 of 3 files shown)


sql/bigconfig.yml
```diff
@@ -8,9 +8,9 @@ saved_metric_definitions:
       threshold:
         type: CONSTANT
         upper_bound: 0
-      metric_schedule:
-        named_schedule:
-          name: default
+      schedule_frequency:
+        interval_type: MINUTES
+        interval_value: 0
```
scholtzan, line 13: There is an open Bigeye bug for this. It should work in theory, but in practice it will break deploys

sql/bigconfig.yml
scholtzan, line 1: maybe just revert changes to this file for now, the other configs look good

sql/bigconfig.yml
kik-kik: Should I just revert it back to default or just remove the scheduling completely?

scholtzan: just revert it back for now. Not defining a schedule will default to a 24h schedule frequency

scholtzan (approved):

PR #6706, opened 2024-12-19
feat: add partnership field to the active_users_aggregates view template

kik-kik:
# feat: add partnership field to the active_users_aggregates view template

## Description

This change ads an additional field to the active_users_aggregates view which maps the distribution_id to a device partner to enable filtering of the KPI dashboards for this specific user segment.

This is an alternative to turning the active_users_aggregates in lookml to support this filtering. For more info on this approach see: https://github.com/mozilla/looker-spoke-default/pull/984

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry/active_users_aggregates/view.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry/active_users_aggregates/view.sql
@@ -32,7 +32,12 @@ SELECT
   app_version_minor,
   app_version_patch_revision,
   app_version_is_major_release,
-  os_grouped
+  os_grouped,
+  CASE
+    WHEN distribution_id LIKE "%vivo%"
+      THEN "vivo"
+    ELSE "other"
+  END AS partnership,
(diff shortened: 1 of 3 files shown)


lucia-vargas-a (commented):
Do you want to add partnership to `telemetry.active_users_aggregates` to reflect in Looker?

lucia-vargas-a (commented):
LGTM. @bochocki are you happy to merge this before the code freeze?

lucia-vargas-a:
@kik-kik This change may result in a slower respond due to the additional transformations. Since the KPI dashboards query `telemetry.active_users_aggregates` directly and it's end of year when these Dashboards are in high demand, I'd recommend merging after the code freeze in Jan 2. Are you happy with that?

kik-kik:
We chatted with @bochocki about this and he mentioned that after December 15th was fine to make these changes. Also, this transformation would only apply when used and even then I don't think this would cause any performance issues.

bochocki (approved):

PR #6740, opened 2024-12-31
Deng 6889 managed backfills default browser agg - set shredder mitigation false

wwyc:
## Description

Setting shredder mitigation to true is causing failure in airflow ([link](https://workflow.telemetry.mozilla.org/dags/bqetl_backfill_initiate/grid?dag_run_id=scheduled__2024-12-30T23%3A00%3A00%2B00%3A00&task_id=initiate_backfill.process_backfill&tab=mapped_tasks&map_index=0)).  Setting it to false for now until managed backfills are done.
Some of these checks for shredder mitigation should probably be in CI as well.

* DENG-6889

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7133)

### DIFF
--- a/dags.yaml
+++ b/dags.yaml
@@ -2056,22 +2056,3 @@ bqetl_default_browser_aggregates:
   schedule_interval: 0 22 * * *
   tags:
     - impact/tier_3
-
-bqetl_fx_desktop_background_defaultagent_aggregates:
-  default_args:
-    depends_on_past: false
-    email:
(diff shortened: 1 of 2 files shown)


lucia-vargas-a (approved):
This check fails because shredder mitigation is set-up to run for tables with a previous version to compare with, thus for v2 or above.
In this case, the correct process would be to set `shredder_mitigation = true` in v2.

PR #6775, opened 2025-01-08
[Bug 1940489] Logic to skip stable view generation for specific apps

scholtzan:
## Description

https://bugzilla.mozilla.org/show_bug.cgi?id=1940489

related to https://github.com/mozilla/bigquery-etl/commit/caf20b4544db4531ace5a93fc7f9b4714916e72c

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7174)

### DIFF
--- a/bqetl_project.yaml
+++ b/bqetl_project.yaml
@@ -377,6 +377,9 @@ generate:
       - moso_mastodon_backend
       - mozphab
       - mozregression
+  stable_views:
+    skip_apps:
+    - mlhackweek_search
 
 retention_exclusion_list:
 - sql/moz-fx-data-shared-prod/search_derived/acer_cohort_v1
--- a/bqetl_project.yaml
+++ b/bqetl_project.yaml
@@ -378,7 +378,7 @@ generate:
       - mozphab
       - mozregression
   stable_views:
-    skip_apps:
+    skip_datasets:
     - mlhackweek_search
 
 retention_exclusion_list:
--- a/sql_generators/stable_views/__init__.py
+++ b/sql_generators/stable_views/__init__.py
@@ -409,7 +409,7 @@ def generate(target_project, output_dir, log_level, parallelism, use_cloud_funct
         last
         for k, (*_, last) in groupby(schemas, lambda t: t.bq_dataset_family)
         if k
-        not in ConfigLoader.get("generate", "stable_views", "skip_apps", fallback=[])
+        not in ConfigLoader.get("generate", "stable_views", "skip_datasets", fallback=[])
     ]
 
     id_token = get_id_token()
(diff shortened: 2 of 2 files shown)


bqetl_project.yaml
```diff
@@ -377,6 +377,9 @@ generate:
       - moso_mastodon_backend
       - mozphab
       - mozregression
+  stable_views:
+    skip_apps:
+    - mlhackweek_search
```
scholtzan, line 382: this has resolved itself in the meantime, but keeping it here for reference

BenWu (approved):
Another thing that could be done for glean schemas is to check if the bq_dataset_family is still in https://probeinfo.telemetry.mozilla.org/v2/glean/app-listings.  That way there's one less step to remember to do

bqetl_project.yaml
```diff
@@ -377,6 +377,9 @@ generate:
       - moso_mastodon_backend
       - mozphab
       - mozregression
+  stable_views:
+    skip_apps:
```
whd, line 381: Using `app` here is a bit confusing since that's a [glean-level concept](https://github.com/mozilla/probe-scraper/?tab=readme-ov-file). The usage in code is of  [`bq_dataset_family`](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/main/schemas/metadata/metaschema/metaschema.1.schema.json#L14) which is a normalization of [namespace](https://docs.telemetry.mozilla.org/concepts/pipeline/http_edge_spec.html#postput-request). Anyway, we want this to work for both glean and non-glean applications and I don't really like `skip_bq_dataset_families` as the name here. I also don't like `skip_namespaces` since that would map to the non-normalized `mlhackweek-search`. But if possible we should have `apps` mean the same thing as the other usage in https://github.com/mozilla/bigquery-etl/blob/main/bqetl_project.yaml, which all appear to be glean apps.

whd (approved):
Apart from the slight confusing around naming this LGTM.

I hope that this combined with other changes like https://github.com/mozilla/bigquery-etl/pull/6769/files and https://github.com/mozilla/bigquery-etl/blob/main/bqetl_project.yaml#L306 is sufficient to get the next bqetl run after merge  to not include any generated metadata for the app. 

For `mlhackweek-search`, after we removed the explicit metadata, we auto-generated dataset metadata anyway via https://github.com/mozilla/private-bigquery-etl/commit/aeb1eaee3a6266628252040a24dfea15e5efc107#diff-ae9cc0a54ce066382670ff34f825a4f79a6eed6d9ae6b365e3d03a08b6feedf7. I hope this change will implicitly make it so that doesn't happen.

The only reason a bqetl PR was needed for mlhackweek-search was because it had workgroup-confidential config. For other apps though the deletion process will now always include a PR to stop stable generation as part of process. Per above I'm hoping this can be done in a single PR but it might need to be chunked.

Issue #6716, opened 2024-12-19
Continue to propagate operational access on datasets with deprecated tables in them


whd:
/CC @chelseybeck 

e.g. https://github.com/mozilla/private-bigquery-etl/commit/eaa3b64a48ce1cbd9aa792e9892e4cf3748ad77a is not the expected generated metadata.

[Slack discussion](https://mozilla.slack.com/archives/C01E8GDG80N/p1734644989351609?thread_ts=1734625002.069079&cid=C01E8GDG80N)

This only happens when (a) operational write access is provisioned via bqetl metadata (b) the dataset is workgroup-confidential (c) there's a deprecated table in the dataset. This hadn't happened until today.

The logic at https://github.com/mozilla/bigquery-etl/blob/main/bigquery_etl/cli/metadata.py#L86 needs to be more nuanced: `roles/bigquery.dataEditor` access which is assumed to be operational should be retained on a dataset grant.

We probably want to explicitly remove `roles/bigquery.dataViewer` but we could also allowlist roles instead i.e. keep only `roles/bigquery.dataEditor`. I'm not sure what the best approach is. I suppose you could ask what you'd expect to happen to `roles/bigquery.metadataViewer`  which is used occasionally as a deciding factor.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7099)

PR #6782, opened 2025-01-08
[#6716] Retain workgroup permissions

scholtzan:
## Description

See https://github.com/mozilla/bigquery-etl/issues/6716

If I understand the issue correctly, we want to retain dataset permissions of some roles when tables get deprecated. Currently, `workgroup_access` is set to `[]` in `dataset_metadata.yaml` when a table gets deprecated. 

This change adds a configuration option where we can list roles whose permissions should be retained on deprecation. I added `roles/bigquery.dataEditor` for now. 

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7186)

### DIFF
--- a/bigquery_etl/cli/metadata.py
+++ b/bigquery_etl/cli/metadata.py
@@ -57,6 +57,9 @@ def update(name: str, sql_dir: Optional[str], project_id: Optional[str]) -> None
     table_metadata_files = paths_matching_name_pattern(
         name, sql_dir, project_id=project_id, files=["metadata.yaml"]
     )
+    retained_dataset_roles = ConfigLoader.get(
+        "deprecation", "retain_dataset_roles", fallback=[]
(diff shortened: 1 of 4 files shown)


whd (approved):
> If I understand the issue correctly, we want to retain dataset permissions of some roles when tables get deprecated. 

That's correct. While it looks like there's no diff here, the test matches the behavior we're looking for. I think starting with an allowlist of retained roles makes sense. If we get fancy in the future with custom roles or whatever we can switch to a dissallowlist if that's easier to maintain.

My expectation is that this would generate a diff obviating https://github.com/mozilla/private-bigquery-etl/pull/529 and allowing us to revert https://github.com/mozilla/private-bigquery-etl/pull/519, but those are both in the private repo.

PR #6783, opened 2025-01-09
feat(DENG-6890): Update mobile kpi support metric aggregates to support dma analysis

kik-kik:
# feat(DENG-6890): Update mobile kpi support metric aggregates to support dma analysis

## Description

This change aims to add the following two fields to engagement, retention and new_profiles KPI support metric aggregated datasets to support DMA related analysis needed.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7188)

### DIFF
--- a/sql_generators/mobile_kpi_support_metrics/templates/active_users.view.sql
+++ b/sql_generators/mobile_kpi_support_metrics/templates/active_users.view.sql
@@ -3,7 +3,10 @@ CREATE OR REPLACE VIEW
   `{{ project_id }}.{{ dataset }}.{{ name }}`
 AS
 SELECT
-  * EXCEPT (isp),
+  * EXCEPT (isp) REPLACE(
+    -- Lower device_manufacturer as in some cases the same manufacturer value has different casing.
+    LOWER(device_manufacturer) AS device_manufacturer
+  ),
   CASE
     WHEN LOWER(isp) = "browserstack"
       THEN CONCAT("{{ friendly_name }}", " ", isp)
--- a/sql_generators/mobile_kpi_support_metrics/templates/active_users.view.sql
+++ b/sql_generators/mobile_kpi_support_metrics/templates/active_users.view.sql
@@ -47,5 +47,15 @@ SELECT
   -- Adding isp at the end because it's in different column index in baseline table for some products.
   -- This is to make sure downstream union works as intended.
   isp,
+  CASE
+    WHEN STARTS_WITH(device_model, "iPad") AND normalized_os = "iOS"
+      THEN "iPad"
+    WHEN STARTS_WITH(device_model, "iPhone") AND normalized_os = "iOS"
+      THEN "iPhone"
+    WHEN normalized_os = "Android"
+      THEN "Android"
+    ELSE
+      CAST(NULL AS STRING)
+  END AS device_type,
 FROM
   `{{ project_id }}.{{ dataset }}.baseline_clients_last_seen`
--- a/sql_generators/mobile_kpi_support_metrics/templates/active_users.view.sql
+++ b/sql_generators/mobile_kpi_support_metrics/templates/active_users.view.sql
@@ -48,9 +48,9 @@ SELECT
   -- This is to make sure downstream union works as intended.
   isp,
   CASE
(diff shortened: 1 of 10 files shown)


sql_generators/mobile_kpi_support_metrics/templates/active_users.view.sql
```diff
@@ -3,7 +3,10 @@ CREATE OR REPLACE VIEW
   `{{ project_id }}.{{ dataset }}.{{ name }}`
 AS
 SELECT
-  * EXCEPT (isp),
+  * EXCEPT (isp) REPLACE(
```
bani, line 6: why were we removing ISP from this view?

sql_generators/mobile_kpi_support_metrics/templates/engagement.query.sql
```diff
@@ -1,4 +1,24 @@
...
+  SELECT
+    submission_date,
+    device_manufacturer,
+    COUNT(*) AS device_manufacturer_count,
+  FROM
+  `{{ project_id }}.{{ dataset }}.engagement_clients`
+  WHERE
+    {% raw %}
+    {% if is_init() %}
+      submission_date < CURRENT_DATE
```
bani, line 12: is this the condition that will run through when we backfill the table?
It seems we'll have vastly different criteria for the first run vs the rest as many smaller manufacturers could pass the threshold of 2k if we look at all historical data

sql_generators/mobile_kpi_support_metrics/templates/engagement.query.sql
```diff
@@ -1,4 +1,24 @@
 {{ header }}
+WITH device_manufacturer_counts AS (
+  SELECT
+    submission_date,
+    device_manufacturer,
+    COUNT(*) AS device_manufacturer_count,
```
bani, line 6: how about adding a rank and doing top 100?
2k seems arbitrary and I wonder how it'll change over time.

sql_generators/mobile_kpi_support_metrics/templates/active_users.view.sql
kik-kik: we, include it at the bottom of the view definition. If I recall correctly and based on the comment there it's because in some of the products that column is in different position in comparison to others causing the downstream union across all products to fail.

sql_generators/mobile_kpi_support_metrics/templates/engagement.query.sql
kik-kik: Honestly, I did not think about it at the time. But that sounds like a fair suggestion. I can make that change.

kik-kik: ok, I made the change. I believe this is what you meant.

sql_generators/mobile_kpi_support_metrics/templates/new_profiles.query.sql
```diff
@@ -1,4 +1,24 @@
 {{ header }}
+WITH device_manufacturer_counts AS (
+  SELECT
+    first_seen_date,
+    device_manufacturer,
```
bani, line 5: we also should lower here, as there are duplicates with upper/lower case

sql_generators/mobile_kpi_support_metrics/templates/new_profiles.query.sql
```diff
@@ -1,4 +1,24 @@
...
+    DENSE_RANK() OVER(PARTITION BY submission_date ORDER BY COUNT(*) DESC) AS manufacturer_rank,
```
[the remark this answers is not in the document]
kik-kik: ok, changed to RANK.

sql_generators/mobile_kpi_support_metrics/templates/active_users.view.sql
```diff
@@ -3,7 +3,10 @@ CREATE OR REPLACE VIEW
...
+    -- Lower device_manufacturer as in some cases the same manufacturer value has different casing.
+    LOWER(device_manufacturer) AS device_manufacturer
```
kik-kik, line 8: Lowering device_manufacturer here so that everything downstream inherits the lowered values.

bani (approved):

Issue #5830, opened 2024-06-24
Union default_table_workgroup_access with table-level workgroup_access in metadata generation logic


whd:
When we implemented retention metadata support via `default_table_workgroup_access`, we introduced a gotcha where unless you specify the value of `default_table_workgroup_access` explicitly in a table's manually created `workgroup_access`, access to that table will be restricted to just the manual grants if there are deprecated tables in the same dataset.

https://github.com/mozilla/bigquery-etl/commit/80dc78673f60c2f4d5141d797ff84ec1875298f2 is an example of this: if there were deprecations in `mozilla_org_derived`, give the way our metadata generation currently works, `workgroup:mozilla-confidential` would have lost access to the table.

We should for non-deprecated tables automatically apply `default_table_workgroup_access` to the table-level metadata.yaml `workgroup_access`, and probably have a CI/lint check to warn against the redundant specification in both places once this is implemented.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-4143)

PR #6811, opened 2025-01-14
#5830 Union default_table_workgroup_access with table-level workgroup_accesss in metadata generation logic

scholtzan:
Fixes https://github.com/mozilla/bigquery-etl/issues/5830

Update the `metadata update` logic to union `default_table_workgroup_access` entries with table-level `workgroup_access` entries

### DIFF
--- a/bigquery_etl/cli/metadata.py
+++ b/bigquery_etl/cli/metadata.py
@@ -10,7 +10,11 @@ import click
 from dateutil.relativedelta import relativedelta
 from google.cloud import bigquery
 
-from bigquery_etl.metadata.parse_metadata import DatasetMetadata, Metadata
+from bigquery_etl.metadata.parse_metadata import (
+    DatasetMetadata,
+    Metadata,
+    WorkgroupAccessMetadata,
+)
 from bigquery_etl.metadata.publish_metadata import publish_metadata
 
 from ..cli.utils import (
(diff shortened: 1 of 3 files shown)


kwindau (approved):

BenWu:
In `sql/moz-fx-data-shared-prod/activity_stream_bi/impression_stats_flat_v1/metadata.yaml` the generated diff, it does 

```
-      # to avoid permissions desync mozilla-confidential access is only granted
-      # at the dataset-level for now
-      # - workgroup:mozilla-confidential
+ - workgroup:mozilla-confidential
```
so it adds mozilla confidential when it shouldn't.  Would that cause issues?  I don't understand what the desync means.

bigquery_etl/cli/metadata.py
```diff
@@ -94,10 +98,41 @@ def update(name: str, sql_dir: Optional[str], project_id: Optional[str]) -> None
...
+                            default_workgroup_member
+                            for default_workgroup_member in default_workgroup_access.get(
+                                "members", []
+                            )
+                            if default_workgroup_member not in table_metadata_members
+                        )
+
+                        table_metadata.workgroup_access[i].members = (
+                            table_metadata_members
+                        )
```
BenWu, line 125: A bit code golfy but I think this is equivalent and would be easier to read
```suggestion
                        table_metadata.workgroup_access[i].members = sorted(
                            set(table_workgroup_access.members)
                            | set(default_workgroup_access.get("members", []))
                        )
```

BenWu:
Having only table-specific permissions in the metadata in `main` makes sense to me.  Keeping redundant entries could also imply that tables can be made more restrictive than the default using metadata

scholtzan:
I filed a follow-up PR that adds a CI check and removes redundant table-level access configurations: https://github.com/mozilla/bigquery-etl/pull/6822/commits/c82eb25f66d1a7ca95ba996b2eda2a99a5ff49f7
The PR depends on the changes made here, so should be merged after

PR #6707, opened 2024-12-19
DENG-6441 Handle usage_deletion_request in Shredder

akkomar:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7086)

PR #6814, opened 2025-01-15
Bug 1932644 - Add the usage-reporting (and its deletion-request) ping to shredder

badboy:
The ping has not yet landed, but once it has we should land the shredder config


badboy:
@akkomar this is not yet complete (desktop) only. However is this the right way to add it?

akkomar:
Yeah, but I have another PR which will handle other applications automatically: https://github.com/mozilla/bigquery-etl/pull/6707

badboy:
> Yeah, but I have another PR which will handle other applications automatically: #6707

D'oh! Thanks.

PR #6822, opened 2025-01-15
Redundant workgroup metadata CI check

scholtzan:
## Description
Follow-up to https://github.com/mozilla/bigquery-etl/pull/6811#issuecomment-2593766594

This adds a new CI check to ensure that there are no table-level access definitions that are redundant with dataset access definitions. This PR also removes configs that are redundant

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7402)

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -289,6 +289,10 @@ jobs:
             - checkout
             - *restore_venv_cache
             - *build
+            - run:
+                name: Validate workgroup access configuration on main
+                command: |
+                  PATH="venv/bin:$PATH" script/bqetl metadata validate-workgroups
             - *attach_generated_sql
             - *copy_staged_sql
             - *authenticate
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -292,7 +292,7 @@ jobs:
             - run:
                 name: Validate workgroup access configuration on main
                 command: |
-                  PATH="venv/bin:$PATH" script/bqetl metadata validate-workgroups
+                  PATH="venv/bin:$PATH" script/bqetl metadata validate-workgroups sql/
             - *attach_generated_sql
             - *copy_staged_sql
             - *authenticate
--- a/bigquery_etl/metadata/validate_metadata.py
+++ b/bigquery_etl/metadata/validate_metadata.py
@@ -408,8 +408,6 @@ def validate(target):
                     if not validate_col_desc_enforced(root, metadata):
                         failed = True
 
-                    if not validate_workgroup_access(metadata, path):
-                        failed = True
                     # todo more validation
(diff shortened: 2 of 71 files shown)


BenWu (approved):

whd (approved):
Diff LGTM, thanks for making this more ergonomic

whd:
One more thing it would be great to add as a CI check: I think we should remove `default_table_workgroup_access` from `main` and require it not be set via a CI check. It should always have the same value as `workgroup_access` and is used exclusively by metadata generation logic to decide whether to set `workgroup_access` differently in generated metadata.

.circleci/workflows.yml
```diff
@@ -289,6 +289,10 @@ jobs:
             - checkout
             - *restore_venv_cache
             - *build
+            - run:
+                name: Validate workgroup access configuration on main
+                command: |
+                  PATH="venv/bin:$PATH" script/bqetl metadata validate-workgroups sql/
```
scholtzan, line 295: Run the metadata workgroup access validation before generated SQL gets copied over. During SQL generation also metadata is generated, which updates and sets the `default_table_workgroup_access` and `workgroup_access` configs and would make the CI check fail.

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -284,6 +284,73 @@ def validate_exclusion_list_expiration_days(metadata, path):
...
+        is_valid = False
+        click.echo(
+            click.style(
+                f"ERROR: default_table_workgroup_access should not be explicity specified in {dataset_metadata_path}. "
+                + "The default_table_workgroup_access configuration will be automatically generated.",
+                fg="red",
+            )
+        )
+
+    return is_valid
```
scholtzan, line 351: Added an additional check to make sure `default_table_workgroup_access` is not being set explicitly. This check runs on non-generated files (which should be equal to files that are on `main`).

scholtzan:
@BenWu @whd Might be good if you can take another look at this. I added a check as described in https://github.com/mozilla/bigquery-etl/pull/6822#issuecomment-2596474076 and removed the `default_table_workgroup_access` from `main`

BenWu (commented):
Shouldn't the `default_table_workgroup_access` still be in the sql.diff because it's generated?  I think the metadata update might need some changes because `default_table_workgroup_access` is never null https://github.com/mozilla/bigquery-etl/blob/3282192a21972f99e9bd9bb0fb98570bfecc07f7/bigquery_etl/metadata/parse_metadata.py#L493-L496

BenWu:
I triggered CI and then cancelled it so it's showing up as failed.  I'll just rerun it

PR #6827, opened 2025-01-16
feat(DENG-7038): Addons Search Detection aggregate

kik-kik:
# feat(DENG-7038): Addons Search Detection aggregate

## Description

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7457)

### DIFF
--- a/sql/moz-fx-data-shared-prod/addons/addon_events_clients/view.sql
+++ b/sql/moz-fx-data-shared-prod/addons/addon_events_clients/view.sql
@@ -1,0 +1,40 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.addons.addon_events_clients`
+AS
+SELECT
+  submission_date,
+  `timestamp`,
+  COALESCE(
+    udf.get_key(event_map_values, 'addonId'),
+    udf.get_key(event_map_values, 'addon_id')
+  ) AS addon_id,
+  COALESCE(
+    udf.get_key(event_map_values, 'addonVersion'),
+    udf.get_key(event_map_values, 'addon_version')
+  ) AS addon_version,
+  -- TODO: why in some redash queries this is hashed_addon_id and in other value? What's the difference?
+  event_string_value AS hashed_addon_id,
+  client_id,
+  sample_id,
+  normalized_channel,
+  country,
+  locale,
+  app_name,
+  app_version,
+  os AS os_name,
+  os_version,
+  event_timestamp,
+  event_category,
+  event_method AS method,
+  event_object AS `object`,
+  event_string_value AS `value`,
+  event_map_values,
+  event_process,
+FROM
+  `moz-fx-data-shared-prod.telemetry.events`
+WHERE
+  submission_date >= '2021-08-09'
+  AND event_category IN (
+    "addonsManager",
+    "addonsSearchDetection"
+  )
--- a/sql/moz-fx-data-shared-prod/addons/addon_events_clients/view.sql
+++ b/sql/moz-fx-data-shared-prod/addons/addon_events_clients/view.sql
@@ -5,12 +5,12 @@ SELECT
   submission_date,
   `timestamp`,
   COALESCE(
-    udf.get_key(event_map_values, 'addonId'),
-    udf.get_key(event_map_values, 'addon_id')
+    `moz-fx-data-shared-prod.udf.get_key`(event_map_values, 'addonId'),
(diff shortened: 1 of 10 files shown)


alekhyamoz (approved):

sql/moz-fx-data-shared-prod/addons/addon_events_clients/view.sql
```diff
@@ -0,0 +1,37 @@
...
+  `timestamp`,
+  COALESCE(
+    `moz-fx-data-shared-prod.udf.get_key`(event_map_values, 'addonId'),
+    `moz-fx-data-shared-prod.udf.get_key`(event_map_values, 'addon_id')
+  ) AS addon_id,
+  COALESCE(
+    `moz-fx-data-shared-prod.udf.get_key`(event_map_values, 'addonVersion'),
+    `moz-fx-data-shared-prod.udf.get_key`(event_map_values, 'addon_version')
+  ) AS addon_version,
+  -- TODO: why in some redash queries this is hashed_addon_id and in other value? What's the difference?
```
willdurand, line 15: This depends on what gets passed. The `hashed_addon_id` is only valid for `install_stats`: https://searchfox.org/mozilla-central/rev/5b061cdc4d40d44988dc61aa941cfbd98e31791f/toolkit/components/telemetry/Events.yaml#233-234

sql/moz-fx-data-shared-prod/addons/addon_events_clients/view.sql
```diff
@@ -0,0 +1,37 @@
...
+  event_method AS method,
+  event_object AS `object`,
+  event_string_value AS `value`,
+  event_map_values,
+  event_process,
+FROM
+  `moz-fx-data-shared-prod.telemetry.events`
+WHERE
+  submission_date >= '2021-08-09'
+  AND event_category IN ("addonsManager", "addonsSearchDetection")
```
willdurand, line 37: I think it'd be good to _not_ do this. `addonsSearchDetection` is its very own thing, could we keep it self-contained by chance?

willdurand (commented):
~~My main question is why are we pulling the legacy Telemetry data and not Glean? Could we switch to Glean for this ticket?~~

Ignore the above, please. The Glean probes for this are relatively new (since August 2024) so that would help there.

I still think we should focus on search detection only and keep that small and self-contained. Could we simplify a bit this patch?

sql/moz-fx-data-shared-prod/addons/addon_events_clients/view.sql
kik-kik: this is just a view on top of telemetry.events just to make it a bit easier to query the addons data and prevent other event categories to be included in the results in case the filter is not included in your query. event_category field is still accessible in this view and can still be used to filter for a specific group of events in the follow up queries. Like we do in search_detection_v1 for example:
https://github.com/mozilla/bigquery-etl/pull/6827/files/774329c67a7bc861d6945d1aadeb70e13f0456b7#diff-a9127827452486e083c54ec89dee402563de06263285a5ef981e53cf50215503R22

kik-kik: Yes, based on the other comment you left along with the description and explanation that hashed_addon_id is only in some cases I removed the alias.

willdurand (approved):

sql/moz-fx-data-shared-prod/addons_derived/dataset_metadata.yaml
```diff
@@ -0,0 +1,11 @@
+friendly_name: Addons Derived
+description: |-
+  Derived data used for Addons related analysis.
+dataset_base_acl: derived
+user_facing: true
```
sean-rose, line 5: I don't think derived datasets can be user-facing.
```suggestion
user_facing: false
```

sql/moz-fx-data-shared-prod/addons_derived/search_detection_v1/schema.yaml
```diff
@@ -0,0 +1,97 @@
+-- Based on telemetry.events: https://docs.telemetry.mozilla.org/concepts/pipeline/event_pipeline
```
sean-rose, line 1: This is invalid YAML.  YAML comments start with `#`.
```suggestion
# Based on telemetry.events: https://docs.telemetry.mozilla.org/concepts/pipeline/event_pipeline
```

PR #6828, opened 2025-01-16
feat(org_mozilla_broken_site_report): Add mobile reports to the unified view.

denschub:
## Description

We use this view as our point of ingestion for user reports. Recently, we also shipped the reporter to Fenix, and we want to ingest that data, too. Not sure if this is the most efficient way of doing this - I originally wanted to UNION the main query, but that'd have me duplicate everything, which felt wrong.

This also replaces the `app_name` with a static string for desktop and mobile. For desktop, this is not changing anything - but for mobile, the `normalized_app_name` column is always `null`. After chatting with @badboy, I learned that [this column is kinda legacy anyway](https://bugzilla.mozilla.org/show_bug.cgi?id=1672191#c1), and we only really need it to see which product the reports came from, so this is an easy way to make sure this is set right.

r? @scholtzan, please :)

n/a

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7524)

### DIFF
--- a/sql/moz-fx-data-shared-prod/org_mozilla_broken_site_report/user_reports/view.sql
+++ b/sql/moz-fx-data-shared-prod/org_mozilla_broken_site_report/user_reports/view.sql
@@ -7,7 +7,7 @@ SELECT
   metrics.text2.broken_site_report_description AS comments,
   metrics.url2.broken_site_report_url AS url,
   metrics.string.broken_site_report_breakage_category AS breakage_category,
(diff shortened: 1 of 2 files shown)


denschub:
Oh, the `user_reports_live` view needs changes, too. I'll update this PR in a bit.

denschub:
Okay, now this is really ready for review. 🙈

scholtzan (approved):

PR #6835, opened 2025-01-17
feat(dryrun): Pass all query parameters when dry-running queries

sean-rose:
## Description
The `bqetl dryrun` command doesn't currently support queries which use any non-date query parameters.  This PR adds support for such queries by passing all expected query parameters when dry-running queries.

This depends on mozilla-services/cloudops-infra#6189 to update the `bigquery_etl_dryrun` Google Cloud function to accept query parameters as an argument.

* https://github.com/mozilla-services/cloudops-infra/pull/6189

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7559)

### DIFF
--- a/bigquery_etl/dryrun.py
+++ b/bigquery_etl/dryrun.py
@@ -38,6 +38,19 @@ except ImportError:
     from backports.cached_property import cached_property  # type: ignore
 
 
+QUERY_PARAMETER_TYPE_VALUES = {
+    "DATE": "2019-01-01",
+    "DATETIME": "2019-01-01 00:00:00",
+    "TIMESTAMP": "2019-01-01 00:00:00",
+    "STRING": "foo",
+    "BOOL": True,
+    "FLOAT64": 1,
+    "INT64": 1,
+    "NUMERIC": 1,
+    "BIGNUMERIC": 1,
+}
+
+
 def get_credentials(auth_req: Optional[GoogleAuthRequest] = None):
     """Get GCP credentials."""
     auth_req = auth_req or GoogleAuthRequest()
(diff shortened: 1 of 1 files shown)


bigquery_etl/dryrun.py
```diff
@@ -38,6 +38,19 @@
...
 
 
+QUERY_PARAMETER_TYPE_VALUES = {
+    "DATE": "2019-01-01",
+    "DATETIME": "2019-01-01 00:00:00",
+    "TIMESTAMP": "2019-01-01 00:00:00",
+    "STRING": "foo",
+    "BOOL": True,
+    "FLOAT64": 1,
+    "INT64": 1,
```
BenWu, line 48: I didn't try every possible alias but `INTEGER` and `FLOAT` (but not `INT`) are accepted as types as well, e.g. `bq query --parameter "val:INTEGER:10" --nouse_legacy_sql "SELECT @val"`.  I would probably think to use those before the 64 versions
```suggestion
    "FLOAT": 1,
    "INT64": 1,
    "INTEGER": 1,
```

bigquery_etl/dryrun.py
```diff
@@ -204,28 +217,30 @@ def dry_run_result(self):
...
-                    *(
-                        param.split(":", 1)[0]
-                        for param in self.metadata.scheduling.get("parameters", [])
-                        if re.fullmatch(r"[^:]+:DATE:{{.*ds.*}}", param)
-                    ),
+
+        query_parameters = []
+        scheduling_metadata = self.metadata.scheduling if self.metadata else {}
+        if date_partition_parameter := scheduling_metadata.get(
+            "date_partition_parameter", "submission_date"
```
BenWu, line 224: Does `submission_date` need to be set if if there's no `date_partition_parameter`?

sean-rose: If `date_partition_parameter` isn't specified at all in the scheduling metadata then it does default to passing a `@submission_date` parameter, so dryrun should too.  However, if `date_partition_parameter` is set to null in the scheduling metadata then no `@submission_date` parameter is passed by default, so this does the same in that case (the `"date_partition_parameter"` key will exist in the scheduling dictionary, so the `get()` will return the null value as `None` which won't pass the `if` check).

BenWu: Oh I see.  I didn't know it did that

BenWu (approved):

PR #6844, opened 2025-01-21
Fix search templates for some of the apps

alekhyamoz:
`SELECT
  DISTINCT
   app_name,
   os,
   normalized_app_name,
FROM `moz-fx-data-shared-prod.search_derived.mobile_search_clients_daily_v2`
WHERE submission_date = "2025-01-01"`
we're seeing locale making it into the os column for some apps


kik-kik (approved):

kik-kik:
For context, this is because later we do a union across the the different templates and with columns having different ordering the locale values make it into the os field for some of the products. This fixes the ordering inside the templates to avoid this from happening.

PR #6847, opened 2025-01-22
feat(DENG-6890): mobile kpi support metrics managed backfill 20250122

kik-kik:
# feat(DENG-6890): mobile kpi support metrics managed backfill 20250122

New fields were added to the tables in order to support DMA analysis. As part of this we need to backfill the corresponding tables for both firefox_ios and fenix. This includes backfill the following tables for both products with data starting 2024-03-05:

- engagement_v1
- retention_v1
- new_profiles_v1

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7618)

### DIFF
--- a/sql/moz-fx-data-shared-prod/fenix_derived/engagement_v1/backfill.yaml
+++ b/sql/moz-fx-data-shared-prod/fenix_derived/engagement_v1/backfill.yaml
@@ -5,11 +5,3 @@
   watchers:
   - kik@mozilla.com
   status: Initiate
-
-2024-05-31:
-  start_date: 2021-01-01
-  end_date: 2024-05-31
-  reason: The table is created, this is to populate it with data.
-  watchers:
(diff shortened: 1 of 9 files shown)


lucia-vargas-a:
> I think if the `metadata.yaml` has `shredder_mitigation: true` then you need to also add it to the `backfill.yaml` file. Or else it will fail in airflow. However, I believe it will also fail when shredder_mitigation is set to true for v1 tables.
> 
> @lucia-vargas-a can you confirm?

Yes, confirmed. Tables labeled with shredder-mitigation need to follow the [corresponding process](https://docs.telemetry.mozilla.org/cookbooks/data_modeling/shredder_mitigation#running-a-managed-backfill-with-shredder-mitigation).

alekhyamoz (approved):
lgtm

PR #6858, opened 2025-01-23
feat: add usage_reporting generator

kik-kik:
# feat: add usage_reporting generator

## Description

Alternative approach to generating the usage_reporting derived tables. This is now unrelated to glean_usage generation and we restrict the generation to specific apps and channels.

What this generator should do:

1. Retrieve a list of apps and their channels generated by probe scraper.
2. Generate the following queries for each channel defined bqetl_project.yaml:
- usage_reporting_daily_v1
- usage_reporting_first_seen_v1
- usage_reporting_last_seen_v1
3. Generate views for each channel for accessing the derived tables:
- usage_reporting_daily
- usage_reporting_first_seen
- usage_reporting_last_seen
4. Generate a union view combining all channel views into a single app view.

---

Usage:

`./bqetl generate usage_reporting`

---

Things that we should do next:
- Add some tests
- refactor the logic, this is a first draft / pass. And can be improved.
- ~Consider if we want to simplify it even further and define the app name along with app ids (channels) inside `bqetl_project.yaml` rather than using probe scraper data.~ Looks like probe scraper is providing us with reliable channel information so maybe we should keep that.


sql_generators/usage_reporting/templates/usage_reporting_clients_daily_v1.query.sql
```diff
@@ -0,0 +1,66 @@
+{{ header }}
+WITH usage_reporting_base AS (
+  SELECT
+    submission_timestamp,
+    DATE(submission_timestamp) AS submission_date,
```
irrationalagent, line 5: I wonder if we should have `app_name` and/or `normalized_app_id` here as well. I'm not entirely sure the difference between them, but I see that `normalized_app_id` is present on the baseline clients_daily tables.

sql_generators/usage_reporting/templates/usage_reporting_clients_daily_v1.schema.yaml
```diff
@@ -0,0 +1,90 @@
...
+    The optional reason the ping was submitted. The specific values for reason are specific to each ping,
+    and are documented in the ping's pings.yaml file.
+
+- mode: NULLABLE
+  name: is_active
+  type: BOOLEAN
+  description: |
+    A flag field indicating whether the specific client was active.
+
+# - mode: NULLABLE
```
irrationalagent, line 85: confirmed that you can remove this. not present in the ping and not approved to be.

sql_generators/usage_reporting/templates/usage_reporting_clients_daily_v1.query.sql
```diff
@@ -0,0 +1,66 @@
...
+  udf.mode_last(ARRAY_AGG(normalized_os_version IGNORE NULLS ORDER BY submission_timestamp ASC)) AS normalized_os_version,
+  udf.mode_last(ARRAY_AGG(app_display_version IGNORE NULLS ORDER BY submission_timestamp ASC)) AS app_display_version,
+  udf.mode_last(ARRAY_AGG(app_build IGNORE NULLS ORDER BY submission_timestamp ASC)) AS app_build,
+  udf.mode_last(ARRAY_AGG(distribution_id IGNORE NULLS ORDER BY submission_timestamp ASC)) AS distribution_id,
+  udf.mode_last(ARRAY_AGG(is_default_browser IGNORE NULLS ORDER BY submission_timestamp ASC)) AS is_default_browser,
+  udf.mode_last(ARRAY_AGG(reason IGNORE NULLS ORDER BY submission_timestamp ASC)) AS reason,
+  {% if "_desktop" in app_name %}
+  COALESCE(LOGICAL_OR(is_active), SUM(browser_engagement_uri_count) > 0 AND SUM(browser_engagement_active_ticks) > 0, FALSE) AS is_active,
+  {% else %}
+  COALESCE(LOGICAL_OR(is_active), SUM(IF(duration BETWEEN 0 AND 100000, duration, 0)) > 0, False) AS is_active,
```
irrationalagent, line 60: I think we might need a `COALESCE(..., 0)` on `duration`, `browser_engagement_uri_count` and `browser_engagement_active_ticks` here. They can be null (duration is often null since its only sent on inactive pings)

bqetl_project.yaml
```diff
@@ -385,6 +385,18 @@ generate:
...
     skip_datasets:
     - mlhackweek_search
+  usage_reporting:
+    apps:
+      fenix:
+        channels:
+        - nightly
+      # firefox_ios:
+      #   channels:
+      #   - nightly
```
irrationalagent, line 395: I checked, we should have data there now https://sql.telemetry.mozilla.org/queries/105052/source

sql_generators/usage_reporting/templates/usage_reporting_clients_daily_v1.schema.yaml
kik-kik: by this you mean `locale`?

sql_generators/usage_reporting/templates/usage_reporting_clients_daily_v1.query.sql
kik-kik: It is available in the union all the app ids:
https://github.com/mozilla/bigquery-etl/pull/6858/files#diff-205ecd62049b013a01eeb8669b85657830ff61de0a2755585e66c3bfe17296d1R10

If you'd prefer to have it already in each app, then I'd suggest we add to the the view on top of the daily_v1 table.

bqetl_project.yaml
kik-kik: I will take a look later. Thanks for letting me know.

sql_generators/usage_reporting/templates/usage_reporting_clients_daily_v1.schema.yaml
kik-kik: removing.

sql_generators/usage_reporting/templates/app_union.view.sql
```diff
@@ -0,0 +1,19 @@
...
+  `{{ project_id }}.{{ app_name }}.{{ view_name }}`
+AS
+{% for channel in channels_info -%}
+{% if not loop.first -%}
+UNION ALL
+{% endif -%}
+SELECT
+  "{{ channel.channel_dataset }}" AS normalized_app_id,
+  -- {% if app_name == "fenix" and view_name == "usage_reporting_clients_daily" %}
+  -- mozfun.norm.fenix_app_info("{{ channel.channel_dataset }}", app_build).channel AS normalized_channel,
```
kik-kik, line 12: @akkomar do we even need the UDF call here? My assumption here is that there were some bad builds in the past that required this, but considering this will be for new clients going forward is it enough if we just use the channel provided by probe scraper?

badboy: Looking at the UDF source I agree: it shouldn't be needed. the app IDs are now distinct (and we only have 3 left)

kik-kik: Thanks @akkomar and @badboy for the additional context. Then unless there are any objections I'd suggest we proceed without the UDF call and just use the value provided by probe-scraper. Considering this is only nightly right now I think this is okay and we can observe to see if we notice anything unexpected. Since this is quite a new thing we can easily tweak this later on.

sql_generators/usage_reporting/templates/usage_reporting_clients_daily_v1.query.sql
```diff
@@ -0,0 +1,60 @@
...
+    metrics.string.usage_app_build AS app_build,
+    metrics.string.usage_reason AS reason,
+    CAST(NULL AS BOOLEAN) AS is_active,  -- Eventually is_active value will come from the client.
+    COALESCE(metrics.timespan.usage_duration.value, 0) AS duration,
+    -- fields only currently available in fenix.
+    {% if app_name == "fenix" %}
+    metrics.string.metrics_distribution_id AS distribution_id,
+    COALESCE(metrics.boolean.metrics_default_browser, FALSE) AS is_default_browser, -- this should eventually also be added to ios and desktop.
+    {% else %}
+    CAST(NULL AS STRING) AS distribution_id,
```
irrationalagent, line 22: distribution_id should start showing up on desktop in the next day or so.

sql_generators/usage_reporting/README.md
```diff
@@ -0,0 +1,53 @@
+# Usage Reporting
+
+Generator responsible for generating queries and views to prepare the usage_reporting ping data for analytics purposes. To large extend, the pipeline generated by this is very similar to that of baseline_clients_daily, last_seen, first_seen created by the `glean_usage` generator. This is in order to maintain a level of consistency across the two data sources and enable comparisons across the two.
```
irrationalagent, line 3:
```suggestion
Generator responsible for generating queries and views to prepare the usage_reporting ping data for analytics purposes. To a large extent, the pipeline generated by this is very similar to that of baseline_clients_daily, last_seen, first_seen created by the `glean_usage` generator. This is in order to maintain a level of consistency across the two data sources and enable comparisons across the two.
```

sql_generators/usage_reporting/templates/usage_reporting_clients_first_seen_v1.metadata.yaml
```diff
@@ -0,0 +1,19 @@
+# {{ header }}
+friendly_name: Clients First Seen Based on the DAU Reporting Ping.
+description: |-
+  A representation of when we saw each `profile_usage_id`
```
irrationalagent, line 4:
```suggestion
  A representation of when we saw each `usage_profile_id`
```

sql_generators/usage_reporting/templates/usage_reporting_clients_daily_v1.metadata.yaml
```diff
@@ -0,0 +1,23 @@
+# {{ header }}
+friendly_name: Clients Daily Based on the DAU Reporting Ping.
+description: |-
+  A daily aggregate of usage_reporting pings per `profile_usage_id`.
```
irrationalagent, line 4:
```suggestion
  A daily aggregate of usage_reporting pings per `usage_profile_id`.
```

sql_generators/usage_reporting/templates/usage_reporting_clients_last_seen_v1.metadata.yaml
```diff
@@ -0,0 +1,19 @@
+# {{ header }}
+friendly_name: Clients Last Seen Based on the DAU Reporting Ping.
+description: |-
+  A daily aggregate of the usage_reporting ping for each `profile_usage_id`
```
irrationalagent, line 4:
```suggestion
  A daily aggregate of the usage_reporting ping for each `usage_profile_id`
```

irrationalagent (approved):
r+wc

bqetl_project.yaml
```diff
@@ -387,6 +387,18 @@ generate:
...
     - mlhackweek_search
+  usage_reporting:
+    apps:
+      fenix:
+        channels:
+        - nightly
+      firefox_ios:
+        channels:
+        - nightly
+      # firefox_desktop:
```
gkatre, line 398: Is the `firefox_desktop` commented out for a reason?
I see the usage of `firefox_desktop` in [clients_daily](https://github.com/mozilla/bigquery-etl/pull/6858/files#diff-37ce5183fad2c4f79538026a3f0d5016dcc64acc1cf7c63767a2b7bd574d4cd4R21)

sql_generators/usage_reporting/templates/usage_reporting_clients_last_seen_v1.query.sql
```diff
@@ -0,0 +1,52 @@
...
+    ) AS days_seen_bits,
+    udf.combine_adjacent_days_28_bits(
+      _previous.days_active_bits,
+      _current.days_active_bits
+    ) AS days_active_bits,
+    udf.combine_adjacent_days_28_bits(
+      _previous.days_created_profile_bits,
+      _current.days_created_profile_bits
+    ) AS days_created_profile_bits
+  )
```
gkatre, line 47: Do we need to add `days_desktop_active_bits` based on `uri_count` and `active_ticks` as done [here](https://github.com/mozilla/bigquery-etl/blob/main/sql_generators/glean_usage/templates/baseline_clients_last_seen_v1.query.sql#L88)?

gkatre (approved):
@kik-kik Thanks! just 2 comments from me. otherwise looks good.

PR #6878, opened 2025-01-24
Specify ping_info fields in event_monitoring_aggregates

BenWu:
## Description

[Dry run](https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/44269/workflows/4b6ec306-8298-405a-bb6c-e3eacedbae88/jobs/513223) is failing with
```
[{'code': 400, 'errors': [{'message': 'Column 6 in UNION ALL has incompatible types: STRUCT<end_time STRING, ...> at [627:5]'
```
because of the new `home_v1` table which has a different order of fields in `ping_info`.  

Sean pointed out that we could reuse [Schema.generate_compatible_select_expression()](https://github.com/mozilla/bigquery-etl/blob/fcc07de2d2f982d1f573b284e5de9be035a21ec7/bigquery_etl/schema/__init__.py#L299-L426) here but due to urgency I'm putting this PR up for now

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7676)

### DIFF
--- a/sql_generators/glean_usage/templates/event_monitoring_aggregates_v1.query.sql
+++ b/sql_generators/glean_usage/templates/event_monitoring_aggregates_v1.query.sql
@@ -69,7 +69,26 @@
       normalized_country_code,
       client_info.app_channel AS channel,
       client_info.app_display_version AS version,
-      ping_info
+      STRUCT (
+        ping_info.end_time,
+        ARRAY(
+          SELECT AS STRUCT
+            key,
+            STRUCT(
+              value.branch,
+              STRUCT(
+                value.extra.type,
+                value.extra.enrollment_id
+              ) AS extra
+            ) AS value
+          FROM
+            UNNEST(ping_info.experiments)
+        ) AS experiments,
+        ping_info.ping_type,
+        ping_info.seq,
+        ping_info.start_time,
+        ping_info.reason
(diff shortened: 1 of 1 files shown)


kwindau (approved):
TY

sql_generators/glean_usage/templates/event_monitoring_aggregates_v1.query.sql
```diff
@@ -69,7 +69,26 @@
...
+            key,
+            STRUCT(
+              value.branch,
+              STRUCT(
+                value.extra.type,
+                value.extra.enrollment_id
+              ) AS extra
+            ) AS value
+          FROM
+            UNNEST(ping_info.experiments)
```
sean-rose, line 85: Note that unnesting isn't guaranteed to preserve the order of the unnested elements.  As long as the order of the experiements array elements doesn't matter this is fine/simpler/preferable, but if order matters you'll need to use `WITH OFFSET ...` and `ORDER BY` to preserve the order.

BenWu: That's a good point.  I'm pretty sure order doesn't matter but I'll add it

PR #6886, opened 2025-01-27
Fix sponsored tiles

alekhyamoz:
[bqetl_sponsored_tiles_clients_daily.sponsored_tiles_clients_daily_v1 ](https://workflow.telemetry.mozilla.org/dags/bqetl_sponsored_tiles_clients_daily/grid) has failed for the last few days and continues to fail with a timeout issue.
Modified the parts of the query to point to `events_stream` instead of `events_unnested`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7694)


skahmann3:
Let's tag @sergiosonline and @jaredsnyder on this one instead. They're more likely to use this table.

jaredsnyder:
I do not feel qualified to review this, might be a better job for one of the ads-attached DEs like @lucia-vargas-a  or @chelseybeck

BenWu (commented):
I'm pretty sure the output is equivalent but I'm not sure this will fix the memory issue since this doesn't change the amount of data scanned.  I think the expensive part of the query is the `fenix_experiments_info` CTE which is still using events_unnested so you may need to adjust the overactive filter or change the query

BenWu:
How did you test the query?  I ran the query with `@submission_date = '2025-01-26'` and I'm still getting `Resources exceeded during query execution: The query could not be executed in the allotted memory. Peak usage: 118% of limit. Top memory consumer(s): aggregate functions and GROUP BY clauses: 100%`.

In particular, just this part fails:
```sql
WITH overactive_android_clients AS (
  -- Find client_ids with over 3 000 000 events in a day,
  -- which could cause errors in the next step due to aggregation overflows.
  SELECT
    client_id AS client_id
  FROM
    `moz-fx-data-shared-prod.fenix.events_stream`
  WHERE
    DATE(submission_timestamp) = "2025-01-26"
  GROUP BY
    client_id
  HAVING
    COUNT(*) > 3000000
),
fenix_experiments_info AS (
  SELECT
    client_info.client_id AS client_id,
    ARRAY_AGG(DATE(submission_timestamp) ORDER BY submission_timestamp DESC)[
      OFFSET(0)
    ] AS submission_date,
    mozfun.map.mode_last(
      ARRAY_CONCAT_AGG(
        mozfun.glean.legacy_compatible_experiments(ping_info.experiments)
        ORDER BY
          submission_timestamp
      )
    ) AS experiments
  FROM
    `moz-fx-data-shared-prod.fenix.events_unnested`
  LEFT JOIN
    overactive_android_clients
    ON overactive_android_clients.client_id = client_info.client_id
  WHERE
    DATE(submission_timestamp) = '2025-01-26'
    AND overactive_android_clients.client_id IS NULL
  GROUP BY
    client_info.client_id
)
SELECT * FROM fenix_experiments_info
```

What I meant is that just changing it to events stream is unlikely to fix it so something else needs to change.  The easiest thing is changing the overactive filter count.  The higher effort way is to rewrite this query to do something like unnest `ping_info.experiments`, filter, then aggregate instead of the `mozfun.map.mode_last` + `ARRAY_CONCAT_AGG`

sergiosonline:
Hey, folks. I broke it down to three queries--desktop, ios, and android--, incorporating the changes in this PR, and it appears that Android's the culprit here as that new query's consistently stopped: 
* Desktop: https://console.cloud.google.com/bigquery?sq=470293616514:b35acfdd54284d8ca51f1d53470eb133
* ios: https://console.cloud.google.com/bigquery?sq=470293616514:4634e988578543df87690e8404bfdbbb
* Android: https://console.cloud.google.com/bigquery?sq=470293616514:2046b2d987c64047b9c37e818f104227

alekhyamoz:
@sergiosonline  the query works if we lower the overactive filter count to 1500000 from 3000000 (as Ben suggested). Please let me know if that works. Thanks!

sql/moz-fx-data-shared-prod/telemetry_derived/sponsored_tiles_clients_daily_v1/query.sql
```diff
@@ -166,7 +166,7 @@ overactive_android_clients AS (
   GROUP BY
     client_id
   HAVING
-    COUNT(*) > 3000000
+    COUNT(*) > 1500000
```
BenWu, line 169: The comment a few lines above should be updated too

BenWu (approved):
Discussed this with Alekhya yesterday and filtering out clients with >1.5 million events seems like a reasonable low-effort solution.  LGTM from my perspective but I don't know if this will affect analysis using this table

sergiosonline (approved):
Thanks for raising. I'm ok with this.

PR #6887, opened 2025-01-27
Specify ping info fields in event_monitoring_live

BenWu:
## Description

Same issue as https://github.com/mozilla/bigquery-etl/pull/6878 where the ping_info fields don't match for the union.  This is just failing dry run because the materialized views don't get redeployed.  This seems to have a significant impact on query performance for `event_monitoring_aggregates_v1` (+200 to 250 slot hours comparing week-over-week but it runs with on-demand) so there's some risk if `firefox_desktop_derived.event_monitoring_live_v1` gets redeployed since it's already expensive.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7695)

### DIFF
--- a/sql_generators/glean_usage/templates/event_monitoring_aggregates_v1.query.sql
+++ b/sql_generators/glean_usage/templates/event_monitoring_aggregates_v1.query.sql
@@ -82,9 +82,7 @@
               ) AS extra
             ) AS value
           FROM
-            UNNEST(ping_info.experiments) WITH OFFSET
-          ORDER BY
-            offset
+            UNNEST(ping_info.experiments)
         ) AS experiments,
         ping_info.ping_type,
         ping_info.seq,
--- a/sql_generators/glean_usage/templates/event_monitoring_live_v1.materialized_view.sql
+++ b/sql_generators/glean_usage/templates/event_monitoring_live_v1.materialized_view.sql
@@ -68,7 +68,26 @@ IF
         normalized_country_code,
         client_info.app_channel AS channel,
         client_info.app_display_version AS version,
-        ping_info
+        STRUCT (
+          ping_info.end_time,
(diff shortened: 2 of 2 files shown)


BenWu:
I just noticed materialized views only get dry run when they get deployed to stage, i.e. get changed (daily), so this fail every CI run on main

sql_generators/glean_usage/templates/event_monitoring_aggregates_v1.query.sql
```diff
@@ -82,9 +82,7 @@
               ) AS extra
             ) AS value
           FROM
-            UNNEST(ping_info.experiments) WITH OFFSET
-          ORDER BY
-            offset
+            UNNEST(ping_info.experiments)
```
sean-rose, line 85: I'm guessing you're removing this for performance reasons, and because the order doesn't really matter?

sql_generators/glean_usage/templates/event_monitoring_live_v1.materialized_view.sql
```diff
@@ -68,7 +68,26 @@ IF
         normalized_country_code,
         client_info.app_channel AS channel,
         client_info.app_display_version AS version,
-        ping_info
+        STRUCT (
```
sean-rose, line 71:
```suggestion
        STRUCT(
```

sean-rose (approved):
r+wc

sql_generators/glean_usage/templates/event_monitoring_aggregates_v1.query.sql
BenWu: Yes I don't expect any noticeable difference but I'm curious to see

PR #6891, opened 2025-01-27
Make validate-metadata run on all files

### LINKED ISSUE #6875: metadata validation is only running for one query
The [validate-metadata CI logs](https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/44244/workflows/f594eebd-bbcc-4cd6-aa34-71919269d20c/jobs/512850/parallel-runs/0/steps/0-108) show that metadata validation only runs for one query and then exits.

This is because the command invokes the dry run command: https://github.com/mozilla/bigquery-etl/blob/9c3949ec2e170ea105a76f09b9321a4951583a1f/bigquery_etl/cli/query.py#L1383-L1385

which then exits the process: https://github.com/mozilla/bigquery-etl/blob/7024b6976e5f8f2f6e365107b3ef97547ee2ab0f/bigquery_etl/cli/dryrun.py#L100-L102

This doesn't happen in private-bigquery-etl because the CI step runs with `--no-dryrun`: https://github.com/mozilla/private-bigquery-etl/blob/main/.circleci/config.yml#L78-L80

Some things we should probably do:
- replace `sys.exit(0)` with `return` in the dryrun function
- add `--no-dryrun` to the CI step in bigquery-etl since dry running is redundant and would take a long time
- look for other instances in the CLI where `sys.exit` may cause similar unexpected behaviour.  From a quick check, I don't see other uses of `sys.exit(0)` or invocations of dryrun

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7667)

BenWu:
## Description

fixes #6875.  `sys.exit(0)` in the dry run command causes the metadata validation to exit early so it doesn't actually validate more than one metadata file ([example](https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/44363/workflows/f0517bc1-771b-4591-8b97-bd7d914ccdf4/jobs/514580)).

CI step changes:
- added `--no-dryrun` to the `validate-metadata` CI step because it's redundant in CI
- added `--skip-format-sql` option because it takes too long to run for every query
- removed `--respect-dryrun-skip` becaus queries aren't run so it should work

I didn't find any other uses of dryrun or `sys.exit(0)` that would have similar behaviour.

* #6875

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7706)

### DIFF
--- a/bigquery_etl/metadata/validate_metadata.py
+++ b/bigquery_etl/metadata/validate_metadata.py
@@ -2,7 +2,6 @@
 
 import logging
 import os
-import sys
 from argparse import ArgumentParser
 from pathlib import Path
 
@@ -111,6 +110,14 @@ def validate_shredder_mitigation(query_dir, metadata):
 
     if has_shredder_mitigation:
         schema_file = Path(query_dir) / SCHEMA_FILE
+        if not schema_file.exists():
+            click.echo(
+                click.style(
+                    f"Table {query_dir} does not have schema.yaml required for shredder mitigation.",
+                    fg="yellow",
+                )
+            )
+            return False
         schema = Schema.from_schema_file(schema_file).to_bigquery_schema()
 
         # This label requires that the query doesn't have id-level columns,
@@ -281,6 +288,10 @@ def validate_retention_policy_based_on_table_type(metadata, path):
     return is_valid
 
 
+class MetadataValidationError(Exception):
+    """Metadata validation failed."""
+
+
 def validate(target):
     """Validate metadata files."""
     failed = False
@@ -322,11 +333,11 @@ def validate(target):
                     # todo more validation
                     # e.g. https://github.com/mozilla/bigquery-etl/issues/924
     else:
-        logging.error(f"Invalid target: {target}, target must be a directory.")
-        sys.exit(1)
+        raise ValueError(f"Invalid target: {target}, target must be a directory.")
 
     if failed:
-        sys.exit(1)
+        # TODO: add failed checks to message
+        raise MetadataValidationError(f"Metadata validation failed for {target}")
 
 
 def validate_datasets(target):
@@ -340,11 +351,12 @@ def validate_datasets(target):
                     path = os.path.join(root, file)
                     _ = DatasetMetadata.from_file(path)
     else:
-        logging.error(f"Invalid target: {target}, target must be a directory.")
-        sys.exit(1)
+        raise ValueError(f"Invalid target: {target}, target must be a directory.")
 
     if failed:
-        sys.exit(1)
+        raise MetadataValidationError(
+            f"Dataset metadata validation failed for {target}"
+        )
 
 
 def main():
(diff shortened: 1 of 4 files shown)


BenWu:
There are a bunch of metadata validation errors, mostly for shredder mitigation.  cc @lucia-vargas-a

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -322,11 +333,11 @@ def validate(target):
                     # todo more validation
                     # e.g. https://github.com/mozilla/bigquery-etl/issues/924
     else:
-        logging.error(f"Invalid target: {target}, target must be a directory.")
-        sys.exit(1)
+        raise ValueError(f"Invalid target: {target}, target must be a directory.")
 
     if failed:
-        sys.exit(1)
+        # TODO: add failed checks to message
```
edugfilho, line 339: Not sure if this is this something you planned to do later or forgot to add now

scholtzan (approved):
Should we remove/disable the shredder mitigation check for now to get this merged? It'll likely require some more extensive changes anyway if we want to rely on the dryrun function

BenWu:
> Should we remove/disable the shredder mitigation check for now to get this merged?

Yes that makes sense.  I'll check with Lucia

edugfilho (approved):

lucia-vargas-a:
> > Should we remove/disable the shredder mitigation check for now to get this merged?
> 
> Yes that makes sense. I'll check with Lucia

Makes sense to me @BenWu. Let me know if I should help with that.
@wwyc We can expect failures during backfills with mitigation due to missing schemas.

sean-rose:
> I didn't find any other uses of dryrun or `sys.exit(0)` that would have similar behaviour.

If the [invoked `dryrun`](https://github.com/mozilla/bigquery-etl/blob/13b2efed5f9c5bba9f864abb544e1d45ae4702aa/bigquery_etl/cli/query.py#L1388-L1395) fails it [calls `sys.exit(1)`](https://github.com/mozilla/bigquery-etl/blob/13b2efed5f9c5bba9f864abb544e1d45ae4702aa/bigquery_etl/cli/dryrun.py#L132), which also seems like it would prematurely end the overall validation process.

I wonder if there's some better pattern we should be using for indicating success/failure of `click`-powered commands rather than calling `sys.exit()`.  I had hopes maybe using `Context.exit()` instead would work more seamlessly with `Context.invoke()`, but Click doesn't appear to have any special logic for handling that.  I guess we could just catch any `SystemExit` exceptions coming out of `invoke()` calls.

BenWu:
@sean-rose I'm less concerned about `sys.exit(1)` since it will cause an error and catch our attention.  If we want to make improvements with this, I think it would be better to just replace all `sys.exit(1)` with regular exceptions so we can then choose to handle them or fail.

PR #6900, opened 2025-01-28
[DO NOT merge] - DENG-7724 - Set retention on client_level tables related to Addons

alekhyamoz:
As part of this PR, I have added Addons client_level tables.

Set the partition expiration to 775

This PR will be merged after January 28th.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7725)

### DIFF
--- a/sql/moz-fx-data-shared-prod/amo_prod/desktop_addons_by_client_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/amo_prod/desktop_addons_by_client_v1/metadata.yaml
@@ -17,6 +17,7 @@ bigquery:
     type: day
     field: submission_date
     require_partition_filter: false
+    expiration_days: 775
   clustering:
     fields:
       - sample_id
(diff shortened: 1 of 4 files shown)


sql/moz-fx-data-shared-prod/amo_prod/desktop_addons_by_client_v1/metadata.yaml
```diff
@@ -17,6 +17,7 @@ bigquery:
     type: day
     field: submission_date
     require_partition_filter: false
+    expiration_days: 775
```
kik-kik, line 20: I wonder if we should define this value of `755` in the `bqetl_project.yaml` config file instead to ensure that the tables where retention policy is set always have this value. This should allow us to lower the risk of an incorrect value being set.

willdurand (approved):

kik-kik (approved):

scholtzan (approved):

PR #6916, opened 2025-01-30
feat(dryrun): Add `--validate-union-schemas` option (DENG-7746)

sean-rose:
## Description
This PR adds a `--validate-union-schemas` option to the `bqetl dryrun` command, which attempts to dryrun subqueries that are unioned together and check whether there are any discrepancies between their schemas which might indicate the union result is incorrect.

Testing this manually it seems to work pretty well for the majority of ETLs/views, but there are some known cases where it fails:
* It doesn't generate valid dryrun queries for unions in correlated subqueries (i.e. where the subquery being unioned is selecting from an outer scope that isn't a separate CTE).  This might be solvable with some clever query rewriting, but I didn't want to make this already complex PR even more complex for this relatively minor edge case.
* SQLGlot mangles type-annotated array literals ([issue](https://github.com/tobymao/sqlglot/issues/4670)), resulting in some invalid dryrun queries (though that has been [fixed](https://github.com/tobymao/sqlglot/pull/4671) and should be included in the next release).

* DENG-7746: Some ETLs and views may be silently unioning data incorrectly


bigquery_etl/cli/dryrun.py
```diff
@@ -151,4 +167,15 @@ def _sql_file_valid(
...
+    if validate_union_schemas:
+        try:
+            success = result.validate_union_schemas()
+        except Exception:
+            click.echo(
+                f"Failed to validate union schemas in {sqlfile}:\n{traceback.format_exc(limit=0)}",
+                err=True,
+            )
+            success = False
+        return success, sqlfile
```
sean-rose, line 179: I mimicked the behavior of the existing `--validate-schemas` option above, where if the option is set the normal dryrun validation doesn't get run.  I'm not sure if that was the intended behavior for the `--validate-schemas` option, but I figured I wouldn't change the status quo for that in this PR.

This also means the `--validate-schemas` and `--validate-union-schemas` options are mutually exclusive (with the `--validate-schemas` option taking precedence).

bigquery_etl/dryrun.py
```diff
@@ -38,10 +41,15 @@
     from backports.cached_property import cached_property  # type: ignore
 
 
-QUERY_PARAMETER_TYPE_VALUES = {
-    "DATE": "2019-01-01",
-    "DATETIME": "2019-01-01 00:00:00",
-    "TIMESTAMP": "2019-01-01 00:00:00",
+TMP_DATASET = "bigquery-etl-integration-test.tmp"
```
sean-rose, line 44: See ([DSRE-65](https://mozilla-hub.atlassian.net/browse/DSRE-65?focusedCommentId=997755)).

bigquery_etl/cli/dryrun.py
BenWu: I think this check would fit better in `bqetl query validate` since it's validation that doesn't (AFAICT) need to dry run the whole query. The validate command can optionally invoke dry run.  The `validate_schema()` method does seem to do a dry run of the whole query

bigquery_etl/schema/__init__.py
```diff
@@ -17,6 +17,12 @@
 SCHEMA_FILE = "schema.yaml"
 
 
+class SchemaAssertError(Exception):
+    """Schema assert error."""
+
+    pass
```
BenWu, line 23: nit: definitely subjective, but I would just treat the string comment as a `pass`
```suggestion
```

bigquery_etl/schema/__init__.py
```diff
@@ -273,6 +279,64 @@ def _traverse(
...
+                f"Different number of fields{parent_field_note} ({len(fields)} vs {len(other_fields)})."
+            )
+
+        for field_number, (field, other_field) in enumerate(
+            zip(fields, other_fields), start=1
+        ):
+            if field["name"] != other_field["name"]:
+                raise SchemaAssertError(
+                    f"Field #{field_number}{parent_field_note} has different names ({field['name']} vs {other_field['name']})."
+                )
```
BenWu, line 302: Could this be changed to something like
```suggestion
                errors.append(
                    f"Field #{field_number}{parent_field_note} has different names ({field['name']} vs {other_field['name']})."
                )
                continue
```
with an exception being raised at the end so the error contains all the conflicting columns.  That would result in dupes like `a != b` `b != a`, but it would be convenient to see everything in one go.

e.g. In https://github.com/mozilla/bigquery-etl/pull/6878 there were two of these errors in the struct

bigquery_etl/cli/dryrun.py
sean-rose: I had assumed `bqetl query` and its subcommands only worked on ETLs, and with this we also want to check views, but looking at the implementation of `bqetl query validate` it seems like it would also process views by default, so I guess that's workable. @scholtzan do you have an opinion on this?

(on a semi-related note, I don't love that `bqetl query validate` defaults to automatically running `bqetl format`, which can mess with your working copy files when running locally unless you remember to specify `--skip-format-sql`)

scholtzan: Yeah, slight preference to move it to `bqetl query validate`.

PR #6918, opened 2025-01-30
feat: Add distribution_id and is_default_browser to baseline_clients_daily and last_seen templates

kik-kik:
# feat: Add distribution_id and is_default_browser to baseline_clients_daily and last_seen templates

## Description

The following two fields have been introduced to firefox_desktop baseline ping:

- metrics.string.usage_distribution_id
- metrics.boolean.usage_is_default_browser

This information is needed for us for two reasons, 1. we need this information to support our ToS efforts. 2. This is an additional step towards us being able to move away from legacy telemetry.

This change affects two templates:

- baseline_clients_daily_v1 - So that we use the new distribution_id value for coming via baseline for firefox_desktop, and add is_default_browser field (only arriving for firefox_desktop currently).
- baseline_clients_last_seen_v1 - Since it consumes all fields from the daily baseline table.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7753)

### DIFF
--- a/sql_generators/glean_usage/templates/baseline_clients_daily_v1.query.sql
+++ b/sql_generators/glean_usage/templates/baseline_clients_daily_v1.query.sql
@@ -24,15 +24,18 @@ WITH base AS (
     normalized_channel,
     normalized_os,
     normalized_os_version,
-    {% if has_distribution_id %}
+    {% if app_name == "fenix" %}
     metrics.string.metrics_distribution_id AS distribution_id,
+    metrics.string.first_session_install_source AS install_source,
+    CAST(NULL AS BOOLEAN) AS is_default_browser,
+    {% elif app_name == "firefox_desktop" %}
+    metrics.string.usage_distribution_id AS distribution_id,
+    CAST(NULL AS STRING) AS install_source,
+    metrics.boolean.usage_is_default_browser AS is_default_browser,
     {% else %}
     CAST(NULL AS STRING) AS distribution_id,
-    {% endif %}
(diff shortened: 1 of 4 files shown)


scholtzan (approved):

sql_generators/glean_usage/templates/baseline_clients_daily_v1.query.sql
```diff
@@ -24,15 +24,18 @@ WITH base AS (
...
-    metrics.string.metrics_distribution_id AS distribution_id,
-    {% else %}
-    CAST(NULL AS STRING) AS distribution_id,
-    {% endif %}
     {% if app_name == "fenix" %}
+    metrics.string.metrics_distribution_id AS distribution_id,
     metrics.string.first_session_install_source AS install_source,
+    CAST(NULL AS BOOLEAN) AS is_default_browser,
+    {% elif app_name == "firefox_desktop" %}
+    metrics.string.usage_distribution_id AS distribution_id,
```
irrationalagent, line 32: out of curiosity, how did you determine that we need the `usage_` prefix here for desktop? i couldn't determine that it would be there from the patch and it hasn't landed in the glean dictionary yet... is there a schema somewhere that you checked?

kik-kik: I just checked the `firefox_desktop_stable.baseline_v1` table in BQ. And I just assumed this is the correct field here.

PR #6930, opened 2025-01-31
Updates to desktop baseline aggregations

gkatre:
## Description

Updates for baseline clients_daily and clients_last_seen to support Terms of Service project and to standardize on the baseline ping schedule across all platforms

Adding the following fields for `clients_daily`:
- windows_build_number
- browser_engagement_uri_count
- browser_engagement_active_ticks
- legacy_telemetry_client_id
- distribution_id
- is_default_browser

For `clients_last_seen` adding the field:
- days_active_dau_bits

Reference: [Data models for Usage ping](https://docs.google.com/document/d/1puDuEvp3FUWA7lF3kdw19EhyhQ3AIE_3ayXUOivLIRI/edit?tab=t.0#heading=h.rzr7iw7miz6n)
 

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7784)


sql_generators/glean_usage/templates/baseline_clients_last_seen_v1.schema.yaml
```diff
@@ -66,3 +66,6 @@ fields:
 - mode: NULLABLE
   name: install_source
   type: STRING
+- mode: NULLABLE
```
kik-kik, line 69: If I recall correctly the baseline_clients_last_seen query does a select * from the daily table. I think we will also have to include the new fields added to the daily in this schema.

kik-kik: Another note, with how the query is structured the column order between the query result in the schema definition is different. I don't think this will break anything in this case, but I also feel like we usually try to maintain the order across the two.

sql_generators/glean_usage/templates/baseline_clients_last_seen_v1.schema.yaml
```diff
@@ -66,3 +66,6 @@ fields:
...
+  name: days_active_dau_bits
```
kik-kik, line 70: Might be worth including a `description` for this field for clarity on how it is different than the `days_active_bits` field.

kik-kik: @irrationalagent another question about naming, my impression was that generally we wanted to avoid directly relating things to `dau`, is this name fine or should we consider calling it something different?

sql_generators/glean_usage/templates/baseline_clients_daily_v1.query.sql
```diff
@@ -137,6 +152,11 @@ windowed AS (
     udf.mode_last(ARRAY_AGG(install_source) OVER w1) AS install_source,
     udf.mode_last(ARRAY_AGG(geo_subdivision) OVER w1) AS geo_subdivision,
     udf.mode_last(ARRAY_AGG(profile_group_id) OVER w1) AS profile_group_id,
+    udf.mode_last(ARRAY_AGG(windows_build_number) OVER w1) AS windows_build_number,
+    SUM(COALESCE(browser_engagement_uri_count,0)) OVER w1 AS browser_engagement_uri_count,
```
kik-kik, line 156:
```suggestion
    SUM(COALESCE(browser_engagement_uri_count, 0)) OVER w1 AS browser_engagement_uri_count,
```

sql_generators/glean_usage/templates/baseline_clients_daily_v1.query.sql
```diff
@@ -137,6 +152,11 @@ windowed AS (
...
+    SUM(COALESCE(browser_engagement_active_ticks,0)) OVER w1 AS browser_engagement_active_ticks,
```
kik-kik, line 157:
```suggestion
    SUM(COALESCE(browser_engagement_active_ticks, 0)) OVER w1 AS browser_engagement_active_ticks,
```

sql_generators/glean_usage/templates/baseline_clients_daily_v1.query.sql
```diff
@@ -24,22 +24,37 @@ WITH base AS (
...
-    CAST(NULL AS STRING) AS install_source,
-    {% endif %}
     metadata.geo.subdivision1 AS geo_subdivision,
     {% if has_profile_group_id %}
     metrics.uuid.legacy_telemetry_profile_group_id AS profile_group_id,
     {% else %}
     CAST(NULL AS STRING) AS profile_group_id,
     {% endif %}
+    {% if app_name == "firefox_desktop" %}
+    client_info.windows_build_number as windows_build_number,
```
kik-kik, line 34:
```suggestion
    client_info.windows_build_number AS windows_build_number,
```

I believe our SQL style guide mentions that we should capitalize reserved words:
https://docs.telemetry.mozilla.org/concepts/sql_style#reserved-words

sql_generators/glean_usage/templates/baseline_clients_daily_v1.query.sql
```diff
@@ -24,22 +24,37 @@ WITH base AS (
     normalized_channel,
     normalized_os,
     normalized_os_version,
-    {% if has_distribution_id %}
```
kik-kik, line 27: Since this is being removed here, could you also include these changes in your PR?

gkatre: Since this is the only place its used, makes sense to remove it from common.py

sql_generators/glean_usage/templates/baseline_clients_last_seen_v1.schema.yaml
gkatre: > See my comment above, maybe we are ok with "overriding" the current days_active_bits field for desktop, however that would be a breaking change.

I agree it would be a breaking change. Also, we would be overloading the definitions. I am inclined towards keeping them separate.

For naming what about `days_desktop_active_bits` to indicate its mainly for desktop computations only?

irrationalagent: OK `days_desktop_active_bits` is fine by me.

sql_generators/glean_usage/templates/baseline_clients_last_seen_v1.query.sql
```diff
@@ -30,11 +31,19 @@ WITH _current AS (
...
     CAST(TRUE AS INT64) & CAST(durations > 0  AS INT64) AS days_active_bits,
     udf.days_since_created_profile_as_28_bits(
       DATE_DIFF(submission_date, first_run_date, DAY)
     ) AS days_created_profile_bits,
+    {% if app_name == "firefox_desktop" %}
+    CAST(TRUE AS INT64) &
+    CAST(browser_engagement_uri_count > 0 AS INT64) &
+    CAST(browser_engagement_active_ticks > 0 AS INT64) AS days_desktop_active_bits,
+    {% else %}
+    CAST(NULL AS INT64) AS days_desktop_active_bits,
```
kik-kik, line 45: I wonder if we should just remove this else part here? Not sure it makes sense to add `days_desktop_active_bits` which is specifically firefox_desktop only field to every single app. But this would also require additional conditional logic inside `_previous` and in the final statement when we combine _previous with _current in the last_seen  and the corresponding schema templates.

kik-kik: @scholtzan what do you think?

sql_generators/glean_usage/templates/baseline_clients_last_seen_v1.query.sql
```diff
@@ -30,11 +31,19 @@ WITH _current AS (
     -- rightmost bit in 'days_since_seen' represents whether the user sent a
     -- baseline ping in the submission_date and similarly, the rightmost bit in
     -- days_active_bits represents whether the user counts as active on that date.
+    -- days_desktop_active_bits represents the official definition of used for desktop active user
```
irrationalagent, line 34:
```suggestion
    -- days_desktop_active_bits represents the official definition of "active user" for desktop
```

sql_generators/glean_usage/templates/baseline_clients_last_seen_v1.schema.yaml
```diff
@@ -1,12 +1,16 @@
...
 - name: days_active_bits
   type: INTEGER
 - name: days_created_profile_bits
   type: INTEGER
-- name: submission_date
-  type: DATE
+- mode: NULLABLE
+  name: days_desktop_active_bits
+  type: INTEGER
+  description: This represents the official definition used for desktop active user
```
irrationalagent, line 13:
```suggestion
  description: This represents the official definition of "active user" for desktop
```

irrationalagent (approved):
OK, lgtm, once we decide what to do with the desktop bits field per @kik-kik 's comment.

sql_generators/glean_usage/templates/baseline_clients_last_seen_v1.query.sql
scholtzan: Either way works.
Are we planning to track activity in a similar manner for mobile? Having something more generic like `days_active_bits` could be defined across applications.

Depending on how gnarly the logic would get, having it only for desktop, if it should be called `days_desktop_active_bits`, would be my preference.

gkatre: I do not think we are planning to have something similar for mobile. 

I will make the change to apply `days_desktop_active_bits` only for desktop

gkatre: @scholtzan When you get a chance could you do a final review before I merge. Thanks!

scholtzan (approved):

PR #6928, opened 2025-01-30
Add minimum row count check for at least 100 rows

kwindau:
## Description

This PR adds a check for a minimum row count of at least 100 per day.  We were alerted in the data-help channel that this ETL job was silently failing with very small # of rows for over a year, so we'd like to add a row count check to be alerted sooner next time if row counts start to drop.

**Reviewer, please follow [this checklist](https://github.com/mozilla/bigquery-etl/blob/main/.github/reviewer_checklist.md)**

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-7782)

PR #6947, opened 2025-02-03
Delete new check, failing due to extra parameter being passed in

kwindau:
## Description

This PR reverts [PR-6928](https://github.com/mozilla/bigquery-etl/pull/6928) - after implementing, we found that the generated task was passing in an extra argument which caused the data check step to fail.
Screenshot 2025-02-03 at 10 06 13 AM

When the task ran, even though there were more than 100 rows, it still failed due to the argument.
`INFO - [base] BigQuery error in query operation: allowed_schema_update_option requires destination_table.`


chelseatroy (approved):
Thank you :)

jasonthomas (approved):

whd (approved):
Per https://github.com/mozilla/bigquery-etl/pull/6928#pullrequestreview-2587379523 we can't move this to bigeye without change control, so this PR has been updated to remove the errant check for now.

PR #6958, opened 2025-02-04
DENG-5180 (mlops) Add outerbounds flow description to cost per flow View

abhi-agg:
## Description
This PR adds flow description field to the existing `outerbounds_cost_per_flow_v1` view.

The flow descriptions are coming from the `outerbounds_flow_description_v1` BQ table which was created in https://github.com/mozilla/bigquery-etl/pull/6957.

* DENG-5180

### DIFF
--- a/sql/moz-fx-data-shared-prod/monitoring_derived/outerbounds_cost_per_flow_run_v1/view.sql
+++ b/sql/moz-fx-data-shared-prod/monitoring_derived/outerbounds_cost_per_flow_run_v1/view.sql
@@ -30,7 +30,7 @@ SELECT
 FROM
   cost_per_flow_run
 LEFT JOIN
-  `moz-fx-data-shared-prod.monitoring_derived.outerbounds_flow_description_v1` as descriptions
+  `moz-fx-data-shared-prod.monitoring_derived.outerbounds_flow_description_v1` AS descriptions
   ON LOWER(cost_per_flow_run.flow_name) LIKE CONCAT('%', LOWER(descriptions.flow_name), '%')
 ORDER BY
   flow_name DESC
--- a/sql/moz-fx-data-shared-prod/monitoring_derived/outerbounds_cost_per_flow_run_v1/view.sql
+++ b/sql/moz-fx-data-shared-prod/monitoring_derived/outerbounds_cost_per_flow_run_v1/view.sql
@@ -1,36 +1,21 @@
 CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.monitoring_derived.outerbounds_cost_per_flow_run_v1`
 AS
-WITH cost_per_flow_run AS (
-  SELECT
-    DATE_TRUNC(usage_start_time, DAY) AS invoice_day,
-    labels.value AS run_id,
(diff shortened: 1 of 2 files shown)


abhi-agg:
This should be merged only after https://github.com/mozilla/bigquery-etl/pull/6957 has been merged.

chelseatroy (commented):
The query does work. It is quite...slow (it took four and a half minutes when I tried it), which is likely due to having a `LIKE` clause whose match starts with % (so it could be anything; everything needs to be checked). 

I think my move here would be to have the table of descriptions include the fully qualified namespaces for all flows and let the MLEs copy and paste the descriptions across them if they're all meant to be the same so that this match can be exact rather than using `LIKE`.

chelseatroy:
> What do you think?

Huh. You're right! I tried the same query again this morning in the same console, and it took 12 seconds. 

I wonder if there was some kind of `StreamingBuffer` issue going on last night for BigQuery that I didn't know about when I ran the query (I didn't check on BQ's service status last night; I just assumed it was working fine, which in this case might've been a bad assumption). My apologies! 

Honestly at _that_ latency I don't think we have a problem here. I'll approve :)

chelseatroy (approved):
:shipit:

abhi-agg:
@chelseatroy I just realized that the descriptions should be a part of `outerbounds_cost_per_flow_v1` view instead of `outerbounds_cost_per_flow_run_v1` view as the latter one contains multiple runs of the same flow, which means each run of the same flow would have had the same description replicated over.

I changed it and added description to `outerbounds_cost_per_flow_v1` view (updated PR title and description accordingly).

I would appreciate a re-review from you 👍🏾

chelseatroy:
Hey @abhi-agg 

So, when I read this change, I realized that I might have misunderstood something in the ticket.

The ticket description says:

> The new[ MLOps cost dashboard](https://mozilla.cloud.looker.com/dashboards/1851?Flow+Name=&Invoice+Day=1+month&Invoice+Day+%28Inference+nonprod%29=&Invoice+Day+%28Inference+prod%29=) provides a super useful breakdown of operation costs per flow, but it does this relying only on the name of the flow, which is not **descriptive enough to understand what exactly it is doing.**

When I read "what exactly it is doing," I think my brain hears "implementation details." 

If the goal is to record implementation details that could explain the expense of a flow, this goes on the run table. The reason is that, if a flow is costing us _a lot of money_, and the team makes updates to that same flow in order to change that, it's possible that what that flow does (which is recorded in the description) changes as well. Having that description recorded per run allows viewers to see why the cost has gone way down (or way up) on the same flow—for example, suppose a team starts with a simple Markov for predicting something. Then they move onto a much more complicated model to get some necessary evaluation metric up, but it costs a lot more to train. The flow name stays the same, but the underlying model changes between runs. Worth noting that, if _this_ is what we're trying to do, there is an additional change that needs to be made, which is that the table's rows need to be incrementally added, rather than rewriting the whole table each time and left-joining it to the current descriptions in the google sheet.

However, I could be misreading that description, and all we really want is to describe the business goal of each flow.

If _that_ is the case, then absolutely, doing it at the flow level will accomplish that, and also the query would take less time as that's a much smaller table.

abhi-agg:
This https://github.com/mozilla/bigquery-etl/pull/6992 should fix the CI check failure on this PR.

PR #6965, opened 2025-02-04
fix: set submission_date as the partition field

kik-kik:
# fix: set submission_date as the partition field for new_profile_activations template


alekhyamoz (approved):
lgtm

PR #7005, opened 2025-02-08
chore: remove monitoring.payload_bytes_error_structured

whd:
## Description

`_all`  has been authorized.

* DSRE-1877


whd:
Ok thanks for checking usage. I'm now inclined to leave the view as-is (authorized) as it is technically more efficient and already authorized. I suspect unauthorizing will cause a transient failure state where Airflow tries to recreate the view after merge but the dataset-level authorization hasn't been removed yet. I could manually remove it to make sure things work but then I risk breaking those queries in the time between that action and the recreate. Basically I agree it's more effort than it's worth to clean this up.

PR #7009, opened 2025-02-10
DENG-7839 - Backfill baseline_clients_last_seen from 2021-12-01

gkatre:
## Description

We need to backfill the firefox_desktop_derived.baseline_clients_last_seen_v1 table from 2021-12-01 to current date. 
This is a test run for the first month before initiating a longer run.

The requirements for the run is that the **backfill needs to run sequentially on a single thread**

I believe this is possible because:
- There is a check for `depends_on_past` for the [backfills can run sequentially](https://github.com/mozilla/bigquery-etl/blob/main/bigquery_etl/cli/query.py#L828)
- `depends_on_past` appears to be set to true for [baseline_clients_last_seen_v1](https://github.com/mozilla/bigquery-etl/blob/main/sql_generators/glean_usage/templates/baseline_clients_last_seen_v1.metadata.yaml#L18)

* DENG-7839


sql/moz-fx-data-shared-prod/firefox_desktop_derived/baseline_clients_last_seen_v1/backfill.yaml
```diff
@@ -0,0 +1,9 @@
+2025-02-10:
+  start_date: 2021-12-01
```
irrationalagent, line 2: just want to flag that `legacy_telemetry_client_id` and `windows_build_number` will not be present in the data for this month, so we won't be able to validate those metrics from this test.

irrationalagent: I'm assuming we can't test on a more recent month because of the dependency issue?

BenWu (commented):
Managed backfills don't seem to support tables with depends_on_past (DENG-3656) which is why the check is failing.  I'm not sure why they're not supported though. @wwyc do you know?

I think the depends_on_past code you linked is used when you run `bqetl query backfill` outside of a managed backfill.  In this case, you could use the [`bqetl_backfill` dag](https://workflow.telemetry.mozilla.org/dags/bqetl_backfill/grid) but it's more manual and riskier since you need to make sure the params are correct

sql/moz-fx-data-shared-prod/firefox_desktop_derived/baseline_clients_last_seen_v1/backfill.yaml
gkatre: > I'm assuming we can't test on a more recent month because of the dependency issue?

That is correct. I also wanted to get an estimate of what it takes to run for one month especially because the runs need to happen sequentially on a single thread

gkatre:
> > Managed backfills don't seem to support tables with depends_on_past ([DENG-3656](https://mozilla-hub.atlassian.net/browse/DENG-3656)) which is why the check is failing. I'm not sure why they're not supported though. @wwyc do you know?
> 
> Managed backfills currently does not support tables with depends_on_past. Since managed backfills results in a staging table and a different dataset than the original prod table, there needs to be additional logic built in to support tables that depends on past.

@BenWu / @wwyc 
Is this additional logic something that can be done fairly quickly for this backfill?

For the `firefox_desktop_derived.baseline_clients_last_seen_v1` table we need to backfill for 3+ years. 
Do you have any other recommendations? Doing a manual run does not seem ideal. I would appreciate any help!

PR #7024, opened 2025-02-12
feat(DENG-2986): Create rolling_cohorts_v2 & cohort_daily_statistics_v2

kwindau:
## Description
This PR creates 2 new tables, that are now built off of active_users instead of unified_metrics which is being deprecated.
- moz-fx-data-shared-prod.telemetry_derived.rolling_cohorts_v2
- moz-fx-data-shared-prod.telemetry_derived.cohort_daily_statistics_v2

* [DENG-2986](https://mozilla-hub.atlassian.net/browse/DENG-2986)

[DENG-2986]: https://mozilla-hub.atlassian.net/browse/DENG-2986?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dags.yaml
```diff
@@ -2149,3 +2149,21 @@ bqetl_marketing_analysis:
     - impact/tier_3
   repo: bigquery-etl
   schedule_interval: 0 12 * * *
+
+bqetl_daily_retention:
```
kik-kik, line 2153: I wonder if instead of creating a new DAG for just those two queries we could use something like `bqetl_analytics_tables` instead?

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
```diff
@@ -0,0 +1,134 @@
+WITH submission_date_activity AS (
+  SELECT
+    client_id,
+    submission_date AS activity_date
+  FROM
+    `moz-fx-data-shared-prod.telemetry.active_users`
+  WHERE
+    submission_date = @submission_date
+    AND is_dau IS TRUE
+  GROUP BY
```
kik-kik, line 10: Why the group by? We do not seem to be doing any aggregations here?

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
```diff
@@ -0,0 +1,134 @@
...
+    client_id,
+    submission_date
```
kik-kik, line 12:
```suggestion
    activity_date
```

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
```diff
@@ -0,0 +1,134 @@
...
+  FROM
+    `moz-fx-data-shared-prod.telemetry_derived.rolling_cohorts_v2`
+  WHERE
+    cohort_date
+    BETWEEN DATE_SUB(@submission_date, INTERVAL 180 DAY)
+    AND DATE_SUB(@submission_date, INTERVAL 1 DAY)
+),
+activity_cohort_match AS (
+  SELECT
+    cohorts_in_range.client_id AS cohort_client_id,
```
kik-kik, line 58: What's the difference between the two client_ids? We're using it as a join key so the two values would be the same, potentially on the right side this would be null?

sql/moz-fx-data-shared-prod/telemetry_derived/rolling_cohorts_v2/query.sql
```diff
@@ -0,0 +1,176 @@
...
+  CAST(NULL AS STRING) AS play_store_attribution_term,
+FROM
+  `moz-fx-data-shared-prod.telemetry.desktop_active_users` au
+LEFT JOIN
+  `moz-fx-data-shared-prod.telemetry.core_clients_last_seen` ccls
+  ON au.client_id = ccls.client_id
+  AND au.submission_date = ccls.submission_date
+LEFT JOIN
+  `moz-fx-data-shared-prod.telemetry.clients_last_seen` cls
+  ON au.client_id = cls.client_id
```
kik-kik, line 117: Looks like we could make this less verbose by using `USING` here instead.

dags.yaml
kwindau: Personally I would prefer a new DAG if that's OK - I like having a separate dag for logically related items, if that makes sense?

sql/moz-fx-data-shared-prod/telemetry_derived/rolling_cohorts_v2/query.sql
kwindau: Will change, good catch

sql/moz-fx-data-shared-prod/telemetry_derived/rolling_cohorts_v2/query.sql
```diff
@@ -0,0 +1,176 @@
...
+  au.first_seen_date = @submission_date
+  AND au.submission_date = @submission_date
+UNION ALL
+--Mobile
+SELECT
+  au.client_id,
+  au.first_seen_date AS cohort_date,
+  au.activity_segment,
+  au.app_display_version AS app_version,
+  CAST(NULL AS string) AS attribution_campaign,
```
kik-kik, line 129:
```suggestion
  CAST(NULL AS STRING) AS attribution_campaign,
```

sql/moz-fx-data-shared-prod/telemetry_derived/rolling_cohorts_v2/query.sql
```diff
@@ -0,0 +1,176 @@
...
+  CAST(NULL AS string) AS attribution_medium,
+  CAST(NULL AS string) AS attribution_source,
+  CAST(NULL AS string) AS attribution_variation,
+  au.city,
+  au.country,
+  au.device_model,
+  au.distribution_id,
+  dflt_brwsr.is_default_browser,
+  au.locale,
+  au.app_name AS normalized_app_name, --do I need to do anything to "normalize" ?
```
kik-kik, line 141: Let's stay consistent and just call it `app_name`. As you can see in the active_users view this field has its name changed to include MozillaOnline or BrowserStack to allow us to easily filter those out:
https://github.com/mozilla/bigquery-etl/blob/generated-sql/sql/moz-fx-data-shared-prod/fenix/active_users/view.sql#L10-L16

sql/moz-fx-data-shared-prod/telemetry_derived/rolling_cohorts_v2/query.sql
```diff
@@ -0,0 +1,176 @@
...
+    FROM
+      `moz-fx-data-shared-prod.focus_ios_derived.metrics_clients_last_seen_v1`
+    WHERE
+      submission_date = @submission_date
+    UNION ALL
+    SELECT
+      a.submission_date,
+      a.client_id,
+      a.is_default_browser,
+      IF(b.isp = 'BrowserStack', CONCAT('Fenix', ' BrowserStack'), 'Fenix') AS normalized_app_name
```
kik-kik, line 18: I'd suggest you reuse exactly the same logic from active_users for setting this value to make sure it can be used as a reliable join key:
https://github.com/mozilla/bigquery-etl/blob/generated-sql/sql/moz-fx-data-shared-prod/fenix/active_users/view.sql#L10-L16

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
kwindau: It's basically to get a distinct of the client IDs and their dates where they were is_DAU (I could have written as select distinct instead if you prefer, either one works)

sql/moz-fx-data-shared-prod/telemetry_derived/rolling_cohorts_v2/query.sql
kik-kik: Alternatively, at this point this could also become a UDF we could reuse everywhere including inside active_users. This would allow us to ensure this logic is consistent in all places whilst being less verbose and less prone to code drift.

dags.yaml
kik-kik: I'm not strongly opposed, maybe only ask from my side would be to come up with a different name. I don't see retention models used anywhere in these queries. Maybe something like `bqetl_cohort_aggregations` or something like that?

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
kwindau: I will write it as a select distinct so it's more obvious what it's doing to readers

dags.yaml
kik-kik: ok aggregates might not be ideal in the name since one of the tables is client level. `bqetl_profile_cohorts`? I'm not sure what the best name is, I hope you understand my concern 😓

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/metadata.yaml
```diff
@@ -0,0 +1,27 @@
...
+  It's a summary of the cohort's activity in that day.
+
+  Note that the values for client attributes are based on the
+  attributes at the time the cohort started (rather than the
+  values at activity time)
+owners:
+- kwindau@mozilla.com
+labels:
+  incremental: true
+  owner1: kwindau
```
kik-kik, line 14: Should we also include `table_type: aggregate` label here?

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/query.sql
kwindau: So basically, imagine that rolling cohorts is a table with the PK of first seen date, and client IDs on that first seen date.  Then it left joins to see who was active each day out of that cohort for the next 180 days.  So if the client ID is null on the activity date from the right table, that means the client who was first seen on that cohort date was not active on that date - hope that helps!

sql/moz-fx-data-shared-prod/telemetry_derived/rolling_cohorts_v2/query.sql
kwindau: Good catch, thanks!

sql/moz-fx-data-shared-prod/telemetry_derived/cohort_daily_statistics_v2/metadata.yaml
kwindau: Good catch, added

dags.yaml
kwindau: I named it originally "bqetl_daily_retention" because it's basically retention for the first 180 days after each first seen date for each cohort on each first seen date - that's essentially what the cohort_daily_statistics table is doing - looking at # of clients active by day for each day of the 180 days after a first seen date.

sql/moz-fx-data-shared-prod/telemetry_derived/rolling_cohorts_v2/query.sql
kwindau: Thank you, I was trying to find where that logic was!

sql/moz-fx-data-shared-prod/telemetry_derived/rolling_cohorts_v2/query.sql
kwindau: So the initial reason I'm renaming it to normalized_app_name is I need to keep the column names the same as the old one so that the dashboards can easily swap out - but I will definitely take your advice and use the logic you sent, thanks for sharing!

dags.yaml
kwindau: Happy to rename to like bqetl_cohort_retention if that's more clear - I think that's a good inbetween - will go with that

Marlene-M-Hirose (approved):

Issue #5711, opened 2024-06-03
Python 3.12 bootstrap - ModuleNotFoundError: No module named 'distutils'


ahal:
The below traceback occurs when running `./bqetl bootstrap` with Python 3.12. The `distutils` module was removed in 3.12 and it looks like `bigquery-etl` depends on an older version of `setuptools` (via `pip-tools` and `nodeenv`) that will need to get updated.

Unfortunately it looks like `pip-tools` doesn't support 3.12 yet (?). So maybe the fix here is to change the supported Python version from `>= 3.11` to `==3.11`.

```
Collecting numpy==1.24.2 (from -r requirements.txt (line 958))
  Downloading numpy-1.24.2.tar.gz (10.9 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 10.9/10.9 MB 42.7 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
ERROR: Exception:
Traceback (most recent call last):
  File "/home/ahal/dev/bigquery-etl/venv/lib64/python3.12/site-packages/pip/_internal/cli/base_command.py", line 180, in exc_logging_wrapper
    status = run_func(*args)
             ^^^^^^^^^^^^^^^
  File "/home/ahal/dev/bigquery-etl/venv/lib64/python3.12/site-packages/pip/_internal/cli/req_command.py", line 248, in wrapper
    return func(self, options, args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ahal/dev/bigquery-etl/venv/lib64/python3.12/site-packages/pip/_internal/commands/install.py", line 377, in run
    requirement_set = resolver.resolve(
                      ^^^^^^^^^^^^^^^^^
  File "/home/ahal/dev/bigquery-etl/venv/lib64/python3.12/site-packages/pip/_internal/resolution/resolvelib/resolver.py", line 92, in resolve
    result = self._result = resolver.resolve(
                            ^^^^^^^^^^^^^^^^^
  File "/home/ahal/dev/bigquery-etl/venv/lib64/python3.12/site-packages/pip/_vendor/resolvelib/resolvers.py", line 546, in resolve
    state = resolution.resolve(requirements, max_rounds=max_rounds)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #7025, opened 2025-02-12
fix: Mark `bqetl` as not compatible with Python 3.12 or later

sean-rose:
## Description
`bqetl` isn't currently compatible with Python 3.12 or 3.13, so this PR updates `pyproject.toml` to indicate that.

* https://github.com/mozilla/bigquery-etl/issues/5711

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,9 +9,9 @@ authors = [
 ]
 description = "Tooling for building derived datasets in BigQuery"
 readme = "README.md"
-requires-python = ">=3.11"
+requires-python = ">=3.11, <3.12"
 dynamic = ["dependencies"]
-version = "2025.1.1"
+version = "2025.2.1"
 
 [project.urls]
 Homepage = "https://github.com/mozilla/bigquery-etl"


pyproject.toml
```diff
@@ -9,9 +9,9 @@ authors = [
 ]
 description = "Tooling for building derived datasets in BigQuery"
 readme = "README.md"
-requires-python = ">=3.11"
+requires-python = ">=3.11, <3.12"
```
BenWu, line 12: Does this only support min/max specifications, i.e. can't do
```suggestion
requires-python = "=3.11"
```

BenWu (commented):
The top-level readme should also be changed to say 3.11 instead of 3.11+

curtismorales (approved):

pyproject.toml
BenWu: Ok that makes sense. > and < works for me

sean-rose: Unfortunately `./bqetl bootstrap` [installs `bqetl`'s dependencies](https://github.com/mozilla/bigquery-etl/blob/d213edbec6f909c2e475d6a8f567aedf2c677224/bqetl#L27) before installing `bqetl` itelf, so issue #5711 still happens even with this in place.

I'm thinking adding a Python version check to the `./bqetl` script would probably be the most user-friendly, but the downside is we'd then have to remember to maintain that Python version check logic.

kik-kik (approved):

pyproject.toml
BenWu: We could have a checklist of things to do when changing the supported versions, e.g. update Dockerfile, pyproject.toml, CI, and bqetl bootstrap, but I think having any check with a clear message is an improvement

PR #7032, opened 2025-02-13
feat: handle cases when SQL syntaxt is invalid and add invalid to the final report

kik-kik:
# feat: handle cases when SQL syntaxt is invalid and add invalid to the final report

Currently, if SQL is found to be invalid we end up with an exception being raised when running bqetl format command and an exception stack trace being printed.

This change is meant to handle these cases more gracefully.

### DIFF
--- a/bigquery_etl/format_sql/format.py
+++ b/bigquery_etl/format_sql/format.py
@@ -8,6 +8,8 @@ from functools import partial
 from multiprocessing.pool import Pool
 from pathlib import Path
 
+from sqlglot.errors import ParseError
+
 from bigquery_etl.config import ConfigLoader
 from bigquery_etl.format_sql.formatter import reformat  # noqa E402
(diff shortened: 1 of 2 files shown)


BenWu (commented):
LGTM I just think there should be a small comment explaining the return value

bigquery_etl/format_sql/format.py
```diff
@@ -43,18 +54,22 @@ def _format_path(check, path):
             fully_referenced_query = query
     except NotImplementedError:
         fully_referenced_query = query  # not implemented for scripts
+    except ParseError:
+        print(f"Invalid syntax found for: {path}")
```
sean-rose, line 58: If the parse error indicates what the actual syntax error is then it'd be helpful to surface that, like maybe
```suggestion
    except ParseError as e:
        print(f"Invalid syntax found for {path}: {e}")
```

PR #7042, opened 2025-02-14
chore(org_mozilla_broken_site_report): Drop the JSON-building from the views.

denschub:
## Description

We originally did that to keep compatibility with another data format that we planned to import. We never imported that data, and we're actually destructing the JSON again on the receiver side, so this isn't super useful.

Exposing all metrics as the new `details` "column" (i.e. record) will allow us to consume all data on our end, but we don't have to maintain that large block of boilerplate code.

This will break existing consumers of the details values - that's fine, as far as we know, we're in full control over everything.

r? @scholtzan

n/a

### DIFF
--- a/sql/moz-fx-data-shared-prod/org_mozilla_broken_site_report/user_reports/view.sql
+++ b/sql/moz-fx-data-shared-prod/org_mozilla_broken_site_report/user_reports/view.sql
@@ -11,59 +11,7 @@ SELECT
   client_info.app_display_version AS app_version,
   normalized_channel AS app_channel,
   normalized_os AS os,
-  TO_JSON_STRING(
-    STRUCT(
-      STRUCT(
-        STRUCT(
-          metrics.text2.broken_site_report_browser_info_app_default_useragent_string AS default_useragent_string,
-          metrics.boolean.broken_site_report_browser_info_app_fission_enabled AS fission_enabled,
-          metrics.string_list.broken_site_report_browser_info_app_default_locales AS app_default_locales
-        ) AS app,
-        STRUCT(
-          metrics.text2.broken_site_report_browser_info_graphics_devices_json AS devices_json,
-          metrics.text2.broken_site_report_browser_info_graphics_drivers_json AS drivers_json,
-          metrics.text2.broken_site_report_browser_info_graphics_features_json AS features_json,
-          metrics.text2.broken_site_report_browser_info_graphics_monitors_json AS monitors_json,
-          metrics.boolean.broken_site_report_browser_info_graphics_has_touch_screen AS has_touch_screen,
-          metrics.string.broken_site_report_browser_info_graphics_device_pixel_ratio AS device_pixel_ratio
-        ) AS graphics,
-        STRUCT(
-          metrics.boolean.broken_site_report_browser_info_prefs_software_webrender AS software_webrender,
-          metrics.boolean.broken_site_report_browser_info_prefs_global_privacy_control_enabled AS global_privacy_control_enabled,
-          metrics.boolean.broken_site_report_browser_info_prefs_installtrigger_enabled AS installtrigger_enabled,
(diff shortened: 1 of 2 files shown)


scholtzan (approved):

sql/moz-fx-data-shared-prod/org_mozilla_broken_site_report/user_reports/view.sql
```diff
@@ -4,66 +4,14 @@ AS
...
-          metrics.boolean.broken_site_report_tab_info_antitracking_has_mixed_display_content_blocked AS has_mixed_display_content_blocked,
-          metrics.boolean.broken_site_report_tab_info_antitracking_is_private_browsing AS is_private_browsing,
-          metrics.boolean.broken_site_report_tab_info_antitracking_has_mixed_active_content_blocked AS has_mixed_active_content_blocked,
-          metrics.boolean.broken_site_report_tab_info_antitracking_has_tracking_content_blocked AS has_tracking_content_blocked,
-          metrics.string.broken_site_report_tab_info_antitracking_block_list AS block_list
-        ) AS antitracking
-      ) AS tab_info
-    )
-  ) AS details
+  metrics as details
```
scholtzan, line 14:
```suggestion
  metrics AS details
```

sql/moz-fx-data-shared-prod/org_mozilla_broken_site_report/user_reports_live/view.sql
```diff
@@ -23,66 +23,14 @@ WITH live_reports AS (
...
-          metrics.boolean.broken_site_report_tab_info_antitracking_has_mixed_display_content_blocked AS has_mixed_display_content_blocked,
-          metrics.boolean.broken_site_report_tab_info_antitracking_is_private_browsing AS is_private_browsing,
-          metrics.boolean.broken_site_report_tab_info_antitracking_has_mixed_active_content_blocked AS has_mixed_active_content_blocked,
-          metrics.boolean.broken_site_report_tab_info_antitracking_has_tracking_content_blocked AS has_tracking_content_blocked,
-          metrics.string.broken_site_report_tab_info_antitracking_block_list AS block_list
-        ) AS antitracking
-      ) AS tab_info
-    )
-  ) AS details
+  metrics as details
```
scholtzan, line 33:
```suggestion
  metrics AS details
```

sql/moz-fx-data-shared-prod/org_mozilla_broken_site_report/user_reports_live/view.sql
```diff
@@ -23,66 +23,14 @@ WITH live_reports AS (
 SELECT
   document_id AS uuid,
   CAST(submission_timestamp AS DATETIME) AS reported_at,
-  metrics.text2.broken_site_report_description AS comments,
-  metrics.url2.broken_site_report_url AS url,
+  metrics.text.broken_site_report_description AS comments,
+  metrics.url.broken_site_report_url AS url,
```
scholtzan, line 27: Stage deploys are failing with 
```
Field name text does not exist in STRUCT
```

Only seems to have `text2`

denschub: Ah, true, the `moz-fx-data-shared-prod.firefox_desktop_live.broken_site_report_v1` table doesn't get the rename-logic that `moz-fx-data-shared-prod.firefox_desktop.broken_site_report` has. urgh. I'll revert the change to the live-table only :(

denschub: Updated the PR.

denschub:
Seems like it worked now. I don't think the one failing job is something I can fix, that's probably me using a fork and not having write-permissions here?

PR #7044, opened 2025-02-14
[DENG-7848] Add retention to backfills_staging_derived and add check

BenWu:
## Description

This sets default table expiry for `backfills_staging_derived` to 30 days.  This only applies to new tables so the existing tables won't be affected

A check for old backfills still in initiate state and an option to skip old backfills in the schedule command were added. Checks are set to 28 days to add some buffer.

* DENG-7848


BenWu:
https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/telemetry_derived/sponsored_tiles_clients_daily_v1/backfill.yaml is causing the validation to fail so that needs to be sorted out before merging this

PR #7045, opened 2025-02-14
DENG-7601 - add more columns to table_storage_trends view

kwindau:
## Description

This PR adds current_physical_bytes to the table_storage_trends view.

* [DENG-7601](https://mozilla-hub.atlassian.net/browse/DENG-7601)

[DENG-7601]: https://mozilla-hub.atlassian.net/browse/DENG-7601?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Marlene-M-Hirose (approved):

BenWu (approved):
This should be good enough for now but in the future we might want to use this view for checking things apart from the impact of retention.  In that case we would probably want the rest of the byte counts like total physical, time travel, and logical to make it easier to track costs or compare cost growth of logical vs physical.  We don't need to add these to the view now but just something to keep in mind

kwindau:
> This should be good enough for now but in the future we might want to use this view for checking things apart from the impact of retention. In that case we would probably want the rest of the byte counts like total physical, time travel, and logical to make it easier to track costs or compare cost growth of logical vs physical. We don't need to add these to the view now but just something to keep in mind

I can add it all, I might as well - that way the Looker explore will have a bunch of different options in case anyone wants to deep dive.  Will add more - thanks Ben!

PR #7050, opened 2025-02-18
feat: mark old retention related datasets as deprecated

kik-kik:
# feat: mark old retention related datasets as deprecated

## Description

Part of clean up as a follow up to creating the new kpi_support_metrics generator (mobile acquisition models). This includes datasets for getting retention and new profile & their activation data.

Once this is merged, following also need to be done:

- Any references, configuration that breaks as the result of the deprecation needs to be updated to point to the new tables or deleted (for example Looker).
- Create a PR to delete corresponding views.
- Delete those views from BigQuery.
- Agree if we can drop those tables from BigQuery, if so drop them.
- If dropped in BigQuery we should also delete the entire dataset directory from BigQuery ETL.


sql/moz-fx-data-shared-prod/fenix_derived/funnel_retention_clients_week_2_v1/metadata.yaml
```diff
@@ -8,9 +10,10 @@ labels:
   schedule: daily
   incremental: true
   table_type: client_level
-scheduling:
-  dag_name: bqetl_analytics_tables
-  depends_on_past: false
+# @kik 2025-02-18: descheduled due to deprecation.
```
kik-kik, line 13: We could consider unscheduling as one of the follow up steps if we think this data may still be in use.

soGaussian (approved):

PR #7062, opened 2025-02-19
DENG-6770: update survey schema with parent sub-field

ogzillla:
## Description
Extends the schema for survey data with a new sub-field, `parent`.
This is a fix for [this](https://workflow.telemetry.mozilla.org/dags/bqetl_alchemer/grid?dag_run_id=scheduled__2025-02-18T12%3A00%3A00%2B00%3A00&tab=logs&task_id=fx_quant_user_research_derived__fxqur_viewpoint_desktop__v1) failing DAG

https://mozilla-hub.atlassian.net/browse/DENG-7875


bigquery_etl/alchemer/response.schema.json
```diff
@@ -41,6 +41,11 @@
         "name": "type",
         "mode": "NULLABLE"
       },
+      {
+        "type": "STRING",
```
sean-rose, line 45: All the example `parent` values in [Alchemer's docs](https://apihelp.alchemer.com/help/surveyresponse-per-question-v5) are integers, so I'm pretty sure this would be better stored as an integer.
```suggestion
        "type": "INTEGER",
```

sean-rose (approved):
r+wc

bigquery_etl/alchemer/response.schema.json
sean-rose: Yes, BigQuery will try to coerce types to the defined schema when loading data, but the load will fail with an error if such coercion fails.

It's unfortunate that the Alchemer API seems to be a bit inconsistent with how it encodes fields.  Their [docs](https://apihelp.alchemer.com/help/surveyresponse-returned-fields-v5#getobject) specify `survey_data[id]` as "numeric" but in your example it's being passed as a string, while `survey_data[section_id]` is also specified as "numeric" and in your example it's being passed as a number literal.

I'd still lean toward defining it as an integer field in BigQuery.

PR #7068, opened 2025-02-20
feat(DENG-7860): Move baseline active users aggregates to different namespace and DAG

kik-kik:
feat(DENG-7860): Move baseline active users aggregates to different namespace and dag

## Description

Two main changes vs the original change:

1. Placing this model inside `firefox_desktop_derived` because this is based on baseline (GLEAN) and this is the corresponding GLEAN namespace.
2. Moving the query to be schedule via: `bqetl_analytics_aggregations` since that DAG already schedules all of the existing `active_users_aggregates` queries.


sql/moz-fx-data-shared-prod/firefox_desktop/baseline_active_users/view.sql
```diff
@@ -2,62 +2,79 @@ CREATE OR REPLACE VIEW
...
-  IFNULL(mozfun.bits28.days_since_seen(days_seen_bits) = 0, FALSE) AS is_daily_user,
-  IFNULL(mozfun.bits28.days_since_seen(days_seen_bits) < 7, FALSE) AS is_weekly_user,
-  IFNULL(mozfun.bits28.days_since_seen(days_seen_bits) < 28, FALSE) AS is_monthly_user,
-  IF(
-    LOWER(IFNULL(isp, '')) <> "browserstack"
-    AND LOWER(IFNULL(distribution_id, '')) <> "mozillaonline",
-    TRUE,
-    FALSE
-  ) AS is_included_kpi
+  END AS activity_segment,
```
kik-kik, line 71: @irrationalagent I'm not sure this naming is really appropriate here, active_segment is based off days_seen_bits which can cause some confusion. Do we need both segments (segment_dau and activity_segment)?

sql/moz-fx-data-shared-prod/firefox_desktop_derived/baseline_active_users_aggregates_v1/metadata.yaml
```diff
@@ -0,0 +1,23 @@
...
+bigquery:
+  time_partitioning:
+    type: day
+    field: submission_date
+    require_partition_filter: true
+    expiration_days: null
+  range_partitioning: null
+  clustering:
+    fields:
+    - app_name
```
kik-kik, line 21: @irrationalagent I guess this clustering makes sense? Usually from high level we'd filter on app_name and then a lot of our segmentation would happen on channel and country level, right?

sql/moz-fx-data-shared-prod/firefox_desktop/baseline_active_users/view.sql
```diff
@@ -2,62 +2,79 @@ CREATE OR REPLACE VIEW
...
       THEN CONCAT('Firefox Desktop', ' ', distribution_id)
     ELSE 'Firefox Desktop'
   END AS app_name,
+  app_display_version AS app_version,
+  normalized_channel AS channel,
+  normalized_os AS os,
+  COALESCE(
+    `mozfun.norm.windows_version_info`(normalized_os, normalized_os_version, windows_build_number),
+    normalized_os_version
+  ) AS os_version_build,
```
kik-kik, line 29: What's the difference between this and normalized_os_version?

sql/moz-fx-data-shared-prod/firefox_desktop/baseline_active_users/view.sql
irrationalagent: Good call out. In our current logic for [desktop_active_users](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/telemetry/desktop_active_users/view.sql#L9-L24) activity segment is based only on days_active_bits, which is analogous to days_desktop_active_bits here. So, I suppose we should mirror that behavior and have a single `activity_segment` column based on `days_desktop_active_bits`. 

If people need to segment usage history based on pings rather than DAU, then days_seen_bits is still available for them to do that manually.

sql/moz-fx-data-shared-prod/firefox_desktop_derived/baseline_active_users_aggregates_v1/metadata.yaml
irrationalagent: this seems to mirror the current clustering for AUA, and I can't think of a great reason for changing it

sql/moz-fx-data-shared-prod/firefox_desktop/baseline_active_users/view.sql
kik-kik: @irrationalagent ok, I updated the query. dau_segment was removed and activity_segment is using `days_desktop_active_bits` now.

sql/moz-fx-data-shared-prod/firefox_desktop_derived/baseline_active_users_aggregates_v1/query.sql
kwindau, line 1: Do you need to have @submission_date parameter here?

kik-kik: Thanks, indeed that was missing. Added it in now.

sql/moz-fx-data-shared-prod/firefox_desktop/baseline_active_users/view.sql
```diff
@@ -2,31 +2,39 @@ CREATE OR REPLACE VIEW
...
+  COALESCE(
+    CAST(NULLIF(SPLIT(normalized_os_version, ".")[SAFE_OFFSET(0)], "") AS INTEGER),
+    0
+  ) AS os_version_major,
+  COALESCE(
```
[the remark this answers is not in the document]
irrationalagent: Could we instead use [this udf here](https://github.com/mozilla/bigquery-etl/blob/main/sql/mozfun/norm/truncate_version/udf.sql)? That looks like it would return `null` in most cases where it would make sense to do so. (unless null values here are problematic for other reasons im not seeing)

sql/moz-fx-data-shared-prod/firefox_desktop/baseline_active_users/view.sql
irrationalagent: I **think**  that normalized_os_version is more or less the same thing that's reported by the client. IIUC, that's not enough to tell windows versions apart for windows 10+. os_version_build derives the "accurate" windows version by referencing the build number.

kik-kik: This seems to be windows specific info, should we rename this field to reflect that more?

sql/moz-fx-data-shared-prod/firefox_desktop/baseline_active_users/view.sql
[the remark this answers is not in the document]
kik-kik: Good call out, I will update this.

sql/moz-fx-data-shared-prod/firefox_desktop/baseline_active_users/view.sql
```diff
@@ -2,31 +2,34 @@ CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.firefox_desktop.baseline_active_users`
 AS
 SELECT
-  *,
+  * EXCEPT (
+    app_display_version,
+    normalized_channel,
+    normalized_os,
+    normalized_os_version,
+    windows_build_number
```
irrationalagent, line 10: I noticed that `windows_build_number` actually makes it into the view for the existing [desktop_active_users](https://mozilla.acryl.io/dataset/urn:li:dataset:(urn:li:dataPlatform:bigquery,moz-fx-data-shared-prod.telemetry.desktop_active_users,PROD)/Columns?highlightedPath&is_lineage_mode=false&schemaFilter=) table. So to keep things consistent with that, maybe we should include it in the client-level view here. 

[PR 7077](https://github.com/mozilla/bigquery-etl/pull/7077/files) will also means that it will make it into the view for [usage_reporting_active_users](https://github.com/mozilla/bigquery-etl/blob/main/sql_generators/usage_reporting/templates/usage_reporting_active_users.view.sql.jinja#L6). In other words, if we include it here, we will have parity with the usage ping version of the table.

HOWEVER, I think in both cases we should NOT add it to the aggregates. Does that make sense?

irrationalagent (approved):
r+wc, see my comment about including windows_build_number for the client level view.

sql/moz-fx-data-shared-prod/firefox_desktop/baseline_active_users/view.sql
kik-kik: ok, I added `windows_build_number` to baseline_active_users. The field is NOT present in the aggregate query.

PR #7069, opened 2025-02-20
feat: update references in the funnels generator for funnel_retention to use the new retention_clients dataset

kik-kik:
# feat: update references in the funnels generator for funnel_retention to use the new retention_clients dataset

blocks: https://github.com/mozilla/bigquery-etl/pull/7050


sql_generators/funnels/configs/android_onboarding.toml
```diff
@@ -189,19 +188,24 @@ aggregation = "count distinct"
 [data_sources.onboarding_events]
 from_expression = """
```
soGaussian, line 189: This table needs to be deprecated as well. Kindly start using the new profiles tables you built going forward

sql_generators/funnels/configs/ios_onboarding_main.toml
```diff
@@ -319,23 +318,26 @@ aggregation = "count distinct"
 [data_sources.onboarding_events]
 from_expression = """
   firefox_ios.firefox_ios_clients ic --each client_id has only one row
```
soGaussian, line 320: Again lets change this table as well to the new new profiles tables

sql_generators/funnels/configs/android_onboarding.toml
kik-kik: Hey @soGaussian , I agree we need to update this but I suggest we do this as a separate follow-up PR.

PR #7070, opened 2025-02-20
Add logic to enforce column descriptions on new tables

kwindau:
## Description
- This PR adds a check to ensure new tables have column descriptions enforced (this check can be turned off if needed).

- When someone runs `./bqetl query create` it will now add a new key to the metadata.yaml file called: `require_column_descriptions=True`

- If this is left as True, when the metadata validation step runs, it will check to ensure column descriptions are on all tables where this is set to True.  If you set to False, it will not run this check.

- Note: In tables where this is not specified, the default value for require_column_descriptions is False

* [DENG-7908](https://mozilla-hub.atlassian.net/browse/DENG-7908)

[DENG-7908]: https://mozilla-hub.atlassian.net/browse/DENG-7908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bigquery_etl/metadata/parse_metadata.py
+++ b/bigquery_etl/metadata/parse_metadata.py
@@ -184,6 +184,7 @@ class Metadata:
     deprecated: bool = attr.ib(False)
     deletion_date: Optional[date] = attr.ib(None)
     monitoring: Optional[MonitoringMetadata] = attr.ib(None)
+    enforce_col_desc: bool = attr.ib(False)
 
     @owners.validator
     def validate_owners(self, attribute, value):
@@ -261,6 +262,7 @@ class Metadata:
         deprecated = False
         deletion_date = None
         monitoring = None
+        enforce_col_desc = False
 
         with open(metadata_file, "r") as yaml_stream:
             try:
@@ -343,6 +345,9 @@ class Metadata:
                         # column needs to be set explicitly
                         monitoring.partition_column_set = True
 
+                if "enforce_col_desc" in metadata:
+                    enforce_col_desc = metadata["enforce_col_desc"]
+
                 return cls(
                     friendly_name,
                     description,
@@ -357,6 +362,7 @@ class Metadata:
                     deprecated,
                     deletion_date,
                     monitoring,
+                    enforce_col_desc,
                 )
             except yaml.YAMLError as e:
                 raise e
--- a/bigquery_etl/metadata/parse_metadata.py
+++ b/bigquery_etl/metadata/parse_metadata.py
@@ -184,7 +184,7 @@ class Metadata:
     deprecated: bool = attr.ib(False)
     deletion_date: Optional[date] = attr.ib(None)
     monitoring: Optional[MonitoringMetadata] = attr.ib(None)
-    enforce_col_desc: bool = attr.ib(False)
+    require_column_descriptions: bool = attr.ib(False)
 
     @owners.validator
     def validate_owners(self, attribute, value):
@@ -262,7 +262,7 @@ class Metadata:
         deprecated = False
         deletion_date = None
         monitoring = None
-        enforce_col_desc = False
+        require_column_descriptions = False
 
         with open(metadata_file, "r") as yaml_stream:
             try:
@@ -345,8 +345,10 @@ class Metadata:
                         # column needs to be set explicitly
                         monitoring.partition_column_set = True
 
-                if "enforce_col_desc" in metadata:
-                    enforce_col_desc = metadata["enforce_col_desc"]
+                if "require_column_descriptions" in metadata:
+                    require_column_descriptions = metadata[
+                        "require_column_descriptions"
+                    ]
 
                 return cls(
                     friendly_name,
@@ -362,7 +364,7 @@ class Metadata:
                     deprecated,
                     deletion_date,
                     monitoring,
-                    enforce_col_desc,
+                    require_column_descriptions,
                 )
             except yaml.YAMLError as e:
                 raise e
--- a/bigquery_etl/metadata/validate_metadata.py
+++ b/bigquery_etl/metadata/validate_metadata.py
@@ -207,7 +207,7 @@ def validate_shredder_mitigation(query_dir, metadata):
 def validate_col_desc_enforced(query_dir, metadata):
     """Check schemas with enforce_col_desc = True comply with requirements."""
     # If column descriptions need to be enforced:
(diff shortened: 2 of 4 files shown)


bigquery_etl/metadata/parse_metadata.py
```diff
@@ -343,6 +345,9 @@ def from_file(cls, metadata_file):
                         # column needs to be set explicitly
                         monitoring.partition_column_set = True
 
+                if "enforce_col_desc" in metadata:
+                    enforce_col_desc = metadata["enforce_col_desc"]
```
scholtzan, line 349: Can we rename this to `enforce_column_descriptions`. Generally, we have been using more descriptive names in the metadata files (even though they are a little bit longer)

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -204,6 +204,35 @@ def validate_shredder_mitigation(query_dir, metadata):
...
+            )
+            return False
+        schema = Schema.from_schema_file(schema_file).to_bigquery_schema()
+
+        for field in schema:
+            # Validate that the query columns have descriptions.
+            if not field.description:
+                click.echo(
+                    f"Column description validation failed, {field.name} does not have "
+                    f"a description in the schema."
```
BenWu, line 228: I think the table name isn't printed until the end if multiple queries are being validated so the error message should the name as well
```suggestion
                    f"Column description validation failed for {query_dir}, "
                    f"{field.name} does not have a description in the schema."
```

bigquery_etl/metadata/parse_metadata.py
```diff
@@ -184,6 +184,7 @@ class Metadata:
     deprecated: bool = attr.ib(False)
     deletion_date: Optional[date] = attr.ib(None)
     monitoring: Optional[MonitoringMetadata] = attr.ib(None)
+    enforce_col_desc: bool = attr.ib(False)
```
sean-rose, line 187: IMO this setting would make more sense nested under `schema`.  Note that if you do that, you'll need to mark `SchemaMetadata`'s existing `derived_from` property as optional to allow specifying just the new property under `schema`.

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -204,6 +204,35 @@ def validate_shredder_mitigation(query_dir, metadata):
     return True
 
 
+def validate_col_desc_enforced(query_dir, metadata):
+    """Check schemas with enforce_col_desc = True comply with requirements."""
+    # If column descriptions need to be enforced:
+    if metadata.enforce_col_desc is True:
```
sean-rose, line 210: If you reverse this conditional you can return early when the new property isn't set and reduce the indentation of the rest of the function.

Also, I think it's more idiomatic Python to simple rely on the "truthiness" of values rather than using explicit comparisons like `is True` or `is not True`.

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -204,6 +204,35 @@ def validate_shredder_mitigation(query_dir, metadata):
...
+            click.echo(
+                click.style(
+                    f"Table {query_dir} does not have schema.yaml required for column descriptions.",
+                    fg="yellow",
+                )
+            )
+            return False
+        schema = Schema.from_schema_file(schema_file).to_bigquery_schema()
+
+        for field in schema:
```
sean-rose, line 223: This is currently only checking top-level columns.  I'd recommend checking all nested fields as well.

BenWu (commented):
I think the implementation is fine but enforcing every column seems restrictive and could end up with most descriptions just being the column name without underscores to get past validation.  Is the expectation in the future that we have good descriptions for every column?  i.e. is that something that should be pointed out in code review?

This isn't a blocker though and this change is ok IMO

tests/cli/test_cli_metadata.py
```diff
@@ -685,3 +686,119 @@ def test_validate_metadata_without_labels(self, runner, capfd):
             captured = capfd.readouterr()
             assert result is None
             assert captured.out == ""
+
+    @patch("google.cloud.bigquery.Client")
+    @patch("google.cloud.bigquery.Table")
```
sean-rose, line 691: I don't think any of these tests you're adding actually need/use these mocked BigQuery SDK objects, so I'd suggest removing them from these tests.

kwindau:
> I think the implementation is fine but enforcing every column seems restrictive and could end up with most descriptions just being the column name without underscores to get past validation. Is the expectation in the future that we have good descriptions for every column? i.e. is that something that should be pointed out in code review?
> 
> This isn't a blocker though and this change is ok IMO

I have the same worries  - I think that's more of a we need to get stricter on code reviews though

bigquery_etl/metadata/parse_metadata.py
sean-rose: IMO `require_column_descriptions` would be slightly more to the point.

tests/cli/test_cli_metadata.py
sean-rose: You'd need to also remove the corresponding test function arguments, and one line within the functions that uses those arguments.  But none of the code being exercised appears to need to connect to BigQuery, and no asserts are being done with those mocks, so I'm pretty sure they can be safely removed.

kwindau: I see now - you're right - fixed that! Thanks Sean!

bigquery_etl/metadata/validate_metadata.py
kwindau: Fixed and added tests, thanks for catching! I didn't even notice

bigquery_etl/metadata/parse_metadata.py
kwindau: Decided to leave as is for now - I debated heavily where to stick this logic and just landed that it was easier to put in metadata for now - we can always move it around later if needed

Marlene-M-Hirose (approved):

bigquery_etl/metadata/parse_metadata.py
sean-rose: That's fine.  To be clear, I was suggesting still having it in `metadata.yaml`, just under the (optional) `schema:` section (not suggesting having it in `schema.yaml`).

PR #7072, opened 2025-02-20
feat(): add composite active users to usage reporting generator

kik-kik:
# feat(): add composite active users to usage reporting generator

Will convert from draft PR once we agree on what the filter should be for the union.

blocked by: https://github.com/mozilla/bigquery-etl/pull/7157

### DIFF
--- a/sql_generators/usage_reporting/templates/composite_active_users.view.sql.jinja
+++ b/sql_generators/usage_reporting/templates/composite_active_users.view.sql.jinja
@@ -1,0 +1,71 @@
+CREATE OR REPLACE VIEW
+  `{{ project_id }}.{{ app_name }}.{{ view_name }}`
+AS
+SELECT
+  submission_date,
+  client_id AS usage_profile_id,
+  first_seen_year,
+  channel,
+  app_name,
+  app_version,
+  app_version_major,
+  app_version_minor,
+  app_version_patch_revision,
+  app_version_is_major_release,
+  country,
+  os,
+  os_version,
+  os_version_major,
+  os_version_minor,
+  {% if app_name == "firefox_desktop" %}
+  os_version_build,
+  {% endif %}
+  distribution_id,
+  is_default_browser,
+  activity_segment,
+  is_dau,
+  is_wau,
+  is_mau,
+  is_daily_user,
+  is_weekly_user,
+  is_monthly_user,
+FROM
+  {% if app_name == "firefox_desktop" %}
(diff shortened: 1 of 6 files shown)


sql_generators/usage_reporting/templates/usage_reporting_active_users.view.sql.jinja
```diff
@@ -3,7 +3,10 @@ CREATE OR REPLACE VIEW
   `{{ project_id }}.{{ app_name }}.{{ view_name }}`
 AS
 SELECT
-  daily.*,
+  daily.* EXCEPT(app_channel, normalized_country_code),
+  app_channel AS channel,
```
kik-kik, line 7: @ogzillla just FYI, this will affect your active_users_aggregates query.

sql_generators/usage_reporting/templates/composite_active_users.view.sql.jinja
```diff
@@ -0,0 +1,50 @@
...
+  activity_segment,
+  is_dau,
+  is_wau,
+  is_mau,
+  is_daily_user,
+  is_weekly_user,
+  is_monthly_user,
+FROM
+  `{{ project_id }}.{{ app_name }}.usage_reporting_active_users`
+-- TODO: will need to add some sort of filter to reduce double counting.
```
kik-kik, line 50: @irrationalagent what filter will make sense here for unioning the baseline and the usage_reporting data? Sounded like date in combination with version should work?

irrationalagent: Is this something that you think we can change in the future without much trouble? I think that for now, something like this makes sense for the usage ping:

```WHERE mozfun.norm.browser_version_info(app_version).major_version >= 136```

And then correspondingly for the active_users part:

```WHERE mozfun.norm.browser_version_info(app_version).major_version < 136```

In the future I think we might want to bump the version number to either 137 or 138. Reason being, there is a bug on Fenix that is just now getting fixed, and will be uplifted into 136 in a march dot release.

I don't think we need a filter on date - we want to take clients on 136+ for all dates, and likewise all clients that haven't updated yet (135 and below) regardless of when they are submitting the pings. For past dates, there will be very few clients on 136 (nightly and beta) but that's ok.

kik-kik: so since this is just a view there is no issues updating this filter later on assuming we do not derive further tables on top of it.

PR #7074, opened 2025-02-20
[DENG-3656] Support depends on past in managed backfills

BenWu:
## Description

Adds managed backfill support for `depends_on_past=True` tables by copying the previous partition (month or day) to the staging table.  This allows backfilling of `clients_last_seen` type tables.  A check was added to look for backfill end dates in the past to attempt to make users think about data continuity issues but it doesn't cover everything.

Tables with null date_partition_parameter like `clients_first_seen` still aren't supported because there isn't a "previous" partition.  A lot of these can be recreated in a single query (e.g. `is_init`) so I don't know if they need managed backfill support but that can be added later if needed.

Also added a backfill for core_clients_last_seen since it's a small table to test the airflow run with.  The backfill won't be set to completed so it won't modify prod.

* DENG-3656

### DIFF
--- a/bigquery_etl/backfill/parse.py
+++ b/bigquery_etl/backfill/parse.py
@@ -73,6 +73,7 @@ class Backfill:
     custom_query_path: Optional[str] = attr.ib(None)
     shredder_mitigation: Optional[bool] = attr.ib(False)
     override_retention_limit: Optional[bool] = attr.ib(False)
+    override_depends_on_past_end_date: Optional[bool] = attr.ib(False)
     billing_project: Optional[str] = attr.ib(None)
 
     def __str__(self):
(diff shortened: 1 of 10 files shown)


BenWu:
Just realized I forgot to modify the query to query the stage table instead of the prod table

kwindau (approved):

bigquery_etl/backfill/parse.py
```diff
@@ -220,36 +226,20 @@ def entries_from_file(
...
-                "end_date": self.end_date,
-                "excluded_dates": sorted(self.excluded_dates),
-                "reason": self.reason,
-                "watchers": self.watchers,
-                "status": self.status.value,
-                "custom_query_path": self.custom_query_path,
-                "shredder_mitigation": self.shredder_mitigation,
-                "override_retention_limit": self.override_retention_limit,
-                "billing_project": self.billing_project,
+                name: value for name, value in entry_dict.items() if value is not None
```
lucia-vargas-a, line 239: Nice!

lucia-vargas-a (approved):
Thanks Ben!

PR #7077, opened 2025-02-21
feat: add windows_build_number to usage_reporting_clients_daily template

kik-kik:
# feat: add windows_build_number to usage_reporting_clients_daily template

This field is needed to accurately determine os version on Windows.


irrationalagent (approved):
Thanks for adding this. i double checked against the other usage fields in the glean dictionary and i think we have everything now/

PR #7089, opened 2025-02-25
feat: add dt to partnership mapping and default to null

kik-kik:
# feat: add dt to partnership mapping and default to null

default to null as it resents the meaning of the value than other (other may imply another meaning rather than lack of partnership)

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry/active_users_aggregates/view.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry/active_users_aggregates/view.sql
@@ -34,9 +34,9 @@ SELECT
   app_version_is_major_release,
   os_grouped,
   CASE
-    WHEN distribution_id LIKE "%vivo%"
+    WHEN STARTS_WITH(distribution_id, "vivo-")
       THEN "vivo"
-    WHEN distribution_id LIKE "%dt%"
+    WHEN STARTS_WITH(distribution_id, "dt-")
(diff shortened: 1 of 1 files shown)


bochocki (approved):
Approving, but just to be sure: are you confident this won't match undesired strings? "%dt%" is a pretty short string to match. For example, "bandwidth", "bedtime", and "bundt cake" would all match. I ran the following query and didn't come up with any results:

SELECT DISTINCT distribution_id
  FROM `moz-fx-data-shared-prod.telemetry.active_users_aggregates_mobile`
 WHERE LOWER(distribution_id) LIKE '%dt%'
   AND submission_date BETWEEN "2024-01-01" AND "2024-06-30"

kik-kik:
> Approving, but just to be sure: are you confident this won't match undesired strings? "%dt%" is a pretty short string to match. For example, "bandwidth", "bedtime", and "bundt cake" would all match. I ran the following query and didn't come up with any results:
> 
> SELECT DISTINCT distribution_id FROM `moz-fx-data-shared-prod.telemetry.active_users_aggregates_mobile` WHERE LOWER(distribution_id) LIKE '%dt%' AND submission_date BETWEEN "2024-01-01" AND "2024-06-30"

Thanks @bochocki , I don't think in this field this should be an issue. But I do see the point you have and tweaked this a little bit to lower the chances of this.

PR #7091, opened 2025-02-25
feat: move udf calls for getting values based on os fields into views

kik-kik:
# feat: move udf calls for getting values based on os fields into views

Not using the results of `norm.truncate_version` and  `norm.windows_version_info`  into the views to allow more flexibility.

Had a chat with @kwindau earlier and it sounds like `norm.windows_version_info` may not always be returning correct / expected results. Hence moving the call to the view instead to allow more flexibility and easier swapping out of how the version values are derived.


kwindau (approved):

irrationalagent (approved):

BenWu:
@kik-kik this is [failing to deploy](https://workflow.telemetry.mozilla.org/dags/bqetl_artifact_deployment/grid?dag_run_id=manual__2025-02-25T18%3A02%3A09.498258%2B00%3A00&task_id=publish_views&map_index=-1&tab=logs) with 
```
oogle.api_core.exceptions.BadRequest: 400 CREATE VIEW has columns with duplicate name os_version_major at [4:1]; reason: invalidQuery, location: query, message: CREATE VIEW has columns with duplicate name os_version_major at [4:1]
```
You can manually drop the three columns from the table if they're ok to delete

PR #7092, opened 2025-02-25
Deng 7935 create views for fx user research team

ogzillla:
## Description
1. Create views atop desktop and mobile tables for Alchemer survey
2. Grant data viewer access to fx-quant-user-research WG

[JIRA](https://mozilla-hub.atlassian.net/browse/DENG-7935)

### DIFF
--- a/sql/moz-fx-data-shared-prod/fx_quant_user_research_derived/dataset_metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/fx_quant_user_research_derived/dataset_metadata.yaml
@@ -13,8 +13,8 @@ workgroup_access:
 - role: roles/bigquery.dataViewer
   members:
   - workgroup:mozilla-confidential
+  - workgroup:fx-quant-user-research/data-viewers
 - role: roles/bigquery.dataEditor
   members:
   - workgroup:fx-quant-user-research/data-developers
-  - workgroup:fx-quant-user-research/data-viewers
 syndication: {}


kik-kik (approved):
LGTM, only question I have. Is there a use case when a user of this data would want to query both mobile and desktop data at the same time or are two separate views for this enough?

sql/moz-fx-data-shared-prod/fx_quant_user_research/dataset_metadata.yaml
```diff
@@ -4,13 +4,10 @@ description: |-
...
-- role: roles/bigquery.dataViewer
-  members:
-  - workgroup:mozilla-confidential
-default_table_expiration_ms: null
 workgroup_access:
 - role: roles/bigquery.dataViewer
   members:
   - workgroup:mozilla-confidential
+  - workgroup:fx-quant-user-research/data-developers
+  - workgroup:fx-quant-user-research/data-viewers
```
sean-rose, line 12: Since `workgroup:mozilla-confidential` is effectively everyone at Mozilla, also granting access to these other specific working groups doesn't really seem necessary.  Or do you want this data to _only_ be accessible to those more specific workgroups?

kik-kik:
> > LGTM, only question I have. Is there a use case when a user of this data would want to query both mobile and desktop data at the same time or are two separate views for this enough?
> 
> There might come a point where the user queries both. Why do you ask? Thinking of a union?

Yes that could be an option. This union could either be a union of the two views or there could be a single view that does it. It would probably also make sense to add a hard coded string column like `"mobile" AS platform,`  for the part union selecting from the mobile table so that downstream if there is a mobile only use case we could filter on this string and avoid the desktop table being scanned.

sql/moz-fx-data-shared-prod/fx_quant_user_research/dataset_metadata.yaml
sean-rose: I wouldn't go so far as to consider this an anti-pattern, I mainly just wanted to make sure you weren't expecting access to be restricted when everyone at Mozilla will have access as it is.

PR #7094, opened 2025-02-25
Add metadata.yml for tasks_v2 + add missing user facing views

ahal:
## Description

I changed the schema of the tasks table, so a `v2` was necessary. I'll attempt to backfill the data from the old `v1` table later. I also added some missing user facing views for tasks and task_runs.

### DIFF
--- a/sql/moz-fx-data-shared-prod/fxci/task_runs/view.sql
+++ b/sql/moz-fx-data-shared-prod/fxci/task_runs/view.sql
@@ -1,0 +1,7 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.fxci.task_runs`
+AS
+SELECT
+  *
+FROM
+  `moz-fx-data-shared-prod.fxci_derived.task_runs_v1`
--- a/sql/moz-fx-data-shared-prod/fxci/tasks/view.sql
+++ b/sql/moz-fx-data-shared-prod/fxci/tasks/view.sql
@@ -1,0 +1,7 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.fxci.tasks`
+AS
+SELECT
+  *
+FROM
+  `moz-fx-data-shared-prod.fxci_derived.tasks_v2`
--- a/sql/moz-fx-data-shared-prod/fxci/worker_metrics/view.sql
+++ b/sql/moz-fx-data-shared-prod/fxci/worker_metrics/view.sql
@@ -1,0 +1,7 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.fxci.worker_metrics`
+AS
+SELECT
+  *
+FROM
+  `moz-fx-data-shared-prod.fxci_derived.worker_metrics_v1`
--- a/sql/moz-fx-data-shared-prod/fxci_derived/task_runs_v1/schema.yaml
+++ b/sql/moz-fx-data-shared-prod/fxci_derived/task_runs_v1/schema.yaml
@@ -1,0 +1,34 @@
+fields:
+  - name: task_id
+    type: STRING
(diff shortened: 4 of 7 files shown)


scholtzan (approved):

ahal:
The deploy is failing because the airflow job hasn't run yet so the `tasks_v2` table doesn't exist. It should exist soon, I'll try rerunning it tomorrow.

ahal:
Still failing, looks like it's because the Telemetry Airflow DAG is configured to create the table under `fxci` already:
https://github.com/mozilla/telemetry-airflow/blob/main/dags/fxci_pulse_export.py#L32

@scholtzan I think this should be changed to create it in `fxci_derived` instead, and then I create a view.sql like in this patch.. is that correct? Sorry, it's taking me awhile but I think I'm starting to get the hang of things ><

scholtzan:
I think adding a `schema.yaml` for `fxci_derived.tasks_v2` would resolve the issue. The stage deploy would then be able to create the table and views

ahal:
I think that would unblock this specific PR, but the deeper issue is that the DAG that populates the table is pointing at the wrong place, see:
https://github.com/mozilla/telemetry-airflow/pull/2171/files

So I think we should land that first, wait for the table to get created, and then land this.

Then we can either backfill the data from the old table into the new one, or honestly what I'm leaning towards.. just start over and not bother.

Aside: due to a bug I fixed late last year, there's only two months of usable data and no one is relying on this ETL quite yet. I also recently changed the `tasks` table schema, so preserving everything is more involved and probably not worth the time it would take.

ahal:
Hmm, I thought this would work but I'm still getting these errors:
```
google.api_core.exceptions.NotFound: 404 Not found: Table *****************************:fxci_derived_moz_fx_data_shared_prod_089b4dd37770b713a126b5b34f60b67ffac38cbb.task_runs_v1 was not found in location US; reason: notFound, message:
Not found: Table *****************************:fxci_derived_moz_fx_data_shared_prod_089b4dd37770b713a126b5b34f60b67ffac38cbb.task_runs_v1 was not found in location US
```

I'm not sure what's up here, I can see the tables existing in the BigQuery console. Is the "location US" bit part of the problem. Do they live in a different location?

ahal:
> I think adding a schema.yaml for fxci_derived.tasks_v2 would resolve the issue. The stage deploy would then be able to create the table and views.

Ah, I guess these are still needed for the views after all. I'll try re-adding them.

ahal:
> > Hmm, I thought this would work but I'm still getting these errors:
> > ```
> > google.api_core.exceptions.NotFound: 404 Not found: Table *****************************:fxci_derived_moz_fx_data_shared_prod_089b4dd37770b713a126b5b34f60b67ffac38cbb.task_runs_v1 was not found in location US; reason: notFound, message:
> > Not found: Table *****************************:fxci_derived_moz_fx_data_shared_prod_089b4dd37770b713a126b5b34f60b67ffac38cbb.task_runs_v1 was not found in location US
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm not sure what's up here, I can see the tables existing in the BigQuery console. Is the "location US" bit part of the problem. Do they live in a different location?
> 
> This error is from the CI staging deployment process where artifacts get deployed to staging for CI tests. The problem is a dummy copy of the `task_runs_v1` table isn't being deployed to the staging environment because there's no `schema.yaml` file for it, and then the CI is trying to deploy a staging copy of the `fxci.task_runs` view referencing that non-existent staging copy of `task_runs_v1`. The best solution is to add `schema.yaml` files for all tables in `fxci_derived` being referenced in the `fxci` views (`task_runs_v1`, `tasks_v2`, etc.).

Oops, didn't notice your comment before posting mine. Thanks for confirming!

PR #7108, opened 2025-02-27
DENG-7965 - Add custom temporary support for relay_backend.events in Shredder

akkomar:
This adds a temporary workaround for Relay table that has `fxa_id` in event extras.
I initially wanted to go with https://mozilla-hub.atlassian.net/browse/DENG-7964 first to move the field to a regular metric, but this might require updates to glean_parser first.
This isn't great, but I haven't found a simpler way. I tried putting an expression to select any `fxa_id` key in `DeleteTarget.field`, but this resulted in correlated subquery which BQ does not support.

I tested this by generating the query locally with `script/shredder_delete --dry_run --start-date=2024-01-01 --end-date=2025-03-02 --only=relay_backend_stable.events_v1` and running it on copied tables.

### DIFF
--- a/bigquery_etl/shredder/delete.py
+++ b/bigquery_etl/shredder/delete.py
@@ -273,7 +273,8 @@ def _override_query_with_fxa_id_in_extras(
             """
                 EXISTS (
                   WITH user_ids AS (
-                  SELECT user_id
+                  SELECT 
+                    user_id_unhashed AS user_id
                   FROM
                   `moz-fx-data-shared-prod.firefox_accounts.fxa_delete_events`
                   WHERE """
--- a/bigquery_etl/shredder/delete.py
+++ b/bigquery_etl/shredder/delete.py
@@ -273,7 +273,7 @@ def _override_query_with_fxa_id_in_extras(
             """
                 EXISTS (
                   WITH user_ids AS (
-                  SELECT 
+                  SELECT
                     user_id_unhashed AS user_id
                   FROM
                   `moz-fx-data-shared-prod.firefox_accounts.fxa_delete_events`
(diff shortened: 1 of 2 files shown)


bigquery_etl/shredder/delete.py
```diff
@@ -253,6 +253,44 @@ class Partition:
...
+        and target.fields[0] == "events[*].extra.fxa_id"
+        and len(sources) == 1
+        and sources[0].table == "firefox_accounts.fxa_delete_events"
+        and sources[0].field == "user_id"
+    ):
+        field_condition = (
+            """
+                EXISTS (
+                  WITH user_ids AS (
+                  SELECT user_id
```
BenWu, line 276:
```suggestion
                  SELECT 
                    user_id_unhashed AS user_id
```

bigquery_etl/shredder/delete.py
```diff
@@ -361,7 +408,7 @@ def create_job(client) -> bigquery.QueryJob:
                 """
             )
         run_tense = "Would run" if dry_run else "Running"
-        logging.debug(f"{run_tense} query: {query}")
+        logging.info(f"{run_tense} query: {query}")
```
akkomar, line 411:
```suggestion
        logging.debug(f"{run_tense} query: {query}")
```
Revert unintentional change

bigquery_etl/shredder/delete.py
```diff
@@ -253,6 +253,45 @@ class Partition:
...
+        and target.fields[0] == "events[*].extra.fxa_id"
+        and len(sources) == 1
+        and sources[0].table == "firefox_accounts.fxa_delete_events"
+        and sources[0].field == "user_id"
+    ):
+        field_condition = (
+            """
+                EXISTS (
+                  WITH user_ids AS (
+                  SELECT
```
akkomar, line 276:
```suggestion
                  SELECT
```

PR #7120, opened 2025-02-28
Deng 7974 implement support for extracting url variables from alchemer

ogzillla:
## Description
1. Enable the retrieval of `url_variables` from Alchemer's API
2. Add new fields to schema:
   - `url_variables` 
   - `language`
   - `date_started`
3. Add type annotations
4. Update unit tests
5. Refactor some functions

[DENG-7974](https://mozilla-hub.atlassian.net/browse/DENG-7974)
[DENG-6770](https://mozilla-hub.atlassian.net/browse/DENG-6770?focusedCommentId=1014465)

[DENG-7974]: https://mozilla-hub.atlassian.net/browse/DENG-7974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DENG-6770]: https://mozilla-hub.atlassian.net/browse/DENG-6770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bigquery_etl/alchemer/survey.py
+++ b/bigquery_etl/alchemer/survey.py
@@ -28,7 +28,7 @@ def date_plus_one(date_string: str) -> str:
 
 
 def make_date_iso_compliant(date_string: str) -> str:
-    """Convert date to ISO-compliant format. '2018-09-27 10:42:16 EDT' becomes 2018-09-27 10:42:16 -04:00'."""
+    """Convert date to ISO-compliant format. E.g. '2018-09-27 10:42:16 EDT' becomes 2018-09-27 10:42:16 -04:00'."""
     tz_mapping = {"EDT": "-04:00", "EST": "-05:00"}
     tz_match = re.search("E[DS]T", date_string)
     if not tz_match:


bigquery_etl/alchemer/survey.py
```diff
@@ -149,9 +203,14 @@ def insert_to_bq(
 @click.option("--api_token", required=True)
 @click.option("--api_secret", required=True)
 @click.option("--destination_table", required=True)
-def main(date, survey_id, api_token, api_secret, destination_table):
+@click.option("--include_url_variables", required=True, default=False)
```
sean-rose, line 206: I don't think this new option should be required, and I believe you need to indicate it's a boolean (either with `is_flag=True` or `type=bool`).
```suggestion
@click.option("--include_url_variables", is_flag=True, default=False)
```
If you do keep it required then you'll also need to update all the other ETLs that use this code.

bigquery_etl/alchemer/survey.py
```diff
@@ -52,22 +71,52 @@ def format_responses(s, date):
...
+        survey_data.append(data)
 
+    return survey_data
 
-def construct_data(survey, date):
-    """Construct response data."""
-    return [format_responses(resp, date) for resp in survey["data"]]
 
+def format_responses(
+    survey_response: dict, date: str, include_url_variables: bool
```
sean-rose, line 81: I'd suggest having `include_url_variables` default to false in all these functions:
```suggestion
    survey_response: dict, date: str, include_url_variables: bool = False
```
That will also have the benefit of not needing to pass that argument to all the test calls of these functions except when you want to actually test that optional functionality by passing `include_url_variables=True`.

bigquery_etl/alchemer/survey.py
```diff
@@ -52,22 +71,52 @@ def format_responses(s, date):
...
+        "submission_date": date,  # this is used as the partitioning field
+        "survey_data": survey_data,
+        "url_variables": url_variables,
+        **{field: survey_response[field] for field in fields},
+    }
+
+    return response
+
+
+def construct_data(survey: dict, date: str, include_url_variables: bool) -> list[dict]:
```
sean-rose, line 106:
```suggestion
def construct_data(survey: dict, date: str, include_url_variables: bool = False) -> list[dict]:
```

bigquery_etl/alchemer/survey.py
```diff
@@ -52,22 +71,52 @@ def format_responses(s, date):
...
+        format_responses(resp, date, include_url_variables) for resp in survey["data"]
+    ]
+
+
+def get_survey_data(
+    survey_id: str,
+    date_string: str,
+    token: str,
+    secret: str,
+    include_url_variables: bool,
```
sean-rose, line 118:
```suggestion
    include_url_variables: bool = False,
```

sql/moz-fx-data-shared-prod/fx_quant_user_research_derived/fxqur_viewpoint_desktop_v1/metadata.yaml
```diff
@@ -22,6 +22,8 @@ scheduling:
       "{{ var.value.surveygizmo_api_secret }}",
       "--destination_table",
       "moz-fx-data-shared-prod.fx_quant_user_research_derived.fxqur_viewpoint_desktop_v1",
+      "--include_url_variables",
+      True
```
sean-rose, line 26: The `True` here is problematic since `arguments` is expected to be a list of strings (plus in case-sensitive YAML booleans are expected to be lowercase).  However, if you take my suggestion from another comment to specify `is_flag=True` for this option then I don't think you'd need to specify the value at all.
```suggestion
      "--include_url_variables"
```
(same goes in the other `metadata.yaml` file)

kwindau (approved):

bigquery_etl/alchemer/response.schema.json
```diff
@@ -78,6 +78,23 @@
...
+      },
+      {
+        "type": "STRING",
+        "name": "userid",
+        "mode": "NULLABLE"
+      }
+    ],
+    "type": "RECORD",
+    "name": "url_variables",
+    "mode": "REPEATED"
```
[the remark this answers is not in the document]
sean-rose: While Alchemer sends both `survey_data` and `url_variables` as dictionaries (at least when they have values), their format is pretty different:  for `survey_data` the dictionary values are the actual values, while for `url_variables` the dictionary values are another nested dictionary with the keys `key` and `value`.

That aside, to me it's less about the format of the data from the API and more about what that data represents, plus the impact of the proposed change.  `survey_data` is predefined fields from Alchemer, which should be consistent across all surveys.  `url_variables` is arbitrary URL parameters passed to surveys, which could vary widely depending on the survey.  And having individual URL parameters explicitly added to the common schema means those get added to all tables this code touches, even if those other surveys don't use those URL parameters, which could be confusing.

Another possible way around this is to make the URL variables it's looking for configurable and then dynamically configure the table schema for that survey accordingly.  For example, instead of `--include_url_variables` you could have `--url_variables` taking a list of URL variable names, and only if that's specified add a (non-repeated) `url_variables` record column with subfields for the specified URL variables.

bigquery_etl/alchemer/survey.py
```diff
@@ -12,35 +12,64 @@
...
-    """Return a nested field of responses for each user."""
-    # `survey_data` is a dict with question ID as the key and question/response
-    # details as the value e.g. survey_data: {'25': {'id': 25, 'type': 'RADIO',
-    # 'question': 'I trust Firefox to help me with my online privacy',
-    # 'section_id': 2, 'answer': 'Agree', 'answer_id': 10066, 'shown': True}}
-    # See https://apihelp.alchemer.com/help/surveyresponse-returned-fields-v5#getobject
+def make_date_iso_compliant(date_string: str) -> str:
+    """Convert date to ISO-compliant format. E.g. '2018-09-27 10:42:16 EDT' becomes 2018-09-27 10:42:16 -04:00'."""
+    tz_mapping = {"EDT": "-04:00", "EST": "-05:00"}
+    tz_match = re.search("E[DS]T", date_string)
```
sean-rose, line 33: BigQuery is going to consider the timestamp strings generated by this code to be invalid due to the space between the time part and the offset.
```suggestion
    """Convert date to ISO-compliant format. E.g. '2018-09-27 10:42:16 EDT' becomes 2018-09-27 10:42:16 -04:00'."""
    tz_mapping = {" EDT": "-04:00", " EST": "-05:00"}
    tz_match = re.search(" E[DS]T$", date_string)
```

sean-rose: Based on manually testing a load job it looks like you're right!

That's surprising because [some relevant BigQuery docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#utc_offset) say
> You can specify the offset from UTC using the following format: `{+|-}H[H][:M[M]]`
> When using this format, no space is allowed between the time zone and the rest of the timestamp.

And running a query like `SELECT TIMESTAMP('2018-09-27 10:42:16 -04:00')` fails with an `Invalid timestamp` error.

I guess BigQuery is currently less strict parsing timestamps in load jobs than it is in normal query jobs, though that might be considered a bug in some sense, so I'd still suggest adjusting this code to produce timestamp values that match what BigQuery's docs say is the expected timestamp format.

PR #7125, opened 2025-03-03
[fxa] add new recovery phones table

clouserw:
## Description

Adds the recovery phones table to the database counts monitoring script.


sql/moz-fx-data-shared-prod/accounts_backend_derived/monitoring_db_counts_v1/query.sql
```diff
@@ -291,6 +300,32 @@ WITH table_counts AS (
       WHERE
         providerId = 2 -- see LinkedAccountProviderIds at https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/lib/types.ts
     )
+  UNION ALL
```
clouserw, line 303: @sean-rose this is kind of a proof of concept.  We're only in US and Canada right now but you can see this is going to get unwieldy.  Is there a better way to get a per-country count easily?  Kind of feels like we're growing past the simple counts schema I started with. :-/

clouserw:
@sean-rose thanks for your comments.  I removed all the per-country stuff here so we can at least get the top level count.  Thanks.

sean-rose (approved):

PR #7135, opened 2025-03-04
feat(DENG-7983): Create fx_accounts_linked_clients_staging_v1

kwindau:
## Description

This PR creates the new table: 
- moz-fx-data-shared-prod.telemetry_derived.fx_accounts_linked_clients_staging_v1

It does not create a corresponding view because this is a table that will be used in creating a different, final table that will then be exposed via a view.  We don't want to expose this since it will have some redundant data that will be provided in a cleaned up, aggregate format via the final table.

Retention is set to a shorter period on this table because the final table will have no retention set, and will have only 1 row per unique linked client ID / client ID combination.

* [DENG-7982](https://mozilla-hub.atlassian.net/browse/DENG-7982)

[DENG-7982]: https://mozilla-hub.atlassian.net/browse/DENG-7982?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/fx_accounts_linked_clients_staging_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/fx_accounts_linked_clients_staging_v1/query.sql
@@ -1,0 +1,21 @@
+WITH client_id_to_account_stg AS (
+  SELECT DISTINCT
+    client_info.client_id AS client_id,
+    metrics.string.client_association_uid AS account_id
+  FROM
+    `moz-fx-data-shared-prod.firefox_desktop.fx_accounts`
+  WHERE
+    client_info.client_id IS NOT NULL
+    AND metrics.string.client_association_uid IS NOT NULL
+    AND DATE(submission_timestamp) >= DATE_SUB(@submission_date, INTERVAL 30 day)
+)
+SELECT DISTINCT
+  a.client_id,
+  b.client_id AS linked_client_id,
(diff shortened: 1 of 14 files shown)


sql/moz-fx-data-shared-prod/telemetry_derived/fx_accounts_linked_clients_staging_v1/query.sql
```diff
@@ -0,0 +1,21 @@
...
+SELECT DISTINCT
+  a.client_id,
+  b.client_id AS linked_client_id,
+  @submission_date AS submission_date
+FROM
+  client_id_to_account_stg a
+JOIN
+  client_id_to_account_stg b
+  ON a.account_id = b.account_id
+  AND a.client_id < b.client_id
```
akkomar, line 21: Why `<`?

akkomar (approved):
LGTM because it breaks the link between client_id and account_id.

sql/moz-fx-data-shared-prod/telemetry_derived/fx_accounts_linked_clients_staging_v1/query.sql
kwindau: for example if client ID 123 is linked to 456, this ensures you only keep the row 123, 456, and not the row, 456, 123

PR #7144, opened 2025-03-06
feat(DENG-7989): disable Bigeye monitoring for all mobile products except firefox_ios and fenix

kik-kik:
# feat(DENG-7989): disable Bigeye monitoring for all mobile products except firefox_ios and fenix

### DIFF
--- a/sql_generators/mobile_kpi_support_metrics/__init__.py
+++ b/sql_generators/mobile_kpi_support_metrics/__init__.py
@@ -206,6 +206,7 @@ class Product:
     friendly_name: str
     search_join_key: str
     is_mobile_kpi: bool = False
+    enable_monitoring: bool = False
     attribution_groups: list[AttributionFieldGroup] = field(
         default_factory=list[AttributionFields.empty]  # type: ignore[valid-type]
     )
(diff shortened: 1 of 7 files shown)


bochocki (approved):
This seems like nice functionality to have. I think it's okay to turn off monitoring for these products for now, but we’ll need to turn monitoring back on once we have the new usage ping integrated into the reporting pipeline, since Focus products still contribute to the KPI.

kik-kik:
> This seems like nice functionality to have. I think it's okay to turn off monitoring for these products for now, but we’ll need to turn monitoring back on once we have the new usage ping integrated into the reporting pipeline, since Focus products still contribute to the KPI.

@bochocki I just want to make sure I understand what you're saying correctly. You mean we should have the monitoring enabled on the data coming from the usage_reporting ping? Generally speaking with the telemetry being disabled for these products and no new data will arrive so I'm not sure it makes sense to continue having those monitors?

bochocki:
> @bochocki I just want to make sure I understand what you're saying correctly. You mean we should have the monitoring enabled on the data coming from the usage_reporting ping? Generally speaking with the telemetry being disabled for these products and no new data will arrive so I'm not sure it makes sense to continue having those monitors?

Yes, sorry, I misunderstood where in the pipeline this monitoring was being used. If no new data is coming in, we don't need to monitor. But we will want to monitor new data coming in from the usage_reporting ping and any aggregate tables that ping gets incorporated into, like an updated `active_users_aggregates`.

PR #7146, opened 2025-03-06
feat(DENG-7984): add focus_android and focus_ios to usage_reporting generator configuration

kik-kik:
# feat(DENG-7984): add focus_android and focus_ios to usage_reporting generator configuration

Adding focus_android and focus_ios to the usage_reporting generator.

### DIFF
--- a/bqetl_project.yaml
+++ b/bqetl_project.yaml
@@ -412,9 +412,13 @@ generate:
       focus_android:
         channels:
         - release
+        - beta
+        - nightly
       focus_ios:
         channels:
         - release
+        - beta
+        - nightly
 
 
 retention_exclusion_list:
--- a/bqetl_project.yaml
+++ b/bqetl_project.yaml
@@ -417,8 +417,6 @@ generate:
       focus_ios:
         channels:
         - release
-        - beta
-        - nightly
 
 
 retention_exclusion_list:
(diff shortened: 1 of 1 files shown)


bqetl_project.yaml
```diff
@@ -408,6 +408,12 @@ generate:
       firefox_desktop:
         # firefox_desktop is a single app containing multiple channels hence why set to null.
         channels: null
+      focus_android:
+        channels:
+        - release
```
irrationalagent, line 413: I see `org_mozilla_focus_nightly` and `org_mozilla_focus_beta`. I can't imagine that there are any substantial amount of users on these channels, but what would be the additional work needed to include them?

kik-kik: Added those.

kwindau (approved):

bqetl_project.yaml
```diff
@@ -408,6 +408,16 @@ generate:
...
+        - beta
+        - nightly
+      focus_ios:
```
irrationalagent, line 416: i didn't see any prerelease channels for focus ios, but its possible i missed them. do they actually exist?

kik-kik: you are correct. Removed the other channels.

PR #7157, opened 2025-03-11
fix: isp column appears to be out of order between last_seen and daily table for some products

kik-kik:
# fix: isp column appears to be out of order between last_seen and daily table for some products. 

This change is to fix the location of the isp column in the last_seen view. Making this change as currently this results in dry run failure for focus_nightly, focus_beta and klar.

blocks: https://github.com/mozilla/bigquery-etl/pull/7072

### DIFF
--- a/sql_generators/glean_usage/templates/baseline_clients_last_seen_v1.query.sql
+++ b/sql_generators/glean_usage/templates/baseline_clients_last_seen_v1.query.sql
@@ -11,11 +11,12 @@ SELECT
   {% if app_name == "firefox_desktop" %}
   CAST(NULL AS INT64) AS days_desktop_active_bits,
   {% endif %}
+  isp,
   -- We make sure to delay * until the end so that as new columns are added
   -- to the daily table we can add those columns in the same order to the end
   -- of this schema, which may be necessary for the daily join query between
   -- the two tables to validate.
-  *
+  * EXCEPT(isp),
 FROM
   `{{ daily_table }}`
 WHERE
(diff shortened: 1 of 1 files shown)


scholtzan (approved):

sean-rose:
I'd lean toward #7119 over this.

kik-kik:
@sean-rose , I'd be okay with that too. Except long term this means we have more points where we need to remember to make updates as columns get added to the daily table. My question here would be, what would be the risk of doing it this way? At least until we can come up with and implement a better / long term solution for this? In general, I'd argue that this specific table should only contain the client_id, sample_id, and the bit fields and nothing else. But since this is currently blocking my changes related to ToS / ToU which is high priority so my preference would be to just take the action of least friction now so that I can get that work completed.

sean-rose:
That makes sense, and I'm fine with you merging this to unblock downstream work, I'd just personally like to see something like #7119 be the ultimate solution.  Only selectively reordering columns when a schema mismatch with incompatible column types causes an outright query error still leaves the possibility of a potentially worse outcome of this query silently doing the wrong thing if there's a mismatched column order with compatible column types and nobody noticing.

sean-rose:
I belatedly realized that [`sql_generators/glean_usage/templates/baseline_clients_last_seen_v1.schema.yaml`](https://github.com/mozilla/bigquery-etl/blob/main/sql_generators/glean_usage/templates/baseline_clients_last_seen_v1.schema.yaml) should probably have also been updated here to reflect the new column order.

PR #7202, opened 2025-03-20
Add new column to slow_startup_events_by_startup_type_v1

kwindau:
## Description

This PR adds a new column "nbr_distinct_users" to the table: 
- moz-fx-data-shared-prod.google_play_store_derived.slow_startup_events_by_startup_type_v1

* [DENG-7931](https://mozilla-hub.atlassian.net/browse/DENG-7931)

[DENG-7931]: https://mozilla-hub.atlassian.net/browse/DENG-7931?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sql/moz-fx-data-shared-prod/google_play_store_derived/slow_startup_events_by_startup_type_v1/query.py
+++ b/sql/moz-fx-data-shared-prod/google_play_store_derived/slow_startup_events_by_startup_type_v1/query.py
@@ -139,8 +139,7 @@ def main():
 
         # Code only set to handle 1 page, error out if more than 1 so it can be fixed
         if "nextPageToken" in result_json:
-            print("next page found, not parsed")
-            raise KeyError
+            raise NotImplementedError("Parsing for next page is not implemented yet.")
 
         # Initialize as none until we find them for each app
         pct_users_w_slow_start_during_cold_start = None
--- a/sql/moz-fx-data-shared-prod/google_play_store_derived/slow_startup_events_by_startup_type_v1/query.py
+++ b/sql/moz-fx-data-shared-prod/google_play_store_derived/slow_startup_events_by_startup_type_v1/query.py
@@ -55,7 +55,7 @@ def create_request_payload_using_logical_dag_date(date_to_pull_data_for):
                 "day": day_after_date_to_pull_data_for_day,
             },
         },
-        "metrics": ["slowStartRate"],
+        "metrics": ["slowStartRate", "distinctUsers"],
         "dimensions": ["startType"],
         "pageSize": 100,
     }
@@ -119,6 +119,7 @@ def main():
             "pct_users_with_slow_start_event_during_cold_app_start": [],
             "pct_users_with_slow_start_event_during_warm_app_start": [],
             "pct_users_with_slow_start_event_during_hot_app_start": [],
+            "nbr_distinct_users": [],
         }
     )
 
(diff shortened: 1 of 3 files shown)


sql/moz-fx-data-shared-prod/google_play_store_derived/slow_startup_events_by_startup_type_v1/query.py
```diff
@@ -136,27 +137,37 @@ def main():
         # Get the data from the result
         result_json = api_call_result.json()
 
+        # Code only set to handle 1 page, error out if more than 1 so it can be fixed
+        if "nextPageToken" in result_json:
+            print("next page found, not parsed")
+            raise KeyError
```
kik-kik, line 143: If we want this to error I'd suggest replacing this with a more suitable exception and I'd just put the message from the print above as the message of the exception. Main thing I'm worried about here, if this error happens how easy can I understand what the error is when looking at the run logs. For example, `NotImplementedError` could be more representative of what is going on here.

The comment above the if I'd maybe expand and say what needs to be done if we start seeing this error.

kwindau: Sure good call, will make the error message more clear

sql/moz-fx-data-shared-prod/google_play_store_derived/slow_startup_events_by_startup_type_v1/query.py
```diff
@@ -136,27 +137,37 @@ def main():
...
-                    "decimalValue"
-                ]["value"]
+            for metric in row["metrics"]:
+                if metric["metric"] == "slowStartRate":
+                    if startup_type == "COLD":
+                        pct_users_w_slow_start_during_cold_start = metric[
+                            "decimalValue"
+                        ]["value"]
+
+                    if startup_type == "WARM":
```
kik-kik, line 160: Are these optional fields? If they should always exist just have no value we could skip the conditions here. If each record should have all you could also consider using `.get(KEY)` to avoid the additional if logic.

sql/moz-fx-data-shared-prod/google_play_store_derived/slow_startup_events_by_startup_type_v1/schema.yaml
```diff
@@ -19,3 +19,7 @@ fields:
   type: NUMERIC
   mode: NULLABLE
   description: The percent of unique users with a slow start event during a hot app start
+- name: nbr_distinct_users
+  type: NUMERIC
```
kik-kik, line 23: Same as above comment.

kik-kik, line 23:
```suggestion
  type: INTEGER
```

kwindau: I originally had it as numeric because I wasn't sure at first what format the number would be but they do seem to always be integer, so updated it.

sql/moz-fx-data-shared-prod/google_play_store_derived/slow_startup_events_by_startup_type_v1/query.py
kwindau: No these aren't optional fields in this case

kwindau: The problem is each record doesn't have all (the way they structure this JSON is weird, that's why I'm doing kinda strange logic here).

PR #7253, opened 2025-03-28
[DENG-8051] Limit concurrency of glam sql generation

BenWu:
## Description

The glam `daily_*` tasks use a lot of memory and occasionally get OOM'd.  I found that it's because the sql generation runs a two python processes per stable table of the target app so fenix spawns 64 processes and each uses a bit of memory.  This changes it to run for 2 tables at a time so max 4 processes.

Memory usage running in a container:
image

First bump is with 2 table concurrency (570MB), second is 3 (900MB), and last two are unlimited (3GB and 6.5GB).

* DENG-8051
* https://bugzilla.mozilla.org/show_bug.cgi?id=1957121


edugfilho (approved):
Thank you for investigating and fixing this!

PR #7254, opened 2025-03-28
[DENG-7884] Add bot_generated flag to generated stable views

scholtzan:
## Description

This adds a flag `bot_generated` to the generated stable ping views to identify records that are coming from bots. This is based on the proposal in https://docs.google.com/document/d/1N8GmbNF2GUspzOwxnDEowWGOwmMYNkw_3IDOQ6MxGi8/edit?tab=t.0#heading=h.14mpx6a8znb7

No data needs to be backfilled at this time as this only changes the view definition. We might propagate the flag into some downstream datasets as part of some follow-up PRs


sql_generators/stable_views/__init__.py
```diff
@@ -23,6 +23,8 @@
 from bigquery_etl.dryrun import get_id_token
 from bigquery_etl.config import ConfigLoader
 
+BOT_GENERATED = 'LOWER(IFNULL(metadata.isp.name, '')) <> "browserstack" AS bot_generated,'
```
ncalexan, line 26: Isn't `bot_generated = true` when the ISP name is *not* "browserstack"?  Is that not backward?

sql_generators/stable_views/__init__.py
```diff
@@ -23,6 +23,8 @@
 from bigquery_etl.dryrun import get_id_token
 from bigquery_etl.config import ConfigLoader
 
+BOT_GENERATED = 'LOWER(IFNULL(metadata.isp.name, "")) = "browserstack" AS bot_generated,'
```
kik-kik, line 26: I wonder if we should move the `,` to be in the sql statement instead of the string here, if someone adds a new field later to the query they may miss the fact it's in the string already.
```suggestion
BOT_GENERATED = 'LOWER(IFNULL(metadata.isp.name, "")) = "browserstack" AS bot_generated'
```

sql_generators/stable_views/__init__.py
```diff
@@ -33,7 +35,8 @@
     {replacements}),
   mozfun.norm.extract_version(client_info.app_display_version, 'major') as app_version_major,
   mozfun.norm.extract_version(client_info.app_display_version, 'minor') as app_version_minor,
-  mozfun.norm.extract_version(client_info.app_display_version, 'patch') as app_version_patch
+  mozfun.norm.extract_version(client_info.app_display_version, 'patch') as app_version_patch,
+  {bot_generated}
```
kik-kik, line 39:
```suggestion
  {bot_generated},
```

sql_generators/stable_views/__init__.py
```diff
@@ -46,6 +49,7 @@
 SELECT
   * REPLACE(
     {replacements}),
+  {bot_generated}
```
kik-kik, line 52:
```suggestion
  {bot_generated},
```

sql_generators/stable_views/__init__.py
scholtzan: Yeah, I wasn't sure about the naming. `is_` is probably more indicative of a boolean flag

sql_generators/stable_views/__init__.py
```diff
@@ -23,6 +23,8 @@
 from bigquery_etl.dryrun import get_id_token
 from bigquery_etl.config import ConfigLoader
 
+BOT_GENERATED = 'LOWER(IFNULL(metadata.isp.name, "")) = "browserstack" AS bot_generated'
```
scholtzan, line 26:
```suggestion
BOT_GENERATED = 'LOWER(IFNULL(metadata.isp.name, "")) = "browserstack" AS is_bot_generated'
```

PR #7290, opened 2025-04-08
fix: remove the additional custom import logic from usage_reporting and patch correct object in unit tests

kik-kik:
# fix: remove the additional custom import logic from usage_reporting and patch correct object in unit tests

Thanks @ogzillla for your input to simplify the import logic.

To verify these changes locally:

```shell
# 1. update pip
venv/bin/python -m pip install --upgrade pip

# 2. Re-install bigquery-etl package
venv/bin/python -m pip install -e .

# 3. Run unit tests
venv/bin/python -m pytest -W ignore::DeprecationWarning tests/sql_generators/usage_reporting/test_usage_reporting.py
```

Both generation via the ./bqetl generate command and when calling the module directly.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ dependencies = {file = ["requirements.in"]}
 [tool.setuptools_scm]
 
 [tool.setuptools.packages.find]
-include = ["bigquery_etl", "bigquery_etl.*"]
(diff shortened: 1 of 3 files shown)


scholtzan (approved):
Tested it locally and works fine. Thanks!

PR #7294, opened 2025-04-08
Add nimbus_recorded_targeting_context queries for fenix and firefox ios

jeddai:
## Description

This PR adds two queries — a query to obtain the previous day's Nimbus recorded targeting contexts for Fenix, and an equivalent one for Firefox iOS.

* [EXP-4608](https://mozilla-hub.atlassian.net/browse/EXP-4608)

[EXP-4608]: https://mozilla-hub.atlassian.net/browse/EXP-4608?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/nimbus_recorded_targeting_context_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/nimbus_recorded_targeting_context_v1/metadata.yaml
@@ -1,6 +1,6 @@
 friendly_name: Nimbus Recorded Targeting Context (Fenix)
 description: |-
-  A query to obtain the
+  A query to obtain the previous day's recorded Nimbus targeting context values for each client.
 owners:
 - chumphreys@mozilla.com
 labels:
--- a/sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/nimbus_recorded_targeting_context_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/nimbus_recorded_targeting_context_v1/metadata.yaml
@@ -1,0 +1,15 @@
+friendly_name: Nimbus Recorded Targeting Context (Fenix)
+description: |-
+  A query to obtain the
+owners:
+- chumphreys@mozilla.com
+labels:
+  incremental: true
+scheduling:
+  dag_name: bqetl_default
+bigquery:
+  time_partitioning:
+    type: day
+    field: submission_date
+    require_partition_filter: true
+    expiration_days: null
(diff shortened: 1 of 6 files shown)


sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/nimbus_recorded_targeting_context_v1/metadata.yaml
```diff
@@ -0,0 +1,15 @@
+friendly_name: Nimbus Recorded Targeting Context (Fenix)
+description: |-
+  A query to obtain the previous day's recorded Nimbus targeting context values for each client.
+owners:
+- chumphreys@mozilla.com
+labels:
+  incremental: true
+scheduling:
+  dag_name: bqetl_default
```
mikewilli, line 9: It might make sense to include these in the `bqetl_experiments_daily` DAG, which queries for things related to experimentation like enrollment and search events.

Otherwise, we'd probably want to create a new DAG for this.

@scholtzan What do you think?

sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/nimbus_recorded_targeting_context_v1/metadata.yaml
```diff
@@ -0,0 +1,15 @@
...
+bigquery:
+  time_partitioning:
+    type: day
+    field: submission_date
+    require_partition_filter: true
+    expiration_days: null
```
mikewilli, line 15: So I've confirmed that this setting will mean data does not expire, which I don't believe is intended or necessary here. I'll let you decide on a reasonable value.

mikewilli (changes requested):
Couple comments, want to make sure the DAG scheduling is appropriate, and that we have an expiration set.

sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/nimbus_recorded_targeting_context_v1/metadata.yaml
jeddai: I'm going to set it to 400 so we'll have ~1yr of data

sql/moz-fx-data-shared-prod/org_mozilla_fenix_derived/nimbus_recorded_targeting_context_v1/metadata.yaml
jeddai: That makes sense to me

mikewilli (approved):
Looks good to me!

After these artifacts get deployed, we can follow this up by auto-generating Looker explores and backfilling the data.

PR #7310, opened 2025-04-11
[fxa] add twilio health metrics

clouserw:
## Description

Adds fxa twilio health metrics


sql/moz-fx-data-shared-prod/firefox_accounts_derived/health_twilio_v1/schema.yaml
```diff
@@ -0,0 +1,25 @@
+fields:
+  - mode: NULLABLE
+    name: date
+    type: DATE
+  - mode: NULLABLE
+    name: category
+    type: STRING
+  - mode: NULLABLE
+    name: count
+    type: FLOAT
```
sean-rose, line 10: Seems like count would probably always be an integer?
```suggestion
    type: INTEGER
```

sean-rose (approved):
r+wc

sql/moz-fx-data-shared-prod/firefox_accounts_derived/health_twilio_v1/schema.yaml
clouserw: I considered that but I haven't worked with Twilio's API and I wanted to stay flexible in case it wasn't.  Their [docs call it a string](https://www.twilio.com/docs/usage/api/usage-record), so.... :-/

clouserw: Anyway, you're probably right.  I changed it.

PR #7367, opened 2025-04-29
bug 1957082: Publish ECH adoption rate

edugfilho:
## Description

This PR publishes the dataset ECH adoption rate.
See bug: [1957082](https://bugzilla.mozilla.org/show_bug.cgi?id=1957082)

* DENG-8381

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/ech_adoption_rate_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/ech_adoption_rate_v1/query.sql
@@ -1,14 +1,14 @@
 WITH unioned AS (
   /* ---------- http3_ech_outcome ---------- */
   SELECT
-    DATE(submission_timestamp) AS day,
+    DATE(submission_timestamp) AS submission_date,
     'http3_ech_outcome' AS metric,
-    h.key AS key,
+    h.key AS label,
     CASE
       WHEN COUNT(DISTINCT client_info.client_id) >= 5000
         THEN metadata.geo.country
       ELSE 'OTHER'
-    END AS country,
+    END AS country_code,
     CASE
       WHEN COUNT(DISTINCT client_info.client_id) >= 5000
         THEN metadata.isp.name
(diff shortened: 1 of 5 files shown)


sql/moz-fx-data-shared-prod/telemetry_derived/ech_adoption_rate_v1/query.sql
```diff
@@ -0,0 +1,150 @@
+WITH unioned AS (
```
akkomar, line 1: Do we really need multiple queries and `UNION ALL` here? IIRC GLAM does something similar without an union, but I might me wrong.

akkomar (commented):
I think you need `schema.yaml` for artifact deployment to work. IIRC `bqetl query schema update` should get you one.

Since there is an active discussion in https://bugzilla.mozilla.org/show_bug.cgi?id=1957082 and PR is in draft mode I'm making a comment review.

sql/moz-fx-data-shared-prod/telemetry_derived/ech_adoption_rate_v1/query.sql
```diff
@@ -0,0 +1,150 @@
...
+      ELSE 'OTHER'
+    END AS isp_name,
+    COUNT(DISTINCT client_info.client_id) AS client_count
+  FROM
+    `moz-fx-data-shared-prod.firefox_desktop.metrics`,
+    UNNEST(metrics.labeled_custom_distribution.http3_ech_outcome) AS h
+  WHERE
+    DATE(submission_timestamp)
+    BETWEEN DATE_SUB(CURRENT_DATE(), INTERVAL 2 DAY)
+    AND CURRENT_DATE()
```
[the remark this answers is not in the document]
edugfilho: My bad here. I was using this query to take a sample of the past two days and that leaked into the PR

sql/moz-fx-data-shared-prod/telemetry_derived/ech_adoption_rate_v1/query.sql
edugfilho: Each of these queries is basically a short version of the ETL, which is much simpler for one or a few metrics, especially for metrics of different types. The full GLAM ETL here would be impractical, because there are too many steps there that don't make much sense for a small amount of metrics. That said, I think I can aggregate all of them into one statement, which will lose readability with maybe better performance, but this already runs in less than 3 minutes...

PR #7402, opened 2025-05-06
[DENG-8432] Filter out nimbus_events.validation_failed and normandy.validation_failed_nimbus_experiment for firefox_desktop_background_update

scholtzan:
## Description

See https://mozilla-hub.atlassian.net/browse/DENG-8432

We saw an increase in `nimbus_events.validation_failed` and `normandy.validation_failed_nimbus_experiment` events that could no longer be processed for `firefox_desktop_background_update.events_stream`. This issue has since been fixed on the FF client (https://phabricator.services.mozilla.com/D247864), but we'll need to filter out these events in the meantime to unblock `firefox_desktop_background_update.events_stream`

### DIFF
--- a/sql_generators/glean_usage/templates/events_stream_v1.query.sql
+++ b/sql_generators/glean_usage/templates/events_stream_v1.query.sql
@@ -144,15 +144,15 @@ CROSS JOIN
     ARRAY(
       SELECT event
       FROM UNNEST(events) AS event
-      WHERE NOT (
+      WHERE (
         app_version_major = 138 AND app_version_minor IN (0, 1)
         AND (
           (event.category = 'nimbus_events' AND event.name = 'validation_failed')
           OR (event.category = 'normandy' AND event.name = 'validation_failed_nimbus_experiment')
         )
-      )
+      ) IS NOT TRUE
     )
-  )
+  ) AS event
   {% else %}
   UNNEST(events) AS event
   {% endif %}
--- a/sql_generators/glean_usage/templates/events_stream_v1.query.sql
+++ b/sql_generators/glean_usage/templates/events_stream_v1.query.sql
@@ -135,7 +135,25 @@ SELECT
 FROM
   base
 CROSS JOIN
+  -- See https://mozilla-hub.atlassian.net/browse/DENG-8432
+  -- Filtering out nimbus and normandy events that are emitted due to 'invalid-feature'.
+  -- The number of these events is too large to be processed, invalid-feature has also
+  -- been removed in more recent versions.
+  {% if app_name == "firefox_desktop_background_update" %}
+  ARRAY(
+    SELECT event
+    FROM UNNEST(events) AS event
+    WHERE NOT (
+      app_version_major = 138 AND app_version_minor IN (0, 1)
+      AND (
+        (event.category = 'nimbus_events' AND event.name = 'validation_failed')
+        OR (event.category = 'normandy' AND event.name = 'validation_failed_nimbus_experiment')
+      )
+    )
+  )
+  {% else %}
   UNNEST(events) AS event
+  {% endif %}
   {% if app_name == "firefox_desktop" %}
     -- See https://mozilla-hub.atlassian.net/browse/DENG-7513
     WHERE
--- a/sql_generators/glean_usage/templates/events_stream_v1.query.sql
+++ b/sql_generators/glean_usage/templates/events_stream_v1.query.sql
@@ -135,19 +135,21 @@ SELECT
 FROM
   base
 CROSS JOIN
+  {% if app_name == "firefox_desktop_background_update" %}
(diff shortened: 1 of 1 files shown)


sql_generators/glean_usage/templates/events_stream_v1.query.sql
```diff
@@ -135,7 +135,25 @@ SELECT
...
+  ARRAY(
+    SELECT event
+    FROM UNNEST(events) AS event
+    WHERE NOT (
+      app_version_major = 138 AND app_version_minor IN (0, 1)
+      AND (
+        (event.category = 'nimbus_events' AND event.name = 'validation_failed')
+        OR (event.category = 'normandy' AND event.name = 'validation_failed_nimbus_experiment')
+      )
+    )
```
scholtzan, line 152: I tried to be more selectively with the filtering here by testing expressions like:

```sql
AND EXISTS (
  SELECT 1
  FROM UNNEST(event.extra) AS kv
  WHERE kv.key = 'reason' AND kv.value = 'invalid-feature'
)
```

or 

```sql
ARRAY_LENGTH(
    ARRAY(
      SELECT 1
      FROM UNNEST(event.extra) AS kv
      WHERE kv.key = 'reason' AND kv.value = 'invalid-feature'
    )
  ) > 0
```
or 

```sql
mozfun.map.get_key(event.extra, "reason") = "invalid-feature"
```

But all of them result in resources being exceeded... Not sure if there is a more efficient way to only filtering these events.

sql_generators/glean_usage/templates/events_stream_v1.query.sql
```diff
@@ -135,7 +135,25 @@ SELECT
...
+  )
+  {% else %}
   UNNEST(events) AS event
+  {% endif %}
```
sean-rose, line 156: In summary, I'd suggest something like this:
```suggestion
JOIN
  UNNEST(events) AS event
  {% if app_name == "firefox_desktop_background_update" %}
    -- See https://mozilla-hub.atlassian.net/browse/DENG-8432
    -- Filtering out nimbus and normandy events that are emitted due to 'invalid-feature'.
    -- The number of these events is too large to be processed, invalid-feature has also
    -- been removed in more recent versions.
    ON (
      (app_version_major = 138 AND app_version_minor IN (0, 1))
      AND (
        (event.category = 'nimbus_events' AND event.name = 'validation_failed')
        OR (event.category = 'normandy' AND event.name = 'validation_failed_nimbus_experiment')
      )
      AND STRUCT('reason' AS key, 'invalid-feature' AS value) IN UNNEST(event.extra)
    ) IS NOT TRUE
  {% endif %}
```

scholtzan: I just tested this query, and it also exceeds resource limits after an hour of runtime.

sql_generators/glean_usage/templates/events_stream_v1.query.sql
```diff
@@ -135,7 +135,27 @@ SELECT
...
+    ARRAY(
+      SELECT event
+      FROM UNNEST(events) AS event
+      WHERE NOT (
+        app_version_major = 138 AND app_version_minor IN (0, 1)
+        AND (
+          (event.category = 'nimbus_events' AND event.name = 'validation_failed')
+          OR (event.category = 'normandy' AND event.name = 'validation_failed_nimbus_experiment')
+        )
+      )
```
sean-rose, line 153:
```suggestion
      WHERE (
        app_version_major = 138 AND app_version_minor IN (0, 1)
        AND (
          (event.category = 'nimbus_events' AND event.name = 'validation_failed')
          OR (event.category = 'normandy' AND event.name = 'validation_failed_nimbus_experiment')
        )
      ) IS NOT TRUE
```

sean-rose (approved):
r+wc

(though I still think doing the filtering in a join condition would be more straightforward and possibly slightly more efficient)

sql_generators/glean_usage/templates/events_stream_v1.query.sql
```diff
@@ -135,7 +135,27 @@ SELECT
...
+    )
+  )
```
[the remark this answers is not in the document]
sean-rose: I believe it works as currently written, though the altered join is technically producing differently shaped results than the default join, which isn't ideal.  However, if you did this and nothing else then references in the `SELECT` like `event.name` wouldn't work, because it would need to be like `event.event.name` for the altered join (which wouldn't work for the normal join).

That could be avoided by changing `SELECT event` to `SELECT AS STRUCT *`, or by using the alternate approach of changing the cross join to a regular join and doing the special filtering in a conditionally-added join condition like I suggested in another comment.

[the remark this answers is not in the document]
sean-rose: Never mind, Ben was right (as I presume you found when actually testing the query, which is what I should have done).

PR #7409, opened 2025-05-08
feat(DENG-8495): Create new glean dataset

kwindau:
## Description

- For the OKR project, creating a new dataset called "glean" that will house the new active_users_aggregates view that combines Glean desktop and Glean mobile app information.
- Traditionally, telemetry was intended to be used for "legacy" data, so want to create a separate place that can house new Glean based views.

* [DENG-8495](https://mozilla-hub.atlassian.net/browse/DENG-8495)

[DENG-8495]: https://mozilla-hub.atlassian.net/browse/DENG-8495?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sql/moz-fx-data-shared-prod/glean/dataset_metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/glean/dataset_metadata.yaml
@@ -1,0 +1,11 @@
+friendly_name: Glean
+description: |-
+  Contains Glean views for combinations of Glean desktop and mobile apps
+dataset_base_acl: view
+user_facing: true
+labels: {}
+workgroup_access:
+- role: roles/bigquery.dataViewer
+  members:
+  - workgroup:mozilla-confidential
+syndication: {}


scholtzan (approved):

sql/moz-fx-data-shared-prod/glean/dataset_metadata.yaml
```diff
@@ -0,0 +1,11 @@
+friendly_name: Glean
+description: |-
+  Contains Glean views for combinations of Glean desktop and mobile apps
```
sean-rose, line 3: Is this going to be specifically for Firefox, or for any/all Mozilla apps?  Because if it's just for Firefox, IMO `firefox` would be a better name.

kwindau: I think any / all Mozilla apps where we want to look at combined Glean views

whd:
A word of caution: this dataset is going to need to be a special case that doesn't map to https://github.com/mozilla-services/mozilla-pipeline-schemas/tree/generated-schemas/schemas/glean/glean. That might be confusing to DE/DSRE since this is now a special case (glean schemas are used to generate other schemas but are not themselves an ingestion namespace).
https://github.com/mozilla-services/cloudops-infra/pull/6410/files fixes the deploy config to allow this dataset to propagate.

whd:
One obvious way to avoid this issue is to say that "glean" is reserved as a user-facing namespace (i.e. bqetl can't create a dataset named glean), as it is already reserved in MPS (ingestion schemas can't be created for a glean app named glean).

But even knowing that glean is reserved in MPS is something I imagine only a few DE could jump to immediately to begin to debug the deployment issue here.

PR #7413, opened 2025-05-08
fix: Rename dataset glean to glean_telemetry

kwindau:
## Description

This is a follow up to [PR-7409](https://github.com/mozilla/bigquery-etl/pull/7409) which caused issues by accidentally using a reserved namespace.  This renames the dataset to `glean_telemetry` instead of `glean`.


whd (approved):
I'm not strongly of the opinion that we should reserve `glean` from bqetl usage but I'm probably not going to get to applying https://github.com/mozilla-services/cloudops-infra/pull/6410/files until tomorrow so this seems fine to me.

PR #7429, opened 2025-05-13
feat(DENG-2407): relax FxA events sql checks

akkomar:
This will fix intermittent check failures.
Now when `accounts_events` explores are hidden and no key/new dashboards use them it's enough if these checks validate that all new events are implemented in `events` ping. We can ignore cases when new event is not implemented in legacy `accounts_events` ping.


kik-kik (approved):
In general should look into moving those checks over to Bigeye.

PR #7439, opened 2025-05-14
Increase history from 180 days to last 768 days

kwindau:
## Description

The table `moz-fx-data-shared-prod.telemetry_derived.cohort_weekly_statistics_v1` has logic to pull all cohorts starting in last 180 days, but extending this to be a larger time range (now all cohorts starting in last 775 days). Note, using 768 here because the logic gets the start of the week so this ensures the start will be within the 775 day range.


danielkberry (approved):
LGTM, thanks for jumping on this so quickly! Will this automatically be rerun when merged or will a job need to be run manually?

PR #7448, opened 2025-05-15
feat: add bqetl_backfill_validate pre-commit hook

kik-kik:
# feat: add bqetl_backfill_validate pre-commit hook

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -55,6 +55,6 @@ repos:
     name: bqetl_backfill_validate
     language: script
     entry: ./bqetl
-    args: [backfill, validate]
+    args: [backfill, validate, --ignore-missing-metadata]
     types: [yaml]
     files: ^.+.backfill.yaml$
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,3 +49,12 @@ repos:
     entry: ./bqetl
     args: [format, --check]
     types: [sql]
+- repo: local
+  hooks:
+  - id: bqetl_backfill_validate
+    name: bqetl_backfill_validate
+    language: script
+    entry: ./bqetl
+    args: [backfill, validate]
+    types: [yaml]
+    files: ^.+.backfill.yaml$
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -49,3 +49,12 @@ repos:
     entry: ./bqetl
     args: [format, --check]
     types: [sql]
+- repo: local
+  hooks:
+  - id: bqetl_backfill_validate
+    name: bqetl_backfill_validate
+    language: script
+    entry: ./bqetl
+    args: [backfill, validate]
```
BenWu, line 58: I think this needs `--ignore-missing-metadata` otherwise it will fail if you don't have all the generated sql

bigquery_etl/cli/backfill.py
```diff
@@ -233,20 +234,26 @@ def create(
...
-    qualified_table_name,
+    asset_name,
     sql_dir,
     project_id,
     ignore_missing_metadata,
 ):
     """Validate backfill.yaml files."""
-    if qualified_table_name:
+    if asset_name:
+        # checking if path instead of "dataset.table_name" format was passed
```
BenWu, line 251: IMO allowing multiple input types is more confusing and trying to coerce a path into a name might cause unexpected errors.  e.g. if you did `bqetl validate sql/project/dataset/table` without a file then the line below would give `sql.project.dataset`

I think it would be better to try to detect if it's a path and raise an error that says something like `Expected qualified table name (project.dataset.table_name), not a path: {qualified_table_name}`

kik-kik: I think the problem here is that pre-commit would pass in the file path. But perhaps that is wrong and there is a way to modify what the command input would be. Honestly, I haven't really worked with pre-commit hooks that much but I do think it would be useful to have this check and have it run at this stage instead of waiting for the CI to tell me this 20 minutes later.

I guess another option could be to include an optional parameter to the command which accepts path, WDYT?

.pre-commit-config.yaml
kik-kik: ah, that's great. We should add that then. To be honest, I missed the fact this option exists.

bigquery_etl/cli/backfill.py
kik-kik: ok, reverted the argument name change and added additional comment for the context of this change / logic

BenWu (approved):

PR #7455, opened 2025-05-16
fix(): usage reporting first seen duplicates

kik-kik:
# fix(): usage reporting first seen duplicates


sql_generators/usage_reporting/templates/usage_reporting_clients_last_seen_v1.query.sql.jinja
```diff
@@ -51,4 +51,6 @@ FROM
   _current
 FULL JOIN
   _previous
+  -- TODO: should we be using app_channel for join here?
```
kik-kik, line 54: @irrationalagent I just noticed in the last_seen we do not use `app_channel` as a join key which in theory should be fine since as far as I know the profile_id (just like client_id) should be unique across all channels. If this indeed is the case, we could just remove it from the join condition in the first_seen query.

irrationalagent: From my understanding, its not possible for a profile to switch channels on mobile. On desktop, I think it IS possible technically, if you have a profile on your machine and you download a different channel you will be able to select that profile on the other channel. I think its a rare edge case but perhaps we should allow for it on desktop only.

kik-kik: alternatively, prior to the join we could set the null values to an empty string. This would also help us resolve this issue.

kik-kik:
This change is no longer relevant, the issue will be resolved by the following PR: https://github.com/mozilla/bigquery-etl/pull/7504

PR #7460, opened 2025-05-16
Use deep copy when merging schemas

BenWu:
## Description

I found this bug while looking at the desktop_crashes generator: https://github.com/mozilla/bigquery-etl/blob/730e33d3d43854f6bee38d38aacf0412927982d9/sql_generators/desktop_crashes/__init__.py#L44-L51

Multiple source schemas are merged into a target schema but since it uses a shallow copy, the  dict references in the target are the same as the ones in the sources so the source schemas get updated unexpectedly.

Deep copying the desktop metrics dict takes ~6.5 ms so performance shouldn't performance shouldn't matter.

* https://bugzilla.mozilla.org/show_bug.cgi?id=1920544

### DIFF
--- a/bigquery_etl/schema/__init__.py
+++ b/bigquery_etl/schema/__init__.py
@@ -7,6 +7,7 @@ from tempfile import NamedTemporaryFile
 from typing import Any, Dict, Iterable, List, Optional
 
 import attr
+import ujson
 import yaml
 from google.api_core.exceptions import NotFound
 from google.cloud import bigquery
@@ -265,7 +266,10 @@ class Schema:
             else:
                 if update and add_missing_fields:
                     # node does not exist in schema, add to schema
-                    columns.append(node.copy())
+                    if node["type"] == "RECORD":  # deep copy record fields
+                        columns.append(ujson.loads(ujson.dumps(node)))
+                    else:
+                        columns.append(node.copy())
                     print(f"Field {node_name} added to {prefix}")
                 else:
                     if not ignore_missing_fields:


bigquery_etl/schema/__init__.py
```diff
@@ -265,7 +266,10 @@ def _traverse(
             else:
                 if update and add_missing_fields:
                     # node does not exist in schema, add to schema
-                    columns.append(node.copy())
+                    if node["type"] == "RECORD":  # deep copy record fields
+                        columns.append(ujson.loads(ujson.dumps(node)))
```
BenWu, line 270: Used `usjon` instead of `copy.deepcopy(node)` because it's faster (15ms vs 6.5ms)

mikewilli: I was interested in whether I should adopt this in other projects so went to look into it, and seems like ujson is in maintenance-only mode (assuming this [link](https://github.com/ultrajson/ultrajson?tab=readme-ov-file#project-status) is the correct project) and recommending people switch to `orjson` (https://pypi.org/project/orjson/).

I guess this is probably not necessary for this change, since it seems like we already have `ujson` in the project deps, but may be worth migrating to `orjson` in the future.

mikewilli (approved):

sean-rose:
This is obviously the correct thing to do, but we should watch out for what this does to memory usage, as we've had issues with CI jobs stalling due to hitting memory limits.

bigquery_etl/schema/__init__.py
BenWu: Oh I didn't know `ujson` was in maintenance-only, I just saw that it was already in the requirements.in.  `orjson` makes sense to use since the interface looks the same

BenWu:
> we should watch out for what this does to memory usage

I agree and I'll do that but I expect this to have negligible impact.  The larger schemas should be in the hundreds of kb and the larger issue is probably python process overhead from multiprocessing

PR #7475, opened 2025-05-21
fix: change the start_date for the attribution_clients managed backfills to be less than 775 days ago

kik-kik:
# fix: change the start_date for the attribution_clients managed backfills to be less than 775 days ago

This is to fix the managed backfill task failure responsible for processing the backfill. Error:
```
[2025-05-20, 11:02:53 UTC] {pod_manager.py:496} INFO - [base] Cannot backfill more than 775 days prior to current date due to retention policies
```


alekhyamoz (approved):
lgtm

PR #7477, opened 2025-05-21
feat: add bqetl_backfill_validate pre-commit hook

kik-kik:
`validate_multiple` command added to support multiple backfill file paths being passed into bqetl for backfill config validation. Example usage:

```shell
/bqetl backfill validate-multiple \
        sql/moz-fx-data-shared-prod/org_mozilla_fenix_nightly_derived/baseline_clients_daily_v1/backfill.yaml \
        sql/moz-fx-data-shared-prod/org_mozilla_firefox_derived/baseline_clients_daily_v1/backfill.yaml
```

This will run validation for all files passed in and exist with status code 1 if any errors are found by the upstream `validate` command.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -55,6 +55,6 @@ repos:
     name: bqetl_backfill_validate
     language: script
     entry: ./bqetl
-    args: [backfill, validate]
+    args: [backfill, validate, --ignore-missing-metadata]
     types: [yaml]
     files: ^.+.backfill.yaml$
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -57,4 +57,4 @@ repos:
     entry: ./bqetl
     args: [backfill, validate-multiple, --ignore-missing-metadata]
     types: [yaml]
-    files: ^.+.backfill.yaml$
+    files: ^.+/backfill\.yaml$
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,6 +49,7 @@ repos:
     entry: ./bqetl
     args: [format, --check]
     types: [sql]
+    exclude: sql_generators/.+\.sql$
 - repo: local
   hooks:
   - id: bqetl_backfill_validate
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -55,6 +55,6 @@ repos:
     name: bqetl_backfill_validate
     language: script
     entry: ./bqetl
-    args: [backfill, validate, --ignore-missing-metadata]
+    args: [backfill, validate-multiple, --ignore-missing-metadata]
     types: [yaml]
     files: ^.+.backfill.yaml$
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,4 +49,12 @@ repos:
     entry: ./bqetl
     args: [format, --check]
     types: [sql]
-    exclude: sql_generators/.+\.sql$
+- repo: local
+  hooks:
+  - id: bqetl_backfill_validate
+    name: bqetl_backfill_validate
+    language: script
+    entry: ./bqetl
+    args: [backfill, validate]
+    types: [yaml]
+    files: ^.+.backfill.yaml$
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -49,4 +49,12 @@ repos:
...
-    exclude: sql_generators/.+\.sql$
+- repo: local
+  hooks:
+  - id: bqetl_backfill_validate
+    name: bqetl_backfill_validate
+    language: script
+    entry: ./bqetl
+    args: [backfill, validate-multiple, --ignore-missing-metadata]
+    types: [yaml]
+    files: ^.+.backfill.yaml$
```
BenWu: `.` before backfill.yaml seems wrong
```suggestion
    files: ^.+/backfill.yaml$
```

bigquery_etl/cli/backfill.py
```diff
@@ -237,13 +237,15 @@ def create(
 @sql_dir_option
 @project_id_option()
 @ignore_missing_metadata_option
+@click.option("--no-exit", default=False, is_flag=True)
```
BenWu: This should have a description since it's not obvious what this is for.  But I think it would be better to raise an exception to catch instead of `sys.exit` so this option wouldn't be needed.

bigquery_etl/cli/backfill.py
```diff
@@ -293,13 +298,67 @@ def validate(
...
+        cmd_args = {
+            "sql_dir": "/".join(sql_dir),
+            "project_id": project_id,
+            "ignore_missing_metadata": True if ignore_missing_metadata else False,
+            "no_exit": True,
+            "qualified_table_name": f"{project_id}.{dataset}.{table_name}",
+        }
+        try:
+            validate_errors = ctx.invoke(validate, **cmd_args)
+            click.echo(validate_errors)
```
BenWu: The error list is already printed in the `validate` cmd

bigquery_etl/cli/backfill.py
```diff
@@ -237,13 +237,15 @@ def create(
...
 def validate(
     ctx,
     qualified_table_name,
     sql_dir,
     project_id,
     ignore_missing_metadata,
+    no_exit,
 ):
     """Validate backfill.yaml files."""
     if qualified_table_name:
```
BenWu: I think it would be simpler to just support `nargs` in the `validate` command since it already supports validating multiple files

Something like
```py
if qualified_table_names:
    backfills_dict = {
        table_name: get_entries_from_qualified_table_name(
            sql_dir, table_name
        ) for table_name in qualified_table_names
    }
```

bigquery_etl/cli/backfill.py
kik-kik: Yes, I was trying to reduce how many changes I'm making to the original `validate` function since I was only looking to wrap around it. In general, I think it could make sense to discuss how cli commands should be structured so we can apply the same patterns across the board.

For now, I will add a description to make it clear what it does.

bigquery_etl/cli/backfill.py
kik-kik: you are correct. Removing this.

bigquery_etl/cli/backfill.py
kik-kik: I believe the main issue here was that pre-commit passes the path rather than the expected format by the validate command (project_id.dataset.table) and we decided we preferred to avoid mixing the expected input in this command.

So we'd still need to have a wrapped around to convert the input into the expected format.

I guess one thing we could do is to do as you suggested and remove all the sys exit calls and replace them with exceptions and have the wrapper command catch them and fail with any errors were raised.

.pre-commit-config.yaml
kik-kik: Would the `/` break this pattern for Windows devices?

bigquery_etl/backfill/validate.py
```diff
@@ -136,3 +136,7 @@ def validate_entries(backfills: List[Backfill], backfill_file: Path) -> None:
         validate_depends_on_past_end_date(backfill_entry, backfill_file)
         validate_old_entry_date(backfill_entry)
     validate_entries_are_sorted(backfills)
+
+
+class BackfillConfigurationError(Exception):
```
kik-kik, line 141: I'm not sure if this is the right place for this.

.pre-commit-config.yaml
BenWu: I'm not sure but we use `/` for other arguments in this file so I assume that works.  Also I just noticed the `.` in the file extension should be escaped

```suggestion
    files: ^.+/backfill\.yaml$
```

bigquery_etl/cli/backfill.py
BenWu: Oh right, I forgot about that

BenWu (approved):

PR #7500, opened 2025-05-23
feat(): Remove status badge from README

kik-kik:
# feat(): Remove status badge from README

Removing status page as now we use multiple workflows and there is a known bug in CircleCI status badge whereby it does not work reliably:

https://circleci.com/docs/status-badges/

"In cases where there are multiple workflows running within a pipeline, the status badge will show the result (red or green) of the latest workflow that has finished running. This is a bug."


badboy (approved):
I have no authority but 🤷

BenWu (approved):
I sometimes use this as a quick way to get to the circleci UI but the check next to the latest commit works as well and is accurate

PR #7504, opened 2025-05-23
feat: remove app_channel from first_seen and last_seen usage_reporting queries

kik-kik:
# feat: remove app_channel from first_seen and last_seen usage_reporting queries


irrationalagent (approved):
OK I think this LGTM - should we do another small backfill to check?

PR #7514, opened 2025-05-26
Bug 1968556 - Check if exists before adding sampled metric to glam queries

BenWu:
## Description

Fix for #7503. Sampled metrics are currently be added to queries for all the pings but the metrics are only in the metrics ping.  This change makes it so that only metrics that exist are added

* https://bugzilla.mozilla.org/show_bug.cgi?id=1968556


edugfilho (approved):
Thanks for fixing this in a way that takes care of other tables, too.

PR #7558, opened 2025-06-02
fix(DENG-8738): NULL date checking in WHERE clause skipping necessary rows

jpetto:
## Description

- this removed part of the WHERE clause was causing the model to incorrectly skip rows that have an expected value of NULL in `reviewed_at`.
- the removed code was intended to fix an out of range error when calling TIMESTAMP_SECONDS in the SELECT clause - but this was remedied by dividing the `reviewed_at` value by 1000 (as it was in milliseconds, unlike `created_at` which is in seconds).

* DENG-8738


csadilek (approved):
🎉

PR #7561, opened 2025-06-03
chore(SubPlat): Add descriptions to some views and columns (DENG-972)

sean-rose:
## Description
And also enable the `require_column_descriptions` setting for these views/ETLs to ensure any new columns added to them in the future will also have descriptions.

* DENG-972: SubPlat consolidated reporting ETL


sql/moz-fx-data-shared-prod/subscription_platform/daily_active_logical_subscriptions/schema.yaml
BenWu, line 1: Doesn't the the `derived_view_schemas` generator already create these with the descriptions from the source table?

sean-rose: I think it only does that when the views select from a single table, and these views now select from multiple tables as of #7471.

PR #7571, opened 2025-06-03
feat(DENG-8338): migrate prospect_item_feed table from snowflake

jpetto:
## Description

migrate prospect_item_feed table from snowflake.

this job is intended to execute after the `prospects_v1` [source table](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/snowflake_migration_derived/prospects_v1/metadata.yaml) is updated.

* DENG-8338

### DIFF
--- a/dags.yaml
+++ b/dags.yaml
@@ -2280,8 +2280,9 @@ bqetl_content_ml_daily:
     Daily extracts for corpus item data to be evaluated for new tab presentation.
   default_args:
     owner: skamath@mozilla.com
-    start_date: "2025-05-05" #TODO: Change to merge date
-    email: ["skamath@mozilla.com", "mlcooper@mozilla.com"]
+    # manually started in Airflow on 2025-05-27
+    start_date: "2025-05-27"
+    email: ["skamath@mozilla.com", "rrando@mozilla.com"]
     retries: 2
     retry_delay: 30m
   tags:
--- a/dags.yaml
+++ b/dags.yaml
@@ -2280,8 +2280,8 @@ bqetl_content_ml_daily:
     Daily extracts for corpus item data to be evaluated for new tab presentation.
   default_args:
     owner: skamath@mozilla.com
-    # manually started in Airflow on 2025-05-27
-    start_date: "2025-05-27"
+    # manually started in Airflow on 2025-05-26
+    start_date: "2025-05-26"
     email: ["skamath@mozilla.com", "rrando@mozilla.com"]
     retries: 2
     retry_delay: 30m
--- a/sql/moz-fx-data-shared-prod/snowflake_migration_derived/prospect_item_feed_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/snowflake_migration_derived/prospect_item_feed_v1/metadata.yaml
@@ -1,0 +1,27 @@
+friendly_name: Prospect Item Feed v1
+description: Model prospect created events for the Fx New Tab from prospects_v1 table
+owners:
+  - skamath@mozilla.com
+  - rrando@mozilla.com
+labels:
+  schedule: daily
+  incremental: true
+  owner1: skamath
+  owner2: rrando
+scheduling:
+  dag_name: bqetl_content_ml_daily
+  # be specific about which tables this depends on so the DAG can execute
+  # jobs in the proper order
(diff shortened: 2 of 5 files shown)


dags.yaml
```diff
@@ -2259,8 +2259,9 @@ bqetl_content_ml_daily:
     Daily extracts for corpus item data to be evaluated for new tab presentation.
   default_args:
     owner: skamath@mozilla.com
-    start_date: "2025-05-05" #TODO: Change to merge date
-    email: ["skamath@mozilla.com", "mlcooper@mozilla.com"]
+    # manually started in Airflow on 2025-05-27
+    start_date: "2025-05-27"
```
sean-rose: This would be slightly more accurate:
```suggestion
    start_date: "2025-05-26"
```
While the `bqetl_content_ml_daily` DAG did first start running on 2025-05-27, in terms of scheduling Airflow is basing that on the start of the "data interval", which for a daily DAG is the preceding day (e.g. the very first DAG run's ID is `scheduled__2025-05-26T04:00:00+00:00`).

sql/moz-fx-data-shared-prod/snowflake_migration_derived/prospect_item_feed_v1/metadata.yaml
```diff
@@ -0,0 +1,27 @@
...
+  schedule: daily
+  incremental: true
+  owner1: skamath
+  owner2: rrando
+scheduling:
+  dag_name: bqetl_content_ml_daily
+  # be specific about which tables this depends on so the DAG can execute
+  # jobs in the proper order
+  referenced_tables:
+    - ["moz-fx-data-shared-prod", "snowflake_migration_derived", "prospects_v1"]
```
sean-rose: It's better to not specify `referenced_tables` unless absolutely necessary, because that completely overrides the tooling's automatic dependency detection logic, which parses the ETL's SQL, finds other ETL tables being referenced (even through intermediate views), and automatically adds appropriate Airflow task dependencies.

So while this technically works, it's a gotcha waiting to happen, where if another ETL table dependency is added in the SQL but whoever's doing that forgets to also update this `referenced_tables` metadata then this ETL won't wait for the associated ETL to complete like it should.

```suggestion
```

sql/moz-fx-data-shared-prod/snowflake_migration_derived/prospect_item_feed_v1/schema.yaml
```diff
@@ -0,0 +1,57 @@
...
+    name: candidate_set_id
+    description: A string identifier for the candidate set to which this run pertains.
+  - mode: NULLABLE
+    type: STRING
+    name: candidate_set_type
+    description: Indicates if candidate set consists of either prospects for curation or candidates for recommendation slates
+  - mode: NULLABLE
+    type: INTEGER
+    name: expires_at
+    description: The date at which this candidate set is no longer "fresh".
```
sean-rose, line 13: Also, I'd normally expect columns that end in `_at` to be timestamps.

If you don't need to keep this as an integer you might consider converting it to a timestamp in the SQL (like is done in `prospects_v1` for `created_at` and `reviewed_at`) and then updating the type here.

However, if you do keep it as an integer then it would be helpful to mention what that integer value means in the column description (e.g. is it seconds, or milliseconds).

dags.yaml
jpetto: ah, good to know! will update.

sql/moz-fx-data-shared-prod/snowflake_migration_derived/prospect_item_feed_v1/metadata.yaml
jpetto: ahhh, i thought i was being helpful by being explicit, but the above makes sense. will remove that directive.

sql/moz-fx-data-shared-prod/snowflake_migration_derived/prospect_item_feed_v1/schema.yaml
jpetto: oh, this is a good catch and a bug. this should be a `TIMESTAMP`. will fix.

sean-rose (approved):

PR #7572, opened 2025-06-04
feat: add final views for enrollment dashboard fully glean

mikewilli:
## Description

Should be the final updates needed to fully reconstruct the experiment enrollments dashboard with only glean. Follow-on to https://github.com/mozilla/bigquery-etl/pull/7557 and https://github.com/mozilla/bigquery-etl/pull/7534

* EXP-4979

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry/experiment_unenrollment_overall_v2/view.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry/experiment_unenrollment_overall_v2/view.sql
@@ -1,0 +1,7 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.experiment_unenrollment_overall_v2`
+AS
+SELECT
+  *
+FROM
+  `moz-fx-data-shared-prod.telemetry_derived.experiment_unenrollment_overall_v2`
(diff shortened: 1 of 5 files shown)


sql/moz-fx-data-shared-prod/telemetry/experiment_unenrollment_overall_v2/view.sql
```diff
@@ -0,0 +1,7 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.telemetry.experiment_unenrollment_overall`
```
scholtzan:
```suggestion
  `moz-fx-data-shared-prod.telemetry.experiment_unenrollment_overall_v2`
```
needs to match file name

scholtzan:
Are we planning to eventually reference the `_v2` tables from the views that don't have a version after validation? Having versioned views is something we only do in rare cases

sql/moz-fx-data-shared-prod/telemetry/experiment_unenrollment_overall_v2/view.sql
mikewilli: Oops good catch!

mikewilli:
>Are we planning to eventually reference the _v2 tables from the views that don't have a version after validation? Having versioned views is something we only do in rare cases

@scholtzan Good question, and yes that's what I was thinking. The problem is we'd like to have duplicates of everything right now so that things can exist side-by-side for comparison/validation. If we'd like to avoid this situation altogether then I can take an alternate route, but I thought this was a reasonable exception to the normal rule. I also considered calling these `_glean` instead of `_v2` but I thought that was worse.

Let me know if you'd like me to do this differently or if you have any suggestions.

scholtzan:
I think it's ok for validation, if we eventually migrate them to a non-versioned view.

mikewilli:
Ok, I'll go ahead with this strategy then.

PR #7582, opened 2025-06-05
chore(SubPlat): Have newer SubPlat consolidated reporting ETLs run hourly (DENG-3736)

sean-rose:
## Description
Fresher subscriptions data is needed for Braze (cc @chelseybeck).

* DENG-3736: Increase frequency for product/subplat data updates


sql/moz-fx-data-shared-prod/mozilla_vpn_derived/users_attribution_v2/metadata.yaml
sean-rose, line 1: I'll need to manually deploy this archive table since the ETL isn't scheduled.

sql/moz-fx-data-shared-prod/mozilla_vpn_derived/funnel_product_page_to_subscribed_v1/metadata.yaml
```diff
@@ -8,6 +8,15 @@ labels:
 scheduling:
   dag_name: bqetl_subplat
   date_partition_parameter: date
+  # Adjust the execution delta for the task dependencies in the bqetl_subplat_hourly DAG
+  # so this depends on the latest bqetl_subplat_hourly DAG run.
+  depends_on:
+    - dag_name: bqetl_subplat_hourly
+      task_id: stripe_external__plan__v1
+      execution_delta: -23h
```
BenWu, line 17: Can this be `-22h`?  As I understand it, -23 would wait for the hourly task with an execution date at 0:30 which would be the one that runs at 1:30 for the next day of data.  So if the hourly task isn't `depends_on_past` and is partitioned, it's possible that the last run for the previous day's data failed when this runs.

I don't know if that's true for any of the tasks here so maybe this is purely academic.  I could also be wrong because time intervals are hard

sean-rose: That's a valid concern, but since in all of these cases the dependent ETLs in question are overwriting the entire table I felt `-23h` was better to ensure the daily ETLs wait for their dependent hourly ETLs that started at the exact same time, so those dependent tables are in a consistent state.  Otherwise there would be a race condition where some hourly-updated dependent tables could have just been updated while others could have been updated an hour ago.

sean-rose: I'm now thinking I should probably clarify this situation a bit more in the associated comments.  How about this?
```python
  # Adjust the execution delta for the task dependencies in the bqetl_subplat_hourly DAG
  # to wait for the concurrent bqetl_subplat_hourly DAG tasks to complete so the
  # dependent tables are in a consistent state.
```

BenWu: Sure that makes sense

sean-rose:
Looking at the `sql.diff` I noticed the generated `ExternalTaskMarker`s have incorrect `execution_date` values.

I've submitted #7613 to fix that.

PR #7588, opened 2025-06-05
fix(DENG-8567): fix remote settings syndicate dataset name

akkomar:


akkomar:
> ## Pull Request Overview
> 
> This PR addresses the remote settings syndicate dataset name issue by introducing a new YAML metadata file that defines dataset details.
> 
>     * Adds the dataset_metadata.yaml file with detailed dataset properties
> 
>     * Updates friendly_name, description, and syndication settings for both stage and prod environments

?!
image

PR #7614, opened 2025-06-10
Include task group in `Task.task_key` if present

sean-rose:
## Description
There could be multiple tasks in a DAG with the same task ID but in different task groups.

I probably should have done this last year when I did #5894.

* https://github.com/mozilla/bigquery-etl/pull/5894


sean-rose:
In the `sql.diff`s you can see this is resulting in a bunch of previously missed `ExternalTaskMarker`s being generated in the `bqetl_glean_usage` and `bqetl_mobile_kpi_metrics` DAGs, since both of those DAGs make heavy use of task groups.

BenWu (approved):

PR #7636, opened 2025-06-12
feat: add bigconfig reference docs

kik-kik:
feat: add bigconfig reference docs

### DIFF
--- a/docs/reference/bigconfig.md
+++ b/docs/reference/bigconfig.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


docs/reference/bigconfig.md
```diff
@@ -0,0 +1,152 @@
+# Bigeye - bigConfig Reference Guide
+
+## Overview
+
+bigConfig is a yaml-based declarative configuration-as-code approach to managing out Bigeye assets and metrics applied to them. This includes specifying which resources (in our case BQ tables) we want to monitor, what metrics / monitors we want to be used on the asset, which collection those should be bundled under, and what should happen if those metrics detect data issues. For example, send a Slack notification.
```
scholtzan:
```suggestion
[bigConfig](https://docs.bigeye.com/docs/bigconfig) is a yaml-based declarative configuration-as-code approach to managing out Bigeye assets and metrics applied to them. This includes specifying which resources (in our case BQ tables) we want to monitor, what metrics / monitors we want to be used on the asset, which collection those should be bundled under, and what should happen if those metrics detect data issues. For example, send a Slack notification.
```

let's link to the docs

scholtzan (approved):

PR #7649, opened 2025-06-13
feat(DENG-8889): increase snowflake migration DAG frequency to hourly

jpetto:
## Description

increases frequency to every hour for DAG that updates models migrated from snowflake. this matches the frequency that was running in DBT.

* DENG-8889

### DIFF
--- a/dags.yaml
+++ b/dags.yaml
@@ -2295,8 +2295,8 @@ bqetl_content_ml_hourly:
     Hourly extracts for corpus item data to be evaluated for new tab presentation.
   default_args:
     owner: skamath@mozilla.com
-    # manually started in Airflow on 2025-05-26
-    start_date: "2025-05-26"
+    # Converted to an hourly DAG on 2025-06-24
+    start_date: "2025-06-24"
     email: ["skamath@mozilla.com", "rrando@mozilla.com"]
     # if the hourly task fails, retry once after a 10 minute wait
     retries: 1
--- a/dags.yaml
+++ b/dags.yaml
@@ -2289,17 +2289,18 @@ bqetl_monitoring_hourly:
     - impact/tier_1
     - repo/bigquery-etl
 
-bqetl_content_ml_daily:
-  schedule_interval: 0 4 * * *
+bqetl_content_ml_hourly:
+  schedule_interval: 30 * * * *
   description: |
-    Daily extracts for corpus item data to be evaluated for new tab presentation.
+    Hourly extracts for corpus item data to be evaluated for new tab presentation.
   default_args:
     owner: skamath@mozilla.com
     # manually started in Airflow on 2025-05-26
     start_date: "2025-05-26"
     email: ["skamath@mozilla.com", "rrando@mozilla.com"]
-    retries: 2
-    retry_delay: 30m
+    # if the hourly task fails, retry once after a 10 minute wait
+    retries: 1
+    retry_delay: 10m
   tags:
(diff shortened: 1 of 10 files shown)


dags.yaml
```diff
@@ -2290,7 +2290,7 @@ bqetl_monitoring_hourly:
     - repo/bigquery-etl
 
 bqetl_content_ml_daily:
```
scholtzan: I think this should be renamed to `bqetl_content_ml_hourly` with the new schedule interval

sean-rose: I agree, and also please change the `schedule: daily` labels on those ETLs to `schedule: hourly` (to be clear that doesn't make a functional difference, but it will avoid misinforming/confusing people).

dags.yaml
```diff
@@ -2290,7 +2290,7 @@ bqetl_monitoring_hourly:
...
-  schedule_interval: 0 4 * * *
+  schedule_interval: 30 * * * *
   description: |
     Daily extracts for corpus item data to be evaluated for new tab presentation.
```
sean-rose:
```suggestion
    Hourly extracts for corpus item data to be evaluated for new tab presentation.
```

dags.yaml
jpetto: ah, that's a good call. i'll need to update the name in the related jobs, too. apologies for being hasty on this one 😅

dags.yaml
```diff
@@ -2290,7 +2290,7 @@ bqetl_monitoring_hourly:
     - repo/bigquery-etl
 
 bqetl_content_ml_daily:
-  schedule_interval: 0 4 * * *
+  schedule_interval: 30 * * * *
```
sean-rose, line 2293: Some additional thoughts:
* Currently this DAG has retries set to 2 and retry delay set to 30 minutes, which means if it's failing the second retry would probably be happening after the subsequent hourly run had already started, which isn't ideal.  I'd suggest reducing one or both of those retry settings to avoid that (also taking into account the time the tasks typically take to complete).
* Speaking of the tasks' runtime, it looks like the recent DAG runs have been completing in a bit under 20 minutes, which is fine, but you'll need to keep an eye on that as things potentially change over time (e.g. ETLs getting added, ETL logic changing, data volume increasing).  It wouldn't be good for these hourly DAG runs to end up taking longer than an hour.
* Currently it looks like this set of ETLs is self-contained with no external dependencies other than Snowplow, but you should be aware that if you do eventually have dependencies on any daily ETLs scheduled via Airflow, then the Airflow task sensors that automatically get generated will likely be incorrect and would have to be manually overridden in the scheduling metadata to avoid issues (the DAG generation logic doesn't currently appropriately handle dependencies between DAGs with fundamentally different frequencies).

dags.yaml
```diff
@@ -2289,17 +2289,18 @@ bqetl_monitoring_hourly:
...
-  schedule_interval: 0 4 * * *
+bqetl_content_ml_hourly:
+  schedule_interval: 30 * * * *
   description: |
-    Daily extracts for corpus item data to be evaluated for new tab presentation.
+    Hourly extracts for corpus item data to be evaluated for new tab presentation.
   default_args:
     owner: skamath@mozilla.com
     # manually started in Airflow on 2025-05-26
     start_date: "2025-05-26"
```
sean-rose:
```suggestion
    # Converted to an hourly DAG on 2025-06-24
    start_date: "2025-06-24"
```

PR #7650, opened 2025-06-14
fix(CI): Update DAG generation so ETL dependencies in the `private-bigquery-etl` repo are detected (DENG-8292)

sean-rose:
## Description
So dependencies from the `private-bigquery-etl` repo don't have to be manually defined via `depends_on` or `external_downstream_tasks` metadata anymore.

This is a counterpart to https://github.com/mozilla/private-bigquery-etl/pull/794.

* DENG-8292: DAG dependencies between bigquery-etl and private-bigquery-etl tables not correctly detected
* https://github.com/mozilla/private-bigquery-etl/pull/794

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -773,7 +773,8 @@ jobs:
             - *restore_venv_cache
             - *build
             - *authenticate
-            - add_ssh_keys:
+            - &add_private_bigquery_etl_ssh_keys
+              add_ssh_keys:
                 # deploy key to private-bigquery-etl
                 fingerprints:
                   - "9d:1e:af:52:78:2c:e8:ec:33:4c:db:cd:5a:ff:70:0a"
@@ -790,7 +791,6 @@ jobs:
                                 git@github.com:mozilla/private-bigquery-etl.git \
                                 ~/private-bigquery-etl
                   rsync --archive ~/private-bigquery-etl/sql/ sql/
-                  rsync --archive ~/private-bigquery-etl/dags.yaml dags.yaml
             - run:
                 name: Generate SQL content
                 no_output_timeout: 30m
@@ -810,7 +810,35 @@ jobs:
                     /tmp/workspace/private-generated-sql/sql/
                   PATH="venv/bin:$PATH" script/bqetl format /tmp/workspace/private-generated-sql/sql/
                   PATH="venv/bin:$PATH" script/bqetl monitoring update /tmp/workspace/private-generated-sql/sql/
-
+            - persist_to_workspace:
+                root: /tmp/workspace
+                paths:
+                  - private-generated-sql
+      - unless:
+          condition: *deploy
+          steps:
+            - *skip
+  private-generate-dags:
+    docker: *docker
+    steps:
+      - when:
+          condition: *deploy
+          steps:
+            - *skip_forked_pr
+            - checkout
+            - *restore_venv_cache
+            - *build
+            - *authenticate
+            - *add_private_bigquery_etl_ssh_keys
+            - run:
+                name: Pull down private-bigquery-etl content
+                command: |
+                  ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
+                  git clone --branch main --depth 1 \
+                    git@github.com:mozilla/private-bigquery-etl.git ~/private-bigquery-etl
+            - run:
+                name: Generate DAGs
+                command: |
                   # Change directory to generate DAGs so `sql_file_path` values are relative to the repo root.
                   export PATH="$PWD/venv/bin:$PATH"
                   export PYTHONPATH="$PWD:$PYTHONPATH"
@@ -834,9 +862,7 @@ jobs:
           condition: *deploy
           steps:
             - *attach_generated_sql
-            - add_ssh_keys:
-                fingerprints:
-                  - "9d:1e:af:52:78:2c:e8:ec:33:4c:db:cd:5a:ff:70:0a"
+            - *add_private_bigquery_etl_ssh_keys
             - run:
                 name: Push to private-generated-sql branch
                 # yamllint disable rule:line-length
@@ -1198,9 +1224,15 @@ workflows:
             branches:
               only:
                 - main
+      - private-generate-dags:
+          filters:
+            branches:
+              only:
+                - main
       - push-private-generated-sql:
           requires:
             - private-generate-sql
+            - private-generate-dags
           filters:
             branches:
               only:
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -791,6 +791,7 @@ jobs:
                                 git@github.com:mozilla/private-bigquery-etl.git \
                                 ~/private-bigquery-etl
                   rsync --archive ~/private-bigquery-etl/sql/ sql/
+                  rsync --archive ~/private-bigquery-etl/sql_generators/ sql_generators/
             - run:
                 name: Generate SQL content
                 no_output_timeout: 30m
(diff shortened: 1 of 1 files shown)


sean-rose:
Note that the latest `sql.diff` showing no differences is inaccurate due to how the `main-generate-sql-and-dags` CI job is also being updated. The previous `sql.diff` shows what the actual changes will be.

.circleci/workflows.yml
```diff
@@ -342,13 +342,29 @@ jobs:
...
+                name: Pull down private-bigquery-etl content
+                command: |
+                  ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
+                  git clone --branch main --depth 1 \
+                    git@github.com:mozilla/private-bigquery-etl.git ~/private-bigquery-etl
             - run:
                 name: Generate DAGs
                 command: |
+                  rm -rf sql/
+                  cp -r /tmp/workspace/private-generated-sql/sql sql
```
BenWu, line 361: Isn't this what's in the removed `- *copy_generated_sql` line?  Could we just add that back?

.circleci/workflows.yml
```diff
@@ -1194,13 +1240,19 @@ workflows:
       # a branch on that internal repository, which may be initially
       # surprising.
       - private-generate-sql:
+          requires:
+            - generate-sql
```
BenWu, line 1200: issue: running these generations sequentially would add another ~25 minutes to the CI run for anything that requires sql generation. IMO it's already too long so I think we need to do this without this dependency 

suggestion: since private-bigquery-etl isn't changed when changes are made here, can we just pull what's in the [private-generated-sql branch](https://github.com/mozilla/private-bigquery-etl/tree/private-generated-sql) and use that to resolve the dag dependencies so that sql generation doesn't need to run twice?

.circleci/workflows.yml
sean-rose: No, `*copy_generated_sql` copies `/tmp/workspace/generated-sql/sql`, while this copies `/tmp/workspace/private-generated-sql/sql`.

.circleci/workflows.yml
sean-rose: > issue: running these generations sequentially would add another ~25 minutes to the CI run for anything that requires sql generation. IMO it's already too long so I think we need to do this without this dependency

I don't believe that's true, as part of these changes is to have the `private-generate-sql` job only run the SQL generators from `private-bigquery-etl` and then combine that with the output of the `generate-sql` job (instead of only (re)running SQL generators from `bigquery-etl`).  In this PR's CI runs the `private-generate-sql` job has been completing in under two minutes (as there currently aren't any SQL generators in `private-bigquery-etl`).

I also triggered a [test CI run for this PR with full SQL generation](https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/49967/workflows/d3a5d22b-f881-4ef0-b1e0-0bf2dd279fe7) and it completed in a little over 45 minutes, which seems like a typical runtime these days for that use case (e.g. all recent `main` branch CI runs have taken longer than that, though they're also doing some extra stuff like publishing container images).

> suggestion: since private-bigquery-etl isn't changed when changes are made here, can we just pull what's in the [private-generated-sql branch](https://github.com/mozilla/private-bigquery-etl/tree/private-generated-sql) and use that to resolve the dag dependencies so that sql generation doesn't need to run twice?

I would have liked to do that if the `private-generated-sql` branch (or some other branch) just contained output from `private-bigquery-etl`.  However, the `private-generated-sql` branch also includes the contents of `sql/` from `bigquery-etl` plus the output of the SQL generators from `bigquery-etl`, which wouldn't be appropriate to use here because a potential change in a `bigquery-etl` PR could be to remove files or no longer generate files which are currently present in the `private-generated-sql` branch.

BenWu: Ok taking another look it makes more sense to me how `private-generate-sql` was changed

.circleci/workflows.yml
```diff
@@ -763,62 +779,86 @@ jobs:
...
+          steps:
+            - *skip_forked_pr
+            - checkout
+            - *restore_venv_cache
+            - *build
+            - *attach_generated_sql
+            - *authenticate
+            - *add_private_bigquery_etl_ssh_keys
+            - run:
+                name: Pull down private-bigquery-etl content
```
BenWu: Why does `private-bigquery-etl` need to be fetched again here if this runs after `private-generate-sql`?

.circleci/workflows.yml
```diff
@@ -763,62 +779,86 @@ jobs:
...
-                  PATH="venv/bin:$PATH" script/bqetl monitoring update /tmp/workspace/private-generated-sql/sql/
+                  $bqetl_script generate all --target-project moz-fx-data-shared-prod
+                  $bqetl_script dependency record --skip-existing sql/
+                  $bqetl_script metadata update sql/
+                  $bqetl_script format sql/
+                  $bqetl_script monitoring update sql/
 
+                  mkdir -p /tmp/workspace/private-generated-sql
+                  cp -r /tmp/workspace/generated-sql/sql /tmp/workspace/private-generated-sql/sql
+                  cp -rf sql/. /tmp/workspace/private-generated-sql/sql
```
BenWu, line 817: should this go before `bqetl dependency record` so that it records the cross-repo depencies or does that not matter?

BenWu (commented):
Generally looks ok to me, just some questions

.circleci/workflows.yml
sean-rose: That doesn't matter, as `bqetl dependency record` is only extracting direct table/view references from the SQL files, not recursively finding table references through referenced views.

.circleci/workflows.yml
sean-rose: Primarily to get `dags.yaml`.  I did consider having the `private-generate-sql` job stash the `dags.yaml` in `/tmp/workspace/private-generated-sql/`, but that didn't feel quite right, and also...

Secondarily to run `bqetl dag generate` in the context of the `private-bigquery-etl` repo, particularly with its own `bqetl_project.yaml` file, in case that project config ever gets used in the DAG generation logic.

BenWu (approved):
Ok LGTM.  There's a lot going on here so thanks for answering my questions.  Can you keep an eye on CI builds on main and diffs in PRs for a bit after merging this to check for edge cases?

.circleci/workflows.yml
scholtzan: This step seems to be almost doing the same as `generate-dags`. Just wondering if we can combine them somehow so that we don't need to run the generation twice. The main difference seems to be where the generated DAGs should be persisted to

scholtzan (approved):

.circleci/workflows.yml
sean-rose: The main difference at the moment is the context/working directory the DAG generation runs in.  As mentioned in my comment above, part of the reason for that is I was speculating that in the future the `bqetl_project.yaml` settings might be used in the DAG generation logic (e.g. adding an option to skip generating certain DAGs).  However, I just realized that if that does end up happening we'd still have an issue with this setup, because the `generate-dags` and `private-generate-dags` each generate all the DAGs with only one repo's `bqetl_project.yaml` being considered (e.g. if there was a skip-DAG-generation option being used in one repo, the job for the other repo wouldn't see that and could generate incorrect Airflow dependencies).

So given that, I think I'll take your suggestion and simplify this to only have one `generate-dags` job that generates and persists all the DAGs with the `bigquery-etl` repo's `bqetl_project.yaml` in effect.  And then if/when `bqetl_project.yaml` settings are ever used in DAG generation logic we can either implement additional logic to combine the multiple `bqetl_project.yaml` files' settings at runtime or just duplicate the appropriate `private-bigquery-etl` DAG settings in the `bigquery-etl` config.

PR #7654, opened 2025-06-16
fix(DENG-8907): improve performance of ETL for prospects_v1 snowflake…

jpetto:
## Description

improve performance of ETL for prospects_v1 snowflake table.

- adds date filtering on base query so as to not rebuild the entire table on every run

* DENG-8907

### DIFF
--- a/sql/moz-fx-data-shared-prod/snowflake_migration_derived/prospects_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/snowflake_migration_derived/prospects_v1/query.sql
@@ -37,7 +37,7 @@ WITH stg_prospects AS (
       AND DATE(derived_tstamp) = @submission_date
     {% endif %}
   QUALIFY
-    ROW_NUMBER() OVER (PARTITION BY happened_at ORDER BY happened_at) = 1
+    ROW_NUMBER() OVER (PARTITION BY event_id ORDER BY dvce_created_tstamp) = 1
 )
 SELECT
   p.prospect_id,
--- a/sql/moz-fx-data-shared-prod/snowflake_migration_derived/prospects_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/snowflake_migration_derived/prospects_v1/query.sql
@@ -33,6 +33,9 @@ WITH stg_prospects AS (
   WHERE
     event_name = 'object_update'
     AND unstruct_event_com_pocket_object_update_1.object = 'prospect'
+    {% if not is_init() %}
+      AND DATE(derived_tstamp) = @submission_date
+    {% endif %}
   QUALIFY
     ROW_NUMBER() OVER (PARTITION BY happened_at ORDER BY happened_at) = 1
 )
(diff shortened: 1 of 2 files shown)


scholtzan (approved):

sql/moz-fx-data-shared-prod/snowflake_migration_derived/prospects_v1/query.sql
```diff
@@ -33,6 +33,9 @@ WITH stg_prospects AS (
   WHERE
     event_name = 'object_update'
     AND unstruct_event_com_pocket_object_update_1.object = 'prospect'
+    {% if not is_init() %}
+      AND DATE(derived_tstamp) = @submission_date
```
sean-rose, line 37: issue (blocking): With [`date_partition_parameter: null` set in `metadata.yaml`](https://github.com/mozilla/bigquery-etl/blob/38bde6514ffd87cb31515b80786bde56d819f8f8/sql/moz-fx-data-shared-prod/snowflake_migration_derived/prospects_v1/metadata.yaml#L15) no `@submission_date` query parameter is being provided (so this won't work) and the ETL is overwriting the entire table (which I'm guessing you probably don't want to use the last day's worth of events to do).

If you intend for this to be a daily incremental table writing to a date-specific table partition each day, then you'll need to update the `metadata.yaml` to remove `date_partition_parameter: null`.

sql/moz-fx-data-shared-prod/snowflake_migration_derived/prospects_v1/query.sql
```diff
@@ -33,6 +33,9 @@ WITH stg_prospects AS (
...
+    {% endif %}
   QUALIFY
     ROW_NUMBER() OVER (PARTITION BY happened_at ORDER BY happened_at) = 1
```
sean-rose: As an aside, this `QUALIFY` filter doesn't make sense to me, as it's filtering out events with the exact same timestamp in an arbitrary manner (i.e. you aren't specifying how the ties should be handled and essentially leaving it up to chance which particular event with the same timestamp makes it through).

IMO events happening to have the exact same timestamp doesn't seem like it should be an issue.  Did you perhaps mean to partition by identifying columns, like is done in the `QUALIFY` clause in the final `SELECT` below?

sql/moz-fx-data-shared-prod/snowflake_migration_derived/prospects_v1/query.sql
jpetto: ah, good catch! this was the first table we migrated, and we were still learning the ins and outs of the system. will update!

sql/moz-fx-data-shared-prod/snowflake_migration_derived/prospects_v1/query.sql
jpetto: ah yeah, you're right. something was lost in translation when we originally built this model. i looked back at the [original DBT source](https://github.com/Pocket/dbt-snowflake/blob/e7880781e1c31a23e4afccabab008857edba2499/models/recsapi_models/prospect/prospects.sql#L47) which has the following:

`qualify row_number() over (partition by event_id order by dvce_created_tstamp) = 1 `

i believe this is intended to de-duplicate events sent from the same device. 

i'll update this model to use the above.

PR #7664, opened 2025-06-18
feat(DENG-8120): model various schedule rates over time and surface

jpetto:
## Description

model various schedule rates over time and surface. used for looker reporting.

* DENG-8927


sql/moz-fx-data-shared-prod/snowflake_migration_derived/scheduled_articles_report_v1/metadata.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - skamath@mozilla.com
+  - rrando@mozilla.com
+labels:
+  schedule: daily
+  incremental: true
+  owner1: skamath
+  owner2: rrando
+scheduling:
+  dag_name: bqetl_content_ml_daily
+bigquery:
```
scholtzan, line 14: The way this table is currently configured it expects the table to be partitioned (which is the default assumption). 

There are a few options:
1. Partition on `scheduled_date`. This only works if for each daily run the `scheduled_date` is identical for all records.

```
time_partitioning:
    field: scheduled_date
    type: day
```

2. Add a separate `submission_date` field and set it to the current `@submission_date` value, and partition the table based on that field.

```
time_partitioning:
    field: submission_date
    type: day
```

3. Override the entire table on every run (likely not what this should do), tell bigquery-etl that the table is not partitioned

```
scheduling:
```


My guess is that option 2 might be applicable here

jpetto: hmm, we need to query this table based on a range of `scheduled_date` - and afaik cannot do much else due to looker constraints. the value of `scheduled_date` will not be identical for all records on each run of this job. those being the constraints, i think option 3 is our only choice.

sql/moz-fx-data-shared-prod/snowflake_migration_derived/scheduled_articles_report_v1/metadata.yaml
```diff
@@ -0,0 +1,20 @@
+friendly_name: Scheduled Articles Report v1
+description: Model article approval, rejected, and removal rates per day per surface
+owners:
+  - skamath@mozilla.com
+  - rrando@mozilla.com
+labels:
+  schedule: daily
+  incremental: true
```
scholtzan:
```suggestion
  incremental: false
```

scholtzan (approved):

PR #7672, opened 2025-06-20
Support upstream dependencies for Fivetran tasks

scholtzan:
## Description

Required for https://mozilla-hub.atlassian.net/browse/AD-917

We need to be able to define upstream dependencies for Fivetran tasks as part of [AD-917](https://mozilla-hub.atlassian.net/browse/AD-917). This PR introduces `depends_on` config options for Fivetran tasks, as well as a `trigger_rule` option for these tasks.

[AD-917]: https://mozilla-hub.atlassian.net/browse/AD-917?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bigquery_etl/query_scheduling/templates/airflow_dag.j2
+++ b/bigquery_etl/query_scheduling/templates/airflow_dag.j2
@@ -200,6 +200,17 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
     )
 
     {% do fivetran_seen.append(fivetran_task) -%}
+
+    {% for fivetran_dependency in fivetran_task.depends_on | sort(attribute='task_id') -%}
+    {% if fivetran_dependency.dag_name == name and fivetran_dependency.get_execution_delta(schedule_interval) in [none, '0h', '0m', '0s'] -%}
+        {% if fivetran_dependency.task_id != task.task_name -%}
+        {{ fivetran_task.task_id }}_sync_start.set_upstream({{ fivetran_dependency.task_id }})
+        {% endif -%}
+    {% else -%}
+        {{ fivetran_task.task_id }}_sync_start.set_upstream(wait_for_{{ fivetran_dependency.task_id | replace('-', '_') }})
+    {% endif -%}
+    {% endfor -%}
+
     {% endif -%}
     {% endfor -%}
     {% endif -%}
--- a/tests/data/dags/simple_test_dag
+++ b/tests/data/dags/simple_test_dag
@@ -70,6 +70,7 @@ with DAG(
         trigger_rule="all_done",
     )
 
+    fivetran_import_1_sync_start.set_upstream(wait_for_fivetran_import_1)
     fivetran_import_2_sync_start = FivetranOperator(
         connector_id="{{ var.value.fivetran_import_2_connector_id }}",
         task_id="fivetran_import_2_task",


bigquery_etl/query_scheduling/task.py
```diff
@@ -205,6 +205,17 @@ class FivetranTask:
...
+    trigger_rule: Optional[str] = attr.ib(None)
+    depends_on: List[TaskRef] = attr.ib([])
+
+    @trigger_rule.validator
+    def validate_trigger_rule(self, attribute, value):
+        """Check that trigger_rule is a valid option."""
+        if value is not None and value not in set(rule.value for rule in TriggerRule):
+            raise ValueError(
+                f"Invalid trigger rule {value}. "
+                "See https://airflow.apache.org/docs/apache-airflow/1.10.3/concepts.html#trigger-rules for list of trigger rules"
```
sean-rose:
```suggestion
                "See https://airflow.apache.org/docs/apache-airflow/2.10.5/core-concepts/dags.html#trigger-rules for list of trigger rules"
```

bigquery_etl/query_scheduling/task.py
```diff
@@ -705,8 +716,14 @@ def with_upstream_dependencies(self, dag_collection):
...
         def _duplicate_dependency(task_ref):
+            fivetran_dependencies = [
+                dep.task_key
+                for fivetran_task in self.depends_on_fivetran
+                for dep in getattr(fivetran_task, "depends_on", [])
+            ]
             return any(
-                d.task_key == task_ref.task_key for d in self.depends_on + dependencies
+                d.task_key == task_ref.task_key
+                for d in self.depends_on + dependencies + fivetran_dependencies
```
sean-rose: This gives me pause, as I don't think we're eliminating direct dependencies based on there being transitive dependencies in any other cases, are we?

Unless there's a specific reason to do this, I'd be inclined to not omit the ETL's own dependency on the same upstream task the Fivetran task will be waiting on.

bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -194,9 +194,56 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
...
+        timeout={{ fivetran_dependency.timeout | format_timedelta | format_repr }},
+        {% endif -%}
+        allowed_states=ALLOWED_STATES,
+        failed_states=FAILED_STATES,
+        pool='DATA_ENG_EXTERNALTASKSENSOR',
+    )
+
+    {% do wait_for_seen.append(fivetran_dependency.task_key) -%}
+    {% endif -%}
+    {% endif -%}
```
sean-rose: It'd be kind of nice if we could continue to have all the `ExternalTaskSensor`s output together rather than having these be in a different location, and also having the `ExternalTaskSensor` constructor logic duplicated.

However, I can see why you opted to put this logic here as it'd be awkward to do the double-nested looping necessary in Jinja where the existing `ExternalTaskSensor` logic is.

Maybe a `Task` method could be added that would return all the Fivetran tasks' dependencies, and then those could simply be appended on [line 93](https://github.com/mozilla/bigquery-etl/blob/d4fa5798bff6c9e427c748d4efbb960c027f3e23/bigquery_etl/query_scheduling/templates/airflow_dag.j2#L93)?

bigquery_etl/query_scheduling/task.py
scholtzan: Yeah, you're right...

bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -90,7 +90,7 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
 {% set wait_for_seen = [] -%}
 {% set fivetran_seen = [] -%}
 {% for task in tasks | sort(attribute='task_name') %}
-    {% for dependency in (task.upstream_dependencies + task.depends_on) | sort(attribute='task_id') -%}
+    {% for dependency in (task.upstream_dependencies + task.depends_on + task.fivetran_dependencies()) | sort(attribute='task_id') -%}
```
sean-rose, line 93: I believe you'll want to do the same [here](https://github.com/mozilla/bigquery-etl/blob/d4fa5798bff6c9e427c748d4efbb960c027f3e23/bigquery_etl/query_scheduling/templates/airflow_dag.j2#L503).

sean-rose: Actually, never mind, because that would add a dependency from the ETL itself to the Fivetran task's intended dependencies.

sean-rose: But I think you'd want to keep the part of the previous Fivetran task logic you had that was setting upstream dependencies for the Fivetran task, which got removed in your latest commit.

bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -200,50 +200,6 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
...
-    {% endif -%}
-    {% endif -%}
-
-    {% if fivetran_dependency.dag_name == name and fivetran_dependency.get_execution_delta(schedule_interval) in [none, '0h', '0m', '0s'] -%}
-        {% if fivetran_dependency.task_id != task.task_name -%}
-        {{ fivetran_task.task_id }}_sync_start.set_upstream({{ fivetran_dependency.task_id }})
-        {% endif -%}
-    {% else -%}
-        {{ fivetran_task.task_id }}_sync_start.set_upstream(wait_for_{{ fivetran_dependency.task_id | replace('-', '_') }})
-    {% endif -%}
```
sean-rose: i.e. this part needs to be kept, otherwise the Fivetran task won't actually have its upstream dependencies set.

sean-rose: Also, I believe this logic needs to get moved down to where the other upstream dependency logic is, after the ETL tasks are generated.  Otherwise, if the Fivetran task is depending on a task in the same DAG it would be trying to reference a Python variable for the ETL task that hadn't been defined yet.

lucia-vargas-a (approved):

bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -194,9 +194,23 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
...
+        trigger_rule="{{ fivetran_task.trigger_rule }}",
+        {%+ endif -%}
     )
 
     {% do fivetran_seen.append(fivetran_task) -%}
+
+    {% for fivetran_dependency in fivetran_task.depends_on | sort(attribute='task_id') -%}
+    {% if fivetran_dependency.dag_name == name and fivetran_dependency.get_execution_delta(schedule_interval) in [none, '0h', '0m', '0s'] -%}
+        {% if fivetran_dependency.task_id != task.task_name -%}
+        {{ fivetran_task.task_id }}_sync_start.set_upstream({{ fivetran_dependency.task_id }})
```
sean-rose, line 207: Per [my earlier comment](https://github.com/mozilla/bigquery-etl/pull/7672#discussion_r2162644440), if this ever happens where a Fivetran task is trying to depend on an ETL task in the same DAG, I'm pretty sure that's going to result in a DAG error because the variable for that ETL task won't have been defined yet at this point.

PR #7684, opened 2025-06-24
Fix Fivetran task dependency template typo

scholtzan:
## Description

Follow up to https://github.com/mozilla/bigquery-etl/pull/7672#discussion_r2164543332

This fixes the Fivetran dependency task generation when a task is in the same DAG. The `set_upstream` code generation gets moved to the bottom to ensure any upstream tasks that are part of the same DAG appear before in the file.


bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -542,4 +532,18 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
 
     {% endfor -%}
     {% endif -%}
+
+    {% if task.depends_on_fivetran != None -%}
+    {% for fivetran_task in task.depends_on_fivetran -%}
```
sean-rose: There's already the same check and loop over `task.depends_on_fivetran` above, so I think the contents of this outer loop could be moved inside that existing loop.

bigquery_etl/query_scheduling/templates/airflow_dag.j2
```diff
@@ -542,4 +532,18 @@ with DAG('{{ name }}', default_args=default_args{%+ if schedule_interval != None
...
+    {% for fivetran_dependency in fivetran_task.depends_on | sort(attribute='task_id') -%}
+    {% if fivetran_dependency.dag_name == name and fivetran_dependency.get_execution_delta(schedule_interval) in [none, '0h', '0m', '0s'] -%}
+        {% if fivetran_dependency.task_id != fivetran_task.task_name -%}
+        {{ fivetran_task.task_id }}_sync_start.set_upstream({{ fivetran_dependency.task_id }})
+        {% endif -%}
+    {% else -%}
+        {{ fivetran_task.task_id }}_sync_start.set_upstream(wait_for_{{ fivetran_dependency.task_id | replace('-', '_') }})
+    {% endif -%}
```
sean-rose, line 540: If multiple ETLs depend on the same Fivetran task the same `set_upstream()` calls could get output multiple times (assuming each ETL's `depends_on_fivetran` metadata specified the nested `depends_on` metadata, which they ought to for consistency).  However, I don't think that would cause any actual problem, the DAG code would just be slightly duplicative.

sean-rose (approved):
r+wc

bigquery_etl/query_scheduling/templates/airflow_dag.j2
sean-rose: I've actually been thinking it'd be better to not have the Fivetran tasks be solely defined in ETL metadata, because when multiple ETLs depend on the same Fivetran task you have to be careful to keep that Fivetran dependency metadata exactly the same in all cases to avoid issues.

Maybe the Fivetran tasks could be defined at the DAG level in `dags.yaml`, and then the ETL dependencies on those Fivetran tasks could be specified via the normal `depends_on` metadata?

scholtzan: Right, it shouldn't cause any issues to have duplicates, but not ideal.

Yeah, we could rethink how Fivetran tasks are specified, though that should go through the proposal process.

PR #7717, opened 2025-07-02
feat(): Update new profile clients view to point to new table (now that its been backfilled)

kik-kik:
# feat(): Update new profile clients view to point to new table (now that its been backfilled)


bqetl_project.yaml
```diff
@@ -499,28 +500,45 @@ retention_exclusion_list:
...
-- sql/moz-fx-data-shared-prod/acoustic_external/suppression_list_v1
+- sql/moz-fx-data-shared-prod/fenix_derived/clients_yearly_v1
+- sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v1  # TODO: check if we still need this
+- sql/moz-fx-data-shared-prod/fenix_derived/attributable_clients_v2  # TODO: check if we still need this
 - sql/moz-fx-data-shared-prod/telemetry_derived/fx_accounts_linked_clients_v1
+# marketing tables
+- sql/moz-fx-data-shared-prod/acoustic_external/suppression_list_v1
+# mobile profile tables
+- sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1
+- sql/moz-fx-data-shared-prod/fenix_derived/new_profile_clients_v1
```
kik-kik, line 532: new_profile_clients_v1 and new_profile_activation_clients_v1 entries added to the list for now. Once a switch over of downstream dependencies is completed and all users of dashboards affected are informed we will remove this from the list and set our standard retention policy on those tables.

bqetl_project.yaml
```diff
@@ -499,28 +500,45 @@ retention_exclusion_list:
...
 - sql/moz-fx-data-shared-prod/fenix_derived/attribution_clients_v1
 - sql/moz-fx-data-shared-prod/firefox_ios_derived/attribution_clients_v1
 - sql/moz-fx-data-shared-prod/focus_android_derived/attribution_clients_v1
 - sql/moz-fx-data-shared-prod/focus_ios_derived/attribution_clients_v1
 - sql/moz-fx-data-shared-prod/klar_android_derived/attribution_clients_v1
 - sql/moz-fx-data-shared-prod/klar_ios_derived/attribution_clients_v1
+#
 - sql/moz-fx-data-shared-prod/firefox_desktop_derived/adclick_history_v1
 - sql/moz-fx-data-shared-prod/fenix_derived/client_adclicks_history_v1
-- sql/moz-fx-data-shared-prod/fenix_derived/clients_yearly_v1
```
kik-kik, line 515: These are not being removed, just reschufling the order in the list to have a logic grouping

kwindau (approved):
Approving, my only nitpick is that the new_profile_clients is switching from joining on client_id only to client_id, normalized channel between active_users and attribution_clients, and in attribution_clients, normalized_channel is currently always null prior to 2023-04-07, so if someone were to in the future re-backfill this, there could be issues prior to that date if they did a backfill greater than 775 days.  One way to prevent this could be adding logic to make the join condition based on the date also (i.e. use client ID only prior to that date, and client_id/normalized_channel after that date).  But this only poses an issue if someone backfills that far back.

sql_generators/mobile_kpi_support_metrics/templates/new_profile_activation_clients.query.sql
```diff
@@ -9,7 +9,6 @@ WITH new_profiles AS (
     city,
     geo_subdivision,
     locale,
-    isp,
```
kwindau, line 12: Would getting rid of isp cause any issues downstream?

kik-kik:
no longer valid.

PR #7720, opened 2025-07-02
feat(): stores update clients references to new profile clients

kik-kik:
# feat(): stores update clients references to new profile clients

This PR aims to update references to firefox_android_clients and firefox_ios_clients to the new new_profile_clients datasets so that the former can be deprecated.

### DIFF
--- a/sql/moz-fx-data-shared-prod/google_ads_derived/android_app_campaign_stats_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/google_ads_derived/android_app_campaign_stats_v1/query.sql
@@ -20,12 +20,12 @@ WITH daily_stats AS (
 activations AS (
   SELECT
     first_seen_date AS `date`,
-    ad_group_id,
-    COUNTIF(activated) AS activated,
+    CAST(REGEXP_EXTRACT(adjust_ad_group, r' \((\d+)\)$') AS INT64) AS ad_group_id,
+    COUNTIF(is_activated) AS activated,
     COUNT(*) AS new_profiles,
     SUM(lifetime_value) AS lifetime_value,
   FROM
-    `moz-fx-data-shared-prod.fenix.firefox_android_clients`
+    `moz-fx-data-shared-prod.fenix.new_profile_activation_clients`
   JOIN
     `moz-fx-data-shared-prod.ltv.fenix_client_ltv`
     USING (client_id)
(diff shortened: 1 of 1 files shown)


kwindau:
Would we be able to make these changes in a V2 table, that way you can backfill the V2 and compare with V1? Then once we are satisfied everything looks good and have run QA checks, we can swap the views to point at the new V2 tables and deprecate the V1 tables at that time?

soGaussian:
The iOS app campaign table is not needed right now.

kik-kik:
> The iOS app campaign table is not needed right now.

@soGaussian are you saying `sql/moz-fx-data-shared-prod/apple_ads_external/ios_app_campaign_stats_v1/query.sql` should be deprecated?

kik-kik:
> > The iOS app campaign table is not needed right now.
> 
> @soGaussian are you saying `sql/moz-fx-data-shared-prod/apple_ads_external/ios_app_campaign_stats_v1/query.sql` should be deprecated?

ok, I will create a follow-up ticket then to deprecate that dataset.

sql/moz-fx-data-shared-prod/apple_ads_external/ios_app_campaign_stats_v1/query.sql
kik-kik, line 1: This has now been marked as deprecated.

sql/moz-fx-data-shared-prod/firefox_ios/ad_activation_performance/view.sql
kik-kik, line 1: Found some differences, dived into a specific client and it appears the new table is reporting activation field correctly. The old dataset not (as per the activation logic). Hard to say why exactly, but this could be because the underlaying dataset we rely on for calculating this changes somehow or potentially the definition of activation changed since when the activation information for that client was generated. 

We only see differences for dates older than (including) 2024 August.

kik-kik: ok, so historically there's definitely bigger differences. And seeing that the aggregation is generated at runtime using client level info so even historically I'd expect more alignment since both datasets should be Shredded so this should not be a factor for these differences. The only thing I can really think of is that since the differences between new_profile count and activation count show up starting August 2024 is that the definition of activation changes around then and this is the definition used by the new dataset (since we seem to have match since then).

Another theory would be that the new dataset was created and backfilled around the time we see differences and historical data differs because at the backfill time the underlaying data (used to create new profiles and activation datasets) may have changed which is why historically we see differences and going forward we do not as both datasets would be updated around the same time.

kik-kik: @soGaussian would you be the main consumer of this dataset?

PR #7728, opened 2025-07-03
feat(): update ltv client references to new new profile clients

kik-kik:
# feat(): update ltv client references to new new profile clients

This is work towards deprecating firefox_ios_clients and firefox_android_clients

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_ios_derived/ltv_states_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_ios_derived/ltv_states_v1/query.sql
@@ -60,16 +60,16 @@ SELECT
   ac.durations,
   ac.ad_clicks,
   ac.total_historic_ad_clicks,
-  c.adjust_network,
-  c.first_reported_country,
-  c.first_reported_isp
+  clients.adjust_network,
+  clients.country AS first_reported_country,
+  CAST(NULL AS STRING) AS first_reported_isp,
 FROM
   ad_clicks ac
 JOIN
-  `moz-fx-data-shared-prod.firefox_ios.firefox_ios_clients` c
-  USING (sample_id, client_id)
+  `moz-fx-data-shared-prod.firefox_ios.new_profile_clients` AS clients
+  USING (client_id)
 WHERE
     -- BrowserStack clients are bots, we don't want to accidentally report on them
-  COALESCE(first_reported_isp, '') != "BrowserStack"
+  app_name <> "Firefox iOS BrowserStack"
     -- Remove clients who are new on this day, but have more/less than 1 day of activity
   AND NOT (days_since_first_seen = 0 AND BIT_COUNT(days_seen_bytes) != 1)


sql/moz-fx-data-shared-prod/fenix_derived/ltv_states_v1/query.sql
```diff
@@ -42,20 +42,20 @@ SELECT
...
-  USING (sample_id, client_id)
+  `moz-fx-data-shared-prod.fenix.new_profile_clients` AS clients
+  USING (client_id)
 LEFT JOIN
   `moz-fx-data-shared-prod.fenix.client_adclicks_history`
   USING (sample_id, client_id)
 WHERE
     -- BrowserStack clients are bots, we don't want to accidentally report on them
-  first_reported_isp != "BrowserStack"
+  app_name <> "Fenix BrowserStack"
```
kik-kik, line 59: Should we also be exluding `Mozilla Online` here @kwindau ?

sql/moz-fx-data-shared-prod/firefox_ios_derived/ltv_states_v1/query.sql
```diff
@@ -60,16 +60,16 @@ SELECT
   ac.durations,
   ac.ad_clicks,
   ac.total_historic_ad_clicks,
-  c.adjust_network,
-  c.first_reported_country,
-  c.first_reported_isp
+  clients.adjust_network,
+  clients.country AS first_reported_country,
+  CAST(NULL AS STRING) AS first_reported_isp,
```
kik-kik, line 65: `isp` field no longer available. As far as I recall the only purpose we used it for was to identify bots. This is now done upstream and this filtering can be done via `app_name`

sql/moz-fx-data-shared-prod/fenix_derived/client_ltv_v1/schema.yaml
```diff
@@ -2,7 +2,7 @@ fields:
   - name: client_id
     mode: NULLABLE
     type: STRING
-    description: "Client ID; uniquely identifies a client. Joinable with fenix.firefox_android_clients."
+    description: "Client ID; uniquely identifies a client."
```
kik-kik, line 5: Removed the last part since it's joinable with anything on client_id.

sql/moz-fx-data-shared-prod/fenix_derived/ltv_states_v1/query.sql
kwindau: I think we are only excluding BrowserStack due to bots, not due to DAU definitions (this is more LTV focused than DAU focused). I think it's ok to leave as is, thanks for checking

kik-kik: @kwindau one more question, is there a use case where we may want to look at this data and filter out Mozilla Online users?

PR #7730, opened 2025-07-03
fix: metadata CODEOWNERS metadata validation in CI

kik-kik:
# fix: metadata CODEOWNERS metadata validation in CI

This is because in CI when deploying to staging the files get written to directories that do not match those directories defined in the CODEWONERS file. This change aims to add support to validate_change_control function for path with wildcards in CODEOWNERS.

Here's an example error message we get in our CLI currently when trying to edit `active_users_aggregates_v3` generator:

```
The Failed to validate 5 metadata files is because path in CODEOWNERS doesn't match the path of the generated queries in the CI
```

e.g. focus_ios_derived_moz_fx_data_shared_prod_03803ba9d249f6bbca0b1337ddb4c1e2e5afe6e9/active_users_aggregates_v3

This seems to be related to how we handle the query namespace in for our staging testing.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,10 +1,10 @@
 # These datasets are subject to the additional change control procedures
 # described in https://docs.google.com/document/d/1TTJi4ht7NuzX6BPG_KTr6omaZg70cEpxe9xlpfnHj9k/
 # Active Users
-sql_generators/active_users_aggregates_v3/templates/ @mozilla/kpi_table_reviewers
-sql/**/active_users_aggregates_v3 @mozilla/kpi_table_reviewers
-sql_generators/active_users_aggregates_v4/templates/ @mozilla/kpi_table_reviewers
-sql/**/active_users_aggregates_v4/ @mozilla/kpi_table_reviewers
+/sql_generators/active_users_aggregates_v3/templates/ @mozilla/kpi_table_reviewers
+/sql/**/active_users_aggregates_v3 @mozilla/kpi_table_reviewers
+/sql_generators/active_users_aggregates_v4/templates/ @mozilla/kpi_table_reviewers
+/sql/**/active_users_aggregates_v4 @mozilla/kpi_table_reviewers
 # Search
 /sql/moz-fx-data-shared-prod/search_terms @whd @jasonthomas
 /sql/moz-fx-data-shared-prod/search_terms_derived @whd @jasonthomas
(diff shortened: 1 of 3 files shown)


CODEOWNERS
```diff
@@ -1,16 +1,10 @@
...
-/sql/moz-fx-data-shared-prod/firefox_desktop_derived/active_users_aggregates_v3 @mozilla/kpi_table_reviewers
-/sql_generators/active_users_aggregates_v4/templates/ @mozilla/kpi_table_reviewers
-/sql/moz-fx-data-shared-prod/firefox_desktop_derived/active_users_aggregates_v4/ @mozilla/kpi_table_reviewers
-/sql/moz-fx-data-shared-prod/fenix_derived/active_users_aggregates_v3/ @mozilla/kpi_table_reviewers
-/sql/moz-fx-data-shared-prod/firefox_ios_derived/active_users_aggregates_v3/ @mozilla/kpi_table_reviewers
-/sql/moz-fx-data-shared-prod/focus_android_derived/active_users_aggregates_v3/ @mozilla/kpi_table_reviewers
-/sql/moz-fx-data-shared-prod/focus_ios_derived/active_users_aggregates_v3/ @mozilla/kpi_table_reviewers
-/sql/moz-fx-data-shared-prod/klar_ios_derived/active_users_aggregates_v3/ @mozilla/kpi_table_reviewers
-/sql/moz-fx-data-shared-prod/klar_android_derived/active_users_aggregates_v3/ @mozilla/kpi_table_reviewers
+sql_generators/active_users_aggregates_v3/templates/ @mozilla/kpi_table_reviewers
```
kik-kik: @BenWu I'm not sure about the starting `/`, is the intention here that we're expecting these directories to exist in the root directory?

BenWu: Yes without the `/` it will match any path in the repo even if it's nested https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file

BenWu (approved):
LGTM but I'm wondering if we could just use the non-stage generated sql for metadata validation

kik-kik:
> LGTM but I'm wondering if we could just use the non-stage generated sql for metadata validation

@BenWu I'll leave this up to the data platform to decide. Feel free to make changes as required or to merge this.

PR #7744, opened 2025-07-08
feat(DENG-8178): build search_clients_daily_v9

kbammarito:
## Description

This PR builds `search_clients_daily_v9` from Glean (migrating from legacy telemetry).

* [DENG-8178](https://mozilla-hub.atlassian.net/browse/DENG-8178)

[DENG-8178]: https://mozilla-hub.atlassian.net/browse/DENG-8178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,333 @@
...
+adblocker_addons AS (
+  SELECT
+  addon_id,
+  addon_name
+  FROM
+  moz-fx-data-shared-prod.revenue.monetization_blocking_addons
+  WHERE
+  blocks_monetization
+),
+-- This table is the Glean equivalent to moz-fx-data-shared-prod.telemetry.clients_daily
```
betling: we can delete this comment since we are doing this differently now with events instead of glean scalars

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,333 @@
...
+    -- date(submission_timestamp) = '2025-06-01'
+    -- and sample_id = 0
+    date(submission_timestamp) = @submission_date
+    AND document_id IS NOT NULL
+  GROUP BY
+    client_id,
+    submission_date
+),
+
+-- join serp to sap including as much client info as possible via sap.counts ping
```
betling: can delete this first comment and leave the next stating what the cte does

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,333 @@
...
+    -- submission_date = '2025-06-01'
+    -- and sample_id = 0
+),
+
+-- this CTE gets all client info and other metrics that are not available in the glean sap.counts events ping
+glean_metrics AS (
+  SELECT
+    client_info.client_id AS client_id,
+    DATE(submission_timestamp) AS submission_date,
+    metrics.string.search_engine_default_display_name AS default_search_engine,
```
Standard8: This field used to be an identifier which actually contained several pieces of information. Per the spreadsheet, I think we should turn this into:

```
metrics.string.search_engine_default_provider_id AS default_search_engine_provider_id,
metrics.string.search_engine_default_partner_code AS default_search_engine_partner_code,
metrics.boolean.search_engine_default_overridden_by_third_party AS default_search_engine_overridden,
```

Those should be the new keys that we use for searches relating to application provided search engines.

I think we could keep the display name, as that's useful for third-party engines, however, I think we should be explicit about what the field is, i.e.

```
metrics.string.search_engine_default_display_name AS default_search_engine_display_name
```

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,333 @@
...
+    metrics.string.search_engine_default_load_path AS default_search_engine_data_load_path,
+    metrics.url2.search_engine_default_submission_url AS default_search_engine_data_submission_url,
+    metrics.string.search_engine_private_display_name AS default_private_search_engine,
```
Standard8: Same comment applies to the `_private` variations.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,333 @@
...
+    AND event.name = 'counts'
+),
+
+-- this CTE gets serp event data
+serp_events AS (
+  SELECT
+    submission_date,
+    impression_id,
+    glean_client_id as client_id,
+    -- SYDNEY AND MARK: do we even need this serp_search_access_point? If we have it from the sap_count ping?
```
skahmann3: When you do the merge, is it consistent across data sources?

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,333 @@
...
+    metrics.string.search_engine_private_load_path AS default_private_search_engine_data_load_path,
+    metrics.url2.search_engine_private_submission_url AS default_private_search_engine_data_submission_url,
+    metrics.string.region_home_region AS user_pref_browser_search_region,
+    metrics.labeled_counter.browser_is_user_default as is_default_browser,
+    metrics.quantity.browser_engagement_max_concurrent_tab_count as browser_engagement_max_concurrent_tab_count,
+    metrics.counter.browser_engagement_active_ticks as browser_engagement_active_ticks,
+    metrics.counter.browser_engagement_tab_open_event_count as browser_engagement_tab_open_event_count,
+    metrics.counter.browser_engagement_uri_count
+    -- SYDNEY AND MARK: do you want that main-ping-like session counts but using glean event timing and sequences?
+    -- ping_info.start_time AS subsession_start_date, -- ping_info.parsed_start_time instead?
```
skahmann3: Does Mark have thoughts on this one? Not familiar with the differences between these fields

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,333 @@
...
+-- This CTE aggregates any fields that need to be counted or summed
+
+sap_serp_glean_aggregates AS (
+
+   SELECT
+      client_id,
+      submission_date,
+      -- this is sap_clients_events stuff getting aggregated
+      COUNT(*) AS sap,
+      -- SYDNEY AND MARK: do we need subsession_hours_sum? This existed in the main ping. We can extract something similar from Glean, but do we need it?
```
skahmann3: I've never used it

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,333 @@
...
+  provider_name,
+  provider_id,
+  partner_code,
+  overridden_by_third_party,
+  search_access_point,
+
+  --sap aggregates
+  sap,
+  -- SYDNEY AND MARK: do we use sap or sap_count (this is the same as sap)?
+  sap AS sap_count,
```
skahmann3: I vote sap so we don't break anything built on the user-facing table: mozdata.search.search_clients_engines_sources_daily

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,333 @@
...
+  normalized_os_version,
+  sample_id,
+  metrics.uuid.legacy_telemetry_profile_group_id AS profile_group_id,
+  metrics.uuid.legacy_telemetry_client_id AS legacy_telemetry_client_id, -- adding this for now so people can join to it if needed
+  ping_info.experiments as ping_info_experiments,
+  ping_info.parsed_start_time as subsession_start_date,
+  ping_info.end_time as subsession_end_time,
+  ping_info.seq as subsession_counter,
+  -- SYDNEY AND MARK: we need alignment on whether to keep the normalized_engine name or provider_name
+  mozfun.map.get_key(event.extra, 'provider_name') as normalized_engine,
```
skahmann3: I vote normalized_engine. Otherwise, we'll need to fix mobile_clients_daily, all the aggregates tables, corresponding Looker explores, and breaks lots of other queries

Standard8: I think ideally, I'd like us to start using `provider_id` and `provider_name` since that matches what the client reports, this makes it easier for engineers to understand what's going on. For example, as an engineer, I know the previous field was generated with `moz-fx-data-shared-prod.udf.normalize_search_engine(scd.engine)`, but that function is private and afaict undocumented, so I can't tell what it does or what its values are without asking.

However, I think I wouldn't be against keeping the `normalized_engine` field depending on what it does, since it would make it easier to query for cases where the user has a Google-like engine installed. For example with the two separate fields, this would be something like:

`provider_id = "google" OR provider_name LIKE "%Google%`.

For filling the `normalized_engine` field, I think we should do something like this:

```
if provider_id = "other" then
  normalized_engine = normalize_search_engine(provider_name)
else 
  normalized_engine = normalize_search_engine(provider_id)
```

That's not sql, but hopefully you get the right idea. I'm assuming we'd still want to wrap the `provider_id` call with `normalize_search_engine`, but as I don't know what it does, I'm not sure. If it helps, this is the list of ids you'll get: https://searchfox.org/mozilla-central/search?q=identifier&path=services%2Fsettings%2Fdumps%2Fmain%2Fsearch-config-v2.json&case=false&regexp=false

The reason is that `provider_id` is guaranteed to be stable for an application provided engine. If we change the display name of an engine, that will update the `provider_name`. The `provider_name` field is really intended for giving information about search engines installed by the user, since there's no id for those.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
Standard8: We do need these for the merge. The SERP events are not always the same source as the SAP, and there are generally more SERPs than SAPs.
 
The possible SAP sources are also a subset of SERP sources.

Though I've just realised that there's some differences between the two, e.g. SAP uses `urlbar-handoff` and SERP uses `urlbar_handoff`. I'm wondering if as it is still "young" if we should fix `sap.counts` reporting to use the same version as the SERP. This would then allow us to do queries to check that various source for SAP result in the same sources for SERP. @skahmann3 Any thoughts on this?

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
Standard8: Sorry, I have no idea about what these are. Generally I've only ever used submission dates.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
Standard8: I don't have an opinion here. I'm not sure what this is for.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
Standard8: It'd be nice to update it to be consistent with the new SERP fields, but I'm fine with keeping it as it is to reduce the need to change things.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,333 @@
...
+    num_ad_clicks,
+    num_non_ad_link_clicks,
+    e.num_other_engagements,
+    num_ads_loaded,
+    num_ads_visible,
+    num_ads_blocked,
+    num_ads_notshowing,
+    ad_blocker_inferred,
+    ad_components.component as ad_click_target
+  FROM `mozdata.firefox_desktop.serp_events` e
```
Standard8: I realise the current `serp_events` doesn't have it, but we could really do with `partner_code` here as well (can we get it added there?

This would allow for better breakdowns by partner code, something that we can't analyse today with the existing tables, yet we should really be taking it into account.

Standard8: We should also have something like `search_engine as provider_id`. Similar to the `sap_source` field, the user may be searching different providers which aren't triggered via the normal search access points. Hence we need to include those as well.

Though could we also change "duckduckgo" to "ddg" within the field so that it matches the `provider_id` field for SAP and the `search_engine_default_provider_id` for the default search engine?

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,333 @@
...
+  --     ) / NUMERIC '3600'
+  -- ) AS subsession_hours_sum,
+      -- this is serp_events stuff getting aggregated
+      COUNT(DISTINCT impression_id) AS serp_impression_count,
+      COUNTIF(is_tagged = TRUE) AS serp_searches_tagged_count,
+      COUNTIF(is_tagged = FALSE) AS serp_organic_count,
+      -- MARK: is this the right way to get follow on searches?
+      COUNTIF((is_tagged = TRUE)
+        AND (search_access_point = 'follow_on_from_refine_on_incontent_search'
+        OR search_access_point = 'follow_on_from_refine_on_SERP')) AS serp_follow_on_searches_tagged_count,
```
Standard8: Yes, that's currently the best way.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,333 @@
...
+  sample_id,
+  profile_group_id,
+  legacy_telemetry_client_id, -- adding this for now so people can join to it if needed
+  experiments,
+  normalized_engine,
+  provider_name,
+  provider_id,
+  partner_code,
+  overridden_by_third_party,
+  search_access_point,
```
Standard8: As noted above, `provider_id`, `partner_code`, `search_access_point` should be from both sap & serp.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,333 @@
...
+  -- SYDNEY AND MARK: do we use sap or sap_count (this is the same as sap)?
+  sap AS sap_count,
+
+  --glean metrics
+  default_search_engine,
+  default_search_engine_data_load_path,
+  default_search_engine_data_submission_url,
+  default_private_search_engine,
+  default_private_search_engine_data_load_path,
+  default_private_search_engine_data_submission_url,
```
Standard8: As noted above, we'll want some slightly different fields than the `default_search_engine` and `default_private_search_engine` ones.

Standard8 (commented):
Thank you for working on this. It is great to see this moving along. Sorry for the delay in responding, there was lots of thinking and checking to do.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
Standard8: We might want to also add the normalized_engine field in here as well - but only filled in from `search_engine` (there's no name field for SERPs, since there's a limited amount.

This would make it easier to search for both tagged and organic SERPs using the normalized_engine field, which you might want to do if you're also including SAP data.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: OK, we removed this

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: OK, we removed this

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: OK

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: OK

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: OK

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: Need to double check this

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: Leaving this as is for now. Let us know what you decide.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: @Standard8: with your first comment, are we understanding correctly that you want `partner_code` added to the `serp_events` table itself (not just us pulling it in through another CTE)?

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: Done

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: @Standard8: we don't see the `provider_id` and `partner_code` fields in the `serp_events` table. Do you want us to pull them in from `serp_impressions` ping (it has `provider` and `partner_code`)? Or are you saying they should be added to the `serp_events` table?

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: We can keep both if that's preferable. Let us know.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: We renamed these as directed

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: done

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
Standard8: Sorry, I realise I wasn't too clear here. Yes, the provider and partner_code data need pulling in from the `serp.impression` event (from `serp_events`). They're dedicated to that event, you can't pull them from the `sap.counts` event, since they're different events with potentially different totals per provider/partner code.

So, in summary, I think this `serp_events` cte needs the following adding:

```
CASE
  WHEN search_engine = "duckduckgo" then "ddg"
  ELSE search_engine
END as provider_id,
partner_code,
```

(This will need `partner_code` also adding to the `mozdata.firefox_desktop.serp_events` table...).

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
Standard8: I believe `provider_id` (aka `provider`) is in the `serp_events` table as [`search_engine`](https://github.com/mozilla/bigquery-etl/blob/d8d41d8a7e34d12fc375d0d262ead4e08e6336ee/sql_generators/serp_events_v2/templates/desktop_query.sql#L72).

I think the `partner_code` should also be added to `serp_events` table if possible - That's an important aspect that we're currently missing.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
Standard8: Let's just go with `sap`. We should document these fields somewhere & can explain it there.

Standard8 (commented):
I've added a few responses. I'm assuming the "done" comments are ones that have been done locally, and not yet pushed to this PR.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
Standard8: > I vote sap so we don't break anything built on the user-facing table: mozdata.search.search_clients_engines_sources_daily

@skahmann3 one thing to note is that the SERP fields are going to be named differently. I think that is important that we change those names, as the existing names are confusing & tend to be unclear.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,333 @@
...
+  is_default_browser,
+
+  --serp events
+  ad_blocker_inferred,
+  ad_click_target
+
+  --serp aggregates
+  serp_impression_count,
+  serp_searches_tagged_count,
+  serp_organic_count,
```
Standard8: `serp_organic_count` -> `serp_searches_organic_count`.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,333 @@
...
+  default_private_search_engine_data_submission_url,
+  user_pref_browser_search_region,
+  is_default_browser,
+
+  --serp events
+  ad_blocker_inferred,
+  ad_click_target
+
+  --serp aggregates
+  serp_impression_count,
```
Standard8: I'm not sure we need this. Isn't this the same as `serp_searches_tagged_count + serp_searches_organic_count` ?

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,333 @@
...
+  serp_searches_tagged_count,
+  serp_organic_count,
+  serp_follow_on_searches_tagged_count,
+  serp_with_ads_organic_count,
+  serp_with_ads_tagged_count,
+  serp_ad_clicks_organic_count,
+  serp_ad_clicks_tagged_count,
+  serp_ad_click_sum,
```
Standard8: Likewise, I'm not sure we need this - it seems to be the count of the organic + tagged.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,333 @@
...
+  serp_ad_blocker_inferred,
+  serp_ad_clicks_target,
+  serp_impression_id,
```
Standard8: I don't think we need the impression id, as this is a summary table?

(sorry for the extra comments, I was just thinking about a few things).

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
betling: no, this is the sum of impressions per client and submission date in the serp_events table (since that table is by impression_id, date).  having this will solve an issue for the nimbus team.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
betling: on this one you are right, we don't need this.  this accidentally made it through from a testing query.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
betling: That is probably correct.  I had only included it because i believe the ad_blocker_inferred and ad_click_target arrays are indexed and/or ordered by them.  kirin is working on optimizing the query and will see if it is needed for that and if not we will remove.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
betling: @Standard8 DE can add partner_code to the serp_events table.  That is not a big lift.  Does it need to be backfilled?  I understand that will take a much longer time if needed but that they could add it to their backlog.  @kbammarito or @alekhyamoz can correct me as needed.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
skahmann3: Yes, we should make the sources match. Open to whether the urlbar-handoff fix is easiest in code or as a UDF

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
skahmann3: Yes SERP fields will have different names. That's fine as long as the updated search metrics (which are equivalent to current legacy metrics) have the same naming so as not to break things/ introduce different trends

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
skahmann3: That's a great idea, Mark. Let's have both with normalized_engine being the one basic users query and the provider fields being the more detailed options. If the provider metrics prove to be useful, maybe we can phase out normalized_engine more naturally over time

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: We've fixed this in the query to be `urlbar_handoff` for SAP

Standard8: > We've fixed this in the query to be `urlbar_handoff` for SAP

Thank you. We have also [fixed it in the client](https://bugzilla.mozilla.org/show_bug.cgi?id=1979318) as this will let us simplify the code in future.

However, let's leave it in the query for now at least - there's still going to be existing versions reporting the old form for a while, so I don't think it hurts to have it changed in both places.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
Standard8: Presumably we can request backfill at a later time? If so, I'd suggest leaving it for now, and if we need a bit more history when we start using the new table, then we can request it.

skahmann3: I believe the plan for "backfill" was to create another versioned table. We'd do a small backfill to ensure metric trends haven't changed, but the final table will concatenate the two tables: v1 if < date and v2 if > date. Stakeholders will keep using legacy data for past results, but future results will be transitioned to the new metrics.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
betling: oh i think i see what you are saying now.  impression count is the same as `serp_searches_tagged_count` + `serp_searches_organic_count` so if nimbus needs it (i don't think they actually do anymore) then we could just create a metric in metric-hub that is `serp_searches_tagged_count` + `serp_searches_organic_count`.  we can drop the impression count.  i'll leave this as unresolved until @kbammarito has a chance to change the code so it doesn't get forgotten.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,948 @@
...
+    `moz-fx-data-shared-prod.udf.normalize_search_engine`(
+      search_engine
+    ) AS serp_provider_id, -- this is engine
+    sap_source AS serp_search_access_point,
+    LOGICAL_AND(ad_blocker_inferred) AS serp_ad_blocker_inferred,
+    COUNTIF(
+      (is_tagged IS TRUE)
+      AND (
+        sap_source = 'follow_on_from_refine_on_incontent_search'
+        OR sap_source = 'follow_on_from_refine_on_serp'
```
irrationalagent: This filter compares against lowercase 'follow_on_from_refine_on_serp', but the emitted value is uppercase SERP. As a result this access point contributes exactly zero follow-on, despite carrying real tagged volume.

I checked against your `kbam_glean_2april` table, rows with source = 'follow_on_from_refine_on_SERP' have non-zero tagged_sap but tagged_follow_on is 0 across every row, while the similar follow_on_from_refine_on_incontent_search maps 1:1 as expected. 

Volume is small but i think its a genuine bug.

Suggestion: LOWER(sap_source) = 'follow_on_from_refine_on_serp' or matching the emitted casing.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,948 @@
...
+    serp_ad_blocker_inferred AS ad_blocker_inferred,
+    serp_num_non_ad_link_clicks AS num_non_ad_link_clicks, -- NEW
+    serp_num_other_engagements AS num_other_engagements, -- NEW
+    serp_num_ads_loaded AS num_ads_loaded, -- NEW
+    serp_num_ads_visible AS num_ads_visible, -- NEW
+    serp_num_ads_blocked AS num_ads_blocked, -- NEW
+    serp_num_ads_notshowing AS num_ads_notshowing, -- NEW
+    NULL AS unknown,
+    NULL AS is_sap_monetizable, -- REVISIT
+    serp_has_adblocker_addon AS has_adblocker_addon,
```
irrationalagent: has_adblocker_addon will never be FALSE, only TRUE or NULL

clients_with_adblocker_addons_cte only ever emits TRUE AS has_adblocker_addon (line 39), and it's left-joined, so non-matching clients come through as NULL rather than FALSE.

checked this against your table, the column takes only two states:  roughly a third of rows TRUE and the remaining two thirds NULL, with no rows set to FALSE. 

v8 produces a proper boolean over the same population, handles this at search_clients_daily_v8/query.sql:86:

```sql
COALESCE(has_adblocker_addon, FALSE) AS has_adblocker_addon
```

scholtzan:
### Integration report
- [View full diff, if available (SQL + DAGs)](https://github.com/mozilla/private-bigquery-etl/blob/bqetl-diff/pr-7744/sql.diff) *(requires private-bigquery-etl access)*
- [Private CI run](https://github.com/mozilla/private-bigquery-etl/actions/runs/31223609059)
- Bigeye monitoring dry-run: ✅ plan clean — [view plan](https://github.com/mozilla/private-bigquery-etl/blob/bqetl-bigeye-plan/bqetl-pr-7744/bigeye_plan.yml) *(requires private-bigquery-etl access)*

scholtzan:
### Integration report
- [View full diff, if available (SQL + DAGs)](https://github.com/mozilla/private-bigquery-etl/blob/bqetl-diff/pr-7744/sql.diff) *(requires private-bigquery-etl access)*
- [Private CI run](https://github.com/mozilla/private-bigquery-etl/actions/runs/31846034080)
- Bigeye monitoring dry-run: ✅ plan clean — [view plan](https://github.com/mozilla/private-bigquery-etl/blob/bqetl-bigeye-plan/bqetl-pr-7744/bigeye_plan.yml) *(requires private-bigquery-etl access)*

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/ctes/13_join_sap_serp.sql
```diff
@@ -0,0 +1,78 @@
...
+  serp_final_cte.profile_age_in_days AS serp_profile_age_in_days,
+  serp_final_cte.serp_counts,
+  serp_final_cte.sessions_started_on_this_day AS serp_sessions_started_on_this_day,
+  serp_final_cte.active_hours_sum AS serp_active_hours_sum,
+  serp_final_cte.serp_scalar_parent_browser_engagement_tab_open_event_count_sum AS serp_scalar_parent_browser_engagement_tab_open_event_count_sum,
+  serp_final_cte.serp_scalar_parent_browser_engagement_total_uri_count_sum AS serp_scalar_parent_browser_engagement_total_uri_count_sum,
+  serp_final_cte.max_concurrent_tab_count_max AS serp_max_concurrent_tab_count_max
+FROM
+  `search_derived.search_clients_daily_v9.serp_final_cte`
+LEFT JOIN
```
irrationalagent: Flagging for discussion, this now ensures that SAP-only counts are dropped. I still think we will want a join pattern that preserves those.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/ctes/14_final.sql
```diff
@@ -0,0 +1,96 @@
...
+  serp_ad_clicks_tagged_count AS ad_clicks_tagged,
+  serp_ad_blocker_inferred AS ad_blocker_inferred,
+  serp_num_non_ad_link_clicks AS num_non_ad_link_clicks, -- NEW
+  serp_num_other_engagements AS num_other_engagements, -- NEW
+  serp_num_ads_loaded AS num_ads_loaded, -- NEW
+  serp_num_ads_visible AS num_ads_visible, -- NEW
+  serp_num_ads_blocked AS num_ads_blocked, -- NEW
+  serp_num_ads_notshowing AS num_ads_notshowing, -- NEW
+  NULL AS unknown,
+  NULL AS is_sap_monetizable, -- REVISIT
```
irrationalagent: Likely will revisit / remove. Talking to sydney, the UDF that feeds this is stale, and the country-based exclusions that matter are already encoded downstream in the pipeline - we will mirror those.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,694 @@
...
+          JSON_VALUE(event_extra.provider_id)
+        )
+    END AS normalized_engine, -- this is "engine" in v8
+    CASE
+      WHEN JSON_VALUE(event_extra.partner_code) = ''
+        THEN NULL
+      ELSE JSON_VALUE(event_extra.partner_code)
+    END AS partner_code,
+    CASE
+      WHEN JSON_VALUE(event_extra.source) = 'urlbar-handoff'
```
irrationalagent: I think there is some additional snake case normalization that has to happen here, checking the data i see both versions of each below:


canonical | variant
-- | --
urlbar_handoff | urlbar-handoff
urlbar_persisted | urlbar-persisted
urlbar_searchmode | urlbar-searchmode
about_home | abouthome
about_newtab | newtab

its small, but worth capturing. 

possibly: REPLACE(source, '-', '_') plus explicit mappings for abouthome → about_home and newtab → about_newtab.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+    %% SAP
+    subgraph SG1["SAP: Customer Enrichment Pipeline"]
+    base_sap("sap_events_with_client_info<br/>---<br/>latest record by:<br/>client_id<br/>submission_date<br/>normalized_engine<br/>partner_code<br/>source")
+
+    sap_enterprise("sap_is_enterprise<br/>---<br/>group by:<br/>client_id<br/>submission_date")
+
+    sap_full_events("sap_events_clients_adblocker_enterprise")
+
+    sap_agg("sap_aggregates<br/>---<br/>group by:<br/>client_id<br/>submission_date<br/>normalized_engine<br/>partner_code<br/>source")
```
irrationalagent, line 22: maybe this is forthcoming, but I don't see partner code in any of the grouping statements

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/README.md
```diff
@@ -0,0 +1,173 @@
...
+    serp_full_events -->|"serp events"| serp_final
+
+    serp_agg -->|"left join using (client_id, submission_date, serp_provider_id, partner_code, serp_search_access_point)"| serp_final
+
+    serp_ad_click -->|"left join using (client_id, submission_date, serp_provider_id, partner_code, serp_search_access_point)"| serp_final
+    end
+
+    %% FINALS
+
+    join_sap_serp("join_sap_serp<br/>---<br/>full outer join on:<br/>client_id<br/>submission_date<br/>engine<br/>partner_code<br/>source")
```
irrationalagent: Related to my other comment on this join, SQL has left join not full outer join. I think it SHOULD probably be full outer join.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/README.md
irrationalagent: this occurs in multiple places.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/ctes/13_join_sap_serp.sql
irrationalagent: a bit more detail: because 14_final.sql reads date and client_id from the SERP side, SAP-only rows don't just lose their SERP columns, they're dropped entirely. Every Baidu SAP disappears (64k events/week) because its not instrumented for SERP. And 14_final.sql:6-11 has a sap_normalized_engine fallback that's unreachable through the left join

irrationalagent (changes requested):

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/ctes/14_final.sql
```diff
@@ -0,0 +1,96 @@
...
+  NULL AS scalar_parent_urlbar_searchmode_tabmenu_sum,
+  NULL AS scalar_parent_urlbar_searchmode_tabtosearch_sum,
+  NULL AS scalar_parent_urlbar_searchmode_tabtosearch_onboard_sum,
+  NULL AS scalar_parent_urlbar_searchmode_topsites_newtab_sum,
+  NULL AS scalar_parent_urlbar_searchmode_topsites_urlbar_sum,
+  NULL AS scalar_parent_urlbar_searchmode_touchbar_sum,
+  NULL AS scalar_parent_urlbar_searchmode_typed_sum,
+  serp_profile_age_in_days AS profile_age_in_days,
+  serp_searches_organic_count AS organic,
+  serp_searches_tagged_count AS tagged_sap,
```
irrationalagent: I think we should consider just removing this. `tagged_sap` is misleading, not just here for glean (where it is SERP) but actually also for legacy: the legacy prob actually ALSO counts page loads for 99% of what we currently call tagged_sap

(legacy tagged_sap comes from browser.search.content.<ap> keyed scalar, key <provider>:tagged:<code> )

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,694 @@
...
+    serp_profile_age_in_days AS profile_age_in_days,
+    serp_searches_organic_count AS organic,
+    serp_searches_tagged_count AS tagged_sap,
+    serp_searches_tagged_count AS tagged_serp,
+    serp_follow_on_searches_tagged_count AS tagged_follow_on,
+    sap,
+    serp_counts,
+    serp_ad_click_target AS ad_click_target,
+    serp_num_ad_clicks AS ad_click,
+    serp_ad_clicks_organic_count AS ad_click_organic,
```
YeonjooSmith: * We aligned to keep ad_clicks_tagged, ad_clicks_organic, ad_clicks_total

* no more `ad_clicks` only which was the v8 naming convention. 

**What v8 does.** Both columns come from one expression (v8 `query.sql:212`) and are disjoint:

```sql
CONCAT('ad-click', IF(REGEXP_CONTAINS(key, ':organic'), ':organic', '')) AS type
```

The legacy key is `<provider>:<type>`, where `type` is one of `organic`, `tagged`, `tagged-follow-on` (`SearchSERPTelemetry.sys.mjs:1728-1730`; `Scalars.yaml` documents only two of the three, the code is authoritative). So v8's `ad_click` is **tagged plus tagged-follow-on**, and the two columns partition the total exactly.

**What v9 does here.** `ad_click` is `SUM(num_ad_clicks)` over all rows regardless of `is_tagged` (line 451), so it contains `ad_click_organic`. The v8 equivalent is computed two lines above, on 449, and currently ships as `ad_clicks_tagged` on line 673.

The two sides do agree on where the boundary sits: `is_tagged` is `info.type.startsWith("tagged")` (`SearchSERPTelemetry.sys.mjs:1509` -> `:2136` -> `serp_events_v2/query.sql:100`), covering the same two types v8 puts in `ad_click`. So `serp_ad_clicks_tagged_count` is an exact analogue, with no grain change involved.

**Suggested change:**

```suggestion
    serp_num_ad_clicks AS ad_click_total,
    serp_ad_clicks_tagged_count AS ad_click_tagged,
    serp_ad_clicks_organic_count AS ad_click_organic,
```

and then line 673 becomes `serp_num_ad_clicks AS ad_click_total`, so the total is still published, under a name that says what it is.

`search_with_ads` / `search_with_ads_organic` on 671-672 already follow this convention, so this would also make the two ad pairs consistent with each other.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,694 @@
...
+    NULL AS scalar_parent_urlbar_searchmode_topsites_newtab_sum,
+    NULL AS scalar_parent_urlbar_searchmode_topsites_urlbar_sum,
+    NULL AS scalar_parent_urlbar_searchmode_touchbar_sum,
+    NULL AS scalar_parent_urlbar_searchmode_typed_sum,
+    serp_profile_age_in_days AS profile_age_in_days,
+    serp_searches_organic_count AS organic,
+    serp_searches_tagged_count AS tagged_sap,
+    serp_searches_tagged_count AS tagged_serp,
+    serp_follow_on_searches_tagged_count AS tagged_follow_on,
+    sap,
```
YeonjooSmith: To output 0 instead of NULL, 
```suggestion
COALESCE(sap, 0) AS sap,
```

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,694 @@
...
+    serp_final_cte.serp_scalar_parent_browser_engagement_total_uri_count_sum AS serp_scalar_parent_browser_engagement_total_uri_count_sum,
+    serp_final_cte.max_concurrent_tab_count_max AS serp_max_concurrent_tab_count_max
+  FROM
+    serp_final_cte
+  LEFT JOIN
+    sap_final_cte
+    ON sap_final_cte.client_id IS NOT DISTINCT FROM serp_final_cte.client_id
+    AND sap_final_cte.submission_date IS NOT DISTINCT FROM serp_final_cte.submission_date
+    AND sap_final_cte.normalized_engine IS NOT DISTINCT FROM serp_final_cte.serp_provider_id
+    AND sap_final_cte.source IS NOT DISTINCT FROM serp_final_cte.serp_search_access_point
```
YeonjooSmith: +1 to `FULL OUTER JOIN` per @irrationalagent in https://github.com/mozilla/bigquery-etl/pull/7744#discussion_r3806137587, and the README already documents it that way.

One thing to pair with it: in `final_cte`, every dimension comes from the SERP side, with `normalized_engine` the only exception. So SAP-only rows would come through with NULL `submission_date`, `client_id`, `source`, `country` and `sample_id`.

`submission_date` is the blocker: `metadata.yaml` partitions on it with `require_partition_filter: true`, so those rows would land in the `__NULL__` partition and be unreachable.

So FULL OUTER plus `COALESCE(serp_x, sap_x)` on the join keys and shared dimensions, ideally in one commit.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/ctes/13_join_sap_serp.sql
kbammarito: this has been reverted to the coalesce and the full outer join

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/README.md
kbammarito: this has been reverted to the coalesce and the full outer join

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/ctes/14_final.sql
kbammarito: removed

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: done

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
YeonjooSmith: Agreed on the fix. Worth flagging it matters more than the label suggests: `source` is one of the four join keys.

With `FULL OUTER`, a pre-142 client emitting `urlbar-searchmode` will not match the SERP row's `urlbar_searchmode`, so the same access point comes out as two rows: one carrying the SERP metrics with `sap` NULL, one carrying the `sap` count with the SERP metrics NULL. Nothing is lost, but the thing the join exists to do, line SAP and SERP up for a given client and access point, does not happen for those rows.

Also worth noting the five pairs are the complete set rather than just what showed up in the data. [`KNOWN_SEARCH_SOURCES`](https://github.com/mozilla-firefox/firefox/blob/91c8ca3faa6ccbb72d65d89401fd31fd3313afc4/browser/components/search/BrowserSearchTelemetry.sys.mjs#L58-L72) maps canonical name to emitted string, and exactly five entries differ: the three `urlbar-*` plus `abouthome` and `newtab`. Every other access point spells the same on both sides and cannot mismatch, so your suggested fix covers all of it.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/schema.yaml
```diff
@@ -0,0 +1,311 @@
...
+- name: partner_code
+  type: STRING
+  mode: NULLABLE
+- name: country
+  type: STRING
+  mode: NULLABLE
+  description: Name of the country in which the activity took place, as determined
+    by the IP geolocation.
+- name: addon_version
+  type: INTEGER
```
irrationalagent: in V8 this is string, probably best to keep it as such.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/schema.yaml
```diff
@@ -0,0 +1,311 @@
...
+  description: The distribution id associated with the install of Firefox.
+- name: locale
+  type: STRING
+  mode: NULLABLE
+  description: Set of language- and/or country-based preferences for a user interface.
+- name: user_pref_browser_search_region
+  type: STRING
+  mode: NULLABLE
+- name: search_cohort
+  type: INTEGER
```
irrationalagent: also a string in V8, recommend changing

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,694 @@
...
+      WHEN partner_code = ''
+        THEN NULL
+      ELSE partner_code
+    END AS partner_code,
+    sap_source AS serp_search_access_point,
+    sample_id,
+    profile_group_id,
+    legacy_telemetry_client_id,
+    normalized_country_code AS country,
+    normalized_app_name AS app_version,
```
YeonjooSmith: `normalized_app_name` is the app name, not the version, which is why this column comes out empty.

`serp_events_v2` already computes the version at its line 84, `mozfun.norm.browser_version_info(client_info.app_display_version)`, and `.version` is the raw string that v8's `app_version` held.

```suggestion
    browser_version_info.version AS app_version,
```

Same substitution applies at line 106 on the SAP side, where the source is `events_stream_v1`, so it would be `client_info.app_display_version AS app_version`. That copy is discarded today but goes live if the join becomes `FULL OUTER`.

**Worth exposing `major_version` alongside it.** Comparing versions as strings is a quiet trap, `'9.0' > '142.0'` lexicographically, and the Firefox 138 cutoff for `sap.counts` means any like-for-like `sap` comparison has to filter on major version. The struct already carries it as NUMERIC:

```sql
    browser_version_info.major_version AS app_major_version,
```

`serp_events_clients_ad_enterprise_cte` and `serp_final_cte` both pass through with `.*`, so that needs only two more lines: `serp_final_cte.app_major_version AS serp_app_major_version` in `join_sap_serp_cte`, and `serp_app_major_version AS app_major_version` in `final_cte`.

It would also match how `os_version` is already handled here, with `os_version_major` and `os_version_minor` derived alongside the raw string.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,694 @@
...
+    serp_default_search_engine_overridden AS default_search_engine_overridden, -- NEW
+    serp_default_private_search_engine_display_name AS default_private_search_engine,
+    serp_default_private_search_engine_load_path AS default_private_search_engine_data_load_path,
+    serp_default_private_search_engine_submission_url AS default_private_search_engine_data_submission_url,
+    serp_default_private_search_engine_partner_code AS default_private_search_engine_partner_code, -- NEW
+    serp_default_private_search_engine_provider_id AS default_private_search_engine_provider_id, -- NEW
+    serp_default_private_search_engine_overridden AS default_private_search_engine_overridden, -- NEW
+    serp_sample_id AS sample_id,
+    serp_subsession_start_time AS subsession_start_time, -- NEW
+    serp_subsession_end_time AS subsession_end_time, -- NEW
```
YeonjooSmith: ln 636 - 638: the column names incorrectly describe what they actually are. 
subsession_start_time (636) holds ping_info.start_time, when this ping's collection window opened. Proposed: ping_start_time
subsession_end_time (637) holds ping_info.end_time, when it closed. Proposed: ping_end_time
subsession_counter (638) holds ping_info.seq, a per-client lifetime count of how many pings of this type have been sent, starting at 0. Proposed: ping_seq

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,801 @@
...
+    default_private_search_engine_display_name AS default_private_search_engine,
+    default_private_search_engine_load_path AS default_private_search_engine_data_load_path,
+    default_private_search_engine_submission_url AS default_private_search_engine_data_submission_url,
+    default_private_search_engine_partner_code, -- NEW
+    default_private_search_engine_provider_id, -- NEW
+    default_private_search_engine_overridden, -- NEW
+    sample_id,
+    subsession_start_time, -- NEW
+    subsession_end_time, -- NEW
+    subsession_counter, -- NEW
```
YeonjooSmith: The column names incorrectly describe what they actually are. `final_cte:761-763`, output names for the values aliased in `join_sap_serp_cte:625`, `:629` and `:633`.

- `subsession_start_time` holds `ping_info.start_time`, when this ping's collection window opened. Proposed: `ping_start_time`
- `subsession_end_time` holds `ping_info.end_time`, when it closed. Proposed: `ping_end_time`
- `subsession_counter` holds `ping_info.seq`, a per-client lifetime count of how many pings of this type have been sent, starting at 0. Proposed: `ping_seq`

These are not v8's subsession fields renamed. Legacy subsessions describe browser sessions, where `subsession_counter` resets per session. Glean's `ping_info` describes ping collection windows. A v8 query referencing `subsession_counter` will keep running against v9 and return plausible numbers that mean something else.

Renaming here changes the output schema without touching upstream:

```suggestion
    ping_start_time, -- NEW
    ping_end_time, -- NEW
    ping_seq, -- NEW
```

That needs the matching aliases updated at `join_sap_serp_cte:625`, `:629` and `:633`. Alternatively alias in place here, `subsession_start_time AS ping_start_time` and so on, if you would rather keep the upstream names.

All three are marked NEW, so nothing depends on them yet and this is free now, a migration after cutover.

Worth noting the aliasing originates in [`serp_events_v2:103-105`](https://github.com/mozilla/bigquery-etl/blob/fb1a9386bb95b68ee0a76c11ef89aeb01d2e2262/sql/moz-fx-data-shared-prod/firefox_desktop_derived/serp_events_v2/query.sql#L103-L105), not in this PR. Renaming locally leaves v9 and `serp_events_v2` disagreeing; fixing at source is more correct but has other consumers. Happy either way, just flagging it is a choice.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,801 @@
...
+            ) AS type,
+            REPLACE(
+              TO_JSON_STRING(experiments[JSON_KEYS(experiments)[OFFSET(0)]].extra.enrollment_id),
+              '"',
+              ''
+            ) AS enrollment_id
+          ) AS extra
+        ) AS value
+      )
+    ] AS experiments
```
YeonjooSmith: The SERP side is a clean passthrough of `serp_events_v2.experiments`, which is already `ARRAY<STRUCT<key, value<branch, extra<type, enrollment_id>>>>`. This block reconstructs that shape from `events_stream_v1.experiments`, which is a `JSON` column. The reconstruction keeps one experiment.

The outer `[ ... ]` is an array literal with `JSON_KEYS(experiments)[OFFSET(0)]` as its only key, so a client in three experiments contributes one. `JSON_KEYS` sorts lexicographically, so it is always the alphabetically first slug, which means the same experiments are dropped every time rather than a random one.

Two related things I checked while looking at this:

**Absent sub-fields come through as the string `null`.** `TO_JSON_STRING` renders an absent or null JSON value as the four-character text `null`, not SQL NULL. So an experiment with no `extra.type` gets `type = 'null'`, and `WHERE type IS NULL` finds nothing. `JSON_VALUE` returns proper NULL and needs no quote stripping.

**`[OFFSET(0)]` errors on an empty JSON object.** `JSON_KEYS(JSON '{}')[OFFSET(0)]` fails with `Array index 0 is out of bounds`. On NULL JSON it safely returns NULL, so this only holds while `experiments` is NULL rather than `{}` for clients in no experiments.

Suggested rewrite:

```suggestion
    ARRAY(
      SELECT AS STRUCT
        k AS key,
        STRUCT(
          JSON_VALUE(experiments[k].branch) AS branch,
          STRUCT(
            JSON_VALUE(experiments[k].extra.type) AS type,
            JSON_VALUE(experiments[k].extra.enrollment_id) AS enrollment_id
          ) AS extra
        ) AS value
      FROM UNNEST(JSON_KEYS(experiments, 1)) AS k
    ) AS experiments
```

The depth argument matters: `JSON_KEYS` recurses by default and returned seven keys for two experiments, including `exp-b.extra.type`.

I ran this against a two-experiment fixture with the SERP side NULL, as it would be on a SAP-only row:

```json
[{"key":"exp-a","value":{"branch":"control","extra":{"type":null,"enrollment_id":null}}},
 {"key":"exp-b","value":{"branch":"treatment","extra":{"type":"rollout","enrollment_id":"e1"}}}]
```

The `COALESCE` at line 634 compiles against it, so the struct names, nesting and field order match. Empty or NULL `experiments` gives a zero-length array with no error.

Since that `COALESCE` prefers the SERP side, this only surfaces on SAP-only rows.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
```diff
@@ -0,0 +1,801 @@
...
+    MAX(UNIX_DATE(DATE(to_utc_string(subsession_start_time)))) - MAX(
+      UNIX_DATE(DATE(safe_parse_timestamp(first_run_date)))
+    ) AS profile_age_in_days,
+    COUNT(*) AS serp_counts,
+    COUNTIF(subsession_counter = 1) AS sessions_started_on_this_day,
+    SUM(browser_engagement_active_ticks / (3600 / 5)) AS active_hours_sum,
+    SUM(
+      browser_engagement_tab_open_event_count
+    ) AS serp_scalar_parent_browser_engagement_tab_open_event_count_sum,
+    SUM(browser_engagement_uri_count) AS serp_scalar_parent_browser_engagement_total_uri_count_sum,
```
YeonjooSmith: These three are ping-scoped counters, so summing them over event rows multiplies them. `max_concurrent_tab_count_max` on the next line uses `MAX` and is fine.

I checked this on 2026-08-18, `sample_id = 0`. The counter value is constant within a ping on both branches, `pings_where_value_varies = 0` across 734,806 and 870,124 pings, which confirms it is a ping-level value repeated on every row from that ping rather than something per-event:

| branch | rows | pings | rows per ping | sum over rows | sum one per ping | inflation |
| --- | --- | --- | --- | --- | --- | --- |
| SAP, `events_stream_v1` | 1,264,971 | 734,806 | 1.72 | 192,270,982 | 84,588,233 | **2.27x** |
| SERP, `firefox_desktop.serp_events` | 2,602,515 | 870,124 | 2.99 | 379,157,133 | 94,563,944 | **4.01x** |

Worth noting the inflation is larger than the rows-per-ping ratio on both sides, 2.27 against 1.72 and 4.01 against 2.99. It is volume-weighted: pings with more search activity also carry higher `active_ticks`, so the heaviest rows get multiplied most. Estimating from the row ratio alone understates it by about a third.

Both paths reach the output, since these are now COALESCEd across the branches at lines 675-690 rather than taken from SERP alone. So a row carries the ~4x figure when a SERP row exists and the ~2.3x figure otherwise, which means the bias is not a constant that could be corrected downstream.

Against v8 these are not comparable at any level, since `clients_daily` holds one value per client-day.

**Two separate problems.** Deduplicating to one row per ping before summing fixes the multiplication. It does not fix the grain: these are client-day measures being computed inside a client / day / engine / access-point group, so a client who searched on two engines still has their day counted twice.

That second half might argue for dropping them rather than repairing them, on the same reasoning as the v8-compatibility columns already removed. They are client-activity measures that rode along because v8's base was `clients_daily`, and the pipeline is already planning to join a DAU source downstream where they would sit naturally. Happy either way, but worth deciding before fixing, since a correctly-deduplicated but still engine-split number would look right and not be.

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/ctes/14_final.sql
```diff
@@ -0,0 +1,70 @@
...
+  scalar_parent_browser_engagement_tab_open_event_count_sum AS tab_open_event_count_sum,
+  active_hours_sum,
+  scalar_parent_browser_engagement_total_uri_count_sum AS total_uri_count,
+  experiments,
+  profile_age_in_days,
+  serp_searches_organic_count AS organic,
+  serp_searches_tagged_count AS tagged_serp,
+  serp_follow_on_searches_tagged_count AS tagged_follow_on,
+  sap,
+  serp_counts,
```
Standard8: I realise we're trying to get to the end here, and we're also trying to make this similar to v8, however, the reason I originally proposed the changes to the naming to the fuller versions, was to make it abundantly clear what they referred to.

For example in the v8 table, we have "sap" and "tagged_sap", however "tagged_sap" is actually relating to tagged SERP, and it confuses me every time I see it. Though the naming is slightly better here.

If we don't want to address this now, so that we can get a summary table for Glean quicker, I'd accept them as they are, but I would ask that we consider it as an enhancement for the next ETL.

Standard8: What's `serp_counts`? Aren't those counts included in the other columns (tagged serp, tagged follow on, organic)?

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/ctes/14_final.sql
```diff
@@ -0,0 +1,70 @@
...
+  serp_ad_click_target AS ad_click_target,
+  serp_num_ad_clicks AS ad_click,
+  serp_ad_clicks_organic_count AS ad_click_organic,
+  serp_with_ads_tagged_count AS search_with_ads,
+  serp_with_ads_organic_count AS search_with_ads_organic,
```
Standard8: I just realised - in theory we don't need both tagged and organic columns. A tagged SERP is one that has a partner code. An organic one, is one without the partner_code. Since we have the `partner_code` in this table (which within a particular row should be the partner code used for the SERP, if any), then we shouldn't need the extra columns.

To put it another way, if `partner_code` is non-null for a particular row, then all the `organic` columns should be 0. If `partner_code` is null, then all the non-organic (tagged) columns should be 0.

Again, I'm happy if we don't want to address it in the interests of time - it'll just mean that we have more columns than necessary.

Standard8 (commented):
It is great to see all the updates and work progressing here. There were a couple of bits I was wondering about. I don't want to slow the work down though, as I realise we do need to get this out soon. So take these as a something we could consider later (or now if it helps to).

irrationalagent (changes requested):

irrationalagent (changes requested):
just a couple of minor points on schema consistency

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/schema.yaml
kbammarito: no longer exists in our dropping of the nulls

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/schema.yaml
kbammarito: no longer exists in our dropping of the nulls

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: done

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/ctes/14_final.sql
kbammarito: We've dropped the tagged_sap from v9 (that was identical to tagged_serp) and just left tagged_serp

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: @kbammarito to also add minor version and patch

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/ctes/14_final.sql
irrationalagent: Hi, yes agreed with your logic here in principle. This is here really for ergonomics - downstream users of these tables in DS have this tagged/organic distinction in their reporting and in their mental models, so making the distinction explicit and having separate metrics is in deference to this precedent (and to discoverability)

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: dropped

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: done

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: I kept normalized_app_name and added the version and its components

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: done and preferred whichever side has enrollments in the coalesce

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/README.md
kbammarito: re-added partner_code into the groupings

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: done and checking on grouping

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/query.sql
kbammarito: done and also updated search_with_ads to be search_with_ads_tagged

sql/moz-fx-data-shared-prod/search_derived/search_clients_daily_v9/ctes/14_final.sql
kbammarito: @Standard8 we've kept it as a total without other conditions (renamed for clarity):

```
    COUNT(*) AS serp_counts_total,
    ...
  FROM
    `mozdata.firefox_desktop.serp_events` -- serp_events_v2 doesn't have the aggregated fields like `num_ads_visible`
  WHERE
  GROUP BY
    client_id,
    submission_date,
    serp_provider_id,
    partner_code,
    serp_search_access_point
```

PR #7746, opened 2025-07-08
feat(DENG-8914): update mobile_view templates inside active_users_aggregates generators so that focus products use composite dataset

kik-kik:
# feat(DENG-8914): update mobile_view templates inside active_users_aggregates generators so that focus products use composite dataset

This is related to changes to focus product telemetry data collection changes. This change should enable us to continue measuring the usage of the focus products.

### DIFF
--- a/bigquery_etl/metadata/validate_metadata.py
+++ b/bigquery_etl/metadata/validate_metadata.py
@@ -428,10 +428,13 @@ def validate(target):
                     if not validate_public_data(metadata, path):
                         failed = True
 
+                    _, project_id, _ = os.path.dirname(root).split("/")
+
                     if not validate_change_control(
                         file_path=root,
                         metadata=metadata,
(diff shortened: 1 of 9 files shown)


irrationalagent:
[so the data looks good,](https://sql.telemetry.mozilla.org/queries/109194/source#268176) there’s one thing i think we still have to fix - the `app_name`. From the composite data source they are `focus_ios` and `focus_android`. from the old AUA datasource they are `Focus iOS` and `Focus Android`.

We definitely want to use the latter names.

I’m guessing the best way to fix that would be to fix the upstream tables, but it might be ok (for now) to add a
```
CASE WHEN app_name = 'focus_ios' THEN 'Focus iOS' END AS app_name
```

or something similar to the relevant parts of the mobile view

irrationalagent (approved):

PR #7757, opened 2025-07-09
DENG-9093 Pull Dashboard Service data from BigEye API and upload to BigQuery

phil-lee70:
## Description

This PR creates a feature to get data from the BigEye API [dashboardService](https://docs.bigeye.com/reference/getdashboarddataseries) and uploads it to BigQuery.

**Sample Query**
[LINK](https://console.cloud.google.com/bigquery?ws=!1m7!1m6!12m5!1m3!1smozdata!2sus-central1!3s9d7cd6f5-29b7-497d-a178-9880745d72c7!2e1)
```
select *
from `moz-fx-data-shared-prod.bigeye_derived.dashboard_service`
;
```

**Output**
Screenshot 2025-07-09 at 2 17 38 PM

* DENG-9093
Story: * DSRE-4576


sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
```diff
@@ -0,0 +1,169 @@
...
+from google.api_core.exceptions import GoogleAPIError
+from google.cloud import bigquery
+from requests.exceptions import RequestException
+
+# Configuration
+TARGET_PROJECT = os.getenv("TARGET_PROJECT", "moz-fx-data-shared-prod")
+TARGET_TABLE = os.getenv(
+    "TARGET_TABLE", "moz-fx-data-shared-prod.bigeye_derived.dashboard_service"
+)
+BIGEYE_API_KEY = os.environ["BIGEYE_API_KEY"]
```
kik-kik, line 14: This needs to be set in the metadata.yaml file. Here's an example of this:
https://github.com/mozilla/private-bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/cinder_external/jobs_raw_v1/metadata.yaml#L15-L17

Basically under scheduling we should have something like this:
```
scheduling:
  dag_name: bqetl_bigeye_derived
  secrets:
  - deploy_target: BIGEYE_API_KEY
    key: bqetl_bigeye_api_key # need to check if this secret is already defined in GSM, if not we need to create it.
```

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/metadata.yaml
```diff
@@ -0,0 +1,17 @@
...
+  BigEye metadata for the dashboard service
+owners:
+- phlee@mozilla.com
+labels:
+  application: bigeye
+  incremental: false
+  schedule: daily
+  table_type: aggregate
+scheduling:
+  dag_name: bqetl_bigeye_derived
```
kik-kik, line 12: This DAG needs to be defined inside [dags.yaml](https://github.com/mozilla/bigquery-etl/blob/main/dags.yaml).

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
```diff
@@ -0,0 +1,169 @@
...
+        # Get and process data
+        df = get_bigeye_data()
+
+        # Validate data
+        validate_dataframe(df)
+
+        # Load to BigQuery
+        load_to_bigquery(df)
+
+        elapsed = time.time() - start_time
```
kik-kik: if we keep the try I'd probably move this outside of it.

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
```diff
@@ -0,0 +1,169 @@
...
+        raise Exception(f"BigQuery load failed: {str(e)}")
+
+
+def main() -> None:
+    """Pull from BigEye API dashboard service endpoint then upload to BigQuery."""
+    start_time = time.time()
+    print("Starting data pipeline...")
+
+    try:
+        # Get and process data
```
kik-kik:
```suggestion
```

Personally I'd suggest avoid comments that do not provide useful information. I already know what is happening by looking at the function call. This comment doesn't really give me anything useful to work with and just adds extra text to parse when looking through the code.

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
```diff
@@ -0,0 +1,169 @@
...
+    try:
+        job = client.load_table_from_dataframe(
+            df,
+            TARGET_TABLE,
+            job_config=bigquery.LoadJobConfig(write_disposition="WRITE_TRUNCATE"),
+        )
+        job.result()
+        print(f"Successfully loaded data to {TARGET_TABLE}")
+    except GoogleAPIError as e:
+        raise Exception(f"BigQuery load failed: {str(e)}")
```
kik-kik: Is this error better in any way than what we'd get if we didn't catch it here and reformat it? If the answer is no, then I'd suggest we simplify the code and remove the try, except here.

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
```diff
@@ -0,0 +1,169 @@
...
+
+# Libraries specific for pulling from API and uploading to bigquery.
+import pandas as pd
+import requests
+from google.api_core.exceptions import GoogleAPIError
+from google.cloud import bigquery
+from requests.exceptions import RequestException
+
+# Configuration
+TARGET_PROJECT = os.getenv("TARGET_PROJECT", "moz-fx-data-shared-prod")
```
kik-kik: I'd suggest we set these up as script arguments to stay consistent with how we do it in other scripts. For example: https://github.com/mozilla/private-bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/cinder_external/jobs_raw_v1/query.py#L95-L108

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
```diff
@@ -0,0 +1,169 @@
...
+from typing import Any, Dict
+
+# Libraries specific for pulling from API and uploading to bigquery.
+import pandas as pd
+import requests
+from google.api_core.exceptions import GoogleAPIError
+from google.cloud import bigquery
+from requests.exceptions import RequestException
+
+# Configuration
```
kik-kik:
```suggestion
```

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
```diff
@@ -0,0 +1,169 @@
+"""Pull from BigEye API from dashboardService and upload to Big Query."""
+
+import json
+import os
+import time
+from typing import Any, Dict
+
+# Libraries specific for pulling from API and uploading to bigquery.
```
kik-kik:
```suggestion
```

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
```diff
@@ -0,0 +1,169 @@
...
+# Configuration
+TARGET_PROJECT = os.getenv("TARGET_PROJECT", "moz-fx-data-shared-prod")
+TARGET_TABLE = os.getenv(
+    "TARGET_TABLE", "moz-fx-data-shared-prod.bigeye_derived.dashboard_service"
+)
+BIGEYE_API_KEY = os.environ["BIGEYE_API_KEY"]
+WORKSPACE_IDS = json.loads(os.getenv("WORKSPACE_IDS", "[463, 508, 509]"))
+API_URL = "https://app.bigeye.com/api/v1/dashboard/calculate-data-points-request"
+
+# Data point types to request
```
kik-kik:
```suggestion
```

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
```diff
@@ -0,0 +1,169 @@
...
+from google.cloud import bigquery
+from requests.exceptions import RequestException
+
+# Configuration
+TARGET_PROJECT = os.getenv("TARGET_PROJECT", "moz-fx-data-shared-prod")
+TARGET_TABLE = os.getenv(
+    "TARGET_TABLE", "moz-fx-data-shared-prod.bigeye_derived.dashboard_service"
+)
+BIGEYE_API_KEY = os.environ["BIGEYE_API_KEY"]
+WORKSPACE_IDS = json.loads(os.getenv("WORKSPACE_IDS", "[463, 508, 509]"))
```
kik-kik: Maybe we should check to make sure the value we get is always a list.

kik-kik: I'd also consider moving this into some sort of config file. We could pass this list as an argument into the script and define the values in metadata.yaml or alternatively if we expect this list to be reused across different bigeye imports then perhaps we could utilize [bqetl_project.yaml](https://github.com/mozilla/bigquery-etl/blob/main/bqetl_project.yaml) for this.

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/metadata.yaml
```diff
@@ -0,0 +1,17 @@
+friendly_name: Dashboard Service
+description: |-
+  BigEye metadata for the dashboard service
```
kik-kik: Could you include a link to the API docs for this specific endpoint here?

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
```diff
@@ -0,0 +1,169 @@
...
+        "accept": "application/json",
+        "content-type": "application/json",
+        "authorization": BIGEYE_API_KEY,
+    }
+
+    try:
+        response = requests.post(API_URL, json=payload, headers=headers, timeout=30)
+        response.raise_for_status()
+        return response.json()
+    except RequestException as e:
```
kik-kik: Does this exception handling add anything of value?

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
```diff
@@ -0,0 +1,169 @@
...
+    try:
+        df = pd.json_normalize(
+            response_data["dataPointSeries"],
+            record_path="dataPoints",
+            meta=["dataPointType"],
+            errors="ignore",
+        )
+        df["workspaceIds"] = df["workspaceIds"].str[0]  # Take first element
+
+        # Drop sourceIds if it exists
```
kik-kik:
```suggestion
```

I'd avoid explaining what my code does, it'd include a comment if there is some nuance to why we have to do something.

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
```diff
@@ -0,0 +1,169 @@
...
+def process_response(response_data: Dict[str, Any]) -> pd.DataFrame:
+    """Process API response into a DataFrame."""
+    try:
+        df = pd.json_normalize(
+            response_data["dataPointSeries"],
+            record_path="dataPoints",
+            meta=["dataPointType"],
+            errors="ignore",
+        )
+        df["workspaceIds"] = df["workspaceIds"].str[0]  # Take first element
```
kik-kik:
```suggestion
        df["workspaceIds"] = df["workspaceIds"].str[0]
```

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
```diff
@@ -0,0 +1,169 @@
...
+                "workspaceIds": "workspace_id",
+                "dataTimestamp": "timestamp",
+            }
+        )
+
+        # Ensure your DataFrame column is datetime type
+        df["timestamp"] = pd.to_datetime(df["timestamp"], unit="s")
+
+        return df
+    except KeyError as e:
```
kik-kik: Unless we intend to do further actions here, I'm not sure this exception handling adds much value.

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
```diff
@@ -0,0 +1,169 @@
...
+
+        # Validate data
+        validate_dataframe(df)
+
+        # Load to BigQuery
+        load_to_bigquery(df)
+
+        elapsed = time.time() - start_time
+        print(f"Pipeline completed successfully in {elapsed:.2f} seconds")
+    except Exception as e:
```
kik-kik: I'm not sure we need to wrap everything here in a try, except. If it fails it's fine, the task will be marked as failed. Looking at the error message here, I'm not sure it will tell us anything extra to make it worth including.

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/schema.yaml
```diff
@@ -0,0 +1,21 @@
+fields:
+- description: |
+    Workspace ID associated with this record.
+  mode: NULLABLE
+  name: workspace_id
+  type: INTEGER
+- description: |
+    Date timestamp associated with this record.
```
kik-kik:
```suggestion
    Timestamp associated with this record.
```

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/schema.yaml
```diff
@@ -0,0 +1,21 @@
...
+  mode: NULLABLE
+  name: timestamp
+  type: DATETIME
```
kik-kik: Is this meant to be DATETIME or TIMESTAMP?

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
```diff
@@ -0,0 +1,169 @@
...
+def get_bigeye_data() -> pd.DataFrame:
+    """Fetch data from Bigeye API for all workspaces and combine into a single DataFrame."""
+    all_data = []
+
+    for workspace_id in WORKSPACE_IDS:
+        try:
+            response_data = make_api_request(workspace_id)
+            df = process_response(response_data)
+            all_data.append(df)
+            print(f"Successfully processed data for workspace {workspace_id}")
```
kik-kik: Could we use the logging library instead of print statements?

An example of logging being used in query.py: https://github.com/mozilla/bigquery-etl/blob/d8d41d8a7e34d12fc375d0d262ead4e08e6336ee/sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_choice_screen_selection_v1/query.py#L201

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/schema.yaml
phil-lee70: DATETIME

kik-kik: In this case, could we rename the field to reflect this? The field being called timestamp makes it confusing.

phil-lee70: I have actually changed it back to simply TIMESTAMP to be less confusing

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
```diff
@@ -0,0 +1,149 @@
...
+        raise ValueError(f"DataFrame is missing required columns: {missing}")
+
+
+def load_to_bigquery(project_id, df: pd.DataFrame) -> None:
+    """Load DataFrame to BigQuery."""
+    client = bigquery.Client(project_id)
+
+    job = client.load_table_from_dataframe(
+        df,
+        TARGET_TABLE,
```
kik-kik, line 129: I'd suggest we make targaet_table a param in this function and pass it in as an argument to make it more flexible in case we want to write to a different target table.

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
```diff
@@ -0,0 +1,149 @@
...
+        "workspaceIds": [workspace_id],
+    }
+
+    headers = {
+        "accept": "application/json",
+        "content-type": "application/json",
+        "authorization": BIGEYE_API_KEY,
+    }
+
+    response = requests.post(API_URL, json=payload, headers=headers, timeout=30)
```
kik-kik, line 56: Do we need to make a post to just get data back here?

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
```diff
@@ -0,0 +1,149 @@
...
+        job_config=bigquery.LoadJobConfig(write_disposition="WRITE_TRUNCATE"),
+    )
+    job.result()
+    logging.info(f"Successfully loaded data to {TARGET_TABLE}")
+
+
+def main() -> None:
+    """Pull from BigEye API dashboard service endpoint then upload to BigQuery."""
+    parser = ArgumentParser(description=__doc__)
+    parser.add_argument("--project", default="moz-fx-data-shared-prod")
```
kik-kik, line 139: I would suggest dataset and maybe the table name as optional arguments into the script so that we could easily run the script locally and write to a temporary table. For example, if we need to debug an issue and do not want to write to prod table.

sql/moz-fx-data-shared-prod/bigeye_derived/dashboard_service_v1/query.py
phil-lee70: yes! [docs](https://docs.bigeye.com/reference/getdashboarddataseries)

PR #7770, opened 2025-07-11
Deng 9088 deprecate firefox reality tables

kwindau:
## Description

This PR deprecates the following Glean generated tables for apps: `firefox_reality` and `firefox_reality_pc`:
- `metrics_clients_last_seen`
- `metrics_clients_daily`
- `clients_last_seen_joined`

* [DENG-9088](https://mozilla-hub.atlassian.net/browse/DENG-9088)


BenWu (commented):
`org_mozilla_vrbrowser` and `org_mozilla_firefoxreality` are the same apps so those per app id tables should also be included https://github.com/mozilla/probe-scraper/blob/e33f830bec2eee0a430d780fbea8faed4715dbde/repositories.yaml#L965

kwindau:
> `org_mozilla_vrbrowser` and `org_mozilla_firefoxreality` are the same apps so those per app id tables should also be included https://github.com/mozilla/probe-scraper/blob/e33f830bec2eee0a430d780fbea8faed4715dbde/repositories.yaml#L965

Thanks @BenWu , will do!

scholtzan (commented):
Maybe keeping a `deprecated_apps` list in https://github.com/mozilla/bigquery-etl/blob/90ab7e99373305739789b418e1d3342038aaf75b/bqetl_project.yaml#L319
could be cleaner? Especially if we deprecate more in the future?
The list could be passed into the templates and used for the condition for the `deprecated` flag

kwindau:
> Maybe keeping a `deprecated_apps` list in
> 
> https://github.com/mozilla/bigquery-etl/blob/90ab7e99373305739789b418e1d3342038aaf75b/bqetl_project.yaml#L319
> 
> could be cleaner? Especially if we deprecate more in the future?
> The list could be passed into the templates and used for the condition for the `deprecated` flag

I really like this idea, will close this PR and go that route instead. Thank you both!

PR #7772, opened 2025-07-11
fix(): app name remapping to usage reporting aggregates view

kik-kik:
# fix(): app name remapping to usage reporting aggregates view

### DIFF
--- a/tests/sql_generators/usage_reporting/expected/moz-fx-data-shared-prod/firefox_desktop/usage_reporting_active_users_aggregates/view.sql
+++ b/tests/sql_generators/usage_reporting/expected/moz-fx-data-shared-prod/firefox_desktop/usage_reporting_active_users_aggregates/view.sql
@@ -3,7 +3,25 @@ CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.firefox_desktop.usage_reporting_active_users_aggregates`
 AS
 SELECT
-  *,
+  * REPLACE (
+    CASE
+      WHEN app_name = "firefox_desktop"
+        THEN "Firefox Desktop"
+      WHEN app_name = "fenix"
+        THEN "Fenix"
+      WHEN app_name = "focus_android"
+        THEN "Focus Android"
+      WHEN app_name = "klar_android"
+        THEN "Klar Android"
+      WHEN app_name = "firefox_ios"
+        THEN "Firefox iOS"
+      WHEN app_name = "focus_ios"
+        THEN "Focus iOS"
+      WHEN app_name = "klar_ios"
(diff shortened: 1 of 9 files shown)


tests/sql_generators/usage_reporting/expected/moz-fx-data-shared-prod/firefox_desktop/usage_reporting_active_users_aggregates/view.sql
```diff
@@ -3,7 +3,25 @@ CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.firefox_desktop.usage_reporting_active_users_aggregates`
 AS
 SELECT
-  *,
+  * REPLACE (
+    CASE
+      WHEN app_name = "firefox_desktop"
```
irrationalagent, line 8: what happens if app_name is `firefox_desktop mozillaonline` (e.g. [here](https://github.com/mozilla/bigquery-etl/pull/7772/files#diff-5aeb5ee1412353c42b4ffcddd658206bdb387fea3d20ecc32139bffe40e254ddR14)) - does that mater?

kik-kik: We could also handle those here. I assumed for reporting we only care about the items I included.

kik-kik: ok, for completeness added the mozilla online cases for fenix and desktop:
https://github.com/mozilla/bigquery-etl/pull/7772/commits/e610b7bb47461a1780d1c0758c901faa42077387

irrationalagent:
I don't think we want to do this change anymore. 

https://github.com/mozilla/bigquery-etl/pull/7776/files should take up the previous renaming that you did on the active_users_aggregates view. The plan is to then backfill `usage_reporting_active_users` - this will cause the "friendly names" to propagate into the materialized aggregates table, so there will be no need for the mapping on the downstream views any longer.

kik-kik:
> I don't think we want to do this change anymore.
> 
> https://github.com/mozilla/bigquery-etl/pull/7776/files should take up the previous renaming that you did on the active_users_aggregates view. The plan is to then backfill `usage_reporting_active_users` - this will cause the "friendly names" to propagate into the materialized aggregates table, so there will be no need for the mapping on the downstream views any longer.

I see the change, the change above will require backfills for all products that are part of this generator. Right?

kik-kik:
This is no longer needed.

PR #7775, opened 2025-07-14
chore(glam): DENG-9113 create clients_hist snapshot and sample clients_hist_new

edugfilho:
## Description

GLAM ETL will take a snapshot of `clients_histogram_aggregates` to read from in release, instead of updating the table "in place".
This allows us to split the job into tasks that process a chunk of samples each time with the first job overwriting the destination table and the subsequent ones appending to it.
https://github.com/mozilla/bigquery-etl/pull/7774

* DENG-9113

### DIFF
--- a/bigquery_etl/glam/templates/clients_histogram_aggregates_v1.sql
+++ b/bigquery_etl/glam/templates/clients_histogram_aggregates_v1.sql
@@ -2,7 +2,7 @@
 {% set aggregate_filter_clause %}
 {% if filter_version %}
   LEFT JOIN
-    glam_etl.{{ prefix }}__latest_versions_v1
+    glam_etl.{{ prefix }} __latest_versions_v1
     USING (channel)
 {% endif %}
 WHERE
@@ -21,7 +21,7 @@ WITH extracted_accumulated AS (
   SELECT
     *
   FROM
-    glam_etl.{{ prefix }}__clients_histogram_aggregates_v1
+    glam_etl.{{ prefix }} __clients_histogram_aggregates_snapshot_v1
     {% if parameterize %}
       WHERE
         sample_id >= @min_sample_id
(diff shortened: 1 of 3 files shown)


bigquery_etl/glam/templates/clients_histogram_aggregates_v1.sql
```diff
@@ -2,7 +2,7 @@
 {% set aggregate_filter_clause %}
 {% if filter_version %}
   LEFT JOIN
-    glam_etl.{{ prefix }}__latest_versions_v1
+    glam_etl.{{ prefix }} __latest_versions_v1
```
BenWu: Did the formatter add spaces? Backticks might stop this from happening
```suggestion
    `glam_etl.{{ prefix }}__latest_versions_v1`
```

sql/moz-fx-glam-prod/glam_etl/firefox_desktop_glam_release__clients_histogram_aggregates_snapshot_v1/init.sql
BenWu, line 1: I think this will add a lot in storage costs even with a clone because the source table gets rewritten and then we need to pay for the time travel storage for the clone after it gets replaced.  This should be fine but something to be aware of

edugfilho: 🤔 what if we delete the snapshot after the source table is rewritten?

bigquery_etl/glam/templates/clients_histogram_aggregates_v1.sql
edugfilho: That did the trick. Thank you

sql/moz-fx-glam-prod/glam_etl/firefox_desktop_glam_release__clients_histogram_aggregates_snapshot_v1/init.sql
edugfilho: https://github.com/mozilla/telemetry-airflow/pull/2223/commits/c65307a124e98cc3f561cd2e9912f92aa6e1c483

BenWu: I expect it to still charge for time travel/fail-safe so it might not make a large difference.  It's fine to just do it and see though

PR #7778, opened 2025-07-14
[DENG-8496] Remove telemtry.modules view and shredder conf

BenWu:
## Description

The modules ping can be removed.  Removing the view isn't strictly necessary because removing the schema will cause the view to stop being generated, but it should help avoid temporary issues like in https://bugzilla.mozilla.org/show_bug.cgi?id=1940489
* DENG-8496


sql_generators/stable_views/__init__.py
```diff
@@ -481,3 +488,7 @@ def generate(target_project, output_dir, log_level, parallelism, use_cloud_funct
             ),
             one_schema_per_dataset,
         )
+
+
+if __name__ == "__main__":
+    generate()
```
BenWu, line 494: I added this to make it easier to run this as a script and attach a debugger.  I forgot to remove it but I think it's fine to keep

scholtzan (approved):

PR #7783, opened 2025-07-15
feat: remove app_channel for focus_ios from usage_reporting_clients_last_seen template

kik-kik:
# feat: remove app_channel for focus_ios from usage_reporting_clients_last_seen template

This field should not exist in this dataset. The field has now been deployed to the dataset and will be dropped once this change is merged.


kik-kik:
no longer valid.

PR #7776, opened 2025-07-14
fix(DENG-9112): Fix usage_reporting_active_users joins

kwindau:
## Description

**This PR consists of 3 main changes.**

**Change 1:**
This PR updates the join used in the SQL generator for the views to use pre-filtered CTEs starting 3/1/25 forward for clients last seen and clients daily. 
- `moz-fx-data-shared-prod.firefox_ios.usage_reporting_active_users`
- `moz-fx-data-shared-prod.focus_ios.usage_reporting_active_users`
- `moz-fx-data-shared-prod.firefox_desktop.usage_reporting_active_users`
- `moz-fx-data-shared-prod.fenix.usage_reporting_active_users`
- `moz-fx-data-shared-prod.focus_android.usage_reporting_active_users`

Once this is merged, Alekhya has a PR ready to backfill the impacted downstream tables here: [PR-7777](https://github.com/mozilla/bigquery-etl/pull/7777)

**Change 2:**
* Updating the `composite_active_users` view logic to only filter on app version for Firefox Desktop and not for anything else, because of an upstream issue where app_version will be null frequently with the way the current logic is written.  The longer term fix will be including app_version in the upstream `usage_reporting_clients_last_seen` tables, and pulling it downstream from there rather than how it is pulled today from `usage_reporting_clients_daily` tables.

**Change 3:**
* Updating 1 and 2, we noticed CI errors related to app_channel missing from the `org_mozilla_focus_ios.usage_reporting_clients_last_seen_v1` schema.yaml when it exists in the production BQ table.  
* To get changes 1 and 2 to merge & pass QA checks, I had to add the app_channel column to the associated schema.yaml and update the associated query for `org_mozilla_focus_ios.usage_reporting_clients_last_seen_v1`. 
* I'm not sure how this app_channel column ended up on the production table in the first place, but it was causing issues because it was on the BQ table but not in the schema.yaml, so we had to add it.  If this column is not needed in the long-term, then someone in the future can remove it, but I saw it already existed on the table and we needed to merge this today, so made this fix to get it to pass the QA checks to allow it to merge.
Screenshot 2025-07-14 at 6 53 31 PM

## Related Tickets & Documents
* [DENG-9112](https://mozilla-hub.atlassian.net/browse/DENG-9112)
[...]

PR #7786, opened 2025-07-15
 fix composite_active_users_aggregates.view.sql.jinja

irrationalagent:
## Description

Fixes an additional filter similar to what was done in https://github.com/mozilla/bigquery-etl/pull/7776

Intention is to match the behavior [here](https://github.com/mozilla/bigquery-etl/blob/main/sql_generators/usage_reporting/templates/composite_active_users.view.sql.jinja#L72-L76)

* DENG-9112

### DIFF
--- a/sql_generators/usage_reporting/templates/composite_active_users_aggregates.view.sql.jinja
+++ b/sql_generators/usage_reporting/templates/composite_active_users_aggregates.view.sql.jinja
@@ -90,11 +90,10 @@ SELECT
   SUM(monthly_users) AS monthly_users,
 FROM
   `{{ project_id }}.{{ app_name }}.usage_reporting_active_users_aggregates`
-WHERE
-  {% if app_name == "firefox_desktop" %}
-    WHERE
-      mozfun.norm.browser_version_info(app_version).major_version >= 136
-  {% endif %}
+{% if app_name == "firefox_desktop" %}
+  WHERE
+    app_version_major >= 136
+{% endif %}
 GROUP BY
   submission_date,
   first_seen_year,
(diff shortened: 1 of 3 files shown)


sql_generators/usage_reporting/templates/composite_active_users_aggregates.view.sql.jinja
```diff
@@ -37,11 +37,7 @@ FROM
   `{{ project_id }}.{{ app_name }}.active_users_aggregates`
   {% endif %}
 WHERE
-  {% if app_name == "fenix" %}
```
irrationalagent, line 40: I don't think this fenix specific behavior is needed - see [this chart](https://sql.telemetry.mozilla.org/queries/109302/source#268525) - we have usage reporting pings for fenix on versions 136, just like the other apps.

sql_generators/usage_reporting/templates/composite_active_users_aggregates.view.sql.jinja
```diff
@@ -94,12 +90,10 @@ SELECT
   SUM(monthly_users) AS monthly_users,
 FROM
   `{{ project_id }}.{{ app_name }}.usage_reporting_active_users_aggregates`
-WHERE
-  {% if app_name == "fenix" %}
```
kik-kik, line 98: @irrationalagent so we do not need any additional filtering here when doing the union?

irrationalagent: Correct, except for desktop we do not receive any usage pings from versions prior to 136.

PR #7803, opened 2025-07-18
feat: add composite key uniqueness metric to new_profile_clients bigconfig

kik-kik:
# feat: add composite key uniqueness metric to new_profile_clients bigconfig

### DIFF
--- a/sql_generators/mobile_kpi_support_metrics/templates/active_users.view.sql
+++ b/sql_generators/mobile_kpi_support_metrics/templates/active_users.view.sql
@@ -61,7 +61,7 @@ SELECT
       CAST(NULL AS STRING)
   END AS device_type,
   EXTRACT(YEAR FROM first_seen_date) AS first_seen_year,
-  app_display_version AS app_version,
+  NULLIF(app_display_version, "") AS app_version,
   `mozfun.norm.browser_version_info`(app_display_version).major_version AS app_version_major,
   `mozfun.norm.browser_version_info`(app_display_version).minor_version AS app_version_minor,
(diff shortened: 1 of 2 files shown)


sql_generators/mobile_kpi_support_metrics/templates/active_users.view.sql
```diff
@@ -61,7 +61,7 @@ SELECT
       CAST(NULL AS STRING)
   END AS device_type,
   EXTRACT(YEAR FROM first_seen_date) AS first_seen_year,
-  app_display_version AS app_version,
+  NULLIF(app_display_version, "") AS app_version,
```
kik-kik, line 64: When looking at the data noticed "" values. Doing this to make sure absence of value is represented with a `null`

sql_generators/mobile_kpi_support_metrics/templates/new_profile_clients.bigconfig.yml
```diff
@@ -21,6 +21,23 @@ tag_deployments:
           - saved_metric_id: is_99_percent_valid_normalized_channel
             rct_overrides:
               - first_seen_date
+      - column_selectors:
+        - name: {{ project_id }}.{{ project_id }}.{{ dataset }}_derived.{{ name }}_{{ version }}.*
```
chelseyklein, line 25:
```suggestion
        - name: {{ project_id }}.{{ dataset }}_derived.{{ name }}_{{ version }}.*
```

should project_id be twice?

chelseyklein: oh, i see that it is above

PR #7807, opened 2025-07-18
Command schema update using base schemas

lucia-vargas-a:
## Description

This PR implements [AD-963](https://mozilla-hub.atlassian.net/browse/AD-963) by adding two options to bqetl command `query schema update`.

These options enable using global and dataset specific schemas to automatically apply business definitions as columns descriptions in tables.

[AD-963]: https://mozilla-hub.atlassian.net/browse/AD-963?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -85,7 +85,6 @@ DEFAULT_CHECKS_FILE_NAME = "checks.sql"
 VIEW_FILE = "view.sql"
 MATERIALIZED_VIEW = "materialized_view.sql"
 NBR_DAYS_RETAINED = 775
-CANONICAL_SCHEMA_PATH = Path("bigquery_etl/schema")
 GLOBAL_SCHEMA_NAME = "global.yaml"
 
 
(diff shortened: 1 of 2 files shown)


bigquery_etl/schema/ads_derived.yaml
```diff
@@ -0,0 +1,352 @@
+fields:
```
lucia-vargas-a, line 1: This YAML schemas are migrated as-is from [private_bqetl](https://github.com/mozilla/private-bigquery-etl/tree/main/canonical) so they can be used with the command.

bigquery_etl/cli/query.py
```diff
@@ -85,6 +85,8 @@
 VIEW_FILE = "view.sql"
 MATERIALIZED_VIEW = "materialized_view.sql"
 NBR_DAYS_RETAINED = 775
+CANONICAL_SCHEMA_PATH = Path("bigquery_etl/schema")
```
scholtzan: we might want to configure this path in: https://github.com/mozilla/bigquery-etl/blob/main/bqetl_project.yaml
might be better than hardcoding it in here?

or alternatively it could be a CLI parameter

lucia-vargas-a: Definitely, moved to the project file.

PR #7815, opened 2025-07-21
feat(): add tests to mobile kpi support metrics generator

kik-kik:
# feat(): add tests to mobile kpi support metrics generator

This adds basic test to the `mobile_kpi_support_metrics` generator to make sure the templates are rendered correctly. This will allow us to work in the future on refactoring the core logic whilst ensuring the output remains the same.


LiamMcFall (approved):

lucia-vargas-a (approved):
I wonder if testing one or two browsers is enough since the logic is the same.

kik-kik:
> I wonder if testing one or two browsers e.g. Fenix is enough since the logic is the same. Or calling the query files from the original location.

I think this is enough, this include fenix, firefox_ios, klar android, and focus android. This should cover all cases.

PR #7843, opened 2025-07-23
[DENG-9121] Use backfill projects for dry runs by default

BenWu:
## Description

Changes dry runs to use a backfill project instead of shared-prod because some dry runs are a lot slower in shared-prod for some queries.  It doesn't really make CI faster because the most problematic queries were already being skipped but it should be more consistent.

This breaks dry runs for queries with non-fully qualified table names which I think is ok because there's a CI check for that but that doesn't run on generated sql.

* DENG-9121

### DIFF
--- a/bqetl_project.yaml
+++ b/bqetl_project.yaml
@@ -182,18 +182,18 @@ dry_run:
   - sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/query.sql
   - sql/moz-fx-data-shared-prod/telemetry_derived/clients_last_seen_v1/query.sql
   - sql/moz-fx-data-shared-prod/telemetry_derived/clients_last_seen_v2/query.sql
-  - sql/moz-fx-data-shared-prod/telemetry_derived/latest_versions/query.sql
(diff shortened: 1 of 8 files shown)


bqetl_project.yaml
```diff
@@ -236,6 +229,29 @@ dry_run:
...
+  - moz-fx-data-backfill-21
+  - moz-fx-data-backfill-22
+  - moz-fx-data-backfill-23
+  - moz-fx-data-backfill-24
+  - moz-fx-data-backfill-25
+  - moz-fx-data-backfill-26
+  - moz-fx-data-backfill-27
+  - moz-fx-data-backfill-28
+  - moz-fx-data-backfill-29
+  - moz-fx-data-backfill-30
```
whd:
```suggestion
```
You should skip backfill-30 for... [reasons](https://github.com/mozilla/global-platform-admin/blob/main/bigquery-reservations/tf/config.yaml#L15-L21)

BenWu:
> Have we asked GCP yet what bottleneck why this workaround is necessary?

Not yet, I'll do that next

BenWu:
I just noticed a lot of changes to task dependencies in the sql.diff so I'll look at that before merging

PR #7845, opened 2025-07-23
feat: create newtab-content aggregates and combine with newtab

LiamMcFall:
## Description

This PR creates an aggregates table for the newtab-content ping. It then also combines that ping data with the existing newtab ping.

* DENG-9057


sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_content_items_daily_v1/query.sql
```diff
@@ -0,0 +1,89 @@
+WITH events_unnested AS (
+  SELECT
+    DATE(submission_timestamp) AS submission_date,
+    -- mozfun.norm.browser_version_info(client_info.app_display_version).major_version AS app_version,
+    140 AS app_version,  -- Placeholder for app_version, adjust as needed
+    normalized_channel AS channel,
+    -- metrics.string.newtab_locale AS locale,
+    "MISSING" AS locale,  -- Placeholder for locale, adjust as needed
```
gkatre: Since we may not be adding locale to the `newtab_content` ping, we can perhaps remove it from the code

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_content_items_daily_v1/query.sql
```diff
@@ -0,0 +1,89 @@
...
+    event_category,
+    event_name,
+    mozfun.map.get_key(event_details, 'corpus_item_id') AS corpus_item_id,
+    SAFE_CAST(mozfun.map.get_key(event_details, 'position') AS INT64) AS position,
+    SAFE_CAST(mozfun.map.get_key(event_details, 'is_sponsored') AS BOOLEAN) AS is_sponsored,
+    SAFE_CAST(
+      mozfun.map.get_key(event_details, 'is_secton_followed') AS BOOLEAN
+    ) AS is_section_followed,
+    mozfun.map.get_key(event_details, 'matches_selected_topic') AS matches_selected_topic,
+    mozfun.map.get_key(event_details, 'newtab_visit_id') AS newtab_visit_id,
```
gkatre: `newtab_visit_id` can be removed. It does not exist in the `newtab_content` ping. Its also not needed in the final aggregation

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_content_items_daily_v1/query.sql
```diff
@@ -0,0 +1,89 @@
...
+  is_sponsored,
+  is_section_followed,
+  matches_selected_topic,
+  received_rank,
+  section,
+  section_position,
+  topic,
+  COUNTIF(event_name = 'impression') AS impression_count,
+  COUNTIF(event_name = 'click') AS click_count,
+  COUNTIF(event_name = 'save') AS save_count,
```
gkatre: `save` event does not exist in the [newtab_content ping](https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/pings/newtab-content?page=1&search=pocket) and can be removed

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_content_historical_items_daily_v1/query.sql
```diff
@@ -0,0 +1,13 @@
+SELECT
```
gkatre: Instead of creating a materialized table to merge the two datasets derived from the `newtab_v1` and `newtab_content_v1` pings, another option is to create an equivalent view — such as `firefox_desktop.newtab_content_daily_aggregates` — in the `firefox_desktop` dataset.

I'll leave it to you to choose whichever approach you feel is most appropriate to implement.

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_content_historical_items_daily_v1/query.sql
```diff
@@ -0,0 +1,13 @@
...
+FROM
+  `moz-fx-data-shared-prod.firefox_desktop_derived.newtab_content_items_daily_v1`
+WHERE
+  submission_date = @submission_date
+UNION ALL
+SELECT
+  *
+FROM
+  `moz-fx-data-shared-prod.firefox_desktop_derived.newtab_items_daily_v1`
+WHERE
```
gkatre: I noticed that you have applied the `major_version >= 140` filter (currently commented) in the `newtab_content_items_daily_v1` query to include versions 140 and above. We may need to do a similar thing while querying the `newtab_items_daily_v1` to filter out versions 140 and above to avoid duplicate counting

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_content_historical_items_daily_v1/query.sql
LiamMcFall: Good point, thank you for sharing an example. 

I think a view does make more sense.

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_items_daily_v1/query.sql
```diff
@@ -50,7 +55,7 @@ SELECT
   channel,
   locale,
   country,
-  mozfun.newtab.scheduled_surface_id_v1(country, locale) AS scheduled_surface_id,
+  newtab_content_surface_id,
```
gkatre: I notice that this is a change to the existing table and an attempt at replacing the udf with a recently introduced field `metrics.string.newtab_content_surface_id`.
While this attempt is headed in the right direction, it may be important to review the data in the ping table and see from which release this change has been applied. Its normal to find data from various releases in the data as users do not always update their browsers as soon as a version is released. Also, when we build data models we would like to design in a way that works with earlier releases as well, in case we need to perform a backfill and reproduce the computations that have worked in prior releases.

An alternative design approach would be to create a new version of this table based on the current requirements:
- remove the metrics associated with pocket `saves` (as this is related to the `Pocket` app thats shut down)
- do the merge from the two ping aggregations all in one table (this also eliminates the need for a view)

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_content_items_daily_v1/query.sql
```diff
@@ -0,0 +1,186 @@
...
+  FROM
+    newtab_events_unnested
+),
+newtab_daily_agg AS (
+  SELECT
+    submission_date,
+    app_version,
+    channel,
+    country,
+    '' AS newtab_content_surface_id,
```
gkatre: Any specific reason to apply `''` to this field?

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_content_items_daily_v1/query.sql
```diff
@@ -0,0 +1,186 @@
...
+    event_category,
+    event_name,
+    mozfun.map.get_key(event_details, 'corpus_item_id') AS corpus_item_id,
+    SAFE_CAST(mozfun.map.get_key(event_details, 'position') AS INT64) AS position,
+    SAFE_CAST(mozfun.map.get_key(event_details, 'is_sponsored') AS BOOLEAN) AS is_sponsored,
+    SAFE_CAST(
+      mozfun.map.get_key(event_details, 'is_section_followed') AS BOOLEAN
+    ) AS is_section_followed,
+    mozfun.map.get_key(event_details, 'matches_selected_topic') AS matches_selected_topic,
+    mozfun.map.get_key(event_details, 'newtab_visit_id') AS newtab_visit_id,
```
gkatre: `newtab_visit_id` may not be relevant to this aggregation as `newtab_content` does not have client or visit information

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_content_items_daily_v1/query.sql
```diff
@@ -0,0 +1,186 @@
...
+    metrics.string.newtab_content_surface_id AS newtab_content_surface_id,
+    timestamp AS event_timestamp,
+    category AS event_category,
+    name AS event_name,
+    extra AS event_details,
+    metrics.quantity.newtab_content_ping_version AS newtab_content_ping_version
+  FROM
+    `moz-fx-data-shared-prod.firefox_desktop.newtab_content`,
+    UNNEST(events)
+  WHERE
```
gkatre, line 111: I noticed that a filter is being applied to _exclude_ the redacted events (`content_redacted != 'true'`) being counted from the `newtab_v1` ping.
Would you be applying a suitable filter to the  `newtab_content` ping to _include only_ those events in the counts that were redacted from `newtab_v1`?

LiamMcFall: Only the `newtab` ping events have a `content_redacted` field ([ex](https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/pocket_click)). The `newtab_content` ping events don't have that field ([ex](https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/newtab_content_click)).

So IMO the best we can do is filter out events from `newtab` where the content was redacted, because that data should be coming from the `newtab_content` ping instead.

gkatre: Based on @rolf-moz 's comment [here](https://mozilla-hub.atlassian.net/browse/DO-2041?focusedCommentId=1109242), there appears to be a way to pick those events from the `newtab-content` ping which are redacted from the `newtab` ping

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_content_items_daily_v1/query.sql
```diff
@@ -0,0 +1,185 @@
...
+  FROM
+    newtab_events_unnested
+),
+newtab_daily_agg AS (
+  SELECT
+    submission_date,
+    app_version,
+    channel,
+    country,
+    CAST(NULL AS STRING) AS newtab_content_surface_id,
```
gkatre: still does not make sense why null for `newtab_content_surface_id`?

LiamMcFall: this was an oversight on my part. I misinterpreted this when doing the initial modeling and thought that this `newtab_content_surface_id` was _only_ a field in newtab-content similar to `newtab_content_ping_version`. Fixed in my recent commit to make sure `newtab_content_surface_id` is being pulled in from newtab events too. Thanks for the catch!

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_content_items_daily_v1/query.sql
```diff
@@ -0,0 +1,191 @@
+WITH newtab_events_unnested AS (
+  SELECT
+    DATE(submission_timestamp) AS submission_date,
+    mozfun.norm.browser_version_info(client_info.app_display_version).major_version AS app_version,
+    normalized_channel AS channel,
+    metrics.string.newtab_locale AS locale,
+    normalized_country_code AS country,
+    metrics.string.newtab_content_surface_id AS newtab_content_surface_id,
```
gkatre, line 8: `metrics.string.newtab_content_surface_id` was introduced recently. Perhaps we could apply [this](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_items_daily_v1/query.sql#L53) udf when the if the value is null

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_content_items_daily_v1/query.sql
```diff
@@ -0,0 +1,191 @@
...
+    section_position,
+    topic,
+    content_redacted,
+    newtab_content_ping_version
+),
+newtab_content_events_unnested AS (
+  SELECT
+    DATE(submission_timestamp) AS submission_date,
+    normalized_channel AS channel,
+    normalized_country_code AS country,
```
rolf-moz: If normalized_country_code comes from Glean, then I think the country code is that of the OHTTP relay. That means it would be US for Canadian users. We are only doing US users anyway, but for reference the country field is passed in the ping here. https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/pings/newtab-content

rolf-moz: use metrics.string.newtab_content_country
Note it may sometimes be missing if a user in the EN_US surface, for example, briefly leaves the country. If in that case you need to have a country you could default to none. If having some country is required you could use normalized_country_code as a backup.

LiamMcFall:
Here is the Looker PR ready to go for when we are ready to deploy this. [PR](https://github.com/mozilla/lookml-generator/pull/1300)

The only schema changes that will happen is the removal of the Pocket Save metrics and the renaming of `scheduled_surface_id` to `newtab_content_surface_id`. The rest of the fields remain the same, with some new ones added.

I can make the minor updates needed to the `Firefox New Tab: Feed Engagement` report to account for these changes and will communicate them in the #hnt-data slack channel.

PR #7848, opened 2025-07-24
chore: delete funnel retention views as no longer needed and the tables used by them have been deprecated for a while now

kik-kik:
# chore: delete funnel retention views as no longer needed and the tables used by them have been deprecated for a while now


chelseybeck (approved):
does what it says it does :)

PR #7855, opened 2025-07-25
fix(ech_adoption_rate): update schema & don't group by handshakes

edugfilho:
## Description

Fixes `ech_adoption_rate` dataset by applying `SUM` on `handshakes` (former `value`) instead of using that column as a group.
Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1957082

I don't think we need to increment the table version, because we're still iterating over its design in the ticket.
 
* DENG-9192


sql/moz-fx-data-shared-prod/telemetry_derived/ech_adoption_rate_v1/query.sql
```diff
@@ -5,13 +5,14 @@ WITH unioned AS (
...
-      WHEN COUNT(DISTINCT client_info.client_id) >= 5000
+      WHEN SUM(v.value) >= 5000
+        OR COUNT(DISTINCT client_info.client_id) >= 5000
         THEN metadata.geo.country
       ELSE 'OTHER'
     END AS country_code,
-    COUNT(DISTINCT client_info.client_id) AS client_count
   FROM
     `moz-fx-data-shared-prod.firefox_desktop.metrics`,
     UNNEST(metrics.labeled_custom_distribution.http3_ech_outcome) AS h,
```
kik-kik, line 18: If there are no http3 ourcomes, would the value be null or an empty array?

sql/moz-fx-data-shared-prod/telemetry_derived/ech_adoption_rate_v1/schema.yaml
```diff
@@ -12,7 +12,7 @@ fields:
   name: key
   type: STRING
 - mode: NULLABLE
-  name: value
+  name: handshakes
```
kik-kik, line 15: Would be nice if you could add a description of what `handshakes` means here.

sql/moz-fx-data-shared-prod/telemetry_derived/ech_adoption_rate_v1/query.sql
```diff
@@ -109,29 +110,21 @@ WITH unioned AS (
...
   key,
-  value,
-  country_code,
-  SUM(client_count) AS total_client_count
+  SUM(handshakes) AS handshakes,
+  SUM(client_count) AS total_client_count,
+  country_code
 FROM
   unioned
 GROUP BY
```
kik-kik, line 125: Not really relevant to this PR, but technically `GROUP BY ALL` is out of preview now so we can use it in prod queries if you want to make them a bit shorter:
https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#group_by_all

kik-kik (approved):

sql/moz-fx-data-shared-prod/telemetry_derived/ech_adoption_rate_v1/query.sql
edugfilho: It will be `null`, hence the filter below
`AND metrics.labeled_custom_distribution.http3_ech_outcome IS NOT NULL`

sql/moz-fx-data-shared-prod/telemetry_derived/ech_adoption_rate_v1/query.sql
BenWu: `GROUP BY ALL` is still against our [sql style guide](https://docs.telemetry.mozilla.org/concepts/sql_style.html) of being explicit (although the guide can change).  IMO there isn't a benefit to making the ETL queries slightly shorter and it's better to have all the columns for readability and make changes to the query a little more intentional, i.e. make you think for a few extra seconds about what you're doing when making changes to ETL

PR #7857, opened 2025-07-25
fix: undeprecate clients_activation_v1 as still in use

kik-kik:
# fix: undeprecate clients_activation_v1 as still in use


alexanderhe88:
Yes please proceed with that. THanks!

PR #7890, opened 2025-07-31
feat: add bigconfig files for braze data

LiamMcFall:
## Description

This PR add bigconfig files for `braze_derived` and `braze_external` datasets.

* DENG-8898


LiamMcFall:
Successfully generated a bigeye plan using `bigeye bigconfig plan` for each table.

sql/moz-fx-data-shared-prod/braze_derived/subscriptions_map_v1/checks.sql
```diff
@@ -1,10 +0,0 @@
---macro checks
-
-#fail
-{{ not_null(["braze_subscription_name", "description", "mozilla_subscription_id", "firefox_subscription_id", "mozilla_dev_subscription_id", "basket_slug"]) }}
-
-#fail
-{{ min_row_count(1) }}
-
-#fail
-{{ is_unique(["mozilla_subscription_id", "firefox_subscription_id", "mozilla_dev_subscription_id"]) }}
```
kik-kik, line 10: I believe this will check that the combination of all of those columns is unique:
https://github.com/mozilla/bigquery-etl/blob/main/tests/checks/is_unique.jinja

the `is_unique` saved metric in bigeye config would check each column individually therefore this is not exactly what you need to translate this check to bigeye metric. You will need to use this:
https://github.com/mozilla/bigquery-etl/pull/7862 (once it is merged)

sql/moz-fx-data-shared-prod/braze_external/users_previous_day_snapshot_v1/checks.sql
```diff
@@ -1,4 +0,0 @@
--- macro checks
-
-#fail
-{{ min_row_count(85000000) }}
```
kik-kik, line 4: Freshness and volume are not quite the same as this metric. You probably want to use the predefined `COUNT_ROWS` metric with a minimum threshold defined:
https://docs.bigeye.com/docs/available-metrics (search for `COUNT_ROWS`)

sql/moz-fx-data-shared-prod/braze_external/changed_users_v1/checks.sql
```diff
@@ -1,10 +0,0 @@
--- macro checks
-
-#warn
-{{ not_null(["external_id", "status", "email", "create_timestamp", "update_timestamp"]) }}
-
-#warn
-{{ min_row_count(1) }}
-
-#warn
-{{ is_unique(["external_id", "email"]) }}
```
kik-kik, line 10: If I recall correctly this will check that the combination of the two is unique and not that each of these columns has unique values which is what the bigeye config would check. So slightly different behaviour.

sql/moz-fx-data-shared-prod/braze_derived/user_profiles_v1/checks.sql
```diff
@@ -1,10 +0,0 @@
--- macro checks
-
-#fail
-{{ not_null(["external_id", "email", "email_subscribe", "has_fxa", "create_timestamp", "update_timestamp"]) }}
-
-#fail
-{{ min_row_count(75000000) }}
```
kik-kik, line 7: Freshness and volume are not quite the same as this metric. You probably want to use the predefined COUNT_ROWS metric with a minimum threshold defined:
https://docs.bigeye.com/docs/available-metrics (search for COUNT_ROWS)

sql/moz-fx-data-shared-prod/braze_derived/user_profiles_v1/checks.sql
```diff
@@ -1,10 +0,0 @@
...
-
-#fail
-{{ is_unique(["external_id", "email", "fxa_id_sha256"]) }}
```
kik-kik, line 10: Should these provide a unique value as a set or should each of these columns contain unique values irrespectively of the other ones?

sql/moz-fx-data-shared-prod/braze_derived/users_v1/checks.sql
```diff
@@ -1,10 +0,0 @@
--- macro checks
-
-#fail
-{{ not_null(["external_id", "email", "email_subscribe", "has_fxa", "create_timestamp", "update_timestamp"]) }}
-
-#fail
-{{ min_row_count(85000000) }}
```
kik-kik, line 7: See other comments related to min_row_count.

sql/moz-fx-data-shared-prod/braze_derived/users_v1/checks.sql
```diff
@@ -1,10 +0,0 @@
...
-
-#fail
-{{ is_unique(["external_id", "email", "fxa_id_sha256"]) }}
```
kik-kik, line 10: See other comments related to is_unique containing more than one column

sql/moz-fx-data-shared-prod/braze_derived/subscriptions_map_v1/checks.sql
chelseyklein: we expect each column to be individually unique in this table, so i don't think a grouped uniqueness check is necessary here

sql/moz-fx-data-shared-prod/braze_external/changed_users_v1/checks.sql
chelseyklein: on this one, we do want it grouped. i think a user could theoretically have multiple accounts using the same email

chelseyklein: but, also, `external_id` should be individually unique - assuming there is a way to do both :)

sql/moz-fx-data-shared-prod/braze_derived/user_profiles_v1/checks.sql
chelseyklein: `external_id` should be individually unique and `fxa_id_sha256` should be individually unique, `email may not always be`

sql/moz-fx-data-shared-prod/braze_derived/users_v1/checks.sql
chelseyklein: `external_id` should be individually unique and `fxa_id_sha256` should be individually unique, `email may not always be`

sql/moz-fx-data-shared-prod/braze_external/changed_users_v1/checks.sql
LiamMcFall: I think I might need to leverage the custom rules workaround [here](https://docs.telemetry.mozilla.org/cookbooks/data_monitoring/bigquery_etl_integration) in order to do combined column uniqueness. I didn't see any way to concatenate or combine these fields as a native functionality within a bigconfig file in the docs 😢

sql/moz-fx-data-shared-prod/braze_derived/subscriptions_map_v1/checks.sql
kik-kik: ok, in that case the bigConfig configuration should be correct.

sql/moz-fx-data-shared-prod/braze_derived/user_profiles_v1/checks.sql
kik-kik: @LiamMcFall I wonder if it would make sense to just remove email from the is_unique list in the bigConfig and leave it at that since we expect the same email to potentially appear more than once?

LiamMcFall: I think so too, if the other 2 are individually unique then it would still satisfy the uniqueness criteria even if emails can appear multiple times.

PR #7925, opened 2025-08-07
Set @mozilla/data-platform-infra-wg CODEOWNER on most files, except sql/, sql_generators/, dags.yaml

scholtzan:
This is required for https://mozilla-hub.atlassian.net/browse/DENG-9290

This PR configures `@mozilla/data-platform-infra-wg` as being required as reviewer for everything except `sql/`, `sql_generators/`, `tests/sql` and `dags.yaml`. The idea is to have more contributors added to the repo with write and review permissions so that they can operate more self-sufficiently without needing a DE to approve and merge their PRs. However, in order to ensure that no critical code gets changed we need to require reviews from DE on these more critical code paths.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -6,6 +6,9 @@
 /tests/sql/
 dags.yaml
 
+# require review on datasets, otherwise permissions could be changed without review
+**/dataset_metadata.yaml @mozilla/data-platform-infra-wg
+
 # These datasets are subject to the additional change control procedures
 # described in https://docs.google.com/document/d/1TTJi4ht7NuzX6BPG_KTr6omaZg70cEpxe9xlpfnHj9k/
 # Active Users
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -3,7 +3,6 @@
 
 # do not require review from data platform for the following files and directories
 /sql/
-/sql_generators/
 /tests/sql/
 dags.yaml
 
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,3 +1,12 @@
+# require @mozilla/data-platform-infra-wg review for everything else
+* @mozilla/data-platform-infra-wg
+
+# do not require review from data platform for the following files and directories
+/sql/
+/sql_generators/
+/tests/sql/
+dags.yaml
+
 # These datasets are subject to the additional change control procedures
 # described in https://docs.google.com/document/d/1TTJi4ht7NuzX6BPG_KTr6omaZg70cEpxe9xlpfnHj9k/
 # Active Users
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -1,3 +1,12 @@
+# require @mozilla/data-platform-infra-wg review for everything else
+* @mozilla/data-platform-infra-wg
+
+# do not require review from data platform for the following files and directories
+/sql/
+/sql_generators/
+/tests/sql/
+dags.yaml
```
scholtzan: Based on the official docs, this will exclude the listed directories from requiring review as no codeowner has been specified for those: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file

kwindau: Should we take sql_generators out of this list?

sean-rose: I'm inclined to agree that SQL generators should perhaps require DE review as they're more complex and some of them have a pretty big blast radius if problems were introduced (e.g. `glean_usage` and `stable_views`).  Maybe if any SQL generators are created by other teams we could exclude specific such SQL generators from requiring DE review?

scholtzan: Agreed. I believe it is pretty rare for users to make changes in `sql_generators/` to begin with and most changes are in collaboration with a DE usually (e.g. funnels)

sean-rose: Also, maybe we should require DE review for `dataset_metadata.yaml` files?  Otherwise restricted access datasets could be reconfigured without our knowledge, and there are conventions for dataset names and ACLs that we need to make sure are followed consistently.

While adding new datasets and changing dataset permissions isn't uncommon, it doesn't seem super frequent, so I'm thinking requiring DE review for adding/changing datasets would hopefully not be too much of a bottleneck.

scholtzan: Yeah, that sounds like a good idea

kwindau (approved):

CODEOWNERS
```diff
@@ -1,3 +1,14 @@
+# require @mozilla/data-platform-infra-wg review for everything else
+* @mozilla/data-platform-infra-wg
```
sean-rose, line 2: Not an issue with this PR per se, but it seems like the @mozilla/data-platform-infra-wg team membership might need to be updated (e.g. @kwindau isn't currently a member).

whd: https://github.com/orgs/mozilla/teams/dataplatform-wg is the autogenerated team from https://protosaur.dev/dawg/workgroup/dataplatform/, which might be reasonable to use in this instance.

PR #7926, opened 2025-08-07
feat(DENG-9319): Create dl_token_ga_attribution_lookup_v2

kwindau:
## Description

This PR creates a new table: 
`moz-fx-data-shared-prod.stub_attribution_service_derived.dl_token_ga_attribution_lookup_v2`

This table has a different grain than the V1 table (4 part key instead of a 3 part key), so chose to make a new version instead.

It also adds "depends_on_past: True" to the original V1 table because I noticed that was missing, and should be on that table as well.

* [DENG-9319](https://mozilla-hub.atlassian.net/browse/DENG-9319)

[DENG-9319]: https://mozilla-hub.atlassian.net/browse/DENG-9319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sql/moz-fx-data-shared-prod/stub_attribution_service_derived/dl_token_ga_attribution_lookup_v2/query.sql
```diff
@@ -0,0 +1,79 @@
...
+  SELECT DISTINCT
+    IFNULL(mozfun.ga.nullify_string(jsonPayload.fields.dltoken), "") AS dl_token,
+    IFNULL(mozfun.ga.nullify_string(jsonPayload.fields.client_id_ga4), "") AS ga_client_id,
+    IFNULL(mozfun.ga.nullify_string(jsonPayload.fields.session_id), "") AS stub_session_id,
+    CAST(@download_date AS date) AS first_seen_date,
+    jsonPayload.fields.dlsource AS download_source --this only recently started getting filled in, during the transition to Firefox.com
+  FROM
+    `moz-fx-stubattribut-prod-32a5.stubattribution_prod.stdout`
+  WHERE
+    DATE(timestamp) = (SELECT CAST(download_date AS date) FROM check_download_date)
```
lucia-vargas-a: Would a CTE be faster? or you prefer less verbose?

sql/moz-fx-data-shared-prod/stub_attribution_service_derived/dl_token_ga_attribution_lookup_v2/checks.sql
```diff
@@ -0,0 +1,5 @@
+#fail
+{{ is_unique(['dl_token', 'ga_client_id', 'stub_session_id', 'download_source']) }}
```
lucia-vargas-a, line 2: If `ga_client_id` is at client level, would `first_seen_date` be unique as well?

sql/moz-fx-data-shared-prod/stub_attribution_service_derived/dl_token_ga_attribution_lookup_v2/metadata.yaml
```diff
@@ -0,0 +1,25 @@
+friendly_name: Download Token Google Analytics Attribution Lookup V2
+description: |-
+  This table lets you lookup GA attribution data for dl_tokens.
+
+  1 row per-(dl_token, ga_client_id, stub_session_id, download_source).
+
+  dl_token - Available in Stub Attribution Service and Telemetry
```
lucia-vargas-a, line 7: Are you adding the notes here so that we keep standard schema descriptions?

lucia-vargas-a (approved):
A few minor comments

sql/moz-fx-data-shared-prod/stub_attribution_service_derived/dl_token_ga_attribution_lookup_v2/query.sql
kwindau: Good catch, thank you @lucia-vargas-a ! I switched it to be simpler so the query would be more efficient.

sql/moz-fx-data-shared-prod/stub_attribution_service_derived/dl_token_ga_attribution_lookup_v2/checks.sql
kwindau: It would be but I try to look for the smallest primary key that is unique, which is the 4 part key (the first seen date is a measure for those 4 dimensions)

sql/moz-fx-data-shared-prod/stub_attribution_service_derived/dl_token_ga_attribution_lookup_v2/metadata.yaml
kwindau: good catch, will make sure these are consistent

PR #7953, opened 2025-08-13
feat(DENG-9250): add syndicate dataset for firefox_app_store_v2

kik-kik:
feat(DENG-9250): add syndicate dataset for firefox_app_store_v2

* DENG-XXXX
* DSRE-XXXX

### DIFF
--- a/sql/moz-fx-data-shared-prod/app_store_v2_syndicate/dataset_metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/app_store_v2_syndicate/dataset_metadata.yaml
@@ -1,0 +1,26 @@
+friendly_name: app_store_syndicate
+description: |
+  Syndicated App Store data from Fivetran. User-facing views into this data can be found
+  in the app_store dataset.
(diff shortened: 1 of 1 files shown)


sql/moz-fx-data-shared-prod/app_store_v2_syndicate/dataset_metadata.yaml
```diff
@@ -0,0 +1,26 @@
+friendly_name: app_store_syndicate
```
sean-rose, line 1: Should this dataset have Firefox in the name, if it's specifically for Firefox and not all Mozilla apps?

sean-rose: Actually, looking in BigQuery I now see the earlier `moz-fx-data-bq-fivetran.firefox_app_store` dataset was syndicated to an `app_store_syndicate` dataset, so this is in keeping with that precedent, which I guess makes sense for consistency.  So I retract my suggestion.

PR #7956, opened 2025-08-14
DENG-8939-Newtab Visits additional metrics and computations

gkatre:
## Description

This PR includes several additional metrics suitable for adhoc analysis at the daily visit level for Firefox desktop newtab.

* DENG-8939
* [Newtab Visits Daily v2 Requirements](https://docs.google.com/document/d/1cimTK12aHjy7lN05P5sLfr70LmpxujEAcqcF5HHNc88/edit?tab=t.0)


sql/moz-fx-data-shared-prod/firefox_desktop/newtab_visits_daily/view.sql
```diff
@@ -1,31 +1,57 @@
...
+      OR is_wallpaper_interaction
+      OR is_other_interaction
+    ) AS is_nonsearch_interaction,
+    (
+      is_content_interaction
+      AND NOT is_sponsored_content_interaction
+    ) AS is_organic_content_interaction,
+    (
+      is_topsite_interaction
+      AND NOT is_sponsored_topsite_interaction
```
jaredsnyder: `is_sponsored_content_interaction` and `is_sponsored_topsite_interaction` will be true if there are _any_ sponsored content/topsite interactions, so this will only be true if the content/topsite are _exclusively_ organic. I think it might be better to write it so it will be true if there is _any_ organic content interaction but I'd like to hear from @ilanasegall on this

land-edi (commented):
@gkatre Since we are updating the newtab tables and will do backfill, can we add sample_id to make it easy to query?

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_visits_daily_v2/query.sql
```diff
@@ -281,6 +289,293 @@ SELECT
...
+    AND SAFE_CAST(mozfun.map.get_key(event_details, 'is_sponsored') AS BOOLEAN)
+  )
+  AND LOGICAL_OR(is_default_ui) AS is_sponsored_topsite_dismissal,
+  LOGICAL_OR(event_category = 'newtab.search.ad' AND event_name = 'impression')
+  AND LOGICAL_OR(is_default_ui) AS is_search_ad_impression,
+  LOGICAL_OR(event_category = 'newtab' AND event_name IN ('weather_impression'))
+  AND LOGICAL_OR(is_default_ui) AS is_widget_impression,
+  LOGICAL_OR(
+    event_category = 'newtab'
+    AND event_name IN ('sections_impression', 'inline_selection_impression')
```
jaredsnyder: @ilanasegall does it make sense for these events to set `is_other_impression` to true?

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_visits_daily_v2/query.sql
```diff
@@ -281,6 +289,293 @@ SELECT
...
+      )
+    ),
+    0
+  ) AS any_interaction_count,
+  IF(
+    LOGICAL_OR(is_default_ui),
+    COUNTIF(
+      event_category = 'newtab.search'
+      AND event_name IN ('issued')
+      AND SAFE_CAST(mozfun.map.get_key(event_details, 'is_sponsored') AS BOOLEAN)
```
jaredsnyder: Is there a reason for this filter?  I'm not seeing `is_sponsored` on the [Glean page for the event](https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/newtab_search_issued)

jaredsnyder: There are also currently no records (on 2025/08/01 anyway) where this column is >0

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_visits_daily_v2/query.sql
```diff
@@ -281,6 +289,293 @@ SELECT
...
+    0
+  ) AS search_ad_impression_count,
+  IF(
+    LOGICAL_OR(is_default_ui),
+    COUNTIF(
+      event_category = 'newtab'
+      AND event_name IN (
+        'weather_change_display',
+        'weather_open_provider_url',
+        'weather_location_selected'
```
jaredsnyder: Add new widget interaction events:
[newtab.widgets_lists_change_display](https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/newtab_widgets_lists_change_display)
[newtab.widgets_lists_user_event](https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/newtab_widgets_lists_user_event)
[newtab.widgets_timer_change_display](https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/newtab_widgets_timer_change_display)
[newtab.widgets_timer_toggle_notification](https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/newtab_widgets_timer_toggle_notification)
[newtab.widgets_timer_user_event](https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/newtab_widgets_timer_user_event)

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_visits_daily_v2/query.sql
```diff
@@ -281,6 +289,293 @@ SELECT
...
+      )
+    ),
+    0
+  ) AS widget_interaction_count,
+  IF(
+    LOGICAL_OR(is_default_ui),
+    COUNTIF(event_category = 'newtab' AND event_name IN ('weather_impression')),
```
jaredsnyder: Add new widget impression events:

[newtab.widgets_lists_impression](https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/newtab_widgets_lists_impression)
[newtab.widgets_timer_impression](https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/newtab_widgets_timer_impression)

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_visits_daily_v2/query.sql
```diff
@@ -281,6 +289,293 @@ SELECT
...
+        'inline_selection_click'
+      )
+    ),
+    0
+  ) AS other_interaction_count,
+  IF(
+    LOGICAL_OR(is_default_ui),
+    COUNTIF(
+      event_category = 'newtab'
+      AND event_name IN ('sections_impression', 'inline_selection_impression')
```
jaredsnyder: @ilanasegall do we want these as "other" impressions?

sql/moz-fx-data-shared-prod/firefox_desktop/newtab_visits_daily/view.sql
jaredsnyder: This implementation is also different than the description in the README, so we should update here or update the description

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_visits_daily_v2/schema.yaml
```diff
@@ -171,3 +171,131 @@ fields:
...
+  description: Array of experiments to associate with each visit
+- name: newtab_weather_enabled
+  type: BOOLEAN
+  description: Whether the weather widget is enabled
+- name: default_search_engine
+  type: STRING
+  description: Engine_id of the default search engine
+- name: default_private_search_engine
+  type: STRING
+  description: Engine_id of the private search engine
```
jaredsnyder:
```suggestion
  description: Engine_id of the default private search engine
```

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_visits_daily_v2/schema.yaml
```diff
@@ -171,3 +171,131 @@ fields:
...
+  description: Whether a search ad was impressed with default UI*
+- name: is_widget_impression
+  type: BOOLEAN
+  description: Whether a widget was impressed with default UI*
+- name: is_other_impression
+  type: BOOLEAN
+  description: Whether other misc events impressed with default UI*
+- name: is_section
+  type: BOOLEAN
+  description: Whether a visit had an impression that belonged to a section with default UI*
```
jaredsnyder: Is there a reason for the `*` at the end of these?  Wondering if it's a holdover from the sheet

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_visits_daily_v2/schema.yaml
```diff
@@ -171,3 +171,131 @@ fields:
...
+  description: Count of other misc interactions
+- name: other_impression_count
+  type: INTEGER
+  description: Count of other misc impressions
+- name: newtab_visit_duration
+  type: INTEGER
+  description: Visit duration in milliseconds (timestamp diff between close and open event)
+- name: newtab_window_inner_height
+  type: INTEGER
+  description: The height of the browser window
```
jaredsnyder:
```suggestion
  description: The height of the browser window in pixels
```

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_visits_daily_v2/schema.yaml
```diff
@@ -171,3 +171,131 @@ fields:
...
+- name: newtab_window_inner_width
+  type: INTEGER
+  description: The width of the browser window
```
jaredsnyder:
```suggestion
  description: The width of the browser window in pixels
```

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_visits_daily_v2/schema.yaml
```diff
@@ -171,3 +171,131 @@ fields:
...
+  type: STRING
+  description: Engine_id of the private search engine
+- name: topsite_rows
+  type: INTEGER
+  description: The number of topsite tile rows configured to be shown on the newtab page
+- name: topsite_sponsored_tiles_configured
+  type: INTEGER
+  description: The number of topsite tiles configured to be shown on newtab
+- name: newtab_blocked_sponsors
+  type: STRING
```
jaredsnyder, line 229: Believe this is actually an array when there are multiple

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_visits_daily_v2/schema.yaml
gkatre: Good catch. There is no need for the `*`. I will remove it

sql/moz-fx-data-shared-prod/firefox_desktop/newtab_visits_daily/view.sql
gkatre: @jaredsnyder 
`is_sponsored_*` or `is_organic_*` are `boolean` **aggregations** at visit level and there could be various interpretations for its definitions. Would it be possible for you and @ilanasegall to decide which implementation is really useful for your use cases? After you have finalized the definitions, i could make the necessary implementation changes.

Note: that we have related metrics `sponsored_*_count` and `organic_*_count`, which I believe has the most value for DS folks. Those metrics will take into account your original concern of sponsored or organic not being exclusive within the same visit.

gkatre: Another thought for @jaredsnyder and @ilanasegall to consider:
With `sponsored_*_count` and `organic_*_count` being available, do you even need the boolean fields `is_sponsored_*` or `is_organic_*`? The `0` or `non-0` values in the count fields could perhaps be used to interpret various scenarios?

ilanasegall: we discussed potentially changing this metric to has_sponsored* or has_organic* such that they could both be true or false and it would make intuitive sense.

sql/moz-fx-data-shared-prod/firefox_desktop/newtab_visits_daily/view.sql
```diff
@@ -1,31 +1,57 @@
...
-    is_topsite_interaction
-    AND NOT is_sponsored_topsite_interaction
-  ) AS is_organic_topsite_interaction,
-  *
+  is_any_interaction,
+  is_nonsearch_interaction,
+  is_organic_content_interaction,
+  is_organic_topsite_interaction,
+  layout_type,
+  mozfun.newtab.determine_tiles_per_row_v1(layout_type, newtab_window_inner_width) AS tiles_per_row,
```
ilanasegall, line 21: is tiles per row the right paradigm now that we're rolling sections out to everyone and we have different-sized cards?

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_visits_daily_v2/query.sql
ilanasegall: I would actually drop "inline_selection_impression" altogether as it was an element for an experiment several months ago and has no entries in the last several months. "sections_impression" should be content.

sql/moz-fx-data-shared-prod/firefox_desktop/newtab_visits_daily/view.sql
gkatre: @ilanasegall Are you saying that there is a possibility for a section layout with a specific window width to have different tile row sizes within the same window?

If so, then it perhaps does not make sense to do this computation

sql/moz-fx-data-shared-prod/firefox_desktop/newtab_visits_daily/view.sql
gkatre: > we discussed potentially changing this metric to has_sponsored* or has_organic* such that they could both be true or false and it would make intuitive sense.

If has_sponsored* and has_organic* are meant to replace all the is_sponsored* or is_organic* in the visits table, this would mean making changes to the existing fields on that table. 
A change like this involving field names/definitions on existing fields, would mean creating a new version of the table (`newtab_visits_daily_v3`) and eventually decommissioning the existing table.

This is an extra bit of effort. But if you and @jaredsnyder are very sure about these new definitions and fields name changes, we should do those changes now rather than later.

gkatre: Another thought:
if has_sponsored* or has_organic* can be computed as:
```
has_sponsored* = TRUE if sponsored_*_count > 0 else FALSE
has_organic* = TRUE if organic_*_count > 0 else FALSE
```

then we do not need to add those fields to the table as they can be computed easily later

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_visits_daily_v2/query.sql
gkatre: @jaredsnyder Thanks for the detailed review!!!

As suggested by @ilanasegall [here](https://github.com/mozilla/bigquery-etl/pull/7956/files/a883085da7fe2a6dda853e005df8d952268bf356#r2286423669), i will remove the "inline_selection_impression"

sql/moz-fx-data-shared-prod/firefox_desktop/newtab_visits_daily/view.sql
jaredsnyder: Confirmed with Nathan Barrett that the only cards that take up multiple columns are billboard and leaderboard, which are their own special cases so the computation is valid.

sql/moz-fx-data-shared-prod/firefox_desktop/newtab_visits_daily/view.sql
jaredsnyder: The logic you suggested for the has_* in your last comment makes sense, and I think it's a valid point that it's a simple enough computation that we could remove them.  I agree with Ilana that I'd like to see the is_* columns renamed so I'd propose one of two options:

- Remove the is_* columns altogether and replace the logic dependent on them in the rollups with `LOGICAL_OR(*_count>0)`
- replace the is_* with the has_* columns as you defined them

This could be a "fast-follow" in a different PR given you'd need to make a v3 table

gkatre: @jaredsnyder @ilanasegall 
As you both confirm the change to remove (or replace) the `is_*` fields is final, if the approach is to fast-follow in a different PR where we remove the `is_*` fields, then we would need to proceed with this PR with the existing implementation of the `is_*` fields.
If we agree to this approach, we can move forward quickly on this PR by ignoring the `is_*` fields for now

jaredsnyder: Fine by me as long as we make a ticket now so this doesn't get forgotten.  @gkatre would you mind creating a new ticket in an appropriate epic/etc, and I can fill out the details?

gkatre: > Fine by me as long as we make a ticket now so this doesn't get forgotten. @gkatre would you mind creating a new ticket in an appropriate epic/etc, and I can fill out the details?

Yes. I will create a ticket for v3. 
And yes, this cannot be delayed (or forgotten). The only way this makes sense is to do v3 for visits next. So that we do not have to follow through with all the subsequent processes on v2...instead we do it all for v3.

Note: in this PR for v2, we are only adding new fields...not making changes to the previously created fields. For v3 we will be removing the old `is_*` fields as well.

But, for this PR, lets make sure all other non `is_*` fields are all good. That way when we fast-forward into v3 we do not need to re-visit everything all over again.

jaredsnyder: That all sounds good!

For this PR, given that we want these columns to be different anyway, maybe we just remove all the new `is_*` columns from this PR?

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_visits_daily_v2/query.sql
gkatre: > Is there a reason for this filter? I'm not seeing `is_sponsored` on the [Glean page for the event](https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/newtab_search_issued)

Good catch! i will remove the is_sponsored condition.


> There are also currently no records (on 2025/08/01 anyway) where this column is >0

After removing the  is_sponsored condition, i do see visits with > 0 for that column for '2025-08-01'

sql/moz-fx-data-shared-prod/firefox_desktop/newtab_visits_daily/view.sql
gkatre: > For this PR, given that we want these columns to be different anyway, maybe we just remove all the new `is_*` columns from this PR?

I am noticing for some of the `is_*`s there are no corresponding `*_counts`. For those missing ones I will convert the `is_*` boolean fields to count fields

jaredsnyder: Sounds good

jaredsnyder (approved):
All comments resolved, the query runs just fine and the view is very simple so I didn't feel the need to check in the console.  Great stuff!

PR #7974, opened 2025-08-19
DO-2075 Added fenix and desktop baseline city seen tables

wwyc:
## Description

Initialize the *baseline_city_seen tables by deriving each client’s first-seen and last-seen city, subdivision and country fields from the stable tables.
Note: This one-time initialization logic will no longer apply once city/subdivision/country fields are nulled in the stable tables.
Ongoing updates: After initialization, the tables will be updated daily via ETL using live tables (appending new clients and advancing last-seen values).

* [DO-2075](https://mozilla-hub.atlassian.net/browse/DO-2075)

[DO-2075]: https://mozilla-hub.atlassian.net/browse/DO-2075?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dags.yaml
+++ b/dags.yaml
@@ -2517,4 +2517,4 @@ bqetl_baseline_clients_city_seen:
   description: Scheduled queries for client city seen tables
   tags:
     - impact/tier_2
-    - repo/bigquery-et
+    - repo/bigquery-etl


sql/moz-fx-data-shared-prod/fenix_derived/clients_city_seen_v1/schema.yaml
```diff
@@ -0,0 +1,41 @@
+fields:
+- name: app_name
```
wwyc: This field was added to support data validations.  Need to confirm if this field is needed and if this is the right name.  For example, this would contain "org_mozilla_firefox".

lucia-vargas-a: `org_mozilla_firefox` is the dataset (app + channel), maybe `app_dataset` or similar is a better name?
-- `app_name` in this case would be Fenix.

sql/moz-fx-data-shared-prod/fenix_derived/clients_city_seen_v1/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: fenix Clients City Seen
+description: |-
+owners:
+  - wichan@mozilla.com
+  - rbaffourawuah@mozilla.com
+labels:
+  change_controlled: true
```
lucia-vargas-a: This label requires [adding the table to the CODEOWNERS](https://mozilla.github.io/bigquery-etl/cookbooks/common_workflows/#setting-up-change-control-to-code-files) file with an owner -preferably a GitHub identity-.

sql/moz-fx-data-shared-prod/fenix_derived/clients_city_seen_v1/schema.yaml
```diff
@@ -0,0 +1,41 @@
...
+  mode: NULLABLE
+  description: Name of the browser app.
+- name: client_id
+  type: STRING
+  mode: NULLABLE
+  description: A UUID uniquely identifying the client.
+- name: first_seen_geo_date
+  type: DATE
+  mode: NULLABLE
+  description: Date when the first seen geo data were captured.
```
lucia-vargas-a: Typo?
```suggestion
  description: Date when the first seen geo data was captured.
```

sql/moz-fx-data-shared-prod/fenix_derived/clients_city_seen_v1/schema.yaml
```diff
@@ -0,0 +1,41 @@
+fields:
+- name: app_name
+  type: STRING
+  mode: NULLABLE
+  description: Name of the browser app.
+- name: client_id
+  type: STRING
+  mode: NULLABLE
+  description: A UUID uniquely identifying the client.
+- name: first_seen_geo_date
```
lucia-vargas-a: Simpler if removing 'geo' in all column names? -- maybe just keep it in the description?

sql/moz-fx-data-shared-prod/fenix_derived/clients_city_seen_v1/schema.yaml
lucia-vargas-a: Is this the actually the date when we first captured geo data for the client_id, or the first_seen-date (where geo data could be NULL)?

sql/moz-fx-data-shared-prod/firefox_desktop_derived/clients_city_seen_v1/query.sql
```diff
@@ -0,0 +1,144 @@
...
+      submission_date,
+      client_id
+    HAVING
+      COUNT(*) > 150000
+  ),
+  windowed_firefox_desktop AS (
+    SELECT
+      submission_date,
+      client_id,
+      sample_id,
```
lucia-vargas-a: `sample_id` seems redundant as it's always the same per `client_id`. You can consider removing it for lower costs (hashing, comparing)

sql/moz-fx-data-shared-prod/firefox_desktop_derived/clients_city_seen_v1/query.sql
```diff
@@ -0,0 +1,144 @@
...
+    LEFT JOIN
+      `moz-fx-data-shared-prod.firefox_desktop_derived.baseline_clients_first_seen_v1` AS cfs
+      USING (client_id)
+    WHERE
+      _n = 1
+  ),
+  clients_city_first_seen_firefox_desktop AS (
+    SELECT
+      client_id,
+      first_seen_date AS first_seen_geo_date,
```
lucia-vargas-a: (See question above) - first_seen_date and the city captured on that date vs. first city seen and on which date.

sql_generators/clients_city_seen_v1/templates/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: {{ app_name }} Clients City Seen
+description: |-
+owners:
+  - wichan@mozilla.com
+  - rbaffourawuah@mozilla.com
+labels:
+  change_controlled: true
```
lucia-vargas-a: Requires adding to CODEOWNERS

sql/moz-fx-data-shared-prod/fenix_derived/clients_city_seen_v1/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+friendly_name: fenix Clients City Seen
+description: |-
```
lucia-vargas-a: How to block the backfill of these tables / are there cases where it's enabled?

sql_generators/clients_city_seen_v1/templates/metadata.yaml
```diff
@@ -0,0 +1,10 @@
...
+#scheduling:
```
lucia-vargas-a: How is the last-seen city updated?

sql/moz-fx-data-shared-prod/fenix_derived/clients_city_seen_v1/schema.yaml
BenWu: `app_id` is what we typically call this. You could also add a channel field to be more clear.

sql_generators/clients_city_seen_v1/__init__.py
```diff
@@ -0,0 +1,182 @@
...
+    "  *\n"
+    "FROM\n"
+    "  clients_city_first_seen_{app_name} AS fs\n"
+    "FULL JOIN\n"
+    "  clients_city_last_seen_{app_name} AS ls\n"
+    "USING (client_id)"
+)
+
+
+def clean(s: str) -> str:
```
BenWu: It would be a lot simpler to just have the unions in the template similar to https://github.com/mozilla/bigquery-etl/blob/main/sql_generators/glean_usage/templates/event_monitoring_live_v1.materialized_view.sql

sql_generators/clients_city_seen_v1/__init__.py
```diff
@@ -0,0 +1,182 @@
...
+    # schema templates
+    schema_template = "schema.yaml"
+    # metadata template
+    metadata_template = "metadata.yaml"
+
+    for browser in Browsers:
+
+        if browser.name == "fenix":
+
+            app_names = [
```
BenWu: The app name and id mappings are all in https://probeinfo.telemetry.mozilla.org/v2/glean/app-listings so these don't need to be hard-coded

e.g.
https://github.com/mozilla/bigquery-etl/blob/cb3b49eb30f077afc93def54cbbd6f9f8f116cae/sql_generators/glean_usage/common.py#L179

sql_generators/clients_city_seen_v1/templates/query.sql
```diff
@@ -0,0 +1,135 @@
...
+    mozfun.glean.parse_datetime(ping_info.end_time) AS parsed_end_time,
+    `moz-fx-data-shared-prod.udf.glean_timespan_seconds`( metrics.timespan.glean_baseline_duration ) AS duration,
+    metadata.geo.city,
+    metadata.geo.subdivision1 AS geo_subdivision1,
+    metadata.geo.subdivision2 AS geo_subdivision2,
+  FROM
+    `{{ project_id }}.{{ app_name }}_stable.baseline_v1`
+  WHERE
+    sample_id = 0 ),
+  with_dates_{{ app_name }} AS (
```
BenWu: It would be easier to review if everything that wasn't used in the output was removed

sql/moz-fx-data-shared-prod/fenix_derived/clients_city_seen_v1/schema.yaml
BenWu: It also looks to me like it's taking the last seen date and not the last seen non-null city so it would be good to be clear on this

sql/moz-fx-data-shared-prod/fenix_derived/clients_city_seen_v1/schema.yaml
wwyc: @BenWu could you share more on how I can add a channel field?  Thanks!

sql/moz-fx-data-shared-prod/fenix_derived/clients_city_seen_v1/metadata.yaml
wwyc: Do you know how/if the first seen tables are being backfilled?

sql_generators/clients_city_seen_v1/templates/metadata.yaml
wwyc: This draft PR only relates to initializing the tables using stable tables and start some discussions - it won't be merged.  The ETL part would be the next step where last seen city would be updated using live tables daily.

sql/moz-fx-data-shared-prod/firefox_desktop_derived/clients_city_seen_v1/query.sql
wwyc: `sample_id` was included for now to help with data validations.  Since this code would only be run once and the PR won't be merged, I think it should be ok?

sql/moz-fx-data-shared-prod/fenix_derived/clients_city_seen_v1/schema.yaml
BenWu: You can use the `normalized_channel` field from the baseline table

sql_generators/clients_city_seen_v1/templates/query.sql
wwyc: can you explain more on what this means?  thanks!

sql/moz-fx-data-shared-prod/fenix_derived/clients_city_seen_v1/schema.yaml
wwyc: It was added to avoid confusion with other tables that use the same column names, such as first_seen_date.

kwindau: I agree with @wwyc , I think this is different than first seen date, it's what date where you first seen in that geo, so the distinct name makes sense to me in this case.

sql/moz-fx-data-shared-prod/fenix_derived/clients_city_seen_v1/metadata.yaml
wwyc: This table cannot be backfilled for more than 30 days since that is the maximum retention in live tables.

sql_generators/baseline_clients_city_seen_v1/templates/query.sql
```diff
@@ -0,0 +1,230 @@
...
+    `moz-fx-data-shared-prod.udf.glean_timespan_seconds`( metrics.timespan.glean_baseline_duration ) AS duration,
+    metadata.geo.city,
+    metadata.geo.subdivision1 AS geo_subdivision1,
+    metadata.geo.subdivision2 AS geo_subdivision2,
+  FROM
+    `{{ project_id }}.{{ app_id }}_stable.baseline_v1`
+  WHERE
+    client_info.client_id IS NOT NULL
+    AND sample_id = 0
+    AND DATE(submission_timestamp) <= "2025-08-25"
```
BenWu: Current date is probably more appropriate for the init so it gets everything

```suggestion
    AND DATE(submission_timestamp) <= CURRENT_DATE()
```

sql_generators/baseline_clients_city_seen_v1/templates/metadata.yaml
```diff
@@ -0,0 +1,37 @@
...
+  dag_name: bqetl_clients_city_seen
+  task_name: {{ app_name }}_{{ table_name }}
+  depends_on_past: true
+  date_partition_parameter: null
+  parameters:
+    - {% raw %}submission_date:DATE:{{ds}}{% endraw %}
+  depends_on:
+    - task_id: copy_deduplicate_all
+      dag_name: copy_deduplicate
+      execution_delta: 1h
```
BenWu: This isn't needed because `copy_deduplicate` is to populate the stable tables and this uses live.  Also `execution_delta` would be 3h in this case

BenWu (commented):
You should remove all the generated files in `sql/` before merging because generated sql doesn't need to be checked in.

The is_init query with a 100% sample might be too expensive run in a single query and timeout.  This is fine for the POC but something to prepare for later

sql_generators/baseline_clients_city_seen_v1/templates/metadata.yaml
wwyc: Thanks for the review!

> You should remove all the generated files in sql/ before merging 
> because generated sql doesn't need to be checked in.

I plan to remove all the generated files in sql/ before merging.  I've added them here just as reference/sharing with DS for now.  

> The is_init query with a 100% sample might be too expensive run in > a single query and timeout. This is fine for the POC but something > to prepare for later

How would you recommend to run this to ensure it would not time out?

> This isn't needed because copy_deduplicate is to populate the 
> stable tables and this uses live. Also execution_delta would be 3h 
> in this case

can you explain more on why execeution_delta would be 3h?  Which DAG should this depend on if not copy_deduplicate to ensure the data is available in the live tables?

BenWu: - One way to initialize the tables would be per sample id.  You could write a script that just runs the query per sample id
- copy_deduplicate is scheduled at 1am and bqetl_clients_city_seen is scheduled at 4am so the execution delta is 3 hours
- There's no dag for the live tables so this wouldn't depend on anything.  It would just assume data is complete after some time past midnight UTC.  This is what copy_deduplicate does

wwyc: It looks like if I add @sample_id then the table would be initialized in parallel per sample_id: https://github.com/mozilla/bigquery-etl/blob/59139fd6fb4b84a653833bfe2084ca02f8951a10/bigquery_etl/cli/query.py#L1460

sql_generators/baseline_clients_city_seen_v1/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+
+def get_app_info_map() -> Dict[str, List[str]]:
+    """Build a map of app name to a list of bq_dataset_families."""
+    app_info = get_app_info()
+
+    app_info = [
+        info
+        for name, info in app_info.items()
+        if name
+        not in ConfigLoader.get("generate", "glean_usage", "skip_apps", fallback=[])
```
BenWu: Are we doing this for all apps or just the firefox ones?  I can't remember if that was decided.

You can add a new entry to bqetl_project that's an allowlist of apps and just have whichever ones you're using for testing.  That way you don't have this and also the hardcoded list below.

sql_generators/baseline_clients_city_seen_v1/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+    app_info_map = get_app_info_map()
+    # browser_app_list = ["firefox_desktop", "fenix"]
+    # app_info_map = {k: app_info_map[k] for k in browser_app_list if k in app_info_map}
+
+    # TODO: listing only a few app_ids for testing
+    app_info_map = {
+        "firefox_desktop": ["firefox_desktop"],
+        "fenix": [
+            "org_mozilla_firefox",
+            "org_mozilla_fenix_nightly",
```
BenWu: `org_mozilla_fenix_nightly` is deprecated and `org_mozilla_fenix` is the active nightly app

sql_generators/baseline_clients_city_seen_v1/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+from jinja2 import Environment, FileSystemLoader
+
+from bigquery_etl.cli.utils import use_cloud_function_option
+from bigquery_etl.config import ConfigLoader
+from bigquery_etl.format_sql.formatter import reformat
+from bigquery_etl.util.common import render, write_sql
+from sql_generators.glean_usage import get_app_info
+
+THIS_PATH = Path(os.path.dirname(__file__))
+TABLE_NAME = os.path.basename(os.path.normpath(THIS_PATH))
```
BenWu: nit: It seems redundant to call normpath and then basename.  Also `Path` and `os.path` don't need to be mixed here

```suggestion
THIS_PATH = Path(__file__).parent
TABLE_NAME = THIS_PATH.name
```

sql_generators/baseline_clients_city_seen_v1/templates/query.sql
```diff
@@ -0,0 +1,306 @@
...
+    {% raw %}
+    {% else %}
+    {% endraw %}
+   `{{ project_id }}.{{ app_id }}_live.baseline_v1`
+    {% raw %}
+    {% endif %}
+    {% endraw %}
+  WHERE
+    client_info.client_id IS NOT NULL
+    AND sample_id = 0
```
BenWu: You should pay attention to how long this takes to init with a 1% sample since it blocks artifact deployment during the init.  It should be fine for now but we might need to do the full initial backfill another way.

sql_generators/baseline_clients_city_seen_v1/templates/metadata.yaml
BenWu: This `depends_on` should still be removed since it's reading from the live tables

sql_generators/baseline_clients_city_seen_v1/__init__.py
wwyc: `org_mozilla_fenix_nightly` doesn't seem to be listed in `deprecated_apps` here: https://github.com/mozilla/bigquery-etl/blob/3be374bf0e0705cba238b1b3947838d9df1a937c/bqetl_project.yaml#L348

How do I know which apps should be included in the final list?

sql_generators/baseline_clients_city_seen_v1/__init__.py
wwyc: We are doing this for all apps (~47?) since we discussed that most tables would be small in size.  
However, for the first phase we would only roll out for the browser apps.

I've added an entry to bqetl_project.

sql_generators/baseline_clients_city_seen_v1/__init__.py
BenWu: That isn't the canonical list of deprecated apps, that's just a partial list for https://mozilla-hub.atlassian.net/browse/DENG-9088.  Deprecated app are listed in probe scraper.  In any case, you shouldn't be using `org_mozilla_fenix_nightly` data over `org_mozilla_fenix` for testing since there isn't much data and it's all from very old clients.

In the `get_app_info_map()` function, you can do
```
if info[0].get("deprecated", False) is not True
```
instead of relying on the the glean_usage config.  That would avoid creating more tech debt since more tables need to be cleaned up for DENG-9088

wwyc: org_mozilla_fenix_nightly doesn't seem to be listed as deprecated here:  https://probeinfo.telemetry.mozilla.org/v2/glean/app-listings

BenWu: Yeah it isn't deprecated there but we stopped building and distributing the app over five years ago.  All I'm saying is you shouldn't be using this as one of the test apps

wwyc: Thanks!  I will not use `org_mozilla_fenix_nightly` as one of the test apps.  But I wonder how do I know what my final allow list would contain since i wont be able to filter using `if info[0].get("deprecated", False) is not True`?

BenWu: It will be fine in the final list if you don't want to specifically ignore it.  It's just that for testing, you should include at least one other channel that actually has data.  Including all of them like you're doing now is fine

sql_generators/baseline_clients_city_seen_v1/templates/query.sql
wwyc: I just did an init with a 2% sample and it took:
- firefox_desktop: 319s
- fenix: 111s
However, if I were to add @sample_id then the table would be initialized in parallel per sample_id: https://github.com/mozilla/bigquery-etl/blob/59139fd6fb4b84a653833bfe2084ca02f8951a10/bigquery_etl/cli/query.py#L1460

Would that be sufficient or is there another way to do the full initial backfill that you would recommend?  Thanks!

wwyc:
@BenWu @soGaussian could you do a final review/approval on this PR? I’d like to merge early next week.

@BenWu is there a time of day you recommend for the merge so initialization doesn’t affect artifact deployments? Thanks!

BenWu:
I'll take a closer look on Monday.

Running it towards the end of the day should be fine so it doesn't block any fixes that need to go out.  I didn't see your comment on how long it would take but it should be fine to do the initialization.  I don't think running in parallel would make it much faster since the query is large enough to use up most of the slots. I just did a test and a 1% query uses up all 2k slots in the backfill reservation for basically the whole duration of the query.

One thing that we should do is make the initialization run in the backfill reservation instead of analysis-and-etl so there isn't slot contention with etl.  I can make that change on Monday.

Also Kiran is planning on doing a large backfill of events_first_seen next week so make sure to coordinate with her so the backfills don't end up fighting for slots

wwyc:
> I'll take a closer look on Monday.
> 
> Running it towards the end of the day should be fine so it doesn't block any fixes that need to go out. I didn't see your comment on how long it would take but it should be fine to do the initialization. I don't think running in parallel would make it much faster since the query is large enough to use up most of the slots. I just did a test and a 1% query uses up all 2k slots in the backfill reservation for basically the whole duration of the query.
> 
> One thing that we should do is make the initialization run in the backfill reservation instead of analysis-and-etl so there isn't slot contention with etl. I can make that change on Monday.
> 
> Also Kiran is planning on doing a large backfill of events_first_seen next week so make sure to coordinate with her so the backfills don't end up fighting for slots

Thank you!  I will check with kiran.  The time it takes to run 2% sample: firefox_desktop: ~319s, fenix: ~111s.

sql_generators/baseline_clients_city_seen_v1/templates/metadata.yaml
```diff
@@ -0,0 +1,35 @@
+friendly_name: Baseline Clients City Seen
+description: |-
+  This table stores the first-seen and last-seen geo attributes for each client_id.
+  The table was initialized from stable tables (with ~2 years of retention),
```
BenWu: Bit of a nit but a lot of the other apps have less than two years and this isn't relevant for any apps that are created after this table is created so I think mentioning an approximate date for when this was created would be good.
```suggestion
  The table was initialized from stable tables (with finite retention) in Oct. 2025,
```

sql_generators/baseline_clients_city_seen_v1/templates/query.sql
```diff
@@ -0,0 +1,306 @@
...
+    AND 'automation' NOT IN (
+      SELECT TRIM(t) FROM UNNEST(SPLIT(metadata.header.x_source_tags, ',')) t
+    )
+  QUALIFY
+    ROW_NUMBER() OVER (PARTITION BY document_id ORDER BY submission_timestamp ) = 1
+    {% raw %}
+    {% endif %}
+    {% endraw %}
+   ),
+  with_dates_{{ app_id }} AS (
```
BenWu: I don't think these session date values are used so these two `with_dates` CTEs can be removed

sql_generators/baseline_clients_city_seen_v1/__init__.py
```diff
@@ -0,0 +1,151 @@
...
+    }
+    """
+    app_info = get_app_info()
+    app_info = [
+        info
+        for name, info in app_info.items()
+        if name
+        in ConfigLoader.get(
+            "generate", "baseline_clients_city_seen", "apps", fallback=[]
+        )
```
BenWu, line 42: Assuming the allowlist is just for this initial phase, when you add the rest of the apps, you'll also need to check if the app has baseline pings since glean.js and server glean doesn't use them. That doesn't need to be done now but might be needed later.

One way to do that would be (I didn't test this):
```py
"glean-core" in requests.get(
  f"https://probeinfo.telemetry.mozilla.org/glean/{info[0]['v1_name']}/dependencies"
).json()
```
glean-usage downloads the schemas but that's slower if you only need to check baseline.

bqetl_project.yaml
```diff
@@ -605,3 +608,5 @@ retention_exclusion_list:
 - sql/moz-fx-data-shared-prod/firefox_ios_derived/client_adclicks_history_v1
 - sql/moz-fx-data-shared-prod/acoustic_external/suppression_list_v1
 - sql/moz-fx-data-shared-prod/telemetry_derived/fx_accounts_linked_clients_v1
+- sql/moz-fx-data-shared-prod/firefox_desktop_derived/baseline_clients_city_seen_v1
+- sql/moz-fx-data-shared-prod/fenix_derived/baseline_clients_city_seen_v1
```
BenWu, line 612: Does `retention_exclusion_list` support something like `sql/moz-fx-data-shared-prod/**/baseline_clients_city_seen_v1`?  Seems inconvenient to need to add each app

wwyc: It doesn't look like it is supported.

soGaussian (commented):
Looks  to me

wwyc:
@BenWu if i can get an approval from your end perhaps I can merge today?  Or would you prefer to merge this one?  Thanks for the thorough review!

BenWu (approved):
You'll need to delete the existing `baseline_clients_city_seen_v1` tables to make the initialization run.  You can merge this towards the end of the day and make a note in the data platform slack channel that artifact deployment will be blocked for a few hours

PR #7979, opened 2025-08-21
feat(DENG-9303): enterprise and consumer segment metrics

kik-kik:
# feat(DENG-9303): enterprise and consumer segment metrics

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_desktop/enterprise_and_consumer_metric_segment_aggregates/view.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_desktop/enterprise_and_consumer_metric_segment_aggregates/view.sql
@@ -2,6 +2,33 @@ CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.firefox_desktop.enterprise_and_consumer_metric_segment_aggregates`
 AS
 SELECT
-  *
+  *,
+  CASE
+    WHEN normalized_channel = "release"
+      AND ((policies_count > 1) OR (policies_is_enterprise = TRUE))
+      THEN "enterprise_release"
+    WHEN normalized_channel = "release"
+      AND (
+        (distribution_id IS NOT NULL)
+        OR (policies_count = 0)
+        OR (policies_is_enterprise = FALSE)
+      )
+      THEN "consumer_release"
+    WHEN normalized_channel = "esr"
+      AND ((policies_count > 1) AND (distribution_id IS NULL))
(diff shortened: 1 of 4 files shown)


sql/moz-fx-data-shared-prod/firefox_desktop_derived/enterprise_and_consumer_metric_segment_clients_v1/query.sql
```diff
@@ -0,0 +1,61 @@
...
+    ALL
+)
+SELECT
+  @submission_date AS submission_date,
+  client_id,
+  sample_id,
+  normalized_channel,
+  distribution_id,
+  is_dau,
+  -- TODO: it appears sometimes those two values can be null in the result,
```
kik-kik: @pissac17 How should we handle nulls here?

sql/moz-fx-data-shared-prod/firefox_desktop_derived/enterprise_and_consumer_metric_segment_clients_v1/query.sql
```diff
@@ -0,0 +1,61 @@
...
+    client_id,
+    normalized_channel AS channel,
+    ARRAY_AGG(distribution_id IGNORE NULLS ORDER BY submission_date DESC)[
+      SAFE_OFFSET(0)
+    ] AS distribution_id,
+  FROM
+    `moz-fx-data-shared-prod.firefox_desktop.baseline_clients_daily`
+  WHERE
+    submission_date
+    BETWEEN DATE_SUB(@submission_date, INTERVAL 27 DAY)
```
kik-kik: @pissac17 does this window need to be this wide or could we make it smaller?

pissac17: @kik-kik is the width impacting overall runtime? If it isn’t a material cost, I’d avoid shrinking it too much. In my tests, going below ~21 days increases the unknown_*rates. Also, we’re only looking back for distribution_id. I think we should add a similar lookback for policies_count and policies_is_enterprise

sql/moz-fx-data-shared-prod/firefox_desktop_derived/enterprise_and_consumer_metric_segment_clients_v1/query.sql
pissac17: Let's leave them as null

PR #7980, opened 2025-08-21
chore(mdn_popularities): incorporate mdn_fred Glean data

caugner:
## Description

MDN launched a new front end (fred), replacing the old one (yari).

This migrates the query of the `mdn_popularities_v1` job to incorporate the `mdn_fred` Glean dataset.

Query tested in Redash: https://sql.telemetry.mozilla.org/queries/110149/source?p_date=d_now

> [!NOTE]
> This job runs on the 1st day of every month, and queries the page view data for the previous month.
> This means the data for August will be queried on September 1st.
>
> As we cut over on August 19th, data for August is split across the `mdn_yari` or `mdn_fred` Glean datasets.
> Once the September 1st run succeeded, we can update the query to use only `mdn_fred`.

Fixes https://github.com/mdn/fred/issues/526.


LeoMcA (approved):
I'd suggest simplifying this query to just use the fred data: half a month's data is still enough of a sample size, and it means we don't have to update this query again

caugner:
> I'd suggest simplifying this query to just use the fred data: half a month's data is still enough of a sample size, and it means we don't have to update this query again

Let's land as is.

The published page views data are absolute numbers, not relative, so this is a good reason to use the full data for August.

PR #7988, opened 2025-08-22
Pocket component table

jaredsnyder:
## Description

This PR creates a table for the content component of the newtab ping.  Idea was to include only tiles specific to the component interaction, and any additional metadata about the visit can be obtained from the visit-level table.

* DENG-8971

## Checklist for review

- [ ] Ensure the validation done in [this notebook](https://colab.research.google.com/drive/1Xk_oo5Zc50tNzh-g7V7HOeC2dFNABJa8#scrollTo=rgxKmr2dV7GL&uniqifier=1) is sufficient
- [ ] Check that included columns are sufficient for analytical needs, keeping in mind metadata about the visit itself can be obtained from the visits-level table
- [ ] validate query and schema

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_desktop/newtab_component_content/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/firefox_desktop/newtab_component_content/metadata.yaml
@@ -1,0 +1,6 @@
+friendly_name: Newtab Component-level Content Table
+description: Contains information about content interactions, both organic and sponsored,
+  from the newtab ping.  Each record represents a specific content position during a specific visit
+owners:
+  - jsnyder@mozilla.com
+  - gkatre@mozilla.com
--- a/sql/moz-fx-data-shared-prod/firefox_desktop/newtab_component_content/view.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_desktop/newtab_component_content/view.sql
@@ -1,0 +1,3 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.firefox_desktop.newtab_component_context`
+AS
--- a/sql/moz-fx-data-shared-prod/firefox_desktop/newtab_component_content/view.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_desktop/newtab_component_content/view.sql
@@ -1,3 +1,7 @@
 CREATE OR REPLACE VIEW
   `moz-fx-data-shared-prod.firefox_desktop.newtab_component_context`
 AS
+SELECT
+  *
+FROM
+  `moz-fx-data-shared-prod.firefox_desktop_derived.newtab_component_context_v1`
--- a/sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_component_content_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_component_content_v1/metadata.yaml
@@ -1,0 +1,28 @@
+friendly_name: Newtab Component-level Content Table
+description: Contains information about content interactions, both organic and sponsored,
+  from the newtab ping.  Each record represents a specific content position during a specific visit
+owners:
+  - jsnyder@mozilla.com
+  - gkatre@mozilla.com
+labels:
+  application: newtab
+  incremental: true
+  schedule: daily
+  dag: bqetl_newtab
+  owner1: gkatre
+  table_type: client_level
+scheduling:
+  dag_name: bqetl_newtab
+bigquery:
+  time_partitioning:
+    type: day
+    field: submission_date
+    require_partition_filter: true
+  clustering:
+    fields:
+      - channel
+      - country
+      - newtab_category
+    expiration_days: 775
+  range_partitioning: null
+references: {}
--- a/sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_component_content_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_component_content_v1/query.sql
@@ -56,7 +56,7 @@ raw_content_info AS (
     CAST(mozfun.map.get_key(event_details, 'is_sponsored') AS BOOLEAN) AS is_sponsored,
     mozfun.map.get_key(event_details, 'format') AS format,
     CAST(LOGICAL_OR(event_name = 'impression') AS INT) AS impression_count,
-    COUNTIF(event_name = 'click') AS clicks_count,
+    CAST(LOGICAL_OR(event_name = 'click') AS INT) AS clicks_count,
     CAST(LOGICAL_OR(event_name = 'dismiss') AS INT) AS dismiss_count,
     CAST(
       LOGICAL_OR(
--- a/sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_component_content_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_component_content_v1/query.sql
@@ -1,0 +1,113 @@
+WITH visit_window_width AS (
+  SELECT
+    MIN(
+      IF(
+        category = 'newtab'
+        AND name = "opened",
+        SAFE_CAST(mozfun.map.get_key(extra, "window_inner_width") AS INT),
+        NULL
+      )
+    ) AS newtab_window_inner_width,
+    mozfun.map.get_key(extra, 'newtab_visit_id') AS newtab_visit_id
+  FROM
+    `moz-fx-data-shared-prod.firefox_desktop_stable.newtab_v1`,
+    UNNEST(events)
+  WHERE
+    DATE(submission_timestamp) = '2025-06-01'
+  GROUP BY
+    newtab_visit_id
+),
+pocket_events_unnested AS (
+  SELECT
+    DATE(submission_timestamp) AS submission_date,
+    sample_id,
+    client_info.client_id AS client_id,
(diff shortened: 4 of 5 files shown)


sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_component_content_v1/metadata.yaml
```diff
@@ -0,0 +1,28 @@
...
+  from the newtab ping.  Each record represents a specific content position during a specific visit
+owners:
+  - jsnyder@mozilla.com
+  - gkatre@mozilla.com
+labels:
+  application: newtab
+  incremental: true
+  schedule: daily
+  dag: bqetl_newtab
+  owner1: gkatre
```
gkatre:
```suggestion
  owner1: jsnyder
```
@jaredsnyder great work! But you dont need to make me the owner 🙂

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_component_content_v1/query.sql
```diff
@@ -0,0 +1,117 @@
...
+  FROM
+    pocket_events_unnested
+  GROUP BY
+    ALL
+),
+-- the purpose of the final CTEs is to apply the necessary UDFs to get rownumber
+content_and_visit_info AS (
+  SELECT
+    raw_content_info.*,
+    mozfun.newtab.determine_grid_layout_v1(
```
gkatre: If i were doing this, would prefer to apply such udfs in the view. This also applies to `mozfun.newtab.determine_tiles_per_row_v1`.
Similar to what I have done [here](https://github.com/mozilla/bigquery-etl/blob/ce07714267e39f3b98596d620ee47cb9e352654d/sql/moz-fx-data-shared-prod/firefox_desktop/newtab_visits_daily/view.sql)

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_component_content_v1/metadata.yaml
```diff
@@ -0,0 +1,30 @@
+friendly_name: Newtab Component-level Content Table
+description: |-
+    Contains information about content interactions, both organic and sponsored,
+    from the newtab ping.  Each record represents a specific content position during a specific visit.
+
+    Note that this is pulling data from the newtab ping and not the newtab_content ping
+owners:
+  - jsnyder@mozilla.com
+  - gkatre@mozilla.com
```
gkatre:
```suggestion
  - jsnyder@mozilla.com
```
Looks like you may have removed the `owner1` tag completely. You may want to add that and set this with your email. Perhaps the same applies for the view

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_component_content_v1/metadata.yaml
jaredsnyder: Sure thing, I put `owner1` back in and put myself as the owner!

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_component_content_v1/query.sql
jaredsnyder: Good suggestion, I'll do that

sql/moz-fx-data-shared-prod/firefox_desktop/newtab_component_content/metadata.yaml
```diff
@@ -0,0 +1,6 @@
+friendly_name: Newtab Component-level Content Table
+description: Contains information about content interactions, both organic and sponsored,
+  from the newtab ping.  Each record represents a specific content position during a specific visit
+owners:
+  - jsnyder@mozilla.com
+  - gkatre@mozilla.com
```
gkatre:
```suggestion
  - jsnyder@mozilla.com
```

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_component_content_v1/query.sql
```diff
@@ -0,0 +1,83 @@
...
+)
+-- aggregate metrics to the visit_id level
+SELECT
+  submission_date,
+  sample_id,
+  country,
+  app_version,
+  client_id,
+  newtab_visit_id,
+  newtab_window_inner_width,
```
gkatre, line 50: > Idea was to include only tiles specific to the component interaction, and any additional metadata about the visit can be obtained from the visit-level table.

The read from the [Description](https://github.com/mozilla/bigquery-etl/pull/7988#issue-3346786592) appears to suggest that non-component specific metadata is going to be obtained from the newtab visit table. This makes sense because duplicating the non-component level fields causes unnecessary redundancy.
To that effect, perhaps the only additional fields needed would be `newtab_visit_id` and `submission_date` (those needed for the join with the visits table)?

sql/moz-fx-data-shared-prod/firefox_desktop/newtab_component_content/view.sql
```diff
@@ -0,0 +1,44 @@
...
+  FROM
+    raw_content_info
+),
+add_tiles_per_row AS (
+  SELECT
+    content_and_visit_info.*,
+    mozfun.newtab.determine_tiles_per_row_v1(
+      layout_type,
+      newtab_window_inner_width
+    ) AS num_tiles_per_row
```
gkatre, line 28: If you plan to use the visits view `firefox_desktop.newtab_visits_daily`, which in my opinion is a better approach, the udfs used here would be available and recomputing these are not necessary.

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_component_content_v1/query.sql
jaredsnyder: I wanted to try to keep the creation of the view/table separate from `newtab_visits_daily` in case we take action on the proposal to populate `newtab_visits_daily_v1` with the component table, I didn't want to create a circular dependency

sql/moz-fx-data-shared-prod/firefox_desktop/newtab_component_content/view.sql
jaredsnyder: See: https://github.com/mozilla/bigquery-etl/pull/7988#discussion_r2302122238

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_component_content_v1/query.sql
gkatre: > I wanted to try to keep the creation of the view/table separate from `newtab_visits_daily` in case we take action on the proposal to populate `newtab_visits_daily_v1` with the component table, I didn't want to create a circular dependency

What you are thinking involves a major design change and may involve changing what you may be doing here. Therefore, not relevant to this PR

jaredsnyder: That's fair, I'd still prefer to keep them independent for now.  Are there serious downsides I might be missing?  I know it is less memory efficient, but is it so much so that this is inadvisable?

gkatre: Its good to keep them (current PR and proposal that's still in an ideation phase) independent. If we are just talking about this PR's scope then perhaps we do not need to materialize any redundant fields?

jaredsnyder: Oh I meant I wanted to keep `newtab_visits_daily` and this table independent, so I'd like to keep `newtab_window_inner_width`, `experiments` and `app_version` for this reason.

I wanted `sample_id` because sometimes it's useful to be able to reduce the size, and `country` is one of the most common to filter on but it could probably be removed. `experiments` is another one that seemed like it'd be used commonly enough that it'd be worth keeping.  @ilanasegall do you have opinions on which fields to keep, trying to balance making this table easy to use and not including too many fields that could be obtained by joining to the visits table.?

gkatre: If this is meant to be a quick PR to serve your and @ilanasegall immediate use cases, then this PR perhaps does not need to go through an elaborate review process.

I would however recommend that a proper data validation done before its deployed and used in any mainstream applications

jaredsnyder: Happy to take feedback on what you think the proper approach to data validation is here.  The [notebook](https://colab.research.google.com/drive/1Xk_oo5Zc50tNzh-g7V7HOeC2dFNABJa8) I created was what I've done so far.  Because there is so little processing and no existing source of truth to compare against, checking the handling of "edge cases" seemed like a reasonable approach.  I also check that the queries run and looked over the results to ensure there wasn't any obvious errors.

Let me know what else you'd like to see done, I'd rather spend a bit more time here and get it right the first time than have to do a follow-up PR to fix avoidable mistakes!

gkatre: > Happy to take feedback on what you think the proper approach to data validation is here.

The visit level table that we did in H1, Nadia had done an extensive data validation which I believe she also documented somewhere to share with other DS folks. Since you are also going to be the consumer of the data, I would let you decide what kind of validation you would like to perform. Would also be good to share with other DS folks for their approval.

PR #7998, opened 2025-08-26
Fix new schemas not being created

BenWu:
## Description

`bqetl query schema update` no longer creates `schema.yaml` if it doesn't already exist since https://github.com/mozilla/bigquery-etl/pull/7807.  This means new tables with no `schema.yaml` wouldn't get deployed.  This affects some glean_usage tables but there haven't been any new apps in the past month

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -2303,20 +2303,20 @@ def _update_query_schema(
     )
 
     changed = True
-    updated_schema = None
 
     if existing_schema_path.is_file():
         existing_schema = Schema.from_schema_file(existing_schema_path)
         old_schema = copy.deepcopy(existing_schema)
         if table_schema:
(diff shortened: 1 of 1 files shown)


bigquery_etl/cli/query.py
```diff
@@ -2303,29 +2303,30 @@ def _update_query_schema(
...
 
     changed = True
-    updated_schema = None
 
     if existing_schema_path.is_file():
         existing_schema = Schema.from_schema_file(existing_schema_path)
         old_schema = copy.deepcopy(existing_schema)
         if table_schema:
-            updated_schema = existing_schema.merge(table_schema)
+            existing_schema.merge(table_schema)
```
BenWu, line 2311: `Schema.merge` doesn't return anything

kwindau (approved):

PR #8001, opened 2025-08-26
feat: creates a dimension model view of the current state of snowflake_migration_derived.corpus_items_updated

LiamMcFall:
## Description

This PR builds a dim view of the current state of s`nowflake_migration_derived.corpus_items_updated` records.

* DENG-9057

### DIFF
--- a/sql/moz-fx-data-shared-prod/snowflake_migration_derived/corpus_items_current_v1/view.sql
+++ b/sql/moz-fx-data-shared-prod/snowflake_migration_derived/corpus_items_current_v1/view.sql
@@ -8,7 +8,7 @@ WITH corpus_item_id_row_nums AS (
     url,
     authors,
     publisher,
-    reviewed_corpus_item_updated_at AS corpus_item_id_updated_at,
+    reviewed_corpus_item_updated_at AS corpus_item_updated_at,
     ROW_NUMBER() OVER (
       PARTITION BY
         approved_corpus_item_external_id


LiamMcFall:
@gkatre To address some of your questions from slack:

```
which attributes that you plan to include in the dim table could be slowly changing?
if any of them are slowly changing, what is the expected assignments in the joins. Meaning, which change would you be applying?
```
`title` and `publisher` look like the 2 dimensions being pulled in that are slowly changing. This shouldn't have any affect on the join that we are doing downstream in the newtab-content modeling. Because we are creating a table that treats `approved_corpus_item_id` as unique, this table will actually fix the row explosion currently happening in the `newtab_items_daily` table.

```
How would the ETL look like when a new update comes in with new values for an existing corpus_item that already exist in the dim table. Most of the ETL daily updates we do today on Fact tables are mainly inserts. The only time we perform updates on previous data is during backfills (which is really dropping the day partitions and re inserting the rows)
```
The way these fields end up being used in this [looker report](https://mozilla.cloud.looker.com/dashboards/2205?Submission+Date=30+day&Scheduled+Surface+ID=%22NEW_TAB_EN_US%22), they are only used to look at the interactions for a `corpus_item_id` at the granularity of the `submission_date`. Because of this, I think we can be okay with either of the SCD being _slightly_ different on different submission dates as it won't impact the way the reporting is being used since these models will run once a day.

sql/moz-fx-data-shared-prod/snowflake_migration_derived/corpus_items_current_v1/view.sql
```diff
@@ -0,0 +1,26 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.snowflake_migration_derived.corpus_items_current_v1`
+AS
+WITH corpus_item_id_row_nums AS (
+  SELECT
+    approved_corpus_item_external_id,
```
gkatre:
```suggestion
    corpus_item_id,
```
This is the correct field name

sql/moz-fx-data-shared-prod/snowflake_migration_derived/corpus_items_current_v1/metadata.yaml
```diff
@@ -0,0 +1,7 @@
+friendly_name: Corpus Items Current V1
+description: |-
+  This table deduplicates and surfaces the most recent version of each corpus item from the
+  corpus_items_updated table. In effect, this table converts the fact table of updates into a
+  dimension table of the most current state.
```
gkatre: Perhaps make it evident that its a _view_ of the transactional updates to the corpus items that picks the most recent updates to the few fields listed in the schema definition

sql/moz-fx-data-shared-prod/snowflake_migration_derived/corpus_items_current_v1/view.sql
```diff
@@ -0,0 +1,26 @@
...
+  `moz-fx-data-shared-prod.snowflake_migration_derived.corpus_items_current_v1`
+AS
+WITH corpus_item_id_row_nums AS (
+  SELECT
+    approved_corpus_item_external_id AS corpus_item_id,
+    title,
+    url,
+    authors,
+    publisher,
+    reviewed_corpus_item_updated_at AS corpus_item_id_updated_at,
```
gkatre:
```suggestion
    reviewed_corpus_item_updated_at AS corpus_item_updated_at,
```

PR #8016, opened 2025-08-28
feat(DENG-8022): update new_downloads and redownloads and total_downloads field references to use the new tap_ prefix

kik-kik:
# feat(DENG-8022): update new_downloads and redownloads and total_downloads field references to use the new tap_ prefix


sql/moz-fx-data-shared-prod/apple_ads_external/ad_group_report_v1/query.sql
```diff
@@ -11,8 +11,8 @@ SELECT
   start_at,
   end_at,
   taps,
-  new_downloads,
-  redownloads,
+  tap_new_downloads AS new_downloads,
+  tap_redownloads AS redownloads,
```
sean-rose, line 15: I thought maybe the the new columns should be coalesced with the old columns to cover historical records, but it looks like [the dbt models already do that](https://github.com/fivetran/dbt_apple_search_ads/blob/ba2f18e161fb904c192a7268ed69647999032920/models/staging/stg_apple_search_ads__ad_group_report.sql#L39-L41), so there's no need to duplicate that logic here.

sql/moz-fx-data-shared-prod/apple_ads_external/ad_group_report_v1/query.sql
```diff
@@ -11,8 +11,8 @@ SELECT
...
   total_downloads,
```
sean-rose: [This comment in the dbt model code](https://github.com/fivetran/dbt_apple_search_ads/blob/ba2f18e161fb904c192a7268ed69647999032920/models/apple_search_ads__ad_group_report.sql#L49-L50) says `total_downloads` will be deprecated and `tap_total_downloads` should be used instead.
```suggestion
  tap_total_downloads AS total_downloads,
```
(and the same goes for the other tables)

kik-kik: I might be misunderstanding the comment here, but that's exactly what I am doing in this PR? I'm updating the references to use the new fields names, I'm just aliasing them to the old names in our queries so that downstream queries not need to be updated.

kik-kik: ok, never mind. I see what you mean now, I just missed this specific field. Thanks for spotting that.

sql/moz-fx-data-shared-prod/apple_ads_external/ad_group_report_v1/query.sql
kik-kik: ok, I made sure all `total_downloads` references were updated to `tap_total_downloads`.

sean-rose (approved):

PR #8019, opened 2025-08-28
feat(DENG-9518): Create new daily exchange rate table

kwindau:
## Description

This PR creates the new table and view: 
- `moz-fx-data-shared-prod.external_derived.fred_daily_exchange_rates_v1`
- `moz-fx-data-shared-prod.external.fred_daily_exchange_rates`

* [DENG-9518](https://mozilla-hub.atlassian.net/browse/DENG-9518)

[DENG-9518]: https://mozilla-hub.atlassian.net/browse/DENG-9518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sql/moz-fx-data-shared-prod/external_derived/fred_daily_exchange_rates_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/external_derived/fred_daily_exchange_rates_v1/metadata.yaml
@@ -1,6 +1,7 @@
 friendly_name: FRED Daily Exchange Rates
 description: |-
   Daily currency exchange rates from the Federal Reserve Bank of St. Louis
+  Does not contain data every day, for example, no data available on weekends &  holidays
 owners:
 - kwindau@mozilla.com
 labels:
--- a/sql/moz-fx-data-shared-prod/external_derived/fred_daily_exchange_rates_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/external_derived/fred_daily_exchange_rates_v1/metadata.yaml
@@ -1,0 +1,24 @@
+friendly_name: Fred Daily Exchange Rates
+description: |-
+  Daily currency exchange rates from the Federal Reserve Bank of St. Louis
+owners:
+- kwindau@mozilla.com
+labels:
+  incremental: true
+  owner1: kwindau
+scheduling:
+  dag_name: bqetl_fred_daily_exchange_rate
+  date_partition_parameter: submission_date
+  arguments: ["--date", "{{ds}}"]
+  secrets:
+  - deploy_target: FRED_API_KEY
+    key: bqetl_fred_daily_exchange_rate__api_key
+bigquery:
+  time_partitioning:
+    type: day
+    field: submission_date
+    require_partition_filter: false
+    expiration_days: null
+  range_partitioning: null
+references: {}
+require_column_descriptions: true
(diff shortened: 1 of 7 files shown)


sql/moz-fx-data-shared-prod/external_derived/fred_daily_exchange_rates_v1/query.py
```diff
@@ -0,0 +1,103 @@
...
+
+# Define inputs
+exchange_rates = {
+    "AU": ["DEXUSAL", "USD", "AUD"],  # USD to AUD
+    "CA": ["DEXCAUS", "CAD", "USD"],  # CAD to USD
+    "CH": ["DEXSZUS", "CHF", "USD"],  # CHF to USD
+    "JP": ["DEXJPUS", "JPY", "USD"],  # JPY to USD
+    "GB": ["DEXUSUK", "USD", "GBP"],  # USD to GBP
+    "EU": ["DEXUSEU", "USD", "EUR"],  # USD to EUR
+}
```
sean-rose: Is there some reason why the direction of exchange rate conversions isn't consistent?  I'd have expected these to either all be to USD or all be from USD, depending on the intended use case.

sql/moz-fx-data-shared-prod/external_derived/fred_daily_exchange_rates_v1/metadata.yaml
```diff
@@ -0,0 +1,24 @@
+friendly_name: Fred Daily Exchange Rates
```
sean-rose:
```suggestion
friendly_name: FRED Daily Exchange Rates
```

sql/moz-fx-data-shared-prod/external_derived/fred_daily_exchange_rates_v1/query.py
```diff
@@ -0,0 +1,103 @@
...
+
+    # Append the data to the BQ table
+    job_config = bigquery.LoadJobConfig(
+        write_disposition=bigquery.WriteDisposition.WRITE_APPEND
+    )
+
+    # Execute the load
+    load_job = client.load_table_from_dataframe(
+        results_df, table_id, job_config=job_config
+    )
```
sean-rose: It'd be better to do the update atomically, either within a transaction, or by loading the data into a specific table partition with the write-truncate disposition.

sean-rose (commented):
We already have an ETL for IMF exchange rates, so I'm curious, why this particular additional source of exchange rates?

kwindau:
> We already have an ETL for IMF exchange rates, so I'm curious, why this particular additional source of exchange rates?

It's a good question. The existing IMF is a monthly report of exchange rates, whereas this is daily-ish (i.e. weekdays / non-holidays).  The DS wanted this to make their daily level forecasting of revenue more accurate.

sql/moz-fx-data-shared-prod/external_derived/fred_daily_exchange_rates_v1/query.py
kwindau: That's a good question - I will check with the DS who requested this if he intended that way or not, I just went off the list of which series he wanted.

kwindau: I agree though, I think I should make them all the same direction - that way the logic is easier when anyone uses this table. Will change it - thanks @sean-rose

kwindau: Ugh this particular API only offers it in certain directions, so I will add logic to make them all consistent. Good catch!

PR #8024, opened 2025-08-29
fixed row_number calc to handle sections with multiple rows

jaredsnyder:
This PR fixes the row_number calculation to account for sections with multiple rows in the sections layout.  The idea is that each card has a size (`tile_size` in the query) associated with it.  The sum of a the sizes of the cards in a row will sum to `num_tiles_per_row`.  This can be used to determine the row number with a window function.

To validate, plotted the position against the row_number for `num_tiles_per_row` of 3 and 4.  This fits with the most common section layouts, where the first row 3 has three medium cards and 4 has one large card and two mediums.  The same plot broken out by card layout is shown to convey this context

image

image

Same plots for grid layout, which shows expected behavior

image


sql/moz-fx-data-shared-prod/firefox_desktop/newtab_component_content/view.sql
```diff
@@ -25,20 +25,53 @@ add_tiles_per_row AS (
...
-        -- the row number is the same as the section position in sections
-        THEN section_position
-        -- for grid, divide the postition by the number of tiles per row and take the floor
-      ELSE FLOOR(position / num_tiles_per_row)
-    END AS INT
+  -- COALESCE with 0 because first element will return NULL due to range filter in partition
+  -- Will always be position 0
+  COALESCE(
+    SAFE_CAST(
+      FLOOR(
```
chelseyklein, line 56: @jaredsnyder mind adding a comment re: why we use floor here? seems like the only fraction we would have is .5 - i'd personally like to understand rounding down in that instance :)

chelseybeck (approved):
approved pending comment at line 56 🙌

sql/moz-fx-data-shared-prod/firefox_desktop/newtab_component_content/view.sql
jaredsnyder: The window function sums up all the `tile_size` of the tiles before it, and tile_size is defined such that a row will sum up to a multiple of `num_tiles_per_row`.  If your tile is in the middle of the row, when you divide the sum by `num_tiles_per_row` you get a non-integer value, and you want to round down to get the row that you are on.  IE you get .5 if you're in the middle of row 0,  If you get 1.5 if you're in the middle of row 1.

chelseyklein: ty!

PR #8030, opened 2025-08-29
HNT-665: Add private new tab pings engagement

mmiermans:
## Description
Upgrade `bqetl_merino_newtab_extract_to_gcs` to:
1. Combine clicks and impressions from the legacy and private pings.
2. Include `report_count`, which is only present as a private ping.

* [HNT-665](https://mozilla-hub.atlassian.net/browse/HNT-665)


sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_extract_v3/query.sql
```diff
@@ -8,6 +8,34 @@ WITH deduplicated_pings AS (
     `moz-fx-data-shared-prod.firefox_desktop_live.newtab_v1`
   WHERE
     submission_timestamp > TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 1 DAY)
+),
+private_pings AS (
+  SELECT
+    submission_timestamp,
+    document_id,
+    events,
+    normalized_country_code
```
rolf-moz: Have you verified the country code here? This may be somewhat incorrect... because we use the OHTTP server you should rely on the 'country' field in the private ping header, which at this case is constrained to US Canda, or not set for other countries.

mmiermans: Do you know off hand whether this field is normalized by the client? If we're not sure, I'll add some normalization in SQL. If it's not already normalized, my concern is that the GCS blob size would spike from thousands of unique country codes. It looks OK so far:
```
code	impressions
NULL	13947
US	    6314
CA	    916
DE	    451
AT	    75
NI	    48
PL	    40
BD	    32
KH	    16
AU	    9
CH	    6
SI	    3
```
We do have a protection in Merino that would avoid loading a big GCS blob and send an alert to our Slack channel.

rolf-moz: Yes, because we are trying to eliminate any means to identify the user, the country should be restricted to the countries available in the surface id, and otherwise it will be empty. It's only set up for EN DE surface now but will expand. See https://searchfox.org/firefox-main/rev/cd6acbe9eaa55fb4da4fbdec275db6eddd4833b8/browser/extensions/newtab/lib/TelemetryFeed.sys.mjs#87

sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_extract_v3/query.sql
```diff
@@ -32,36 +60,43 @@ flattened_newtab_events AS (
...
+      OR TIMESTAMP_MILLIS(
+        SAFE_CAST(mozfun.map.get_key(unnested_events.extra, 'recommended_at') AS INT64)
+      ) > TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 1 DAY)
+    )
 ),
-/* Aggregate clicks and impressions by corpus_item_id and normalized_country_code. */
+/* Aggregate clicks, impressions, and reports by corpus_item_id and normalized_country_code. */
 aggregated_events AS (
   SELECT
     fe.corpus_item_id,
```
rolf-moz, line 69: Does this drop the event if corpus_item_id is null or missing? Since the other ping also has the same event, you might want to make sure that you aren't aggregating a lot of events with null corpus_item_id

rolf-moz (approved):

sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_extract_v3/query.sql
```diff
@@ -32,36 +60,43 @@ flattened_newtab_events AS (
...
     -- Keep only rows with a non-null corpus_item_id
     AND mozfun.map.get_key(unnested_events.extra, 'corpus_item_id') IS NOT NULL
-    -- Only keep the last day's data.
-    AND TIMESTAMP_MILLIS(
-      SAFE_CAST(mozfun.map.get_key(unnested_events.extra, 'recommended_at') AS INT64)
-    ) > TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 1 DAY)
+    -- Only keep the last day's data (reports don't have recommended_at, so make it optional)
+    AND (
+      -- Report events lack recommended_at timestamp, so we include ALL reports here.
+      -- They'll be filtered later to only keep reports for content with recent impressions/clicks.
```
rolf-moz: @mmiermans I was testing the private ping only and found an issue with this section.
The private ping does not have recommended_at

sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_extract_v3/query.sql
```diff
@@ -73,21 +108,33 @@ country_aggregates AS (
...
     impression_count,
-    click_count
+    click_count,
+    report_count
   FROM
     aggregated_events
   WHERE
     -- Gather country (a.k.a. region) specific engagement for all countries that share a feed.
     -- https://mozilla-hub.atlassian.net/wiki/x/JY3LB
     normalized_country_code IN ('US', 'CA', 'DE', 'CH', 'AT', 'BE', 'GB', 'IE')
```
rolf-moz, line 106: Quite a few events in the private ping in the current experiment have have the country code as NULL FYI. I'm not sure if we need to address for the experiment or not by forcing to US.

sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_extract_v3/query.sql
mmiermans: [newtab_content_click](https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/newtab_content_click) and [newtab_content_impression](https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/newtab_content_impression) both mention `recommended_at`. Is it not provided?

sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_extract_v3/query.sql
mmiermans: Yes, line 39 drops events without a corpus_item_id:
```
    AND mozfun.map.get_key(unnested_events.extra, 'corpus_item_id') IS NOT NULL
```

sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_extract_v3/query.sql
mmiermans: I noticed the same; only about 37% coverage. There may be a bug on the front-end, because the client can only receive Sections if it knows itself to be in the US.

rolf-moz: Now that the data has been flowing in higher volume, I don't see an issue anymore!

PR #8031, opened 2025-09-01
feat(DENG-9250): update firefox app store territory report to include new connector data via v2

kik-kik:
# feat(DENG-9250): update firefox app store territory report to include new connector data via v2


kik-kik:
We decided to stick with the approach from: https://github.com/mozilla/bigquery-etl/pull/7971

PR #8038, opened 2025-09-02
Bug 1986081 - Fix events_daily generator

akkomar:
Quick workaround to stop this query failing in some cases.
Note: this is eventually [planned for removal ](https://mozilla-hub.atlassian.net/browse/DENG-7638) so I didn't spend too much time optimizing it.

### DIFF
--- a/sql_generators/events_daily/templates/events_daily_v1/query.sql
+++ b/sql_generators/events_daily/templates/events_daily_v1/query.sql
@@ -64,12 +64,28 @@ joined AS (
   INNER JOIN
     {{ dataset }}.event_types event_types
     USING (category, event)
+),
+grouped AS (
+  -- Workaround for this query failing with `Cannot query rows larger than 100MB limit` error on some days.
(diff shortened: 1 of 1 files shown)


sql_generators/events_daily/templates/events_daily_v1/query.sql
```diff
@@ -64,12 +64,28 @@ joined AS (
   INNER JOIN
     {{ dataset }}.event_types event_types
     USING (category, event)
+),
+grouped AS (
+  -- Workaround for this query failing with `Cannot query rows larger than 100MB limit` error on some days.
+  -- Example is https://bugzilla.mozilla.org/show_bug.cgi?id=1986081 where we didn't have any aggregated rows
+  -- larger than 100MB, but query still failed.
+  -- It seems like adding this CTE forces a simpler, more stable query plan which succeeds.
```
sean-rose, line 72: It might be worth reporting this to Google support for investigation, as that seems like a BigQuery bug.

PR #8046, opened 2025-09-03
DENG-8767: Create events first seen table

kbammarito:
## Description

This PR adds Events First Seen into the Glean Usage SQL generator.

* [DENG-8767](https://mozilla-hub.atlassian.net/browse/DENG-8767)
* [Planning Doc](https://docs.google.com/document/d/1sk9ErviL5dfVZb1Tle8Ea9bYyxbwCxMCp0oOTWxsKQY/edit)

### DIFF
--- a/sql_generators/glean_usage/events_first_seen.py
+++ b/sql_generators/glean_usage/events_first_seen.py
@@ -1,0 +1,113 @@
+"""Generate Events First Seen table."""
+
+import os
+from enum import Enum
+from pathlib import Path
+
+from jinja2 import Environment, FileSystemLoader
+
+from bigquery_etl.format_sql.formatter import reformat
+from bigquery_etl.util.common import render, write_sql
+from sql_generators.glean_usage.common import GleanTable
+
+project_id = "moz-fx-data-shared-prod"
+BASE_TABLE = "events_stream_v1"
+TARGET_TABLE = "events_first_seen"
+VERSION = "v1"
+PREFIX = "events_first_seen"
+PATH = Path(os.path.dirname(__file__))
+
+
+class Browsers(Enum):
+    """Enumeration with browser names and equivalent dataset names."""
+
+    firefox_desktop = "Firefox Desktop"
+
+
+class EventsFirstSeenTable(GleanTable):
+    """Represents generated events_first_seen table."""
+
+    def __init__(self):
+        """Initialize events_first_seen table."""
+        GleanTable.__init__(self)
+        self.target_table_id = f"{TARGET_TABLE}_{VERSION}"
+        self.prefix = PREFIX
+        self.custom_render_kwargs = {}
+
+    def generate_across_apps(
+        self, project_id, apps, output_dir=None, use_cloud_function=True, parallelism=8
+    ):
+        """Generate events_first_seen table.
+
+        The parent folders will be created if not existing and existing files will be overwritten.
+        """
+        env = Environment(loader=FileSystemLoader(str(PATH / "templates")))
+        output_dir = Path(output_dir)
+        browser = Browsers["firefox_desktop"]
+
+        # firefox_desktop_derived/events_first_seen_v1/query.sql
+        query_template = env.get_template(f"{TARGET_TABLE}_{VERSION}.query.sql")
+
+        query_sql = reformat(
+            query_template.render(
+                project_id=project_id,
+                app_name=browser.name,
+                base_table=BASE_TABLE,
+                events_first_seen_table=f"{TARGET_TABLE}_{VERSION}",
+            )
+        )
+
+        write_sql(
+            output_dir=output_dir,
+            full_table_id=f"{project_id}.{browser.name}_derived.{TARGET_TABLE}_{VERSION}",
+            basename="query.sql",
+            sql=query_sql,
+            skip_existing=False,
+        )
+
+        # firefox_desktop_derived/events_first_seen_v1/metadata.yaml
+        write_sql(
+            output_dir=output_dir,
+            full_table_id=f"{project_id}.{browser.name}_derived.{TARGET_TABLE}_v1",
+            basename="metadata.yaml",
+            sql=render(
+                sql_filename=f"{TARGET_TABLE}_{VERSION}.metadata.yaml",
+                template_folder=PATH / "templates",
+                output_dir=output_dir,
+                app_name=browser.name,
+                events_first_seen_table=f"{TARGET_TABLE}_{VERSION}",
+                format=False,
+            ),
+            skip_existing=False,
+        )
+
+        # firefox_desktop/events_first_seen/view.sql
+        view_template = env.get_template(f"{TARGET_TABLE}.view.sql")
+
+        write_sql(
+            output_dir=output_dir,
+            full_table_id=f"{project_id}.{browser.name}.{TARGET_TABLE}",
+            basename="view.sql",
+            sql=reformat(
+                view_template.render(
+                    project_id=project_id,
+                    app_name=browser.name,
+                    events_first_seen_view=TARGET_TABLE,
+                    events_first_seen_table=f"{TARGET_TABLE}_{VERSION}",
+                )
+            ),
+            skip_existing=False,
+        )
+
+        # firefox_desktop/events_first_seen/metadata.yaml
+        write_sql(
+            output_dir=output_dir,
+            full_table_id=f"{project_id}.{browser.name}.{TARGET_TABLE}",
+            basename="metadata.yaml",
+            sql=render(
+                sql_filename=f"{TARGET_TABLE}.metadata.yaml",
+                template_folder=PATH / "templates",
+                format=False,
+            ),
+            skip_existing=False,
+        )
--- a/sql_generators/glean_usage/templates/events_first_seen_v1.metadata.yaml
+++ b/sql_generators/glean_usage/templates/events_first_seen_v1.metadata.yaml
@@ -1,0 +1,36 @@
+friendly_name: Events First Seen
+description: |-
+  Captures the earliest date that we observe an event for a particular client in the events_stream_v1 table.
+owners:
+  - kbammarito@mozilla.com
+  - vsabino@mozilla.com
+labels:
+  incremental: true
+  schedule: daily
+  table_type: client_level
+scheduling:
+  dag_name: bqetl_glean_usage
+  task_group: { { app_name } }
+  date_partition_parameter: null
+  destination_table: >-
+    {% raw %}events_first_seen_v1${{
+    (execution_date - macros.timedelta(hours=24)).strftime("%Y%m%d")
+    }}{% endraw %}
+  parameters:
+    - >-
+      {% raw %}submission_date:DATE:{{ds}}{% endraw %}
+  query_file_path:
+    # explicit query file path is necessary because the destination table
+    # includes a partition identifier that is not in the path
+    {{ output_dir }}/{{ app_name }}_derived/{{ events_first_seen_table }}/query.sql
+
+bigquery:
+  time_partitioning:
+    type: day
+    field: submission_date
+    require_partition_filter: true
+    clustering:
+      fields:
+        - event_category
+        - sample_id
+        - normalized_channel
--- a/sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
+++ b/sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
@@ -1,0 +1,127 @@
+{{ header }}
+
+{% raw %}
+{% if is_init() %}
+{% endraw %}
+with
+eventsstream as (
+    select
+    client_id,
+    profile_group_id,
+    sample_id,
+    event_category,
+    event_name,
+    `event`,
+    null as criteria,
+    min(submission_timestamp) as first_submission_timestamp,
+    min(event_timestamp) as first_event_timestamp,
+    min_by(event_extra, submission_timestamp) as event_extra,
+    min_by(app_version_major, submission_timestamp) as app_version_major,
+    min_by(normalized_channel, submission_timestamp) as normalized_channel,
+    min_by(normalized_country_code, submission_timestamp) as normalized_country_code,
+    min_by(normalized_os, submission_timestamp) as normalized_os,
+    min_by(normalized_os_version, submission_timestamp) as normalized_os_version,
+    from
+    `{{ project_id }}.{{ app_name }}_derived.{{ base_table }}`
+    where
+    -- initialize by looking over all of history
+    date(submission_timestamp) >= date_sub(@submission_date, interval 1 day)
+    group by
+    client_id,
+    profile_group_id,
+    sample_id,
+    event_category,
+    event_name,
+    `event`,
+    criteria
+)
+select * from eventsstream
+
+{% raw %}
+{% else %}
+{% endraw %}
+
+with
+_current as (
+    select
+    @submission_date as submission_date,
+    @submission_date as first_seen_date,
+    client_id,
+    profile_group_id,
+    sample_id,
+    event_category,
+    event_name,
+    `event`,
+    null as criteria,
+    min(submission_timestamp) as first_submission_timestamp,
+    min(event_timestamp) as first_event_timestamp,
+    min_by(event_extra, submission_timestamp) as event_extra,
+    min_by(app_version_major, submission_timestamp) as app_version_major,
+    min_by(normalized_channel, submission_timestamp) as normalized_channel,
+    min_by(normalized_country_code, submission_timestamp) as normalized_country_code,
+    min_by(normalized_os, submission_timestamp) as normalized_os,
+    min_by(normalized_os_version, submission_timestamp) as normalized_os_version,
+    from
+    `{{ project_id }}.{{ app_name }}_derived.{{ base_table }}`
+    where
+    date(submission_timestamp) = @submission_date
+    group by
+    submission_date,
+    first_seen_date,
+    sample_id,
+    client_id,
+    profile_group_id,
+    sample_id,
+    event_category,
+    event_name,
+    `event`,
+    criteria
+),
+-- query over all of history to see whether the client_id has shown up before
+_previous as (
+    select
+    client_id,
+    profile_group_id,
+    sample_id,
+    event_category,
+    event_name,
+    `event`,
+    null as criteria,
+    first_submission_timestamp,
+    first_event_timestamp,
+    event_extra,
+    app_version_major,
+    normalized_channel,
+    normalized_country_code,
+    normalized_os,
+    normalized_os_version
+    from
+    `{{ project_id }}.{{ app_name }}_derived.{{ events_first_seen_table }}` -- events_first_seen_table doesn't yet exist
+    where
+    first_seen_date > '2023-01-01'
+    and first_seen_date < @submission_date
+),
+_joined as (
+  --switch to using separate if statements instead of 1
+  --because dry run is struggling to validate the final struct
+  select
+    if(
+      _previous.client_id is null
+      or _previous.first_seen_date >= _current.first_seen_date,
+      _current,
+      _previous
+    ).*
+  from
+    _current
+  full join
+    _previous
+    using (client_id)
+)
+
+select
+*
+from _joined
+
+{% raw %}
+{% endif %}
+{% endraw %}
--- a/sql_generators/glean_usage/templates/events_first_seen_v1.schema.yaml
+++ b/sql_generators/glean_usage/templates/events_first_seen_v1.schema.yaml
@@ -1,0 +1,49 @@
+fields:
+  - mode: NULLABLE
+    name: submission_date
+    type: DATE
+  - mode: NULLABLE
+    name: first_seen_date
+    type: DATE
+  - mode: NULLABLE
+    name: client_id
+    type: STRING
+  - mode: NULLABLE
+    name: profile_group_id
+    type: STRING
+  - mode: NULLABLE
+    name: sample_id
+    type: INTEGER
+    - mode: NULLABLE
+    name: event_category
+    type: STRING
+    - mode: NULLABLE
+    name: event_name
+    type: STRING
+    - mode: NULLABLE
+    name: criteria
+    type: STRING
+    - mode: NULLABLE
+    name: first_submission_timestamp
+    type: TIMESTAMP
+    - mode: NULLABLE
+    name: first_event_timestamp
+    type: TIMESTAMP
+    - mode: NULLABLE
+    name: event_extra
+    type: STRING
+    - mode: NULLABLE
+    name: app_version_major
+    type: STRING
+    - mode: NULLABLE
+    name: normalized_channel
+    type: STRING
+    - mode: NULLABLE
+    name: normalized_country_code
+    type: STRING
+    - mode: NULLABLE
+    name: normalized_os
+    type: STRING
+    - mode: NULLABLE
+    name: normalized_os_version
+    type: STRING
--- a/sql_generators/glean_usage/__init__.py
+++ b/sql_generators/glean_usage/__init__.py
@@ -21,6 +21,7 @@ from sql_generators.glean_usage import (
     event_error_monitoring,
     event_flow_monitoring,
     event_monitoring_live,
+    events_first_seen,
     events_stream,
     events_unnested,
     glean_app_ping_views,
@@ -42,6 +43,7 @@ GLEAN_TABLES = [
     event_monitoring_live.EventMonitoringLive(),
     event_error_monitoring.EventErrorMonitoring(),
     event_flow_monitoring.EventFlowMonitoring(),
+    events_first_seen.EventsFirstSeenTable(),
     events_stream.EventsStreamTable(),
 ]
 
--- a/sql_generators/glean_usage/common.py
+++ b/sql_generators/glean_usage/common.py
@@ -146,6 +146,8 @@ def table_names_from_baseline(baseline_table, include_project_id=True):
         events_view=f"{prefix}.events",
         events_stream_table=f"{prefix}_derived.events_stream_v1",
         events_stream_view=f"{prefix}.events_stream",
+        events_first_seen_table=f"{prefix}_derived.events_first_seen_v1",
+        events_first_seen_view=f"{prefix}.events_first_seen",
     )
 
 
@@ -449,7 +451,7 @@ class GleanTable:
             target_table=f"{target_dataset}_derived.{self.target_table_id}",
             app_name=app_name,
             enable_monitoring=enable_monitoring,
-            deprecated_app = deprecated_app,
+            deprecated_app=deprecated_app,
         )
         render_kwargs.update(self.custom_render_kwargs)
 
--- a/sql_generators/glean_usage/templates/events_first_seen.metadata.yaml
+++ b/sql_generators/glean_usage/templates/events_first_seen.metadata.yaml
@@ -1,0 +1,7 @@
+{{ header_yaml }}
+friendly_name: Events First Seen
+description: |-
+  Captures the earliest date that we observe an event for a particular client in the events_stream_v1 table.
+owners:
+  - kbammarito@mozilla.com
+  - vsabino@mozilla.com
--- a/sql_generators/glean_usage/templates/events_first_seen.view.sql
+++ b/sql_generators/glean_usage/templates/events_first_seen.view.sql
@@ -1,0 +1,9 @@
+{{ header }}
+
+CREATE OR REPLACE VIEW
+  `{{ project_id }}.{{ app_name }}.{{ events_first_seen_view }}`
+AS
+SELECT
+  *
+FROM
+  `{{ project_id }}.{{ app_name }}_derived.{{ events_first_seen_table }}`


sql_generators/glean_usage/events_first_seen.py
```diff
@@ -0,0 +1,113 @@
...
+        self, project_id, apps, output_dir=None, use_cloud_function=True, parallelism=8
+    ):
+        """Generate events_first_seen table.
+
+        The parent folders will be created if not existing and existing files will be overwritten.
+        """
+        env = Environment(loader=FileSystemLoader(str(PATH / "templates")))
+        output_dir = Path(output_dir)
+        browser = Browsers["firefox_desktop"]
+
```
scholtzan: Currently `events_first_seen` is generated for every app. You probably need something like this to skip generation for apps that are not explicitly configured:

```python
include_apps = ConfigLoader.get(
    "generate", "glean_usage", "events_first_seen", "include_apps", fallback=[]
) # this would have firefox_desktop
apps = [app[0] for app in apps if app[0]["app_name"] in include_apps]
if apps == []:
            return
```

scholtzan:
I think CI might be failing because in the generated metadata.yaml file the `task_group` isn't correctly resolved:

```yaml
scheduling:
  dag_name: bqetl_glean_usage
  task_group: { { app_name } }
```

sql_generators/glean_usage/templates/events_first_seen_v1.metadata.yaml
```diff
@@ -0,0 +1,36 @@
...
+owners:
+  - kbammarito@mozilla.com
+  - vsabino@mozilla.com
+labels:
+  incremental: true
+  schedule: daily
+  table_type: client_level
+scheduling:
+  dag_name: bqetl_glean_usage
+  task_group: { { app_name } }
```
scholtzan:
```suggestion
  task_group: {{ app_name }}
```
the extra spaces might cause issues here

sql_generators/glean_usage/events_first_seen.py
kbammarito: I just redid this to use `event_flow_monitoring.py` as a template. Seems to work now. Thanks!

sql_generators/glean_usage/events_first_seen.py
```diff
@@ -0,0 +1,145 @@
...
+        view_filename = f"{TARGET_TABLE}.view.sql"
+
+        view_sql = render(
+            sql_filename=view_filename,
+            template_folder=PATH / "templates",
+            format=False,
+            **render_kwargs,
+        )
+
+        # firefox_desktop_derived/events_first_seen_v1/metadata.yaml
```
scholtzan:
```suggestion
        # firefox_desktop/events_first_seen_v1/metadata.yaml
```

sql_generators/glean_usage/templates/events_first_seen_v1.metadata.yaml
```diff
@@ -0,0 +1,37 @@
...
+    # explicit query file path is necessary because the destination table
+    # includes a partition identifier that is not in the path
+    {{ output_dir }}/{{ app_name }}_derived/{{ events_first_seen_table }}/query.sql
+  depends_on_past: true
+  date_partition_parameter: null
+
+bigquery:
+  time_partitioning:
+    type: day
+    field: submission_date
```
scholtzan:
```suggestion
    field: first_seen_date
```

These tables are usually partitioned on `first_seen_date`

sql_generators/glean_usage/templates/events_first_seen_v1.metadata.yaml
```diff
@@ -0,0 +1,37 @@
...
+    require_partition_filter: true
```
scholtzan:
```suggestion
    require_partition_filter: false
```
Partition filter requirement is usually disabled for these kinds of tables

sql_generators/glean_usage/templates/events_first_seen_v1.metadata.yaml
```diff
@@ -0,0 +1,37 @@
...
+    clustering:
+      fields:
+        - event_category
+        - sample_id
+        - normalized_channel
```
scholtzan:
```suggestion
        - normalized_channel
        - submission_date
```
It may be worth considering clustering on submission_date. I'm guessing lot's of queries will filter on submission_date so this could improve performance

sql_generators/glean_usage/templates/events_first_seen_v1.metadata.yaml
```diff
@@ -0,0 +1,37 @@
...
+  schedule: daily
+  table_type: client_level
+scheduling:
+  dag_name: bqetl_glean_usage
+  task_group: {{ app_name }}
+  date_partition_parameter: null
+  destination_table: >-
+    {% raw %}events_first_seen_v1${{
+    (execution_date - macros.timedelta(hours=24)).strftime("%Y%m%d")
+    }}{% endraw %}
```
scholtzan:
```suggestion
```
The query updates the entire table (because it joins "current" and "previous"), so we don't need to specify a partition here

sql_generators/glean_usage/templates/events_first_seen_v1.metadata.yaml
```diff
@@ -0,0 +1,37 @@
...
+  parameters:
+    - >-
+      {% raw %}submission_date:DATE:{{ds}}{% endraw %}
+  query_file_path:
+    # explicit query file path is necessary because the destination table
+    # includes a partition identifier that is not in the path
+    {{ output_dir }}/{{ app_name }}_derived/{{ events_first_seen_table }}/query.sql
```
scholtzan:
```suggestion
```

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,131 @@
...
+    normalized_channel,
+    normalized_country_code,
+    normalized_os,
+    normalized_os_version
+    from
+    `{{ project_id }}.{{ app_name }}_derived.{{ events_first_seen_table }}` -- events_first_seen_table doesn't yet exist
+    where
+    first_seen_date > '2023-01-01'
+    and first_seen_date < @submission_date
+    and submission_date = @submission_date -- added this because it needs a partition filter
```
scholtzan:
```suggestion
```
Removing the partition filter requirement in the metadata doesn't require this anymore

sql_generators/glean_usage/templates/events_first_seen_v1.metadata.yaml
```diff
@@ -0,0 +1,28 @@
...
+labels:
+  schedule: daily
+  table_type: client_level
+scheduling:
+  dag_name: bqetl_glean_usage
+  task_group: {{ app_name }}
+  parameters:
+    - >-
+      {% raw %}submission_date:DATE:{{ds}}{% endraw %}
+  depends_on_past: true
```
scholtzan, line 16:
```suggestion
  depends_on_past: true
  date_partition_parameter: null
```

alekhyamoz (approved):
lgtm

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,130 @@
...
+    min_by(app_version_major, submission_timestamp) as app_version_major,
+    min_by(normalized_channel, submission_timestamp) as normalized_channel,
+    min_by(normalized_country_code, submission_timestamp) as normalized_country_code,
+    min_by(normalized_os, submission_timestamp) as normalized_os,
+    min_by(normalized_os_version, submission_timestamp) as normalized_os_version,
+    from
+    `{{ project_id }}.{{ app_name }}_derived.{{ base_table }}`
+    where
+    -- initialize by looking over all of history
+    date(submission_timestamp) >= date_sub(@submission_date, interval 1 day)
```
BenWu: I think this will fail deployment because is_init doesn't add the submission_date parameter.

Also is this meant to be `<=`  based on the comment or do you only want the most recent day to be initialized?

BenWu (changes requested):
Sorry to remove it from the queue but I think this will fail to deploy

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,140 @@
+{{ header }} -- the header doesn't seem to work properly
```
BenWu: The header isn't automatically added for `generate_across_apps` so you would need to add it yourself. https://github.com/mozilla/bigquery-etl/blob/69d9173b5d4406b244a4d1cb855ad835e8dfcb4b/sql_generators/glean_usage/common.py#L527-L532

But `generate_across_apps` seems like the wrong choice here since it will only run once, meaning if you added another app to the bqetl_project.yaml, it would only generate a query for one app.  `generate_per_app_id` is more appropriate since events_stream is per app id

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,140 @@
...
+    `{{ project_id }}.{{ app_name }}_derived.{{ base_table }}`
+    where
+    -- initialize by looking over all of history
+    date(submission_timestamp) >= '2023-01-01'
+    and profile_group_id is not null
+    and event_category not in (
+      'media.playback',
+      'nimbus_events',
+      'uptake.remotecontent.result'
+    )
```
BenWu, line 30: If you add `sample_id = @sample_id`, it will run the init per sample id.  This is the only way to do this without the query timing out.
```suggestion
    )
    AND sample_id = @sample_id
```
It will block deployments for however many hours it runs so I would recommend either removing the is_init completely or making only run for the `MIN(DATE(submission_timestamp))`, and then rely on a managed backfill.

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
BenWu, line 1: nit: can you run `bqetl format sql_generators/glean_usage/templates/events_first_seen_v1.query.sql` to format this?  It would make this a little nicer to read.  You'll need to remove all the `{% raw %}` and `{% endraw %}` first, then add them back in afterwards

sql_generators/glean_usage/events_first_seen.py
```diff
@@ -0,0 +1,145 @@
...
+        """Initialize events_first_seen table."""
+        self.per_app_id_enabled = False
+        self.per_app_enabled = False
+        self.across_apps_enabled = True
+        self.prefix = PREFIX
+        self.target_table_id = f"{TARGET_TABLE}_{VERSION}"
+        self.custom_render_kwargs = {}
+        self.base_table_name = "events_stream_v1"
+
+    def generate_across_apps(
```
BenWu: per the other comment, this should be a `generate_per_app_id`

sql_generators/glean_usage/events_first_seen.py
```diff
@@ -0,0 +1,145 @@
...
+        self, project_id, apps, output_dir=None, use_cloud_function=True, parallelism=8
+    ):
+        """Generate a query across all apps."""
+        if not self.across_apps_enabled:
+            return
+
+        # output_dir = Path(output_dir)
+        browser = Browsers["firefox_desktop"]
```
BenWu: The enum value doesn't seem to be used anywhere.  Can this be removed?  I don't think this needs to be hardcoded anyway because the app name is in the `apps` dict.

sql_generators/glean_usage/events_first_seen.py
```diff
@@ -0,0 +1,145 @@
...
+        apps = [app[0] for app in apps if app[0]["app_name"] in include_apps]
+
+        render_kwargs = dict(
+            project_id=project_id,
+            output_dir=Path(output_dir),
+            base_table=self.base_table_name,
+            app_name=browser.name,
+            events_first_seen_table=f"{TARGET_TABLE}_{VERSION}",
+            events_first_seen_view=f"{TARGET_TABLE}",
+            apps=apps,
```
BenWu: This doesn't seem to be used
```suggestion
```

BenWu (commented):
I have a couple non-trivial comments. Let me know if I can clarify anything

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
kbammarito: As discussed, I'm removing the entire _is_init_ block and doing a managed backfill instead

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
kbammarito: this has been implemented

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
kbammarito: As discussed, we are reverting back to the _is_init_ block

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,135 @@
...
+    min_by(normalized_os, submission_timestamp) AS normalized_os,
+    min_by(normalized_os_version, submission_timestamp) AS normalized_os_version,
+  FROM
+    `{{ project_id }}.{{ app_name }}_derived.events_stream_v1`
+  WHERE
+    -- initialize by looking over all of history
+    DATE(submission_timestamp) >= '2023-01-01'
+    AND profile_group_id IS NOT NULL
+    AND event_category NOT IN ('media.playback', 'nimbus_events', 'uptake.remotecontent.result')
+    AND sample_id = @sample_id
```
BenWu: You can copy what Winnie did in https://github.com/mozilla/bigquery-etl/pull/8258/files#diff-9ce6039183364ba26975e76377b309479853a423201303e143d83eb28afe9aed

sql_generators/glean_usage/events_first_seen.py
sean-rose, line 1: Note that if my PR #8345 gets merged before this PR, then this code will need to be updated to match the updated `GleanTable` class:
- The `per_app_id_enabled` attribute is being renamed `per_app_channel_enabled`.
- The `generate_per_app_id()` method is being renamed `generate_per_app_channel()` and will have slightly different args.
  - You'll be able to use the new `app_name` arg instead of that `app_id` logic (which was a misnomer anyway because that was really getting the "dataset family", not the app ID).
- The `generate_per_app()` method will have slightly different args.
  - You'll be able to use the new `app_name` arg instead of getting it from the first app channel.

sql_generators/glean_usage/events_first_seen.py
```diff
@@ -0,0 +1,73 @@
...
+        self,
+        project_id,
+        app_info,
+        output_dir=None,
+        use_cloud_function=True,
+        parallelism=8,
+        id_token=None,
+        all_base_tables_exist=None,
+    ):
+        """Generate the events_stream table query per app_name."""
```
sean-rose: **suggestion (non-blocking):** This doc string looks copy-pasted from `events_stream.py` and should be updated.

sql_generators/glean_usage/events_first_seen.py
sean-rose: #8345 has been merged, but with some differences from the changes I listed before.  I'll add some comments with suggestions for how to update this code appropriately.

sql_generators/glean_usage/events_first_seen.py
```diff
@@ -0,0 +1,73 @@
...
+        """Initialize events_first_seen table."""
+        GleanTable.__init__(self)
+        self.target_table_id = TARGET_TABLE_ID
+        self.prefix = PREFIX
+        self.per_app_enabled = True
+        self.per_app_id_enabled = True
+        self.across_apps_enabled = False
+        self.cross_channel_template = "cross_channel.view.sql"
+        self.base_table_name = "events_v1"
+        self.custom_render_kwargs = {}
```
sean-rose: **suggestion (blocking):** The `custom_render_kwargs` attribute was renamed `common_render_kwargs` in #8345.
```suggestion
        self.common_render_kwargs = {}
```

sql_generators/glean_usage/events_first_seen.py
```diff
@@ -0,0 +1,73 @@
...
+            "generate", "glean_usage", "events_first_seen", "include_apps", fallback=[]
+        ):
+            super().generate_per_app_id(
+                project_id,
+                baseline_table,
+                output_dir,
+                use_cloud_function,
+                app_info,
+                id_token=id_token,
+            )
```
sean-rose, line 70: **suggestions (blocking):** For `generate_per_app_id()`:
- The `app_info` argument was replaced with `app_name` and `app_id_info` arguments in #8345.
- `pydocstyle` requires a docstring for public methods.
- All arguments should be passed through to the base `generate_per_app_id()` method.

```suggestion
    def generate_per_app_id(
        self,
        project_id,
        baseline_table,
        app_name,
        app_id_info,
        output_dir=None,
        use_cloud_function=True,
        parallelism=8,
        id_token=None,
    ):
        """Generate the events_first_seen table query per app_id."""
        # Include only selected apps
        if app_name in ConfigLoader.get(
            "generate", "glean_usage", "events_first_seen", "include_apps", fallback=[]
        ):
            super().generate_per_app_id(
                project_id,
                baseline_table,
                app_name,
                app_id_info,
                output_dir=output_dir,
                use_cloud_function=use_cloud_function,
                parallelism=parallelism,
                id_token=id_token,
            )
```

sql_generators/glean_usage/events_first_seen.py
```diff
@@ -0,0 +1,73 @@
...
+    ):
+        """Generate the events_stream table query per app_name."""
+        target_dataset = app_info[0]["app_name"]
+        if target_dataset in ConfigLoader.get(
+            "generate", "glean_usage", "events_first_seen", "include_apps", fallback=[]
+        ):
+
+            super().generate_per_app(
+                project_id, app_info, output_dir, id_token=id_token
+            )
```
sean-rose, line 96: **suggestions (blocking):** For `generate_per_app()`:
- The `app_info` argument was replaced with `app_name` and `app_ids_info` arguments in #8345.
- As I mentioned in an earlier comment, the method's docstring should be corrected to not refer to `events_stream`.
- All arguments should be passed through to the base `generate_per_app()` method.

```suggestion
    def generate_per_app(
        self,
        project_id,
        app_name,
        app_ids_info,
        output_dir=None,
        use_cloud_function=True,
        parallelism=8,
        id_token=None,
        all_base_tables_exist=None,
    ):
        """Generate the events_first_seen table query per app_name."""
        if app_name in ConfigLoader.get(
            "generate", "glean_usage", "events_first_seen", "include_apps", fallback=[]
        ):
            super().generate_per_app(
                project_id,
                app_name
                app_ids_info,
                output_dir=output_dir,
                use_cloud_function=use_cloud_function,
                parallelism=parallelism,
                id_token=id_token,
                all_base_tables_exist=all_base_tables_exist,
            )
```

sean-rose (commented):
You'll also probably need to be careful the next time you merge `main` into this branch, because in #8345 I made a number of changes to the `__init__.py` and `common.py` files which overlap with some of the changes you've made in this PR (e.g. some of the same `black` code formatting changes, but also actual logic changes).

sql_generators/glean_usage/templates/events_first_seen_v1.metadata.yaml
```diff
@@ -0,0 +1,29 @@
...
+bigquery:
+  time_partitioning:
+    type: day
+    field: event_first_seen_date
+    require_partition_filter: false
+  clustering:
+    fields:
+      - event_category
+      - sample_id
+      - normalized_channel
```
BenWu: You generally want to have the lower cardinality fields earlier in the clustering order for more effective pruning.  `event_category` first is fine though since every query on this table should be filtering by it
```suggestion
      - normalized_channel
      - sample_id
```

BenWu (approved):
Just one minor comment

sql_generators/glean_usage/templates/events_first_seen_v1.metadata.yaml
```diff
@@ -0,0 +1,29 @@
...
+  time_partitioning:
+    type: day
+    field: event_first_seen_date
+    require_partition_filter: false
+  clustering:
+    fields:
+      - event_category
+      - normalized_channel
+      - sample_id
+      - submission_date
```
bani: I don't think you need to cluster on submission_date as we already have the partitioning in first_seen_date. Country would be more useful.

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,131 @@
+{{ header }}
+
+{% raw %}
+{% if is_init() %}
```
bani, line 4: Wouldn't the init table require the same columns as the regular one? Like event_first_seen_date and submission_date?

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,131 @@
...
+{% endraw %}
+
+WITH eventsstream AS (
+  SELECT
+    client_id,
+    `event`,
+    CAST(NULL AS string) AS criteria,
+    min_by(profile_group_id, submission_timestamp) AS profile_group_id,
+    min_by(sample_id, submission_timestamp) AS sample_id,
+    min_by(event_category, submission_timestamp) AS event_category,
```
bani: I find this syntax unexpected. Event is defined by concatenating event_category and event_name. I think it'd be less confusing to not "aggregate" it.

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,131 @@
+{{ header }}
+
+{% raw %}
+{% if is_init() %}
+{% endraw %}
+
+WITH eventsstream AS (
+  SELECT
+    client_id,
+    `event`,
```
bani: why is event the only one with backticks?

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,131 @@
...
+    AND profile_group_id IS NOT NULL
+    AND event_category NOT IN ('media.playback', 'nimbus_events', 'uptake.remotecontent.result')
+  GROUP BY
+    submission_date,
+    event_first_seen_date,
+    client_id,
+    `event`,
+    criteria
+),
+-- query over all of history to see whether the client_id has shown up before
```
bani: client_id and event

sql_generators/glean_usage/templates/events_first_seen_v1.schema.yaml
```diff
@@ -0,0 +1,52 @@
+fields:
```
bani, line 1: I think we are expected to start writing descriptions for documentation?

kbammarito: Yes, I was going to add descriptions and general documentation after we were sure the query was how we want it

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
kbammarito: Updated to read `-- query over all of history to see whether the client_id, event and criteria combination has shown up before`

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
kbammarito: This was because `event` can be a reserved keyword. It doesn't actually change anything. But looking now, it doesn't look like `event` is even a keyword in BigQuery SQL anyway. I can remove the backticks if they're confusing.

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
kbammarito: Thanks for flagging! I think I'd had them before but I thought they'd been removed for a reason I now cannot recall. I've added them back in and updated the tests.

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
kbammarito: I did this so I could explicitly group by only `client_id, event, criteria` and thus specifically show the granularity as **one row per client_id, event and criteria, with all other fields coming from the earliest submission_timestamp**. Are you saying to also group by `event_category` and `event_name`?

sql_generators/glean_usage/templates/events_first_seen_v1.metadata.yaml
kbammarito: Updated

sql_generators/glean_usage/templates/events_first_seen_v1.metadata.yaml
```diff
@@ -0,0 +1,29 @@
+friendly_name: Events First Seen
```
bani, line 1: did we need something in the metadata to not apply the data retention policy to this table?

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
bani: The event column is just a query helper added very late in the pipeline, while event_category and event_name are the actual data columns, so yes, I think having them in the group by conveys better what the data is about.

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
bani: Up to you

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
kbammarito: Updated

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
kbammarito: Keeping them so as not to have so many changes

sql_generators/glean_usage/templates/events_first_seen_v1.metadata.yaml
kbammarito: It's [here](https://github.com/mozilla/bigquery-etl/pull/8046/files#diff-7e34d2d38e27a920e060691809970185fc8f8b0bf3d3e7c7fb4792ed527c3c11R551) in the retention exclusion list

sql_generators/glean_usage/templates/events_first_seen_templating.yaml
```diff
@@ -0,0 +1,29 @@
...
+#     criteria:
+#       - name: <name of criteria>
+#         sql: <SQL expression>
+
+# To correctly output the name of the criteria inside single quotes, wrap it in single and then double quotes. E.g., "'sidebar'"
+
+apps:
+  firefox_desktop:
+    criteria:
+      - name: CAST(NULL as string)
```
sean-rose, line 14: **suggestions (non-blocking):**
- I'd advise against using null as a special criteria name value here because nulls are harder to work with and easy to misinterpret.  Instead I'd suggest using some more meaningful name like `any_event` (or even a less scrutable special value like `*` would be better than null IMO).
- I'd also advise against using raw SQL syntax for the criteria names, as that's making it difficult to configure things correctly (as evidenced by the comment above about how to YAML-encode the SQL strings for criteria names, which IMO ideally shouldn't be necessary).
  - Even if you decide you want to stick with using null as a special criteria name value, you could encode that as a YAML `null` literal here and then do something like `{{ ("'" ~ item["name"] ~ "'") if item["name"] else "CAST(NULL AS STRING)" }} AS criteria` in the ETL query template.

And if you accept both of those suggestions you could also encode the criteria in a more natural way as a dictionary, like:
```
apps:
  <app name>:
    criteria:
      <criteria name>: |
        <criteria SQL>
```
(in which case you'd want to loop over the criteria in the ETL query template like `{% for criteria_name, criteria_sql in criteria.items() %}`)

sql_generators/glean_usage/templates/events_first_seen_templating.yaml
```diff
@@ -0,0 +1,29 @@
...
+        sql: true
+      - name: "'chatbot_usage'"
+        sql: event_category = 'genai.chatbot' AND event_name = 'sidebar_toggle' AND JSON_VALUE(event_extra.provider) <> 'none'
+      - name: "'smart_tabgroup_save'"
+        sql: event_category = 'tabgroup' AND ((event_name = 'smart_tab_suggest' AND JSON_VALUE(event_extra.action) LIKE 'save%') OR (event_name = 'smart_tab_topic' AND JSON_VALUE(event_extra.action) = 'save'))
```
sean-rose, line 19: **suggestion (non-blocking):** IMO these criteria SQL values would be easier to read/edit if they were formatted on multiple lines in YAML block scalars, like:
```suggestion
        sql: |
          event_category = 'tabgroup'
          AND (
            (event_name = 'smart_tab_suggest' AND JSON_VALUE(event_extra.action) LIKE 'save%')
            OR (event_name = 'smart_tab_topic' AND JSON_VALUE(event_extra.action) = 'save')
          )
```
(if you accept this suggestion I'd recommend consistently using YAML block scalars for all criteria SQL values, not just the very long ones)

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,164 @@
...
+{% if is_init() %}
+{% endraw %}
+
+{% for item in criteria %}
+
+(
+WITH eventsstream AS (
+  SELECT
+  DATE(MIN(submission_timestamp)) as submission_date,
+  DATE(MIN(submission_timestamp)) as event_first_seen_date,
```
sean-rose: **questions (non-blocking):**
- Why have `submission_date` and `event_first_seen_date` columns with the exact same value?
- Why have these date columns when you have the more precise `first_submission_timestamp` column?

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,164 @@
...
+  {% if app_id == 'firefox_desktop' -%}
+  AND profile_group_id IS NOT NULL
+  {% endif %}
+        -- below is the templated criteria
+  AND ({{ item["sql"] }})
+  GROUP BY
+  client_id,
+  `event`,
+  event_category,
+  event_name,
```
sean-rose, line 54: **question (blocking):** Are you absolutely certain you want to group by the event/category/name?

Some of the configured criteria could match multiple types of events, so grouping by event/category/name means you could get multiple rows per client for those criteria (one for each distinct event type that matches the criteria's conditions).

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,164 @@
...
+  min_by(profile_group_id, submission_timestamp) AS profile_group_id,
+  min_by(sample_id, submission_timestamp) AS sample_id,
+  MIN(submission_timestamp) AS first_submission_timestamp,
+  MIN(event_timestamp) AS first_event_timestamp,
+  min_by(event_extra, submission_timestamp) AS event_extra,
+  min_by(app_version_major, submission_timestamp) AS app_version_major,
+  min_by(normalized_channel, submission_timestamp) AS normalized_channel,
+  min_by(normalized_country_code, submission_timestamp) AS normalized_country_code,
+  min_by(normalized_os, submission_timestamp) AS normalized_os,
+  min_by(normalized_os_version, submission_timestamp) AS normalized_os_version
```
sean-rose: **issues (blocking):**
- Events submitted in the same `events` ping will all have the same `submission_timestamp`, so there could easily be ties based on `submission_timestamp` values and these `MIN_BY()` calls might not end up selecting the properties for the actual first such event.
- Even if you also compare using `event_timestamp` to break `submission_timestamp` ties it's currently impossible to guarantee there won't be any ties, as events could also have the same `event_timestamp` (though my planned solution for [DENG-9800](https://mozilla-hub.atlassian.net/browse/DENG-9800) might eventually help with this). And if ties happen then the multiple `MIN_BY()` calls here aren't necessarily guaranteed to all end up getting their data from the same event record.

So I'd suggest using `event_timestamp` in the comparison and getting all the necessary data in a single aggregate call.  Unfortunately, `event_timestamp` can have some wild values (e.g. far in the past or future) so a simple `COALESCE(event_timestamp, submission_timestamp)` probably isn't safe to use, and `MIN_BY()` doesn't allow comparing multiple values. However, you could do something like this:
```sql
  ARRAY_AGG(
    STRUCT(
      profile_group_id,
      sample_id,
      submission_timestamp AS first_submission_timestamp,
      event_timestamp AS first_event_timestamp,
      event_extra,
      app_version_major,
      normalized_channel,
      normalized_country_code,
      normalized_os,
      normalized_os_version
    )
    ORDER BY
      submission_timestamp,
      event_timestamp NULLS LAST
    LIMIT 1
  )[0].*
```

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,164 @@
...
+  FROM
+  `{{ project_id }}.{{ app_id }}_derived.events_stream_v1`
+  WHERE
+        -- initialize by looking over all of history
+  DATE(submission_timestamp) >= '2023-01-01'
+        -- AND sample_id >= @sample_id
+        -- AND sample_id < @sample_id + @sampling_batch_size
```
sean-rose: **issue (possibly blocking):** As it stands the init query for `firefox_desktop_derived.events_first_seen_v1` is going to be scanning petabytes of data, so I suspect having the init query run per sample ID may be necessary to avoid having that take an exceedingly long time (blocking the artifact deployment process, and possibly even hitting the 6-hour query runtime limit).

@BenWu what do you think?  Any advice on how to estimate how long such a huge query is likely to take?

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,164 @@
...
+  AND event_category NOT IN ('media.playback', 'nimbus_events', 'uptake.remotecontent.result')
```
sean-rose, line 43: **suggestion (non-blocking):** It'd be helpful to have a comment explaining why these particular event categories are being excluded.

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,164 @@
...
+        -- if app_id is firefox_desktop, filter for where profile_group_id is not null
```
sean-rose, line 44: **suggestions (non-blocking):**
- This comment explains what is being done, but we could already tell that by looking at the code in question.  Better would be having a comment that explains _why_ this is only including Firefox Desktop events that have `profile_group_id` values.
- In any case, the comment should go within the `{% if app_id == 'firefox_desktop' -%}` block (otherwise the generated ETL queries for other apps will confusingly contain the comment but not the associated SQL code).

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
sean-rose: **suggestion (non-blocking):** Apparently `normalized_os_version` ends up being `10.0` for both Windows 10 and 11 (though IMO that seems like a bug we should fix), so I'd suggest also including `client_info.windows_build_number` to allow differentiating between Windows 10 and 11 (e.g. [DENG-8570](https://mozilla-hub.atlassian.net/browse/DENG-8570)).

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,164 @@
...
+        _previous
+      ).*
+    FROM
+      _current
+    FULL OUTER JOIN
+      _previous
+      ON _current.client_id = _previous.client_id
+          AND _current.event = _previous.event
+          AND (_current.criteria = _previous.criteria
+              OR (_current.criteria IS NULL AND _previous.criteria IS NULL))
```
sean-rose, line 161: - If you decide not to group by event/category/name then the `event` join condition here should be removed.
- If you decide to avoid using null criteria name values then the `criteria` join condition here could be simplified.

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
sean-rose, line 1: **suggestions (non-blocking):**
- Formatting this file in a more `bqetl format`-like way would help with readability.
- There's a fair amount of duplicated logic between the init and non-init parts of the SQL which would be good to avoid if possible.

sql_generators/glean_usage/templates/events_first_seen_v1.schema.yaml
```diff
@@ -0,0 +1,52 @@
...
+  - mode: NULLABLE
+    name: submission_date
+    type: DATE
```
sean-rose: **nitpick:** IMO it's good to keep the schema field attributes consistently in the order `name`, `type`, `mode`, `description`, `fields` for readability, rather than having the attributes be sorted alphabetically like this.

sql_generators/glean_usage/events_first_seen.py
```diff
@@ -0,0 +1,96 @@
...
+    def __init__(self):
+        """Initialize events_first_seen table."""
+        GleanTable.__init__(self)
+        self.target_table_id = TARGET_TABLE_ID
+        self.prefix = PREFIX
+        self.per_app_enabled = True
+        self.per_app_id_enabled = True
+        self.across_apps_enabled = False
+        self.cross_channel_template = "cross_channel_events_first_seen.view.sql"
+        self.base_table_name = "events_v1"
```
sean-rose, line 27: **suggestion (non-blocking):** While this doesn't really matter since you're using an opt-in approach, it would be more accurate to define the base table as `events_stream_v1`.
```suggestion
        self.base_table_name = "events_stream_v1"
```

sql_generators/glean_usage/events_first_seen.py
```diff
@@ -0,0 +1,96 @@
...
+                project_id,
+                baseline_table,
+                app_name,
+                app_id_info,
+                output_dir=output_dir,
+                use_cloud_function=use_cloud_function,
+                parallelism=parallelism,
+                id_token=id_token,
+                custom_render_kwargs={
+                    "app_id": app_id_info["bq_dataset_family"],
```
sean-rose, line 67: **suggestion (non-blocking):** IMO calling this value `app_id` is a bit confusing/misleading, so I'd suggest changing this to something like `app_id_dataset` (and updating the associated Jinja).

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
kbammarito: @bani is this still what you wanted, given the new criteria you provided me?

bani: Good point, for my use case that has this issue (smart tab group save) it's fine to have separate rows one for topic and one for suggestion.
But I see TOU also has the same behaviour so we should check with @soGaussian as well.

kbammarito: OK thanks. @soGaussian please let us know about your criteria. I would like to avoid extensive conditionals in the jinja if possible.

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
kbammarito: @sean-rose per our convo, I'll add (uncomment) this back in

sql_generators/glean_usage/events_first_seen.py
sean-rose: Never mind! It turns out [the `glean_usage` generator currently assumes base tables are in `*_stable` datasets](https://github.com/mozilla/bigquery-etl/blob/bb849d8c2e475c91ae822a3b75234ae61fc8884e/sql_generators/glean_usage/__init__.py#L153-L154), so if you applied my suggestion the `EventsFirstSeenTable` code wouldn't get called at all! 😬

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
soGaussian: For ToU use case, it is also fine to have separate rows each event as well.

kbammarito: Thanks all. I'll leave it as is, then. Thanks @sean-rose for flagging!

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
kbammarito: I implemented this with the `COALESCE(event_timestamp, '9999-12-31 23:59:59')` (per our convo). Thanks!

kbammarito: Done. Thanks!

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,176 @@
...
+        )
+        ORDER BY
+          submission_timestamp,
+          COALESCE(event_timestamp, '9999-12-31 23:59:59')
+        LIMIT 1
+      )[0].*
+  FROM
+    `{{ project_id }}.{{ app_id }}_derived.events_stream_v1`
+  WHERE
+    DATE(submission_timestamp) = DATE(@submission_timestamp)
```
sean-rose: **suggestion (blocking):** The query parameter defined in `metadata.yaml` is `@submission_date`, not `@submission_timestamp`.
```suggestion
    DATE(submission_timestamp) = @submission_date
```

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,176 @@
...
+      normalized_channel,
+      normalized_country_code,
+      normalized_os,
+      normalized_os_version,
+      windows_build_number
+    FROM
+      `{{ project_id }}.{{ events_first_seen_table }}`
+    WHERE
+      DATE(first_submission_timestamp) > '2023-01-01'
+      AND DATE(first_submission_timestamp) < DATE(@submission_timestamp)
```
sean-rose: **suggestion (blocking):** The query parameter defined in `metadata.yaml` is `@submission_date`, not `@submission_timestamp`.
```suggestion
      AND DATE(first_submission_timestamp) < @submission_date
```

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,176 @@
...
+      app_version_major,
+      normalized_channel,
+      normalized_country_code,
+      normalized_os,
+      normalized_os_version,
+      windows_build_number
+    FROM
+      `{{ project_id }}.{{ events_first_seen_table }}`
+    WHERE
+      DATE(first_submission_timestamp) > '2023-01-01'
```
sean-rose: **suggestion (blocking):** This condition should either exactly match what's in the init query:
```suggestion
      DATE(first_submission_timestamp) >= '2023-01-01'
```
Or, since the `events_first_seen_v1` tables are not being configured to require a partition column filter, this condition could be removed.

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,176 @@
...
+    DATE(submission_timestamp) = DATE(@submission_timestamp)
+    AND event_category NOT IN ('media.playback', 'nimbus_events', 'uptake.remotecontent.result')
+        -- if app_id is firefox_desktop, filter for where profile_group_id is not null
+    {% if app_id == 'firefox_desktop' -%}
+    AND profile_group_id IS NOT NULL
+    {% endif %}
+        -- below is the templated criteria
+    AND ({{ item["sql"] }})
+  GROUP BY
+    DATE(submission_timestamp),
```
sean-rose: **suggestion (blocking):** There's no need to group by `DATE(submission_timestamp)` here, as that expression is not being used in the select list outside of an aggregation.
```suggestion
```

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,176 @@
...
+      DATE(first_submission_timestamp) > '2023-01-01'
+      AND DATE(first_submission_timestamp) < DATE(@submission_timestamp)
+  ),
+  _joined AS (
+    --switch to using separate if statements instead of 1
+    --because dry run is struggling to validate the final struct
+    SELECT
+      IF(
+        _previous.client_id IS NULL
+        OR date(_previous.first_submission_timestamp) >= date(_current.first_submission_timestamp),
```
sean-rose: **suggestion (non-blocking):** With the `_current` CTE using the condition `DATE(submission_timestamp) = @submission_date` and the `_previous` CTE using the condition `DATE(first_submission_timestamp) < @submission_date`, I believe the date condition here is unnecessary.
```suggestion
        _previous.client_id IS NULL,
```

tests/sql/moz-fx-data-shared-prod/firefox_desktop_derived/events_first_seen_v1/test_aggregation/query_params.yaml
```diff
@@ -0,0 +1,4 @@
+---
+- name: submission_timestamp
+  type: TIMESTAMP
+  value: 2025-10-29T10:00:00
```
sean-rose: **suggestion (blocking):** The query parameter defined in `metadata.yaml` is `submission_date`, not `submission_timestamp`.
```suggestion
- name: submission_date
  type: DATE
  value: 2025-10-29
```

tests/sql/moz-fx-data-shared-prod/firefox_desktop_derived/events_first_seen_v1/moz-fx-data-shared-prod.firefox_desktop_derived.events_first_seen_v1.schema.json
```diff
@@ -0,0 +1,87 @@
...
+  {
+    "mode": "NULLABLE",
+    "name": "first_submission_timestamp",
+    "type": "TIMESTAMP"
+  },
+  {
+    "mode": "NULLABLE",
+    "name": "event_first_seen_date",
+    "type": "DATE"
+  },
```
sean-rose: **suggestion (blocking):** There is no longer an `event_first_seen_date` column.
```suggestion
```

tests/sql/moz-fx-data-shared-prod/firefox_desktop_derived/events_first_seen_v1/test_aggregation/expect.yaml
```diff
@@ -0,0 +1,137 @@
+---
+# Test 1: New event - uses current data
+# - submission_date: 2025-10-29
+  # event_first_seen_date: 2025-10-29
```
sean-rose: **suggestion (non-blocking):** Leaving these previous column values commented out seems unnecessary.  I'd suggest removing them throughout this file.

tests/sql/moz-fx-data-shared-prod/firefox_desktop_derived/events_first_seen_v1/test_init/expect.yaml
```diff
@@ -0,0 +1,126 @@
+---
+# Different events for same client_id
+# - submission_date: 2025-10-29
+  # event_first_seen_date: 2025-10-29
```
sean-rose: **suggestion (non-blocking):** Leaving these previous column values commented out seems unnecessary.  I'd suggest removing them throughout this file.

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
kbammarito: Resolving this comment since we discussed it yesterday with Vanessa

sql_generators/glean_usage/templates/events_first_seen_v1.metadata.yaml
```diff
@@ -0,0 +1,29 @@
...
+description: |-
+  Captures the earliest date that we observe an event for a particular client in the events_stream_v1 table.
+owners:
+  - kbammarito@mozilla.com
+  - vsabino@mozilla.com
+labels:
+  schedule: daily
+  table_type: client_level
+scheduling:
+  dag_name: bqetl_glean_usage
```
sean-rose, line 11: **issue (blocking):** If this is merged as is, the ETL is going to start running as scheduled, which could conflict with the table initialization process that's going to take 7+ hours.

The easiest solution at the moment would probably be to have the ETL be unscheduled initially:
```suggestion
  # Temporarily unschedule this ETL until the table initialization has been able to complete.
  #dag_name: bqetl_glean_usage
```
And then schedule it in a subsequent PR after the table initialization has been completed.

(the same concern would also apply when generating `events_first_seen_v1` ETLs for additional apps with large amounts of events data in the future)

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,174 @@
...
+      normalized_channel,
+      normalized_country_code,
+      normalized_os,
+      normalized_os_version,
+      windows_build_number
+    FROM
+      `{{ project_id }}.{{ events_first_seen_table }}`
+    WHERE
+      DATE(first_submission_timestamp) >= '2023-01-01'
+      AND DATE(first_submission_timestamp) < @submission_date
```
sean-rose, line 142: **issue (blocking):** This `_previous` CTE should only include records for the particular criteria, but it's including all records, which will cause duplicate records to be generated on every ETL run in a compounding manner.

```suggestion
      AND DATE(first_submission_timestamp) < @submission_date
      AND criteria IS NOT DISTINCT FROM {{ item["name"] }}
```

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,174 @@
...
+    ),
+  -- query over all of history to see whether the client_id, event and criteria combination has shown up before
+  _previous AS (
+    SELECT
+      first_submission_timestamp,
+      client_id,
+      `event`,
+      event_category,
+      event_name,
+      {{ item["name"] }} AS criteria,
```
sean-rose: **suggestion (non-blocking):** IMO it would be better to just select the previously saved `criteria` value in this `_previous` CTE.
```suggestion
      criteria,
```

sean-rose (changes requested):

sql_generators/glean_usage/templates/events_first_seen_v1.query.sql
```diff
@@ -0,0 +1,175 @@
...
+      _previous
+      ON _current.client_id = _previous.client_id
+          AND _current.event = _previous.event
+          AND (_current.criteria = _previous.criteria
+              OR (_current.criteria IS NULL AND _previous.criteria IS NULL))
+  )
+SELECT
+  *
+FROM
+  _joined
```
sean-rose, line 166: **suggestion (non-blocking):** This could be changed to only select new records for the targeted date, and then the ETL wouldn't need to overwrite the entire table (which would hopefully improve the ETL performance).

```suggestion
  )
  SELECT
    _current.*
  FROM
    _current
  LEFT JOIN
    _previous
    ON _current.client_id = _previous.client_id
    AND _current.event = _previous.event
    AND (
      _current.criteria = _previous.criteria
      OR (_current.criteria IS NULL AND _previous.criteria IS NULL)
    )
  WHERE
    _previous.client_id IS NULL
```

sql_generators/glean_usage/templates/events_first_seen_v1.metadata.yaml
```diff
@@ -0,0 +1,30 @@
...
+  arguments: ["--billing-project=moz-fx-data-backfill-4"]
+  task_group: {{ app_name }}
+  parameters:
+    - >-
+      {% raw %}submission_date:DATE:{{ds}}{% endraw %}
+  depends_on_past: true
+  date_partition_parameter: null
```
sean-rose, line 17: **suggestion (non-blocking):** If you change the ETL to not overwrite the entire table.
```suggestion
  depends_on_past: true
```

sean-rose (approved):
r+wc

kbammarito:
We think the merge was failing due to this PR being quite large (too many commits to squash). @BenWu squashed it and shortened the commit message. So we made [a new PR](https://github.com/mozilla/bigquery-etl/pull/8638) that cherry picks the squashed commit and merged successfully. Closing this PR.

PR #8050, opened 2025-09-03
DENG-8953--Newtab Clients additional metrics and computations

gkatre:
## Description

This PR includes several additional metrics suitable at the daily client level for Firefox desktop newtab aggregated from the newtab visits table

* DENG-8953

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_clients_daily_v2/query.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_clients_daily_v2/query.sql
@@ -31,7 +31,7 @@ SELECT
     DISTINCT IF(nonsearch_interaction_count > 0, newtab_visit_id, NULL)
   ) AS nonsearch_engagement_visits,
   COUNT(
-    DISTINCT IF(is_content_interaction, newtab_visit_id, NULL)
+    DISTINCT IF(any_content_interaction_count > 0, newtab_visit_id, NULL)
   ) AS any_content_engagement_visits,
   SUM(any_content_click_count) AS any_content_click_count,
   SUM(any_content_impression_count) AS any_content_impression_count,
@@ -41,12 +41,12 @@ SELECT
   SUM(organic_content_click_count) AS organic_content_click_count,
   SUM(organic_content_impression_count) AS organic_content_impression_count,
   COUNT(
-    DISTINCT IF(is_sponsored_content_interaction, newtab_visit_id, NULL)
+    DISTINCT IF(sponsored_content_interaction_count > 0, newtab_visit_id, NULL)
   ) AS sponsored_content_engagement_visits,
   SUM(sponsored_content_click_count) AS sponsored_content_click_count,
   SUM(sponsored_content_impression_count) AS sponsored_content_impression_count,
   COUNT(
-    DISTINCT IF(is_topsite_interaction, newtab_visit_id, NULL)
+    DISTINCT IF(any_topsite_interaction_count > 0, newtab_visit_id, NULL)
   ) AS any_topsite_engagement_visits,
   SUM(any_topsite_click_count) AS any_topsite_click_count,
   SUM(any_topsite_impression_count) AS any_topsite_impression_count,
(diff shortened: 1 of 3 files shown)


sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_clients_daily_v2/query.sql
```diff
@@ -56,14 +56,73 @@ SELECT
...
+    DISTINCT IF(widget_interaction_count > 0, newtab_visit_id, NULL)
+  ) AS widget_engagement_visits,
+  COUNT(
+    DISTINCT IF(other_interaction_count > 0, newtab_visit_id, NULL)
+  ) AS others_engagement_visits,
+  -- New fields added Aug 2025
+  ANY_VALUE(sample_id) AS sample_id,
+  ANY_VALUE(profile_group_id) AS profile_group_id,
+  `moz-fx-data-shared-prod.udf.mode_last`(ARRAY_AGG(geo_subdivision)) AS geo_subdivision,
+  ANY_VALUE(experiments) AS experiments,
```
jaredsnyder, line 73: I'd argue that we should just exclude `experiments` from the table because it can change mid-day.  I would say any metrics that segment by experiment should be derived from the visits table, but I'd like to hear what @ilanasegall has to say on this one

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_clients_daily_v2/query.sql
```diff
@@ -56,14 +56,73 @@ SELECT
...
+  ANY_VALUE(newtab_weather_enabled) AS newtab_weather_enabled,
+  ANY_VALUE(default_search_engine) AS default_search_engine,
+  ANY_VALUE(default_private_search_engine) AS default_private_search_engine,
+  ANY_VALUE(topsite_rows) AS topsite_rows,
+  ANY_VALUE(topsite_sponsored_tiles_configured) AS topsite_sponsored_tiles_configured,
+  ANY_VALUE(newtab_blocked_sponsors) AS newtab_blocked_sponsors,
```
jaredsnyder, line 79: It might be better to use `mode_last` here to get the most common value rather than `ANY_VALUE`

land-edi (approved):
Looks good!

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_clients_daily_v2/schema.yaml
```diff
@@ -127,3 +127,148 @@ fields:
...
+- name: topsite_rows
+  type: INTEGER
+  description: Number of topsite rows shown on newtab page
+- name: topsite_sponsored_tiles_configured
+  type: INTEGER
+  description: Number of sponsored topsite tiles configured
+- name: newtab_blocked_sponsors
+  type: STRING
+  description: List of advertiser domains blocked by the user
+  mode: REPEATED
```
jaredsnyder, line 184: It'd be good to note the aggregation method on these, because it could in theory change mid-day. This is probably rare.  The way we want to do it depends whether we keep using `ANY_VALUE` vs `mode_last`

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_clients_daily_v2/query.sql
```diff
@@ -56,14 +56,73 @@ SELECT
...
+  SUM(sponsored_topsite_interaction_count) AS sponsored_topsite_interaction_count,
+  SUM(content_thumbs_up_count) AS content_thumbs_up_count,
+  SUM(content_thumbs_down_count) AS content_thumbs_down_count,
+  SUM(search_ad_click_count) AS search_ad_click_count,
+  SUM(search_ad_impression_count) AS search_ad_impression_count,
+  SUM(widget_interaction_count) AS widget_interaction_count,
+  SUM(widget_impression_count) AS widget_impression_count,
+  SUM(other_interaction_count) AS other_interaction_count,
+  SUM(other_impression_count) AS other_impression_count,
+  AVG(newtab_visit_duration) AS avg_newtab_visit_duration,
```
jaredsnyder, line 123:
```suggestion
  AVG(newtab_visit_duration) AS avg_newtab_visit_duration,
  SUM(newtab_visit_duration) AS cumulative_newtab_visit_duration,
```

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_clients_daily_v2/schema.yaml
```diff
@@ -127,3 +127,148 @@ fields:
...
+  description: Count of widget impressions
+- name: other_interaction_count
+  type: INTEGER
+  description: Count of other non-content interactions
+- name: other_impression_count
+  type: INTEGER
+  description: Count of other non-content impressions
+- name: avg_newtab_visit_duration
+  type: FLOAT64
+  description: Average duration of newtab visit (ms)
```
jaredsnyder, line 274:
```suggestion
- name: avg_newtab_visit_duration
  type: FLOAT64
  description: Average duration of newtab visit (ms)
- name: cumulative_newtab_visit_duration
  type: FLOAT64
  description: Total duration of newtab visits on this day (ms)
```

jaredsnyder (changes requested):
A few small things

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_clients_daily_v2/schema.yaml
gkatre: > It'd be good to note the aggregation method on these, because it could in theory change mid-day. This is probably rare. 

As I see this, its a higher level aggregation from the visit. I would expect such edge cases to happen for each of the higher level aggregations.

If the analysis you typically care about is more granular, I would think you would be using the visit level table for those use cases.

At this stage perhaps we do not need to be overly concerned about those rare edge cases?

If there is a good justification and a consensus to make changes, we could open a PR if needed.

cc: @ilanasegall

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_clients_daily_v2/query.sql
gkatre: [Commented here](https://github.com/mozilla/bigquery-etl/pull/8050#discussion_r2322699151). This seems related

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_clients_daily_v2/query.sql
gkatre: `cumulative_newtab_visit_duration` appears to be a new metrics request. I will create a new PR for it

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_clients_daily_v2/query.sql
gkatre: This seems somewhat similar to [this](https://github.com/mozilla/bigquery-etl/pull/8050#discussion_r2322699151) and may need to be decided if experiments at higher level granularity is needed. If so, how is it going to be used.

Before we decide to exclude, lets make sure we all understand the usage in all respects. I can create a separate PR if we need to remove the column. We do not need to rush into making these decisions

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_clients_daily_v2/query.sql
jaredsnyder: Thanks!

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_clients_daily_v2/query.sql
jaredsnyder: The difference with experiments is that it is that mid-day changes are much more common, so I don' t think we can dismiss it as an edge case.  In a two week experiment, day 1 and day 14 will both have different values during the day, so that 2/14 days where this value is misleading.  Obviously not everyone is always in an experiment, etc, but we're probably closer to 10% of days than 1% of days on this one.

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_clients_daily_v2/schema.yaml
jaredsnyder: Yeah that's fair, though in the future I think it'd be good if part of the validation was quantifying how often it occurs so we're not going off assumptions.

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_clients_daily_v2/query.sql
ilanasegall: instead of any_value, I'd like to see the most up to date experiment list for a client day. we may miss a tiny portion of an unenrollment day, but this way we get the entire day for an enrollment and don't miss any actions that might be taken at the beginning of an experiment.

gkatre: The reason I was saying "no need to rush" is because removing a column does not involve backfilling. Its only going to be a PR change with a table change to remove a column. 
But if we decide we want the experiment field later, that would involve a backfill.
So for now, until we have everyone onboard, I will leave it there.

jaredsnyder: I agree with Ilana, let's keep it but grab the most recent value rather than `ANY_VALUE`

sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_clients_daily_v2/query.sql
ilanasegall: for weather_enabled, i'd like to return true if there is at least one true for the day

PR #8065, opened 2025-09-05
[DENG-9426] Use local schemas for glean_app_ping_views and implement retry mechanism to get schema

scholtzan:
This implements some optimizations for generated `glean_app_ping_views`. As the `stable_views` generator runs before `glean_usage` and also generates view schemas, these local schemas can be re-used if available. Also some dry runs occasionally fail for getting table schemas. This PR implements a retry mechanism (up to 3 times which seemed enough in my local tests) or raises an exception to avoid the issues described in https://mozilla-hub.atlassian.net/browse/DENG-9426

### DIFF
--- a/sql_generators/glean_usage/glean_app_ping_views.py
+++ b/sql_generators/glean_usage/glean_app_ping_views.py
@@ -19,7 +19,7 @@ from pathos.multiprocessing import ThreadingPool
 
 from bigquery_etl import ConfigLoader
 from bigquery_etl.format_sql.formatter import reformat
-from bigquery_etl.schema import Schema
+from bigquery_etl.schema import Schema, SCHEMA_FILE
 from bigquery_etl.util.common import get_table_dir, write_sql
 from sql_generators.glean_usage.common import GleanTable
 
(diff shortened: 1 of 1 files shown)


sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -114,20 +116,51 @@ def _process_ping(ping_name):
...
+                            )
+                        except Exception as e:
+                            print(
+                                f"Attempt {attempt + 1}: Failed to get schema for {channel_dataset_view}: {e}"
+                            )
+
                 cached_schemas[channel_dataset] = deepcopy(schema)
 
                 if schema.schema["fields"] == []:
                     # check for empty schemas (e.g. restricted ones) and skip for now
```
BenWu: This comment isn't accurate anymore so I think it can be removed.  Is the cloud function able to read every stable table?  I assume it can if this doesn't fail
```suggestion
```

BenWu (approved):

scholtzan:
I'll merge this on Monday, just because it potentially touches quite a few generated views.

PR #7293, opened 2025-04-08
Bug 1957476 - Add attr/dist metrics to client_info in stable views

BenWu:
## Description

This changes the stable views to mirror glean attribution and distribution metrics into `client_info` per [this doc](https://docs.google.com/document/d/1TIZhpBeZcJSEnIZJwj0Cj9saL2Tfs0X6oi_4gFU35eM/edit?tab=t.0#heading=h.dc06cblplbtl).

The metrics are `metrics.object.glean_attribution_ext` and `metrics.object.glean_distribution_ext` and they get added to client info as `client_info.attribution.ext` and `client_info.distribution.ext` as JSON fields.  If the metrics don't exist in the schema, the fields are still added as nulls.

@chutten can you confirm the field names and structure is what you expect?

## Related Tickets & Documents
* https://bugzilla.mozilla.org/show_bug.cgi?id=1957476

**Reviewer, please follow [this checklist](https://github.com/mozilla/bigquery-etl/blob/main/.github/reviewer_checklist.md)**

PR #8066, opened 2025-09-05
Add base client_info.{attribution|distribution} to events_stream

chutten:
## Description

This PR augments events_stream with attribution and distribution information added in https://bugzilla.mozilla.org/show_bug.cgi?id=1955428

It misses the extended attribution and distribution information that is carried in metrics. Probably needs something like #7293  to get those into place.


chutten:
Actually, upon reflection, since events_stream derives from the `events` _view_, it ought to have `ext` without any special handling. This might do it all.

BenWu (approved):
Yes, should do what you want since it's querying the view

Issue #6655, opened 2024-12-11
sqlglot incompatibilites with INFORMATION_SCHEMA identifiers


scholtzan:
A recent update to sqlglot results in `INFORMATION_SCHEMA` table identifiers being parsed differently. The `INFORMATION_SCHEMA` part is now always quoted and parts are interpreted differently. This messes with some of our substitution logic in https://github.com/mozilla/bigquery-etl/blob/9814402cd387a9c13b05c08bf6f87ca6493dfd28/bigquery_etl/util/common.py#L245-L256

The sqlglot change that caused this: https://github.com/tobymao/sqlglot/commit/84f78aafd5d7e74da407167cd394d2bff0718cfb#diff-3cdd876ac935bf47c0213f4f49aac3319401a3f4d0012d58d716d9c814a6e2faR609

We should try updating this logic so we can update to the most recent sqlglot version

data-sync-user:
➤ Rob Miller commented:

Anna Scholtz to engage w sqlglot devs (they’re very active) to see if there can be a fix from their end. If that doesn’t bear fruit, we’ll make this an InfraDay task.

BenWu:
I have a partial fix in https://github.com/mozilla/bigquery-etl/compare/main...benwu/qualify-ref-fix that does fix handling information schemas but it's failing with https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/telemetry_derived/clients_first_seen_v2/checks.sql with the jinja templating which isn't a proper query but it probably means there are some edge cases.  I'll come back to this at some point

BenWu:
Also another approach is seeing if we can get https://sqlglot.com/sqlglot/optimizer/qualify_tables.html to do what we need

data-sync-user:
➤ Sean Rose commented:

IMO the simplest/best way forward at this point would be to simply not try to qualify table references in checks.sql files (especially since those are being deprecated in favor of Bigeye): https://github.com/mozilla/bigquery-etl/pull/8410 ( https://github.com/mozilla/bigquery-etl/pull/8410|smart-link ) 

Then we could use the solution Ben Wu came up with as is.

data-sync-user:
➤ Sean Rose commented:

https://github.com/mozilla/bigquery-etl/pull/8411 ( https://github.com/mozilla/bigquery-etl/pull/8411|smart-link )

PR #8068, opened 2025-09-05
Add support for unquoted hyphenated project names in sql formatter

BenWu:
## Description

This allows the sql formatter to handle things like `moz-fx-data-shared.telemetry.main` without turning it into `moz - fx - data - shared.telemetry.main` 

This might make https://github.com/mozilla/bigquery-etl/issues/6655 easier to deal with

### DIFF
--- a/bigquery_etl/format_sql/tokenizer.py
+++ b/bigquery_etl/format_sql/tokenizer.py
@@ -706,11 +706,17 @@ class Identifier(Token):
     pattern = re.compile(r"[A-Za-z_][A-Za-z_0-9]*|`(?:\\.|[^\\`])+`")
 
 
+class ProjectIdentifier(Identifier):
+    """Identifier for a GCP project, can contain hyphens unlike Identifier."""
+
+    pattern = re.compile(r"[A-Za-z](?:[A-Za-z_0-9-]*[A-Za-z_0-9])?|`(?:\\.|[^\\`])+`")
+
+
 class QualifiedIdentifier(Identifier):
     """Fully or partially qualified identifier for a column, table, or other database object."""
 
     pattern = re.compile(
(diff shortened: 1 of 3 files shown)


scholtzan (approved):

bigquery_etl/format_sql/tokenizer.py
```diff
@@ -706,11 +706,17 @@ class Identifier(Token):
     pattern = re.compile(r"[A-Za-z_][A-Za-z_0-9]*|`(?:\\.|[^\\`])+`")
 
 
+class ProjectIdentifier(Identifier):
+    """Identifier for a GCP project, can contain hyphens unlike Identifier."""
+
+    pattern = re.compile(r"[A-Za-z](?:[A-Za-z_0-9-]*[A-Za-z_0-9])?|`(?:\\.|[^\\`])+`")
```
sean-rose, line 712: I believe the doubled backslashes are incorrect since this is a raw string.  But since ``[^`]`` should also match dots, I think this slightly simplified version without backslashes should work just as well:
```suggestion
    pattern = re.compile(r"[A-Za-z](?:[A-Za-z_0-9-]*[A-Za-z_0-9])?|`[^`]+`")
```

sean-rose: Actually, nevermind, I misinterpreted what that was intending to do.  I thought it was trying to escape the dot, but it's matching/eating potential backslash escape sequences within the backtick-quoted string.  Carry on. 🫡

sean-rose: And it looks like that was copied+modified from `Identifier.pattern` above, which it turns out [I wrote](https://github.com/mozilla/bigquery-etl/blame/d81fb331f5acd8658bc2cfdf35106c50249c1f3e/bigquery_etl/format_sql/tokenizer.py#L706). 😅

PR #8073, opened 2025-09-08
Update schema in events stream script if needed

BenWu:
## Description

Desktop and background update [failing](https://workflow.telemetry.mozilla.org/dags/bqetl_glean_usage/grid?dag_run_id=scheduled__2025-09-07T02%3A00%3A00%2B00%3A00&task_id=firefox_desktop_background_update.firefox_desktop_background_update_derived__events_stream__v1&tab=logs) because https://github.com/mozilla/bigquery-etl/pull/8066 added columns. This script uses a copy job which doesn't support `schema_update_options` so they can't add the columns like the regular queries.  I already manually updated the desktop schema because it has downstream things.

### DIFF
--- a/sql_generators/glean_usage/templates/events_stream_v1.query.py
+++ b/sql_generators/glean_usage/templates/events_stream_v1.query.py
@@ -64,7 +64,7 @@ def main(submission_date, billing_project, destination_table, temp_dataset, slic
             query=query,
             job_config=bigquery.QueryJobConfig(
                 destination=temp_table,
-                write_disposition=bigquery.WriteDisposition.WRITE_TRUNCATE,
+                write_disposition=bigquery.WriteDisposition.WRITE_APPEND,
                 query_parameters=[
                     bigquery.ScalarQueryParameter(
                         "min_sample_id", "INT64", min_sample_id
@@ -76,6 +76,7 @@ def main(submission_date, billing_project, destination_table, temp_dataset, slic
                         "submission_date", "DATE", submission_date.date()
                     ),
                 ],
+                schema_update_options=bigquery.SchemaUpdateOption.ALLOW_FIELD_ADDITION,
             ),
         )
         logging.info(


sql_generators/glean_usage/templates/events_stream_v1.query.py
```diff
@@ -77,12 +78,19 @@ def main(submission_date, billing_project, destination_table, temp_dataset, slic
...
-        print(
+        logging.info(
             f"Writing sample id {min_sample_id} to {max_sample_id} to {temp_table}: {job.path}"
         )
         job.result()
         temp_tables.append(job.destination)
 
+    # update destination table schema if the source has changed
+    temp_table_schema = client.get_table(temp_tables[0]).schema
+    if not temp_table_schema != destination_table.schema:
```
sean-rose: This condition seems like it would only pass if the schemas are the same (double negative = positive)?

BenWu: Oops 🤦

I had it as `not Schema.from_bigquery_schema(temp_table_schema).equals...` first

sql_generators/glean_usage/templates/events_stream_v1.query.py
```diff
@@ -77,12 +78,19 @@ def main(submission_date, billing_project, destination_table, temp_dataset, slic
...
-        print(
+        logging.info(
             f"Writing sample id {min_sample_id} to {max_sample_id} to {temp_table}: {job.path}"
         )
         job.result()
         temp_tables.append(job.destination)
 
+    # update destination table schema if the source has changed
+    temp_table_schema = client.get_table(temp_tables[0]).schema
+    if temp_table_schema != destination_table.schema:
```
sean-rose, line 90: One other minor concern is if we ever have column descriptions in this table, presumably the temp table won't have those column descriptions, so this method of comparing schemas would detect that as a difference.

However, there currently aren't column descriptions, so we can deal with that issue when the time comes.

sean-rose (approved):
r+wc

sql_generators/glean_usage/templates/events_stream_v1.query.py
BenWu: That's a good point.  I just tested this and it retains descriptions if I change it to  `WriteDisposition.WRITE_APPEND` with `SchemaUpdateOption.ALLOW_FIELD_ADDITION` so I'll do that.  The bqetl `Schema.equals` would also work

PR #8083, opened 2025-09-09
Replace dry run check in glean_usage generator to check referenced tables exist

scholtzan:
This PR speeds up glean_usage generation by a few minutes by reducing the number of dryruns when checking if referenced tables exist. Instead, a different mechanism is used to determine whether all upstream tables exist, based on the schema existence of configured "baseline" tables.
This PR also implements local caching for pipeline schemas and reduces some other API calls.


bigquery_etl/schema/stable_table_schema.py
```diff
@@ -68,6 +71,23 @@ def prod_schemas_uri():
...
+    # check if cached file exists and load it
+    if os.path.exists(cache_file):
+        print(f"Loading cached schemas from {cache_file}")
+        try:
+            with open(cache_file, "rb") as f:
+                return pickle.load(f)
+        except (pickle.PickleError, EOFError, OSError) as e:
+            print(f"Failed to load cached schemas: {e}, re-downloading...")
+
+    print(f"Downloading schemas from {schemas_uri}")
```
scholtzan, line 90: I noticed that multiple generators need access to the schemas and download them anew multiple times (likely related to them running in different processes). Caching the schemas in a local file helps with that

bqetl_project.yaml
```diff
@@ -410,6 +410,32 @@ generate:
...
+      - experimenter_cirrus
+      - firefox_desktop_background_defaultagent
+      - firefox_desktop_background_tasks
+      - firefox_crashreporter
+      - monitor_cirrus
+      - mozilla_vpn
+      - mozillavpn_backend_cirrus
+      - pine
+      - thunderbird_android
+      - thunderbird_desktop
```
scholtzan, line 438: Turns out we didn't create `metrics_clients_last_seen` and `clients_last_seen_joined` tables for these apps that got added recently as the schema can't be generated automatically due to the query referencing itself. It seems no one was missing those tables, so maybe we can just keep skipping their generation for now.

sql_generators/glean_usage/templates/metrics_clients_last_seen.metadata.yaml
```diff
@@ -30,6 +30,9 @@ bigquery:
       - sample_id
 monitoring:
   enabled: {% if enable_monitoring %}true{% else %}false{% endif %}
+schema:
+  derived_from:
+  - table: ['moz-fx-data-shared-prod', '{{ app_name }}', 'metrics_clients_daily_v1']
```
scholtzan, line 35: This is what was blocking the schemas from getting generated automatically. For newly added apps `metrics_clients_last_seen` and `clients_last_seen_joined` should get deployed automatically now.

sql_generators/glean_usage/common.py
```diff
@@ -149,24 +149,6 @@ def table_names_from_baseline(baseline_table, include_project_id=True):
...
-            404 == e.get("code")
-            or (
-                403 == e.get("code")
-                and "bigquery.tables.create denied" not in e.get("message")
-            )
-            for e in dryrun.errors()
-        ]
-    )
-
-
```
scholtzan, line 169: This check was taking most of the time due to the dry run. This PR implements a different way of checking whether certain `glean_usage` tables should get generated based on the existence of the baseline tables.

bigquery_etl/schema/stable_table_schema.py
```diff
@@ -132,4 +152,12 @@ def get_stable_table_schemas() -> List[SchemaFile]:
         )
     ]
 
+    # Cache the processed schemas
+    try:
+        with open(cache_file, "wb") as f:
+            pickle.dump(schemas, f)
+        print(f"Cached schemas to {cache_file}")
+    except (pickle.PickleError, OSError) as e:
+        print(f"Failed to cache schemas: {e}")
```
BenWu, line 161: Did you see errors when testing this?  Would this be from things like multiple processes trying to write to same file at the same time?

sql_generators/derived_view_schemas/__init__.py
```diff
@@ -8,6 +8,7 @@
 
 from bigquery_etl.cli.utils import use_cloud_function_option
 from bigquery_etl.dryrun import get_id_token
+from bigquery_etl.schema.stable_table_schema import get_stable_table_schemas
```
BenWu: Was this left in by mistake?

sql_generators/glean_usage/__init__.py
```diff
@@ -159,6 +159,25 @@ def get_tables(table_name="baseline_v1"):
         for baseline_table in get_tables(table_name=table.base_table_name)
     ]
 
+    baseline_tables = {}
+    unique_baseline_table_names = {table.base_table_name for table in GLEAN_TABLES}
```
BenWu: Should this be `base_table` instead of `baseline`? Baseline is confusing to me since `events_v1` would also be here

bigquery_etl/schema/stable_table_schema.py
scholtzan: I don't think I saw this during testing, but it could be possible that due multiple processes are trying to write to the same cache file simultaneously

sql_generators/glean_usage/__init__.py
scholtzan: Yeah good point. I updated the naming

PR #8084, opened 2025-09-10
fix(events_stream): Avoid mixing up experiment `type` and `enrollment_id` data (DENG-9632)

sean-rose:
## Description
Currently it appears 56 apps have event pings where the `ping_info.experiments[].value.extra` field is of type `STRUCT<type STRING, enrollment_id STRING>`, while 31 apps have event pings where the `ping_info.experiments[].value.extra` field is of type `STRUCT<enrollment_id STRING, type STRING>`, and for the latter cases the current events stream ETL logic is unfortunately reversing the experiment `type` and `enrollment_id` values (DENG-9632).

This PR fixes the events stream ETL so both flavors of event ping experiments schema are handled properly.

* DENG-9632: Events stream ETL can mix up experiment type and enrollment ID data

### DIFF
--- a/sql_generators/glean_usage/templates/events_stream_v1.query.sql
+++ b/sql_generators/glean_usage/templates/events_stream_v1.query.sql
@@ -23,21 +23,6 @@ RETURNS json AS (
   )
 );
 
--- convert array of key value pairs to a json object
--- values are nested structs and will be converted to json objects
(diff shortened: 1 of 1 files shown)


akkomar (approved):
LGTM, I ran an isolated test.
Test failures seem temporary, restarted CI.

PR #8087, opened 2025-09-10
bug(DENG-9614): remove schema.yaml files for BigEye API data

phil-lee70:
## Description

This PR fixes the bug that occurs when BigEye updates their API and adds new fields which results in the `schema.yaml` file not matching what is in BigQuery. This breaks the pipeline and causes bugs every few weeks like [this one](phil-schema-9614).

After some discussion, there were 2 options:

1) Update the `query.py` script to drop any new columns and if there are new columns that need to be added, manually update the schema. 
2) Remove `schema.yaml` files altogether since the schema is determined by the source.

This PR chooses options 2, though not ideally best practice, it is less complex to maintain and the tables involved are small (~1k-2k rows).

* DENG-9614


sql/moz-fx-data-shared-prod/bigeye_derived/collection_service_v1/metadata.yaml
```diff
@@ -10,6 +10,7 @@ labels:
   table_type: aggregate
 scheduling:
   dag_name: bqetl_bigeye_derived
+  arguments: ['--schema_update_option=ALLOW_FIELD_ADDITION']
```
BenWu, line 13: These are argument for the query.py.  The script doesn't accept this argument so this shouldn't be here. You would want to put this in the in the `LoadJobConfig` but it's already doing this it's `WRITE_TRUNCATE` on an unpartitioned table.

PR #8096, opened 2025-09-11
Bug 1988207 - Look for schema in output_dir in glean_app_ping_views

BenWu:
## Description

Fun series of bugs causing publish_view errors:
- dry runs in `stable_views` sometimes timeout and cause no schema.yaml to be created
- the existing schema lookup uses `/sql` but CI outputs the generated sql to `/tmp/workspace/generated-sql/sql/` so it's never used in CI
- the views that don't have schema.yaml will get schemas created from dry run (`Schema.for_table`) which doesn't get descriptions
- `generate_compatible_select_expression` looks at descriptions in the comparison so every field that has a description is considered missing from the view with no schema.yaml

* https://bugzilla.mozilla.org/show_bug.cgi?id=1988207
 
### DIFF
--- a/sql_generators/glean_usage/glean_app_ping_views.py
+++ b/sql_generators/glean_usage/glean_app_ping_views.py
@@ -117,11 +117,13 @@ class GleanAppPingViews(GleanTable):
                     continue
 
                 # look for schema in output_dir because bqetl generate all runs stable_views first
-                existing_schema_path = (
+                sql_dir = (
                     output_dir
-                    / channel_dataset
-                    / view_name
-                    / SCHEMA_FILE
+                    or Path(ConfigLoader.get("default", "sql_dir", fallback="sql"))
+                    / project_id
+                )
+                existing_schema_path = (
+                    sql_dir / channel_dataset / view_name / SCHEMA_FILE
                 )
 
                 schema = None


bigquery_etl/schema/__init__.py
```diff
@@ -363,16 +363,26 @@ def recurse_fields(
...
-                        or len(node_path) > max_unnest_depth > 0
-                        # field not in unnest allowlist
-                        or (
-                            unnest_allowlist is not None
-                            and node_path[0] not in unnest_allowlist
+                    if (
+                        dtype == source_schema_nodes[node["name"]]["type"]
+                        and (
+                            # fields must match for records, otherwise only type needs to match
+                            # note that this will be false if field descriptions don't match
```
BenWu, line 370: This is fixable but I don't feel like spending time on this since this is the existing behaviour anyway

sql_generators/glean_usage/glean_app_ping_views.py
```diff
@@ -116,10 +116,9 @@ def _process_ping(ping_name):
                 if channel_dataset_view in ignored_pings:
                     continue
 
-                sql_dir = Path(ConfigLoader.get("default", "sql_dir", fallback="sql"))
+                # look for schema in output_dir because bqetl generate all runs stable_views first
                 existing_schema_path = (
-                    sql_dir
-                    / "moz-fx-data-shared-prod"
+                    output_dir
```
scholtzan:
```suggestion
                # look for schema in output_dir because bqetl generate all runs stable_views first
                sql_dir = output_dir or Path(ConfigLoader.get("default", "sql_dir", fallback="sql")) / project_id
                existing_schema_path = (
                    sql_dir
                    / project_id
```
We might want to handle the case where `output_dir` hasn't been set

PR #8105, opened 2025-09-16
Parallelize dependency record, metadata update, monitoring update (generate-sql CI task)

scholtzan:
## Description

This PR parallelizes the following commands that run as part of the `generate-sql` CI task:
* `./bqetl dependency record`
* `./bqetl metadata update`
* `./bqetl monitoring update`

This shaves off around 2.5 minutes from the `generate-sql` CI task (e.g. compare with https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/52810/workflows/910fe13d-e797-4b56-964d-4bb9eda82430/jobs/655557).

### DIFF
--- a/bigquery_etl/cli/metadata.py
+++ b/bigquery_etl/cli/metadata.py
@@ -75,7 +75,7 @@ def update(
     )
 
     # group table metadata files by dataset
-    datasets = {}
+    datasets: dict[Path, list[Path]] = {}
     for table_metadata_file in table_metadata_files:
         dataset_path = Path(table_metadata_file).parent.parent
         if dataset_path not in datasets:
--- a/bigquery_etl/cli/metadata.py
+++ b/bigquery_etl/cli/metadata.py
@@ -74,122 +74,139 @@ def update(
         "deprecation", "retain_dataset_roles", fallback=[]
     )
 
+    # group table metadata files by dataset
+    datasets = {}
+    for table_metadata_file in table_metadata_files:
+        dataset_path = Path(table_metadata_file).parent.parent
+        if dataset_path not in datasets:
+            datasets[dataset_path] = []
+        datasets[dataset_path].append(table_metadata_file)
+
+    # process each dataset in parallel
     if parallelism > 1:
         with Pool(parallelism) as pool:
             pool.map(
-                partial(_update_single_metadata_file, retained_dataset_roles),
-                table_metadata_files,
+                partial(_update_dataset_metadata, retained_dataset_roles),
+                datasets.items(),
             )
     else:
-        for table_metadata_file in table_metadata_files:
-            _update_single_metadata_file(retained_dataset_roles, table_metadata_file)
+        for dataset_path, dataset_table_files in datasets.items():
+            _update_dataset_metadata(
+                retained_dataset_roles, (dataset_path, dataset_table_files)
+            )
 
     return None
 
 
-def _update_single_metadata_file(retained_dataset_roles, table_metadata_file):
-    """Update a single metadata file."""
-    try:
-        dataset_metadata_path = (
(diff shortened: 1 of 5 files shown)


bigquery_etl/dependency.py
```diff
@@ -224,9 +296,18 @@ def show(paths: Tuple[str, ...], without_views: bool):
...
     help="Skip files with existing references rather than failing",
 )
-def record(paths: Tuple[str, ...], skip_existing):
+@click.option(
+    "--parallelism",
+    "-p",
+    default=8,
+    type=int,
+    help="Number of threads for parallel processing",
+)
```
scholtzan, line 311: Can't use `parallelism_option` here as it causes a circular import error:

```
  File "<string>", line 1, in <module>
  File "/bigquery_etl/dependency.py", line 16, in <module>
    from bigquery_etl.cli.utils import parallelism_option
  File "/bigquery_etl/cli/__init__.py", line 13, in <module>
    from ..cli.backfill import backfill
  File "/bigquery_etl/cli/backfill.py", line 49, in <module>
    from ..cli.query import backfill as query_backfill
  File "/bigquery_etl/cli/query.py", line 48, in <module>
    from ..dependency import get_dependency_graph
ImportError: cannot import name 'get_dependency_graph' from partially initialized module 'bigquery_etl.dependency' (most likely due to a circular import)
```

Redefining this option was the easiest solution here

scholtzan:
The diff shows some changes to `partition_column_set` which is a Bigeye configuration. The change to them being set to `false` looks correct to me (not sure why it was true before). None of these tables have partition columns configured and the default is setting that value to false.

bigquery_etl/cli/metadata.py
```diff
@@ -168,8 +190,8 @@ def update(name: str, sql_dir: Optional[str], project_id: Optional[str]) -> None
         if table_metadata_updated:
             table_metadata.write(table_metadata_file)
             click.echo(f"Updated {table_metadata_file}")
-
-    return None
+    except Exception as e:
+        click.echo(f"Error processing {table_metadata_file}: {e}", err=True)
```
BenWu: I think this should raise the exception since that's what's happening now.  Continuing through failure might cause bugs that are hard to track down

bigquery_etl/cli/metadata.py
```diff
@@ -61,22 +61,44 @@ def metadata(ctx):
...
-def update(name: str, sql_dir: Optional[str], project_id: Optional[str]) -> None:
+@parallelism_option()
+def update(
+    name: str, sql_dir: Optional[str], project_id: Optional[str], parallelism: int
+) -> None:
     """Update metadata yaml file."""
-    table_metadata_files = paths_matching_name_pattern(
-        name, sql_dir, project_id=project_id, files=["metadata.yaml"]
+    table_metadata_files = list(
+        set(
```
BenWu: Is it possible for paths_matching_name_pattern to return duplicates?  It would be a bit cleaner to put this logic in the function instead of each caller

scholtzan: Duplicates get added in `paths_matching_name_pattern` because the code path:
```python
for root, _, _ in os.walk(pattern, followlinks=True):
            for file in files:
                matching_files.extend(
                    map(Path, glob(f"{root}/**/{file}", recursive=True))
                )
```
iterates of every directory and subdirectory and searches for specific files (metadata.yaml in this case) in each of them. E.g. files get added for `sql/moz-fx-cjms-prod-f3c7` (searches all subdirectories due to `/**/`) and also for `sql/moz-fx-cjms-prod-f3c7/cjms_bigquery`.
I moved the deduplication logic into the function itself.

kwindau (approved):

BenWu (approved):

PR #8110, opened 2025-09-17
feat(DENG-9303): add enterprise metrics v1 aggregate dataset

kik-kik:
# feat(DENG-9303): add enterprise metrics v1 aggregate dataset

### DIFF
--- a/sql/moz-fx-data-shared-prod/firefox_desktop_derived/enterprise_metrics_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/firefox_desktop_derived/enterprise_metrics_v1/metadata.yaml
@@ -16,7 +16,4 @@ bigquery:
     field: submission_date
     require_partition_filter: false
     expiration_days: null
-  clustering:
-    fields:
-    - normalized_channel
 deprecated: false


sql/moz-fx-data-shared-prod/firefox_desktop_derived/enterprise_metrics_v1/query.sql
```diff
@@ -0,0 +1,11 @@
+SELECT
+  submission_date,
+  normalized_channel,
+  distribution_id,
```
kik-kik: Is there any way in which we could bucket distribution_id together? There's a lot of different possible values which causes each date to have many a lot of entries.

pissac17: I think we can exclude the distribution id in the aggregate table and keep it in the client level table.

kik-kik: Removed as per our discussions. I also added a filter in the client level dataset to only include esr and release channel clients since those appear to be the only ones relevant to enterprise.

pissac17 (approved):

PR #8117, opened 2025-09-18
Bug 1989142 - Set fixed date for event_monitoring_live view

BenWu:
## Description

event_monitoring_aggregates has been broken since 2025-09-16 so the [looker dashboard](https://mozilla.cloud.looker.com/dashboards/1452?Event+Name=%22extensions.data.storage_local_error%22&App+Name=Firefox+for+Android&Window+Start+Time=14+days&Channel=release) shows 0 events.  Switching to materialized views will allow it to show data until it gets fixed 

* https://bugzilla.mozilla.org/show_bug.cgi?id=1989142


scholtzan (approved):

sql_generators/glean_usage/templates/event_monitoring_live.view.sql
```diff
@@ -21,7 +21,9 @@ CREATE OR REPLACE VIEW
         FROM
           `{{ project_id }}.{{ dataset['bq_dataset_family'] }}_derived.event_monitoring_live_v1`
         WHERE
-          DATE(submission_date) > DATE_SUB(CURRENT_DATE(), INTERVAL 2 DAY)
+          -- workaround for event_monitoring_aggregates outage https://bugzilla.mozilla.org/show_bug.cgi?id=1989142
+          -- DATE(submission_date) > DATE_SUB(CURRENT_DATE(), INTERVAL 2 DAY)
+          DATE(submission_date) > '2025-09-15'
```
sean-rose, line 26: Sorry for the late comment, but for a more robust workaround would it perhaps make sense to use Jinja to continuously update the hard-coded dates in these views, [like is done in the `event_monitoring_live_v1` materialized views](https://github.com/mozilla/bigquery-etl/blob/0a4b046b3c1d8c14ed1993387244ca23b21be665/sql_generators/glean_usage/templates/event_monitoring_live_v1.materialized_view.sql#L63)?

BenWu: Can you explain what makes it more robust?  I'm setting it specifically to that date because we don't have data after that.  The materialized views are set that way because they doesn't support `CURRENT_DATE` since it's not deterministic

sean-rose: Ah, nevermind, I misunderstood the nature of this fix, and thought we might have to periodically update the date for performance reasons (like in the materialized views).  But now I get this is a hard cutoff for specifically when the upstream ETL stopped working recently.  My apologies for the unwarranted comment!

PR #8126, opened 2025-09-19
fix(DENG-9617): re-deploy experiment_events_live_v1 materialized view

mikewilli:
## Description

#8116 updated `experiment_events_live_v1` materialized view, but it did not get re-deployed, which appears to be due to the `CREATE IF NOT EXISTS` statement. This changes that to `CREATE OR REPLACE` so it will get re-deployed.

* DENG-9617

### DIFF
--- a/sql_generators/experiment_monitoring/templates/experiment_events_live_v1/materialized_view.sql
+++ b/sql_generators/experiment_monitoring/templates/experiment_events_live_v1/materialized_view.sql
@@ -1,7 +1,6 @@
 -- Generated via ./bqetl generate experiment_monitoring
-CREATE MATERIALIZED VIEW
-IF
-  NOT EXISTS {{ dataset }}_derived.{{ destination_table }}
+CREATE OR REPLACE MATERIALIZED VIEW
+  {{ dataset }}_derived.{{ destination_table }}
   PARTITION BY DATE(partition_date)
   OPTIONS
     (enable_refresh = TRUE, refresh_interval_minutes = 5)


mikewilli:
The real issue here is that we explicitly don't re-deploy materialized views, and this needs to be done manually for now.

Also, we don't use this `OR REPLACE` syntax for any other materialized views, so I'm going to close this PR to keep that consistent here as well.

sql_generators/experiment_monitoring/templates/experiment_events_live_v1/materialized_view.sql
```diff
@@ -1,7 +1,6 @@
 -- Generated via ./bqetl generate experiment_monitoring
-CREATE MATERIALIZED VIEW
-IF
-  NOT EXISTS {{ dataset }}_derived.{{ destination_table }}
+CREATE OR REPLACE MATERIALIZED VIEW
```
sean-rose, line 2: Some code assumes all materialized views start with `CREATE MATERIALIZED VIEW` ([example 1](https://github.com/mozilla/bigquery-etl/blob/08bd44a73ca223361ef9f3ccef1c2c6a07fb99d6/bigquery_etl/cli/stage.py#L175), [example 2](https://github.com/mozilla/bigquery-etl/blob/08bd44a73ca223361ef9f3ccef1c2c6a07fb99d6/bigquery_etl/dryrun.py#L218)), so I wouldn't recommend merging this as is.

And as Ben pointed out in another comment, we're intentionally not automatically redeploying materialized views (with the obvious drawback that redeploying them requires manual intervention).

sean-rose (changes requested):

PR #8128, opened 2025-09-19
feat(fxa): monitor key fetch tokens table

clouserw:
## Description

This PR adds monitoring for the key_fetch_tokens table.  Not sure why it wasn't already there. It's been a while since I did this but I think this is all I need to add.  /cc @sean-rose 

Thanks!

* none

### DIFF
--- a/sql/moz-fx-data-shared-prod/accounts_backend_derived/monitoring_db_counts_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/accounts_backend_derived/monitoring_db_counts_v1/query.sql
@@ -90,6 +90,15 @@ WITH table_counts AS (
     )
   UNION ALL
   SELECT
+    'key_fetch_tokens' AS table_name,
+    COUNT(*) AS total_rows
+  FROM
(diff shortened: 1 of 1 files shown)


sql/moz-fx-data-shared-prod/accounts_backend_derived/monitoring_db_counts_v1/query.sql
```diff
@@ -89,6 +89,15 @@ WITH table_counts AS (
       'UTC'
     )
   UNION ALL
+  SELECT
+    'key_fetch_tokens' AS table_name,
```
sean-rose: **suggestion (non-blocking):** Since it seems these initial straightforward monitoring subqueries are in alphabetical order by `table_name` value, I think ideally this `key_fetch_tokens` subquery would go before the `linked_accounts` subquery above.

clouserw: fixed!

PR #8130, opened 2025-09-20
DENG-9727: Added mode last struct retain nulls udf

wwyc:
## Description

Added a `mode_last_struct_retain_nulls` UDF that returns the most frequent STRUCT in an array; if there’s a tie, it selects the latest occurrence. Use this to pick a single, self-consistent set of related fields (e.g., city, subdivision(s), country) together, rather than aggregating each field separately.  Retain nulls.

* DENG-9727

### DIFF
--- a/sql/mozfun/map/mode_last_struct_retain_nulls/udf.sql
+++ b/sql/mozfun/map/mode_last_struct_retain_nulls/udf.sql
@@ -30,222 +30,194 @@ CREATE OR REPLACE FUNCTION map.mode_last_struct_retain_nulls(entries ANY TYPE) A
 -- Tests
 SELECT
   -- 1) Most frequent wins (Berlin appears twice)
-  assert.equals(
-    TO_JSON_STRING(
-      STRUCT(
-        'Berlin' AS city,
-        'BE' AS subdivision1,
-        CAST(NULL AS STRING) AS subdivision2,
-        'DE' AS country
-      )
+  assert.struct_equals(
+    STRUCT(
+      'Berlin' AS city,
+      'BE' AS subdivision1,
+      CAST(NULL AS STRING) AS subdivision2,
+      'DE' AS country
     ),
-    TO_JSON_STRING(
-      map.mode_last_struct_retain_nulls(
-        [
-          STRUCT(
-            'Berlin' AS city,
-            'BE' AS subdivision1,
-            CAST(NULL AS STRING) AS subdivision2,
-            'DE' AS country
-          ),
-          STRUCT(
-            'Munich' AS city,
-            'BY' AS subdivision1,
-            CAST(NULL AS STRING) AS subdivision2,
-            'DE' AS country
-          ),
-          STRUCT(
-            'Berlin' AS city,
-            'BE' AS subdivision1,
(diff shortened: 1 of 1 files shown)


sql/mozfun/map/mode_last_struct_retain_nulls/udf.sql
```diff
@@ -0,0 +1,251 @@
...
+      last_pos DESC
+    LIMIT
+      1
+  )
+);
+
+-- Tests
+SELECT
+  -- 1) Most frequent wins (Berlin appears twice)
+  assert.equals(
```
BenWu: You should be able to use `assert.struct_equals` for all of these which will do `TO_JSON_STRING` for you and make this easier to read and edit.
```suggestion
  assert.struct_equals(
```

sql/mozfun/map/mode_last_struct_retain_nulls/udf.sql
```diff
@@ -0,0 +1,251 @@
...
+            'Berlin' AS city,
+            'BE' AS subdivision1,
+            CAST(NULL AS STRING) AS subdivision2,
+            'DE' AS country
+          )   -- latest among the tied
+        ]
+      )
+    )
+  ),
+-- 6) NULL struct occurs most frequently -> expect NULL
```
BenWu, line 167: This makes sense for this udf but for https://github.com/mozilla/bigquery-etl/pull/7974, wouldn't we want to consider Berlin as the first/last seen city in this case?

BenWu (commented):
Is there a difference in output with `stats.mode_last_retain_nulls`?  I remembered that bigquery recently(?) added support for grouping by structs so it's possible that `stats.mode_last_retain_nulls` works as-is.  Can you try putting these test cases in that udf?

sql/mozfun/map/mode_last_struct_retain_nulls/udf.sql
wwyc: This UDF would first be applied to stable tables to produce one row per client per day (mirroring baseline_clients_daily). We should retain NULLs at this stage: clients in cities with populations <15k or in locations MaxMind can’t map should remain NULL. If we drop NULLs here, we’d misrepresent a client’s true location and only capture them when they travel to a resolvable city. Downstream, the city_seen table can then keep only the non-NULL city values after this step.  I hope that makes sense?

wwyc:
> Is there a difference in output with `stats.mode_last_retain_nulls`? I remembered that bigquery recently(?) added support for grouping by structs so it's possible that `stats.mode_last_retain_nulls` works as-is. Can you try putting these test cases in that udf?

I tried putting these test cases in `stats.mode_last_retain_nulls` and they all passed.  So I think I can just use that udf and close this PR?

BenWu:
Yeah this one can be closed if the other udf does what you need

sql/mozfun/map/mode_last_struct_retain_nulls/udf.sql
BenWu: I'm not sure that matches what I was thinking so I might be misunderstanding where this is going to be used.  I'll see if it makes sense when I look at where the udf is used

PR #8135, opened 2025-09-22
feat: build win10 user targeting for braze to start ip warming

LiamMcFall:
## Description

This PR creates the basis of what we need to target 14 day inactive windows 10 firefox users. 

* DENG-9051

### DIFF
--- a/sql/moz-fx-data-shared-prod/subscription_platform_derived/stripe_logical_subscriptions_history_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/subscription_platform_derived/stripe_logical_subscriptions_history_v1/query.sql
@@ -77,14 +77,6 @@ plan_services AS (
   GROUP BY
     plan_id
 ),
-paypal_subscriptions AS (
-  SELECT DISTINCT
-    subscription_id
-  FROM
-    `moz-fx-data-shared-prod.stripe_external.invoice_v1`
-  WHERE
-    JSON_VALUE(metadata, '$.paypalTransactionId') IS NOT NULL
-),
 subscriptions_history_charge_summaries AS (
   SELECT
     history.id AS subscriptions_history_id,
@@ -226,7 +218,11 @@ SELECT
       FORMAT_TIMESTAMP('%FT%H:%M:%E6S', history.subscription_first_active_at)
     ) AS id,
     'Stripe' AS provider,
-    IF(paypal_subscriptions.subscription_id IS NOT NULL, 'PayPal', 'Stripe') AS payment_provider,
+    IF(
+      history.subscription.collection_method = 'send_invoice',
+      'PayPal',
+      'Stripe'
+    ) AS payment_provider,
     history.subscription.id AS provider_subscription_id,
     subscription_item.id AS provider_subscription_item_id,
     history.subscription.created AS provider_subscription_created_at,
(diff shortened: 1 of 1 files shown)


LiamMcFall:
I need to be able to to query `moz-fx-data-shared-prod.accounts_db_external.fxa_emails_v1` in order to get the email addresses for FX accounts that don't already have an existing Braze user associated with it. As such, I can't currently run this query in full until I have that access.

LiamMcFall:
I will open a ticket to get myself access to `workgroup:accounts-confidential` so that I can get this query fully tested/productionalized.

sql/moz-fx-data-shared-prod/braze_derived/fxa_win10_users_v1/metadata.yaml
```diff
@@ -0,0 +1,12 @@
+friendly_name: Fxa Win10 Users
```
chelseyklein:
```suggestion
friendly_name: Mozilla Account Win10 Users
```
Thinking since we switched from using Fx Account to Mozilla Account we might want to rename this...something like mza in place of fxa? Might complicate things too much...so it's up to you :)

LiamMcFall:
@chelseybeck 

OKAY! I have made some changes.

Ultimately while I don't think we _need_ an intermediate table, I decided that I would like to have one to separate the steps of what is happening:

1. `braze_derived.fxa_win10_users_daily_v1` looks at a single day based on submission date and builds the list of windows 10 users who have been inactive for 14 days. I removed the previous logic that looked at 14-21 days since this will likely run for multiple days before the first email is sent. Looking at a bigger window of days would just make the table less efficient to build on subsequent runs. This table will only have the inactive users for the most recent submission date.

2. `braze_derived.fxa_win10_users_historical_v1` will be our table that contains ALL users who have been added to the list. It will be partitioned by submission date and each day will look at `braze_derived.fxa_win10_users_daily_v1` and add any new users to the list that are not already contained in the table based on their FxA id. This table will be our historical record of all users who will have been added to the email group. It also generates external ids for any users that do not already have one. If a user reactivates and then goes inactive for 14 days at any point they will still only show up in this table the FIRST time they go inactive. If we get approval for follow up emails, this logic can be adjusted. Additionally, this table is currently storing only locale en-US as that is the only locale we want to sync while doing IP warming. As we approach launch date for the full campaign, October 14 2025, we will add in the rest of the desired locales.

3. `braze_external.win10_users_sync_v1` will be the table that actually gets synced into Braze. Each day it will be overwritten with the users who have been added to `braze_derived.fxa_win10_users_historical_v1` for that submission date.

The daily and historical tables could hypothetically be a single table, but I like the separation of their functions. I think it also allows us more flexibility in terms of what we will need to change if approval is granted to send follow up emails.

LiamMcFall:
- braze_derived.test_fxa_win10_users_daily_v1
- braze_derived.test_fxa_win10_users_historical_v1
- braze_external.test_win10_users_sync_v1

The above tables can all be view in BQ. I ran a few days worth of backfill and made sure that we did not have duplicate in the historical table. `test_fxa_win10_users_daily_v1` and `test_win10_users_sync_v1` will be as of 9/29 submission date, and `test_fxa_win10_users_historical_v1` has data for 9/27-9/29.

sql/moz-fx-data-shared-prod/braze_derived/fxa_win10_users_v1/metadata.yaml
LiamMcFall: yeah I think I want to leave this as is for right now. I'm open to that change, but I also think I am trying to keep things as consistent as possible with how this project is being talked about which has currently been all in terms of Fx Accounts....

chelseybeck (approved):
great work, @LiamMcFall 🙌

PR #8140, opened 2025-09-22
[DENG-9732] Filter out remotecontent events from event etl

BenWu:
## Description

This filters the [high volume `uptake.remotecontent.result` events](https://bugzilla.mozilla.org/show_bug.cgi?id=1989348) from legacy and glean event tables. This might fix the [`event_monitoring_aggregates_v1` failures](https://bugzilla.mozilla.org/show_bug.cgi?id=1989142) but I'm still waiting for the queries to finish.

I also rearranged the `event_monitoring_aggregates_v1` query a bit to aggregate events per ping before aggregating per app since I found that it uses slightly less slot time and shuffle 

* DENG-9732

### DIFF
--- a/bqetl_project.yaml
+++ b/bqetl_project.yaml
@@ -300,6 +300,8 @@ view:
     - sql/moz-fx-data-shared-prod/activity_stream/tile_id_types/view.sql
     - sql/moz-fx-data-shared-prod/pocket/pocket_reach_mau/view.sql
     - sql/moz-fx-data-shared-prod/telemetry/buildhub2/view.sql
+    # https://bugzilla.mozilla.org/show_bug.cgi?id=1990074
+    - sql/moz-fx-data-shared-prod/monitoring/event_monitoring_live/view.sql
     # Dataset glam-fenix-dev:glam_etl was not found
     # TODO: this should be removed if views are to be automatically deployed
     - sql/glam-fenix-dev/glam_etl/**/view.sql
--- a/sql/moz-fx-data-shared-prod/firefox_desktop_derived/enterprise_metrics_clients_v1/backfill.yaml
+++ b/sql/moz-fx-data-shared-prod/firefox_desktop_derived/enterprise_metrics_clients_v1/backfill.yaml
@@ -4,7 +4,7 @@
   reason: Need to redo backfill to make sure the data matches the query.
   watchers:
   - kik@mozilla.com
-  status: Initiate
+  status: Complete
   shredder_mitigation: false
   override_retention_limit: false
 
--- a/sql/moz-fx-data-shared-prod/firefox_desktop_derived/enterprise_metrics_v1/backfill.yaml
+++ b/sql/moz-fx-data-shared-prod/firefox_desktop_derived/enterprise_metrics_v1/backfill.yaml
@@ -1,0 +1,9 @@
+2025-09-23:
+  start_date: 2023-08-11
+  end_date: 2025-09-23
+  reason: Initial backfill of the dataset.
+  watchers:
+  - kik@mozilla.com
+  status: Initiate
+  shredder_mitigation: false
+  override_retention_limit: false
--- a/sql/moz-fx-data-shared-prod/firefox_desktop_derived/report_content_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/firefox_desktop_derived/report_content_v1/query.sql
@@ -10,12 +10,12 @@ WITH events AS (
     mozfun.map.get_key(event.extra, 'topic') AS topic,
     mozfun.map.get_key(event.extra, 'url') AS url
   FROM
-    `moz-fx-data-shared-prod.firefox_desktop.newtab` AS e
+    `moz-fx-data-shared-prod.firefox_desktop.newtab_content` AS e
   CROSS JOIN
     UNNEST(e.events) AS event
   WHERE
     DATE(submission_timestamp) = @submission_date
-    AND event.category = 'newtab'
+    AND event.category = 'newtab_content'
     AND event.name = 'report_content_submit'
 )
 SELECT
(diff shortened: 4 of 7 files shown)


sql/moz-fx-data-shared-prod/telemetry_derived/event_events_v1/query.sql
```diff
@@ -53,30 +53,34 @@ FROM
...
     event_category = 'security'
     AND event_method = 'unexpectedload'
     AND mozfun.map.get_key(event_map_values, 'contenttype') = 'TYPE_STYLESHEET'
-    AND mozfun.norm.truncate_version(app_version, 'major')
+    AND mozfun.norm.extract_version(app_version, 'major')
     BETWEEN 84
     AND 87
-  )
-  AND
+  ) IS FALSE
```
sean-rose: **issue:** Using `IS FALSE` means that the event will be excluded if one or more of the fields being checked is null (including event extras being checked not being present) while the rest of the non-null field conditions match.  

If that isn't the intended behavior then I'd recommend using `IS NOT TRUE` instead (same goes for the other usages of `IS FALSE`).

sql_generators/glean_usage/templates/event_monitoring_aggregates_v1.query.sql
```diff
@@ -8,17 +8,26 @@ WITH
...
+        @submission_date AS submission_date,
+        TIMESTAMP_ADD(
+          TIMESTAMP_TRUNC(submission_timestamp, HOUR),
+            -- Aggregates event counts over 60-minute intervals
+          INTERVAL(DIV(EXTRACT(MINUTE FROM submission_timestamp), 60) * 60) MINUTE
+        ) AS window_start,
+        TIMESTAMP_ADD(
+          TIMESTAMP_TRUNC(submission_timestamp, HOUR),
+          INTERVAL((DIV(EXTRACT(MINUTE FROM submission_timestamp), 60) + 1) * 60) MINUTE
+        ) AS window_end,
```
sean-rose: **suggestion (non-blocking):** I see this was existing code just being moved into this CTE, but this logic doesn't make sense to me since the extracted minute will always be less than 60, in which case `DIV(EXTRACT(MINUTE FROM submission_timestamp), 60)` would always evaluate to zero, and this could be simplified:

```suggestion
        TIMESTAMP_TRUNC(submission_timestamp, HOUR) AS window_start,
        TIMESTAMP_ADD(
          TIMESTAMP_TRUNC(submission_timestamp, HOUR),
          INTERVAL 1 HOUR
        ) AS window_end,
```

Or am I missing something?

sql_generators/glean_usage/templates/event_monitoring_aggregates_v1.query.sql
```diff
@@ -30,22 +39,35 @@ WITH
...
+        {% endif %}
+      GROUP BY
+        submission_date,
+        window_start,
+        window_end,
+        event_category,
+        event_name,
+        event_extra_key,
+        country,
+        normalized_app_name,
```
sean-rose: **suggestion (blocking):** I presume trying to group by `normalized_app_name` would fail given you removed that column from this CTE above.
```suggestion
```

sql_generators/glean_usage/templates/event_monitoring_aggregates_v1.query.sql
```diff
@@ -30,22 +39,35 @@ WITH
...
+        channel,
+        version,
+        experiment,
+      experiment_branch
```
sean-rose: **nitpick:** Indentation.
```suggestion
        experiment,
        experiment_branch
```

sql_generators/glean_usage/templates/event_monitoring_aggregates_v1.query.sql
```diff
@@ -30,22 +39,35 @@ WITH
...
-        INTERVAL(DIV(EXTRACT(MINUTE FROM submission_timestamp), 60) * 60) MINUTE
-      ) AS window_start,
-      TIMESTAMP_ADD(
-        TIMESTAMP_TRUNC(submission_timestamp, HOUR),
-        INTERVAL((DIV(EXTRACT(MINUTE FROM submission_timestamp), 60) + 1) * 60) MINUTE
-      ) AS window_end,
-      * EXCEPT (submission_timestamp),
-      COUNT(*) AS total_events,
+      "{{ dataset['canonical_app_name'] }}" AS normalized_app_name,
+      * REPLACE (SUM(total_events) AS total_events),
```
sean-rose: **issue:** This is changing the order of the columns being returned, which seems like something we should avoid if we can.

```suggestion
      submission_date,
      window_start,
      window_end,
      event_category,
      event_name,
      event_extra_key,
      country,
      "{{ dataset['canonical_app_name'] }}" AS normalized_app_name,
      channel,
      version,
      experiment,
      experiment_branch,
      SUM(total_events) AS total_events,
```

sql_generators/glean_usage/templates/events_stream_v1.query.sql
```diff
@@ -153,6 +153,12 @@ CROSS JOIN
         AND event.name = 'enrollment_status'
         AND app_version_major = 140
       )
+      -- See https://mozilla-hub.atlassian.net/browse/DENG-9732
+      AND (
+        event.category = "uptake.remotecontent.result"
+        AND event.name IN ("uptake_remotesettings", "uptake_normandy")
+        AND mozfun.norm.extract_version(client_info.app_display_version, 'major') >= 143
```
sean-rose: **suggestion (non-blocking):** Here we have an `app_version_major` column readily available.
```suggestion
        AND app_version_major >= 143
```

sql_generators/glean_usage/templates/event_monitoring_aggregates_v1.query.sql
BenWu: Yes it's equivalent.  I'll change it

sql_generators/glean_usage/templates/event_monitoring_aggregates_v1.query.sql
BenWu: The query works because `normalized_app_name` is in the source table but yes this should be removed

sean-rose (approved):

PR #8157, opened 2025-09-24
feat: unschedule app_store_choice_screen_selection_v1 for now

kik-kik:
# feat: unschedule app_store_choice_screen_selection_v1 for now

This is because there appears to be a wider issue currently around the browser choice data with ongoing discussions. Until then descheduling this task to reduce the noise.


kwindau (commented):
LGTM

PR #8162, opened 2025-09-25
Bug 1990074 - Exclude deprecated apps from event monitoring

BenWu:
## Description

I tested and this allows `monitoring.event_monitoring_live` to be deployed from airflow-gke-prod

Removes
- pine
- org_mozilla_fenix_nightly
- org_mozilla_fennec_aurora
- org_mozilla_tv_firefox
- org_mozilla_vrbrowser
- mozilla_lockbox
- org_mozilla_ios_lockbox
- burnham
- org_mozilla_connect
- org_mozilla_firefoxreality
- org_mozilla_bergamot
- firefox_translations
- viu_politica
- treeherder

* https://bugzilla.mozilla.org/show_bug.cgi?id=1990074

### DIFF
--- a/sql_generators/glean_usage/event_monitoring_live.py
+++ b/sql_generators/glean_usage/event_monitoring_live.py
@@ -96,17 +96,11 @@ class EventMonitoringLive(GleanTable):
     ):
         # Cache app_info to avoid repeated calls
         cached_app_info = get_app_info()
-        
+
         # Get the app ID from the baseline_table name.
         # This is what `common.py` also does.
         app_id = re.sub(r"_stable\..+", "", baseline_table)
         app_id = ".".join(app_id.split(".")[1:])
-
-        # Skip any not-allowed app.
-        if app_id in ConfigLoader.get(
-            "generate", "glean_usage", "events_monitoring", "skip_apps", fallback=[]
-        ):
-            return
 
         tables = table_names_from_baseline(baseline_table, include_project_id=False)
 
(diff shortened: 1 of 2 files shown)


sql_generators/glean_usage/event_monitoring_live.py
```diff
@@ -121,6 +115,23 @@ def generate_per_app_id(
             "generate", "glean_usage", "events_monitoring", "events_tables", fallback={}
         )
 
+        deprecated = [
+            app_dataset.get("deprecated", False)
+            for _, app in cached_app_info.items()
+            for app_dataset in app
+            if dataset == app_dataset["bq_dataset_family"]
+        ][0] is True
```
sean-rose: **question:** Are there any cases where there multiple results in this list?

If so, I wonder if using `all()` might be more appropriate than just checking the first result.

sql_generators/glean_usage/event_monitoring_live.py
```diff
@@ -121,6 +115,23 @@ def generate_per_app_id(
             "generate", "glean_usage", "events_monitoring", "events_tables", fallback={}
         )
 
+        deprecated = [
+            app_dataset.get("deprecated", False)
+            for _, app in cached_app_info.items()
```
sean-rose: **suggestion (non-blocking):**
```suggestion
            for app in cached_app_info.values()
```

sean-rose (approved):
r+wc

sql_generators/glean_usage/event_monitoring_live.py
BenWu: Yes it's one entry per channel so this is checking the first channel which happens to always be release right now but I don't think there's anything enforcing that.

Deprecation can be set at both the app and channel level but there aren't any instances where it's set in the channel and there's more than one channel.  So setting `all()` won't change anything right now but it still makes sense to do

PR #8167, opened 2025-09-26
[DENG-6243] Use event_monitoring_live in aggregates query

BenWu:
## Description

Implementing DENG-6243 to use the materialized view that has the same data because I don't want to keep trying to make the query use less memory.  This fixed the query in my tests and only took 11 minutes to run.  

This makes event_monitoring_aggregates have duplicates from the live table but I found that duplicates are <0.1% based on https://sql.telemetry.mozilla.org/queries/110886/source which is a lot lower than I thought.

* DENG-6243
* https://bugzilla.mozilla.org/show_bug.cgi?id=1989142


sql_generators/glean_usage/templates/event_monitoring_aggregates_v1.query.sql
```diff
@@ -5,44 +5,106 @@ WITH
...
-            event.category = "uptake.remotecontent.result"
-            AND event.name IN ("uptake_remotesettings", "uptake_normandy")
-            AND mozfun.norm.extract_version(client_info.app_display_version, 'major') >= 143
-            AND sample_id != 0
-          ) IS NOT TRUE
-        {% endif %}
+        DATE(submission_date) = @submission_date
+    )
+  {% else %}
+    {% for events_table in event_tables_per_dataset[dataset['bq_dataset_family']] -%}
```
BenWu, line 31: below this is whitespace changes so setting hide whitespace will make it easier to read

sql_generators/glean_usage/templates/event_monitoring_aggregates_v1.metadata.yaml
```diff
@@ -13,6 +13,14 @@ scheduling:
   arguments: [
     "--billing-project", "moz-fx-data-backfill-3"
   ]
+  {% if manual_refresh_apps -%}
+  depends_on:
+  {%- for app in manual_refresh_apps %}
+    - dag_name: bqetl_materialized_view_refresh
+      task_id: {{ app }}_derived__event_monitoring_live__v1
+      execution_delta: -23h45m  # wait for the hourly run at 1:45
```
sean-rose: **suggestion (blocking):** The execution delta is applied to the data interval starts of the runs, which for the 01:45 hourly run would be 00:45, so to get from 02:00 the previous day to 00:45 the next day would be 22:45.
```suggestion
      execution_delta: -22h45m  # wait for the hourly run at 1:45
```

sean-rose: With @BenWu out and the `firefox_desktop_derived.event_monitoring_aggregates_v1` ETL failing again, I'm going ahead and merging this PR with my suggested code change.

sean-rose (approved):

PR #8201, opened 2025-10-03
feat(SubPlat): Implement ended_reason for Google and Apple logical subscriptions (DENG-4043)

phil-lee70:
## Description
Create `ended_reason` field for Google and Apple logical subscriptions.

* DENG-4043


sql/moz-fx-data-shared-prod/subscription_platform_derived/apple_logical_subscriptions_history_v1/query.sql
```diff
@@ -284,6 +284,26 @@ SELECT
       WHEN 5
         THEN 'Revoked'
     END AS provider_status,
+    -- API Docs enumerations :
+    -- https://developer.apple.com/documentation/appstoreserverapi/status
+    -- https://developer.apple.com/documentation/appstoreserverapi/expirationintent
+    CASE
+      WHEN history.subscription_is_active IS NOT TRUE
+        THEN NULL
```
sean-rose: **suggestion (blocking):** This is making `ended_reason` null for all subscriptions which _have_ ended, which is the reverse of what you want.
```suggestion
      WHEN history.subscription_is_active
        THEN NULL
```

sql/moz-fx-data-shared-prod/subscription_platform_derived/apple_logical_subscriptions_history_v1/query.sql
```diff
@@ -284,6 +284,26 @@ SELECT
...
+      WHEN (
+          history.subscription.status = 2
+          AND history.subscription.renewal_info.expiration_intent IN (1, 3)
+          OR history.subscription.status = 5
+        )
+        THEN 'User Initiated'
```
sean-rose: **suggestion (blocking):**
```suggestion
      WHEN (
          history.subscription.status = 2
          AND history.subscription.renewal_info.expiration_intent IN (1, 3)
        )
        OR history.subscription.status = 5
        THEN 'User Initiated'
```

sql/moz-fx-data-shared-prod/subscription_platform_derived/apple_logical_subscriptions_history_v1/query.sql
```diff
@@ -284,6 +284,26 @@ SELECT
...
+      WHEN (
+          history.subscription.status = 2
+          AND history.subscription.renewal_info.expiration_intent = 2
+          OR history.subscription.status = 3
+        )
+        THEN 'Payment Failure'
```
sean-rose: **suggestion (blocking):**
```suggestion
      WHEN (
          history.subscription.status = 2
          AND history.subscription.renewal_info.expiration_intent = 2
        )
        OR history.subscription.status = 3
        THEN 'Payment Failure'
```

sql/moz-fx-data-shared-prod/subscription_platform_derived/apple_logical_subscriptions_history_v1/query.sql
```diff
@@ -284,6 +284,26 @@ SELECT
...
+      ELSE 'Other'
+    END AS ended_reason,
```
sean-rose: **issue (blocking):** The `ended_reason` field should be immediately after `ended_at`, like it is in the `schema.yaml`.

sql/moz-fx-data-shared-prod/subscription_platform_derived/apple_logical_subscriptions_history_v1/query.sql
```diff
@@ -284,6 +284,26 @@ SELECT
...
         THEN 'Revoked'
     END AS provider_status,
+    -- API Docs enumerations :
+    -- https://developer.apple.com/documentation/appstoreserverapi/status
+    -- https://developer.apple.com/documentation/appstoreserverapi/expirationintent
+    CASE
+      WHEN history.subscription_is_active IS NOT TRUE
+        THEN NULL
+      WHEN (
+          history.subscription.status = 2
```
sean-rose: **nitpick:** In the existing ETL logic I added trailing comments wherever inscrutable API values were being used in conditions to explain what they meant, and IMO it would be nice to keep doing that consistently (linking to the docs is good, but saving devs a trip to the docs is even more helpful).
```suggestion
          history.subscription.status = 2  -- 2 = expired
```
(ditto for all the other numeric API values used in the conditions you're adding here)

sql/moz-fx-data-shared-prod/subscription_platform_derived/apple_logical_subscriptions_history_v1/query.sql
```diff
@@ -284,6 +284,26 @@ SELECT
...
+          AND history.subscription.renewal_info.expiration_intent IN (1, 3)
+          OR history.subscription.status = 5
```
sean-rose: **suggestion (non-blocking):** Since how to treat revoked Apple subscriptions was non-obvious and something we had to clarify by asking others, IMO that would warrant an explanatory line comment above this line (e.g. explaining that admins aren't currently revoking Apple subscriptions so we can assume all such cases are from the users themselves).

sql/moz-fx-data-shared-prod/subscription_platform_derived/apple_logical_subscriptions_history_v1/schema.yaml
```diff
@@ -175,6 +175,11 @@ fields:
     description: |-
       When the subscription ended.
       This will be null for active subscriptions.
+  - name: ended_reason
+    type: STRING
+    mode: NULLABLE
+    description: |-
+      Reason why the subscription ended.
```
sean-rose, line 182: **suggestion (non-blocking):** Since the values for the `ended_reason` field are entirely novel, I think it would be helpful to list the possible values with a description for each (similar to how I did that for some of the provider-specific fields in the changelog ETLs):
```suggestion
      Reason why the subscription ended.

      Possible values:
        * `Admin Initiated` - An admin at Mozilla ended the subscription, either directly or by deleting the user's Mozilla Account.
        * `User Initiated` - The user ended their subscription, either directly or by turning off auto-renewal or deleting their Mozilla Account.
        * `Payment Failure` - The subscription was unable to auto-renew due to a payment failure.
        * `Other` - The subscription ended for some other reason, such as unverified Mozilla Accounts being automatically deleted.
```
(if we do this, the `ended_reason` descriptions should be kept the same across all the logical subscription tables, including the Stripe one)

sql/moz-fx-data-shared-prod/subscription_platform_derived/google_logical_subscriptions_history_v1/query.sql
```diff
@@ -179,6 +179,18 @@ SELECT
       history.mozilla_account_id,
       TO_HEX(SHA256(history.mozilla_account_id)) AS mozilla_account_id_sha256,
       history.subscription.country_code,
+      CASE
+        WHEN history.subscription_is_active IS NOT TRUE
+          THEN NULL
```
sean-rose: **suggestion (blocking):**
```suggestion
        WHEN history.subscription_is_active
          THEN NULL
```

sql/moz-fx-data-shared-prod/subscription_platform_derived/google_logical_subscriptions_history_v1/query.sql
```diff
@@ -179,6 +179,18 @@ SELECT
...
+        WHEN history.subscription.cancel_reason = 0
+          THEN 'User Initiated'
+        WHEN history.subscription.cancel_reason = 1
+          AND history.subscription.payment_state = 0
+          THEN 'Payment Failure'
```
sean-rose: **issue (blocking):** Unfortunately this appears to be a messy situation:
- While a fair number of Google subscriptions _eventually_ end up with `cancel_reason = 1` (system canceled) and `payment_state = 0` (payment pending), almost none of them have that combination of values _immediately_ after becoming inactive based on their expiry time being in the past (I only see one such subscription thus far).
  - Looking at Google subscriptions that just became inactive where `cancel_reason` is null and `payment_state = 0` (payment pending), this appears to be a case of late-arriving `cancel_reason` data:
    - 17% of those later end up with `cancel_reason = 0` (user canceled).
    - 74% of those later end up with `cancel_reason = 1` (system canceled).
    - 9% of those never got a `cancel_reason` value.
  - I believe the main impact this will have is on the "Subscription End" events, which are based on the subscription's data immediately after it becomes inactive.  However, depending on how late-arriving the `cancel_reason` data is, this might also impact the daily/monthly logical subscription records to some extent.
- [Google's docs for the `paymentState` field](https://developers.google.com/android-publisher/api-ref/rest/v3/purchases.subscriptions#SubscriptionPurchase.FIELDS.payment_state) specifically say "_Not present for canceled, expired subscriptions_", though clearly we do have plenty of expired Google subscription records with values in that field (and I'm pretty sure there isn't a SubPlat bug like there was for the Apple Firestore records). 🤷 

So I suggest we just categorize all subscriptions with `cancel_reason = 1` (system cancelled) as "Payment Failure", but I'm on the fence about how to handle the cases where `cancel_reason` is null.  I can think of two approaches that could make sense...
1. Fall back to categorizing those cases where `cancel_reason` is null as "Other":
   ```suggestion
           WHEN history.subscription.cancel_reason = 1
             THEN 'Payment Failure'
   ```
2. Preemptively categorize those cases where `cancel_reason` is null and `payment_state = 0` (payment pending) as "Payment Failure", even though that will probably end up being wrong ~17-26% of the time:
   ```suggestion
           WHEN history.subscription.cancel_reason = 1
             OR (
               history.subscription.cancel_reason IS NULL
               AND history.subscription.payment_state = 0
             )
             THEN 'Payment Failure'
   ```

I'd probably lean toward option 1 as being slightly less misleading, but what do you think?

sql/moz-fx-data-shared-prod/subscription_platform_derived/google_logical_subscriptions_history_v1/query.sql
```diff
@@ -179,6 +179,18 @@ SELECT
...
+        WHEN history.subscription.cancel_reason = 3
+          THEN 'Admin Initiated'
+        ELSE 'Other'
+      END AS ended_reason,
```
sean-rose: **issue (blocking):** The `ended_reason` field should be immediately after `ended_at`, like it is in the `schema.yaml`.

sql/moz-fx-data-shared-prod/subscription_platform_derived/google_logical_subscriptions_history_v1/query.sql
phil-lee70: I am thinking reason 1 as well - it seems less ambiguous and wrong 17-26% of the time is a bit too much

sql/moz-fx-data-shared-prod/subscription_platform_derived/apple_logical_subscriptions_history_v1/query.sql
```diff
@@ -300,6 +300,30 @@ SELECT
...
+        )
+        -- admins are not revoking Apple subscriptions so we can assume such cases are from users
+        OR history.subscription.status = 5 -- 5 = auto-renewable subscription is revoked
+        THEN 'User Initiated'
+      WHEN (
+          history.subscription.status = 2 -- 2 = auto-renewable subscription is expired
+          AND history.subscription.renewal_info.expiration_intent = 2 -- 2 = Billing error
+        )
+        OR history.subscription.status = 3 -- 3 = auto-renewable subscription is in a billing retry period
+        THEN 'Payment Failure'
```
sean-rose, line 324: **nitpicks:**
- IMO putting two spaces before end-of-line comments is better for readability.
- Though Apple's docs say "_The auto-renewable subscription is ..._" in all the `status` value descriptions, I think we could omit that repetitive part for brevity without sacrificing clarity, as we're only dealing with auto-renewable subscriptions here.
- It looks like `bqetl format` kind of mangled the `expiration_intent IN (1, 3)` line with the addition of your end-of-line comments, so I'd suggest an alternate layout that's more readable.
- The comment for `expiration_intent = 3` isn't quite accurate.
```suggestion
      WHEN (
          history.subscription.status = 2  -- 2 = expired
          AND history.subscription.renewal_info.expiration_intent IN (
            1,  -- 1 = customer canceled their subscription
            3   -- 3 = customer didn’t consent to a price increase or conversion that requires their consent
          )
        )
        -- admins are not revoking Apple subscriptions so we can assume such cases are from users
        OR history.subscription.status = 5  -- 5 = revoked
        THEN 'User Initiated'
      WHEN (
          history.subscription.status = 2  -- 2 = expired
          AND history.subscription.renewal_info.expiration_intent = 2  -- 2 = Billing error
        )
        OR history.subscription.status = 3  -- 3 = in a billing retry period
        THEN 'Payment Failure'
```

sql/moz-fx-data-shared-prod/subscription_platform_derived/google_logical_subscriptions_history_v1/query.sql
```diff
@@ -238,6 +238,19 @@ SELECT
...
+      -- https://developers.google.com/android-publisher/api-ref/rest/v3/purchases.subscriptions#SubscriptionPurchase.FIELDS.cancel_reason
+      CASE
+        WHEN history.subscription_is_active
+          THEN NULL
+        WHEN history.subscription.cancel_reason = 0 -- 0 = User canceled the subscription
+          THEN 'User Initiated'
+        WHEN history.subscription.cancel_reason = 1 -- 1 = Subscription was canceled by the system, for example because of a billing problem
+          THEN 'Payment Failure'
+        WHEN history.subscription.cancel_reason = 3 -- 3 = Subscription was canceled by the developer
+          THEN 'Admin Initiated'
```
sean-rose: **nitpick:** IMO putting two spaces before end-of-line comments is better for readability.
```suggestion
        WHEN history.subscription.cancel_reason = 0  -- 0 = User canceled the subscription
          THEN 'User Initiated'
        WHEN history.subscription.cancel_reason = 1  -- 1 = Subscription was canceled by the system, for example because of a billing problem
          THEN 'Payment Failure'
        WHEN history.subscription.cancel_reason = 3  -- 3 = Subscription was canceled by the developer
          THEN 'Admin Initiated'
```

sql/moz-fx-data-shared-prod/subscription_platform_derived/google_logical_subscriptions_history_v1/schema.yaml
```diff
@@ -173,6 +173,11 @@ fields:
     description: |-
       When the subscription ended.
       This will be null for active subscriptions.
+  - name: ended_reason
+    type: STRING
+    mode: NULLABLE
+    description: |-
+      Reason why the subscription ended.
```
sean-rose, line 180: **suggestion (blocking):** The `ended_reason` description here should match the updated `apple_logical_subscriptions_history_v1.subscription.ended_reason` description.

(and the same goes for the `stripe_logical_subscriptions_history_v1.subscription.ended_reason` description, though updating that one could wait until a subsequent PR when you're passing the `ended_reason` field through to downstream ETLs & views)

sean-rose (approved):
r+wc

PR #8209, opened 2025-10-06
Lower initialization parallelism and run in backfill slots

BenWu:
## Description

Related to https://github.com/mozilla/bigquery-etl/pull/7974 and https://github.com/mozilla/bigquery-etl/pull/8046 which have large backfills that would be simpler in the is_init than in a managed backfill since they can run in a single query.  The current parallelism is fairly high at 10 since it's unlikely to run 10 queries without slot contention, and smaller queries will be fast anyway so parallelism wouldn't matter as much.  Lowering the parallelism to reduce the chances of running out of memory or query timeouts.  Also moving to backfill reservation so it doesn't get in the way of etl if it runs overnight.

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -80,7 +80,7 @@ QUERY_NAME_RE = re.compile(r"(?P<dataset>[a-zA-z0-9_]+)\.(?P<name>[a-zA-z0-9_]+)
 VERSION_RE = re.compile(r"_v[0-9]+")
 DESTINATION_TABLE_RE = re.compile(r"^[a-zA-Z0-9_$]{0,1024}$")
 DEFAULT_DAG_NAME = "bqetl_default"
-DEFAULT_INIT_PARALLELISM = 10
+DEFAULT_INIT_PARALLELISM = 3
 DEFAULT_CHECKS_FILE_NAME = "checks.sql"
 VIEW_FILE = "view.sql"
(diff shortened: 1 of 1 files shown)


bigquery_etl/cli/query.py
```diff
@@ -1599,7 +1599,7 @@ def _initialize(query_file):
                         dataset=dataset,
                         query_file=query_file,
                         arguments=arguments,
-                        parallelism=parallelism,
+                        parallelism=2,
```
sean-rose: **question (blocking):** Is this perhaps a hard-coded testing value that got left in accidentally?

BenWu: No it's intentionally set to 2 because I don't think sample_id parallelism should be tied to the table parallelism.  And 2 sample ids at a time seems appropriate.

sean-rose (approved):

bigquery_etl/cli/query.py
sean-rose: OK, thanks for clarifying.  It might be nice to have that as a global variable setting like `INIT_SAMPLE_ID_PARALLELISM` alongside `DEFAULT_INIT_PARALLELISM` for clarity, and since the two settings interact (init parallelism * sample ID parallelism = max possible init parallelism).

BenWu: Sure I'll add that

PR #8231, opened 2025-10-09
fix(mdn): handle duplicate UTM parameters

caugner:
## Description

Updates the `mdn.legacy_action` and `mdn.legacy_page` views to keep only the last occurrence of each UTM key (later one wins).

Resolves the `Scalar subquery produced more than one element` error with [this query](https://sql.telemetry.mozilla.org/queries/96339) when using `mdn.legacy_page` instead of `mdn_yari.page`.

* (none)


sql/moz-fx-data-shared-prod/mdn/legacy_action/view.sql
```diff
@@ -102,10 +102,20 @@ CREATE OR REPLACE VIEW
...
+                SELECT
+                  REGEXP_EXTRACT(kv, r'^utm_(.*?)=') AS key,
+                  REGEXP_EXTRACT(kv, r'=(.*)$') AS value,
+                  off
+                FROM
+                  UNNEST(REGEXP_EXTRACT_ALL(JSON_VALUE(event_extra.url), r'[?&](utm_[^&]+)')) AS kv
+                  WITH OFFSET off
+              )
+            QUALIFY
+              ROW_NUMBER() OVER (PARTITION BY key ORDER BY off DESC) = 1
```
LeoMcA, line 118: I had to read up on this, and it seems correct: I presume you've tested it?

Is it possible to do this a bit more simply with something like `LIMIT 1` in the subquery?

caugner: I did indeed test it.

Here's a potential alternative using `ARRAY_AGG()` and `GROUP BY` on the outer query:

```sql
SELECT
  key,
  (ARRAY_AGG(value ORDER BY off DESC LIMIT 1))[OFFSET(0)] AS value
FROM (
  SELECT
    REGEXP_EXTRACT(kv, r'^utm_(.*?)=') AS key,
    REGEXP_EXTRACT(kv, r'=(.*)$') AS value,
    off
  FROM UNNEST(REGEXP_EXTRACT_ALL(JSON_VALUE(event_extra.url), r'[?&](utm_[^&]+)')) AS kv
  WITH OFFSET off
)
GROUP BY key
```

Is this what you had in mind?

LeoMcA: It's not really any simpler is it! Approved as is 🚀

PR #8285, opened 2025-10-17
fix: fix the win10 targeting model to correctly account for inactive accounts

LiamMcFall:
## Description

This PR corrects a bug in our windows 10 user targeting.
- accounts were being pulled from the submission_date. It should have been pulling from pings ~14 days ago.
- an account could show up multiple times in users_services_last_seen. This has been updated to remove any account from syncing that had any activity in users_services_last_seen from within the last 2 weeks.

The erroneous data from 2025-10-01 to 2025-10-14 will remain in this table to ensure that users who have already been emailed but may qualify again in the future do not get re-added to the table. This table will correctly sync starting with a submission date of 2025-10-16.

* DENG-9051

### DIFF
--- a/sql/moz-fx-data-shared-prod/braze_derived/fxa_win10_users_historical_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/braze_derived/fxa_win10_users_historical_v1/query.sql
@@ -34,30 +34,43 @@ WITH win10_users AS (
   FROM
     `moz-fx-data-shared-prod.firefox_desktop.fx_accounts`
   WHERE
-    DATE(submission_timestamp) = @submission_date
+    DATE(submission_timestamp) = DATE_SUB(@submission_date, INTERVAL 14 day)
     AND `mozfun.norm.windows_version_info`(
       client_info.os,
       client_info.os_version,
       client_info.windows_build_number
     ) = 'Windows 10'
 ),
-last_seen_14_days AS (
-  SELECT
-    user_id_sha256,
-    MIN(days_seen_bits) AS last_seen_min
+seen_within_14_days AS (
+  SELECT DISTINCT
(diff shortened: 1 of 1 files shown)


sql/moz-fx-data-shared-prod/braze_derived/fxa_win10_users_historical_v1/query.sql
```diff
@@ -34,30 +34,43 @@ WITH win10_users AS (
...
-    last_seen_min = 16384
+    AND mozfun.bits28.active_in_range(days_seen_bits, -13, 14)
+),
+last_seen_14_days AS (
+  SELECT DISTINCT
+    last_seen.user_id_sha256,
+    last_seen.submission_date
+  FROM
+    `moz-fx-data-shared-prod.accounts_backend_derived.users_services_last_seen_v1` AS last_seen
+  LEFT JOIN
```
chelseyklein, line 59: should this be an inner join? looks like you want to filter on `seen_recently` but a left join would include users not in that CTE, right?

LiamMcFall: That's actually what I want. Sometimes users show up multiple times (across different services) for the same submission date in users_services_last_seen_v1. So I want to exclude any that show up in `seen_recently` at all

LiamMcFall: So `last_seen_14_days` is the full scope of superset potential accounts to target based on inactivity and then I want to make sure I am excluding any that may have shown up anywhere in the last 2 weeks so I am finding those users in the `seen_within_14_days` subset and doing that join in order to remove them from the superset.

PR #8288, opened 2025-10-19
Add openai package to BQETL

kwindau:
## Description

This PR adds the Python package `openai` to the BQETL repo. This package will be used to help summarize information as part of the "Market Intel Bot" project.


kwindau:
> Neither approve nor deny, but it seems odd to me to add a new package that isn't being used. What's the reasoning for adding this without any functionality depending on it?

@mikewilli `openai` depends on `distro`, `httpcore` and `httpx`, and `httpx` depends on `anyio`.  So basically it just added all the dependencies it needed that it didn't have to add `openai`.  

Also I just wanted to make sure people were OK with me adding the package before I add in the code I have that actually uses the package.

mikewilli:
>@mikewilli openai depends on distro, httpcore and httpx, and httpx depends on anyio. So basically it just added all the dependencies it needed that it didn't have to add openai.

I understand that `openai` has dependencies of its own getting added here, I just meant `openai` itself. It's being added but not used anywhere in the code.

>Also I just wanted to make sure people were OK with me adding the package before I add in the code I have that actually uses the package.

I'd suggest we don't add packages independent of functionality. As an example, if we were adding a CLI to an existing app that didn't have one, we wouldn't add `click` in a PR and then go add the CLI functionality later. It's hard to determine whether its worth including without knowing why.

If you're interested in gathering feedback before spending the time implementing functionality, perhaps a proposal would be a better initial route?

kwindau:
> > @mikewilli openai depends on distro, httpcore and httpx, and httpx depends on anyio. So basically it just added all the dependencies it needed that it didn't have to add openai.
> 
> I understand that `openai` has dependencies of its own getting added here, I just meant `openai` itself. It's being added but not used anywhere in the code.
> 
> > Also I just wanted to make sure people were OK with me adding the package before I add in the code I have that actually uses the package.
> 
> I'd suggest we don't add packages independent of functionality. As an example, if we were adding a CLI to an existing app that didn't have one, we wouldn't add `click` in a PR and then go add the CLI functionality later. It's hard to determine whether its worth including without knowing why.
> 
> If you're interested in gathering feedback before spending the time implementing functionality, perhaps a proposal would be a better initial route?

I can just submit this change to add the package in the PR with the actual code using the package if that's the preferred method, no worries! Will close this and add it into the other PR.

PR #5659, opened 2024-05-24
[DENG-3889] Rewrite events_stream_v1 for better performance

BenWu:
DENG-3889

Changes:
- replace two (out of five) of the javascript udfs usages with a sql udf
- rearrange ctes so most of the udfs are called per event ping rather than on the unnested events

### performance

1% sample (sample_id = 1) on firefox desktop for 2024-05-22:
changes | slot hours | job id (backfill-2)
-- | -- | --
base | 24.5 | bquxjob_7a85e33a_18fab4919e3
rearranged | 5.7 | bquxjob_4ababcd5_18fab3b62ea
sql udfs | 4.3 | bquxjob_129253ac_18fab52d5c9
rearranged + sql udfs | 3.4 | bquxjob_4145760d_18fab765f03

100% of firefox desktop for 2024-05-22 (both overwriting a clustered partition):

changes | slot hours | job id (backfill-2)
-- | -- | --
base (airflow run) | 3910 😵‍💫 | bqjob_r6deaccf950e8d7f5_0000018fa35c4fce_1
rearranged + sql udfs | 699 | bquxjob_5293faeb_18fab5ab707

I'm mostly confident the output is equivalent because this doesn't throw any errors:
```sql
SELECT
  mozfun.assert.json_equals(from_map_event_extra(event.extra), mozfun.json.from_map(event.extra)),
  mozfun.assert.json_equals(from_map_experiment(ping_info.experiments), mozfun.json.from_map(ping_info.experiments)),
FROM
  `moz-fx-data-shared-prod.firefox_desktop_stable.events_v1`
CROSS JOIN 
  UNNEST(events) AS event
WHERE 
  DATE(submission_timestamp) IN ('2024-05-22')
  AND sample_id = 1
```
Also ran this for fenix and ios

Checklist for reviewer:

- [ ] Commits should reference a bug or github issue, if relevant (if a bug is referenced, the pull request should include the bug number in the title).
- [ ] If the PR comes from a fork, trigger integration CI tests by running the [Push to upstream workflow](https://github.com/mozilla/bigquery-etl/actions/workflows/push-to-upstream.yml) and provide the `<username>:<branch>` of the fork as parameter. The parameter will also show up
in the logs of the `manual-trigger-required-for-fork` CI task together with more detailed instructions.
- [ ] If adding a new field to a query, ensure that the schema and dependent downstream schemas have been updated.
[...]

PR #8291, opened 2025-10-20
refactor(mozfun): Convert JavaScript UDFs in `mozfun.json` into SQL UDFs (DENG-9782)

sean-rose:
## Description
The [JSON_KEYS](https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#json_keys) function is no longer in preview as of [2025-09-16](https://cloud.google.com/bigquery/docs/release-notes#September_16_2025), and this should improve the performance of queries which use those UDFs.

* DENG-9782: Convert JavaScript UDFs in `mozfun.json` into SQL UDFs

### DIFF
--- a/sql/moz-fx-data-shared-prod/search_terms/sanitization_job_languages/view.sql
+++ b/sql/moz-fx-data-shared-prod/search_terms/sanitization_job_languages/view.sql
@@ -8,4 +8,4 @@ SELECT
 FROM
   `moz-fx-data-shared-prod.search_terms_derived.sanitization_job_metadata_v2`
 CROSS JOIN
-  UNNEST(mozfun.json.js_extract_string_map(approximate_language_proportions_json)) AS lang
+  UNNEST(mozfun.json.extract_string_map(approximate_language_proportions_json)) AS lang
--- a/sql/moz-fx-data-shared-prod/subscription_platform_derived/nonprod_stripe_subscriptions_history_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/subscription_platform_derived/nonprod_stripe_subscriptions_history_v1/query.sql
@@ -144,7 +144,7 @@ product_capabilities AS (
   FROM
     `moz-fx-data-shared-prod`.stripe_external.nonprod_product_v1 AS products
   JOIN
-    UNNEST(mozfun.json.js_extract_string_map(metadata)) AS metadata_items
+    UNNEST(mozfun.json.extract_string_map(metadata)) AS metadata_items
     ON metadata_items.key LIKE 'capabilities%'
   JOIN
     UNNEST(SPLIT(metadata_items.value, ",")) AS capability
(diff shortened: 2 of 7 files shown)


sql/mozfun/json/from_nested_map/udf.sql
```diff
@@ -1,21 +1,45 @@
...
+                      []
+                    ),
+                    COALESCE(ARRAY_AGG(key_value_pair.value ORDER BY key_value_pair_offset), [])
+                  )
+                FROM
+                  UNNEST(JSON_QUERY_ARRAY(input[key])) AS key_value_pair
+                  WITH OFFSET AS key_value_pair_offset
+                WHERE
+                  key_value_pair.key IS NOT NULL
+                  AND key_value_pair.value IS NOT NULL
```
BenWu, line 28: question(non-blocking): Could the `from_map_json` be reused here?  I don't know if this works as-is but something like:
```suggestion
                SELECT
                  json.from_map_json(input[key])
```

BenWu (approved):

sql/mozfun/json/from_nested_map/udf.sql
sean-rose: My aim for this PR was to keep the results exactly the same as the previous JavaScript implementations, and using `from_map_json` would convert number-like and boolean-like string values into numbers/booleans, which this function wasn't doing before.

sean-rose: Digging into this a bit further, it looks like the `from_map` & `from_map_json` functions were intended to be used with maps where numeric/boolean values always end up encoded as strings (e.g. in event extras), so trying to parse the string values make sense for that use case.

In contrast, this `from_nested_map` function appears to be used for labeled metrics, where numeric/boolean values are not encoded as strings (e.g. labeled counter metric values are actual integers, and labeled boolean metric values are actual booleans), so trying to parse string values is unnecessary for that use case, and could accidentally mangle labeled string metric values.

BenWu: Ok, I didn't notice that it was parsing as strings.  That's fine then

sql/mozfun/json/from_map_json/udf.sql
sean-rose, line 1: @BenWu with this change we could potentially try switching the events stream ETLs back to calling `mozfun.json.from_map` rather than the `from_map_event_extra` temp function you implemented in #5659.  However, that would incur a bit of extra processing compared to the temp function since `mozfun.json.from_map` converts the input array to JSON and passes it to this function, rather than just operating directly on the SQL array.

Or alternatively it could make sense to change the `mozfun.json.from_map` function's `input ANY TYPE` argument to be `input ARRAY<STRUCT<key STRING, value STRING>>` and replace its implementation with the same code as the events stream ETL's `from_map_event_extra` temp function, since there's really no technical reason the SQL array needs to be converted into JSON and passed to this function.

BenWu: Although it would be better to have less code, I'm not sure it's worth the time to change it if it's working.  So I would leave it as-is but you can give it a try if you want to.  Changing the events stream side would require a bit of performance testing to at least see if there's any impact

PR #8300, opened 2025-10-21
GENAI-2114 Redo prior calculation for Merino Thompson sampling of content recommendations

rolf-moz:
## Description

Priors used for Merino thompson sampling are based on schedule_corpus_item_id which is being phased out. They also didn’t take into account the newtab_content ping which was causing a big drop in US impression per day value.

Comparison of the result are in the JIRA ticket to protect data.

**Duration of job**
 Duration
    2 sec
Bytes processed
    3.42 GB
Bytes billed
    3.42 GB
Slot milliseconds
    403457

Before PR
Slot time was over 3 days (I canceled before it was finished)

https://mozilla-hub.atlassian.net/browse/GENAI-2114

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_priors_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_priors_v1/query.sql
@@ -54,7 +54,7 @@ per_region_ctr AS (
   FROM
     aggregated_events
   WHERE
-    impression_count > 2000
+    impression_count > 2000 -- 1/min_ctr protects against CTRs form low sample size
 ),
   -- Average impressions per item per region (rounded)
 per_region_impressions_per_item AS (


sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_priors_v1/query.sql
```diff
@@ -1,74 +1,62 @@
...
+    country AS region,    -- rename here
+    impression_count,
+    click_count
+  FROM
+    `moz-fx-data-shared-prod.firefox_desktop_derived.newtab_content_items_daily_combined_v1`
   WHERE
-    TIMESTAMP_MILLIS(recommended_at) >= params.start_timestamp
-    AND TIMESTAMP_MILLIS(recommended_at) < params.end_timestamp
+    corpus_item_id IS NOT NULL
+    AND country IS NOT NULL
```
mmiermans, line 25: Do you want to ignore events without a country? I think it would bring the global count down a bit, but it may be worth checking how much.

sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_priors_v1/query.sql
```diff
@@ -1,74 +1,62 @@
...
+    b.corpus_item_id,
+    b.region,
+    SUM(b.impression_count) AS impression_count,
+    SUM(b.click_count) AS click_count
   FROM
-    flattened_newtab_events
+    base b
+  INNER JOIN
+    corpus_items c
+    ON b.corpus_item_id = c.corpus_item_id
```
mmiermans, line 41: Could you measure the performance before and after, and add that data to the PR description? In the past I heard about some concerns joining data in BigQuery across these tables. My guess has always been that it's would be fine, but I may be wrong.

sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_priors_v1/query.sql
```diff
@@ -1,74 +1,62 @@
...
+    corpus_item_id,
     region,
     SAFE_DIVIDE(click_count, impression_count) AS ctr,
     impression_count,
     click_count
   FROM
     aggregated_events
   WHERE
-    impression_count > 0
+    impression_count > 2000
```
mmiermans: Could you add a comment about how this value was derived, in case someone wants to recompute it in the future?

sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_priors_v1/query.sql
rolf-moz: There are already jobs that join now on the snowflake data.

https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/firefox_desktop_derived/newtab_content_items_daily_combined_v1/view.sql

PR #8314, opened 2025-10-22
feat(SubPlat): add ended_reason to service subscriptions (DENG-4043)

phil-lee70:
## Description

This PR passes the `ended_reason` field downstream for service subscriptions.

* DENG-4043

### DIFF
--- a/sql/moz-fx-data-shared-prod/subscription_platform_derived/service_subscription_events_v1_live/view.sql
+++ b/sql/moz-fx-data-shared-prod/subscription_platform_derived/service_subscription_events_v1_live/view.sql
@@ -138,7 +138,7 @@ auto_renew_change_events AS (
   SELECT
     `timestamp`,
     'Auto-Renew Change' AS type,
-    subscription.ended_reason AS reason,
+    IF(subscription.auto_renew, 'Auto-Renew Enabled', 'Auto-Renew Disabled') AS reason,
     service_subscriptions_history_id,
     subscription,
     old_subscription


sql/moz-fx-data-shared-prod/subscription_platform/daily_active_service_subscriptions/schema.yaml
```diff
@@ -431,6 +431,17 @@ fields:
     description: |-
       Projected annual recurring revenue for the subscription in US dollars.
       This will be null for Stripe subscriptions prior to 2023-02-27 (DENG-754).
+  - name: ended_reason
```
sean-rose: **issue (blocking):** In the views `ended_reason` won't appear after the additional revenue-related fields the views themselves are adding.  This should be moved up before `country_vat_rate`.

sql/moz-fx-data-shared-prod/subscription_platform/daily_active_service_subscriptions/schema.yaml
```diff
@@ -431,6 +431,17 @@ fields:
...
+    type: STRING
+    mode: NULLABLE
+    description: |-
+      Reason why the subscription ended.
+      Possible values:
+        * `Admin Initiated` - An admin at Mozilla ended the subscription, either directly or by deleting the customer's Mozilla Account.
+        * `Customer Initiated` - The customer ended their subscription, either directly or by turning off auto-renewal or deleting their Mozilla Account.
+        * `Payment Failure` - The subscription was unable to auto-renew due to a payment failure.
+        * `Other` - The subscription ended for some other reason, such as unverified Mozilla Accounts being automatically deleted.
+        * `Downgrade` - The subscription ended due to the customer moving to a lower tier subscription.
```
sean-rose: **suggestion (non-blocking):** IMO it makes most sense to list "Other" last, so I'd suggest listing "Downgrade" before "Other". (and same in all the other `schema.yaml` files)

sql/moz-fx-data-shared-prod/subscription_platform/service_subscriptions/schema.yaml
```diff
@@ -409,3 +409,14 @@ fields:
   mode: NULLABLE
   description: |-
     Projected annual recurring revenue for the subscription in US dollars.
+- name: ended_reason
```
sean-rose: **issue (blocking):** In the views `ended_reason` won't appear after the additional revenue-related fields the views themselves are adding.  This should be moved up before `country_vat_rate`.

sean-rose (commented):
**suggestion (blocking):** You also need to use `ended_reason` as the `reason` for service subscription "Subscription End" events.

sql/moz-fx-data-shared-prod/subscription_platform_derived/service_subscription_events_v1_live/view.sql
```diff
@@ -139,7 +138,7 @@ auto_renew_change_events AS (
   SELECT
     `timestamp`,
     'Auto-Renew Change' AS type,
-    IF(subscription.auto_renew, 'Auto-Renew Enabled', 'Auto-Renew Disabled') AS reason,
+    subscription.ended_reason AS reason,
```
sean-rose: **issue (blocking):** This will result in the `reason` for `Auto-Renew Change` events always being null, and the `ended_reason` values shouldn't be used as the `reason` for `Auto-Renew Change` events in any case.

phil-lee70: >  You also need to use ended_reason as the reason for service subscription "Subscription End" events.

So what field to add ended_reason then?

sean-rose: Toward the top of this file you already changed the logic for `Subscription End` events to use `subscription.ended_reason` as the `reason` for those events, which was the appropriate thing to do.  But this logic for `Auto-Renew Change` events should be left as is.

PR #8317, opened 2025-10-22
DENG 9478 create city seen for firefox desktop background apps

wwyc:
## Description

DS has asked to create city seen derived tables for firefox background apps

* DENG-9478


bqetl_project.yaml
```diff
@@ -530,6 +530,9 @@ generate:
     - focus_android
     - klar_ios
     - klar_android
+    - firefox_desktop_background_update
```
BenWu, line 533: This is fine to do but baseline pings were disabled in https://bugzilla.mozilla.org/show_bug.cgi?id=1971373 which I think went out in Firefox 142 so this isn't really reliable since it's incomplete.  Anyone asking for this should be aware of that

PR #8345, opened 2025-10-27
fix(glean_usage): Bug fixes and refactors for the `glean_usage` SQL generator

sean-rose:
## Description
I ran into a number of issues when trying to work on the `glean_usage` SQL generator code for events stream ETLs/views (DENG-9546), and decided to fix those in this separate PR.

### Bug fixes:
- Correctly filter out tables in all datasets associated with skipped apps.
- Pass all `GleanTable` method arguments consistently.
- Correct documentation of `prefix` parameter.
- Fix reference to nonexistent `generate.glean_usage.events_stream.skip_datasets` config.
- Don't update `GleanTable` object attributes in methods that are called in parallel.

### Refactors:
- Indicate the return type of `get_app_info()`.
- Pass `app_name` to `GleanTable`'s `generate_per_app_id()` and `generate_per_app()` methods.
- Use `app_name` values for "app" configs.
- Remove superfluous `generate.glean_usage.bigconfig.skip_app_metrics` config.
- Rename `custom_render_kwargs` object attributes to `common_render_kwargs` to reinforce the fact that they're shared between method invocations.
- Apply code changes required by `black`, `isort`, and `pydocstyle`.

N/A

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -41,7 +41,7 @@ pytest-pydocstyle==2.4.0
 pytest-xdist==3.8.0
 pytest==8.4.2
 PyYAML==6.0.3
-rich-click==1.9.3
+rich-click==1.9.4
 smart_open==6.4.0
 sqlglot==25.28.0
 sqlparse==0.5.3
--- a/requirements.txt
+++ b/requirements.txt
@@ -2174,9 +2174,9 @@ rich==12.6.0 \
     # via
     #   rich-click
     #   typer
-rich-click==1.9.3 \
-    --hash=sha256:60839150a935604df1378b159da340d3fff91f912903e935da7cb615b5738c1b \
-    --hash=sha256:8ef51bc340db4d048a846c15c035d27b88acf720cbbb9b6fecf6c8b1a297b909
+rich-click==1.9.4 \
+    --hash=sha256:af73dc68e85f3bebb80ce302a642b9fe3b65f3df0ceb42eb9a27c467c1b678c8 \
+    --hash=sha256:d70f39938bcecaf5543e8750828cbea94ef51853f7d0e174cda1e10543767389
     # via -r requirements.in
 rpds-py==0.8.10 \
     --hash=sha256:08166467258fd0240a1256fce272f689f2360227ee41c72aeea103e9e4f63d2b \
--- a/sql/moz-fx-data-shared-prod/external_derived/market_intel_bot_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/external_derived/market_intel_bot_v1/metadata.yaml
@@ -19,5 +19,5 @@ scheduling:
     key: bqetl_chrome_releases__open_ai_api_key
   - deploy_target: MARKET_INTEL_BOT_GITHUB_ACCESS_TOKEN
     key: bqetl_market_intel_bot__github_token
-  # - deploy_target: SLACK_MARKET_INTEL_BOT_WEBHOOK_URL
-  #   key: bqetl_market_intel_bot__webhook_url
+  - deploy_target: SLACK_MARKET_INTEL_BOT_WEBHOOK_URL
+    key: bqetl_market_intel_bot__webhook_url
--- a/sql/moz-fx-data-shared-prod/external_derived/market_intel_bot_v1/query.py
+++ b/sql/moz-fx-data-shared-prod/external_derived/market_intel_bot_v1/query.py
@@ -43,16 +43,16 @@ OUTPUT_FPATH_8 = OUTPUT_BASE + "UpcomingHolidays/WebScraping_"
 
 # Pull in the different keys from GSM
 OPENAI_API_TOKEN = os.getenv("DATA_ENG_OPEN_AI_API_KEY")
-# SLACK_WEBHOOK = os.getenv("SLACK_MARKET_INTEL_BOT_WEBHOOK_URL")
+SLACK_WEBHOOK = os.getenv("SLACK_MARKET_INTEL_BOT_WEBHOOK_URL")
 GITHUB_ACCESS_TOKEN = os.getenv("MARKET_INTEL_BOT_GITHUB_ACCESS_TOKEN")
 
 # If any aren't found, raise an error
 if not OPENAI_API_TOKEN:
     raise ValueError("Environment variable DATA_ENG_OPEN_AI_API_KEY is not set!")
-# if not SLACK_WEBHOOK:
-#     raise ValueError(
-#         "Environment variable SLACK_MARKET_INTEL_BOT_WEBHOOK_URL is not set!"
-#     )
+if not SLACK_WEBHOOK:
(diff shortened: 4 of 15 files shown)


sql_generators/glean_usage/__init__.py
```diff
@@ -87,7 +87,7 @@
 def generate(
     target_project, output_dir, parallelism, exclude, only, app_name, use_cloud_function
 ):
-    """Generate per-app_id queries and views, and per-app dataset metadata and union views.
```
scholtzan, line 90: When this was set up initially calling it App ID instead of channel was somewhat intentional as channel usually refers to nightly, beta, release but the generator is usually generating artifacts for specific app ID names. 
Each app ID corresponds to a channel, so I'm ok with keeping the rename change here (though some apps have channels that are based on multiple app IDs, like fenix).

bqetl_project.yaml
```diff
@@ -377,33 +370,33 @@ generate:
...
       - ads_backend  # restricted dataset, we don't want to include it aggregate view
-      - org_mozilla_fenix_nightly
-      - org_mozilla_fennec_aurora
       # event pings for this app were removed but still high volume due to nimbus bugs in versions 138 to 140
       # events are still sent in the background-update ping, but they are being skipped
       # See: https://mozilla-hub.atlassian.net/browse/DENG-9715
       - firefox_desktop_background_update
+      skip_app_channels:
+      - org.mozilla.fenix.nightly
+      - org.mozilla.fennec.aurora
```
BenWu: These are app ids so I don't think calling this `skip_app_channels` is right
```suggestion
      skip_app_ids:
      - org.mozilla.fenix.nightly
      - org.mozilla.fennec.aurora
```

sql_generators/glean_usage/__init__.py
sean-rose: Yeah, the term "app" meaning multiple things in Glean is confusing and unfortunate.  Ideally I'd prefer things to be disambiguated in the Glean metadata itself (e.g. changing `app_name` to something like `app_group` or `app_family`), but I'm guessing that's probably more work than it's worth at this point.

I chose the term "app channel" based on the naming of the `channels:` arrays in [probe-scraper/repositories.yaml](https://github.com/mozilla/probe-scraper/blob/main/repositories.yaml), but as you point out that's now giving "app channel" multiple meanings.

To try to avoid causing further confusion, how about I take a cue from [this `glean_app_ping_views.py` code](https://github.com/mozilla/bigquery-etl/blob/51bc458d371041a522dba7484611393169e50185/sql_generators/glean_usage/glean_app_ping_views.py#L114) and use the term "channel app" instead?
- The `GleanTable.per_app_channel_enabled` attribute would be renamed `per_channel_app_enabled`.
- The `GleanTable.generate_per_app_channel()` method would be renamed `generate_per_channel_app()`, and its `app_channel_info` arg would be renamed `channel_app_info`.
- The `GleanTable.generate_per_app()` method's `app_channels_info` arg would be renamed `channel_apps_info`.

bqetl_project.yaml
sean-rose: Fair enough.

Or if [my latest suggestion in the other thread](https://github.com/mozilla/bigquery-etl/pull/8345#discussion_r2467088808) to use the term "channel app" instead is agreeable, then perhaps `skip_channel_apps` would make sense here?

sql_generators/glean_usage/__init__.py
BenWu: "channel app" doesn't make much sense to me since it only makes sense in the context of that code as the app info belonging to the current channel in the loop.  To me, per app id is the most appropriate because that's exactly what the code does.  per app channel adds a layer of abstraction but still makes sense because that's basically what it does, apart from fenix which has multiple ids per channel.  

I don't think the change is necessary because everybody working on this should know/learn what an app id is.  app id is what glean has always used so I would prefer not to hide that detail here.  I agree that a glean "app" can be confusing but as long as app name and app id are clearly disambiguated here, I think that issue is fixed

sean-rose: OK, I switched things back to using the term `app_id`, rather than `app_channel` or `channel_app`.

sean-rose:
> * Correctly filter out tables in all datasets associated with skipped apps.
> * Fix reference to nonexistent `generate.glean_usage.events_stream.skip_datasets` config.

FYI, these bug fixes are resulting in the generated code for the following views/ETLs being removed:
```
moz-fx-data-shared-prod.bergamot.events_stream
moz-fx-data-shared-prod.firefox_echo_show.events_stream
moz-fx-data-shared-prod.firefox_fire_tv.events_stream
moz-fx-data-shared-prod.firefox_reality.events_stream
moz-fx-data-shared-prod.firefox_reality_pc.events_stream
moz-fx-data-shared-prod.lockwise_android.events_stream
moz-fx-data-shared-prod.lockwise_ios.events_stream
moz-fx-data-shared-prod.mozilla_lockbox.events_stream
moz-fx-data-shared-prod.mozilla_lockbox_derived.events_stream_v1
moz-fx-data-shared-prod.org_mozilla_bergamot.events_stream
moz-fx-data-shared-prod.org_mozilla_bergamot_derived.events_stream_v1
moz-fx-data-shared-prod.org_mozilla_connect_firefox.events_stream
moz-fx-data-shared-prod.org_mozilla_connect_firefox_derived.events_stream_v1
moz-fx-data-shared-prod.org_mozilla_firefoxreality.events_stream
moz-fx-data-shared-prod.org_mozilla_firefoxreality_derived.events_stream_v1
moz-fx-data-shared-prod.org_mozilla_ios_lockbox.events_stream
moz-fx-data-shared-prod.org_mozilla_ios_lockbox_derived.events_stream_v1
moz-fx-data-shared-prod.org_mozilla_social_nightly.baseline_clients_daily
moz-fx-data-shared-prod.org_mozilla_social_nightly.baseline_clients_first_seen
moz-fx-data-shared-prod.org_mozilla_social_nightly.baseline_clients_last_seen
moz-fx-data-shared-prod.org_mozilla_social_nightly.events_stream
moz-fx-data-shared-prod.org_mozilla_social_nightly_derived.baseline_clients_daily_v1
moz-fx-data-shared-prod.org_mozilla_social_nightly_derived.baseline_clients_first_seen_v1
moz-fx-data-shared-prod.org_mozilla_social_nightly_derived.baseline_clients_last_seen_v1
moz-fx-data-shared-prod.org_mozilla_social_nightly_derived.events_stream_v1
moz-fx-data-shared-prod.org_mozilla_tv_firefox.events_stream
moz-fx-data-shared-prod.org_mozilla_tv_firefox_derived.events_stream_v1
moz-fx-data-shared-prod.org_mozilla_vrbrowser.events_stream
moz-fx-data-shared-prod.org_mozilla_vrbrowser_derived.events_stream_v1
```

BenWu (approved):

PR #8358, opened 2025-10-28
feat(DENG-9505): migrating amo datasets to glean

kik-kik:
# feat(DENG-9505): migrating amo datasets to glean

This PR introduces the following changes:

- Add `firefox_desktop_stats_installs_v1` inside `addons_derived` that is based on Glean telemetry.
- Add `firefox_desktop_addons_by_clients_v1` inside `addons_derived` that is based on Glean telemetry.
- Add `fenix_addons_by_clients_v1` inside `addons_derived` that is based on Glean telemetry.
- Add `stats_dau_v1` inside `addons_derived` that is based on addons Glean telemetry ping (and only includes clients with app_version 148 and later).
- Add `amo_stats_dau_legacy_source_v1` inside `amo_prod` that limits which clients to include in the aggregation (pre 148 major version) using app_version_major to make the union of the existing and new aggregations as simple as possible.

Things still to do as follow-up:

- Validate the results are the same or as close as we'd expect them to be (for Desktop since data source changes).
- If everything is in order migrate over the amo_prod views to use the new datasets (this will likely be some sort of union between the existing and new logic)

### DIFF
--- a/sql/moz-fx-data-shared-prod/addons_derived/stats_dau_v1/schema.yaml
+++ b/sql/moz-fx-data-shared-prod/addons_derived/stats_dau_v1/schema.yaml
@@ -6,6 +6,7 @@ fields:
 - name: addon_id
   type: STRING
   mode: NULLABLE
+  description: Unique identifier of an addon.
 - name: dau
   type: INTEGER
   mode: NULLABLE
@@ -13,60 +14,78 @@ fields:
 - name: dau_by_addon_version
   type: RECORD
   mode: REPEATED
+  description: Record containing addon version and the corresponding DAU count.
   fields:
   - name: key
     type: STRING
     mode: NULLABLE
+    description: Addon version.
   - name: value
     type: INTEGER
     mode: NULLABLE
+    description: DAU count for the corresponding addon version.
 - name: dau_by_app_version
   type: RECORD
   mode: REPEATED
+  description: Record containing app version and the corresponding DAU count.
   fields:
   - name: key
     type: STRING
     mode: NULLABLE
+    description: App version.
   - name: value
     type: INTEGER
     mode: NULLABLE
+    description: DAU count for the corresponding app version.
 - name: dau_by_fenix_build
   type: RECORD
   mode: REPEATED
+  description: Record containing fenix build and the corresponding DAU count.
   fields:
   - name: key
     type: STRING
     mode: NULLABLE
+    description: Build identifier.
   - name: value
     type: INTEGER
     mode: NULLABLE
+    description: DAU count for the corresponding fenix build.
 - name: dau_by_locale
   type: RECORD
   mode: REPEATED
+  description: Record containing locale and the corresponding DAU count.
   fields:
   - name: key
     type: STRING
     mode: NULLABLE
+    description: locale.
   - name: value
     type: INTEGER
     mode: NULLABLE
+    description: DAU count for the corresponding locale.
 - name: dau_by_country
   type: RECORD
   mode: REPEATED
+  description: Record containing country and the corresponding DAU count.
   fields:
   - name: key
     type: STRING
     mode: NULLABLE
+    description: Country.
   - name: value
     type: INTEGER
     mode: NULLABLE
+    description: DAU count for the corresponding country.
 - name: dau_by_app_os
   type: RECORD
   mode: REPEATED
+  description: Record containing os and the corresponding DAU count.
   fields:
   - name: key
     type: STRING
     mode: NULLABLE
+    description: OS.
   - name: value
     type: INTEGER
     mode: NULLABLE
+    description: DAU count for the corresponding os.


rpl:
Hi @kik-kik, it took me a bit longer to comment here than I originally promised, sorry about that. 

I spent some time yesterday looking more closely into the conventions used in some other of the directories in this repo, and putting together some mermaid graphs to better visualize the current state and feasibility/pros/cons of the organization and naming that William and myself have been briefly discussing yesterday.

Hopefully the data flow diagram may be helpful enough to highlight the ideas/proposals that I'm going to describe below 🤞

## Proposed tweaks to linearize/simplify the new derived datasets and view

The current data flow for the `amo_glean` and `amo_glean_derived` directories/namespaces, as currently defined in this pull request, seems to look more or less like in the following diagram:

amo_glean-dataflow-diagram

Whereas the current data flow for the amo_prod bigquery-etl directory looks a bit more linear (meaning that data sources flows into derived datasets and derived datasets into the views, without the views becoming the source of other derived datasets):

amo_prod-dataflow-diagram

I'm wondering if we could simplify the data flows from this pull request a bit and make it a more linear data flow and to mirror the `amo_prod` structure a bit more closely, e.g. something like the following (I'm keeping all the view and table names the same to make it easier to match this diagram with the derived tables and views as currently defined in this PR):

amo_glean-proposed-dataflow-diagram

### fenix.addons vs. org_mozilla_firefox/…/org_mozilla_fenix.addons

NOTE: as a side note, in the diagram right above `fenix_addons_by_client_v1` is getting the mobile builds data from a single `fenix.addons` data source instead of the per-channel ones (`org_mozilla_firefox.addons`, `org_mozilla_fenix.addons` etc.). @kik-kik this is what I was also asking couple of days ago over slack when I noticed that `fenix.addons` in Redash seems to have data from all mobile channels (similarly to how `firefox_desktop.addons` provides access to the Glean `addons` ping data from all desktop channels). would that be possible?

## Naming conventions and organization of current and proposed data flow diagrams

As I mentioned during our meeting yesterday, William and myself have been wondering if we could organize the current and new stuff with a naming convention that would feel a bit more unified instead of splitting the current and new derived datasets and views in completely separate namespaces and with different naming conventions (which may also not make it immediately clear how they are actually related to each other and combined into an unified view including data for both the old and new clients).

@willdurand would the following data flow diagram be matching what you had in mind? (or at least going in that direction)

amo_prod-proposed-merged-dataflow-diagram-final-state

If I'm not mistaken, that would still follow the naming conventions used in this repository (in particular the derived datasets to be located in an amo_prod_derived directory / namespace as described in https://github.com/mozilla/bigquery-etl/blob/main/docs/cookbooks/creating_a_derived_dataset.md ) while also matching a bit more closely what William was describing me yesterday.

### Backward compatible step

The data flow and organization shown in the previous diagram would be technically backward incompatible with some queries (and related dashboard) that currently exist in Redash that are currently (and wrongly) using the derived dataset table directly instead of the corresponding view, and so it would break those redash queries as soon as we would be moving the current derived dataset into the `amo_prod_derived` namespace.

(e.g. the redash query `85611`, part of the `live-language-switching` redash dashboard as an example, not a great one given that the dashboard seems to have been last refreshed in 2024, but there are very likely more).

We can track these down this mistakes easily enough in redash, but fixing them may require some back and forth with the owners of the queries and dashboads involved.

To avoid having to deal with that as soon as we land these new derived datasets and views, I'm wondering if it may be a good idea to go first through an initial “backward compatible” state, for which we don't move the current derived dataset just yet (and instead we defer that to a separate followup step), something like the following:

amo_prod-proposed-merged-dataflow-diagram-step1-backward-compatible-state

@kik-kik @willdurand wdyt? does this sounds like a reasonable intermediate state (and path to the final state) worth considering?

## followup: extend amo_stats_install to also account for fenix install events

This is not something to deal with in this pull request, but putting together these data flow diagrams made me notice that at the moment `amo_stats_installs` is only including Firefox Desktop install events, due to historical reasons (because `amo_stats_installs_v3` used to be taking the data from the legacy telemetry.events data source which was only available for desktop builds).

The new one is now using the Glean events (at the moment it mirrors the current datasets and only takes data from `firefox_desktop.events`) and those are available on mobile too, and so as a separate followup we should also consider to extend that to also include the Fenix Glean events. @willdurand wdyt? is what I'm describing here a reasonable follow from an AMO perspective too, right?

kik-kik:
Hi @rpl 

Let me try responding to your comment section by section.

> Proposed tweaks to linearize/simplify the new derived datasets and view

Maybe I'm misunderstanding this point. The lineage here is already pretty straight forward. It sounds like here we're considering the views as the "bloat"? In general, having views for a query is a pretty standard convention. The view often acts as a "gateway" to the underlying physical table. This is helpful for example if we ever need to create a new version of the query in which case we could just update the view and let the downstream processes not be concerned with this.

Right now it's more less like this:

```
(fenix) channel_1           -> amo_glean.addons -> fenix_by_clients                   -> amo_dau
(fenix) channel_2          -> amo_glean.addons -> fenix_by_clients                   -> amo_dau
desktop (all_channels) -> amo_glean.addons -> firefox_desktop_by_clients  -> amo_dau
```

Above another option would be to create fenix.addons which would be the union of all fenix channels and then the clients queries could consume from fenix.addons (for fenix) and firefox_desktop.addons (for desktop).

```
firefox_desktop.events -> amo_install_stats
```

> fenix.addons vs. org_mozilla_firefox/…/org_mozilla_fenix.addons

Yes, firefox_desktop.addons (only firefox desktop) and amo_glean.addons (firefox_desktop + fenix) contains all channels. On fenix we see different namespaces for fenix (different for each channel) because each channel is an individual app (I'm not fully aware of why all don't just use a single end-point for sending telemetry regardless of the channel but instead each is treated separately vs on desktop where all channels land in a single table). This is why we need to union all channels together on Fenix.

I guess one question here I'd have is if the amo_glean.addons should contain both firefox desktop and fenix data (as it currently is set up) or if we want a stronger separation between the two?

> Naming conventions and organization of current and proposed data flow diagrams

I would still be in suggesting we separate glean and legacy data on namespace level instead of just treating it as v2. If we want more closely aligned naming conventions between the two then I suggest we update the existing namespace to be something like `amo_legacy_derived` and `amo_legacy` for views. We could then use `amo_prod` as the namespace that combines the two. I'd avoid having a namespace that has resources that are partially based of one vs the other source as I think this will make the deprecation of legacy more tricky later on.

And yes, another option would be to keep `amo_prod` as is and consider it the `legacy` namespace and create a new one like `amo` for combined glean + legacy data once we reach that stage.

> followup
Cool, thanks for mentioning this and yes let's deal with this as a follow-up.

---

@rpl please let me know if that addresses your thoughts or if I completely missed the mark there.

willdurand:
@kik-kik I am sorry to insist but I second Luca's points, and I would also like to see the following implemented in this PR as well (this image is the last image in Luca's comment):

image

`amo_prod` and `amo_dev` are the two namespaces AMO and the Add-ons team are using, and I'd like to keep it that way. Although right now it's just stats stuff, it might be different in the future but there is no need to make things more complicated and possibly confusing.

I am not really in favor of any `legacy` prefix but if that makes it easier to identify things to remove in the future, we can potentially move things under `amo_prod/legacy/` and `amo_prod_derived/legacy/`. About BC break: we can identify what is querying the derived tables, but AMO reads the views, and others should as well.

We should also have new derived tables with higher versions as well, not completely new tables with a version that's lower or equal to what we already have because that'd be very confusing.

In a follow-up, we can then move the existing derived tables into `amo_prod_derived`. That would leave us with `amo_prod`, `amo_prod_derived` and `amo_dev`. We can deal with `amo_dev` separately as well.

kik-kik:
Hey @willdurand and @rpl ,

unfortunately your proposal to have `amo_prod/legacy/` is simply not possible as our tool simply does not support this. Additionally, I'm not sure you noticed, but the current folder structure in this project directly maps to the resources in BigQuery which the suggestion would not fit.

I will also insist on creating a new namespace to separate glean from legacy data for better clarity on the source of the data to fit our existing design and structure inside BigQuery. Using different version (v2 for glean based query) is also not something I believe to be an appropriate approach as version bump implies the newer version of the query supersedes the previous versions and this case this is not quite accurate since both remain current and will continue to co-exist for some time.

I'm happy for `amo_prod` to serve as the namespace that combines the two and acts as a flexible abstraction layer to reduce downstream impact on any changes to the underlying tables or structure, but I still believe the two telemetry source should be clearly separated to avoid confusion, reduce risks of errors, make the validation work easier, and the later deprecation easier.

edit: One other approach here could be to put the product specific datasets (like addons by client) in fenix_derived and firefox_desktop_derived namespaces respectively and have amo namespace act as a union of all amo data.

rpl:
I'd like to find a middle ground that would allow us to move forward in the short term and land into a final state that we can agree on.

Follows a new proposal (where step1 represents the proposal for this pull request, while step2 and step3 describes the proposed next step and final stage I'm proposing as part of this middle ground proposal).

I'd tried to include in this proposal the points made so far from both of you, and so hopefully it may be a reasonable middle ground (or at least a base to get to a middle ground we can agree on and move forward).

@kik-kik @willdurand Let me know what do you think about this proposal and any tweak to it that may help us to get to an agreement.

### Step 1 - new namespace, derived datasets and views (this PR)

How about we start by using naming conventions similar to the ones in the following "step 1" data flow diagram?

step 1 - New Proposed AMO Stats Dataflow Diagram 2026-01-19

At this stage we are only aiming to introduce the new namespaces, derived datasets and views with the following naming conventions:

`addons_stats_derived` (instead of `amo_glean_derived`)
- this namespace is meant to group the new datasets, its name highlights it is meant to host derived datasets related to addons-related stats based on the Glean telemetry data sent by desktop and android mobile clients (while the `amo_prod` at the final stage step3 will only be used to group the views meant to be directly by AMO)
- The derived datasets follows the following convention (where `CLIENTTYPE` is going to be `desktop` or `fenix`, `N` the derived datasets version numbers, currently just `1`):
  - `active_addons_by_CLIENTTYPE_client_vN`
  - `installed_addons_by_CLIENTTYPE_client_vN`

`addons_stats` (instead of `amo_glean`):
- this namespace is meant to group the new intermediate views (which only include the new derived datasets), its name highlights that it is the namespace for the view related to `addons_stats_derived` only.

`amo_prod`:
- at this stage this namespace is still left as is (will be started to be tweaked starting from the following step).

### Step 2 - Merge old clients and new clients data in the final amo_stats view amo_stats_dau and amo_stats_installs

At this stage we aim to start using the new views for the AMO dau and install stats, while not yet moving them out of `amo_prod` and in new namespaces the already existing derived datasets, and so this data flow graph is the same as the previous one with the addition of two new intermediate views old including the data for the old clients for which we still need to use the current derived datasets:

- `dau_oldclients`  and `installs_oldclients` two new views (that only includes data from the current `amo_prod` derived datasets)
- `amo_prod.amo_stats_dau` to include data from both `amo_prod.dau_oldclients` and `addons_stats.dau` (with the data got from the new derived datasets)
- similarly `amo_prod.amo_stats_installs` to include data from both `amo_prod.installs_oldclients` and `addons_stats.installs`.

step 2 - New Proposed AMO Stats Dataflow Diagram 2026-01-19

### Step 3 - Move old clients derived dataset and intermediate views into new namespace and gets amo_prod into its final shape

At this stage we aim to get `amo_prod` into its final state (and only be grouping the final views that AMO is meant to be consuming), and so this step will introduce a breaking change for Redash queries that were wrongly using the versioned derived datasets tables.

With this final step:

- `amo_prod` is meant to only group the two final views that AMO is consuming directly:  `amo_stats_dau` and `amo_stats_installs` (which are derived by the intermediate views `dau` and `installs` from the two namespaces `addons_stats` and `addons_stats_oldclients`)
- `amo_prod` derived datasets to be moved to the new namespace `addons_stats_oldclients_derived` (which as the namespace suggests groups the derived datasets gathering data from the old clients, like `addons_stats_derived` gathers data from the new clients)
- intermediate views introduced in `amo_prod` in step2 to be moved into a new `addons_stats_oldclients` namespace:
  - `amo_prod.dau_oldclients` renamed into `addons_stats_oldclients.dau`
  - `amo_prod.installs_oldclients` renamed into `addons_stats_oldclients.installs`

step 3 - New Proposed AMO Stats Dataflow Diagram 2026-01-19

willdurand:
I am fine with Luca's last proposal, especially the final state:

image

sql/moz-fx-data-shared-prod/addons_derived/fenix_addons_by_client_v1/schema.yaml
```diff
@@ -0,0 +1,62 @@
+# TODO: confirm if we want normalized_channel to be included here.
```
kik-kik: @rpl ah this is an open question, do we want to have a split here by channel or keep it out from this result as it has been so far?

sql/moz-fx-data-shared-prod/addons_derived/firefox_desktop_addons_by_client_v1/metadata.yaml
```diff
@@ -0,0 +1,23 @@
+---
+friendly_name: Desktop addons by client
+description: >-
+  Clients_daily-like table that records only the dimensions and addon info
+  necessary to power daily the amo_glean_derived.stats_dau_v1 query.
```
rpl: Nit: this description is referring to the previous namespace name and we should be updating it to use the one actually used in the updated pull request (`addons_derived`). 

The metadata of the fenix counterpart of this derived seems to be using already the right new namespace name in the derived data set description.

sql/moz-fx-data-shared-prod/addons_derived/firefox_desktop_stats_installs_v1/metadata.yaml
```diff
@@ -0,0 +1,30 @@
...
+description: >
+  Daily install statistics to power addons.mozilla.org stats pages.
+  See bug 1654330. Note that this table uses a hashed_addon_id
+  defined as `TO_HEX(SHA256(addon_id))` because the underlying event
+  pings have limitations on length of properties attached to events
+  and addon_id values are sometimes too long. The AMO stats application
+  looks up records in this table based on the hashed_addon_id.
+
+  The hashed_addon_id is only valid for install_stats:
+  https://searchfox.org/mozilla-central/rev/5b061cdc4d40d44988dc61aa941cfbd98e31791f/toolkit/components/telemetry/Events.yaml#233-234
```
rpl: Nit: this is currently pointing to the legacy telemetry events metadata, let's update it to point to the corresponding Glean event metadata: https://searchfox.org/firefox-main/rev/ba7a7a320649794a9948b32156f5b438fef5ce7a/toolkit/mozapps/extensions/metrics.yaml#347-349

sql/moz-fx-data-shared-prod/addons_derived/stats_dau_v1/metadata.yaml
```diff
@@ -0,0 +1,29 @@
...
+friendly_name: AMO Stats DAU
+description: |-
+  Daily user statistics to power addons.mozilla.org stats pages.
+  See: bug-1572873.
+
+  Each row in this table represents a particular addon on a particular day
+  and provides all the information needed to populate the various
+  "Daily Users" plots for the AMO stats dashboard.
+
+  NOTE: Includes only clients that reported major app_version 148 or higher.
```
rpl, line 11: praise: @kik-kik Thanks for adding this note explicitly in these derived dataset descriptions, very much appreciated.

sql/moz-fx-data-shared-prod/addons_derived/fenix_addons_by_client_v1/schema.yaml
rpl: @kik-kik Thanks for highlighting this TODO for me.

From what I recall that would not be required based on how AMO leverages the final view derived from these derived datasets at the moment, and so I feel like creating views split by channel (or day aggregated by channel) may not be necessary/worth it at the moment.

Nonetheless, including the normalized_channel here may be valuable for investigations purpose (eg. if we'll need to investigate something odd we are seeing in the data, having the channel at this level may make it slightly more precise and easier to dig into the data specific to a particular channel).

And so I'd not be against including the normalized_channel at this level of the data pipeline (and same on the firefox_desktop counterpart), without aggregating per-channel at the stats_dau_v1 derived dataset level, but from my perspective it may not be actually necessary or mandatory at the moment.

---

@willdurand do you agree about what I'm describing above? TLDR: dau stats aggregated by normalized_channel not needed or used from an AMO perspective, including normalized_channel at this derived dataset level (without use it for aggregating by channel at stats_dau_v1 level) may be useful for investigation but not strictly necessary or mandatory at the moment.

sql/moz-fx-data-shared-prod/addons_derived/firefox_desktop_stats_installs_v1/metadata.yaml
kik-kik: the link should now be updated.

sql/moz-fx-data-shared-prod/addons_derived/firefox_desktop_stats_installs_v1/query.sql
```diff
@@ -0,0 +1,152 @@
...
+  )
+);
+
+WITH install_stats AS (
+  SELECT
+    DATE(submission_timestamp) AS submission_date,
+    client_info.client_id,
+    extract_addon_info(event.extra).*,
+  FROM
+    `moz-fx-data-shared-prod.firefox_desktop.events`,
```
scholtzan: might be more performant to use `events_stream` here instead since it's already unnested

sql/moz-fx-data-shared-prod/addons/firefox_desktop_stats_installs/view.sql
```diff
@@ -0,0 +1,37 @@
...
+    SELECT AS STRUCT
+      IFNULL(`key`, 'Unknown') AS `key`,
+      `value`
+    FROM
+      UNNEST(downloads_per_content)
+  ) AS downloads_per_content,
+  ARRAY(
+    SELECT AS STRUCT
+      IFNULL(`key`, 'Unknown') AS `key`,
+      value
```
rpl: question: @kik-kik is there any reason why `value` is being wrapped by quotes in the first two but not on this one and the other one that follows below at line 32 of this view.sql? in the other one view.sql that follows (sql/moz-fx-data-shared-prod/addons/stats_dau/view.sql) it seems all of them are.

If there is no reason, it may be good to tweak this one for consistency (or to add a inline comment if there is an actual reason).

sql/moz-fx-data-shared-prod/addons_derived/firefox_desktop_stats_installs_v1/query.sql
rpl: I'll add my +1 on @scholtzan comment, I double-checked my most recent queries on redash and I've been using `events_stream` for that same reason once I discovered it did exist (I think since I saw that data source in a query that you or chutten did show me some time ago).

I feel it may be a good followup but one that doesn't need to block this PR, I assume that the data in `events_stream` is the same and just unnested and so it may be worth to look into that as a separate followup, but before we will be using these derived dataset on the AMO side. 

@kik-kik how that sounds to you?

sql/moz-fx-data-shared-prod/addons/firefox_desktop_stats_installs/view.sql
kik-kik: no, I just missed it. Wrapping the field name in ` ` ` is just to let the interpreter know it's not a reserved word but rather a field with a name that happens to be the same as one of the reserved words. I will fix this now

sql/moz-fx-data-shared-prod/addons_derived/firefox_desktop_stats_installs_v1/query.sql
kik-kik: ah thanks, your suggestion makes sense as far as I can see so updated the query to use events_stream instead.

PR #8361, opened 2025-10-28
feat(events_stream): Expose event extras as individual fields in `events_stream` views (DENG-9546)

sean-rose:
## Description
So the possible event extras are more easily visible/accessible (e.g. in BigQuery console, Looker, and Redash).

* DENG-9546: Expose event extras as individual fields in events stream BigQuery views

### DIFF
--- a/sql_generators/glean_usage/events_stream.py
+++ b/sql_generators/glean_usage/events_stream.py
@@ -211,9 +211,9 @@ def get_glean_app_event_extras_by_type(
             dependency_v1_name, ping, cutoff_date
         )
         for (
-            depencency_extra_type,
-            depencency_extras,
+            dependency_extra_type,
+            dependency_extras,
         ) in dependency_extras_by_type.items():
-            extras_by_type[depencency_extra_type].update(depencency_extras)
+            extras_by_type[dependency_extra_type].update(dependency_extras)
 
     return extras_by_type


sql_generators/glean_usage/templates/cross_channel_events_stream.query.sql
```diff
@@ -1,18 +1,27 @@
...
   `{{ project_id }}.{{ target_view }}`
 AS
-{% for (dataset, channel) in datasets %}
-{% if not loop.first -%}
-UNION ALL
-{% endif -%}
+WITH events_stream_union AS (
+  {% for (dataset, channel) in datasets %}
+    {% if not loop.first -%}
+      UNION ALL BY NAME
```
sean-rose, line 9: Using `UNION ALL BY NAME` isn't currently needed in production as all apps' app-ID-specific `events_stream_v1` tables have the same deployed schema at the moment, but it could easily become needed in production as the column order of the `events_stream_v1` ETL query has changed over time (e.g. https://github.com/mozilla/bigquery-etl/pull/5659 changed the position of the `metrics` column).

Using `UNION ALL BY NAME` _is_ currently needed for deploys to staging, which dynamically calculate the `events_stream_v1` table schemas, and since the `events_stream_v1` ETLs pass through the underlying `events_v1` table columns the `events_stream_v1` schemas can vary when the `events_v1` schemas vary (e.g. most Firefox Android events tables like `moz-fx-data-shared-prod.org_mozilla_firefox_stable.events_v1` have the `metrics` column in the middle of the table, but `moz-fx-data-shared-prod.org_mozilla_fenix_stable.events_v1` has the `metrics` column at the end of the table for some reason).

Unfortunately, `UNION ALL BY NAME` doesn't help with struct fields being in different orders, so changes like https://github.com/mozilla/bigquery-etl/pull/8066 updating `events_stream_v1` ETLs to add the `attribution` and `distribution` fields in the middle of the `client_info` struct could easily come back to bite us and break these union queries.  However, we can deal with that if/when it happens.

sql_generators/glean_usage/templates/cross_channel_events_stream.query.sql
```diff
@@ -1,18 +1,27 @@
...
+    {% endif -%}
+    SELECT
+      "{{ dataset }}" AS normalized_app_id,
+      e.*
+      {% if app_name == "fenix" -%}
+        REPLACE(mozfun.norm.fenix_app_info("{{ dataset }}", client_info.app_build).channel AS normalized_channel),
+      {% elif datasets|length > 1 -%}
+        REPLACE("{{ channel }}" AS normalized_channel),
+      {% endif -%}
+    FROM `{{ project_id }}.{{ dataset }}_derived.events_stream_v1` AS e
```
sean-rose, line 19: This is switching from unioning the `events_stream` views to unioning the `events_stream_v1` tables, because the `events_stream` views will now have `extras` structs themselves which we'd just have to exclude anyway.

sql_generators/glean_usage/events_stream.py
```diff
@@ -126,4 +148,72 @@ def generate_per_app(
...
+            )
+            or 775
+        )
+        cutoff_date = date.today() - timedelta(days=(ping_delete_after_days - 1))
+
+    metrics = list(get_glean_app_metrics(v1_name).values())
+    for metric in metrics:
+        if metric["type"] == "event":
+            for history in metric["history"]:
+                last_date = datetime.fromisoformat(history["dates"]["last"]).date()
```
scholtzan, line 202: What does the `history["dates"]["last"]` mean in the context of a probe?

sql_generators/glean_usage/events_stream.py
```diff
@@ -126,4 +148,72 @@ def generate_per_app(
...
+    for dependency in repository["dependencies"]:
+        dependency_v1_name = get_glean_dependency_v1_name(dependency)
+        dependency_extras_by_type = get_glean_app_event_extras_by_type(
+            dependency_v1_name, ping, cutoff_date
+        )
+        for (
+            depencency_extra_type,
+            depencency_extras,
+        ) in dependency_extras_by_type.items():
+            extras_by_type[depencency_extra_type].update(depencency_extras)
```
BenWu: type o
```suggestion
        for (
            dependency_extra_type,
            dependency_extras,
        ) in dependency_extras_by_type.items():
            extras_by_type[dependency_extra_type].update(dependency_extras)
```

sql_generators/glean_usage/events_stream.py
```diff
@@ -126,4 +148,72 @@ def generate_per_app(
...
+                .get("delete_after_days")
+            )
+            or 775
+        )
+        cutoff_date = date.today() - timedelta(days=(ping_delete_after_days - 1))
+
+    metrics = list(get_glean_app_metrics(v1_name).values())
+    for metric in metrics:
+        if metric["type"] == "event":
+            for history in metric["history"]:
```
BenWu, line 201: q(non-blocking): Is this iteration to get all the historical types of an extra?

sql_generators/glean_usage/events_stream.py
sean-rose: I believe it's the timestamp of the last commit when that particular historical version of the metric was present in the source code (i.e. the timestamp of the commit ID in `history["git-commits"]["last"]`), but I'm not super familiar with probe-scraper so I could be wrong.

sql_generators/glean_usage/events_stream.py
```diff
@@ -126,4 +148,72 @@ def generate_per_app(
...
+@cache
+def get_glean_app_event_extras_by_type(
+    v1_name: str, ping: str = "events", cutoff_date: Optional[date] = None
+) -> dict[str, set[str]]:
+    """Return the Glean app's event extra keys for the specified ping, grouped by type."""
+    extras_by_type = defaultdict(set)
+    repository = get_glean_app_repository(v1_name)
+
+    if not cutoff_date:
+        ping_delete_after_days = (
```
BenWu, line 182: Just fyi, it's possible that one day we will support setting retention in dependencies so the way it's using the passed in date may longer be relevant: DENG-7765.  But nothing to do now and the dependency would likely be set lower than the app so using an earlier cutoff date is fine

BenWu (approved):

sql_generators/glean_usage/events_stream.py
sean-rose: Thanks for catching that!  Can't believe I let that slip through. 😬

sql_generators/glean_usage/events_stream.py
sean-rose: It's to get all of the event's configured extras during the time period in question.  Extras changing type is one common scenario (particularly since the `type` setting isn't required).  Other scenarios could be extras being renamed, or extras being removed.

sql_generators/glean_usage/events_stream.py
sean-rose: Thanks for the heads up!  I've added myself as a watcher on DENG-7765.

Since events stream currently only uses the core events ping I don't think that would actually be a factor at the moment, but if we ever do events stream for other pings it may well become a factor (DENG-8596).

PR #8362, opened 2025-10-29
DENG-9849: Update copy dedup code to null out city and subdivision fields for browser apps

wwyc:
## Description

This PR updates the copy dedup logic to null out city and subdivision fields for selected apps.

* DENG-9849


bigquery_etl/copy_deduplicate.py
```diff
@@ -13,17 +13,22 @@
...
 from google.cloud import bigquery
 
 from bigquery_etl.cli.utils import table_matches_patterns
+from bigquery_etl.config import ConfigLoader
 from bigquery_etl.util.bigquery_id import sql_table_id
 from bigquery_etl.util.client_queue import ClientQueue
 from bigquery_etl.util.common import TempDatasetReference
 
 from .cli.utils import parallelism_option, project_id_option
+from .shredder.config import get_glean_channel_to_app_name_mapping
```
BenWu: This function should be moved to somewhere like `bigquery_etl.util.common` if it's being used outside of shredder

bigquery_etl/copy_deduplicate.py
```diff
@@ -13,17 +13,22 @@
...
+
+GLEAN_APP_LISTINGS_URL = "https://probeinfo.telemetry.mozilla.org/v2/glean/app-listings"
```
BenWu: I don't think this is used here
```suggestion
```

bqetl_project.yaml
```diff
@@ -521,6 +521,9 @@ generate:
     - focus_android
     - klar_ios
     - klar_android
+  null_geo:
+    apps:
+    - klar_android
```
BenWu: If this is going to be a separate config from the `baseline_clients_city_seen` one, I think this shouldn't be nested in `generate`.

Also it would simplify the code a bit if app ids were used here.  That seems more appropriate to me because it matches the granularity of copy deduplicate

BenWu (commented):
Approach looks ok to me so far

bqetl_project.yaml
```diff
@@ -525,6 +525,12 @@ generate:
     - firefox_desktop_background_defaultagent
     - firefox_desktop_background_tasks
 
+geo_deprecation:
+  app_id:
+  - org_mozilla_ios_klar
+  table_id:
+  - baseline_v1
```
akkomar: How will we ultimately configure this?
This approach includes a product of (any listed app x any listed table) - is this intended?
This is also opt-in, so any new ping/app will have to be added manually. Are we planning to switch this to opt-out?

bigquery_etl/copy_deduplicate.py
```diff
@@ -73,14 +75,58 @@
...
+
+
+def _select_geo(live_table: str, client: bigquery.Client) -> str:
+    """Build a SELECT REPLACE clause that NULLs metadata.geo.* if applicable."""
+    _, dataset_id, table_id = live_table.split(".")
+    app_id = dataset_id.removesuffix("_live")
+    included_apps = set(ConfigLoader.get("geo_deprecation", "app_id", fallback=[]))
+    included_tables = set(ConfigLoader.get("geo_deprecation", "table_id", fallback=[]))
+
+    if (app_id not in included_apps) or (table_id not in included_tables):
```
wwyc: add check for client_id

bqetl_project.yaml
wwyc: Thanks for the comment! Right now, we’re deprecating city/subdivision fields only for selected app_ids.  Once we have the full list, we can switch to an opt-out model.

bigquery_etl/copy_deduplicate.py
```diff
@@ -73,14 +75,72 @@
...
+
+
+def _select_geo(live_table: str, client: bigquery.Client) -> str:
+    """Build a SELECT REPLACE clause that NULLs metadata.geo.* if applicable."""
+    _, dataset_id, table_id = live_table.split(".")
+
+    excluded_tables = set(
+        ConfigLoader.get("geo_deprecation", "skip_tables", fallback=[])
+    )
+    if any(excluded in table_id for excluded in excluded_tables):
```
BenWu: This feels too imprecise since e.g. `newtab_content` would be excluded by having `newtab` in the list.  Maybe that's desired in this case, but I think the list should be more explicit on which pings are excluded
```suggestion
    if table_id.removesuffix("_v1") in excluded_tables:
```

bigquery_etl/copy_deduplicate.py
```diff
@@ -73,14 +75,72 @@
...
+        "      (SELECT AS STRUCT"
+        "         metadata.geo.* REPLACE ("
+        "           CAST(NULL AS STRING) AS city,"
+        "           CAST(NULL AS STRING) AS subdivision1,"
+        "           CAST(NULL AS STRING) AS subdivision2"
+        "         )"
+        "      ) AS geo"
+        "    )"
+        " ) AS metadata)"
+    )
```
BenWu: A multiline string would make this easier to maintain
```suggestion
    return """
        REPLACE (
        (SELECT AS STRUCT
           metadata.* REPLACE (
             (SELECT AS STRUCT
                metadata.geo.* REPLACE (
                  CAST(NULL AS STRING) AS city,
                  CAST(NULL AS STRING) AS subdivision1,
                  CAST(NULL AS STRING) AS subdivision2
                )
             ) AS geo
           )
        ) AS metadata)
    """
```

wwyc:
@BenWu if i can get an approval from your end perhaps I can merge next week? Thanks for the thorough review!

PR #8371, opened 2025-10-30
fix: add glean_usage/templates/events_stream_v1.schema.yaml

kik-kik:
# fix: add glean_usage/templates/events_stream_v1.schema.yaml

caused by: https://github.com/mozilla/bigquery-etl/pull/8363

It looks like now schema files in the glean_usage generator are required for query.py templates. This seems to have caused the `'events_stream_v1.schema.yaml' not found in search path` error in our deployment.


kik-kik:
This will not work as the schema will vary for different apps. Closing for this reason.

PR #8393, opened 2025-11-04
[SVCSE-3019] Move regrets-reporter apps to glean_usage skip_apps

scholtzan:
## Description

Stable views are still being generated because I moved the regrets reporter apps into the wrong skip section. This should fix it.

* https://mozilla-hub.atlassian.net/browse/SVCSE-3019

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/doh_adoption_rate_v1/backfill.yaml
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/doh_adoption_rate_v1/backfill.yaml
@@ -4,4 +4,4 @@
   reason: Inital backfill of DOH adoption rate data after table changes
   watchers:
   - ascholtz@mozilla.com
-  status: Initiate
+  status: Complete


bqetl_project.yaml
```diff
@@ -361,6 +361,11 @@ generate:
     - regrets_reporter
     - tiktokreporter_android
     - tiktokreporter_ios
+    - regrets_reporter
+    - org_mozilla_ios_tiktokreporter
+    - org_mozilla_tiktokreporter
+    - regrets_reporter_ucs
+    - org_mozilla_ios_tiktok_reporter_tiktok_reportershare
```
sean-rose, line 368: **issues (blocking):**
- `regrets_reporter` is already in the `generate.glean_usage.skip_apps` list, and after #8345 I'm pretty sure that's properly blocking `glean_usage` from doing anything with that app, including its app ID `regrets.reporter.ucs`.
- Similarly, `tiktokreporter_android` and `tiktokreporter_ios` are already in the `generate.glean_usage.skip_apps` list, which should also be blocking `glean_usage` from doing anything with its app IDs `org.mozilla.ios.TikTok-Reporter`, `org.mozilla.ios.TikTok-Reporter.TikTok-ReporterShare`, and `org.mozilla.tiktokreporter`.

scholtzan:
Nvm, generated-sql is now up-to-date without the regrets-reporter datasets. It took a while to propagate after removing them from the repositories.yaml: https://github.com/mozilla/probe-scraper/pull/958
My guess it's that it is related to caching

bqetl_project.yaml
```diff
@@ -486,11 +491,6 @@ generate:
   stable_views:
     skip_datasets:
     - mlhackweek_search
-    - regrets_reporter
-    - org_mozilla_ios_tiktokreporter
-    - org_mozilla_tiktokreporter
-    - regrets_reporter_ucs
-    - org_mozilla_ios_tiktok_reporter_tiktok_reportershare
```
sean-rose, line 493: **issue (blocking):** I wouldn't recommend removing this, as it _should_ work, except there appears to be a bug in [this `stable_views` code](https://github.com/mozilla/bigquery-etl/blob/991f37d85e868e5ec3cfef0c378b359889486795/sql_generators/stable_views/__init__.py#L460-L472) where the `skip_datasets` filter should be applied to the `schemas` list rather than just to the `one_schema_per_dataset` list (and if the `skip_datasets` filter is applied to the `schemas` list it wouldn't need to be applied to the `one_schema_per_dataset` list, since that's based directly on the `schemas` list).

sean-rose: > there appears to be a bug in [this `stable_views` code](https://github.com/mozilla/bigquery-etl/blob/991f37d85e868e5ec3cfef0c378b359889486795/sql_generators/stable_views/__init__.py#L460-L472) where the `skip_datasets` filter should be applied to the `schemas` list

I've submitted a fix for that bug: https://github.com/mozilla/bigquery-etl/pull/8394

PR #8395, opened 2025-11-04
feat(DENG-9467): set up terms_of_use generator

kik-kik:
# feat(DENG-9467): set up terms_of_use generator

## Description

This adds a new generator for generating ToS related queries for 3 different applications:

- firefox_ios - terms_of_use_status_v1
- fenix - terms_of_use_status_v1
- firefox_desktop - terms_of_use_status_v1 && terms_of_use_events_v1

This is to support ToS related analysis.

### DIFF
--- a/sql_generators/terms_of_use/terms_of_use.py
+++ b/sql_generators/terms_of_use/terms_of_use.py
@@ -77,9 +77,9 @@ def generate_terms_of_use(target_project: str, output_dir: Path):
             "header": HEADER,
             "dag_name": dag_name,
             "project_id": target_project,
-            "bigeye_collection": app_config.get("bigconfig", dict()).get("collection")
+            "bigeye_collection": app_config.get("bigeye", dict()).get("collection")
             or bigeye_defaults["collection"],
-            "slack_notification_channel": app_config.get("bigconfig", dict()).get(
+            "slack_notification_channel": app_config.get("bigeye", dict()).get(
                 "notification_channel"
             )
             or bigeye_defaults["notification_channel"],


sql_generators/terms_of_use/templates/terms_of_use_status_v1/query.sql.jinja
```diff
@@ -0,0 +1,76 @@
...
+    isp_name,
+    terms_of_use_version_accepted,
+    terms_of_use_date_accepted,
+    {% if app_name == "firefox_desktop" %}
+    legacy_telemetry_client_id,  -- firefox_desktop exclusive field.
+    {% endif %}
+FROM
+  `{{ project_id }}.{{ app_name }}_derived.{{ table_name }}`
+WHERE
+  DATE(submission_timestamp) <= @submission_date
```
kik-kik: TODO: need to think more about whether this should be `<>` instead.

sql_generators/terms_of_use/templates/terms_of_use_events_v1/query.sql.jinja
```diff
@@ -0,0 +1,33 @@
+-- {{ header }}
+SELECT
+  submission_timestamp,
+  metrics.uuid.messaging_system_client_id AS legacy_telemetry_client_id,
```
kik-kik, line 6: @soGaussian @irrationalagent , should `client_id` also be included in this query?

sql_generators/terms_of_use/templates/terms_of_use_events_v1/schema.yaml.jinja
```diff
@@ -0,0 +1,51 @@
...
+  name: normalized_event
+  type: STRING
+  description: |
+    The events for the infobar are different than the events for the modal, this field contains standardized modal events so they look like the infobar events.
+
+- mode: NULLABLE
+  name: messaging_surface
+  type: STRING
+  description: |
+    TODO
```
kik-kik: @soGaussian could you help me come up with accurate description of this field?

sql_generators/terms_of_use/templates/terms_of_use_status_v1/query.sql.jinja
```diff
@@ -0,0 +1,95 @@
...
+        UNBOUNDED PRECEDING
+        AND UNBOUNDED FOLLOWING
+    )
+)
+SELECT
+  -- update entry if `terms_of_use_version_accepted` or `terms_of_use_date_accepted` value changes:
+  IF(
+    _previous.client_id IS NULL
+    OR (
+      (_current.terms_of_use_version_accepted <> _previous.terms_of_use_version_accepted)
```
kik-kik, line 84: Should we also add an additional check here to make sure we're not overwriting "future" entries?

sql_generators/terms_of_use/templates/terms_of_use_events_v1/schema.yaml.jinja
soGaussian: The surface the terms of use prompt/impression was served

sql_generators/terms_of_use/templates/terms_of_use_events_v1/query.sql.jinja
soGaussian: For now we will not use glean client_id in any downstream aggregation or joins and so I will recommend we keep it out for now

sql_generators/terms_of_use/templates/terms_of_use_status_v1/query.sql.jinja
soGaussian: I think the version is sufficient here

sql_generators/terms_of_use/templates/terms_of_use_events_v1/query.sql.jinja
kik-kik: For now? But that means we plan to use it later? The main concern here is if we add it later then wouldn't we need to recompute the table to make sure the value is present in historical rows @soGaussian ?

sql_generators/terms_of_use/templates/terms_of_use_status_v1/query.sql.jinja
kik-kik: Sorry, what I meant here. Should we have an additional guard to not overwrite entries from _previous CTE if their accepted date is newer than the one in the _current CTE?

sql_generators/terms_of_use/templates/terms_of_use_events_v1/query.sql.jinja
soGaussian: To be honest I dont think I can argue for down/up side of adding the glean client id telemetry. I am ok with whatever you deem necessary

kwindau (approved):

PR #8399, opened 2025-11-05
Add job to create propensity stats table for use in ranking 

rolf-moz:
## Description

We are seeking to collect some data about relative clickthrough rates of items in a section independent of content. This may be applied in calculating live ranking data in the newtab_merino_extract_v3 job that runs every 20 minutes.

Specifically, we might reduce or increase the estimated number of real-time clicks based on which position the user clicked on based on the artifact created in this job.

As part of ## Related Tickets & Documents
https://mozilla-hub.atlassian.net/browse/GENAI-1831)


sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_propensity_v1/query.sql
```diff
@@ -0,0 +1,185 @@
...
+        ANY_VALUE(x.value)
+      FROM
+        UNNEST(ev.extra) AS x
+      WHERE
+        x.key = 'corpus_item_id'
+    ) AS corpus_item_id,
+    ev.name AS event_name
+  FROM
+    params,
+    `moz-fx-data-shared-prod.firefox_desktop_stable.newtab_content_v1` AS e
```
mmiermans, line 54: I don't think it matters much because this job runs weekly, but would `moz-fx-data-shared-prod.firefox_desktop_derived.newtab_content_items_daily_combined_v1` have the data you need?

mmiermans (commented):
I think you'll need a job similar to [this one](https://github.com/rolf-moz/bigquery-etl/tree/9ef539acb3a6dde0ba9140ef17f94835024c82f3/sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_priors_to_gcs_v1) to write it to GCS.

sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_propensity_v1/query.sql
rolf-moz: No because we are looking here at the exact moment that the corpus item was created, so we're able to look at it's CTR at a time when it is being ranked randomly independent of the article. Otherwise the data we'd get would be biased by the article ranking.

PR #8407, opened 2025-11-07
[DENG-10153] Ignore derived_view_schemas generator in CI

scholtzan:
## Description

The `derived_view_schemas` SQL generator is with about 7min the longest running generator (makes about 1/2 of the 15min long generate-sql generation process). This generator creates schemas for views in order to propagate description from referenced tables. These schemas are only useful before deploying a view, so in order to save time, the generator can be skipped in CI. This will result in view schemas being removed from the generated-sql branch, but these schemas aren’t generally useful to users or used anywhere other than for deployment.

An implication of this change is that the `derived_view_schemas always` needs to run before view deploys. 

Depends on https://github.com/mozilla/telemetry-airflow/pull/2279

* https://mozilla-hub.atlassian.net/browse/DENG-10153

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -591,7 +591,10 @@ jobs:
 
                     cp -r sql/ /tmp/workspace/generated-sql/sql
                     # Don't depend on dry run for PRs
+                    # Ignore derived_view_schemas as this will add descriptions to view which take a long time;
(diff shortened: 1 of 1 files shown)


scholtzan:
I won't merge before next week, just in case this breaks the deployment

PR #8413, opened 2025-11-10
[DENG-10166] Blobless checkout in CI

scholtzan:
## Description

Using the `blobless` checkout method in CI reduces the checkout time from around 30s to 5s. Blobless clones reduce the amount of data fetched from the remote by asking the remote to filter out objects that are not attached to the current commit: https://circleci.com/docs/reference/configuration-reference/#checkout

* https://mozilla-hub.atlassian.net/browse/DENG-10166

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -4,7 +4,6 @@ version: 2.1
 orbs:
   gcp-gcr: circleci/gcp-gcr@0.16
   docker: circleci/docker@2.6
-  python: circleci/python@2.1.1
   gcp-cli: circleci/gcp-cli@3.2
 
 parameters:
@@ -38,7 +37,7 @@ executors:
 jobs:
   build:
     docker: &docker
-      - image: python:<< pipeline.parameters.python-version >>
+      - image: cimg/python:<< pipeline.parameters.python-version >>
     steps:
       - checkout:
           method: blobless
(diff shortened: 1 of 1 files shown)


.circleci/workflows.yml
```diff
@@ -38,22 +37,23 @@ executors:
 jobs:
   build:
     docker: &docker
-      - image: python:<< pipeline.parameters.python-version >>
+      - image: cimg/python:<< pipeline.parameters.python-version >>
```
scholtzan, line 40: Using the convience image supposedly also provides [faster spin-up times](https://circleci.com/docs/guides/execution-managed/circleci-images/). Based on my tests, the difference was not very noticeable.

.circleci/workflows.yml
```diff
@@ -38,22 +37,23 @@ executors:
...
+          method: blobless
       - &restore_venv_cache
         restore_cache:
           keys:
             # when lock files change, use increasingly general
             # patterns to restore cache
             - &python_cache_key
               # yamllint disable-line rule:line-length
-              python-<< pipeline.parameters.python-version >>-packages-v1-{{ .Branch }}-{{ checksum "requirements.in" }}-{{ checksum "requirements.txt" }}
+              python-<< pipeline.parameters.python-version >>-packages-v2-{{ .Branch }}-{{ checksum "requirements.in" }}-{{ checksum "requirements.txt" }}
```
scholtzan, line 51: invalidate cache since new image is being used

.circleci/workflows.yml
```diff
@@ -38,22 +37,23 @@ executors:
 jobs:
   build:
     docker: &docker
-      - image: python:<< pipeline.parameters.python-version >>
+      - image: cimg/python:<< pipeline.parameters.python-version >>
     steps:
-      - checkout
+      - checkout:
+          method: blobless
```
sean-rose, line 43: **question:** I'm confused, because according to [the docs you referenced](https://circleci.com/docs/reference/configuration-reference/#checkout) `blobless` is the default method.  So isn't this a no-op?

scholtzan: Yeah, that's what I thought too, but it seems to default to a `full` checkout (e.g. a task where [it's not specified to compare](https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/54852/workflows/5d6a3583-5d48-4178-b424-7aeae4700d98/jobs/696558))

PR #8421, opened 2025-11-11
fix(SubPlat): Try to avoid having Apple/Google subscription records' Mozilla Account IDs change after the fact (DENG-9565)

sean-rose:
## Description
Not all Apple/Google subscription records have Mozilla Account ID values, so we fall back to trying to find the Mozilla Account ID in the subscription's other nearby records.  However, during backfills of SubPlat ETLs we noticed that some historical Apple subscription records were having their Mozilla Account ID value changed retroactively, which ideally shouldn't happen.

The changes in this PR try to avoid that happening by looking at records for the same purchase/period first, and looking at preceding records first before following records.

While this will cause more existing historical Apple/Google subscription records to have their Mozilla Account ID value changed retroactively the next time those are backfilled, it ought to reduce how often this problem occurs going forward.

* DENG-9565: Fix some known issues in SubPlat consolidated reporting ETLs

### DIFF
--- a/sql/moz-fx-data-shared-prod/subscription_platform_derived/google_logical_subscriptions_history_v1/query.sql
+++ b/sql/moz-fx-data-shared-prod/subscription_platform_derived/google_logical_subscriptions_history_v1/query.sql
@@ -27,26 +27,18 @@ WITH subscriptions_history AS (
     -- the Mozilla Account ID from the subscription's other nearby records.
     COALESCE(
       subscription.metadata.user_id,
-      FIRST_VALUE(subscription.metadata.user_id IGNORE NULLS) OVER (
-        PARTITION BY
-          original_subscription_purchase_token
-        ORDER BY
-          valid_from,
-          valid_to
-        ROWS BETWEEN
-          1 FOLLOWING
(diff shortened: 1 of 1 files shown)


phil-lee70 (approved):
Approved. Just curious, do we have an approximate estimate how often this fallback logic gets used in production?

sean-rose:
> Just curious, do we have an approximate estimate how often this fallback logic gets used in production?

That logic technically gets used for all Apple/Google subscriptions, because SubPlat creates the initial Firestore records for those without a user ID before immediately updating them with the user ID.  However, this change doesn't effect that scenario because there aren't both preceding and following records with user IDs to choose from in that case (just following records).

Thus far, it looks like a slight majority of Apple/Google subscriptions have had such cases where the user ID is missing and there are both preceding and following records with user IDs to choose from (53% of Apple subscriptions and 66% of Google subscriptions).

However, the edge cases where this logic change will make an actual difference aren't that common, as I estimate only 66 historical `apple_logical_subscriptions_history_v1` records and 2 historical `google_logical_subscriptions_history_v1` records will have their Mozilla Account ID values changed as a result of this PR.

PR #8434, opened 2025-11-13
fix(glean_usage): Add new `subscription_platform_backend_cirrus` app to skip lists

sean-rose:
## Description
Like the other Cirrus apps.

* https://github.com/mozilla/probe-scraper/pull/963

### DIFF
--- a/bqetl_project.yaml
+++ b/bqetl_project.yaml
@@ -420,6 +420,7 @@ generate:
       - mozilla_vpn
       - mozillavpn_backend_cirrus
       - pine
+      - subscription_platform_backend_cirrus
       - thunderbird_android
       - thunderbird_desktop
     clients_last_seen_joined:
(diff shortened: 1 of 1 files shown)


bqetl_project.yaml
```diff
@@ -434,6 +435,7 @@ generate:
       - mozilla_vpn
       - mozillavpn_backend_cirrus
       - pine
+      - subscription_platform_backend_cirrus
```
BenWu, line 438: Is this needed even if the app doesn't have a baseline ping table?  I thought it would already be skipped

sean-rose: Ahh, good point!  Since this new `subscription_platform_backend_cirrus` app depends on `glean-server` rather than `glean-core` like the previous Cirrus apps I guess it won't have most of the core ping tables like those other apps did (except for `events_v1`).  I'll close this PR.

sean-rose:
Never mind, it turns out this isn't actually needed since this app doesn't depend on `glean-core` like the other Cirrus apps.

PR #8445, opened 2025-11-13
fix(query): Have `bqetl query run` redeploy the table schema if necessary to add back field descriptions

sean-rose:
## Description
Because when an ETL overwrites the entire table the field descriptions unfortunately get removed.

N/A

### DIFF
--- a/bigquery_etl/cli/query.py
+++ b/bigquery_etl/cli/query.py
@@ -1096,6 +1096,8 @@ def _run_query(
             destination_table
             and "$" not in destination_table
             and "--dry_run" not in query_arguments
+            and "--append_table" not in query_arguments
+            and "--noreplace" not in query_arguments
             and not is_backfill
         ):
             schema_file = query_file.parent / SCHEMA_FILE


bigquery_etl/cli/query.py
```diff
@@ -1089,6 +1089,23 @@ def _run_query(
             # run the query as shell command so that passed parameters can be used as is
             subprocess.check_call(["bq"] + query_arguments, stdin=query_stream)
 
+        # If the entire table was overwritten then redeploy the table schema to add back field descriptions.
+        if destination_table and "$" not in destination_table:
```
BenWu: Checking and skipping for a `--dry_run` argument would be a good idea

bigquery_etl/cli/query.py
```diff
@@ -1089,6 +1089,23 @@ def _run_query(
...
+                    schema_destination_table = destination_table.replace(":", ".")
+                else:
+                    schema_destination_table = ".".join(
+                        (
+                            (project_id or default_project),
+                            (dataset_id or default_dataset),
+                            destination_table,
+                        )
+                    )
+                schema.deploy(schema_destination_table)
```
BenWu, line 1119: non-blocking: I think this could fail if there are concurrent writes to the same table and `client.get_table`/`client.update_table` get called from multiple processes around the same time.  I expect this will be fine since I don't think there would be a case in which we have concurrent writes to an entire destination table, but worth keeping in mind

sean-rose: Yeah, the Airflow DAG generation logic sets `task_concurrency=1` for ETLs that overwrite the entire table, so under normal operation I don't think that could happen.

sean-rose: I guess one other scenario where there could be concurrent updates to a table is when the `bqetl_artifact_deployment` DAG's `publish_new_tables` or `publish_metadata` tasks are running at the same time as the ETL itself.  However, since the schema deployment process updates the table immediately after fetching it, I don't think there's much of a chance of that actually failing due to the table getting modified between those two calls and having a different ETag.

sean-rose:
> Possibly blocking issue: tables have a maximum of 5 operations per 10 seconds ([doc](https://docs.cloud.google.com/bigquery/quotas#standard_tables)) and that could be hit during managed backfills of a small depends_on_past table. Those queries would run sequentially so it's unlikely to have more than 5 finish in 10 seconds, especially since the schema deploy takes time, but it's a small risk. It's probably ok to keep this as-is if there's no simple solution to this

I updated the logic to not redeploy the table schema for backfills. 👍

bigquery_etl/cli/query.py
```diff
@@ -1089,6 +1091,33 @@ def _run_query(
             # run the query as shell command so that passed parameters can be used as is
             subprocess.check_call(["bq"] + query_arguments, stdin=query_stream)
 
+        # If the entire table was overwritten then redeploy the table schema to add back field descriptions.
+        if (
+            destination_table
+            and "$" not in destination_table
+            and "--dry_run" not in query_arguments
+            and "--append_table" not in query_arguments
+            and "--noreplace" not in query_arguments
```
sean-rose, line 1100: Technically there are additional ways these arguments could be passed, such as `--dry_run=true`, `--append_table=true`, or `--replace=false`, as well as variants without the equal signs which would be more complex to handle correctly.

However, in practice our existing code seems to consistently specify these arguments in this form, so I'm inclined to keep this simple and not try to handle all the possible variants.

bigquery_etl/cli/query.py
BenWu: Yeah I think that's unlikely enough to not need solving

PR #8477, opened 2025-11-18
[DENG-10167] Speed up DAG generation

scholtzan:
## Description

This PR speeds up the DAG generation process by:
1. caching the stable schemas that get downloaded as part of the `generate-sql` step and reuses them during DAG generation (save around 1.5minutes). If the `generate-sql` step gets skipped then it will still download schemas
2. parallelizing extracting the table references needed for generating DAG task dependencies (saving around 1 minute)

* https://mozilla-hub.atlassian.net/browse/DENG-10167

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -673,6 +673,7 @@ jobs:
             - *restore_venv_cache
             - *build
             - *attach_generated_sql
+            - *restore_schema_cache
             - *copy_generated_sql
             - add_ssh_keys:
                 fingerprints:


.circleci/workflows.yml
```diff
@@ -610,6 +619,7 @@ jobs:
                       --ignore derived_view_schemas \
                       --output-dir /tmp/workspace/generated-sql/sql/ \
                       --target-project moz-fx-data-shared-prod
+                    PATH="venv/bin:$PATH" script/bqetl format /tmp/workspace/generated-sql/sql/
```
scholtzan, line 631: We don't need to run the formatting if we use the existing `generated-sql` branch since anything on there is already properly formatted. The formatting does take 2-3 minutes to run

kwindau (approved):

.circleci/workflows.yml
BenWu: IIRC a lot (most?) of that is due to qualifying table references.  That could be another thing to optimize but it seems hard to do

PR #8485, opened 2025-11-20
feat(query): Add `bqetl query render` support for Jinja imports/includes from outside the ETL directory (DENG-9950)

sean-rose:
## Description
This implements the accepted approach from [Proposal: Add `bqetl query render` support for Jinja macros defined outside the `query.sql` file](https://docs.google.com/document/d/1OazrnKmC11SBxHfVUFoO03yxrO3MWeW9sz6rx6QXxjo/edit?tab=t.0).

* DENG-9950: Add `bqetl query render` support for Jinja macros defined outside the `query.sql` file

### DIFF
--- a/bigquery_etl/util/common.py
+++ b/bigquery_etl/util/common.py
@@ -8,6 +8,7 @@ import re
 import string
 import tempfile
 import warnings
+from functools import cache
 from pathlib import Path
 from typing import List, Optional, Set, Tuple
 from uuid import uuid4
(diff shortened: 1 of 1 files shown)


bigquery_etl/util/common.py
```diff
@@ -122,14 +123,22 @@ def render(
...
+                    + (template_folder_path / sql_filename).read_text()
                 )
 
                 file_loader = FileSystemLoader(f"{str(checks_template.parent)}")
                 env = Environment(loader=file_loader)
                 main_sql = env.get_template(checks_template.name)
         else:
-            file_loader = FileSystemLoader(f"{template_folder}")
+            # Add the bigquery-etl project root to the search path to support Jinja imports/includes.
+            file_loader_search_paths = [template_folder_path, ROOT]
```
BenWu, line 147: non-blocking: does adding the root to the loader have any measurable difference in run time or memory usage?  I would guess it's insignificant but I have no idea

sean-rose: No, because the `FileSystemLoader` isn't doing anything like eagerly loading/indexing the contents of the search path directories.  When trying to resolve template/include/import paths it's just joining those input paths with each configured search path in turn and testing if the resulting file path exists, with the first one winning.

scholtzan (approved):

PR #8486, opened 2025-11-20
Dryrun caching

scholtzan:
## Description

This PR adds a cache to `DryRun` which caches query and table schemas for a configurable amount of time and as long as the query content is the same. This reduces the schema generation time slightly which is executed before table artifact deploys (locally it saved 3 to 4 minutes).

### DIFF
--- a/bigquery_etl/dryrun.py
+++ b/bigquery_etl/dryrun.py
@@ -248,15 +248,15 @@ class DryRun:
         cache_input = f"{sql}|{self.project}|{self.dataset}|{self.table}"
         return hashlib.sha256(cache_input.encode()).hexdigest()
 
-    def _get_cached_result(self, cache_key, ttl_seconds=None):
-        """Load cached dry run result from disk."""
-        if ttl_seconds is None:
-            ttl_seconds = ConfigLoader.get("dry_run", "cache_ttl_seconds", fallback=900)
-
+    @staticmethod
+    def _get_cache_dir():
+        """Get the cache directory path."""
         cache_dir = Path(tempfile.gettempdir()) / "bigquery_etl_dryrun_cache"
         cache_dir.mkdir(parents=True, exist_ok=True)
-        cache_file = cache_dir / f"dryrun_{cache_key}.pkl"
+        return cache_dir
 
+    def _read_cache_file(self, cache_file, ttl_seconds):
+        """Read and return cached data from a pickle file with TTL check."""
         try:
             if not cache_file.exists():
                 return None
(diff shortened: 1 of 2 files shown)


scholtzan:
dryrun failing due to unrelated error in `sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_propensity_v1/query.sql`

scholtzan:
I won't merge this before next week.

wwyc (approved):

bigquery_etl/dryrun.py
```diff
@@ -225,6 +243,145 @@ def get_sql(self):
...
+    def _save_cached_result(self, cache_key, result):
+        """Save dry run result to disk cache using atomic write."""
+        cache_dir = Path(tempfile.gettempdir()) / "bigquery_etl_dryrun_cache"
+        cache_dir.mkdir(parents=True, exist_ok=True)
+        cache_file = cache_dir / f"dryrun_{cache_key}.pkl"
+
+        try:
+            # write to temporary file first, then atomically rename
+            # this prevents race conditions where readers get partial files
+            temp_file = Path(str(cache_file) + f".tmp.{os.getpid()}")
```
BenWu: I think there's a possible edge case if this is called from a thread pool where the threads have the same PID an extra random string might be better.
```suggestion
            temp_file = Path(str(cache_file) + f".tmp.{os.getpid()}.{os.urandom(4).hex()}")
```

bigquery_etl/dryrun.py
```diff
@@ -225,6 +243,145 @@ def get_sql(self):
...
+        if ttl_seconds is None:
+            ttl_seconds = ConfigLoader.get("dry_run", "cache_ttl_seconds", fallback=900)
+
+        cache_dir = Path(tempfile.gettempdir()) / "bigquery_etl_dryrun_cache"
+        cache_dir.mkdir(parents=True, exist_ok=True)
+        # table identifier as cache key
+        table_cache_key = hashlib.sha256(table_identifier.encode()).hexdigest()
+        cache_file = cache_dir / f"table_metadata_{table_cache_key}.pkl"
+
+        try:
```
BenWu: The cache read and write code is almost identical between the metadata and the dry run result.  Moving the common code into a common function would make this easier to maintain

bigquery_etl/cli/query.py
```diff
@@ -2300,6 +2300,9 @@ def _update_query_schema(
         query_schema = Schema.from_query_file(
             query_file_path,
             content=sql_content,
+            project=project_name,
+            dataset=dataset_name,
+            table=table_name,
```
sean-rose, line 2305: It turns out this particular change is causing problems with deploying ETLs to stage:
- https://github.com/mozilla/bigquery-etl/pull/8598

PR #8498, opened 2025-11-24
DENG-10202 Changed geo deprecation to opt out config

wwyc:
## Description

The current set up from #8362  uses an `opt-in` list to deprecate geo (`city` and `subdivision`) fields.

This PR changes the set up to use an `opt-out` approach to exempt specific apps and tables.  In this case, any apps not listed in the config file (including future apps added) will have geo fields deprecated by default.

* DENG-10202


bigquery_etl/cli/utils.py
```diff
@@ -250,3 +252,25 @@ def temp_dataset_option(
         help="Dataset where intermediate query results will be temporarily stored, "
         "formatted as PROJECT_ID.DATASET_ID",
     )
+
+
+def get_glean_channel_to_app_name_mapping() -> Dict[str, str]:
```
BenWu: I recommend renaming this to something more general now that it's used in more places
```suggestion
def get_glean_app_id_to_app_name_mapping() -> Dict[str, str]:
```

BenWu: This should get cached otherwise copy_deduplicate is going to make ~900 calls to https://probeinfo.telemetry.mozilla.org/v2/glean/app-listings
```suggestion
from functools import cache

@cache
def get_glean_channel_to_app_name_mapping() -> Dict[str, str]:
```

bigquery_etl/copy_deduplicate.py
```diff
@@ -98,21 +101,22 @@ def _select_geo(live_table: str, client: bigquery.Client) -> str:
...
-    app_id = dataset_id.removesuffix("_live")
-    included_apps = set(
-        ConfigLoader.get("geo_deprecation", "include_app_ids", fallback=[])
-    )
-    if app_id not in included_apps:
-        return ""
-
     table = client.get_table(live_table)
 
+    # only glean tables have this label
```
BenWu, line 123: Does this mean we're explicitly only doing this for glean apps?  I don't remember if this was discussed already

bqetl_project.yaml
```diff
@@ -561,25 +559,12 @@ generate:
...
-  - firefox_desktop_background_defaultagent
-  - firefox_desktop_background_tasks
+  skip_apps:
+  - ads_backend
+  - mozilla_vpn
+  - mozillavpn_backend_cirrus
+  - accounts_backend
+  - accounts_frontend
   skip_tables:
   - newtab
```
BenWu, line 573: `newtab_content` is already covered by the `include_client_id` check but I think we should list as well since it's specifically called out in the decision brief, just to be clear on what we're intentionally excluding.
```suggestion
  - newtab
  - newtab_content
```

bigquery_etl/copy_deduplicate.py
wwyc: yes we are only deprecating the geo fields for glean apps/tables for now.  After the glean migration is completed the legacy tables would be deprecated.

kwindau (approved):

tests/test_copy_deduplicate.py
```diff
@@ -0,0 +1,461 @@
...
+    def test_select_geo_when_not_glean_app(
+        self,
+        mock_mapping,
+        mock_config_get,
+        geo_deprecation_config_side_effect,
+    ):
+        """If app_id is not in the glean mapping, we should get an empty string."""
+        mock_mapping.return_value = GLEAN_MAPPING
+        mock_config_get.side_effect = geo_deprecation_config_side_effect
+        mock_client = Mock(spec=bigquery.Client)
```
BenWu: This setup is in every test so they could be defined as fixtures

BenWu (approved):
One suggestion but otherwise LGTM

PR #8517, opened 2025-11-27
fix: add additional spelling of country name for code ST

kik-kik:
# fix: add additional spelling of country name for code ST

Currently this is causing a few country null values inside `firefox_ios_derived.app_store_funnel_v1`


kik-kik:
Actually, this entry already exists in the CSV. It looks like the BQ table has not been correctly updated.

PR #8522, opened 2025-11-29
Metadata label and CI validation for table levels based on medallion architecture

lucia-vargas-a:
Enables adding a metadata label to set the table level as one of [gold, silver, bronze].

* [DENG-10161](https://mozilla-hub.atlassian.net/browse/DENG-10161)

[DENG-10161]: https://mozilla-hub.atlassian.net/browse/DENG-10161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bigquery_etl/cli/utils.py
+++ b/bigquery_etl/cli/utils.py
@@ -4,11 +4,13 @@ import fnmatch
 import os
 import re
 from fnmatch import fnmatchcase
+from functools import cache
 from glob import glob
 from pathlib import Path
-from typing import Iterator, List, Optional, Tuple
+from typing import Dict, Iterator, List, Optional, Tuple
 
 import click
+import requests
 from google.auth.exceptions import DefaultCredentialsError
 from google.cloud import bigquery
 
@@ -23,6 +25,7 @@ CHECKS_FILE_RE = re.compile(
     r"^.*/([a-zA-Z0-9-]+)/([a-zA-Z0-9_]+)/([a-zA-Z0-9_]+(_v[0-9]+)?)/"
     r"(?:checks\.sql)$"
 )
+GLEAN_APP_LISTINGS_URL = "https://probeinfo.telemetry.mozilla.org/v2/glean/app-listings"
 
 
 def is_valid_dir(ctx, param, value):
@@ -250,3 +253,26 @@ def temp_dataset_option(
         help="Dataset where intermediate query results will be temporarily stored, "
         "formatted as PROJECT_ID.DATASET_ID",
     )
+
+
+@cache
+def get_glean_app_id_to_app_name_mapping() -> Dict[str, str]:
+    """Return a dict where key is the channel app id and the value is the shared app name.
+
+    e.g. {
+        "org_mozilla_firefox": "fenix",
+        "org_mozilla_firefox_beta": "fenix",
+        "org_mozilla_ios_firefox": "firefox_ios",
+        "org_mozilla_ios_firefoxbeta": "firefox_ios",
+    }
+    """
+    response = requests.get(GLEAN_APP_LISTINGS_URL)
+    response.raise_for_status()
+
+    app_listings = response.json()
+
+    return {
+        app["bq_dataset_family"]: app["app_name"]
+        for app in app_listings
+        if "bq_dataset_family" in app and "app_name" in app
+    }
--- a/bigquery_etl/copy_deduplicate.py
+++ b/bigquery_etl/copy_deduplicate.py
@@ -20,13 +20,16 @@ import click
 from google.api_core.exceptions import BadRequest
 from google.cloud import bigquery
 
-from bigquery_etl.cli.utils import table_matches_patterns
+from bigquery_etl.cli.utils import (
+    get_glean_app_id_to_app_name_mapping,
+    parallelism_option,
+    project_id_option,
+    table_matches_patterns,
+)
 from bigquery_etl.config import ConfigLoader
 from bigquery_etl.util.bigquery_id import sql_table_id
 from bigquery_etl.util.client_queue import ClientQueue
 from bigquery_etl.util.common import TempDatasetReference
-
-from .cli.utils import parallelism_option, project_id_option
 
 QUERY_TEMPLATE = """
 WITH
@@ -97,6 +100,13 @@ def _has_field_path(schema: List[bigquery.SchemaField], path: List[str]) -> bool
 def _select_geo(live_table: str, client: bigquery.Client) -> str:
     """Build a SELECT REPLACE clause that NULLs metadata.geo.* if applicable."""
     _, dataset_id, table_id = live_table.split(".")
+    channel_to_app_name = get_glean_app_id_to_app_name_mapping()
+    app_id = re.sub("_live$", "", dataset_id)
+
+    excluded_apps = set(ConfigLoader.get("geo_deprecation", "skip_apps", fallback=[]))
+    app_name = channel_to_app_name.get(app_id)
+    if app_name in excluded_apps:
+        return ""
 
     excluded_tables = set(
         ConfigLoader.get("geo_deprecation", "skip_tables", fallback=[])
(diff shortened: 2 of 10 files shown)


kwindau (approved):

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -215,6 +218,148 @@ def validate_shredder_mitigation(query_dir, metadata):
...
 
 
+def count_schema_fields(schema):
+    """Returns the count of fields and how many have descriptions in a given BigQuery schema."""
+    if not schema or not isinstance(schema, (list, tuple)):
+        return (0, 0)
+
+    fields = descriptions = 0
+    base = list(schema)
+    while base:
```
kik-kik, line 229: Maybe I'm missing something here, but base is an iterator and we iterate though each element once. It looks like this could be a bit more concise if we used a for loop instead:
```python
fields = len(base)

for field in base:
   if getattr[...]

return [...]
```

kik-kik: ok, I see. Later you are modifying the object we are iterating over here.

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -215,6 +218,148 @@ def validate_shredder_mitigation(query_dir, metadata):
...
+        4: "bigeye_predefined_metrics",
+        5: "change_control",
+        6: "column_descriptions_all",
+        7: "column_descriptions_70_percent"
+    }
+
+    level_requirements = {
+        "gold": [1, 2, 3, 4, 5, 6],
+        "silver": [1, 2, 3, 4, 7],
+        "bronze": [1]
```
kik-kik: Does everything need to fall in the bronze category at the minimum or do we plan to allow datasets to be defined in the repository without this label?

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -215,6 +218,148 @@ def validate_shredder_mitigation(query_dir, metadata):
...
+    }
+    results = {}
+    missing = []
+
+    if not metadata.level:
+        return True
+    else:
+        # Check that the level is unique and not empty.
+        if len(metadata.level) != 1:
```
kik-kik: We should consider the level value validation inside the metadata class:
https://github.com/mozilla/bigquery-etl/blob/a27d6b02ed50c8315671cac974fee3e4d32744d5/bigquery_etl/metadata/parse_metadata.py#L178-L201

bigquery_etl/metadata/parse_metadata.py
```diff
@@ -198,6 +198,7 @@ class Metadata:
     deletion_date: Optional[date] = attr.ib(None)
     monitoring: Optional[MonitoringMetadata] = attr.ib(None)
     require_column_descriptions: bool = attr.ib(False)
+    level: bool = attr.ib(False)
```
kik-kik: Is False a valid value for level? Or should it be None to indicate a lack of assigned value?

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -215,6 +218,148 @@ def validate_shredder_mitigation(query_dir, metadata):
...
+            if rx.search(obj):
+                path_found = True
+
+    missing = predefined_metrics - found_types
+
+    click.echo({
+        "path_found": path_found,
+        "found_types": sorted(found_types),
+        "missing_types": sorted(missing),
+    })
```
akkomar: Do we want this or it's an accidental debug log?

bigquery_etl/metadata/parse_metadata.py
akkomar: This looks like a copy-paste typo from a line above. We probably want this to be a string.

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -215,6 +218,148 @@ def validate_shredder_mitigation(query_dir, metadata):
...
+
+    if not metadata.level:
+        return True
+    else:
+        # Check that the level is unique and not empty.
+        if len(metadata.level) != 1:
+            click.echo(f"❌ERROR. Invalid level in metadata, A table or view can have max. one of {[*level_requirements]} level assigned.")
+            return False
+
+        level = metadata.level[0] if isinstance(metadata.level, list) else metadata.level
```
akkomar: Why allow setting level to list of strings instead of a single value?

bigquery_etl/metadata/parse_metadata.py
sean-rose: IMO an enum would make most sense.

bigquery_etl/metadata/validate_metadata.py
lucia-vargas-a: We plan to allow for no label, the unittest would be [test_level_is_not_set](https://github.com/mozilla/bigquery-etl/blob/a27d6b02ed50c8315671cac974fee3e4d32744d5/tests/cli/test_cli_metadata.py#L460)

bigquery_etl/metadata/validate_metadata.py
lucia-vargas-a: It's an attempt to print as much detail as possible for developers, but I think the print below does the work by telling what is missing, so I'll remove this one.

bigquery_etl/metadata/validate_metadata.py
lucia-vargas-a: Updated the docstrings and output to clarify that no level label is also an option.

bigquery_etl/metadata/parse_metadata.py
lucia-vargas-a: It should be None, corrected.

bigquery_etl/metadata/validate_metadata.py
lucia-vargas-a: Adding flexibility for a single element string e.g. ["gold"], I've noticed is a common syntax in the metadata, although it might give the wrong idea of allowing multiple values. Currently, the CI check would allow a string or a single-element list, also string.  @akkomar should we remove that flexibility and allow only string?

akkomar: Yes, I'd suggest removing the list flexibility and only accepting a string. Accepting ["gold"] but rejecting ["gold", "silver"] is inconsistent - it gives the impression that multiple levels might be allowed, only to fail validation when
  someone tries it.

bigquery_etl/metadata/parse_metadata.py
```diff
@@ -363,6 +411,9 @@ def from_file(cls, metadata_file):
                         "require_column_descriptions"
                     ]
 
+                if "level" in metadata:
+                    level = metadata["level"]
```
akkomar, line 394: I think if you follow [Sean's suggestion](https://github.com/mozilla/bigquery-etl/pull/8522#discussion_r2577751329) you could convert to `AssetLevel` here:
```suggestion
                    level = AssetLevel(metadata["level"])
```
This way invalid values fail at parse time and you don't need a separate validator above.

lucia-vargas-a: Yes this is tidy, though I'd personally prefer the detailed debugging messages in the validator, in case of CI failures.

bigquery_etl/metadata/validate_metadata.py
```diff
@@ -215,6 +218,148 @@ def validate_shredder_mitigation(query_dir, metadata):
...
+
+    missing = predefined_metrics - found_types
+
+    if any(metric in missing for metric in predefined_metrics):
+        click.echo(f"ERROR: Missing Bigeye metrics: {missing}.")
+        return False
+    return True
+
+
+def validate_asset_level(query_dir, metadata):
```
wwyc, line 274: how about missing owners?

lucia-vargas-a: In my opinion, it makes sense to require owners, at the minimum for level Gold.
Could you add it to [Requirements for each level](https://docs.google.com/document/d/1K8f4Jm0pRJOTbQpUs4KALnPmUSWg_mmJ1L_1484yEeg/edit?tab=t.0#heading=h.pu9y4swccxnc) for which levels you think it makes sense?

PR #8524, opened 2025-12-01
fix: newtab_merino_propensity_v1 schema validation error

kik-kik:
# fix: newtab_merino_propensity_v1 schema validation error

Also making sure when the value of `section_position` is set to null we cast it to INTEGER to match the schema.


rolf-moz:
thanks for fixing this!

Issue #8530, opened 2025-12-01
Support stage deploys for syndicated views


scholtzan:
If a syndicated view is referenced in an artifact that gets deployed to stage, then the stage deploys fail since there is no local query folder for the syndicated view. We could simply pull in the schemas via dryruns and create the artifacts locally that are needed for stage deploys

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DENG-10279)

PR #8532, opened 2025-12-01
Support deploying syndicated views to stage

scholtzan:
## Description

Fixes https://github.com/mozilla/bigquery-etl/issues/8530

If any artifacts reference a syndicated views then stage deploys are currently failing with:
```
Unrecognized name: submission_timestamp
```
Stage deploys try to determine the schemas of referenced tables via dryruns if they don't have a locally defined schema.
The syndicated views aren't partitioned on `submission_timestamp` (unlike stable tables), so we need to make sure that the partitioning on `submission_timestamp` is only used when the dependency is a stable table.


sean-rose:
Could we instead avoid trying to deploy syndicated views to stage at all (including not rewriting references to syndicated views)?

scholtzan:
> Could we instead avoid trying to deploy syndicated views to stage at all (including not rewriting references to syndicated views)?

Yeah. I made the change and tested it on `sql/moz-fx-data-shared-prod/app_store/firefox_app_store_territory_source_type_report` which references a syndicated view. https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/55684/workflows/ebf8426f-5fec-4dff-adb1-330223e6e719/jobs/713840

bigquery_etl/cli/stage.py
```diff
@@ -306,26 +307,47 @@ def _view_dependencies(artifact_files, sql_dir):
...
-                        schema = Schema.for_table(
-                            project=project,
-                            dataset=dataset,
-                            table=name,
-                            id_token=id_token,
-                            partitioned_by=partitioned_by,
+                # check if dataset is used for syndication
+                # if so, skip creating schema and dependencies unless a local file exists
+                dataset_metadata_path = (
+                    Path(sql_dir) / project / dataset / "dataset_metadata.yaml"
```
sean-rose: **suggestion (non-blocking):** If you import `DATASET_METADATA_FILE` from the `parse_metadata` module then this could be
```suggestion
                    Path(sql_dir) / project / dataset / DATASET_METADATA_FILE
```

bigquery_etl/cli/stage.py
```diff
@@ -306,26 +307,47 @@ def _view_dependencies(artifact_files, sql_dir):
...
-                        schema.to_yaml_file(path / SCHEMA_FILE)
-
-                if not file_exists_for_dependency:
-                    (path / QUERY_SCRIPT).write_text(
-                        "# Table stub generated by stage deploy"
-                    )
-                    view_dependencies.add(path / QUERY_SCRIPT)
+                        is_syndicated = dataset_metadata.dataset_base_acl == "syndicate"
+                    except Exception:
+                        pass
```
sean-rose: **question (non-blocking):** Since you're already checking whether the `dataset_metadata.yaml` exists first, what other exceptions are you anticipating could occur?

Personally, I'd be inclined to either not use `try:` for this, or print a warning when an exception occurs and is being ignored.

bigquery_etl/cli/stage.py
```diff
@@ -306,26 +307,47 @@ def _view_dependencies(artifact_files, sql_dir):
...
+
+                # Only create stubs if not syndicated OR if a local file already exists
+                if not is_syndicated or file_exists_for_dependency:
```
sean-rose: **question (blocking):** Are you certain this logic is correct?  The only thing that comes to mind is maybe you were trying to account for non-syndicate content in syndicate datasets, but even in that scenario it doesn't make sense to me that the logic in this block should run when files already exist for the dependency.

One additional complication is it looks like the logic that sets `file_exists_for_dependency` assumes the dependency's project is the same as the referring view's project, which seems like a bug (none of the other logic in this method makes that assumption).

bigquery_etl/cli/stage.py
```diff
@@ -306,26 +307,47 @@ def _view_dependencies(artifact_files, sql_dir):
...
+                    path = Path(sql_dir) / project / dataset / name
+                    if not path.exists():
+                        path.mkdir(parents=True, exist_ok=True)
+                        # don't create schema for wildcard and metadata tables
+                        if "*" not in name and name != "INFORMATION_SCHEMA":
+                            partitioned_by = None
+
+                            if dataset.endswith("_stable"):
+                                partitioned_by = "submission_timestamp"
```
sean-rose: **suggestion (blocking):** As some views reference live tables, I'd think we'd also need to account for those:
```suggestion
                            if any(dataset.endswith(suffix) for suffix in ("_live", "_stable")):
                                partitioned_by = "submission_timestamp"
```

Though IMO it would make more sense to also avoid trying to deploy referenced live/stable tables to stage.  And if we did that, are there any other scenarios where we'd actually need to create these stub table schemas for deploy to stage?

bigquery_etl/cli/stage.py
```diff
@@ -306,26 +307,47 @@ def _view_dependencies(artifact_files, sql_dir):
...
+                                id_token=id_token,
+                                partitioned_by=partitioned_by,
+                            )
+                            schema.to_yaml_file(path / SCHEMA_FILE)
+
+                    if not file_exists_for_dependency:
+                        (path / QUERY_SCRIPT).write_text(
+                            "# Table stub generated by stage deploy"
+                        )
+                        view_dependencies.add(path / QUERY_SCRIPT)
```
sean-rose, line 338: **suggestion (non-blocking):** To me it would make sense to only create the `query.py` stub files if the `schema.yaml` file was actually created (i.e. move this logic into the same block that writes the `schema.yaml` file), because a `query.py` table can't be deployed without a `schema.yaml`.

sean-rose (commented):
Thanks for being willing to change the approach!  Though given that change, the PR title should probably be updated.

bigquery_etl/cli/stage.py
sean-rose: And if files do already exist for tables referenced by the view then it seems like those should be added to the set of view dependencies, but that doesn't seem like it's being done in either the existing logic or this updated logic.

sean-rose:
> Yeah. I made the change and tested it on `sql/moz-fx-data-shared-prod/app_store/firefox_app_store_territory_source_type_report` which references a syndicated view. https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/55684/workflows/ebf8426f-5fec-4dff-adb1-330223e6e719/jobs/713840

It doesn't seem like that CI job actually published the `app_store.firefox_app_store_territory_source_type_report` view to stage though?

bigquery_etl/cli/stage.py
scholtzan: I updated the logic. My intention was to handle cases where datasets have syndication configured but also have other content/tables.

bigquery_etl/cli/stage.py
scholtzan: Yeah, we can handle live and stable tables like the syndicated views here and not publish them to stage

bqetl_project.yaml
```diff
@@ -42,7 +42,6 @@ dry_run:
   - sql/moz-fx-data-shared-prod/monitoring_derived/metadata_completeness_v1/query.sql
   - sql/moz-fx-data-shared-prod/firefox_desktop_derived/desktop_installs_v1/query.sql
   - sql/moz-fx-data-experiments/monitoring/query_cost_v1/query.sql
-  - sql/moz-fx-data-shared-prod/app_store/**/*.sql
```
scholtzan, line 45: just for testing. will remove once successfully deployed

scholtzan:
Ah, the service account used to deploy artifacts to the staging project doesn't have access to shared-prod. I believe we decided not to grant read access as there is a risk that production data might get leaked.
I think since this is the case I might go back to my initial proposed solution of getting the schemas for the syndicate datasets and deploying those: https://github.com/mozilla/bigquery-etl/pull/8532/commits/0ebe6259df65dca22e35a87098fb06843ff60e48

bigquery_etl/cli/stage.py
```diff
@@ -311,7 +311,11 @@ def _view_dependencies(artifact_files, sql_dir):
                     path.mkdir(parents=True, exist_ok=True)
                     # don't create schema for wildcard and metadata tables
```
sean-rose, line 312: **suggestion (non-blocking):** As evidenced by some of our confusion during this PR, I think it'd be helpful to have a comment explaining why this approach is needed.

bigquery_etl/cli/stage.py
```diff
@@ -311,7 +311,11 @@ def _view_dependencies(artifact_files, sql_dir):
...
                     if "*" not in name and name != "INFORMATION_SCHEMA":
-                        partitioned_by = "submission_timestamp"
+                        partitioned_by = None
+
+                        if dataset.endswith("_stable"):
```
sean-rose: **suggestion (blocking):** As some views reference live tables, I'd think we'd also need to account for those:
```suggestion
                        if any(dataset.endswith(suffix) for suffix in ("_live", "_stable")):
```

bigquery_etl/cli/stage.py
```diff
@@ -311,7 +311,11 @@ def _view_dependencies(artifact_files, sql_dir):
                     path.mkdir(parents=True, exist_ok=True)
                     # don't create schema for wildcard and metadata tables
                     if "*" not in name and name != "INFORMATION_SCHEMA":
```
sean-rose, line 317: **issue (non-blocking):** This doesn't create a `schema.yaml` for wildcard tables and `INFORMATION_SCHEMA` views, but it does still create a directory for them, and the `if not file_exists_for_dependency:` block below would also create a `query.py` stub for them (IMO it only makes sense to create `query.py` if `schema.yaml` is also being created).

And then there's also the issue of the logic that sets `file_exists_for_dependency` assuming the dependency's project is the same as the referring view's project, which seems like a bug.

I'm not marking these issues as blocking since they aren't directly related to this PR's purpose, and maybe you'd prefer we address them in a separate PR.

scholtzan: I created a ticket for the `file_exists_for_dependency` issue: https://github.com/mozilla/bigquery-etl/issues/8545

PR #8938, opened 2026-03-02
Add Claude automated code review workflow

alekhyamoz:
This adds an action to run automated code reviews on PRs using Claude.
Triggers:
  - Once automatically when a PR is opened or marked ready for review
  - On demand by commenting "@claude" on a PR


.github/workflows/claude-review.yml
```diff
@@ -0,0 +1,141 @@
...
+          track_progress: true
+          use_sticky_comment: true
+          include_fix_links: false
+
+          # Custom prompt for BigQuery ETL review
+          prompt: |
+            REPO: ${{ github.repository }}
+            PR NUMBER: ${{ github.event.pull_request.number || github.event.issue.number }}
+
+            You are reviewing code for Mozilla's BigQuery ETL infrastructure. This repository contains:
```
akkomar: I think we should start with something minimal, observe, and add instructions here only if Claude is missing something. This seems too long and at least some information here is redundant (e.g. "This repository contains:" is already captured in AGENTS.md, or "read the project instructions in AGENTS.md" is not needed, Claude reads this automatically).

akkomar (changes requested):

.github/workflows/claude-review.yml
alekhyamoz: noted. I will test with fork and keep you posted.

.github/workflows/claude-review.yml
```diff
@@ -0,0 +1,50 @@
...
+
+jobs:
+  review:
+    if: |
+      (
+        github.event_name == 'pull_request' &&
+        github.event.pull_request.draft == false &&
+        github.event.pull_request.head.repo.full_name == github.repository
+      ) || (
+        github.event_name == 'issue_comment' &&
```
alekhyamoz, line 22: @akkomar  the pattern that build job is gated by is 
- pull_request → only for non-fork PRs
- pull_request_target → only for fork PRs (runs with secrets but from a trusted ref)

For issue_comment, the fork status isn't available in the if expression,the solution is to add an early fork-check step that short-circuits the job for fork PRs

.github/workflows/claude-review.yml
```diff
@@ -0,0 +1,50 @@
...
+        contains(github.event.comment.body, '@claude')
+      )
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+      id-token: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
```
akkomar:
```suggestion
        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
```

.github/workflows/claude-review.yml
```diff
@@ -0,0 +1,50 @@
...
+        with:
+          fetch-depth: 1
+
+      - name: Run Claude Code Review
+        uses: anthropics/claude-code-action@v1
```
akkomar:
```suggestion
        uses: anthropics/claude-code-action@df37d2f0760a4b5683a6e617c9325bc1a36443f6  # v1.0.75
```

alekhyamoz:
@akkomar made changes addressing to key points
1. workflow is gated the same way as build job
2. Scan it with zizmor
3. if we're allowed to enable this without approvals

.github/workflows/claude-review.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Checkout repository
+        if: steps.fork-check.outputs.is_fork != 'true'
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          fetch-depth: 1
+          persist-credentials: false
+
+      - name: Run Claude Code Review
+        if: steps.fork-check.outputs.is_fork != 'true'
+        uses: anthropics/claude-code-action@df37d2f0760a4b5683a6e617c9325bc1a36443f6  # v1.0.75
```
akkomar: Let's bump to the latest version:
```suggestion
        uses: anthropics/claude-code-action@fefa07e9c665b7320f08c3b525980457f22f58aa  # v1.0.111
```

akkomar (approved):
LGTM, I think we can try this out in this form first without running on forked PRs.
This is also good to be merged per https://bugzilla.mozilla.org/show_bug.cgi?id=2020431.
I'd like @scholtzan to take a look before this lands.

Test failures seem not related, rebase/update should fix them.

scholtzan (approved):

PR #9168, opened 2026-04-07
add files for new_profile_churn_clients_v1

Marlene-M-Hirose:
## Description

This PR adds file for `glean_telemetry_derived.new_profile_churn_clients_v1`
* DENG-10676
* DSRE-XXXX


sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,622 @@
...
+    DATE_DIFF(max_date, first_seen_date, DAY) AS days_data_available,
+    days_since_last_data,
+    MAX(
+      IF(
+        submission_date = DATE_ADD(first_seen_date, INTERVAL 0 DAY)
+        AND mozfun.bits28.days_since_seen(days_desktop_active_bits) = 0,
+        "active",
+        NULL
+      )
+    ) AS day_0,
```
BryanBaird, line 151: Instead of just "active" vs NULL, I think the three possible values should be Boolean TRUE/FALSE/NULL.

NULL makes sense when the date is actually just not available and it's impossible for any client to have a known value, but otherwise Bools are probably more efficient and easier to use downstream.

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,622 @@
+WITH activity_range AS (
+  SELECT
+    client_id,
+    first_seen_date,
+    submission_date,
+    days_desktop_active_bits
+  FROM
+    `moz-fx-data-shared-prod.firefox_desktop.baseline_clients_last_seen`
```
BryanBaird, line 8: Is there a preference for `baseline_clients_last_seen` over `baseline_active_users`? I'm not sure if all the other client info columns are the same across the tables, but it might be easier to use the `is_dau` field already available in the active users table than recalculating from `days_desktop_active_bits` each time.

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,622 @@
...
+      IF(
+        submission_date = DATE_ADD(first_seen_date, INTERVAL 0 DAY)
+        AND mozfun.bits28.days_since_seen(days_desktop_active_bits) = 0,
+        "active",
+        NULL
+      )
+    ) AS day_0,
+    MAX(
+      IF(
+        submission_date = DATE_ADD(first_seen_date, INTERVAL 1 DAY)
```
BryanBaird, line 154: Could we compute these more easily upstream with a general `DATE_DIFF(submission_date, first_seen_date, interval day) AS days_since_first_seen` in one of the first CTE steps?

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,622 @@
...
+      AND day_22 IS NULL
+      AND day_23 IS NULL
+      AND day_24 IS NULL
+      AND day_25 IS NULL
+      AND day_26 IS NULL
+      AND day_27 IS NULL
+      AND day_28 IS NULL
+      AND day_29 IS NULL
+      AND day_30 IS NULL
+      THEN "immediately_churned"
```
BryanBaird, line 585: I think this needs an ELSE clause for when they _do_ have at least one day of activity in the initial 28 day window.

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,622 @@
...
+      AND day_26 IS NULL
+      AND day_27 IS NULL
+      AND day_28 IS NULL
+      AND day_29 IS NULL
+      THEN "churned"
+  END AS churned_after_2_days,
+  -- Client is in the cohort but never active across the entire 30-day window
+  -- NULL if day 30 data not yet available
+  CASE
+    WHEN days_data_available < 30
```
BryanBaird, line 552: This one should be able to calculate with 28 days available, not the full 30. (Even if they're active right after the first 28 days, we'd still call them initially churned.)

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/schema.yaml
```diff
@@ -0,0 +1,308 @@
...
+  mode: NULLABLE
+  description: User churned after two days
+- name: immediately_churned
+  type: STRING
+  mode: NULLABLE
+  description: User was not active in the 30 days since first seen. Immediately churned
+- name: day_0
+  type: STRING
+  mode: NULLABLE
+  description: Active first day - day 0
```
BryanBaird:
```suggestion
  description: Whether the client qualified as DAU on their first seen date (Day 0).
```

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,622 @@
...
+      AND day_5 IS NULL
+      AND day_6 IS NULL
+      THEN "churn_risk"
+    WHEN (day_0 = "active" OR day_1 = "active")
+      AND day_2 IS NULL
+      AND day_3 IS NULL
+      AND day_4 IS NULL
+      AND day_5 IS NULL
+      AND day_6 IS NULL
+      THEN "churn_risk"
```
BryanBaird, line 474: I think this can be simplified and condensed -- it shouldn't need two separate WHEN clauses, just:
* Day 2 is FALSE
* Day 3 is FALSE
* Day 4 is FALSE
* Day 5 is FALSE
* Day 6 is FALSE

The Day 0 and Day 1 values are logically irrelevant in the definition that was set up.

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,622 @@
...
+  first_seen_attribution_source,
+  first_seen_attribution_term,
+  first_seen_distribution_name,
+  is_desktop,
+  policies_is_enterprise,
+  -- No activity on days 0-4; at risk of churning by day 5
+  -- Clients active on day 0 or day 1 are excluded (they become churn_risk_after_7_days instead)
+  -- NULL if day 4 data not yet available
+  CASE
+    WHEN days_data_available < 4
```
BryanBaird:
```suggestion
    WHEN days_data_available < 5
```

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,622 @@
...
+      THEN NULL
+    WHEN day_0 IS NULL
+      AND day_1 IS NULL
+      AND day_2 IS NULL
+      AND day_3 IS NULL
+      AND day_4 IS NULL
+      THEN "churn_risk"
+  END AS churn_risk_after_5_days,
```
BryanBaird: Similar to comment below on churn risk after 7 days case, I think this can be simplified and combined a bit because Day 0 activity is ignored for this metric definition. Combining with some of the other comments on data type and False vs NULL output, I think you get something like:

```suggestion
    WHEN day_1 IS FALSE
      AND day_2 IS FALSE
      AND day_3 IS FALSE
      AND day_4 IS FALSE
      THEN TRUE
     ELSE FALSE
  END AS churn_risk_after_5_days,
```

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,622 @@
...
+        NULL
+      )
+    ) AS day_2,
+    MAX(
+      IF(
+        submission_date = DATE_ADD(first_seen_date, INTERVAL 3 DAY)
+        AND mozfun.bits28.days_since_seen(days_desktop_active_bits) = 0,
+        "active",
+        NULL
+      )
```
BryanBaird, line 174: How much does it improve performance to calculate all of the days as separate columns like this? I'm sure it has some parallelization benefits over the other draft approach that uses several self-joins, but I wonder if those are offset by not being able to use bitwise operations the denser `days_{desktop_}active_bits` fields from fewer days.

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,622 @@
...
+    first_seen.distribution.name AS first_seen_distribution_name,
+    IF(
+      LOWER(IFNULL(last_seen.isp, '')) <> 'browserstack'
+      AND LOWER(
+        IFNULL(COALESCE(last_seen.distribution_id, distribution_mapping.distribution_id), '')
+      ) <> 'mozillaonline',
+      TRUE,
+      FALSE
+    ) AS is_desktop,
+    last_seen.policies_is_enterprise,
```
BryanBaird, line 123: This wasn't in the spec previously, but if we are going to persist with the client-level version of the table as well as the aggregates, it's probably worth including `sample_id` and partitioning by that to support the querying best practices we use on other client-level tables.

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
Marlene-M-Hirose: `baseline_active_users` is built from `baseline_clients_last_seen` combined with two other tables. if we want all of the columns from `baseline_active_users`, it is easy to switch out. 

That being said:
If we use baseline_active_users in activity_range: possible and slightly cleaner — is_dau = TRUE replaces the bits check, and the table is smaller (only active days, not all "recently seen" days). Marginal cost reduction.

baseline_clients_last_seen in client_attributes: can't be replaced entirely because it has more columns — baseline_active_users doesn't have attribution, distribution, OS, or the days_desktop_active_bits needed for activity_segment.

Bottom line: We could swap activity_range to use baseline_active_users, but since the expensive part is the client_attributes joins, the overall cost savings would be small.

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
Marlene-M-Hirose: It isn't only about performance, it's also about cost. The joins have to shuffle across slots and if we are doing multiple self joins as in the draft approach, that will end up being more costly than the single join - each LEFT JOIN activity_range AS day1, day4, day6, etc. triggers its own shuffle. Since activity_range isn't materialized, BigQuery re-scans and re-shuffles the underlying table for each of the 6 joins — so we would be paying that shuffle cost 6 times on what is ultimately the same data. 

So you would have more dense bitwise operations, but it would be more costly if we do the self joins.

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
Marlene-M-Hirose: refactored to do this

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
Marlene-M-Hirose: done

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
Marlene-M-Hirose: refactored

Marlene-M-Hirose: Is this NULL or FALSE? I still have NULL.

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
Marlene-M-Hirose: refactored

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
BryanBaird: That all makes sense to me. There's definitely a subjective question about what seems simpler and more maintainable, given that neither source table has everything we want out of the box, so if there isn't a concrete cost benefit one way or the other I'm more than happy to just stick with this approach.

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,714 @@
...
+  -- NULL if day 6 data not yet available
+  CASE
+    WHEN days_data_available < 6
+      THEN NULL
+    WHEN day_2 = FALSE
+      AND day_3 = FALSE
+      AND day_4 = FALSE
+      AND day_5 = FALSE
+      AND day_6 = FALSE
+      THEN "churn_risk"
```
BryanBaird:
```suggestion
      THEN TRUE
      ELSE FALSE
```

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,714 @@
...
+      AND day_20 = FALSE
+      AND day_21 = FALSE
+      AND day_22 = FALSE
+      AND day_23 = FALSE
+      AND day_24 = FALSE
+      AND day_25 = FALSE
+      AND day_26 = FALSE
+      AND day_27 = FALSE
+      AND day_28 = FALSE
+      THEN "churned"
```
BryanBaird:
```suggestion
      THEN TRUE
      ELSE FALSE
```

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,714 @@
...
+      AND day_21 = FALSE
+      AND day_22 = FALSE
+      AND day_23 = FALSE
+      AND day_24 = FALSE
+      AND day_25 = FALSE
+      AND day_26 = FALSE
+      AND day_27 = FALSE
+      AND day_28 = FALSE
+      AND day_29 = FALSE
+      THEN "churned"
```
BryanBaird:
```suggestion
      THEN TRUE
      ELSE FALSE
```

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,714 @@
...
+      AND day_22 = FALSE
+      AND day_23 = FALSE
+      AND day_24 = FALSE
+      AND day_25 = FALSE
+      AND day_26 = FALSE
+      AND day_27 = FALSE
+      AND day_28 = FALSE
+      AND day_29 = FALSE
+      AND day_30 = FALSE
+      THEN "immediately_churned"
```
BryanBaird:
```suggestion
      THEN TRUE
      ELSE FALSE
```

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,714 @@
...
+      OR day_23 = TRUE
+      OR day_24 = TRUE
+      OR day_25 = TRUE
+      OR day_26 = TRUE
+      OR day_27 = TRUE
+      OR day_28 = TRUE
+      OR day_29 = TRUE
+      OR day_30 = TRUE
+      THEN "not_immediately_churned"
+    ELSE NULL
```
BryanBaird:
```suggestion
    
```

I think this can all be cut if we mimic the approach for the other metrics.

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,714 @@
...
+      AND day_21 = FALSE
+      AND day_22 = FALSE
+      AND day_23 = FALSE
+      AND day_24 = FALSE
+      AND day_25 = FALSE
+      AND day_26 = FALSE
+      AND day_27 = FALSE
+      AND day_28 = FALSE
+      AND day_29 = FALSE
+      AND day_30 = FALSE
```
BryanBaird:
```suggestion
```
"Immediate churn" only depends on first 28 days.

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,714 @@
...
+      AND day_3 = FALSE
+      AND day_4 = FALSE
+      AND day_5 = FALSE
+      AND day_6 = FALSE
+      THEN "churn_risk"
+  END AS churn_risk_after_7_days,
+  -- Active on day 0, then no activity for the next 28 days
+  -- NULL if day 28 data not yet available
+  CASE
+    WHEN days_data_available < 28
```
BryanBaird:
```suggestion
    WHEN days_data_available < 29
```
I think we need 29 days (Day 0 active + 28 days inactive).

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,714 @@
...
+      AND day_28 = FALSE
+      THEN "churned"
+  END AS churned_after_1_day,
+  -- Active on days 0 and 1, then no activity for the next 28 days
+  -- NULL if day 29 data not yet available
+  CASE
+    WHEN days_data_available < 29
+      THEN NULL
+    WHEN day_0 = TRUE
+      AND day_1 = TRUE
```
BryanBaird:
```suggestion
  -- Active on days 0 and/or 1, then no activity for the next 28 days
  -- NULL if day 29 data not yet available
  CASE
    WHEN days_data_available < 30
      THEN NULL
    WHEN (day_0 = TRUE
      OR day_1 = TRUE)
```
Subtle tweaks to business logic here as well.

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/schema.yaml
```diff
@@ -0,0 +1,308 @@
...
+  type: BOOLEAN
+  mode: NULLABLE
+  description: Indicates if the client is included in the desktop KPI.
+- name: policies_is_enterprise
+  type: BOOLEAN
+  mode: NULLABLE
+  description: Attempt to determine if the user is an enterprise user based on various
+    signals.
+- name: churn_risk_after_5_days
+  type: STRING
```
BryanBaird: With the refactoring, these will all be Booleans instead of Strings, so we'll want to make sure to keep the column documentation up-to-date as well.

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/schema.yaml
```diff
@@ -0,0 +1,308 @@
...
+  mode: NULLABLE
+  description: User has not been active in the first five days. Potential churn risk
```
BryanBaird:
```suggestion
  description: Is the user a high risk of already churning, based on their inactivity levels in the five days after they are first seen? 
```

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/schema.yaml
```diff
@@ -0,0 +1,308 @@
...
+- name: churn_risk_after_7_days
+  type: STRING
+  mode: NULLABLE
+  description: User has not been active in the first seven days. Potential churn risk
```
BryanBaird:
```suggestion
  description: Is the user a high risk of already churning, based on their inactivity levels in the seven days after they are first seen?
```

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/schema.yaml
```diff
@@ -0,0 +1,308 @@
...
+- name: churned_after_1_day
+  type: STRING
+  mode: NULLABLE
+  description: User churned after one day
```
BryanBaird:
```suggestion
  description: Shows if the client was active on their first seen date, but then immediately churned with no DAU activity for the following 28 days.
```

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/schema.yaml
```diff
@@ -0,0 +1,308 @@
...
+- name: churned_after_2_days
+  type: STRING
+  mode: NULLABLE
+  description: User churned after two days
```
BryanBaird:
```suggestion
  description: Shows if the client was active at some point within the 48 hours after they are first seen, but then immediately churned with no DAU activity for the following 28 days.
```

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/schema.yaml
```diff
@@ -0,0 +1,308 @@
...
+- name: immediately_churned
+  type: STRING
+  mode: NULLABLE
+  description: User was not active in the 30 days since first seen. Immediately churned
```
BryanBaird:
```suggestion
  description: User was not active in any of the the 28 days from the day they are first seen (i.e., qualified as "churned" immediately).
```

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/metadata.yaml
```diff
@@ -0,0 +1,21 @@
+friendly_name: New Profile Churn Clients
+description: |-
+  Please provide a description for the query
```
BryanBaird: Probably good to note up-front that this implementation is Desktop only. That's consistent with the business requirements, but helpful to mitigate confusion for future consumers!

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,714 @@
...
+  attribution_term,
+  distribution_name,
+  first_seen_attribution_campaign,
+  first_seen_attribution_content,
+  first_seen_attribution_medium,
+  first_seen_attribution_source,
+  first_seen_attribution_term,
+  first_seen_distribution_name,
+  is_desktop,
+  policies_is_enterprise,
```
BryanBaird, line 513: In the temp table, it looks like about 30% of rows are NULL for all of these values, and I'm not sure why. 🤔 It doesn't seem to be the case in `moz-fx-data-shared-prod.firefox_desktop.baseline_clients_last_seen`, but I don't know where the join/CTE logic might be introducing that many missing values.

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
BryanBaird: Oof, yeah, the shuffling and compute costs are definitely parts I have no real intuition about. Definitely a good thing to optimize for!

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/metadata.yaml
Marlene-M-Hirose: I will add that to the description. And I'll add Glean to the friendly name/description

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
Marlene-M-Hirose: I noticed that. I'll do a deeper investigation next since we have the business logic mostly figured out.

Marlene-M-Hirose: Ok the problem is that in the test table, we had cohort_date going back to December 2, 2025. However, I limited the attribution data to January 1, 2026. Therefore, the join would result in `missing` data. I'm building a temporary table so that should go away. I'll be sure to cover that when I put this table into production

Marlene-M-Hirose:
@BryanBaird , just so you know, the code as it stands process more than 2 TB of data. When this is run for one single day, the code will process over 1TB of data. For your purposes, I will restrict the amount of data by putting a filter on the sample_id. Adding a filter of `sample_id <= 10` drops the amount of data processed to ~190 GB.

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,682 @@
...
+      AND day_28 = FALSE
+      THEN TRUE
+    ELSE FALSE
+  END AS churned_after_1_day,
+  -- Active on days 0 and/or 1, then no activity for the next 28 days
+  -- NULL if day 29 data not yet available
+  CASE
+    WHEN days_data_available < 30
+      THEN NULL
+    WHEN (day_0 = TRUE OR day_1 = TRUE)
```
BryanBaird:
```suggestion
  -- Active on Day 1, then no activity for the next 28 days
  -- NULL if day 29 data not yet available
  CASE
    WHEN days_data_available < 30
      THEN NULL
    WHEN (day_1 = TRUE)
```

Sorry, I got the business logic slightly wrong with my previous recommendation -- day 0 can be true or false, but day 1 needs to be TRUE.

A client that was active on Day 0 but then _not_ on Day 1 would be churned_after_1_day, and the two categories should be mutually exclusive. They will get combined into the `cumulative_` versions of the metrics in the aggregation stage.

sql/moz-fx-data-shared-prod/glean_telemetry_derived/new_profile_churn_clients_v1/query.sql
```diff
@@ -0,0 +1,682 @@
...
+  FROM
+    activity_range
+  CROSS JOIN
+    max_available_date
+  GROUP BY
+    client_id,
+    sample_id,
+    first_seen_date,
+    max_date,
+    days_since_last_data
```
BryanBaird, line 470: Something in the aggregating logic means that the `day_x` values are only populated for the latest available data -- every other day is `null`, which breaks the downstream business logic.

E.g. in the temp table, `cohort_date = '2026-02-15` shows a lot of rows where `day_14` is the only non-null date column. So all `churn_risk` values are FALSE (they use `IS FALSE` in their logic) and all `churned_` columns are NULL (they use `= FALSE` in their logic), because none of them can be satisfied with only a value on that one date.

Marlene-M-Hirose: Ah... what changes need to be made then?

Ok, looks like something went wrong with the test data. I've created a new test table with `sample_id <= 10` where this case isn't happening anymore.

BryanBaird (approved):
Based on the adjustments to the business logic and our back-and-forth testing with the sample data, I can say this Looks Good To Me from the perspective of the business requirements being satisfied!

There is an identified and known discrepancy between how Glean and Legacy telemetry have different counts for Desktop DAU, especially on the first seen date. But that is an upstream issue beyond the scope of this specific task.

I can't speak to whether there are other Data Engineering cost/efficiency considerations, but in terms of getting the desired output I can call this **approved** ✅.

PR #9199, opened 2026-04-13
DO NOT MERGE: DENG-10944 Test Pipeline for Generating Metadata Descriptions 

gkabbz:
DO NOT MERGE: DRAFT ONLY

* DENG-XXXX
* DSRE-XXXX


script/metadata/field_profiler.py
```diff
@@ -0,0 +1,615 @@
...
+            " low-cardinality columns store the full value/frequency list."
+        ),
+    ),
+    bigquery.SchemaField(
+        "example_value",
+        "STRING",
+        mode="NULLABLE",
+        description=(
+            "A single representative value sampled from the column. Only populated for high-cardinality columns."
+            " PII-like values (emails, UUIDs, etc.) are stored as-is here but should not be surfaced in descriptions."
```
chelseyklein, line 94: @gkabbz is that the behavior we want? i ran this on `braze_derived` restricted dataset and got pii in the form of user emails in this column - luckily was in a sandbox vs nonprod

i'm finding this difficult to mitigate. if you think this column is necessary, i think it should be in a restricted dataset that few people have access to

we could add a function that mitigates this some by preventing a scan of probable pii columns...but may need legal review?

gkabbz: I agree. I'd chatted with @akkomar around how we want to handle it. For this table to be useful for the data classification work, we do need to see where we have PII. So I think the best solution here is to restrict this in a dataset where only DE and DPE at most have access to it specifically for the classification side of things. The other thing I haven't thought about in this job is how the updates would happen. Assuming we want to schedule this and keep it fresh the most recent run is the most important. I don't think we need to have history. So at most we wouldn't have a large number of PII in here.

akkomar: Even DE+DPE is too broad IMO. Not only because of PII, but restricted datasets.
Another aspect we need to consider is that some classes of data should not be sent to external LLM providers.

PR #9205, opened 2026-04-14
feat(DENG-): update attribution_clients.query.sql template to include the new play_store_attribution ping

kik-kik:
# feat: update attribution_clients.query.sql template to include the new play_store_attribution ping

## Description

play_store_attribution attribution is no longer send using first-session ping, instead now this information is shared using the new play-store-attribution ping. This change updated the attribution_clients query template to account for this. Once updated and deployed we should be able to re-run the pipeline for affected dates to ensure play store attribution information is correctly propagated.

### DIFF
--- a/sql_generators/mobile_kpi_support_metrics/mobile_kpi_support_metrics.py
+++ b/sql_generators/mobile_kpi_support_metrics/mobile_kpi_support_metrics.py
@@ -44,9 +44,10 @@ BIGEYE_NOTIFICATION_SLACK_CHANNEL = "#de-bigeye-triage"
 class AttributionPings(Enum):
     """An enumerator containing a list of pings that could be the source of attribution information."""
 
-    metrics: str = "metrics"
-    first_session: str = "first_session"
-    baseline: str = "baseline"
+    metrics = "metrics"
+    first_session = "first_session"
+    baseline = "baseline"
+    play_store_attribution = "play_store_attribution"
 
 
 @dataclass
@@ -106,7 +107,10 @@ class AttributionFields:
     )
     play_store = AttributionFieldGroup(
         name="play_store",
-        source_pings=[AttributionPings.first_session],
+        source_pings=[
+            AttributionPings.first_session,
+            AttributionPings.play_store_attribution,
+        ],
         fields=[
             {
                 "name": "play_store_attribution_campaign",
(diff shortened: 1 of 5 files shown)


sql_generators/mobile_kpi_support_metrics/mobile_kpi_support_metrics.py
```diff
@@ -44,9 +44,10 @@
 class AttributionPings(Enum):
     """An enumerator containing a list of pings that could be the source of attribution information."""
 
-    metrics: str = "metrics"
-    first_session: str = "first_session"
-    baseline: str = "baseline"
+    metrics = "metrics"
```
kik-kik, line 47: This is to fix mypy errors.

sql_generators/mobile_kpi_support_metrics/mobile_kpi_support_metrics.py
```diff
@@ -295,7 +299,7 @@ def generate_mobile_kpi_support_metrics(target_project: str, output_dir: click.P
     The parent folders will be created if not existing and existing files will be overwritten.
     """
     env = Environment(loader=FileSystemLoader(str(GENERATOR_ROOT / "templates")))
-    output_dir = Path(output_dir) / target_project
+    project_output_dir = Path(str(output_dir)) / target_project
```
kik-kik, line 302: Fixing mypy errors.

Marlene-M-Hirose (approved):

sql_generators/mobile_kpi_support_metrics/templates/attribution_clients.query.sql
```diff
@@ -286,7 +348,7 @@ SELECT
   COALESCE(first_session_ping.adjust_info, metrics_ping.adjust_info) AS adjust_info,
   {% endif %}
   {% if 'play_store' in product_attribution_group_names %}
-  first_session_ping.play_store_info,
+  COALESCE(play_store_attribution_ping.play_store_info, first_session_ping.play_store_info) AS play_store_info,
```
scholtzan, line 354: Just making sure this is the correct order of operation here:
As I understand first session ping doesn't sent this information anymore, do we want to fall back to `play_store_attribution_ping.play_store_info` in cases where `first_session_ping.play_store_info` doesn't exist? E.g. if it's still available in historical tables should this still use `first_session_ping`? If so those two values would need to be switched.

But if `play_store_attribution_ping` should always have precedence, then this is fine.

kik-kik: what I did instead is I added a date filter in the play_store_attribution_ping_base CTE. I like `play_store_attribution_ping` being at the front of the COALESCE as to me it feels like it clearly communicates that now this is the preferred source of information.

PR #9211, opened 2026-04-15
Optimize Merino New Tab Interactions Query

rolf-moz:
## Description
There are three fixes in here:

With the rollout of additional sections and propensity based query calculations, the cost of the merino extract query has gone up over time. 96% of events worldwide are in the newtab_content ping. 

Therefore we can drop the newtab ping and rely on the newtab_content ping exclusively

Background is in this doc https://docs.google.com/document/d/1ByJJ3IdKRhiOT4wjjdYTbs8P4MNHfWq6T9bTVIBW1zo/edit?tab=t.0

This image shows per country adoption. US over 95%, with FR, IT and IN in the 86% percent range, but it should be improving.
image

A manual run shows 70% reduction in slot time.

2.
Our propensity https://github.com/mozilla/bigquery-etl/pull/9154 upgrade includes an 'any' field that shows the propensity weight at a position independent of tile format. Adding 'any' fallback allows better propensity calculations when we have new formats such as the new Daily Briefing format.

3. This job sometimes gets delayed at midnight EST or UTC. I was told a while ago that we could boost the priority and set it to 2. Someone should review this.

### DIFF
--- a/sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_extract_to_gcs_v3/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_extract_to_gcs_v3/metadata.yaml
@@ -10,6 +10,7 @@ labels:
   owner1: cbeck
 scheduling:
   dag_name: bqetl_merino_newtab_extract_to_gcs
+  priority: 2
   arguments:
   - --source-project=moz-fx-data-shared-prod
   - --source-dataset=telemetry_derived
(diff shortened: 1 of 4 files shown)


mmiermans (approved):
Nice job figuring out the possible source of the daily delays! I'm not sure that I'm the right person to sign off on `priority: 2`, but it would be great if that fixed this problem:

Image

The rest of the PR looks good to me.

BenWu (approved):

PR #9212, opened 2026-04-15
Add event_counts_glean_v2 table

BenWu:
## Description

Creating a `event_counts_glean_v2` because the v1 table is hard to use and I want to rename and add some columns.  Differences:
- add ping type dimension
- add app id and name because the normalized_app_name is hard to remember
- Add event_extras_length as an approximation of the size of the event extras


sql_generators/glean_usage/templates/event_counts_glean_v2.query.sql
```diff
@@ -0,0 +1,64 @@
...
+        app_name,
+        normalized_app_name,
+        ping_type,
+        event_category,
+        event_name,
+        normalized_channel,
+        normalized_country_code,
+        app_version
+    )
+  {% if not outer_loop.last -%}
```
akkomar:
```suggestion
  {% if not (outer_loop.last and loop.last) %}
```
Otherwise this would break if the last app has >1 event tables I think.

akkomar (approved):

PR #9214, opened 2026-04-15
DENG-10886: Metadata completeness global.yaml update

gkabbz:
## Description
As part of the metadata completeness project, this PR makes updates to the global.yaml for fields identified via data profiling as appearing across multiple datasets without global descriptions or with inaccurate ones. 

- Adds: branch, app_version_patch, and is_bot_generated as new entries;
- Adds build_id as an alias of app_build_id (confirmed same value via query lineage);
- Updates os_version description with accurate cross-platform version examples.## Related Tickets & Documents

**IMPORTANT:** This is a critical change from a definition accuracy perspective. As the next step in the metadata project would be to cascade these definitions to a large number of tables. 

Relevant Tickets
* DENG-10886


bigquery_etl/schema/global.yaml
```diff
@@ -36,12 +38,20 @@ fields:
...
+  type: INTEGER
+  description: The patch version of the user visible app version string for the browser,
+    e.g. "142.1.3" has patch version 3
 - name: app_name
   type: STRING
   description: The name of the browser.
 - name: architecture
   type: STRING
   description: The architecture of the device, (e.g. "arm", "x86").
+- name: branch
```
kik-kik, line 51: thought: the phrase / word `branch` seems like a very generic word and feels like something that could have multiple definitions depending on context. For example, if we have a Github table with open or merged PRs we may also end up with a field called `branch`. So I'm not sure if this specific term should live in a global schema definition.

kik-kik (approved):
Changes look fine, I did leave one thought on one of the fields though. Please take a look at it before merging.

bigquery_etl/schema/global.yaml
```diff
@@ -36,12 +38,20 @@ fields:
   type: INTEGER
   description: The minor version of the user visible app version string for the browser,
     e.g. "142.1.3" has minor version 1
+- name: app_version_patch
+  type: INTEGER
+  description: The patch version of the user visible app version string for the browser,
+    e.g. "142.1.3" has patch version 3
 - name: app_name
   type: STRING
   description: The name of the browser.
```
lucia-vargas-a, line 47: E.g. VPN,  or should we limit to telemetry data?
```suggestion
  description: The name of the browser or application.
```

bigquery_etl/schema/global.yaml
```diff
@@ -36,12 +38,20 @@ fields:
...
 - name: architecture
   type: STRING
   description: The architecture of the device, (e.g. "arm", "x86").
+- name: branch
+  type: STRING
+  description: The Nimbus experiment branch to which the client was assigned
+    (e.g., "control", "treatment-a", "treatment-b").
 - name: attribution_campaign
   type: STRING
   description: The attribution campaign; similar or the same as UTM campaign.
```
lucia-vargas-a, line 57: We could take the [attribution definitions from firefox_android_clients](https://github.com/mozilla/bigquery-etl/blob/db844c037cf4adc3c4211f12de79259ee54459fe/sql/moz-fx-data-shared-prod/fenix_derived/firefox_android_clients_v1/schema.yaml#L84-L98), I collected those from the [adjust docs](https://help.adjust.com/en/article/links#campaign-structure-parameters).
```suggestion
  description: The utm_content this install is attributed to. Reported by the install referrer service, not Adjust.
```

bigquery_etl/schema/global.yaml
```diff
@@ -152,8 +166,9 @@ fields:
   description: The operating system code based on the Common Platform Enumeration CPE.
 - name: os_version
   type: INTEGER
-  description: Version of the operating system version running at the
-    client. E.g. "100.9.11".
+  description: The normalized version of the operating system running at the
+    client (e.g., "10.0" for Windows, "24.6.0" for macOS, "18.6.2" for iOS,
+    "16" for Android).
   aliases:
```
lucia-vargas-a, line 172: I think we should separate os from normalized_os because it's imprece to say they're aliases since one is normalized and the other isn't. Then the definition would be ok for normalized_os.

There is one problem tho in that os_version exists as a non-normalized raw data point and also exists in  `active_users_aggregates` (AUA) as os_version , simply because it's a shorter easy-to-query name.
So we should just enmsure we don't update that schema.
Same happens with fields that have a raw version and a normalized version e.g. os, channel.

lucia-vargas-a (approved):

gkatre:
## Re-Review: PR #9214 — Applied Base Schema Decision Tree Analysis

I re-reviewed this PR using the base-schema-audit decision tree (applied in order):
1. **Product namespace** — product-specific prefix/feature? → `app_<product>.yaml`
2. **Cross-dataset breadth** — appears in 2+ distinct datasets and is generic? → `global.yaml`
3. **Single-dataset generic** — one dataset only? → `<dataset_name>.yaml`
4. **No schema evidence** — defer / remove

---

### `branch` → `global.yaml` ✓ confirmed, but description needs scoping

Verified across 6 distinct datasets (`fenix_derived`, `firefox_desktop_derived`, `firefox_ios_derived`, `org_mozilla_fenix_derived`, `org_mozilla_ios_firefox_derived`, `telemetry_derived`) — clearly passes the Step 2 cross-dataset threshold, so `global.yaml` is the right file.

One structural nuance: in the majority of tables, `branch` is **nested inside an experiment struct**, not a top-level field. Worth confirming the cascade tooling handles nested field promotion correctly, otherwise this entry won't reach most of its actual usages.

Agree with kik-kik's concern about generality. Suggested description:

```yaml
description: In Nimbus experimentation contexts, the branch to which the client
  was enrolled (e.g., "control", "treatment-a", "treatment-b"). Do not apply this
  description to tables where "branch" refers to a different concept (e.g., source
  control or release branches).
```

---

### `app_version_patch` → should NOT be in `global.yaml` — two problems

Only found in **1 table** (`telemetry_derived/fx_accounts_active_daily_clients_v1`). The decision tree routes single-dataset fields to `telemetry_derived.yaml`, not `global.yaml`.

Additionally there is a **type mismatch**: the actual `schema.yaml` for that table has `type: NUMERIC`, but the PR proposes `type: INTEGER`. Cascading with `INTEGER` would conflict with the real column type.

**Recommendation:** remove from this PR. Add to `telemetry_derived.yaml` with `type: NUMERIC` once more table evidence exists to justify promotion to `global.yaml`.

---

### `is_bot_generated` → no schema evidence, remove from PR

Found in **zero** `schema.yaml` files across the entire repo. The decision tree cannot classify a field with no data presence — adding it to `global.yaml` is premature and it will never cascade anywhere useful until it exists in actual table schemas.

**Recommendation:** remove from this PR and re-introduce once it appears in at least one real table schema.

---

### `build_id` alias for `app_build_id` → placement correct, but verify scope

`build_id` appears in 2 datasets (`telemetry_derived`, `firefox_desktop_derived`) — passes Step 2. However, several of those tables are installer/funnel tables where `build_id` is described as _"Build ID of the installed product. Absent for a failed stub installation"_ — close to `app_build_id` but not identical. The alias is reasonable given the author confirmed equivalence via query lineage, but that confirmation should explicitly cover installer tables too.

---

### `os_version` / `normalized_os_version` alias → remove the alias

Both fields are genuinely global (`os_version` in 18 datasets, `normalized_os_version` in 8), but they are **not the same field**. In `active_users_aggregates`, `os_version` is the raw version string (`"100.9.11"`), while `normalized_os_version` is the pipeline-processed variant. The alias would cause the cascade to overwrite raw `os_version` fields with a "normalized" description — exactly the risk @lucia-vargas-a flagged.

**Recommendation:** remove the `normalized_os_version` alias from the `os_version` entry. Add `normalized_os_version` as its own separate top-level entry in `global.yaml` using the updated cross-platform description from this PR.

---

### `app_name` description

Agree with @lucia-vargas-a — broadening to _"The name of the browser or application"_ is correct and should be accepted.

---

### Summary

| Field | Decision Tree | Right File? | Action |
|---|---|---|---|
| `branch` | Step 2 → `global.yaml` | ✅ Yes | Tighten description; confirm nested cascade works |
| `app_version_patch` | Step 3 → `telemetry_derived.yaml` | ❌ No | Remove; fix type to `NUMERIC`; move to `telemetry_derived.yaml` |
| `is_bot_generated` | No schema evidence | ❌ No | Remove until field exists in actual table schemas |
| `build_id` alias | Step 2 → `global.yaml` | ✅ Yes | Verify installer table equivalence |
| `os_version` alias | Semantic mismatch | ❌ No | Remove alias; add `normalized_os_version` as its own entry |
| `app_name` description | n/a (existing field) | — | Accept @lucia-vargas-a suggestion |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

bigquery_etl/schema/global.yaml
gkabbz: I'll make this change and resubmit for review.

bigquery_etl/schema/global.yaml
gkabbz: I hear you. But unfortunately I don't think we can change the name here as that name is part of the nimbus data collection. Given its usage - I'm seeing it used over 2000 times I think it makes sense for it to be in the global file. For github type branch, the design of the metadata completeness introduces a dataset.yaml file which takes precedence over the global.yaml so for the github branch definitions, those would be defined there for reuse.

bigquery_etl/schema/global.yaml
gkabbz: I think you meant to refer to campaign but said content. But either way, I've updated the description to specifically call out it's what is in the `utm_campaign` field in marketing campaigns. I didn't include the reference to the referrer service, be it install referrer or adjust as the `utm_campaign` is used pretty consistently for marketing campaigns even though the mechanism of how it gets reported might be different for both mobile and desktop. This way the definition works more globally i.e. attribution_campaign == the field where utm_campaign ends up living.

gkabbz:
Hey @gkatre. I'm going to ignore the claude review since I resolved the branch comment from Kik and made the changes Lucia suggested. The other changes recommended regarding the `is_bot` and `app_version_patch` are not in scope as this PR isn't touching those fields.

lucia-vargas-a (approved):

PR #9216, opened 2026-04-16
feat(etl): Update subscription deletion ETL scripts

david1alvarez:
Because:

* Current glean deletion ETL query duplicates the nimbus_user_id derivation logic that lives in the fxa repository

This commit:

* Updates the script to match the paired [PR](https://github.com/mozilla/fxa/pull/20381), while still supporting the existing approach

Closes #PAY-3438

### DIFF
--- a/sql/moz-fx-data-shared-prod/subscription_platform_backend_cirrus_derived/delete_events_v2/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/subscription_platform_backend_cirrus_derived/delete_events_v2/metadata.yaml
@@ -1,0 +1,18 @@
+---
+friendly_name: Subscription Platform Backend Cirrus Delete Events
+description: Deletion events for subscription platform backend cirrus, derived
+  from payments-api GKE logs containing pre-derived nimbus_user_id
+owners:
+  - dalvarez@mozilla.com
+labels:
+  incremental: true
+  schedule: daily
+  table_type: client_level
+scheduling:
+  dag_name: bqetl_fxa_events
+bigquery:
+  time_partitioning:
+    type: day
+    field: submission_timestamp
+    require_partition_filter: false
+  require_column_description: true
--- a/sql/moz-fx-data-shared-prod/subscription_platform_backend_cirrus_derived/delete_events_v2/schema.yaml
+++ b/sql/moz-fx-data-shared-prod/subscription_platform_backend_cirrus_derived/delete_events_v2/schema.yaml
@@ -1,0 +1,9 @@
+fields:
+- name: submission_timestamp
+  type: TIMESTAMP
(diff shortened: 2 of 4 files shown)


sql/moz-fx-data-shared-prod/subscription_platform_backend_cirrus_derived/delete_events_v2/metadata.yaml
```diff
@@ -0,0 +1,17 @@
...
+  incremental: true
+  schedule: daily
+  table_type: client_level
+scheduling:
+  dag_name: bqetl_fxa_events
+bigquery:
+  time_partitioning:
+    type: day
+    field: submission_timestamp
+    require_partition_filter: false
```
BenWu, line 17: The expectation is that new tables have column descriptions in the schema.yaml and this should be set
```suggestion
    require_partition_filter: false
require_column_descriptions: true
```
Creating tables via the cli will add this but can get missed when copy/pasting

sql/moz-fx-data-shared-prod/subscription_platform_backend_cirrus_derived/delete_events_v2/schema.yaml
```diff
@@ -0,0 +1,7 @@
+fields:
+- name: submission_timestamp
+  type: TIMESTAMP
+  mode: NULLABLE
```
BenWu, line 4: A description is useful here because the meaning of `submission_timestamp` here doesn't match the meaning we're used to having everywhere else (logged event vs. telemetry ping).  I know the `delete_events_v1` table doesn't have this but this would be an improvement
```suggestion
  mode: NULLABLE
  description: ...
```

BenWu (commented):
Just some comments on column descriptions but othwerwise LGTM

sql/moz-fx-data-shared-prod/subscription_platform_backend_cirrus_derived/delete_events_v2/metadata.yaml
david1alvarez: Updated to add `require_column_description: true`

sql/moz-fx-data-shared-prod/subscription_platform_backend_cirrus_derived/delete_events_v2/schema.yaml
david1alvarez: Updated to add `description: Earliest timestamp of the account deletion log entry for this nimbus_user_id on the given date` and `description: Pre-derived Nimbus user identifier logged by payments-api at account deletion time`

BenWu (approved):
I'll merge this when CI finishes

PR #9221, opened 2026-04-17
Update ads_client_operations_and_errors_hourly_v1 to use live tables and add is_init

curtismorales:
## Description

Tweak `ads_client_operations_and_errors_hourly_v1` query to use live tables and do an initial backfill.

The amount of data that will be processed in the init is quite small, so I think it's fine to backfill this way.


chelseyklein (approved):
haven't tested but logic looks right 🙌

PR #9229, opened 2026-04-20
feat(nimbus): Use metric-config-parser for feature monitoring configs

### LINKED ISSUE #15387: Move configuration files to experimenter github repo
feature monitoring configs like [https://github.com/mozilla/bigquery-etl/blob/.../sql_generators/nimbus_feature_monitoring/templates/firefox_desktop.yaml](https://github.com/mozilla/bigquery-etl/blob/5550aeed344dd635d0de26f5f62d1d69d9a5e36d/sql_generators/nimbus_feature_monitoring/templates/firefox_desktop.yaml) should live in the metric hub repo. The config format should be changed from YAML to TOML, and the source table property is_event_stream should be changed to type so that we explicitly support specific table types, e.g. metrics (currently defined by is_event_stream: false) event_stream, and clients daily (not specifically supported yet). The bigquery-etl generator python should then use metric-config-parser to get the configs.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6732)

yashikakhurana:
Relates to EXP-6732

## Description

- Replace yaml + tomllib with NimbusFeatureMonitoringSpec from metric-config-parser
- Change is_events_stream bool to type string (event_stream / metrics / clients_daily)
- Add --config-path CLI option to point generator at metric-hub checkout
- Configs now live in metric-hub/nimbus_feature_monitoring/ as TOML files

* EXP-6732
* closes https://github.com/mozilla/experimenter/issues/15387  


yashikakhurana:
blocked on this https://github.com/mozilla/metric-hub/pull/1396

yashikakhurana:
blocked on this one https://github.com/mozilla/metric-hub/pull/1406

sql/moz-fx-data-shared-prod/firefox_desktop/urlbar_events_daily/view.sql
mikewilli, line 7: Is this related?

sql_generators/nimbus_feature_monitoring/__init__.py
```diff
@@ -202,12 +200,11 @@ def generate_queries(project, path, write_dir):
...
             features.append(
                 Feature(
-                    name=feat_name,
-                    project=feat.pop("project", project),
-                    dataset=feat.pop("dataset", f"{dataset}_derived"),
-                    ratios=feat.pop("ratios", []),
+                    name=feat.nimbus_slug(),
+                    project=project,
+                    dataset=f"{dataset}_derived",
+                    ratios=[],
```
mikewilli: Should this be populated with something?

sql_generators/nimbus_feature_monitoring/templates/nimbus_feature_monitoring_v1/query.sql
```diff
@@ -67,7 +67,7 @@ rollouts_cte AS (
               {% endif %}
             )
           {% elif metric.data_type == "event" %}
-            {% if source_table.is_events_stream %}
+            {% if source_table.type == "event_stream" %}
```
mikewilli: Hm I didn't catch this on the metric-config-parser PR, but we should probably use `"events_stream"` here to match the table naming. Can we update this and also update on the config and config parser?

And while you're on the config parser, it might be good to define `type` as an enum as well.

sql_generators/nimbus_feature_monitoring/__init__.py
```diff
@@ -47,7 +46,7 @@ class SourceTable:
     dimensions: list[Dimension]
     analysis_unit_id: str = "client_info.client_id"
     time_partition_field: str = "submission_timestamp"
-    is_events_stream: bool = False
+    type: str = "metrics"
```
mikewilli: Do you think it makes sense to define a separate `SourceTable` vs what's in metric-config-parser? And if a new class is warranted, should it live here or in the parser side?

mikewilli (dismissed):
Left a few comments, but I think this also warrants some tests. You should be able to add them in a `tests/` dir at the same level as the `templates/` dir. I see some examples, like in `sql_generators/events_daily/`.

sql_generators/nimbus_feature_monitoring/__init__.py
yashikakhurana: The split is intentional — SourceTable can't move to metric-config-parser because:

project is the --target-project CLI arg, not something known at config parse time
Dimension computes SQL expressions (e.g. IF(field IN (...), field, default)) that are bigquery-etl-specific rendering logic
That said, there is duplication between SourceTable and SourceTableSpec (both declare table_name, type, analysis_unit_id). We could reduce that by having SourceTable hold a reference to the SourceTableSpec and only add the extra fields (project, dimensions: list[Dimension], last_dimensions, non_last_dimensions) rather than re-declaring everything. Happy to refactor if you'd prefer that approach.

sql_generators/nimbus_feature_monitoring/__init__.py
yashikakhurana: Added ratios: list[list[str]] = attr.Factory(list) to FeatureSpec in metric-hub and updated from_dict to read it.

yashikakhurana:
@mikewilli here are the changes https://github.com/mozilla/metric-hub/pull/1415

sql_generators/nimbus_feature_monitoring/__init__.py
mikewilli: Gotcha, I think you could define the class in metric-config-parser and instantiate it here if you wanted, but I'm fine with it living here if you prefer that (and I could be wrong or there might be other benefits to it living here).

And I don't have a strong preference on the approaches, but there is a pattern in metric-config-parser of having the *Spec inside a *Configuration class, like https://github.com/mozilla/metric-hub/blob/566b2edca603d086375cb4ff69895bfd0d65b41a/lib/metric-config-parser/metric_config_parser/experiment.py#L102-L111, so I'd probably lean that way.

yashikakhurana: I dont have preference, but let me know if you think that is better

mikewilli: Sure, let's follow the existing convention and go ahead with that.

yashikakhurana: okay done changes

tests/sql_generators/nimbus_feature_monitoring/test_generate_queries.py
mikewilli, line 1: Ok, this is my fault for steering you wrong, but I confirmed with others that these actually need to go in `tests/sql_generators/nimbus_feature_monitoring` to get run. Sorry about that!

sql_generators/nimbus_feature_monitoring/__init__.py
```diff
@@ -39,22 +39,40 @@ def __post_init__(self):
 
 
 @dataclass
-class SourceTable:
-    name: str
+class SourceTableConfiguration:
+    """Wraps a SourceTableSpec with rendering context needed to generate SQL."""
+
+    spec: SourceTableSpec
```
mikewilli: Maybe a more specific name?
```suggestion
    source_table_spec: SourceTableSpec
```

mikewilli (changes requested):
LGTM, except for my mistake on the test location.

mikewilli (approved):

PR #9232, opened 2026-04-21
Add support for ./bqetl target migrate-branch

scholtzan:
## Description

Adds `./bqetl --target <name> target migrate-branch <old_branch>` for moving an existing target deployment from a previous branch to the currently checked-out branch. This is useful after renaming a local git branch, so existing dev artifacts match the new branch.

- Renames datasets and/or artifacts (depending on where git.branch appears in the target's templates) by replacing the sanitized old-branch string with the new one.
- Tables are copied in place and the originals deleted. Views, materialized views, and routines are skipped during migration and recreated by a deploy step prompted at the end
- Rewrites references (FROM clauses, routine calls, manifests) in local SQL/YAML files under the target project dir, so downstream queries under the target path keep pointing at the migrated artifacts.
- When the target template includes `git.commit`, only the newest commit's datasets are migrated, and the commit hash is rewritten to current HEAD so the next deploy lands at the matching location instead of creating a parallel set of artifacts.

### DIFF
--- a/bigquery_etl/cli/target.py
+++ b/bigquery_etl/cli/target.py
@@ -19,6 +19,7 @@ from ..util.common import block_coding_agents
 from ..util.target import (
     IAM_ROLE_MAP,
     MANIFEST_FILENAME,
+    _get_git_context,
     _reapply_shared_access,
     render_artifact_prefix_pattern,
     render_dataset_pattern,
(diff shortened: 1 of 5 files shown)


bigquery_etl/cli/target.py
```diff
@@ -353,6 +359,367 @@ def share(
...
+
+    Useful after renaming a local git branch: rewrites the sanitized branch
+    string in dataset and/or artifact names so existing artifacts match the
+    current (new) branch without redeploying from scratch. Must be run
+    while checked out on the destination branch.
+
+    If the target template includes git.commit, only the newest commit's
+    datasets are migrated (older commit deployments are left behind) and
+    the commit hash is rewritten to current HEAD so the next deploy lands
+    at the matching location.
```
scholtzan, line 374: This made the most sense for me when I tested it out, but I'm open to changing that behaviour

bigquery_etl/cli/target.py
```diff
@@ -353,6 +359,365 @@ def share(
...
+    _rewrite_target_references(sql_dir, project_id, rewrite_content)
+    _prompt_and_deploy(ctx, client, project_id, plan, sql_dir, yes)
+
+
+def _make_transform(pairs):
+    """Build a str->str substring-replace transform from (old, new) pairs."""
+
+    def transform(s):
+        for old, new in pairs:
+            s = s.replace(old, new)
```
BenWu, line 491: Looks like this can replace all occurrences in the input which could cause incorrect replacements if someone uses a simple branch name.  Maybe using regex to look only for dataset/table names in qualified table ids would work.

I don't think this is a blocker because it's unlikely to occur and probably hard to solve, but I think it's a risk worth pointing out.

BenWu (approved):
Some smarter replacement should be considered but it's not a blocker

bigquery_etl/cli/target.py
scholtzan: I added a comment to the docs

PR #9242, opened 2026-04-22
[DENG-10973] Generate dag with copy_deduplicate task_markers

BenWu:
## Description

This generates a new dag that only has task markers for tasks directly downstream of copy_deduplicate.  copy_deduplicate will have a task marker on the tasks in this new dag.  This allows state changes in copy_deduplicate to automatically propagate down without needing to manually keep copy_deduplicate up to date.

Draft for the copy_deduplicate dag changes: https://github.com/mozilla/telemetry-airflow/pull/2380/changes

I did a small validation of this setup with https://github.com/mozilla/telemetry-airflow/commit/9903e4e076572a0c8151598826b6faa355d339ae

* DENG-10973

### DIFF
--- a/bigquery_etl/query_scheduling/copy_deduplicate_task_markers.py
+++ b/bigquery_etl/query_scheduling/copy_deduplicate_task_markers.py
@@ -1,8 +1,8 @@
 """Generate the copy_deduplicate_task_markers DAG.
 
 Generate an intermediate DAG between copy_deduplicate in the telemetry-airflow repo and
-its dependent DAGs.  This DAG contains `ExternalTaskMarker`s for the downstream tasks/DAGs so
-that setting state on the upstream copy_deduplicate tasks will propagate to downstream tasks.
+its dependent DAGs. This DAG contains `ExternalTaskMarker`s for the downstream tasks/DAGs so
+that clearing tasks in the upstream copy_deduplicate tasks will propagate to downstream tasks.
 This allows the task markers to stay updated without needing to change the code in telemetry-airflow.
 """
 
@@ -61,7 +61,11 @@ def build_markers_context(dag_collection) -> dict:
         for task in dag.tasks:
             deps = (task.upstream_dependencies or []) + (task.depends_on or [])
             for dep in deps:
+                # not copy_deduplicate
                 if dep.dag_name != SOURCE_DAG_NAME:
+                    continue
+                # already looked at task + dag
+                if dag.name in consumers_by_source[dep.task_id]:
                     continue
                 source_schedule = source_schedules.get(dep.task_id)
                 if source_schedule is None:
@@ -77,7 +81,7 @@ def build_markers_context(dag_collection) -> dict:
                         f"{dag.name} ({downstream_schedule!r})."
                     )
                 delta_seconds = _execution_delta_seconds(delta_string)
-                consumers_by_source[dep.task_id].setdefault(dag.name, delta_seconds)
+                consumers_by_source[dep.task_id][dag.name] = delta_seconds
 
     sources: List[tuple] = []
     for source_task_id in sorted(consumers_by_source):
@@ -98,9 +102,6 @@ def build_markers_context(dag_collection) -> dict:
         "schedule_interval": markers_schedule,
         "owner": TELEMETRY_ALERTS_EMAIL,
         "email": [TELEMETRY_ALERTS_EMAIL],
-        "start_date_year": 2019,
-        "start_date_month": 7,
-        "start_date_day": 25,
         "tags": [
             "impact/tier_1",
             "repo/bigquery-etl",
--- a/bigquery_etl/query_scheduling/copy_deduplicate_task_markers.py
+++ b/bigquery_etl/query_scheduling/copy_deduplicate_task_markers.py
@@ -1,0 +1,128 @@
+"""Generate the copy_deduplicate_task_markers DAG.
+
+Generate an intermediate DAG between copy_deduplicate in the telemetry-airflow repo and
+its dependent DAGs.  This DAG contains `ExternalTaskMarker`s for the downstream tasks/DAGs so
+that setting state on the upstream copy_deduplicate tasks will propagate to downstream tasks.
+This allows the task markers to stay updated without needing to change the code in telemetry-airflow.
+"""
+
+from collections import defaultdict
+from pathlib import Path
+from typing import Dict, List
+
+from black import FileMode, format_file_contents
+from jinja2 import Environment, PackageLoader
+
+from bigquery_etl.query_scheduling.task import EXTERNAL_TASKS
+from bigquery_etl.query_scheduling.utils import (
+    TELEMETRY_ALERTS_EMAIL,
+    schedule_interval_delta,
+)
+
+TASK_MARKERS_DAG_NAME = "copy_deduplicate_task_markers"
+TASK_MARKERS_TEMPLATE = "copy_deduplicate_task_markers.j2"
+SOURCE_DAG_NAME = "copy_deduplicate"
+
+
+def _execution_delta_seconds(delta_string: str) -> int:
+    """Parse a `<int>s` timedelta string into an int number of seconds."""
+    if delta_string is None:
+        return 0
+    if not delta_string.endswith("s"):
+        raise ValueError(
+            f"schedule_interval_delta should end with 's', got {delta_string!r}"
+        )
+    return int(delta_string[:-1])
+
+
+def build_markers_context(dag_collection) -> dict:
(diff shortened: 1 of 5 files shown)


bigquery_etl/query_scheduling/copy_deduplicate_task_markers.py
```diff
@@ -0,0 +1,128 @@
...
+        sources.append((source_task_id, consumers))
+
+    # Use copy_deduplicate's own schedule so execution_dates line up 1:1
+    markers_schedule = next(iter(source_schedules.values()))
+
+    return {
+        "name": TASK_MARKERS_DAG_NAME,
+        "source_dag_id": SOURCE_DAG_NAME,
+        "schedule_interval": markers_schedule,
+        "owner": TELEMETRY_ALERTS_EMAIL,
```
BenWu, line 103: I didn't set an explicit owner since this dag shouldn't fail unless copy_deduplicate also fails, but airflow weirdness does happen.  I'm open to setting myself or akomar who is the owner of copy_deduplicate for airflow triage assignment but I don't think it's necessary.

bigquery_etl/query_scheduling/templates/copy_deduplicate_task_markers.j2
```diff
@@ -0,0 +1,84 @@
...
+    )
+
+    with TaskGroup("{{ source_task_id }}_task_markers") as {{ source_task_id }}_task_markers:
+        {% for consumer in consumers %}
+        ExternalTaskMarker(
+            task_id="{{ consumer.dag_name }}__wait_for_{{ source_task_id }}",
+            external_dag_id="{{ consumer.dag_name }}",
+            external_task_id="wait_for_{{ source_task_id }}",
+            {% if consumer.execution_delta_seconds != 0 -%}
+            execution_date="{% raw %}{{{% endraw %} (logical_date + datetime.timedelta(seconds={{ consumer.execution_delta_seconds }})).isoformat() {% raw %}}}{% endraw %}",
```
BenWu, line 65: This uses timedelta instead of the `execution_date.replace` that copy_deduplicate currently uses because it can just reuse the value calculated value for the task sensor.  It should also be more flexible and work for sensors that are > ~24 hours

bigquery_etl/query_scheduling/templates/copy_deduplicate_task_markers.j2
```diff
@@ -0,0 +1,84 @@
...
+from airflow.utils.task_group import TaskGroup
+import datetime
+
+from utils.constants import ALLOWED_STATES, FAILED_STATES
+
+docs = """
+### {{ name }}
+
+Generated DAG that propagates `copy_deduplicate` state changes to every
+downstream bqetl DAG via `ExternalTaskMarker`s.
```
sean-rose: **quibble:** Both this and the similar wording in the `bigquery_etl/query_scheduling/copy_deduplicate_task_markers.py` docstring sound a bit misleading to me, as the only state changes this DAG will be involved in propagating is when the `copy_deduplicate` tasks are cleared recursively.  It'll have nothing to do with other state changes like success/failure.

bigquery_etl/query_scheduling/templates/copy_deduplicate_task_markers.j2
```diff
@@ -0,0 +1,84 @@
...
+
+tags = {{ tags }}
+
+with DAG(
+    "{{ name }}",
+    default_args=default_args,
+    schedule_interval="{{ schedule_interval }}",
+    doc_md=docs,
+    tags=tags,
+    catchup=False,
```
sean-rose: **suggestion (non-blocking):** I think catchup should ultimately be enabled for this DAG to match how `copy_deduplicate` is configured.  However, with this DAG's start date being set to 2019-07-25 to match `copy_deduplicate`'s start date, enabling catchup in this PR would result in nearly seven years of historical DAG runs getting triggered, which wouldn't be good (not least because the external task sensors for most of those historical DAG runs would get stuck waiting for DAG runs which are never going to happen again).

So I'd suggest either:
- Keeping the 2019 start date, and enable catchup in a follow-up PR after this DAG has had at least one DAG run.
- Changing the start date to be the current date and enabling catchup in this PR.

I'd personally lean toward the latter (also because the start date would then be more reflective of reality).

bigquery_etl/query_scheduling/templates/copy_deduplicate_task_markers.j2
```diff
@@ -0,0 +1,84 @@
...
+        external_dag_id="{{ source_dag_id }}",
+        external_task_id="{{ source_task_id }}",
+        execution_delta=datetime.timedelta(seconds=0),
+        check_existence=True,
+        mode="reschedule",
+        poke_interval=datetime.timedelta(minutes=5),
+        allowed_states=ALLOWED_STATES,
+        failed_states=FAILED_STATES,
+        pool="DATA_ENG_EXTERNALTASKSENSOR",
+    )
```
sean-rose: **note:** Using external task sensors in this DAG isn't technically necessary (e.g. `EmptyOperator` tasks would also work). The external task markers in `copy_deduplicate` just need a specific task to target (it's unfortunate they don't support targeting task groups).

However, I'm fine with using external task sensors here, as that is in keeping with the usual way these cross-DAG dependencies are set up (sensor pointing upstream, task marker pointing downstream at that sensor).

bigquery_etl/query_scheduling/templates/copy_deduplicate_task_markers.j2
```diff
@@ -0,0 +1,84 @@
...
+    with TaskGroup("{{ source_task_id }}_task_markers") as {{ source_task_id }}_task_markers:
+        {% for consumer in consumers %}
+        ExternalTaskMarker(
+            task_id="{{ consumer.dag_name }}__wait_for_{{ source_task_id }}",
+            external_dag_id="{{ consumer.dag_name }}",
+            external_task_id="wait_for_{{ source_task_id }}",
+            {% if consumer.execution_delta_seconds != 0 -%}
+            execution_date="{% raw %}{{{% endraw %} (logical_date + datetime.timedelta(seconds={{ consumer.execution_delta_seconds }})).isoformat() {% raw %}}}{% endraw %}",
+            {%- else -%}
+            execution_date="{% raw %}{{{% endraw %} logical_date.isoformat() {% raw %}}}{% endraw %}",
```
sean-rose: **note:** The `execution_date` parameter is optional and internally defaults to this exact same value, so this could be omitted.

bigquery_etl/query_scheduling/copy_deduplicate_task_markers.py
```diff
@@ -0,0 +1,128 @@
...
+                delta_string = schedule_interval_delta(
+                    source_schedule, downstream_schedule
+                )
+                if delta_string is None:
+                    raise ValueError(
+                        f"Cannot compute execution_delta between {SOURCE_DAG_NAME} ({source_schedule!r}) and "
+                        f"{dag.name} ({downstream_schedule!r})."
+                    )
+                delta_seconds = _execution_delta_seconds(delta_string)
+                consumers_by_source[dep.task_id].setdefault(dag.name, delta_seconds)
```
sean-rose: **question:** Is `setdefault()` being used to handle some particular edge case?

Otherwise, it seems like this could simply be
```suggestion
                consumers_by_source[dep.task_id][dag.name] = delta_seconds
```

Or an alternative way to get the same `setdefault()` effect of not overwriting an existing entry could be to add `or dag.name in consumers_by_source[dep.task_id]` to the `continue` conditions at the top of the loop.

sean-rose (approved):
r+wc

Thanks for implementing this!

bigquery_etl/query_scheduling/templates/copy_deduplicate_task_markers.j2
BenWu: I didn't consider that.  I think `EmptyOperator` would be better because there would be fewer waiting sensors and lower chance of things failing from transient issues.  It loses the "external dag" button in the extra links but that shouldn't matter because we shouldn't need to actually look at this dag.

bigquery_etl/query_scheduling/copy_deduplicate_task_markers.py
BenWu: Not an edge case, just making it explicit that there can be repeated keys and we only need one.  I'll change it to a `continue`.  I recall meaning to do that at some point but forgot

bigquery_etl/query_scheduling/templates/copy_deduplicate_task_markers.j2
BenWu: I'll change the start date

PR #9278, opened 2026-04-28
feat(DENG-10738): add new fields to metrics_clients_daily and metrics_clients_last_seen

kbammarito:
## Description

This PR adds the following fields to `metrics_clients_daily` and `metrics_clients_last_seen`:

- `is_large_device`
- `device_manufacturer`
- `isp_name`

* DENG-10738


wwyc (approved):

sql_generators/glean_usage/templates/metrics_clients_daily.query.sql
```diff
@@ -21,6 +21,12 @@
         {{ metrics[app_name][metric].sql }} AS {{ metric }},
       {% endfor -%}
     {% endif -%}
+    {% if app_name == "fenix" -%}
```
kik-kik, line 24: Maybe should add this to the very end of the query since in the schema update this field will be added to at the end. This might be important in case there are any unions or queries downstream that rely on column order.

kbammarito: It works now (with the new fields going to the end) because the rest of the query is `desktop` and nothing else

PR #9119, opened 2026-03-30
Structured data for SUMO

lucia-vargas-a:
## Description
Add retrieval table for Kitsune/SUMO with embedding, sentiment scor and LLM-generated summary, category, topics and entities.

## Related Tickets & Documents
* [DENG-10847](https://mozilla-hub.atlassian.net/browse/DENG-10847)

**Reviewer, please follow [this checklist](https://github.com/mozilla/bigquery-etl/blob/main/.github/reviewer_checklist.md)**

[DENG-10847]: https://mozilla-hub.atlassian.net/browse/DENG-10847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #9287, opened 2026-04-29
feat(dependency): Implement config for skipping dependency extraction

lucia-vargas-a:
## Description

sqlglot fails to parse queries using Vertex AI. This is [reported to sqlglot](https://github.com/tobymao/sqlglot/issues/7582) and will be deployed soon. For now -and maybe also worth having for the future-, it requires handling to unblock PRs using this feature e.g. [PR-9119)](https://github.com/mozilla/bigquery-etl/pull/9119).

* DENG-10847

### DIFF
--- a/bigquery_etl/dependency.py
+++ b/bigquery_etl/dependency.py
@@ -3,6 +3,7 @@
 import re
 import sys
 from concurrent.futures import ProcessPoolExecutor, as_completed
+from functools import cache
 from glob import glob
 from itertools import groupby
 from pathlib import Path
@@ -18,6 +19,29 @@ from bigquery_etl.schema.stable_table_schema import get_stable_table_schemas
 from bigquery_etl.util.common import alter_sql_for_sqlglot, render
 
 stable_views = None
+
+
+@cache
+def skip_dependency():
+    """Return a list of configured SQL files for which dependency parsing should be skipped."""
+    root = ConfigLoader.project_dir
+    result = []
+    for pattern in ConfigLoader.get("dependency", "skip", fallback=[]):
+        if any(c in pattern for c in ("*", "?", "[")):
+            # Wildcard pattern: expand via glob relative to repo root
+            result.extend(
+                str(Path(p).relative_to(root))
+                for p in glob(str(root / pattern), recursive=True)
+            )
+        else:
+            # Exact path: validate it is specific enough to avoid broad matching
+            if "/" not in pattern:
+                raise click.ClickException(
+                    f"dependency skip pattern '{pattern}' is too broad — "
+                    "must include a directory component (e.g. 'sql/project/dataset/table/query.sql')"
+                )
+            result.append(pattern)
+    return result
 
 
 def _raw_table_name(table: sqlglot.exp.Table) -> str:
@@ -71,6 +95,9 @@ def extract_table_references(sql: str) -> List[str]:
 
 def extract_table_references_without_views(path: Path) -> Iterator[str]:
     """Recursively search for non-view tables referenced in the given SQL file."""
+    if any(str(path).endswith(s) for s in skip_dependency()):
+        return
+
     # handle both global and local stable_views for thread/process safety
     global stable_views
     local_stable_views = stable_views
@@ -148,6 +175,8 @@ def _process_single_file(
     path: Path, without_views: bool = False
 ) -> Tuple[Path, List[str]]:
     """Process a single SQL file to extract table references."""
+    if any(str(path).endswith(s) for s in skip_dependency()):
+        return path, []
     try:
         if without_views:
             return path, list(extract_table_references_without_views(path))
--- a/bigquery_etl/dependency.py
+++ b/bigquery_etl/dependency.py
@@ -8,7 +8,7 @@ from glob import glob
 from itertools import groupby
 from pathlib import Path
 from subprocess import CalledProcessError
-from typing import Dict, Iterator, List, Tuple
+from typing import Dict, Iterator, List, Optional, Tuple
 
 import rich_click as click
 import sqlglot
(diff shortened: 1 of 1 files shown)


bigquery_etl/dependency.py
```diff
@@ -20,6 +21,29 @@
 stable_views = None
 
 
+@cache
+def skip_dependency():
+    """Return a list of configured SQL files for which dependency parsing should be skipped."""
```
sean-rose, line 26: **question:** Some dependency parsing is already skipped for ETLs that have `scheduling.referenced_tables` set in their `metadata.yaml` files.  What are the scenarios where that's insufficient for this new AI functions use case?

bigquery_etl/dependency.py
```diff
@@ -20,6 +21,29 @@
...
+    root = ConfigLoader.project_dir
+    result = []
+    for pattern in ConfigLoader.get("dependency", "skip", fallback=[]):
```
sean-rose, line 29: **quibble:** Since the title of this PR is "_Skip dependency parsing for queries using Vertex AI functions_" I'd rather expect it to be adding a `dependency.skip` section in `bqetl_project.yaml` to exclude specific queries using Vertex AI functions.  I'm guessing you're wanting to merge this PR first, then add such `dependency.skip` entries in #9119, which makes sense, but in that case I think the title of this PR should more accurately be something like "_feat(dependency): Implement config for skipping dependency extraction_".

bigquery_etl/dependency.py
sean-rose, line 62: **issue (blocking):** The skip dependency logic needs to be added in `extract_table_references()`.

And I think this is technically the only place it would absolutely have to be added, as both `extract_table_references_without_views()` and `_process_single_file()` end up calling `extract_table_references()` (but keeping the skip dependency logic and returning early in those functions is also still fine).

sean-rose (commented):
> sqlglot fails to parse queries using Vertex AI. ... For now -and maybe also worth having for the future-, it requires handling to unblock PRs using this feature e.g. [PR-9119)](https://github.com/mozilla/bigquery-etl/pull/9119).

It looks like the current CI issue with #9119 is that the Vertex AI API isn't enabled in the `moz-fx-data-integration-tests` project.  Where were you seeing the SQLGlot parsing failures happen?

bigquery_etl/dependency.py
lucia-vargas-a: Yes I'm adding it in the [PR where I introduce a query that uses Vertex functions](https://github.com/mozilla/bigquery-etl/pull/9119/changes#diff-7e34d2d38e27a920e060691809970185fc8f8b0bf3d3e7c7fb4792ed527c3c11), should I move it here?

lucia-vargas-a: Updated the title

sean-rose: > Yes I'm adding it in the [PR where I introduce a query that uses Vertex functions](https://github.com/mozilla/bigquery-etl/pull/9119/changes#diff-7e34d2d38e27a920e060691809970185fc8f8b0bf3d3e7c7fb4792ed527c3c11), should I move it here?

No, I think it makes sense to do it in that other PR since the ETLs in question don't currently exist.  It was just that the PR title implied otherwise, so thanks for updating that.

bigquery_etl/dependency.py
lucia-vargas-a: Makes sense, it could also be added to `_get_references` where is missing but that'd not be future proofed, so I moved to extract_table_references, thank you!

bigquery_etl/dependency.py
lucia-vargas-a: What I observe is that sqlglot thinks  AI in (AI.GENERATE and AI.EMBED) is a table and treats it as a dependency.

lucia-vargas-a:
Hey @sean-rose the [original error](https://github.com/mozilla/bigquery-etl/actions/runs/23917599811/job/69755478266?pr=9119#step:5:18) is not happening anymore. We could add this as a safeguard, but doesn't look like a blocker now. Let me know,

bigquery_etl/dependency.py
lucia-vargas-a: + added an optional path parameter to avoid modifying other calls to this function.

bigquery_etl/dependency.py
sean-rose: Interesting.  It looks like that was happening both for `bqetl format` (which there's already a `format.skip` config for) and `bqetl dependency record` (which is what this PR could help with).

It seems that SQLGlot added support for BigQuery's quirky AI function syntax in [v30.3.0](https://github.com/tobymao/sqlglot/blob/main/CHANGELOG.md#v3030---2026-04-07), but we're still using v28.5.0 (apparently because [we have too many open Dependabot PRs](https://mozilla.slack.com/archives/C01E8GDG80N/p1777578041033329)).

bigquery_etl/dependency.py
sean-rose: Oops, that was my bad.  I got so focused on the fact that `extract_table_references()` was the one place where SQLGlot was actually being used that I missed it didn't have access to the file path to filter on. 🤦 

So the way you had it before was appropriate.

sean-rose:
> Hey @sean-rose the [original error](https://github.com/mozilla/bigquery-etl/actions/runs/23917599811/job/69755478266?pr=9119#step:5:18) is not happening anymore. We could add this as a safeguard, but doesn't look like a blocker now. Let me know,

Yeah, I'd be inclined to not merge this since it isn't currently a blocker, and the underlying issue has been fixed in SQLGlot v30.3.0.

Also, entirely omitting dependencies isn't ideal, as those are used for various purposes.  I think a better solution to this kind of thing could be to have the existing `scheduling.referenced_tables` ETL metadata override settings also apply in these other use cases like `bqetl dependency record`.

sean-rose:
> I think a better solution to this kind of thing could be to have the existing `scheduling.referenced_tables` ETL metadata override settings also apply in these other use cases like `bqetl dependency record`.

I just remembered there's already a workaround for `bqetl dependency record`, which is to run it with the `--skip-existing` option and hard-code `references:` in `metadata.yaml` ([example](https://github.com/mozilla/bigquery-etl/blob/bef8ef089681df2bb365de39a1954dc5edcd2ff2/sql/moz-fx-data-shared-prod/org_mozilla_ios_focus_derived/baseline_clients_last_seen_v1/metadata.yaml#L42-L49)).  And in fact the `generate-sql` CI job does already run `bqetl dependency record` with the `--skip-existing` option ([code](https://github.com/mozilla/bigquery-etl/blob/bef8ef089681df2bb365de39a1954dc5edcd2ff2/.github/workflows/build.yml#L380-L382)).

lucia-vargas-a:
The problem that originated this is solved in SQLGlot v30.3.0.

PR #9303, opened 2026-05-04
[DENG-4042] Add backfill support for workgroup-restricted datasets

BenWu:
## Description

Allows managed backfills to work with non-mozilla-confidential datasets by copying the permissions from the prod dataset and table when creating the staging and backup tables.  Broad dataset permissions on `backfills_staging_derived` are replaced with data owner for the airflow service accounts.

I tested this in a sandbox project.  Also added two test tables that I plan on adding backfills for later. 

Should be merged with https://github.com/mozilla-services/cloudops-infra/pull/5694 which creates the backfill base ACL and https://github.com/mozilla/telemetry-airflow/pull/2381 which adds the new CLI arg.  This should also be merged when there aren't active backfills since the staging tables we would lose access to the staging tables.  We will lose read access to existing backup tables but I think that's acceptable.

* DENG-4042


sql/moz-fx-data-shared-prod/backfills_staging_derived/dataset_metadata.yaml
```diff
@@ -6,9 +6,10 @@ user_facing: false
 default_table_expiration_ms: "2592000000"  # 30 days
```
whd, line 7: Per jira discussion this dataset's workgroup access should match https://github.com/mozilla-services/cloudops-infra/pull/5694/changes i.e. drop `dataViewer` and `dataOwner` and set `dataset_base_acl` to `backfill`.

scholtzan (approved):

PR #9315, opened 2026-05-05
feat(fxci): add Azure worker costs

jwmossmoz:
## Description

Adds Azure VM cost rows to the existing `fxci_derived.worker_costs_v1` table.

Azure costs are folded into the same worker-cost table as GCP with a nullable `cloud_provider` field. The separate Azure worker cost table was removed after review feedback.

This PR also adds `moz-fx-data-shared-prod.azure_billing_syndicate` metadata so the Azure billing export can follow the normal bigquery-etl cross-project access pattern instead of reading `moz-fx-data-billing-prod-9147.azure_billing_raw` directly.

This PR no longer tries to add Azure task attribution. `task_run_costs_v1` remains scoped to the existing GCP worker metrics path for now. Azure task attribution should be a follow-up once replacement worker metrics are available from `docker-etl/jobs/fxci-taskcluster-export`; the older `taskclusteretl.worker_metrics` source is being decommissioned.

## Reviewer Notes

This follows the earlier FXCI cost PR shape from #5758, #6007, #6335, and #6374, but keeps the current change focused on the billing side.

Live data checks changed the implementation in a few places:

- The Azure billing tables available in BigQuery are `fxci_daily_actual_*` and `taskcluster_daily_actual_*`; `trusted_fxci_daily_actual_*` is not present.
- The load tables contain export snapshots, so the query joins through the matching `*_lookup` tables.
- Azure `ResourceId` casing varies, so VM matching and extraction are case-insensitive.
- New schema fields are appended at the end of the schema file to match BigQuery column-order behavior.

Access flow:

- This PR creates the bqetl-side syndicate dataset metadata for `moz-fx-data-shared-prod.azure_billing_syndicate`.
- [mozilla/global-platform-admin#6256](https://github.com/mozilla/global-platform-admin/pull/6256) adds the source-side `google_bigquery_syndicated_dataset` module for the existing `azure_billing_raw` dataset with `create_dataset = false`.
- Once both changes land and the Terraform PR is applied, this query should read Azure billing data through `moz-fx-data-shared-prod.azure_billing_syndicate.*`, matching the existing GCP billing syndication pattern.
- [mozilla/global-platform-admin#6235](https://github.com/mozilla/global-platform-admin/pull/6235) is a related Azure billing ETL change. It redeploys the Cloud Function that populates these tables, so the Terraform apply should wait until that PR's Atlantis plan is no longer active.
- [DENG-11127](https://mozilla-hub.atlassian.net/browse/DENG-11127) tracks possible Data Engineering help if the syndication or permissions work needs their assistance.

This PR is from a fork because I do not have push access to `mozilla/bigquery-etl`; integration CI may need the manual upstream trigger from the reviewer checklist.

## Validation

- `./bqetl format --check sql/moz-fx-data-shared-prod/fxci_derived/worker_costs_v1/query.sql sql/moz-fx-data-shared-prod/fxci_derived/task_run_costs_v1/query.sql`
- `./bqetl query validate --no-dryrun fxci_derived.worker_costs_v1`
- `./bqetl query validate --no-dryrun fxci_derived.task_run_costs_v1`
- `./bqetl metadata validate-workgroups sql/moz-fx-data-shared-prod/azure_billing_syndicate/dataset_metadata.yaml`
- `git diff --check`
- Checked against `docs/reference/recommended_practices.md`: versioned query dirs, metadata/schema files, incremental labels, partition filters, escaped keyword columns, and cross-project access through a syndicate dataset.
- Ran live BigQuery checks against the Azure billing exports to confirm the lookup join, available export table names, case-insensitive VM extraction, and non-null VM names on recent sample dates.
- Dry-ran the Azure-only query fragment with the account that can read the Azure billing export.
- Dry-ran `task_run_costs_v1` with the account that can read the existing FXCI derived tables.

A full local dry run of the combined `worker_costs_v1` query is blocked until the `azure_billing_syndicate` views exist and the source-side Terraform authorization is applied.

## Related

- [RELOPS-2330](https://mozilla-hub.atlassian.net/browse/RELOPS-2330)
- [DENG-11127](https://mozilla-hub.atlassian.net/browse/DENG-11127)
- [mozilla/global-platform-admin#6256](https://github.com/mozilla/global-platform-admin/pull/6256)
- [mozilla/global-platform-admin#6235](https://github.com/mozilla/global-platform-admin/pull/6235)

[RELOPS-2330]: https://mozilla-hub.atlassian.net/browse/RELOPS-2330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DENG-11127]: https://mozilla-hub.atlassian.net/browse/DENG-11127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jwmossmoz:
Leaving in draft to review with @ahal and folks within data engineering as this is my first attempt at getting this working.

sql/moz-fx-data-shared-prod/fxci_derived/worker_costs_azure_v1/schema.yaml
```diff
@@ -0,0 +1,19 @@
...
+  mode: NULLABLE
+- name: instance_id
+  type: STRING
+  mode: NULLABLE
+- name: total_cost
+  type: FLOAT
+  mode: NULLABLE
+- name: usage_start_date
+  type: DATE
+  mode: NULLABLE
```
ahal: It looks like the schema is that same as the normal worker_costs table. Is there a way we can avoid having a separate table for Azure workers? Or is a separate table desirable?

If we merge them into a single table, maybe we can add a `cloud_provider = {azure|gcp}` key in case consumers ever want to filter only one or the other

sql/moz-fx-data-shared-prod/fxci_derived/worker_metrics_azure_v1/query.sql
```diff
@@ -0,0 +1,84 @@
...
+        THEN 3
+      WHEN "taskFinish"
+        THEN 4
+      WHEN "instanceReboot"
+        THEN 5
+      WHEN "instanceShutdown"
+        THEN 6
+    END AS event_order
+  FROM
+    `moz-fx-data-shared-prod.taskclusteretl.worker_metrics`,
```
ahal: Taskcluster ETL is being decomissioned so we can't rely on it. The whole point of this repo new ETL was to provide an alternative to it :).

Unfortunately this means we'll need to write code in:
https://github.com/mozilla/docker-etl/tree/main/jobs/fxci-taskcluster-export

that populates the metrics data for us. I think for now, this PR can just focus on the cost aspect, and we can work on the metrics aspect as we go.

ahal (changes requested):
Thanks, this is a great start!

sql/moz-fx-data-shared-prod/fxci_derived/worker_costs_v1/metadata.yaml
```diff
@@ -1,15 +1,19 @@
...
 labels:
   incremental: true
   owner1: ahalberstadt
   dag: bqetl_fxci
 scheduling:
   dag_name: bqetl_fxci
   task_name: fxci_worker_cost__v1
+  arguments: ['--schema_update_option=ALLOW_FIELD_ADDITION']
+schema:
+  allow_field_addition: true
```
ahal: Why do we need `allow_field_addition`?

ahal (approved):
Thanks this looks reasonable. Assuming it's been tested and works, no objections from me. I'll flag Data SRE as well as I'm not very good at SQL.

sql/moz-fx-data-shared-prod/fxci_derived/worker_costs_v1/metadata.yaml
jwmossmoz: Due to an additional cloud_provider column to `worker_costs_v1` based on https://github.com/mozilla/bigquery-etl/blob/main/docs/reference/recommended_practices.md#dynamic-schemas

sql/moz-fx-data-shared-prod/fxci_derived/worker_costs_v1/schema.yaml
```diff
@@ -1,4 +1,7 @@
 fields:
+- name: cloud_provider
+  type: STRING
+  mode: NULLABLE
```
BenWu: As a convention, new columns should be added at the end of the schema file.  This is because new columns are always added to the end of the bigquery table.

sql/moz-fx-data-shared-prod/fxci_derived/worker_costs_v1/metadata.yaml
BenWu: The wording is confusing but `allow_field_addition: true` isn't necessary here because the schema will be updated during deployment using the schema.yaml.  It would be needed if the schema might be updated at query-time, e.g. the query does `SELECT *` from an upstream table and the upstream table might have fields added.
```suggestion
```

sql/moz-fx-data-shared-prod/fxci_derived/worker_costs_v1/query.sql
```diff
@@ -1,4 +1,36 @@
+WITH azure_billing AS (
+  SELECT
+    daily_load.subscriptionName,
+    daily_load.`date`,
+    daily_load.ResourceId,
+    daily_load.resourceLocation,
+    daily_load.costInBillingCurrency,
+    daily_load.chargeType
+  FROM
+    `moz-fx-data-billing-prod-9147.azure_billing_raw.fxci_daily_actual_load` AS daily_load
```
BenWu: There are a few service accounts that would need access to these tables, `default-workloads@moz-fx-data-airflow-gke-prod.iam.gserviceaccount.com`, `bigquery-etl-dryrun@moz-fx-data-shared-prod.iam.gserviceaccount.com`, and probably more.  Is that already being setup somewhere else?  One way to do this would be to create authorized views in `moz-fx-data-shared-prod` for these tables following https://mozilla-hub.atlassian.net/wiki/spaces/IP/pages/473727279/Importing+Data+from+OLTP+Databases+to+BigQuery+via+Federated+Queries

jwmossmoz: Addressed by adding the bqetl-side `azure_billing_syndicate` dataset metadata and the companion source-side Terraform PR: https://github.com/mozilla/global-platform-admin/pull/6256

sql/moz-fx-data-shared-prod/azure_billing_syndicate/dataset_metadata.yaml
```diff
@@ -0,0 +1,19 @@
...
+  syndicated_tables:
+    - fxci_daily_actual_load
+    - fxci_daily_actual_lookup
+    - taskcluster_daily_actual_load
+    - taskcluster_daily_actual_lookup
+  administer_views: false
+workgroup_access:
+  - role: roles/bigquery.dataViewer
+    members:
+      - workgroup:mozilla-confidential/data-viewers
```
BenWu, line 19: Just double-checking: this is fine to be available to all employees right?  I assume this is correct since it matches the GCP billing table.

BenWu (approved):
This looks correct to me.  I'll hold off on merging this until the syndication change is ready to be applied so we don't get stuck with a broken job

PR #9306, opened 2026-05-04
feat(CI): Store DAGs in new `generated-dags` and `private-generated-dags` branches (DENG-10984)

sean-rose:
## Description
So Airflow's Git sync process doesn't need to download hundreds of megabytes of unnecessary SQL files (DENG-10984).

## Related Tickets & Documents
* DENG-10984: Airflow pods' `git-sync-init` containers can be slow

**Reviewer, please follow [this checklist](https://github.com/mozilla/bigquery-etl/blob/main/.github/reviewer_checklist.md)**

PR #9318, opened 2026-05-05
chore(CI): Remove temporary CI logic from #9306 (DENG-10984)

sean-rose:
## Description
Now that https://github.com/mozilla/bigquery-etl/pull/9306 has taken effect some of the logic necessary for that transition can be removed.

I'm also taking this opportunity to make the `sync-bigquery-etl-submodule` job's logic more consistent with the `sync-private-bigquery-etl-submodule` job.

* DENG-10984: Airflow pods' `git-sync-init` containers can be slow
* https://github.com/mozilla/bigquery-etl/pull/9306
* https://github.com/mozilla/bigquery-etl/pull/8867

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1400,9 +1400,11 @@ jobs:
         run: |
           git clone --branch main git@github.com:mozilla/telemetry-airflow-dags.git
           cd telemetry-airflow-dags
-          # Manually clone bigquery-etl submodule to avoid initializing private repos
-          rm -rf bigquery-etl
-          git clone --branch generated-dags --depth 1 https://github.com/mozilla/bigquery-etl.git bigquery-etl
+          git submodule update --init --depth 1 bigquery-etl
+          cd bigquery-etl
+          git fetch --depth=1 origin generated-dags
+          git reset --hard FETCH_HEAD
+          cd ..
           git add bigquery-etl
           git commit --allow-empty -m "Automatic commit from bigquery-etl commit ${{ github.sha }} [skip ci]"
           git push origin main
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1125,7 +1125,6 @@ jobs:
           cd generated-sql/
           rm -rf sql/
           cp -r /tmp/workspace/generated-sql/sql sql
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -1402,9 +1400,11 @@ jobs:
         run: |
           git clone --branch main git@github.com:mozilla/telemetry-airflow-dags.git
           cd telemetry-airflow-dags
-          # Manually clone bigquery-etl submodule to avoid initializing private repos
-          rm -rf bigquery-etl
-          git clone --branch generated-dags --depth 1 https://github.com/mozilla/bigquery-etl.git bigquery-etl
+          git submodule update --init --depth 1 bigquery-etl
```
sean-rose, line 1403: **note:** I believe specifying the `bigquery-etl` submodule in this `git submodule update` command should avoid initializing the `private-bigquery-etl` submodule (prior to https://github.com/mozilla/bigquery-etl/pull/8867 which implemented the manual cloning approach it was calling `git submodule update --init --recursive` with no particular submodule specified).

.github/workflows/build.yml
```diff
@@ -1402,9 +1400,11 @@ jobs:
...
+          cd bigquery-etl
+          git fetch --depth=1 origin generated-dags
+          git reset --hard FETCH_HEAD
```
sean-rose, line 1406: **note:** One benefit of using this `git fetch` + `git reset` approach instead of `git pull` is that if we ever need to manually modify the generated DAG branches for some reason then this will handle that more gracefully.

scholtzan (approved):

.github/workflows/build.yml
sean-rose: Although maybe that point is moot, per this comment from Claude:
> `git submodule update --init --depth 1 bigquery-etl` checks out the commit the parent repo's submodule pointer currently references

So in a situation where we've manually changed the `generated-dags` branch history the existing `bigquery-etl` submodule pointer may reference a no-longer-existent commit and fail.

Given that, I don't have a strong preference on which approach we use for updating submodules, except that I think we should use the same approach in both jobs to minimize the number of Git edge cases that need to be considered/handled.

PR #9319, opened 2026-05-05
fix: remove duplicate fields from the final select

kbammarito:
## Description

This PR removes duplicate fields from the final `select`. `device_manufacturer` is an exact duplicate and `isp_name` already exists as `isp`.

Also, manually dropping the fields. i.e.,
```sql
ALTER TABLE `moz-fx-data-shared-prod.fenix_derived.clients_last_seen_joined_v1`
DROP COLUMN device_manufacturer_1,
DROP COLUMN isp_name;
```

* https://bugzilla.mozilla.org/show_bug.cgi?id=2037207


sql_generators/glean_usage/templates/clients_last_seen_joined.query.sql
```diff
@@ -30,8 +30,17 @@ SELECT
...
   baseline.normalized_channel,
-  * EXCEPT (submission_date, normalized_channel, client_id, sample_id, is_default_browser),
+  * EXCEPT (
+      submission_date,
+      normalized_channel,
+      client_id,
+      sample_id,
+      is_default_browser,
+      device_manufacturer,
+      isp_name
```
wwyc: question: is there a reason why it has to be named `isp_name`? it seems like `isp` is more aligned with existing column names

wwyc (approved):
approving this for now to fix the DAG failures
however, unless there is a specific reason for naming the column `isp_name`, it should be aligned with existing columns as `isp`

sql_generators/glean_usage/templates/clients_last_seen_joined.query.sql
kbammarito: Yeah, I'm removing it completely (not adding it back in, like I did with `device_manufacturer` on line 42). `isp_name` is from the already-created `metrics_*` generators ([PR](https://github.com/mozilla/bigquery-etl/pull/9278)), but not needed, since it already exists as `isp` from `baseline`:

```
baseline: metadata.isp.name AS isp,
metrics: mozfun.stats.mode_last(
        ARRAY_AGG(metadata.isp.name ORDER BY submission_timestamp)
      ) AS isp_name,
```

wwyc: I think the upstream field should also be named `isp` so you don’t need to reference `isp_name` here at all — you can just handle it the same way as `device_manufacturer`.

kbammarito: Resolving this for now so we can get the DAGs fixed. I added a comment to explain `isp_name`. Will revisit.

PR #9321, opened 2026-05-06
fix: correct the addon filter inside dev_amo_stats_dau_v1

kik-kik:
# fix: correct the addon filter inside dev_amo_stats_dau_v1


badboy (approved):
uh oh, bad case of not-nesting.
r+ with the correct nesting now.

lucia-vargas-a (approved):

PR #9346, opened 2026-05-11
add visit level widget table

jaredsnyder:
This PR adds a visit-level aggregate table for widgets.

Widgets key experiments related to widgets are currently happening, and widgets are expected to be rolled out widely in June.  This table enables the analysis of widget interactions at the visit level and will be used to create a client-level table that can be used in experiments.

The table was modeled on `firefox_desktop_derived.newtab_visits_daily_v1`.  It copies the same strategy for creating dimensions for visits.

The table was manually verified for several visits, as documented here: https://docs.google.com/document/d/1d4H1JegVWJ4EhvAvVd95VD0WftS8nozdD6xitx81ji0/edit?tab=t.0


sql/moz-fx-data-shared-prod/firefox_desktop_derived/widgets_visit_daily_v1/query.sql
```diff
@@ -0,0 +1,172 @@
...
+  visit_aggregations.* EXCEPT (submission_date, client_id, newtab_visit_id),
+  user_action_counts_summary.user_action_counts,
+FROM
+  widget_metrics
+LEFT JOIN
+  visit_aggregations
+  USING (submission_date, client_id, newtab_visit_id)
+LEFT JOIN
+  user_action_counts_summary
+  USING (submission_date, client_id, newtab_visit_id, widget_name, widget_size)
```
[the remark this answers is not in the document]
jaredsnyder: Updated schema to clarify this case

sql/moz-fx-data-shared-prod/firefox_desktop_derived/widgets_visit_daily_v1/metadata.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - jsnyder@mozilla.com
+labels:
+  application: newtab
+  incremental: true
+  schedule: daily
+  dag: bqetl_newtab
+  owner1: jsnyder
+  table_type: client_level
+scheduling:
+  dag_name: bqetl_newtab
```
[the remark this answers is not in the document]
jaredsnyder: removed

sql/moz-fx-data-shared-prod/firefox_desktop_derived/widgets_visit_daily_v1/metadata.yaml
```diff
@@ -0,0 +1,30 @@
+friendly_name: Widgets Visit Daily
```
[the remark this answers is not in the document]
jaredsnyder: added

sql/moz-fx-data-shared-prod/firefox_desktop/widgets_visit_daily/metadata.yaml
```diff
@@ -0,0 +1,9 @@
+friendly_name: Widgets Visit Daily
+description: |-
+  Daily aggregation of newtab widget impression and user-event metrics.
+  The granularity is one row per Firefox Desktop (newtab_visit_id, widget_name, widget_size)
+  per submission_date, unpacking the `widgets_impression` and `widgets_user_event`
+  events from the `newtab` event category. Adds a derived `windows_version` column
+  over the underlying `firefox_desktop_derived.widgets_visit_daily_v1` table.
```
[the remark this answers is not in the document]
jaredsnyder: removed

sql/moz-fx-data-shared-prod/firefox_desktop_derived/widgets_visit_daily_v1/query.sql
```diff
@@ -0,0 +1,167 @@
...
+      mozfun.norm.browser_version_info(client_info.app_display_version).major_version AS INT64
+    ) AS app_version,
+    normalized_os AS os,
+    normalized_os_version AS os_version,
+    client_info.windows_build_number AS windows_build_number,
+    normalized_channel AS channel,
+    client_info.locale AS locale,
+    normalized_country_code AS country,
+    sample_id,
+    metrics.string.newtab_homepage_category AS homepage_category,
```
land-edi, line 16: Most of all these columns are available in the newtab_visits_daily data. Do we need to reduplicate them here? Or maybe another option can be to include these widgets data into the newtab_visits by add new columns?

land-edi (approved):
I am wondering whether we should create a new dataset for widgets given it is copying all the same columns we already have in newtab_visits_daily, which include today everything except widgets. Do you think it is will be better to just include the few widget columns you are creating itno that dataset? Beside that, everything else looks good

sql/moz-fx-data-shared-prod/firefox_desktop_derived/widgets_visit_daily_v1/query.sql
jaredsnyder: I would say so but I think it's up for debate. My argument is that they are all dimensions that we may want to slice the data by, and that operations is easiest and most efficient by leaving them in this table.  The counterpoint is that you could leave them out and, when necessary, join them on from `newtab_visits_daily`.

sql/moz-fx-data-shared-prod/firefox_desktop_derived/widgets_visit_daily_v1/query.sql
```diff
@@ -0,0 +1,167 @@
...
+  FROM
+    `moz-fx-data-shared-prod.firefox_desktop_stable.newtab_v1`,
+    UNNEST(events)
+  WHERE
+    DATE(submission_timestamp) = @submission_date
+    AND category = 'newtab'
+    -- include `opened` so that is_default_ui can be derived per visit
+    AND name IN ('opened', 'widgets_impression', 'widgets_user_event', 'widgets_enabled')
+),
+visit_aggregations AS (
```
gkatre, line 54: Extending the conversation that @land-edi started about using the newtab_visits_daily table, have you considered using it in lieu of the `visit_aggregations` CTE? That way you would not need to reaggregate the data using exactly the same `GROUP BY`.
Also, the need for `name IN ('opened')` [line 52] may not be necessary because many of those fields are already computed in the newtab_visits_daily table.

If you choose this path, you would need to do the following:
- events_unnested CTE would only need the widget related fields
- the final merge [line 162]: the visit_aggregations CTE will be replaced by the newtab_visits_daily table filtered by the @submission_date to pick one table partition

jaredsnyder: I did consider using newtab_visits, the main argument not to is that I add a few extra columns (submission_timestamp and windows_build_number) and since I had to extract them anyway it felt like I might as well pull all the dimensions and avoid having an additional join.

jaredsnyder:
Did a quick run of the most recent file and compared to the validated v1 version.  There were <1000 discrepancies and the enabled and disabled columns look fine in the aggregate so I think we're good.  Updated the validation doc with this info.

sql/moz-fx-data-shared-prod/firefox_desktop_derived/widgets_visit_daily_v1/query.sql
jaredsnyder: I'm going to leave it as is because it's a little more efficient and frankly I don't want to have to re-validate 😅

PR #9325, opened 2026-05-06
Use target-based deploys for stage deploys

scholtzan:
## Related Tickets & Documents
* DENG-XXXX
* DSRE-XXXX

**Reviewer, please follow [this checklist](https://github.com/mozilla/bigquery-etl/blob/main/.github/reviewer_checklist.md)**

PR #9350, opened 2026-05-11
Use target-based deploys for stage deploys

scholtzan:
## Description

This PR switches CI stage deploys from the legacy `bqetl stage deploy/clean` commands to the `--target stage` framework with `--isolated`. Some fixes to the deploy mechanism that were discovered during testing.
Tests I ran:
* Make change to `clients_last_seen` template -> deployed 200+ artifacts in a few minutes: https://github.com/mozilla/bigquery-etl/actions/runs/25578721133/job/75093595952
    * This got previously stuck on deploys (https://github.com/mozilla/bigquery-etl/pull/9278)
* Deploys on queries with UDFs, SQL tests, materialized views: https://github.com/mozilla/bigquery-etl/actions/runs/25578721133/job/75093595952

(changes copied from https://github.com/mozilla/bigquery-etl/pull/9325, but new PR to clear up history)


.github/workflows/build.yml
```diff
@@ -479,19 +479,31 @@ jobs:
...
+        run: |
+          cat > bqetl_targets.yaml <<'EOF'
+          stage:
+            project_id: moz-fx-data-integration-tests
+            dataset: "{{ git.branch }}_{{ git.commit }}{% if run_id %}_{{ run_id }}{% endif %}"
+            artifact_prefix: "{{ artifact.project_id }}_{{ artifact.dataset_id }}_"
+            grant_dryrun_access: true
+            expire_after_hours: 12
+            rewrite_tests: true
+          EOF
```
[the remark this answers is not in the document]
scholtzan: This is intentional. We don't want to commit the file because otherwise users might accidentally commit their changes to the file

BenWu (approved):
I don't see any CI runs for this branch that deployed anything to stage so I think you could add some trivial change to a query to test the workflow before merging

.github/workflows/build.yml
```diff
@@ -479,19 +479,31 @@ jobs:
...
       - *auth-gcp-dryrun
       - *auth-gcp-stage
+      - &write-bqetl-targets
+        name: Write bqetl_targets.yaml
+        run: |
+          cat > bqetl_targets.yaml <<'EOF'
+          stage:
+            project_id: moz-fx-data-integration-tests
+            dataset: "{{ git.branch }}_{{ git.commit }}{% if run_id %}_{{ run_id }}{% endif %}"
+            artifact_prefix: "{{ artifact.project_id }}_{{ artifact.dataset_id }}_"
```
sean-rose: **suggestion:** For readability:
```suggestion
            artifact_prefix: "{{ artifact.project_id }}__{{ artifact.dataset_id }}__"
```

.github/workflows/build.yml
```diff
@@ -479,19 +479,31 @@ jobs:
...
+            grant_dryrun_access: true
+            expire_after_hours: 12
+            rewrite_tests: true
+          EOF
       - name: Deploy changes to stage
         timeout-minutes: 120
         env:
           SKIP_STAGE_DEPLOYS: ${{ github.event.inputs.skip-stage-deploys }}
           GOOGLE_GHA_ID_TOKEN: ${{ steps.auth-dryrun.outputs.id_token }}
+          BQETL_RUN_ID: ${{ github.run_id }}
```
sean-rose: **question:** In https://github.com/mozilla/bigquery-etl/pull/8943 @BenWu opted to use `run_id` + `run_number`. Should we do something similar here?

After re-reading the [GitHub docs](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context) it seems like `run_id` and `run_number` work nearly the same way, except `run_number` values are smaller since they start from 1. So I'm thinking we should either:
- Leave this as is, so all attempts of a specific workflow run target the same stage dataset.
- Include `run_attempt` in the run ID value passed to `bqetl stage` commands, to target distinct stage datasets for each workflow run attempt.

Attempt-specific datasets seems like overkill, and would probably complicate the cleanup process, so I'd lean toward leaving this as is.

.github/workflows/build.yml
```diff
@@ -479,19 +479,31 @@ jobs:
...
         run: |
           if [ "$SKIP_STAGE_DEPLOYS" != "true" ]; then
             PATHS="$(git diff --no-index --name-only --diff-filter=d generated-sql/sql sql)" || true
             echo $PATHS
             git diff --no-index --diff-filter=d generated-sql/sql sql | head -1000 || true
-            PATH=".venv/bin:$PATH" script/bqetl stage deploy \
-              --dataset-suffix="${GITHUB_SHA}_${GITHUB_RUN_ID}" \
-              --remove-updated-artifacts \
+            PATH=".venv/bin:$PATH" script/bqetl --target stage deploy \
+              --tables --views --isolated \
```
sean-rose: **suggestion:** It feels incongruous to use the `BQETL_RUN_ID` environment variable here, but the `--run-id` argument for the `bqetl stage target clean` command below.  I'd suggest consistently using the same approach for both.

Personally I'd lean toward using an explicit `--run-id` argument, and maybe even removing the `BQETL_RUN_ID`/`GITHUB_RUN_ID` fallback options as unnecessary complexity.

bigquery_etl/cli/target.py
```diff
@@ -81,6 +84,9 @@ def target(ctx):
     git.branch is in artifact_prefix, only matching tables within datasets are
     deleted. Both filters apply when branch appears in both templates.
 
+    --run-id deletes artifacts for a specific run id. Reads from $BQETL_RUN_ID
+    / $GITHUB_RUN_ID when not passed.
```
sean-rose: **suggestion:** Correction:
```suggestion
    --run-id deletes artifacts for a specific run id. Reads from $BQETL_RUN_ID
    when not passed.
```

bigquery_etl/cli/target.py
```diff
@@ -111,6 +119,13 @@ def target(ctx):
     default=None,
     help="Delete deployments for a specific branch.",
 )
+@click.option(
+    "--run-id",
+    "--run_id",
+    "run_id",
+    default=None,
+    help="Delete deployments for a specific run id (defaults to $BQETL_RUN_ID / $GITHUB_RUN_ID if set).",
```
sean-rose: **suggestion:** Correction:
```suggestion
    help="Delete deployments for a specific run id (defaults to $BQETL_RUN_ID if set).",
```

bigquery_etl/cli/target.py
```diff
@@ -134,34 +150,44 @@ def target(ctx):
...
     )
     artifact_re = None
-    if (
-        branch
-        and target_config.raw_artifact_prefix
-        and "git.branch" in target_config.raw_artifact_prefix
+    if target_config.raw_artifact_prefix and (
+        (branch and "git.branch" in target_config.raw_artifact_prefix)
+        or (run_id and "run_id" in target_config.raw_artifact_prefix)
     ):
```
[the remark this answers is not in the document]
sean-rose: I disagree with the bot's suggestion, and think matching on `{{ run_id` is unnecessarily specific/fragile.

Instead I'd suggestion something like:
```suggestion
    if target_config.raw_artifact_prefix and (
        (branch and re.search(r"\bgit\.branch\b", target_config.raw_artifact_prefix))
        or (run_id and re.search(r"\brun_id\b", target_config.raw_artifact_prefix))
):
```

bigquery_etl/cli/target.py
```diff
@@ -134,34 +149,45 @@ def target(ctx):
...
 @block_coding_agents
 @click.pass_context
-def clean(ctx, older_than, branch, all_deployments, dry_run, yes, sql_dir):
+def clean(ctx, older_than, branch, run_id, all_deployments, dry_run, yes, sql_dir):
     """Clean up target environment deployments."""
-    if not any([older_than, branch, all_deployments]):
+    # Only fall back to BQETL_RUN_ID (opt-in) — not GITHUB_RUN_ID, which is
+    # auto-set on every Actions runner and would silently filter a local
+    # `target clean` invoked from a CI-adjacent shell. Skip the fallback
+    # entirely when --all is set, to avoid a surprising mutex error.
```
sean-rose: **suggestion:** Saying the mutex error is "surprising" implies to me that it's non-obvious, in which case I think it would be helpful to actually describe the error (at least briefly) rather than just allude to it.

bigquery_etl/util/target.py
```diff
@@ -228,19 +248,60 @@ def _get_targets_file() -> Path:
...
-        }
+        repo = git.Repo(get_bqetl_project_root() or ROOT)
+        try:
+            branch = repo.active_branch.name
+        except TypeError:
+            branch = env_branch
+        try:
+            commit = repo.head.commit.hexsha
+        except Exception:
+            commit = env_commit
```
sean-rose: **suggestion:** IMO it would be better to check for detached head state directly rather than relying on the Git SDK raising a `TypeError`, and `repo.head.commit.hexsha` seems to work fine in detached head state (plus the outer `try` already catches generic exceptions).
```suggestion
        branch = repo.active_branch.name if not repo.head.is_detached else env_branch
        commit = repo.head.commit.hexsha
```

bigquery_etl/util/target.py
```diff
@@ -473,6 +535,22 @@ def _existing_artifact_file(source_dir: Path) -> Optional[Path]:
     return None
 
 
+def default_partition_for(dataset: str) -> Optional[str]:
+    """Guess the partition-filter column from a Mozilla dataset suffix.
```
sean-rose, line 560: **question:** Why are we guessing? Can't we get the partition column name from the table's metadata?

bigquery_etl/util/target.py
```diff
@@ -593,12 +690,21 @@ def _udf_dep_paths(udf_names: List[str]) -> Set[Path]:
...
+        if udf in seen or udf not in raw_routines:
             continue
+        seen.add(udf)
         for transitive in accumulate_dependencies([], raw_routines, udf):
             if transitive in raw_routines:
                 paths.add(Path(raw_routines[transitive].filepath))
+                # test_dependencies aren't picked up by accumulate_dependencies;
+                # queue them so their own deps get walked too.
+                for test_dep in raw_routines[transitive].test_dependencies:
+                    if test_dep not in seen:
```
sean-rose: **question:** Shouldn't this also check if the test dependency is already in `queue`?
```suggestion
                    if test_dep not in queue and test_dep not in seen:
```

bigquery_etl/util/target.py
scholtzan: Yes, that's a good point. We can use the partitioning if there is a `metadata.yaml` file. We'll still have to fall back to this method though for tables that are not managed in bigquery-etl (e.g. stable tables)

PR #9353, opened 2026-05-12
Fork PR approval gate

scholtzan:
## Description

This PR replaces per-job `environment: external-fork` approval prompts with a single gate workflow (`fork-gate.yml`). One maintainer approval per fork PR push now authorizes the whole build run, instead of ~6 separate clicks. `build.yml` is triggered for fork PRs via `workflow_run` after the gate completes successfully. It checks out `workflow_run.head_sha` so it runs against exactly the SHA the maintainer approved. There is a new `report-checks` job that posts a check run per upstream job to the PR's head SHA, so the PR UI continues to show per-job pass/fail (which `workflow_run` doesn't surface natively).

Testing this without merging is a little difficult. I propose we merge this and then open a fork PR to test that everything is working as intended. If not, we can revert this change.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1522,6 +1522,7 @@ jobs:
       - test-bqetl
       - integration
       - generate-sql
+      - private-generate-sql 
       - docs
       - deploy-changes-to-stage
       - dry-run-sql
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1512,7 +1512,7 @@ jobs:
   # the fork-PR Checks UI without needing edits here.
   report-checks:
     name: Report PR checks
-    if: github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success'
+    if: always() && github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success'
     needs:
       - decide-runs
       - build
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1474,6 +1474,11 @@ jobs:
   # Only runs when build.yml is triggered via `workflow_run` (i.e. after the
   # fork-gate approval). For same-repo PRs the native pull_request checks
   # already appear in the PR UI, so this job is unnecessary there.
+  #
+  # !!! COUPLING: the `needs:` list below must include every job whose status
+  # should appear in the fork-PR Checks UI. If you ADD a job to this workflow
+  # that should be visible to fork-PR reviewers, ADD it to `needs:` here too.
+  # Forgetting means the new job runs but its status never reaches the PR.
   report-checks:
     name: Report PR checks
     if: github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success'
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -62,8 +63,11 @@ jobs:
           persist-credentials: false
           filter: blob:none
           fetch-depth: 0
-          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          ref: ${{ github.event.workflow_run.head_sha || github.event.pull_request.head.sha || github.sha }}
       - id: changed-files-routines
+        # Skip path-based gating for fork PRs reaching us via workflow_run;
+        # PR event payload isn't available, so we run everything conservatively.
+        if: github.event_name != 'workflow_run'
```
scholtzan: The most noticeable change is that for forks no CI checks will run until someone approves the run. I don't think that's a blocker though.

.github/workflows/build.yml
```diff
@@ -1453,3 +1469,61 @@ jobs:
...
           git push origin main
+
+  # Surface upstream job results as Checks-API check runs on the PR's head SHA.
+  # Only runs when build.yml is triggered via `workflow_run` (i.e. after the
+  # fork-gate approval). For same-repo PRs the native pull_request checks
+  # already appear in the PR UI, so this job is unnecessary there.
+  report-checks:
+    name: Report PR checks
+    if: github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success'
+    needs:
```
scholtzan, line 1516: This step is needed otherwise the triggered workflow jobs won't appear as checks in the PR UI

BenWu (approved):

.github/workflows/build.yml
```diff
@@ -1453,3 +1500,104 @@ jobs:
...
+  # fork-gate approval). For same-repo PRs the native pull_request checks
+  # already appear in the PR UI, so this job is unnecessary there.
+  #
+  # `needs:` enforces ordering (run after all upstream jobs complete). The
+  # actual list of jobs to report is derived dynamically from the Actions Jobs
+  # API at runtime, so new jobs added to this workflow automatically appear in
+  # the fork-PR Checks UI without needing edits here.
+  report-checks:
+    name: Report PR checks
+    if: github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success'
```
akkomar:
```suggestion
    if: always() && github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success'
```
issue: this won't run if any of the build jobs fail. Per https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idneeds:
> If you would like a job to run even if a job it is dependent on did not succeed, use the always() conditional expression in jobs.<job_id>.if

.github/workflows/build.yml
```diff
@@ -1453,3 +1500,104 @@ jobs:
...
+    needs:
+      - decide-runs
+      - build
+      - verify-requirements
+      - verify-format-yaml
+      - verify-format-sql
+      - test-bqetl
+      - integration
+      - generate-sql
```
akkomar, line 1524:
```suggestion
      - generate-sql
      - private-generate-sql 
```

akkomar (approved):
r+wc

PR #9366, opened 2026-05-13
feat(nimbus): Experiment enrollment funnel data

yashikakhurana:
## Description
This PR adds a new daily ETL job that exports a cumulative enrollment funnel rollup to GCS for Experimenter visualization and alerting.

For each live recipe, it queries the nimbus_targeting_context ping tables across firefox_desktop, firefox_ios, and fenix, takes the latest evaluation per client per recipe, and groups by (app_name, slug, status, reason, conflict_slug) with a distinct client count. The result is exported to gs://mozanalysis/enrollment_counts/enrollment_funnel_latest.json — the same bucket and cadence as the existing enrollment alert data — for Experimenter to consume.

* [EXP-6871](https://mozilla-hub.atlassian.net/browse/EXP-6871) — ETL: enrollment funnel rollup in bigquery-etl
* [EXP-6870](https://mozilla-hub.atlassian.net/browse/EXP-6870) — Enrollment Funnel Visualization (parent epic)

[EXP-6871]: https://mozilla-hub.atlassian.net/browse/EXP-6871?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EXP-6870]: https://mozilla-hub.atlassian.net/browse/EXP-6870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


yashikakhurana:
Quick flag on the schema.yaml — wanted to check if this is the right approach before merging.

The stage deploy only processes files in the PR diff, so experiment_enrollment_alert_data_v1 (same structure, no schema.yaml) isn't affected. But if someone modifies it in a future PR it would hit the same failure.

Since our script writes to GCS and creates no BQ table, adding schema.yaml feels a bit odd — the stage deploy will create a staging BQ table that doesn't exist in prod. Is this the intended pattern for Python GCS exporters? Or would it make more sense to patch stage.py to skip query.py files without a schema.yaml (similar to how view files are already handled via DryRun.skipped_files() at line 606)? Happy to go either way, just didn't want to assume.
cc @scholtzan  and @mikewilli

sql/moz-fx-data-experiments/monitoring/experiment_enrollment_funnel_v1/schema.yaml
mikewilli, line 1: Maybe I missed it, but I don't think we're writing to a bigquery table, are we? If not, I don't think we need a schema. Might be useful to do so though?

sql/moz-fx-data-experiments/monitoring/experiment_enrollment_funnel_v1/query.py
```diff
@@ -0,0 +1,203 @@
...
+  SELECT normandy_slug AS slug
+  FROM `moz-fx-data-experiments.monitoring.experimenter_experiments_v1`
+  WHERE start_date IS NOT NULL
+    AND (end_date IS NULL OR end_date >= DATE_SUB(@run_date, INTERVAL 90 DAY))
+),
+all_apps_raw AS (
+  -- firefox_desktop
+  SELECT
+    'firefox_desktop' AS app_name,
+    t.client_info.client_id AS client_id,
```
mikewilli, line 44: Probably want the `profile_group_id` here, or maybe use the app from the experiments table?

sql/moz-fx-data-experiments/monitoring/experiment_enrollment_funnel_v1/query.py
```diff
@@ -0,0 +1,203 @@
...
+    (SELECT value FROM UNNEST(event.extra) WHERE key = 'slug') AS slug,
+    (SELECT value FROM UNNEST(event.extra) WHERE key = 'status') AS status,
+    (SELECT value FROM UNNEST(event.extra) WHERE key = 'reason') AS reason,
+    (SELECT value FROM UNNEST(event.extra) WHERE key = 'conflict_slug') AS conflict_slug,
```
mikewilli, line 48: Branch might be helpful too.

sql/moz-fx-data-experiments/monitoring/experiment_enrollment_funnel_v1/query.py
```diff
@@ -0,0 +1,203 @@
...
+    t.submission_timestamp
+  FROM `moz-fx-data-shared-prod.firefox_desktop.nimbus_targeting_context` t,
+  UNNEST(t.events) AS event
+  WHERE DATE(t.submission_timestamp) BETWEEN @min_start_date AND @run_date
+    AND event.category = 'nimbus_events'
+    AND event.name = 'enrollment_status'
```
mikewilli, line 56: The `min_start_date` is 2023-06-06 right now, and that plus the CROSS JOIN on the unnested events makes this is a huge query. Can we do a 1% sample for each of these? Or maybe we should consider creating a separate ETL just to pull out the enrollment_status events that could go upstream of this?

sql/moz-fx-data-experiments/monitoring/experiment_enrollment_funnel_v1/query.py
```diff
@@ -0,0 +1,203 @@
...
+  -- fenix
+  SELECT
+    'fenix' AS app_name,
+    t.client_info.client_id AS client_id,
+    (SELECT value FROM UNNEST(event.extra) WHERE key = 'slug') AS slug,
+    (SELECT value FROM UNNEST(event.extra) WHERE key = 'status') AS status,
+    (SELECT value FROM UNNEST(event.extra) WHERE key = 'reason') AS reason,
+    (SELECT value FROM UNNEST(event.extra) WHERE key = 'conflict_slug') AS conflict_slug,
+    t.submission_timestamp
+  FROM `moz-fx-data-shared-prod.org_mozilla_fenix.nimbus_targeting_context` t,
```
mikewilli: I believe this is the nightly dataset, you likely want `fenix`:
```suggestion
  FROM `moz-fx-data-shared-prod.fenix.nimbus_targeting_context` t,
```

sql/moz-fx-data-experiments/monitoring/experiment_enrollment_funnel_v1/schema.yaml
yashikakhurana: we are not writing to it, but ci is erroring it out if I dont provide it https://github.com/mozilla/bigquery-etl/pull/9366#issuecomment-4448135413

sql/moz-fx-data-experiments/monitoring/experiment_enrollment_funnel_v1/query.py
jaredlockhart: Sorry I should've specified in the design that it was expected we'd be subsampling here.  I used 10% when I prototyped and it seemed fine, we can do smaller if operationalizing it is even more strenuous, we can try 1% and see if that's good enough?

sql/moz-fx-data-experiments/monitoring/experiment_enrollment_funnel_v1/query.py
yashikakhurana: I believe client_id is the right choice here since Nimbus uses it as the randomization unit for bucket assignment — but happy to be corrected if profile_group_id is preferred for this use case. What do you think?

mikewilli: `profile_group_id` is used for bucketing on desktop, and a different ID is used for mobile (`client_id` is used for analysis, but there's a separate `nimbus_id` that is used as the randomization unit for bucketing). So I think it makes sense to use `profile_group_id` for desktop, and `client_id` for mobile.

sql/moz-fx-data-experiments/monitoring/experiment_enrollment_funnel_v1/query.py
```diff
@@ -0,0 +1,211 @@
...
+        # No active experiments — write empty files and exit cleanly.
+        storage_client = storage.Client(args.project)
+        bucket = storage_client.bucket("mozanalysis")
+        empty = json.dumps({"v1": {}})
+        for path in [
+            f"{args.gcs_folder}/enrollment_funnel_v1_{args.date}.json",
+            f"{args.gcs_folder}/enrollment_funnel_v1_latest.json",
+        ]:
+            bucket.blob(path).upload_from_string(empty, content_type="application/json")
+        return
```
mikewilli, line 161: Should we log a warning or something before returning? This seems like a "something is broken" scenario. I assume we're tracking this separately though, so maybe not worth adding more noise here.

sql/moz-fx-data-experiments/monitoring/experiment_enrollment_funnel_v1/schema.yaml
```diff
@@ -0,0 +1,29 @@
...
+    description: Enrollment status (Enrolled, NotEnrolled, Disqualified, WasEnrolled).
+  - name: reason
+    type: STRING
+    mode: NULLABLE
+    description: Reason for the status (Qualified, NotTargeted, NotSelected, FeatureConflict, OptOut, etc.).
+  - name: conflict_slug
+    type: STRING
+    mode: NULLABLE
+    description: Slug of the recipe holding the feature slot when reason is FeatureConflict. NULL on iOS/Fenix until SDK fix lands.
+  - name: clients
```
mikewilli: Might be helpful to give this a name that indicates that it is a count, like `client_count` or similar?

sql/moz-fx-data-experiments/monitoring/experiment_enrollment_funnel_v1/query.py
mikewilli: I'll add that I'm fine to defer to you and @jaredlockhart on this if you think `client_id` makes more sense, I just know that desktop defaults to using the `profile_group_id` now (as of about a year ago).

mikewilli (commented):
Looks good overall, just a couple small comments to double-check before merging. I'd give this a r+wc but since it's marked for automerge I'll just leave a comment for now.

sql/moz-fx-data-experiments/monitoring/experiment_enrollment_funnel_v1/query.py
yashikakhurana: @mikewilli Tried t.client_info.profile_group_id but it doesn't exist in the client_info struct on nimbus_targeting_context — getting "Field name profile_group_id does not exist in STRUCT<android_sdk_version STRING, app_build STRING, app_channel STRING, ...>".

mikewilli: Hm, I don't see the `profile_group_id` on this table, only a [UUID metric](https://mozilla.acryl.io/dataset/urn:li:dataset:(urn:li:dataPlatform:bigquery,mozdata.firefox_desktop.nimbus_targeting_context,PROD)/Columns?highlightedPath=%5Bversion%3D2.0%5D.%5Btype%3Dstruct%5D.%5Btype%3Dstruct%5D.metrics.%5Btype%3Dstruct%5D.uuid.%5Btype%3Dstring%5D.legacy_telemetry_profile_group_id&is_lineage_mode=false&schemaFilter=) `legacy_telemetry_profile_group_id`, and I'm not sure if that's what you want.

I think the options are:
1. add `profile_group_id` to this table
2. just use `client_id` and not worry about the small discrepancy

yashikakhurana: I am okay with using the client id, but let me know what you think is best?

jaredlockhart (approved):
Yes! Awesome @yashikakhurana thank you 🙏 

Yes to confirm, `client_id` is fine here, the distinction between client and profile group here isn't super important, what's more important is the drop offs as you move through the funnel, which is the same in either case.

So yes looks great let's land it ty @yashikakhurana 🙏 🎉

mikewilli (approved):

PR #9385, opened 2026-05-19
docs: drop stale public-org-membership note for fork PRs

akkomar:
The cookbook still described the per-job author_association split where public Mozilla org members got the `dev` environment for build/lint jobs and `external-fork` only for integration tests. That branching was removed by #9353 (Fork PR approval gate) in favor of a single per-SHA approval in fork-gate.yml that gates every job for every fork PR.


alekhyamoz (approved):
lgtm

PR #9387, opened 2026-05-19
Create widget client table

jaredsnyder:
Create rollup of the widget visit-day table to the client-day table.  It was modeled on `newtab_clients_daily` and creates dimensions the same way.

Validation doc: https://docs.google.com/document/d/1d4H1JegVWJ4EhvAvVd95VD0WftS8nozdD6xitx81ji0/edit?tab=t.0


sql/moz-fx-data-shared-prod/firefox_desktop_derived/widgets_client_daily_v1/query.sql
```diff
@@ -0,0 +1,123 @@
...
+    client_id,
+    widget_name,
+    widget_size
+),
+client_metrics AS (
+  SELECT
+    submission_date,
+    client_id,
+    widget_name,
+    widget_size,
```
land-edi, line 42: I am wondering if we should apply the mode_last function to the size as well. It is likely that a user with multiple visits can have different size for the same widget. And this might break the granularity of 1 row per day, user, widget.

Everything else looks good!

jaredsnyder: I wanted to keep widget_size as something that got grouped by to enable calculations like CTR by widget size

But you are right it does lead to some counter-intuitive results.  I tried to call them out in the schema.yaml with qualifiers on the impressions field.  There are also some examples in the validation doc

gkatre (commented):
Following up on our chat earlier today: if we are interested in analyzing new client behavior and trends, we could add a LEFT JOIN with `moz-fx-data-shared-prod.firefox_desktop_derived.baseline_clients_first_seen_v1` to bring in the `first_seen_date` field. That would let us identify whether a client is new.

land-edi (approved):

land-edi (approved):
Let's 🚢

PR #9367, opened 2026-05-14
feat: initiate backfill for firefox_desktop nimbus feature monitoring tables

yashikakhurana:
## Summary

- Adds `backfill.yaml` for all 24 `firefox_desktop_derived` nimbus feature monitoring tables
- Backfill range: 2025-11-14 to 2026-05-10 (initial 6-month backfill)
- All tables are newly generated from `sql_generators/nimbus_feature_monitoring`

## Tables

All `firefox_desktop_derived.nimbus_feature_monitoring_*_v1` tables:
- `address_autofill_feature`, `infobar`, `aboutwelcome`, `feature_callout`, `fxms_bmb_button`
- `fxms_message_1/7/16/20/21/22`, `background_task_message`, `preonboarding`
- `newtab_trainhop_addon`, `newtab_sponsored_content`, `new_tab_sections_experiment`
- `newtab_trainhop`, `pocket_newtab`, `newtab_private_ping`, `newtab_merino_ohttp`
- `newtab_ad_sizing_experiment`, `newtab_promo_card`, `newtab_inferred_personalization`
- `urlbar`

Ref: [6753](https://mozilla-hub.atlassian.net/browse/EXP-6753 )

PR #9388, opened 2026-05-19
feat(nimbus): move feature monitoring tables to nimbus_feature_monitoring dataset

yashikakhurana:
## Summary

Per @mikewilli's suggestion in #9367, consolidates all feature monitoring tables under a single shared dataset instead of per-app `_derived` datasets, and shortens table names by dropping the redundant `nimbus_feature_monitoring_` prefix.

**Before:**
- `moz-fx-data-shared-prod.firefox_desktop_derived.nimbus_feature_monitoring_urlbar_v1`

**After:**
- `moz-fx-data-shared-prod.nimbus_feature_monitoring.urlbar_v1`

Same change applies to all feature monitoring tables across all apps.

### DIFF
--- a/sql_generators/nimbus_feature_monitoring/__init__.py
+++ b/sql_generators/nimbus_feature_monitoring/__init__.py
@@ -206,9 +206,7 @@ def generate_queries(project, path, write_dir):
                                         event_name=metric.pop(
                                             "event_name", metric_name
                                         ),
-                                        source_type=source_tables[
-                                            source_name
-                                        ].type,
+                                        source_type=source_tables[source_name].type,
                                         **metric,
                                     )
                                 )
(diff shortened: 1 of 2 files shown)


sql_generators/nimbus_feature_monitoring/__init__.py
```diff
@@ -240,7 +242,9 @@ def generate_queries(project, path, write_dir):
             }
 
             feature_name_sql = feature.name.replace("-", "_").lower()
-            table = f"nimbus_feature_monitoring_{feature_name_sql}_v1"
+            # Prefix with the app dataset to avoid collisions when multiple apps
+            # define a feature with the same Nimbus slug.
+            table = f"{dataset}_{feature_name_sql}_v1"
```
mikewilli, line 245: Yea makes sense to me. I think this is better than having a dataset per app.

mikewilli (approved):
Makes sense to me. One thing that occurred to me is that the consolidated view could grow quite large and I don't remember if we're doing any clustering or anything on the feature tables? But we can worry about that later if it's slow or otherwise problematic.

PR #9409, opened 2026-05-20
Keep hidden files in stage deploy artifacts

scholtzan:
## Description

Add `include-hidden-files: true` to the `staged-generated-sql` upload artifact step. The `.bqetl_target_info.yaml` manifest files (dotfiles) were silently excluded by upload-artifact v4+, which defaults to omitting hidden files. Without them, the dryrun job can't map stage paths back to the skip list, causing skip-listed queries to fail dryrun in CI. See https://github.com/mozilla/bigquery-etl/actions/runs/26177369598/job/77011766034?pr=9402

Tested in https://github.com/mozilla/bigquery-etl/actions/runs/26179373260/job/77018731680?pr=9408

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -588,7 +588,6 @@ jobs:
         with:
           name: staged-generated-sql
           path: /tmp/workspace/staged-generated-sql
-          include-hidden-files: true
 
   dry-run-sql:
     name: Dry Run SQL
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -588,6 +588,7 @@ jobs:
         with:
           name: staged-generated-sql
           path: /tmp/workspace/staged-generated-sql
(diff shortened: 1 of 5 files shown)


sean-rose:
**question:** Why are those `.bqetl_target_info.yaml` files named with a leading dot?  Could we simply drop the leading dot from the filenames instead?

Including hidden files makes me slightly nervous from a security perspective, as sometimes tools put sensitive things in dot files (e.g. `.env`).  Are we sure GitHub Actions isn't doing anything like that?

chelseyklein (approved):
ty!

scholtzan:
> Why are those .bqetl_target_info.yaml files named with a leading dot? Could we simply drop the leading dot from the filenames instead?

Yeah, that's a fair point. I initially figured since these files are only used by the tooling and are not useful to users of the tool they could be hidden.

PR #9412, opened 2026-05-20
Change merino extract job frequency to every 10 minutes 

rolf-moz:
## Description

We have recently moved the newtab_merino_extract job to ONLY use the newtab_content ping and ignore the newtab ping. This has dramatically improved speed and reduced costs from an average of $100 per day to around $50 a day, despite increased workloads like propensity calculations!

https://mozilla.cloud.looker.com/explore/monitoring/bigquery_usage?qid=BrkdURGGK5h4U8lQkTrmD4&origin_space=1058&toggle=fil,vis

Doc with more background here https://docs.google.com/document/d/1ByJJ3IdKRhiOT4wjjdYTbs8P4MNHfWq6T9bTVIBW1zo/edit?tab=t.0

We would like to use that added 'budget' to increase frequency of the job from every 20 minutes to every 10 minutes. 

This is important because in the US, Canada and soon Germany we're updating content every 30 minutes and want to rank it quickly!

Increasing the frequency will also increase the update interval on Merino.  


mmiermans (approved):
Yay! I bet this will result in higher CTR. 🤞

PR #9441, opened 2026-05-26
fix(bug-2040539): add parameter @submission_date to metadata.yaml and fix partition filter required error

kik-kik:
# fix(bug-2040539): add parameter @submission_date to metadata.yaml and fix partition filter required error

This should address the following error we get when running the job:

```
Query parameter 'submission_date' not found at [5:45]
```

During dry-run also found that partition filter is required on `glean_telemetry_derived.new_profile_churn_clients_v1`, this does not appear to align with the metadata configuration.


sql/moz-fx-data-shared-prod/glean_telemetry_derived/recent_new_profile_churn_clients_v1/query.sql
kik-kik, line 1: @BryanBaird this appears to be a relatively heavy query, are we sure we need it? If we do, it might be worth seeing if any changes can be made to make it a bit more "lightweight".

cc: @alekhyamoz this needs a new owner.

gkabbz (approved):

PR #9445, opened 2026-05-26
[AIMOD-1214] ]Save a daily snapshot of propensity

mlplyler:
## Description

We wrote newtab content propensity by tile position to a table that was replaced daily.

We now want a historical propensity record by date.

Python code now sends to both legacy and historical (by date) table 

* [AIMOD-1214]

[AIMOD-1214]: https://mozilla-hub.atlassian.net/browse/AIMOD-1214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rolf-moz:
Thinking about this, one way we could update it to be more consistent with the data team design patterns
-Create a parallel v2 directory that has the date column - it may have duplicate query.py code
-Have the query.py take a date argument
-Structure yml files so that we could support backfill type jobs in the future  so we can be assured every date is populated.

Abandon the _v1 - keep running for several weeks then turn off

sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_propensity_v2/schema.yaml
```diff
@@ -0,0 +1,22 @@
+fields:
+- mode: NULLABLE
+  name: snapshot_date
+  type: DATE
+- mode: NULLABLE
+  name: snapshot_at
+  type: TIMESTAMP
+- mode: NULLABLE
```
rolf-moz, line 14: we may also want to add a string column for layout at the same time

BenWu:
+1 on @rolf-moz's suggestions. Having a single query.py write to multiple tables is potentially confusing when we triage issues since it isn't standard behaviour. I think you would need a query.py in `newtab_merino_propensity_history_v1/` anyway in order for it to be deployed as part of the deployment process.

rolf-moz:
@mlplyler I think you need to rebase.

One requested change -- Can you add a country column? If this initial analysis is US only then we can mark it US otherwise it should be marked Null.

Then we're good and I think I can trigger the CI

mlplyler:
> @mlplyler I think you need to rebase.
> 
> One requested change -- Can you add a country column? If this initial analysis is US only then we can mark it US otherwise it should be marked Null.
> 
> Then we're good and I think I can trigger the CI

The query I copied has a hardcoded "US", so yes US only for now, but I added the country column.

rolf-moz (approved):

PR #9448, opened 2026-05-26
backfill widget_client_daily 

jaredsnyder:
add yaml to initiate backfill


sql/moz-fx-data-shared-prod/firefox_desktop_derived/widgets_client_daily_v1/backfill.yaml
```diff
@@ -0,0 +1,11 @@
+2026-05-26:
+  start_date: 2026-04-01
```
land-edi, line 2: Widgets were introduced in March, can this be 03/01 instead?

jaredsnyder: It could be, but the visits table only goes back to 04/01 so you'd need to backfill March in the visits table first.  You could use this PR as a template for doing that!

land-edi (approved):

PR #9462, opened 2026-05-28
chore: Add backstage-gke wg to syndicate datasets

quiiver:
## Description
According to a comment on MZCLD-3137 Authorizing the public views is not sufficient and the underlying syndicated datasets also need authorization

* MZCLD-3137

### DIFF
--- a/sql/moz-fx-data-shared-prod/mozcloud_derived/dataset_metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/mozcloud_derived/dataset_metadata.yaml
@@ -7,3 +7,4 @@ workgroup_access:
 - role: roles/bigquery.dataViewer
   members:
   - workgroup:mozilla-confidential/data-viewers
+  - workgroup:platform/backstage-gke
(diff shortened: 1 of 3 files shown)


whd (commented):
> According to a comment on MZCLD-3137 Authorizing the public views

To be clear about terminology, "authorizing" here is not the same as https://docs.cloud.google.com/bigquery/docs/authorized-views or https://docs.cloud.google.com/bigquery/docs/authorized-datasets: it's actually the lack of authorized views in the public dataset that makes this PR necessary. If those views were instead authorized this PR would not be necessary, so another way to achieve the same result is to add `authorized: true` metadata to all the existing user-facing views.

BenWu (commented):
Did you want to give it permission to [mozcloud_derived](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/mozcloud_derived/dataset_metadata.yaml) as well for the `service_uptime_v1` table?  Just double-checking you don't need another PR

quiiver:
@BenWu actually that's not a bad idea. thank you

quiiver:
@whd I just saw your comment. is the preferred action here to authorize the views or to add the workgroup access to the underlying datasets?

BenWu (approved):
IMO if the service account is expected to have access to all the tables in the syndicated dataset, then this is better than authorizing each view since you don't need to remember to apply it to each one. Also deployment of normal views is easier to track than authorized views.

PR #9384, opened 2026-05-18
feat(DENG-11078): ingest statcounter data

kbammarito:
## Description

This PR ingests daily [Statcounter](https://gs.statcounter.com/browser-market-share) data for:
- mobile and desktop
- worldwide and by regions

## Related Tickets & Documents
* DENG-11078

**Reviewer, please follow [this checklist](https://github.com/mozilla/bigquery-etl/blob/main/.github/reviewer_checklist.md)**

PR #9465, opened 2026-05-29
fix(DENG-11078): preserve --parameter flags for python-query tables in bqetl check run

kbammarito:
## Description

This PR fixes `bqetl check run` dropping `--parameter=*` flags for Python-query tables, which caused Airflow `bigquery_dq_check` tasks to fail with `Query parameter 'date' not found` whenever a `checks.sql` macro emits `@date` (or any other parameterized reference).

**Testing:** I ran `./bqetl check run moz-fx-data-shared-prod.statcounter_derived.browser_market_share_worldwide_v1 --marker=warn --parameter=date:DATE:2026-05-20 ; echo "exit=$?"` before the fix to verify that it first failed with the same error and then succeeded afterward.

* DENG-11078
* https://github.com/mozilla/bigquery-etl/pull/9384

### DIFF
--- a/bigquery_etl/cli/check.py
+++ b/bigquery_etl/cli/check.py
@@ -205,9 +205,10 @@ def run(ctx, dataset, project_id, sql_dir, marker, dry_run):
     for checks_file, project_id, dataset_id, table in paths_matching_checks_pattern(
         dataset, sql_dir, project_id=project_id
     ):
-        query_args = (
-            [] if (Path(checks_file).parent / "query.py").exists() else ctx.args
-        )
+        if (Path(checks_file).parent / "query.py").exists():
+            query_args = [a for a in ctx.args if a.startswith("--parameter=")]
+        else:
+            query_args = ctx.args
         _run_check(
             checks_file,
             project_id,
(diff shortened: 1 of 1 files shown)


kik-kik:
@kbammarito it sounds like this should resolve your specific issue. Did you have a chance to make sure this will not break execution of existing bq checks. For example, bqetl checks against query.py files that do not require any parameters, and query.sql files. Would want to be on the safe side here so that we do not end up breaking our ETL pipelines.

BenWu (approved):
To check the impact of this, these are the query.py that have checks.sql:
- `sql/moz-fx-data-shared-prod/external_derived/gdp_v1`
- `sql/moz-fx-data-shared-prod/external_derived/imf_exchange_rates_v1`
- `sql/moz-fx-data-shared-prod/external_derived/monthly_inflation_v1`
- `sql/moz-fx-data-shared-prod/external_derived/population_v1`
- `sql/moz-fx-data-shared-prod/external_derived/quarterly_inflation_v1`
- `sql/moz-fx-data-shared-prod/statcounter_derived/browser_market_share_regions_v1`
- `sql/moz-fx-data-shared-prod/statcounter_derived/browser_market_share_worldwide_v1`
- `sql/moz-fx-data-shared-prod/telemetry_derived/newtab_merino_propensity_v1`

`row_count_within_past_partitions_avg` is the only check template that uses bigquery parameters, and `browser_market_share_regions_v1` and  `browser_market_share_worldwide_v1` are currently the only one that use it. So this shouldn't affect anything else right now.  But I think the solution is correct anyway so I'm not too worried about the impact

PR #9479, opened 2026-06-02
feat(DENG-11185): Kitsune weekly contributor metrics

phil-lee70:
## Description

Breaks the `kitsune_weekly_contributors.sql` dashboard query into a maintainable, scheduled base-table → metrics-table → view stack in `sumo_metrics_derived`, following the existing `*_base_v1` / `*_daily_v1` / `*_kpis` pattern in that dataset.

## What's added

`kitsune_contributor_answers_base_v1` — answer-level fact table (one row per qualifying answer: non-spam, active user, Firefox product, non-OP-self-answer). Materializes the dashboard query's `main` CTE. Sourced from `sumo_syndicate.*`.

`kitsune_contributor_metrics_weekly_v1` — weekly contributor metrics (weekly/new contributors, top-contributor %, retention rate by tier, each with 4-week moving averages and lags), built on the base table.

`sumo_contributor_kpis` — thin view exposing curated weekly KPI columns for the dashboard layer.
Both tables are incremental: false (full daily rebuild) on the existing bqetl_sumo_metrics DAG.

## Key design decisions

**Weekly grain for the metrics table** — the measures are non-additive distinct counts over weekly/trailing-4-week windows, so they can't be reconstructed from daily aggregates via a view.

**Full-history weeks** — the weeks series is anchored to `MIN(answer_date)` instead of the dashboard's trailing `CURRENT_DATE - 1 YEAR`, so each full refresh reproduces the entire time series.

**Dropped** `CURRENT_DATE()`-relative flags (`is_newbie`, `days_since_last_answer`, etc.) — "as of now" snapshots that don't belong in a historical fact and aren't used downstream.

* DENG-11185

### DIFF
--- a/sql/moz-fx-data-shared-prod/sumo_metrics_derived/kitsune_contributor_answers_base_v1/bigconfig.yml
+++ b/sql/moz-fx-data-shared-prod/sumo_metrics_derived/kitsune_contributor_answers_base_v1/bigconfig.yml
@@ -1,0 +1,24 @@
+type: BIGCONFIG_FILE
+tag_deployments:
+- collection:
+    name: Operational Checks
+    notification_channels:
+    - slack: '#de-bigeye-triage'
+  deployments:
+  - column_selectors:
+    - name: moz-fx-data-shared-prod.moz-fx-data-shared-prod.sumo_metrics_derived.kitsune_contributor_answers_base_v1.*
+    metrics:
+    - metric_type:
+        type: PREDEFINED
+        predefined_metric: FRESHNESS
+      metric_name: freshness
+      metric_schedule:
+        named_schedule:
+          name: Default Schedule - 13:00 UTC
+    - metric_type:
+        type: PREDEFINED
+        predefined_metric: VOLUME
+      metric_name: volume
+      metric_schedule:
+        named_schedule:
+          name: Default Schedule - 13:00 UTC
--- a/sql/moz-fx-data-shared-prod/sumo_metrics_derived/kitsune_contributor_answers_base_v1/metadata.yaml
+++ b/sql/moz-fx-data-shared-prod/sumo_metrics_derived/kitsune_contributor_answers_base_v1/metadata.yaml
@@ -1,0 +1,43 @@
+friendly_name: Kitsune Contributor Answers Base
+description: |
+  Answer-level fact table for the SUMO (Kitsune) support forum, with one row per
+  qualifying answer used to measure contributor activity.
+
+  Grain: Answer
+
+  Each row is a single answer that satisfies all of the following (matching the
+  `main` CTE of the kitsune_weekly_contributors dashboard query):
+  - Not spam (kitsune_answers_raw.is_spam IS NOT TRUE)
+  - Posted by an active user (kitsune_auth_user.is_active = TRUE)
+  - On a Firefox product (firefox, mobile, ios, firefox-enterprise)
+  - Not an OP self-answer (answer creator <> question creator)
+
+  Product mapping:
+    firefox            -> Firefox desktop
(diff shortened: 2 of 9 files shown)


phil-lee70:
## Validation results — equivalence vs. original dashboard query

Both new tables were validated against the original `kitsune_weekly_contributors.sql` dashboard query. Both layers reproduce the original exactly. Run date: 2026-06-02.

### Base table — `kitsune_contributor_answers_base_v1` vs. the `main` CTE

| Check | Result |
|-------|--------|
| Row counts | ✅ Match — both 801,749 rows |
| Grain (`answer_id`) | ✅ Unique — 801,749 distinct, 0 duplicates |
| Exact diff (identity columns) | ✅ 0 rows differ in either direction |

Identity columns compared: `answer_id, answer_date, creator_username, question_id, question_date, product, locale, num_replies_from_others, op_reply_back`. Time-varying columns (votes, logins, last_reply) and presentation columns (URLs, email) were excluded since they change between ETL runs. Confirms the `sumo_syndicate.*` sources are faithful mirrors of `moz-fx-sumo-prod.sumo.*`, and that the refactored OP-self-answer exclusion is behaviorally identical.

### Weekly table — `kitsune_contributor_metrics_weekly_v1` vs. the dashboard final output

Compared over 53 overlapping weeks (2025-06-02 → 2026-06-01).

| Group | Result |
|-------|--------|
| Raw per-week metrics | ✅ 0 of 53 weeks differ — exact match |
| Windowed metrics (`_4w_ma`, `lagged_*`) | ✅ 46 of 53 match; differ only on the first 7 weeks of the dashboard window |

Raw metrics: `weekly_contributor_count, new_contributors, top_contributor_percentage, retention_rate, retention_rate_top, retention_rate_regular`.

The 7 edge-week differences are an **expected boundary artifact**, not a logic discrepancy: the dashboard's `weeks` array starts at `CURRENT_DATE - 1 YEAR` with no preceding weeks, so its 4-week moving averages and lags are truncated at the start. The new table is built on full history, so its values for those weeks are the more correct of the two. The raw metrics on those same 7 weeks still match exactly; only the derived MA/lag columns differ, and they converge to identical values from week 8 onward.

**Conclusion:** the migration is sound — both the base table and the weekly metrics table reproduce the original query, with the only differences being expected and improved behavior at the window edge.

phil-lee70:
@kpham-mozilla RE: In this PR the relative flags (is_newbie, days_since_last_answer, etc.) being dropped from the base table - just needed to confirm what the intention of these flags are:

1. "As of the answer" (point-in-time, stable) — put it in the base table

If you want "was this contributor a newbie when they posted this answer," recompute relative to answer_date instead of CURRENT_DATE():

date_joined > DATE_SUB(answer_date, INTERVAL 1 YEAR) AS is_newbie_at_answer

This is immutable, stable across refreshes, and is the correct way to carry such a flag in a historical fact. is_newbie / is_returning translate cleanly this way.

2. "As of now" (current snapshot, intentionally volatile) — keep it out of the base table

login_last_week and days_since_last_answer are inherently "as of today" recency signals — they only make sense as current state. 

Recommendation:

is_newbie / is_returning → add to the base fact as *_at_answer (point-in-time). Stable, validates cleanly.
login_last_week / days_since_last_answer → expose in a view (or contributor dimension), never materialized into the base table.

kpham-mozilla:
@phil-lee70  you can actually drop the relative flags (is_newbie, days_since_last_answer, etc.) from the base tables as they are not being used in the CXOR dashboard. Those are remnants of the original query pulled from the SUMO Heartbeat Dashboard

kpham-mozilla (approved):

PR #9480, opened 2026-06-02
fix(DENG-11708): fix month parsing, update error handling

kbammarito:
## Description

This PR checks for both full and abbreviated months, updates cleanup error handling and continues to next date even if prior date fails.

* DENG-11078


sql/moz-fx-data-shared-prod/statcounter_derived/pipeline.py
chelseyklein:
```suggestion
       ValueError: If the percent column is missing, if the date string
           cannot be parsed as either abbreviated or full month format,
           or if the embedded date does not match partition_date.
```

chelseyklein (approved):
Not blocking...could be a future enhancement...but it'd be good to have tests on the statcounter pipeline. the month-parsing logic could regress...a unit test for `rename_columns` with duplicate date headers would be valuable; as would a test verifying an unparseable month raises `ValueError`; plus the error-isolation logic

PR #9481, opened 2026-06-02
feat(DENG-11189): Remove conversion event query code

kik-kik:
# feat(DENG-11189): Remove conversion event query code


scottbedwell (approved):
Reviewed and this looks great to me.

PR #9494, opened 2026-06-03
Add doc for supported/unsupported backfill operations

BenWu:
## Description

This is hard to remember and claude sometimes doesn't get the right answer for some of these so explicitly listing things out will help. Also moved the backfilling doc to a separate file as it was getting long


akkomar (approved):
🎉

PR #9503, opened 2026-06-03
DENG-11204 - Productionizing Dataset Profiling Table

gkatre:
#### ***Updated to a single table design ***

Adds a weekly job that profiles columns of one or more source datasets and writes the stats to a single partitioned table, exposed through an unversioned view for the schema_enricher tooling to read.

## What's added
```
sql/moz-fx-data-shared-prod/data_governance_metadata_derived/
├── column_profiles_v1/          # the job + destination table
│   ├── query.py                 # self-contained profiler (google-cloud-bigquery + stdlib only)
│   ├── metadata.yaml            # weekly scheduling, partitioning, clustering, 90-day retention
│   └── schema.yaml              # 12-column schema (profiled_at DATE partition key)
└── column_profiles/             # the read surface
    ├── view.sql                 # unversioned passthrough view over column_profiles_v1
    ├── metadata.yaml
    ├── schema.yaml              # column descriptions for the view
    └── README.md                # architecture, query/pruning guidance, runbook
dags.yaml                        # new bqetl_data_governance_metadata DAG (Mon 04:00 UTC)
```

## Design notes

- **Single table, multi-dataset:** One weekly task profiles every base table across `--source-datasets`. Keeps reruns/backfills idempotent (`profiled_at` based partitioning).
   - Table maintenance: Partition retention is currently set at 90 days. Based on weekly runs the max partitions are expected to be around ~13
- **Unversioned view `column_profiles`:** Conforms to `bigquery-etl` convention. 
- **Reproducibility:** for a given --date, both the scanned source window (anchored to --date − 7 days) and the 1% sample bucket (seeded from --date+table) are deterministic — so reruns and backfills of a date reproduce the same data in the same partition.
- **Partition pruning through the view:** consumers should always filter on profiled_at. Because the latest-snapshot read doesn't otherwise constrain the partition, use a recent-window floor so BigQuery prunes underlying table.
example: `WHERE DATE(profiled_at) >= CURRENT_DATE() - INTERVAL 14 DAY`
- The 14-day window would contain the most recent weekly run and bounds the scan to ~1–2 partitions (based on weekly runs). 
- Details and query examples in the README (sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles/README.md#querying-the-view-column_profiles).

> [!NOTE]
> - **Adding/Changing dataset:** Datasets are a job parameter and can be set/altered in airflow job runs manually or updated in `../column_profiles_v1/metadata.yaml` for automated runs
> - **Profiling a subset of tables:** To profile a few tables within a dataset can be done manually on the command-line using `bqetl` CLI.
> - These details are explained in the README (sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles/README.md)

* DENG-11204


sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles/view.sql
```diff
@@ -0,0 +1,15 @@
+CREATE OR REPLACE VIEW
+  `moz-fx-data-shared-prod.data_governance_metadata_derived.column_profiles`
+AS
+SELECT
+  *
+FROM
+  `moz-fx-data-shared-prod.data_governance_metadata_derived.column_profiles_telemetry_derived_v1`
```
lucia-vargas-a: This will silently omit tables from the view if the developer forgets to add it. I recommend a wildcard so that new tables are automatically covered.

```suggestion
  `moz-fx-data-shared-prod.data_governance_metadata_derived.column_profiles_*`
```

lucia-vargas-a: **Note:** This suggestion will include all tables with the prefix so we need to ensure naming consistency and that test tables are not included.

sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles_v1/query.py
```diff
@@ -0,0 +1,730 @@
...
+    """Execute the profiling query and return (column_stats_dict, gb_scanned)."""
+    partition_filter = get_partition_filter(client, table)
+    query = build_profile_query(table, columns, partition_filter)
+
+    job = client.query(query)
+    rows = list(job.result())
+    gb_scanned = job.total_bytes_processed / 1e9
+
+    row = rows[0]
+    total_rows = row.total_rows
```
lucia-vargas-a, line 465: Should we simply write zero when the table has rows, but the scanned 1 week-old partition + 1% sample doesn't?
Or even better if we scan the lattest partition with data, instead of the last 7 days?

```suggestion
    total_rows = row.total_rows
    if total_rows == 0:
      return {}, gb_scanned
```

lucia-vargas-a (commented):
A few suggestions, mostly for precision

gkabbz:
Issue (blocking): @gkatre It looks like the design you are taking is separate table for each dataset -> gets unioned in a view. I think the design should be a single table that has all the profiled tables. We can still manage each dataset as a task but we'd pass the source dataset as an argument to a single job that writes to one shared table. For idempotency on reruns, we'd need to delete the rows (WHERE profiled_at = @date AND source_dataset = X) before writing, which avoids partition contention without needing separate tables.                                                            
                                                                                                                   
The reason I prefer this design is that the union view has a manual step. Every new dataset requires a view update. So either way we'll have a manual step so I'd rather it be on the deleting since that won't happen regularly vs on the adding. If it gets missed on adding, cross-dataset column lookups silently return incomplete results, which is hard to catch downstream.

sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles_v1/query.py
gkatre: @lucia-vargas-a good call! I will implement the change in the new version. Thanks!

sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles/view.sql
gkatre: This is going to change in the new version - just one table for multiple datasets as discussed today morning

lucia-vargas-a (approved):

PR #9595, opened 2026-06-17
chore: move firefox install to final run command installing dependencies

kik-kik:
# chore: move firefox install to final run command installing dependencies

It seems now when building the image firefox does not always get installed as intended. Moving the install command to the final run command installing dependencies. Locally this seems to be working more consistently now.


scottbedwell (approved):
lgtm

scottbedwell (approved):
lgtm

PR #9614, opened 2026-06-22
fix(data_governance_metadata): profile tables with all-NULL sample_id

akkomar:
column_profiles_v1 takes a 1% sample via `WHERE sample_id = <bucket>` whenever a table has a `sample_id` column. Tables sourced from server-side Glean pings carry no `client_id` so their `sample_id` is entirely NULL: the sample matches zero rows and the table profiles to zero columns with only a "Scanned slice empty" info log.

Changed in this PR:
Detect whether sample_id is populated and fall back to `TABLESAMPLE SYSTEM (1 PERCENT)` when it is not. The sampling decision moves out of build_profile_query / build_nested_profile_queries (which now take an explicit tablesample flag) into _run_profile_query.

* https://mozilla-hub.atlassian.net/browse/DENG-11270

### DIFF
--- a/sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles_v1/query.py
+++ b/sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles_v1/query.py
@@ -236,7 +236,7 @@ def _sample_bucket(run_date: str, table: str) -> int:
     Seeding by run_date makes a given --date reproducible; including the table
     name spreads buckets across tables instead of always sampling the same one.
     """
-    return random.Random(f"{run_date}:{table}").randint(1, 99)
+    return random.Random(f"{run_date}:{table}").randint(0, 99)
 
 
 def build_profile_query(
--- a/sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles_v1/query.py
+++ b/sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles_v1/query.py
@@ -14,6 +14,7 @@ The profiling logic is vendored verbatim from
 ``google.cloud.bigquery`` plus the Python standard library — the bigquery-etl
 runner image installs dependencies with ``--no-deps``.
 """
+
 import argparse
 import datetime
 import json
@@ -243,19 +244,20 @@ def build_profile_query(
     columns: list[tuple[str, str, str]],
     partition_filter: str | None = None,
     sample_id: int | None = None,
+    tablesample: bool = False,
 ) -> str:
     """Build a single query that profiles every profilable column in one pass.
 
     Profiles scalar and leaf tiers only. nested_leaf fields are profiled
     separately via build_nested_profile_queries (UNNEST-based).
 
-    sample_id selects the 1% bucket for tables with a sample_id column; the
-    caller passes a value derived deterministically from the run date so a
-    given --date always samples the same bucket.
+    Sampling is decided by the caller: a non-None sample_id adds a
+    ``sample_id = <bucket>`` filter (deterministic per run date), while
+    tablesample adds TABLESAMPLE SYSTEM (1 PERCENT) for tables whose sample_id
+    column is unpopulated. The two are mutually exclusive.
     """
     excluded_tiers = {"undocumented", "pii_suppressed", "nested_leaf", "scalar_array"}
     profilable = [(fp, dt) for fp, dt, tier in columns if tier not in excluded_tiers]
-    has_sample_id = any(fp == "sample_id" for fp, _ in profilable)
 
     col_stats = []
     for field_path, data_type in profilable:
@@ -273,15 +275,18 @@ def build_profile_query(
 
     select = ",\n        ".join(["COUNT(*) AS total_rows"] + col_stats)
 
+    from_clause = f"`{table}`"
+    if tablesample:
+        from_clause += " TABLESAMPLE SYSTEM (1 PERCENT)"
+
     where_clauses = []
     if partition_filter:
         where_clauses.append(partition_filter)
-    if has_sample_id:
-        bucket = sample_id if sample_id is not None else random.randint(1, 99)
-        where_clauses.append(f"`sample_id` = {bucket}")
+    if sample_id is not None:
+        where_clauses.append(f"`sample_id` = {sample_id}")
 
     where = f"WHERE {' AND '.join(where_clauses)}" if where_clauses else ""
-    return f"SELECT {select} FROM `{table}` {where}"
+    return f"SELECT {select} FROM {from_clause} {where}"
 
 
 def build_nested_profile_queries(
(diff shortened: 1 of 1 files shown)


scholtzan:
### Integration report
- [View full diff, if available (SQL + DAGs)](https://github.com/mozilla/private-bigquery-etl/blob/bqetl-diff/pr-9614/sql.diff) *(requires private-bigquery-etl access)*
- [Private CI run](https://github.com/mozilla/private-bigquery-etl/actions/runs/27947220773)

sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles_v1/query.py
```diff
@@ -273,28 +275,35 @@ def build_profile_query(
 
     select = ",\n        ".join(["COUNT(*) AS total_rows"] + col_stats)
 
+    from_clause = f"`{table}`"
+    if tablesample:
+        from_clause += " TABLESAMPLE SYSTEM (1 PERCENT)"
```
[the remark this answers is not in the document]
akkomar: Partition pruning / cost issue doesn't reproduce.

Block-level all-or-nothing on small tables: good catch, but not a regression. To make it deterministic I added a fallback in a follow-up commit: if the sampled scan returns 0 rows, retry the partition unsampled.

scholtzan:
### Integration report
- [View full diff, if available (SQL + DAGs)](https://github.com/mozilla/private-bigquery-etl/blob/bqetl-diff/pr-9614/sql.diff) *(requires private-bigquery-etl access)*
- [Private CI run](https://github.com/mozilla/private-bigquery-etl/actions/runs/27960501333)

sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles_v1/query.py
```diff
@@ -333,17 +338,20 @@ def build_nested_profile_queries(
 
         select = ",\n        ".join(["COUNT(*) AS total_rows"] + col_stats)
 
+        from_clause = f"`{table}`"
+        if tablesample:
+            from_clause += " TABLESAMPLE SYSTEM (1 PERCENT)"
```
kik-kik, line 343: suggestion: would it make sense for tablesample to be an int that specifies the percent value used in the TABLESAMPLE SYSTEM clause?

sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles_v1/query.py
```diff
@@ -372,6 +380,29 @@ def get_partition_filter(
...
+    match no rows and the table would profile to zero columns, so the caller
+    uses this to decide whether to fall back to TABLESAMPLE. Checks the same
+    slice the profiler scans (the partition_filter), short-circuiting via
+    LIMIT 1 so the common populated case stays cheap.
+    """
+    where_clauses = ["`sample_id` IS NOT NULL"]
+    if partition_filter:
+        where_clauses.append(partition_filter)
+    query = (
+        f"SELECT EXISTS(SELECT 1 FROM `{table}` "
```
kik-kik, line 400: question: just want to make sure I understand this correctly. Basically this will return 1 for any records that is returned and then as long as there is at least one record with 1 we will return True here to indicate that we are good?

Would would this still not be a case if some rows had sample_id of `null`?

sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles_v1/query.py
akkomar: No, it's a binary choice between bucket sampling and block sampling. Both fixed at 1% so the fallback doesn't change the sample size, and there's no caller that needs a different percent.

sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles_v1/query.py
akkomar: Correct. It's "any populated," not "all populated". But sample_id is derived from client_id so a mixed table doesn't occur.

sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles_v1/query.py
gkatre: For `sample_id` (when value is present) the sampling mechanism is 1% (values 0-99). If the `TABLESAMPLE` fallback is set as a 1% constant to match, it may be ok to set `tablesample` as a boolean. Is there a reason to make it variable?

Also, something to note (not major): The `TABLESAMPLE SYSTEM (1 PERCENT)` would return a random 1% and may not be a deterministic sample between this nested and the main query (`build_profile_query`)

akkomar: It IS a boolean.

sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles_v1/query.py
gkatre: Correct.

But this led me to think about another scenario when table (or partition) is small and materializes into only a few sample_ids. In this situation, the bucket that gets selected could be different from what is actually in the table (a situation similar to some or very few sample_id populated).

sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles_v1/query.py
gkatre:
```suggestion
    return random.Random(f"{run_date}:{table}").randint(0, 99)
```
Perhaps a minor fix

gkatre (approved):
@akkomar Thanks for the fixes and updates!

sql/moz-fx-data-shared-prod/data_governance_metadata_derived/column_profiles_v1/query.py
akkomar: Yeah, but that could only really theoretically happen in non-telemetry tables with some special sample_id semantics.

scholtzan:
### Integration report
- [View full diff, if available (SQL + DAGs)](https://github.com/mozilla/private-bigquery-etl/blob/bqetl-diff/pr-9614/sql.diff) *(requires private-bigquery-etl access)*
- [Private CI run](https://github.com/mozilla/private-bigquery-etl/actions/runs/27981011619)

scholtzan:
### Integration report
- [View full diff, if available (SQL + DAGs)](https://github.com/mozilla/private-bigquery-etl/blob/bqetl-diff/pr-9614/sql.diff) *(requires private-bigquery-etl access)*
- [Private CI run](https://github.com/mozilla/private-bigquery-etl/actions/runs/27981289251)

PR #9698, opened 2026-07-14
feat(customer_experience): add authorship and engagement columns to kitsune_retrieval_index

kpham-mozilla:
## Description

Adds six new output columns to the Kitsune retrieval index
(`customer_experience_derived.kitsune_retrieval_index_v1` and the downstream
`customer_experience.kitsune_retrieval_index` view) so consumers can see
question/answer authorship and engagement signals.

### New columns

**Authorship** (from `creator_username`)
- `question_creator` (STRING) — username of the question author
- `answer_creator` (STRING) — username of the answer author; NULL when no answer

**Engagement** (from `sumo_syndicate.kitsune_questions_plus`)
- `page_views` (INTEGER) — page views the question has received
- `is_solved` (BOOLEAN) — whether the question is marked solved
- `is_locked` (BOOLEAN) — whether the thread is locked (the query filters
  `is_locked = FALSE`, so this is effectively always FALSE for stored rows)
- `num_votes` (INTEGER) — "me too" / interest votes on the question

### Changes
- `kitsune_retrieval_index_v1/query.sql` — select the six new columns.
- `kitsune_retrieval_index_v1/schema.yaml` — declare the new fields.
- `kitsune_retrieval_index/schema.yaml` (view) — declare the new fields; the view
  is `SELECT * EXCEPT(product)` so it inherits the columns with no view.sql change.

### Notes
- Local `bqetl query validate` dry-run was not run (expired GCP
  application-default credentials); CI runs the same dry-run + schema check.
- Assumed column types are best-effort; if the CI dry-run reports different types
  for any of the six, the schema.yaml files will need a quick adjustment.

### DIFF
--- a/sql/moz-fx-data-shared-prod/customer_experience/kitsune_retrieval_index/schema.yaml
+++ b/sql/moz-fx-data-shared-prod/customer_experience/kitsune_retrieval_index/schema.yaml
@@ -7,6 +7,14 @@ fields:
   type: INTEGER
   mode: NULLABLE
   description: Unique identifier for the support question in Kitsune (SUMO).
+- name: question_creator
+  type: STRING
+  mode: NULLABLE
+  description: Kitsune (SUMO) username of the user who created the support question.
+- name: answer_creator
+  type: STRING
+  mode: NULLABLE
+  description: Kitsune (SUMO) username of the user who created the answer; NULL when no answer exists.
 - name: answer_id
   type: INTEGER
   mode: NULLABLE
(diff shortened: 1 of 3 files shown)


scholtzan:
### Integration report
- [View full diff, if available (SQL + DAGs)](https://github.com/mozilla/private-bigquery-etl/blob/bqetl-diff/pr-9698/sql.diff) *(requires private-bigquery-etl access)*
- [Private CI run](https://github.com/mozilla/private-bigquery-etl/actions/runs/29327808979)
- Bigeye monitoring dry-run: ✅ plan clean — [view plan](https://github.com/mozilla/private-bigquery-etl/blob/bqetl-bigeye-plan/bqetl-pr-9698/bigeye_plan.yml) *(requires private-bigquery-etl access)*

sql/moz-fx-data-shared-prod/customer_experience/kitsune_retrieval_index/schema.yaml
```diff
@@ -7,6 +7,14 @@ fields:
   type: INTEGER
   mode: NULLABLE
   description: Unique identifier for the support question in Kitsune (SUMO).
+- name: question_creator
+  type: STRING
+  mode: NULLABLE
+  description: Kitsune (SUMO) username of the user who created the support question.
```
lucia-vargas-a, line 13: We are intentionally avoiding user's data (PII in these public datasets, what would be the use cases?

kelimuttu: Hi Lucia,

At the moment, my primary use case for this is to detect duplicate questions. If we have other way to do it, I don't mind not having it.

lucia-vargas-a: Hey! what would be a duplicate - E.g. exact same text or repeated question_id (which I would not expect)

lucia-vargas-a: ...With the definition of duplicates, we deduplicate in the query so that the data is unique and PII free.

kelimuttu: Usually, the same user posted similar things in a short period of time. For example: 
* https://support.mozilla.org/en-US/questions/1591384
* https://support.mozilla.org/en-US/questions/1591387
* https://support.mozilla.org/en-US/questions/1591388

PR #9733, opened 2026-07-27
Add overview to managed backfill doc

BenWu:
## Description

Adding a summary at the top to make it more accessible to newer users

### DIFF
--- a/docs/cookbooks/backfilling_a_table.md
+++ b/docs/cookbooks/backfilling_a_table.md
@@ -2,6 +2,30 @@
 
 **Note** For large sets of data, follow the [recommended practices](https://mozilla.github.io/bigquery-etl/reference/recommended_practices/#backfills) for backfills.
 
+## Overview
+
+A backfill reprocesses historical partitions of a table, for example after a query is fixed or
+upstream data is corrected, so the older data reflects the change. Scheduled runs only produce new
+partitions going forward, so past data has to be rewritten separately.
+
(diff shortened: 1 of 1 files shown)


docs/cookbooks/backfilling_a_table.md
```diff
@@ -2,6 +2,29 @@
...
+
+A backfill reprocesses historical partitions of a table, for example after a query is fixed or
+upstream data is corrected, so the older data reflects the change. Scheduled runs only produce new
+partitions going forward, so past data has to be rewritten separately.
+
+You start a managed backfill by editing a `backfill.yaml` file in the table's directory and opening
+a pull request. Most of the process is automated, and you wait for Slack notifications between steps.
+The process is:
+
+1. Create an entry in the table's `backfill.yaml` (using `bqetl backfill create`) and open a PR. See [Initiating the backfill](#initiating-the-backfill).
```
scholtzan: I'd find it helpful to have an example config somewhere here. You can in theory piece the config together thought the "What can be backfilled with bqetl backfill" section, but I imagine having a common config to copy-paste + change some dates would be helpful.

scholtzan (approved):
r+wc

docs/cookbooks/backfilling_a_table.md
BenWu: I would have a default recommendation of using the interactive mode of `bqetl backfill create` since copy-pasting can miss some nuance and cause errors. That's why I made the interactive mode. What I can do is link an example of a basic one but recommend the CLI.

PR #9736, opened 2026-07-27
feat(jira): Add jira events integration to capture a changelogs as events

quiiver:
## Description

This PR adds the ability to create tables of Jira **change events** — one row per thing that happened to a ticket — scoped per table by Jira project and arbitrary JQL, fixed at the table's `query.py`. It adds the reusable integration plus the first table using it, `jira_events_derived.srein_events_v1`, covering the SREIN (MozCloud Intake) project.

The existing `jira_tickets_derived` tables are point-in-time snapshots (current status, current assignee). They can't answer how a ticket *moved* — time in each status, assignee churn, time to first response. These tables can.

### What's here

**Reusable integration** (`bigquery_etl/jira/`)
- `client.py` (new) — extracted from `issues.py`: credential loading, timestamp parsing, and the two Jira pagination styles (`nextPageToken` for search, `startAt` for changelog/comments). Adds retry with backoff honouring `Retry-After`, which `issues.py` didn't need at a handful of calls but a ~3,200-call seed certainly does. `issues.py` is refactored onto it with **no output change** — there's a regression test asserting its mapped output is unchanged.
- `events.py` (new) — event schema, changelog/comment → row mapping, fetch orchestration, and the BigQuery load.

**First table** — `sql/moz-fx-data-shared-prod/jira_events_derived/srein_events_v1/`, daily on the existing `bqetl_jira_tickets` DAG. No `dags.yaml` change needed.

### Event model

Jira has no single "events" endpoint, and the obvious event names don't map 1:1 onto Jira concepts:

| Concept | Jira source | `event_type` |
| --- | --- | --- |
| opened | issue's own `fields.created` + `reporter` (not in the changelog) | `created` |
| status change | changelog item where `field == "status"` | `status_change` |
| updated | **no such Jira event** — every update is one or more field changes | `field_change`, with `field` naming which |
| commented | `/issue/{key}/comment` (not in the changelog) | `commented` |

One changelog *history* with several *items* (one save changing status and assignee) becomes **separate rows** sharing `event_id` and `event_ts` — that's what makes it an event stream rather than a change-set log.

### Decisions worth a reviewer's attention

- **Comment bodies are stored, flattened from ADF to plain text.** API v3 returns them as an ADF document, so `bigquery_etl/jira/adf.py` reduces one to text for a single `comment_body` column. This costs nothing extra to fetch — the comment endpoint already returns bodies in the payload we download and has no `fields` parameter to exclude them, so we were previously discarding data we had already paid for. The table carries the same free text as the tickets themselves and is ACL'd to match, so a reader here could already read the tickets.
- **The ADF reduction is deliberately lossy and forgiving.** Formatting marks are dropped; nodes carrying text in `attrs` (mentions, emoji, status lozenges, smart links) contribute that attribute; embedded media contributes nothing; and an unrecognised node type is traversed for children then skipped, so a node type Jira adds later degrades to missing text rather than a failed run. Blank lines are dropped rather than preserved, because nested blocks each emit a break and blank lines track ADF nesting depth more than authorial intent. **One known loss:** a `link` mark's href is dropped, so linked text keeps its label but loses the URL. Validated against seven real SREIN-1603 comments, two of which are pinned as regression tests.
- **`Last Comment` and `Rank` changelog fields are excluded.** Jira writes a synthetic `Last Comment` entry on every comment, so keeping it would double-count the `commented` events — which carry a comment id and the body, unlike that entry. `Rank` is backlog-reordering noise. (Measured on SREIN-1548: 9 of 29 changelog items were `Last Comment`, 1 was `Rank`.)
- **Memory is O(1) in events.** Fetch is a generator and rows stream to a temp NDJSON file, so a seed over a large project doesn't materialise millions of dicts. This also makes the zero-event case a clean truncating load, where `load_table_from_json([])` may raise.
- **`(issue_key, event_id, field)` is *not* unique** — one save can emit several items naming the same field (three attachments → three `Attachment` items under one history id). Please don't add a uniqueness check on it; it would false-positive.

### Re-runs are safe, and the CLI is two modes

Every event this table records bumps the issue's `updated`: creation sets it, a changelog write is by definition a modification, and commenting updates it too (verified on live SREIN data — `updated` tracks the last comment to within ~2s). So the daily JQL bounds `updated` **only from below**, at `D-1` for timezone slack, and `events_on_date` filters precisely from there. A re-run of any date therefore selects a superset of the issues with events on that date and rebuilds the partition completely — clearing an old Airflow task is an ordinary repair, not a silent partial overwrite. Re-running a long-past date simply fetches more issues the further back it reaches.

That makes the CLI two modes, each WRITE_TRUNCATE-ing exactly what it fetched, with no modifiers:

| Flag | Target |
| --- | --- |
| `--date D` | one partition, rebuilt from a complete fetch |
| `--seed` | the base table, from full history — needed only for the initial load |

`_validate` is correspondingly small: the two modes are mutually exclusive, one is required, and `--date` must parse — all before any network or BigQuery call. There is no staleness guard, no write-disposition flags, no chunked seeding, and no `--jql` override, because the CLI can no longer express an unsafe combination. (Earlier revisions of this PR had all four; see the commit history for why each was removed rather than guarded.)

### New dataset justification

Per the reviewer checklist item on new derived datasets — this data is **not** already available:

- `jira_service_desk_syndicate` (Fivetran) has `issue_field_history`, but its ACL is scoped to `workgroup:service-desk/*`, the connector covers service-desk projects rather than arbitrary ones like SREIN, and it carries **no comment events**.
- It is deliberately a new dataset rather than an extension of `jira_tickets_derived`: that dataset holds issue *snapshots*, and mixing per-event streams into it would conflate two different grains. `jira_events_derived` will hold one events table per Jira project.

ACL mirrors `jira_tickets_derived` (`dataset_base_acl: derived`, `user_facing: false`, `roles/bigquery.dataViewer` for `workgroup:mozilla-confidential/data-viewers`).

### Testing

100 tests in `tests/jira/`, HTTP mocked with `responses` against payload shapes captured from real SREIN data. Coverage includes the retry/backoff paths (including transport-level failures, not just 5xx), both pagination styles, generator laziness (so the O(1) memory property can't silently regress), the changelog call count, both argument guards, 27 tests over the ADF flattener including two real SREIN comment bodies, and a schema-contract test binding `schema.yaml` and `metadata.yaml` to what `load_events` actually sends to BigQuery — the two schema declarations previously had nothing keeping them in sync.

### Deployment

After merge the table is created by artifact deployment, then history must be **seeded manually once** (~1,600 issues, ~3,200 API calls, 15–30 min). This is the only thing `--seed` is for; ongoing repair is just re-running a date.

```
PYTHONPATH=<repo root> python sql/moz-fx-data-shared-prod/jira_events_derived/srein_events_v1/query.py --seed
```

### Known open item

`CHANGELOG_ITEMS_KEY = "values"` follows Atlassian's documented `PageBeanChangelog` shape but **could not be verified against the live API** (no Jira credentials in the dev environment). A guard raises loudly if it's wrong, so the failure mode is a red task on the first issue rather than silently writing zero changelog events. Worth confirming before the first seed:

```
curl -su "$JIRA_USERNAME:$JIRA_TOKEN" \
  'https://mozilla-hub.atlassian.net/rest/api/3/issue/SREIN-1548/changelog?maxResults=1' | jq keys
```

* DENG-11416


whd:
Have we considered extending our existing use of https://fivetran.com/docs/connectors/applications/jira instead?

scholtzan:
### Integration report
- [View full diff, if available (SQL + DAGs)](https://github.com/mozilla/private-bigquery-etl/blob/bqetl-diff/pr-9736/sql.diff) *(requires private-bigquery-etl access)*
- [Private CI run](https://github.com/mozilla/private-bigquery-etl/actions/runs/30296244768)
- Bigeye monitoring dry-run: ✅ plan clean — [view plan](https://github.com/mozilla/private-bigquery-etl/blob/bqetl-bigeye-plan/bqetl-pr-9736/bigeye_plan.yml) *(requires private-bigquery-etl access)*

scholtzan:
### Integration report
- [View full diff, if available (SQL + DAGs)](https://github.com/mozilla/private-bigquery-etl/blob/bqetl-diff/pr-9736/sql.diff) *(requires private-bigquery-etl access)*
- [Private CI run](https://github.com/mozilla/private-bigquery-etl/actions/runs/30303504744)
- Bigeye monitoring dry-run: ✅ plan clean — [view plan](https://github.com/mozilla/private-bigquery-etl/blob/bqetl-bigeye-plan/bqetl-pr-9736/bigeye_plan.yml) *(requires private-bigquery-etl access)*

scholtzan:
### Integration report
- [View full diff, if available (SQL + DAGs)](https://github.com/mozilla/private-bigquery-etl/blob/bqetl-diff/pr-9736/sql.diff) *(requires private-bigquery-etl access)*
- [Private CI run](https://github.com/mozilla/private-bigquery-etl/actions/runs/30307917971)
- Bigeye monitoring dry-run: ✅ plan clean — [view plan](https://github.com/mozilla/private-bigquery-etl/blob/bqetl-bigeye-plan/bqetl-pr-9736/bigeye_plan.yml) *(requires private-bigquery-etl access)*

scholtzan:
### Integration report
- [View full diff, if available (SQL + DAGs)](https://github.com/mozilla/private-bigquery-etl/blob/bqetl-diff/pr-9736/sql.diff) *(requires private-bigquery-etl access)*
- [Private CI run](https://github.com/mozilla/private-bigquery-etl/actions/runs/30309376323)
- Bigeye monitoring dry-run: ✅ plan clean — [view plan](https://github.com/mozilla/private-bigquery-etl/blob/bqetl-bigeye-plan/bqetl-pr-9736/bigeye_plan.yml) *(requires private-bigquery-etl access)*

quiiver:
@whd and I had a chance to chat about fivetran and jira and while we both agree that using fivetran in a broader manner (i.e. sync all of the 'mozilla-confidential' jira projects) would be a more ideal 'platform feature', it would also be a much larger lift and would require more coordination. Landing this additional integration is relatively cheap and allows us to validate if having jira data in this manner would be helpful.

BenWu (approved):

